PR #63, opened 2019-09-04
Anchor api

### LINKED ISSUE #51: Implement GetByAnchorID with index
```
GetByAnchorID
Make sure that the “getByAnchorId” does not return anchor data after evict date has been reached.
This solves the issue that the bucket might be evicted already and the index is still existing. -> Then the evict has to assure that it clears the index one after another.

given anchor id = 𝐴𝐼𝐷
𝑡𝑒𝑣𝑖𝑐𝑡=𝐴𝑛𝑐ℎ𝑜𝑟𝐸𝑣𝑖𝑐𝑡𝐷𝑎𝑡𝑒𝑠.𝑔𝑒𝑡(𝐴𝐼𝐷) => O(1)
𝑎𝑛𝑐ℎ𝑜𝑟𝐷𝑎𝑡𝑎=𝑟𝑒𝑎𝑑_𝑐ℎ𝑖𝑙𝑑_𝑠𝑡𝑜𝑟𝑎𝑔𝑒(𝑡𝑒𝑣𝑖𝑐𝑡,𝐴𝐼𝐷) => 𝑂(𝑙𝑒𝑛(𝑐𝑜𝑛𝑐𝑎𝑡(𝑡𝑒𝑣𝑖𝑐𝑡,𝐴𝐼𝐷))) => 𝑂(1)
```

vimukthi-git:
closes #51


src/api.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// Anchor RPC methods.
+#[rpc]
+pub trait AnchorRpcApi {
+    /// Returns the next valid index (aka nonce) for given account.
+    #[rpc(name = "anchor_getAnchorById")]
+    fn get_anchor_by_id(&self, id: Hash) -> Result<AnchorData<Hash, BlockNumber>>;
+}
+
+pub struct Anchors<B, E, Block: BlockT, RA> {
```
vimukthi-git: a unit test for this stuff is on the way.

vimukthi-git: this is pretty complicated to set up the dependancies for the test :| . Since I have func tests update coming for this api, I will add this test later when I clarify how to add a proper test for this.

src/api.rs
```diff
@@ -0,0 +1,53 @@
...
+        Client<B, E, Block, RA>: ProvideRuntimeApi,
+        <Client<B, E, Block, RA> as ProvideRuntimeApi>::Api: AnchorApi<Block>
+{
+    fn get_anchor_by_id(&self, id: Hash) -> Result<AnchorData<Hash, BlockNumber>> {
+        let api = self.client.runtime_api();
+        let best = self.client.info().chain.best_hash;
+        let at = BlockId::hash(best);
+        api.get_anchor_by_id(&at, id).ok().unwrap().ok_or(Error {
+            code: ErrorCode::ServerError(RUNTIME_ERROR),
+            message: "Unable to find anchor ".into(),
```
pstehlik: remove trailing space?

runtime/src/lib.rs
```diff
@@ -309,7 +311,21 @@ pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, Call, SignedExt
 /// Executive: handles dispatch to the various modules.
 pub type Executive = executive::Executive<Runtime, Block, system::ChainContext<Runtime>, Runtime, AllModules>;
 
+decl_runtime_apis! {
+	/// The API to query anchoring info.
+	pub trait AnchorApi {
+		fn get_anchor_by_id(id: Hash) -> Option<AnchorData<Hash, BlockNumber>>;
```
pstehlik: I think there is a crucial piece missing here and that is for the caller to be able to find out until when the anchor is valid. Meaning the return would be better served by containing the expiration date as well.

vimukthi-git: @pstehlik Technically speaking the there is already a storage function RPC autogenerated for that because of the `AnchorEvictDates get(get_anchor_evict_date): map T::Hash => u32;` index in storage. That's why we are omitting it here. Also if the anchor is invalid it would have been evicted and this function would not return any data, other than an error.

PR #64, opened 2019-09-05
Anchor child storage

### LINKED ISSUE #49: Implement child tree based anchor buckets and commit extrinsic
```

// child storage for anchor data represented as a double map
AnchorStorage: Map EvictDate => AnchorID => AnchorData

#### Commit

Anchor data = $AnchorData$
Anchor ID = $AID$
Commit date = $t_{commit}$
User defined evict Date = $t_p$
Calculate User defined anchor period = $p_u$ = $t_p$ - $t_{commit}$
System defined minimum anchor period = $p_s$
Calculate actual anchor period = $p_{actual}$ = {$p_u$ < $p_s$ ? $p_s$ : $p_u$} (properly define case stmt :)
Calculate actual evict date = $t_{evict}$ = $t_{commit}$ + $p_{actual}$

**Store in child storage** -> $set\_child\_storage(t_{evict},  AID, AnchorData)$
**Store evict date index** -> $AnchorEvictDates.put(AID, t_{evict})$
**Calculate anchor index** = $AIDX = CurrentAnchorIndex + 1$
**Store anchor index** -> $AnchorIndexes.put(AIDX, AID)$
**Store current anchor index** -> $CurrentAnchorIndex = AIDX$

Total storage per anchor = $S_{anchor} = 2 * storage(AID) + storage(AnchorData) + storage(AIDX)$
Cost per storage unit per time unit = $c_{unit}$
Calculate anchor storage cost = $c_{storage} = S_{anchor} * c_{unit} * p_{actual}$  

Const gas cost (gas) = $c_{insert}$
Total tx cost = $c_{tx} = c_{storage} + c_{insert}$

complexity = $O(1)$ since trie inserts costs at most length of the anchor id string and its a fixed length hash.
```

vimukthi-git:
closes #49


runtime/src/anchor.rs
```diff
@@ -100,11 +119,20 @@ decl_module! {
 
 
             let block_num = <system::Module<T>>::block_number();
-            <Anchors<T>>::insert(anchor_id, AnchorData {
+            let child_storage_key = common::child_storage_key(storage_days);
```
vimukthi-git: the name `storage_days` is confusing here since it sounds relative to current date, will rename it to mean its the number of days from the epoch (non relative)

runtime/src/anchor.rs
```diff
@@ -459,12 +512,31 @@ mod tests {
...
+                doc_root,
+                <Test as system::Trait>::Hashing::hash_of(&0),
+                1567589844087
+            ));
+            a = Anchor::get_anchor_by_id(anchor_id2).unwrap();
+            assert_eq!(a.id, anchor_id2);
+            assert_eq!(a.doc_root, doc_root);
+            assert_eq!(Anchor::get_anchor_evict_date(anchor_id2), 18144);
+            assert_eq!(Anchor::get_anchor_id_by_index(2), anchor_id2);
+            assert_eq!(Anchor::get_anchor_id_by_index(Anchor::get_current_index()), anchor_id2);
```
pstehlik: if testing the index update, wouldn't it be better to pass the index `2` in here instead of get_current_index? Because what if the logic inside the code was broken and the index would always be stuck to 1? then the second get would still work but still be broken in terms of logic.

runtime/src/common.rs
```diff
@@ -0,0 +1,50 @@
...
+    let days = (ts / MS_PER_YEAR) as u32;
+    if remainder == 0 {
+        days
+    } else {
+        days + 1
+    }
+}
+
+/// create a child storage key from the given specific key
+pub fn child_storage_key(specific_key: u32) -> Vec<u8> {
```
pstehlik: would also call this `get_xxx` or `generate_xxx` to be clear what the method does.

runtime/src/anchor.rs
```diff
@@ -85,13 +97,20 @@ decl_module! {
...
+        pub fn commit(origin, anchor_id_preimage: T::Hash, doc_root: T::Hash, proof: T::Hash, eviction_date: T::Moment) -> Result {
             let who = ensure_signed(origin)?;
+            ensure!(<timestamp::Module<T>>::get() < eviction_date, "Not a valid storage date");
+
+            // validate the eviction date
+            let eviction_date_u64 = TryInto::<u64>::try_into(eviction_date)
+                .map_err(|_e| "Can not convert eviction date to u64")
+                .unwrap();
+            let storage_days = common::get_days_since_epoch(eviction_date_u64);
+            ensure!(Self::storage_max_days() >= storage_days, "Not a valid storage date");
```
vimukthi-git: make it clear that max days is dynamic and relative to today (max days from now)

runtime/src/common.rs
```diff
@@ -0,0 +1,50 @@
+use codec::{Encode};
+use rstd::vec::Vec;
+
+// default substrate child storage root
+const CHILD_STORAGE_DEFAULT_PREFIX: &[u8] = b":child_storage:default:";
+
+const MS_PER_YEAR: u64 = 86400000;
```
vimukthi-git: ms per day

runtime/src/anchor.rs
vimukthi-git: If I understand correctly the test case you suggest is the one above this line? :)

runtime/src/anchor.rs
```diff
@@ -50,6 +53,15 @@ decl_storage! {
         // Anchors store the map of anchor Id to the anchor
         Anchors get(get_anchor): map T::Hash => AnchorData<T::Hash, T::BlockNumber>;
 
+        // index to find the eviction date given and anchor id
```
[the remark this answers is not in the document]
vimukthi-git: is it not clear from the comment ?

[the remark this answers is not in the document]
vimukthi-git: oops i see now :)

PR #68, opened 2019-09-10
Pre-commit evict bucket edge case test

vimukthi-git:
There is a problem with `determine_pre_anchor_eviction_bucket` logic, since it allows eviction of pre-commits at `current_block + 1` when ,
`u32_current_block - (u32_current_block % expiration_horizon) + expiration_horizon < Self::expiration_duration_blocks()` (eg: every (n * 2400) - 1 block). This is a test case to identify the issue.

A solution might be to pass `pre_commit_expiry_block` to the `determine_pre_anchor_eviction_bucket` function and calculate eviction relative to that instead of the `current_block`.

### DIFF
--- a/runtime/src/anchor.rs
+++ b/runtime/src/anchor.rs
@@ -970,6 +970,46 @@ mod tests {
     }
 
     #[test]
+    fn pre_commit_at_4799_and_then_evict_before_expire_and_collaborator_succeed_commit() {
+        with_externalities(&mut new_test_ext(), || {
+            let pre_image = <Test as system::Trait>::Hashing::hash_of(&0);
+            let anchor_id = (pre_image).using_encoded(<Test as system::Trait>::Hashing::hash);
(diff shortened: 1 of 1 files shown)


runtime/src/anchor.rs
```diff
@@ -969,6 +969,46 @@ mod tests {
...
+                Origin::signed(1),
+                anchor_id,
+                signing_root
+            ));
+
+            assert_eq!(Anchor::determine_pre_anchor_eviction_bucket(start_block), 4800);
+            let a = Anchor::get_pre_anchor(anchor_id);
+            assert_eq!(a.expiration_block, expiry_block);
+            // the edge case bug - pre-commit eviction time is less than its expiry time
+            assert_eq!(Anchor::determine_pre_anchor_eviction_bucket(start_block) < a.expiration_block, true);
```
vimukthi-git: core of the problem

pstehlik:
I think we should fix this rather than just "documenting" the erronerous behavior via a test case, no?

PR #69, opened 2019-09-10
Adding detailed rust doc comments for anchor module

vimukthi-git:


runtime/src/anchor.rs
```diff
@@ -46,20 +53,20 @@ pub trait Trait: system::Trait + timestamp::Trait {}
 decl_storage! {
     trait Store for Module<T: Trait> as Anchor {
 
-        // Pre Anchors store the map of anchor Id to the pre anchor, which is a lock on an anchor id to be committed later
-        PreAnchors get(get_pre_anchor): map T::Hash => PreAnchorData<T::Hash, T::AccountId, T::BlockNumber>;
+        /// PreCommits store the map of anchor Id to the pre anchor, which is a lock on an anchor id to be committed later
```
pstehlik: Still using `pre anchor` here.
Also, consistency in wording would be great - it is a pre-commit. not PreCommit or pre commit.

vimukthi-git: here its referring to the name of the data structure, the way its intended as they are alway camel case.

pstehlik:
LGTM

vimukthi-git:
Thanks @pstehlik

PR #74, opened 2019-09-18
Amber testnet

vimukthi-git:


src/chain_spec.rs
```diff
@@ -20,8 +20,10 @@ pub enum Alternative {
     Development,
     /// Whatever the current runtime is, with simple Alice/Bob auths.
     LocalTestnet,
-    /// Fulvous testnet with whatever the current runtime is, with simple Alice/Bob auths and sudo account set by environment.
+    /// Fulvous testnet with whatever the current runtime is and with Alice/Bob as validators.
     Fulvous,
+    /// Amber testnet with whatever the current runtime is and persistent disks and with Alice/Bob as validators.
```
pstehlik: Alice and bob only for now (in this first iteration), right? In the persistent testnet (which we might also make public) we'd want to have non-standard keys set up.

vimukthi-git: yes thats the plan. I already discussed with @mikiquantum on how to execute an arbitrary RPC on after deployment to set validator keys.
Just that I want to test out the permanant storage setup first.

PR #86, opened 2019-09-30
add move_anchor dispatch

### LINKED ISSUE #84: Schedule for Commit
```json=
{
    "anchorID": currentVersion
}
```
- Validates that anchor exists in centchain
- Emits event to be processed by bridge operators with anchorID-docRoot

vedhavyas:
Once this is in, I will add the functional tests.

Closes #84

### DIFF
--- a/runtime/src/anchor/mod.rs
+++ b/runtime/src/anchor/mod.rs
@@ -57,8 +57,8 @@ decl_event! (
     where
         <T as system::Trait>::Hash,
     {
-        // MoveAnchor event is triggered when the anchor should be moved to a different chain.
-        // AnchorID and its DocRoot are sent as part of the event.
+        /// MoveAnchor event is triggered when the anchor should be moved to a different chain.
+        /// AnchorID and its DocRoot are sent as part of the event.
         MoveAnchor(Hash, Hash),
     }
 );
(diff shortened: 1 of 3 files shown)


runtime/src/anchor/mod.rs
```diff
@@ -91,6 +105,8 @@ decl_storage! {
 decl_module! {
     pub struct Module<T: Trait> for enum Call where origin: T::Origin {
 
+        fn deposit_event<T>() = default;
```
mikiquantum: What is the purpose of this? Apologize in advance for my lack of knowledge on RUST/Substrate

runtime/src/anchor/mod.rs
```diff
@@ -237,6 +253,17 @@ decl_module! {
 
             Ok(())
         }
+
+        // move_anchor dispatch call when anchor by anchor_id is to be moved to another chain
+        pub fn move_anchor(origin, anchor_id: T::Hash) -> Result {
+            // ensure signed origin
+            ensure_signed(origin)?;
```
mikiquantum: Is this origin the validator signature? or the client/requester signature?

vedhavyas:
> Since we run the bridges and they should take care of not doing actions twice (for $$ reasons). And we do not keep track of the moved anchors yet, the bridges should take care of what anchors have been processed already (for a given anchorID), so we do not end up in a DDOS attack and bridge funds are wiped.
Just to keep in mind, for the bridge design.

I have discussed these with @vimukthi-git as well. For the first version, I think we should be good. We anyway discard this as soon as we start implemeting bucket logic

runtime/src/anchor/mod.rs
vedhavyas: this is the module function that is used to send an event. `default` implies we are choosing to use the substrate's default event triggering logic

runtime/src/anchor/mod.rs
vedhavyas: This validates the sender of the dispatch call. Ethereum does it out of the box but in substrate different accounts can send a dispatch call. For this function `ensure-signed` is enough. For sudo users, we would be using a different fucntion. 

@vimukthi-git can shed more light on this maybe ?

vimukthi-git: This is basically the signer of the extrinsic, or the entity who originated the call. Ensure signed is used whenever we want to make sure that the extrinsic is signed. Signing account is the account which pays the transaction fees if any also.

mikiquantum: That means that the chain needs to know the public key to verify the signature. Is that a previous setup needed for each account? Or it is implicit like in Ethereum where public key is the address (simplifying)? Or the public key goes along the `origin` object?

vedhavyas: Its like ethereum i guess.

vimukthi-git: Yup its like ethereum, account addresses are public keys.

PR #92, opened 2019-10-10
add proof functions

### LINKED ISSUE #90: Validate sorted hashes proof function
- Create a Validate proof function that takes list of proofs and return a bool 
- Create a Function to return bundeldHash

vedhavyas:
Closes #90


runtime/src/proofs.rs
```diff
@@ -0,0 +1,49 @@
...
+    hash: Vec<u8>,
+    sorted_hashes: Vec<Vec<u8>>,
+}
+
+/// validates each proof and return true if all the proofs are valid
+/// else returns false
+pub fn validate_proofs(doc_root: &Vec<u8>, proofs: &Vec<Proof>) -> bool {
+    let mut res = false;
+    for proof in proofs.iter() {
+        res = res & validate_proof(doc_root, proof)
```
vimukthi-git: is this intended to be bitwise `and`?

runtime/src/proofs.rs
```diff
@@ -0,0 +1,48 @@
+use rstd::vec::Vec;
+
+#[derive(Debug)]
+pub struct Proof {
+    hash: Vec<u8>,
```
vimukthi-git: may be makes sense to use the in built `Hash` type here or make this fixed array

runtime/src/proofs.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        hash = runtime_io::blake2_256(hash.as_slice()).to_vec()
+    }
+
+    hash != *doc_root
+}
+
+pub fn bundled_hash(proofs: Vec<&Proof>) -> [u8; 32] {
+    let mut hash: Vec<u8> = Default::default();
+    for proof in proofs.iter() {
```
vimukthi-git: idiomatic way in rust is to use iterator operations on the stream(eg: https://github.com/centrifuge/centrifuge-chain/blob/master/runtime/src/anchor/mod.rs#L359) instead of `for` loop :)

runtime/src/proofs.rs
```diff
@@ -0,0 +1,48 @@
...
+            hash = sh;
+        } else {
+            // hash is less than sorted_hash, so append sorted hash to hash
+            hash.extend(sh);
+        }
+
+        hash = runtime_io::blake2_256(hash.as_slice()).to_vec()
+    }
+
+    hash != *doc_root
```
vimukthi-git: why `*doc_root`?

runtime/src/proofs.rs
vedhavyas: MB

runtime/src/proofs.rs
vedhavyas: You mean the runtime hash ?

runtime/src/proofs.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    false
+}
+
+// appends deposit_address and all the hashes from the proofs and returns keccak hash of the result.
+pub fn bundled_hash(proofs: Vec<Proof>, deposit_address: [u8; 20]) -> [u8; 32] {
+    let hash = proofs
+        .into_iter()
+        .fold(deposit_address.to_vec(), |mut acc, proof: Proof| {
```
mikiquantum: `deposit_addr+proofs` is the order? or `proofs+deposit_addr`?

runtime/src/proofs.rs
```diff
@@ -0,0 +1,282 @@
...
+            return true;
+        }
+        matches.push(hash.clone())
+    }
+
+    false
+}
+
+// appends deposit_address and all the hashes from the proofs and returns keccak hash of the result.
+pub fn bundled_hash(proofs: Vec<Proof>, deposit_address: [u8; 20]) -> [u8; 32] {
```
mikiquantum: Thanks for separating this function!

runtime/src/proofs.rs
vedhavyas: Deposit_address+Proofs as per your comment on the Asset contract PR

runtime/src/proofs.rs
```diff
@@ -0,0 +1,282 @@
...
+    for proof in proofs.into_iter() {
+        matches.push(proof);
+        hash = hash_of(hash, proof);
+        if matches.contains(&hash) {
+            return true;
+        }
+        matches.push(hash.clone())
+    }
+
+    false
```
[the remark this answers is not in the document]
vedhavyas: Yeah, we can add return but infer return is added here

runtime/src/proofs.rs
```diff
@@ -0,0 +1,282 @@
...
+    let mut h: Vec<u8> = Vec::with_capacity(64);
+    if a < b {
+        h.extend_from_slice(&a);
+        h.extend_from_slice(&b);
+    } else {
+        h.extend_from_slice(&b);
+        h.extend_from_slice(&a);
+    }
+
+    runtime_io::blake2_256(&h)
```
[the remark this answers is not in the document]
vedhavyas: For proof verification, we are going with Blake2b. If we decide to change this, this is the place to change it

runtime/src/proofs.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    let mut matches = Vec::new();
+    matches.push(doc_root);
+    return proofs
+        .iter()
+        .map(|proof| {
+            validate_proof(
+                &mut matches,
+                proof.hash.clone(),
```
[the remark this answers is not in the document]
vedhavyas: If it doesn't implemeent copy trait, it will be moved. Hence we clone it. But for fixed size array, we do not need to since it implements the copy trait

runtime/src/proofs.rs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+fn validate_proof(matches: &mut Vec<[u8; 32]>, hash: [u8; 32], proofs: Vec<[u8; 32]>) -> bool {
+    // if hash is already cached earlier
+    if matches.contains(&hash) {
+        return true;
+    }
+
+    let mut hash = hash;
+    for proof in proofs.into_iter() {
```
vimukthi-git: this code might be better optimised if we have a mapping of `match => proof` to begin with. Right now its a bit hard to understand whats going on. May be better naming would also help.

vedhavyas: We do not have a `mapping` for runtime, hence I went with Vec. OPen for other suggestions as well

PR #93, opened 2019-10-15
Integrate staking module (WIP)

philipstanislaus:


src/chain_spec.rs
```diff
@@ -43,18 +44,20 @@ fn get_from_pubkey_hex<TPublic: Public>(pubkey_hex: &str) -> <TPublic::Pair as P
...
     )
 }
 
 /// Helper function to obtain grandpa and babe keys from pubkey strings
-pub fn get_authority_keys_from_pubkey_hex(grandpa: &str, babe: &str) -> (GrandpaId, BabeId) {
+pub fn get_authority_keys_from_pubkey_hex(grandpa: &str, babe: &str) -> (GrandpaId, BabeId, ImOnlineId) {
     (
         get_from_pubkey_hex::<GrandpaId>(grandpa),
         get_from_pubkey_hex::<BabeId>(babe),
+        get_from_pubkey_hex::<ImOnlineId>(seed),
```
vimukthi-git: pass the pubkey string

runtime/src/lib.rs
```diff
@@ -258,11 +269,154 @@ impl balances::Trait for Runtime {
...
+impl authorship::Trait for Runtime {
+	type FindAuthor = session::FindAccountFromAuthorIndex<Self, Babe>;
+	type UncleGenerations = UncleGenerations;
+	type FilterUncle = ();
+	type EventHandler = Staking;
+}
+
+type SessionHandlers = (Grandpa, Babe, ImOnline, AuthorityDiscovery);
+
+impl_opaque_keys! {
```
vimukthi-git: session keys struct is already implemented in our runtime you can just add the new one that is `im_online`

runtime/src/lib.rs
```diff
@@ -282,9 +436,17 @@ construct_runtime!(
 		Timestamp: timestamp::{Module, Call, Storage, Inherent},
 		Babe: babe::{Module, Call, Storage, Config, Inherent(Timestamp)},
 		Grandpa: grandpa::{Module, Call, Storage, Config, Event},
+		Authorship: authorship::{Module, Call, Storage, Inherent},
 		Indices: indices::{default, Config<T>},
 		Balances: balances,
+		Session: session::{Module, Call, Storage, Event, Config<T>},
+		Staking: staking::{default, OfflineWorker},
+		Council: collective::<Instance1>::{Module, Call, Storage, Origin<T>, Event<T>, Config<T>},
+		Treasury: treasury::{Module, Call, Storage, Event<T>},
```
vimukthi-git: It would be nicer I think if we could avoid complexity of having a treasury and counsil for atleast the POC.

docker-compose.yml
```diff
@@ -1,76 +1,98 @@
...
-      restart_policy:
-        condition: on-failure
+    # deploy:
+    #   replicas: 1
+    #   resources:
+    #     limits:
+    #       cpus: "1"
+    #       memory: 2000M
+    #   restart_policy:
+    #     condition: on-failure
```
vedhavyas: Is comment expected ?

runtime/Cargo.toml
```diff
@@ -122,11 +182,36 @@ features = ['derive']
...
+
+[dependencies.srml-staking-reward-curve]
+default_features = false
+git = 'https://github.com/paritytech/substrate.git'
+package = 'srml-staking-reward-curve'
+rev = '67b73c468e44839c5809245703d067dc75940076'
+
+[dependencies.staking]
+default_features = false
+git = 'https://github.com/paritytech/substrate.git'
```
vedhavyas: Is the reward curve package a hard dependency for staking module ? Asking because you suggested we deprioritise this

src/chain_spec.rs
```diff
@@ -212,24 +241,56 @@ fn testnet_genesis(
...
+        session: Some(SessionConfig {
+			keys: initial_authorities.iter().map(|x| {
+				(x.0.clone(), session_keys(x.2.clone(), x.3.clone(), x.4.clone()))
+			}).collect::<Vec<_>>(),
+		}),
+		staking: Some(StakingConfig {
+            // The current era index.
+			current_era: 0,
+            // Minimum number of staking participants before emergency conditions are imposed.
+			minimum_validator_count: 1,
```
vedhavyas: Is the minimum number we have decided upon ?

runtime/src/lib.rs
philipstanislaus: Changed.

docker-compose.yml
philipstanislaus: No, removed.

src/chain_spec.rs
philipstanislaus: No, but this parameter has been added to the parameter discussion spreadsheet: https://docs.google.com/spreadsheets/d/1eGc-jEMWLs5qzgYRDPo7VnZgJgWwRn_GjK4v1scuVCg/edit#gid=0

runtime/Cargo.toml
philipstanislaus: Good point – it is not, but there is no harm using the default one. I think we should deprioritize implementing our own custom reward curve – the current thinking is around exponentially falling rewards over time. That is much more work to implement than just using the existing one.

runtime/src/lib.rs
philipstanislaus: Treasury is needed to collect fees/rewards and slashed tokens. 

I removed council for now, but we should add it at some point, because right now we rely on the root account for access to the treasury. Added this as an issue: https://github.com/centrifuge/centrifuge-chain/pull/93/commits/ec57451a687b1f844927b5bf5d1351bcc92c75a0#diff-11f21b7f145f98241270cf825e0aaab5R463-R464

PR #110, opened 2019-10-30
rename Amber => Flint

vimukthi-git:


vimukthi-git:
@mikiquantum yes. Most work is to adjust deployed things.

PR #119, opened 2019-11-17
Substrate Upgrade

philipstanislaus:


runtime/src/anchor/mod.rs
```diff
@@ -204,7 +204,7 @@ decl_module! {
             };
 
             let anchor_data_encoded = anchor_data.encode();
-            runtime_io::set_child_storage(&child_storage_key, anchor_id.as_ref(), &anchor_data_encoded);
+            runtime_io::storage::child_set(&child_storage_key, anchor_id.as_ref(), &anchor_data_encoded);
```
mikiquantum: does this mean that the newer version has changed the file/package structure?

scripts/init.sh
```diff
@@ -5,12 +5,8 @@ set -e
...
 if [ -z $CI_PROJECT_NAME ] ; then
-   #rustup update nightly
+   rustup update nightly
    rustup update stable
 fi
 
-rustup target add wasm32-unknown-unknown --toolchain $RUST_TOOLCHAIN
-
-# Install wasm-gc. It's useful for stripping slimming down wasm binaries.
-command -v wasm-gc || \
```
mikiquantum: I noticed that when I first built the project I needed this wasm-gc , is there any reason why we are removing it?

src/service.rs
```diff
@@ -89,46 +75,64 @@ macro_rules! new_full_start {
             )?;
 
             import_setup = Some((babe_block_import, grandpa_link, babe_link));
-
             Ok(import_queue)
+        // })?
```
mikiquantum: Is there any reason why we comment this? Is it just to isolate the problem first?

runtime/src/lib.rs
```diff
@@ -8,39 +8,35 @@
...
 
-use babe::AuthorityId as BabeId;
-use client::{
-    block_builder::api::{self as block_builder_api, CheckInherentsResult, InherentData},
-    decl_runtime_apis, impl_runtime_apis, runtime_api as client_api,
-};
-use grandpa::fg_primitives;
-use grandpa::{AuthorityId as GrandpaId, AuthorityWeight as GrandpaWeight};
-use primitives::{crypto::key_types, OpaqueMetadata};
+use primitives::OpaqueMetadata;
```
mikiquantum: Is this our implementation of the node-template file updated to the latest released?

runtime/src/anchor/mod.rs
philipstanislaus: Exactly 👍

runtime/src/lib.rs
philipstanislaus: Not sure I understand your question, but yes, I did follow the changes in the node-template file. We are using a bunch more than the default template though.

src/service.rs
philipstanislaus: Exactly, I commented it out to fix compilation issues, has been re-added now.

scripts/init.sh
philipstanislaus: True, the substrate team changed that requirement, the wasm optimization is happening within the wasm build stage now.

PR #128, opened 2019-12-09
run rust fmt

vedhavyas:
Just ran rustfmt on the project. Maybe we should enforce this like we do it for Go


philipstanislaus:
Closing this for now, since it will complicate upgrading substrate.

PR #132, opened 2019-12-16
update nft proof validations to support static proofs

vedhavyas:
Update NFT proof validations with static tree structure

### DIFF
--- a/runtime/src/proofs.rs
+++ b/runtime/src/proofs.rs
@@ -93,12 +93,12 @@ fn validate_proof(matches: &mut Vec<[u8; 32]>, hash: [u8; 32], proofs: Vec<[u8; 
 // prefill the required matches before proof validation
 fn pre_matches(static_proofs: [[u8; 32]; 3], doc_root: [u8; 32]) -> (bool, Vec<[u8; 32]>) {
     let mut matches = Vec::new();
-    let data_root = static_proofs[0];
-    let sibling_root = static_proofs[1];
+    let basic_data_root = static_proofs[0];
+    let zk_data_root = static_proofs[1];
     let signature_root = static_proofs[2];
-    matches.push(data_root);
-    matches.push(sibling_root);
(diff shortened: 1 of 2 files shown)


runtime/src/proofs.rs
```diff
@@ -66,12 +90,29 @@ fn validate_proof(matches: &mut Vec<[u8; 32]>, hash: [u8; 32], proofs: Vec<[u8;
     false
 }
 
+// prefill the required matches before proof validation
+fn pre_matches(static_proofs: [[u8; 32]; 3], doc_root: [u8; 32]) -> (bool, Vec<[u8; 32]>) {
+    let mut matches = Vec::new();
+    let data_root = static_proofs[0];
```
mikiquantum: Since the inner signing tree is static as well, I would rename:
- `data_Root` to `basic_data_root` or `left_data_root`
- `sibling_root` to `zk_data_root` or `right_data_root`
WDYT?

runtime/src/proofs.rs
```diff
@@ -66,12 +90,29 @@ fn validate_proof(matches: &mut Vec<[u8; 32]>, hash: [u8; 32], proofs: Vec<[u8;
     false
 }
 
+// prefill the required matches before proof validation
+fn pre_matches(static_proofs: [[u8; 32]; 3], doc_root: [u8; 32]) -> (bool, Vec<[u8; 32]>) {
```
mikiquantum: Is the `static_proofs: [[u8; 32]; 3]` length check implicit here, or do we need to add an extra check on required length of 3, to avoid out of bounds exception?

vedhavyas: its implicit here. It should only have 3 and not more
If any missing, then the hash check fails

mikiquantum: Cool, because it is initialized to 0s to the ones that are not provided, right?

PR #134, opened 2019-12-18
Upgrade

philipstanislaus:


runtime/src/constants.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// <https://research.web3.foundation/en/latest/polkadot/BABE/Babe/#6-practical-results>
+    pub const MILLISECS_PER_BLOCK: Moment = 6000;
+    pub const SECS_PER_BLOCK: Moment = MILLISECS_PER_BLOCK / 1000;
+
+    pub const SLOT_DURATION: Moment = MILLISECS_PER_BLOCK;
+
+    // 1 in 4 blocks (on average, not counting collisions) will be primary BABE blocks.
+    pub const PRIMARY_PROBABILITY: (u64, u64) = (1, 4);
+
+    pub const EPOCH_DURATION_IN_BLOCKS: BlockNumber = 30; // 10 * MINUTES;
```
mikiquantum: Is the 10*MINUTES comment still relevant?

philipstanislaus: Good catch, will remove

PR #144, opened 2020-01-08
Add council and democracy module

philipstanislaus:


runtime/src/lib.rs
```diff
@@ -302,11 +302,80 @@ impl pallet_staking::Trait for Runtime {
...
 }
 
+parameter_types! {
+	pub const LaunchPeriod: BlockNumber = 28 * 24 * 60 * MINUTES;
+	pub const VotingPeriod: BlockNumber = 28 * 24 * 60 * MINUTES;
+	pub const EmergencyVotingPeriod: BlockNumber = 3 * 24 * 60 * MINUTES;
+	pub const MinimumDeposit: Balance = 100 * 1_000_000_000_000_000_000;
+	pub const EnactmentPeriod: BlockNumber = 30 * 24 * 60 * MINUTES;
+	pub const CooloffPeriod: BlockNumber = 28 * 24 * 60 * MINUTES;
+	pub const PreimageByteDeposit: Balance = 1 * 10_000_000_000_000_000;
```
vedhavyas: do we have these parameters documented somewhere to look at?

src/chain_spec.rs
```diff
@@ -134,20 +129,7 @@ impl Alternative {
                             get_authority_keys_from_seed("Bob"),
                         ],
                         get_account_id_from_seed::<sr25519::Public>("Alice"),
-                        vec![
```
vedhavyas: Did we just moved them around or something changed here ?

runtime/src/lib.rs
```diff
@@ -593,3 +665,57 @@ impl_runtime_apis! {
 		}
 	}
 }
+
+#[cfg(test)]
+mod tests {
```
vedhavyas: are these tests part of the Council and Democracy?

src/chain_spec.rs
philipstanislaus: Yes, we have a spreadsheet, which is pinned to our #crcl_flim on Slack.

runtime/src/lib.rs
philipstanislaus: Yes, we have a spreadsheet, which is pinned to our #crcl_flim on Slack.

src/chain_spec.rs
philipstanislaus: These were moved here: https://github.com/centrifuge/centrifuge-chain/pull/144/files/3710754192fe895eb86de34533692a05e2511cec#diff-6740deafebf0ba9153f241785add2eefR240

runtime/src/lib.rs
philipstanislaus: No, but they were to small for their own PR. I just copied them over from substrate.

philipstanislaus:
@vedhavyas want to approve this?

PR #157, opened 2020-01-21
Require fee change origin to be defined in trait, set it to simple ma…

philipstanislaus:
…jority of council

### DIFF
--- a/runtime/src/anchor/tests.rs
+++ b/runtime/src/anchor/tests.rs
@@ -51,6 +51,7 @@ impl pallet_timestamp::Trait for Test {
 
 impl fees::Trait for Test {
     type Event = ();
+    type FeeChangeOrigin = frame_system::EnsureRoot<u64>;
 }
 
 parameter_types! {
--- a/runtime/src/fees.rs
+++ b/runtime/src/fees.rs
@@ -9,12 +9,15 @@ use frame_support::{
     traits::{Currency, ExistenceRequirement, WithdrawReason},
     weights::SimpleDispatchInfo,
 };
-use frame_system::{self as system, ensure_signed};
+use frame_system::{self as system, ensure_root};
+use sp_runtime::traits::EnsureOrigin;
 
 /// The module's configuration trait.
 pub trait Trait: frame_system::Trait + pallet_balances::Trait {
     /// The overarching event type.
     type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
+    /// Required origin for changing fees
+	type FeeChangeOrigin: EnsureOrigin<Self::Origin>;
 }
 
 #[derive(Encode, Decode, Default, Clone, PartialEq)]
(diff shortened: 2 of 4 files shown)


runtime/src/anchor/tests.rs
```diff
@@ -51,6 +51,7 @@ impl pallet_timestamp::Trait for Test {
 
 impl fees::Trait for Test {
     type Event = ();
+    type FeeChangeOrigin = frame_system::EnsureRoot<u64>;
```
pstehlik: I am confused as to why Root would work here - or is the council automatically treated as root?

pstehlik: yeah... brainfart.... this is the test file. all good.

runtime/src/fees.rs
```diff
@@ -9,12 +9,15 @@ use frame_support::{
...
-use frame_system::{self as system, ensure_signed};
+use frame_system::{self as system, ensure_root};
+use sp_runtime::traits::EnsureOrigin;
 
 /// The module's configuration trait.
 pub trait Trait: frame_system::Trait + pallet_balances::Trait {
     /// The overarching event type.
     type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
+    /// Required origin for changing fees
+	type FeeChangeOrigin: EnsureOrigin<Self::Origin>;
```
pstehlik: so what's the allowed origin? sorry - not super clear here for me as i am reviewing in-between things. maybe a quick call or skip my review.

philipstanislaus: All good, great question. 

This just defines the `FeeChangeOrigin` type to be needed and implement the `EnsureOrigin` trait. Further down in the code, `T::FeeChangeOrigin::try_origin(origin)` is called, which is verifying that the `origin` of the extrinsic satisfied the FeeChangeOrigin. 

In `lib.rs`, we define `type FeeChangeOrigin = pallet_collective::EnsureProportionAtLeast<_1, _2, AccountId, CouncilCollective>;`, where `pallet_collective::EnsureProportionAtLeast<_1, _2, AccountId, CouncilCollective>;` implements the `EnsureOrigin` trait. 

Hope that makes sense – if not, please call me :)

philipstanislaus:
@vedhavyas any thoughts or can I merge this?

runtime/src/fees.rs
```diff
@@ -111,9 +113,11 @@ impl<T: Trait> Module<T> {
         }
     }
 
-    fn can_change_fee(_who: T::AccountId) -> DispatchResult {
-        //TODO add auth who can change fees
-        //        ensure!(<validatorset::Module<T>>::is_validator(who), "Not authorized to change fees.");
+    fn can_change_fee(origin: T::Origin) -> DispatchResult {
+        T::FeeChangeOrigin::try_origin(origin)
+            .map(|_| ())
+            .or_else(ensure_root)?;
```
vedhavyas: we fallback here to check if the fee is changed by the root ?

runtime/src/fees.rs
```diff
@@ -180,8 +186,12 @@ mod tests {
         type Version = ();
         type ModuleToIndex = ();
     }
+    parameter_types! {
+        pub const One: u64 = 1;
+    }
     impl Trait for Test {
         type Event = ();
+        type FeeChangeOrigin = EnsureSignedBy<One, u64>;
```
vedhavyas: nice

runtime/src/fees.rs
philipstanislaus: Exactly. If `try_origin` fails, it will return an `Err(OuterOrigin)` result. The `or_else` will unwrap the `OuterOrigin` and use it as an argument to `ensure_root`.

PR #159, opened 2020-01-29
Amber genesis

philipstanislaus:


runtime/src/lib.rs
```diff
@@ -208,7 +208,7 @@ parameter_types! {
     pub const TransactionBaseFee: Balance = 100 * MICRO_RAD;
     pub const TransactionByteFee: Balance = 1 * MICRO_RAD;
 	// setting this to zero will disable the weight fee.
-	pub const WeightFeeCoefficient: Balance = 1_000;
+	pub const WeightFeeCoefficient: Balance = 100_000_000;
```
vedhavyas: we missed this as part of the fee PR?

src/chain_spec.rs
```diff
@@ -42,6 +39,55 @@ pub type ChainSpec = sc_service::ChainSpec<
...
+		})
+	}
+
+	pub(crate) fn from(s: &str) -> Option<Self> {
+		match s {
+			"dev" => Some(Alternative::Development),
+			"local" => Some(Alternative::LocalTestnet),
+			"fulvous" => Some(Alternative::Fulvous),
+			"flint" => Some(Alternative::Flint),
+			"" | "amber" => Some(Alternative::Amber),
```
vedhavyas: I think being explicit for amber would be good. Having an empty `""` would be good chioce for mainnet Chainspc

runtime/src/lib.rs
philipstanislaus: No, that is a different fee component. The fee PR only changed the authorization check of changes to state rent. This here is adjusting how much RAD each weight unit of an extrinsic will cost.

src/chain_spec.rs
philipstanislaus: Agree, will change.

README.md
```diff
@@ -27,84 +27,91 @@ cargo build --release
 
 ## Run
 
+### Testnets
+
+Centrifuge has multiple testnets online.
+
+1. Fulvous is an ephemeral testnet that is tracking `master`.
```
mikiquantum: I would change this to: `Fulvous` is an ephemeral testnet for internal dev purposes, and testing internal integrations with all the centrifuge components. Not recommended for external usage due to its nature and purpose.

src/chain_spec.rs
```diff
@@ -343,6 +256,93 @@ fn get_default_properties(token: &str) -> sc_service::Properties {
...
+		"fulvous",
+		fulvous_genesis,
+		vec![],
+		None,
+		Some("flvs"),
+		Some(get_default_properties("TRAD")),
+		Default::default(),
+	)
+}
+
```
mikiquantum: Are you planning on adding flint and amber here as well? Or how are we generating the genesis config file from code for them?

README.md
```diff
@@ -27,84 +27,91 @@ cargo build --release
...
+  --bootnodes /ip4/35.198.166.26/tcp/30333/p2p/QmNpeu3bJhESzriWMLRcxRgSCYDGQ6GdBHnJAf8bJexAd5
+```
+
+#### 2. Flint
+
+To run a node:
+
+```bash
+./target/release/centrifuge-chain \
+  --chain=flint \
```
mikiquantum: Do we need to ensure here the usage of the specific genesis version, instead of latest. 
`--chain=flint-cc2-spec.json`

mikiquantum:
@philipstanislaus Cancelled build since it was stuck running for more than 5 hours. A compilation error was triggered though.

src/chain_spec.rs
philipstanislaus: No, I'm following the pattern of Polkadot here. This gives us one source of truth for each testnet, instead of 2 (code + generated genesis).

README.md
philipstanislaus: No, `--chain=flint` will resolve to the latest file, since that's loaded in `chain_spec.rs`

README.md
philipstanislaus: I like that, changed.

philipstanislaus:
> Looks good overall, some small comments.
> Out of curiosity, if we are not defining amber as part of the chain-spec.rs? How did you generate the genesis file and with which params?

Copied from flint. This is the pattern Polkadot uses for Kusama, see https://github.com/paritytech/polkadot/blob/master/cli/src/chain_spec.rs

mikiquantum:
@philipstanislaus thanks for the comments. Lets say that we need to make changes in Amber genesis file, What is the best way yo know the endowed accounts, initial validator set ... so we can create a new version?

philipstanislaus:
> @philipstanislaus thanks for the comments. Lets say that we need to make changes in Amber genesis file, What is the best way yo know the endowed accounts, initial validator set ... so we can create a new version?

Not sure I follow - the idea is to just edit in the json file directly. Do you see any issue with that?

mikiquantum:
@philipstanislaus that answers my question! I meant that, thanks!

PR #169, opened 2020-02-06
Clean up for audit

philipstanislaus:


philipstanislaus:
@vedhavyas do we still need move_anchor? There's no functionality currently attached to that method, other than dispatching an event.

vedhavyas:
I'm removing it as part of my PR @philipstanislaus

PR #170, opened 2020-02-06
update proofs comments and remove move_anchor

vedhavyas:


runtime/src/anchor/mod.rs
```diff
@@ -210,18 +206,6 @@ decl_module! {
...
-            let today_in_days_from_epoch = TryInto::<u64>::try_into(<pallet_timestamp::Module<T>>::get())
-                .map(common::get_days_since_epoch)
-                .map_err(|_e| "Can not convert timestamp to u64")
-                .unwrap();
-
-            // we use the fee config setup on genesis for anchoring to calculate the state rent
-            let fee = <fees::Module<T>>::price_of(Self::fee_key()).unwrap() *
-                <T as pallet_balances::Trait>::Balance::from(stored_until_date_from_epoch - today_in_days_from_epoch);
-            <fees::Module<T>>::pay_fee_given(who, fee)?;
-
```
philipstanislaus: Good idea to move this up, it's a condition that should be filled before we actually commit. 👍

runtime/src/proofs.rs
```diff
@@ -52,6 +49,9 @@ pub fn validate_proofs(doc_root: H256, proofs: &Vec<Proof>, static_proofs: [H256
         .fold(true, |acc, b| acc && b);
 }
 
+// computes blake2 256 sorted hash of the a and b
+// if a < b: blake256(a+b)
+// else: blake256(b+a)
```
philipstanislaus: @vedhavyas do we still need this? Have not really been involved in this, but I remember vaguely that hashes are now always sorted?

vedhavyas: yeah, we still need this. only the signing_root, signature_root, and the two data roots are sorted. Rest all profs which lead up to these above rots are calculated through sorted hashes

vedhavyas: An Example of this would be proof for a field in Invoice document. The proofs that come with this are to be sorted to lead up to Data rot

philipstanislaus:
Makes sense, thanks for the clarification!

PR #171, opened 2020-02-12
Format some parts for better consistency and readability

c410-f3r:
It is common to configure Rustfmt on CI to have an homogeneous
formatting or at least run `cargo fmt` once in a while. This PR does the
following regarding this area:

- Use 4 spaces indentation instead of \t
- Join multiple lines into a single line
- Remove double \n (Empty lines)
- Add newline to the end of a file


philipstanislaus:
Thanks Caio, but we do not apply fmt by purpose to stay as close to upstream as possible. Happy to apply this once it has been applied to paritytech/substrate.

PR #182, opened 2020-02-21
Upgrade 2

philipstanislaus:


runtime/src/lib.rs
```diff
@@ -72,11 +76,11 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
...
     authoring_version: 10,
     // Per convention: if the runtime behavior changes, increment spec_version
-    // and set impl_version to equal spec_version. If only runtime
+    // and set impl_version to 0. If only runtime
     // implementation changes and behavior does not, then leave spec_version as
     // is and increment impl_version.
-    spec_version: 198,
-    impl_version: 198,
+    spec_version: 221,
+    impl_version: 0,
```
vedhavyas: why the change here ?

philipstanislaus: Yes, it changed in Kusama from `impl_version` starting at `spec_version` to `impl_version` starting at 0.

PR #187, opened 2020-03-02
Add mainnet preview chain spec

philipstanislaus:


runtime/src/lib.rs
```diff
@@ -387,7 +387,7 @@ impl pallet_collective::Trait<CouncilCollective> for Runtime {
 parameter_types! {
 	pub const CandidacyBond: Balance = 1000 * RAD;
 	pub const VotingBond: Balance = 50 * CENTI_RAD;
-	pub const TermDuration: BlockNumber = 7 * DAYS;
+	pub const TermDuration: BlockNumber = 1 * DAYS;
```
mikiquantum: What is the impact of this change?

philipstanislaus: @mikiquantum we discussed this with Martin and Cassidy – the idea is to have very short times on soft launch, and change them back up to 4 days for main launch through an upgrade.

PR #190, opened 2020-03-10
setup github actions

ivanceras:


ivanceras:
I was able to merge the PR. Perhaps, we need to setup PR needs to be reviewed first.

PR #197, opened 2020-03-16
transfer anchor state rent to block author

vedhavyas:
closes centrifuge/roadmap#130

This PR ensures the state rent is paid to block author.

`pallet_authorship::author()` returns either the primary or secondary author of the current block this is being included in. 
More on this is found here - https://github.com/paritytech/substrate/blob/master/frame/babe/src/lib.rs#L207

I have updated the `fees` module to take the `to` account to which the fees will be transferred to. Even though we use this for anchor state rent, it would be more reusable to take `from` and `to` rather than `fees` module implementing `pallet_authorship`

### DIFF
--- a/runtime/src/fees.rs
+++ b/runtime/src/fees.rs
@@ -4,7 +4,7 @@ use frame_support::{
     decl_event, decl_module, decl_storage,
     dispatch::DispatchResult,
     ensure,
-    traits::{Currency, ExistenceRequirement, WithdrawReason},
+    traits::{Currency, ExistenceRequirement},
     weights::SimpleDispatchInfo,
 };
 use frame_system::{self as system, ensure_root};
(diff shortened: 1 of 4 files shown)


runtime/src/fees.rs
```diff
@@ -71,21 +71,25 @@ decl_module! {
...
     /// Pay the given fee
-    pub fn pay_fee_given(who: T::AccountId, fee: T::Balance) -> DispatchResult {
-        let _ = <pallet_balances::Module<T> as Currency<_>>::withdraw(
-            &who,
+    pub fn pay_fee_given_to(
+        from: T::AccountId,
+        to: T::AccountId,
+        fee: T::Balance,
+    ) -> DispatchResult {
+        let _ = <pallet_balances::Module<T> as Currency<_>>::transfer(
```
philipstanislaus: Is there a better way than using a transfer for fees? The transfer looks in Polkascan like the anchorer did intentionally transfer to another party, while in reality the intent was to just pay fees. 

Maybe withdraw and deposit instead?

vedhavyas: Wouldn't withdraw and deposit appear in the polkscan?

vedhavyas:
> I feel that the fee module should always pay the fee to the block author, never to anyone else. That would simplify the API and remove the requirement of the anchor module to depend on the authorship module. What do you think?

I do not have a strong opinion about this

Issue #179, opened 2020-02-19
Integrate multi account pallet


philipstanislaus:
https://github.com/paritytech/substrate/issues/4834

philipstanislaus:
- [ ] Add tests for genesis 
- [ ] Create repository
- [ ] Add to our chain

PR #203, opened 2020-03-17
Integrate multi-account pallet

ivanceras:
This change is [Reviewable](https://reviewable.io/reviews/centrifuge/centrifuge-chain/203)

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -499,7 +499,7 @@ impl nfts::Trait for Runtime {
 }
 
 parameter_types! {
-	pub const MultiAcountSigDepositBase: Balance = 30 * CENTI_RAD;
+	pub const MultiAccountSigDepositBase: Balance = 30 * CENTI_RAD;
 	pub const MultiAccountDepositBase: Balance = 30 * CENTI_RAD;
 	pub const MultiAccountSigDepositFactor: Balance = 5 * CENTI_RAD;
 	pub const MultiAccountDepositFactor: Balance = 5 * CENTI_RAD;
@@ -512,7 +512,7 @@ impl substrate_pallet_multi_account::Trait for Runtime {
     type MaxSignatories = MultiAccountMaxSignatories;
     type MultiAccountDepositBase = MultiAccountDepositBase;
     type MultiAccountDepositFactor =  MultiAccountDepositFactor;
-    type MultisigDepositBase = MultiAcountSigDepositBase;
+    type MultisigDepositBase = MultiAccountSigDepositBase;
     type MultisigDepositFactor = MultiAccountSigDepositFactor;
 }
 
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -155,6 +155,7 @@ parameter_types! {
 	pub const MaxSignatories: u16 = 100;
 }
 
+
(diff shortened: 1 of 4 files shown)


ivanceras:
resolves #179

runtime/src/lib.rs
```diff
@@ -524,6 +543,7 @@ construct_runtime!(
 		Anchor: anchor::{Module, Call, Storage},
 		Fees: fees::{Module, Call, Storage, Event<T>, Config<T>},
 		Nfts: nfts::{Module, Call, Event<T>},
+                MultiAccount: substrate_pallet_multi_account::{Module, Call, Storage, Event<T>},
```
vedhavyas: space ?

runtime/src/lib.rs
```diff
@@ -497,6 +498,24 @@ impl nfts::Trait for Runtime {
     type Event = Event;
 }
 
+parameter_types! {
```
vedhavyas: are these parameters already discussed ?

runtime/src/lib.rs
ivanceras: I just realized that the substrate code base is using tabs as opposed to spaces(as most projects are).

runtime/src/lib.rs
ivanceras: probably not yet discussed, however @philipstanislaus told me to just copy the same values as the parameter types of `utility` pallet

philipstanislaus: These are the values I proposed a few weeks back in the deposit comparison. I have just triggered the discussion again on Slack, but they will most likely stay like this. Let's merge the PR and change them later if needed.

runtime/src/lib.rs
```diff
@@ -497,6 +498,24 @@ impl nfts::Trait for Runtime {
...
+	pub const MultiAcountSigDepositBase: Balance = 30 * CENTI_RAD;
```
philipstanislaus:
```suggestion
	pub const MultiAccountSigDepositBase: Balance = 30 * CENTI_RAD;
```

runtime/src/lib.rs
philipstanislaus: Yeah, we do not rustfmt to stay as close to Substrate (upstream) as possible to simplify upgrades. I hope they will rustfmt their codebase soon though ;)

runtime/src/lib.rs
ivanceras: Thanks for catching that.

runtime/src/lib.rs
ivanceras: The paritytech/substrate repo is consistent with their [formatting](https://github.com/paritytech/substrate/blob/master/.editorconfig) which they are using tabs for all their rust code. I'll recreate a `rustfmt.toml` close equivalent of their configuration. I'll make a separate PR for formatting changes, after all the pending PR are merged, otherwise there will be a lot of formatting conflicts.

ivanceras:
@philipstanislaus @vedhavyas  can we merge this?

runtime/src/lib.rs
philipstanislaus: Sounds good, thanks!

philipstanislaus:
@ivanceras I don’t see the genesis changes I mentioned above yet. Good to merge after that!

PR #204, opened 2020-03-19
data member should be public then other mod/pallet can access.

juniuszhou:
Set all data members in anchor and proofs as public. Code change also include some code formating.


juniuszhou:
Close it since I just notice there is a method to get doc root. And it is a better way to access value via member method.

PR #205, opened 2020-03-19
Add method to access leaf hash in proof.

juniuszhou:
Add method to access leaf hash in proof.


runtime/src/proofs.rs
```diff
@@ -17,6 +18,10 @@ impl Proof {
             sorted_hashes,
         }
     }
+
+    pub fn get_leaf_hash(&self) -> H256 {
```
vedhavyas: can we make `leaf_hash` field in proofs public instead ?

runtime/src/proofs.rs
```diff
@@ -1,5 +1,6 @@
 use codec::{Decode, Encode};
 use sp_core::H256;
+use sp_runtime::RuntimeDebug;
```
vedhavyas: why do we need this ?

runtime/src/anchor/mod.rs
```diff
@@ -13,7 +13,7 @@ use frame_support::{
     storage::child::{self, ChildInfo},
     weights::SimpleDispatchInfo,
 };
-use frame_system::ensure_signed;
+use frame_system::{self as system, ensure_signed};
```
vedhavyas: why do we need this ?

runtime/src/proofs.rs
juniuszhou: Yes, we can. I implemented it this way before. Then I found in anchor mod, a get_doc_root provided instead of setting data member public. I think it is better to following the team coding style.

runtime/src/proofs.rs
juniuszhou: It is from merge of old version code in master. I will remove it.

runtime/src/anchor/mod.rs
juniuszhou: The same as last comment.

runtime/src/anchor/mod.rs
```diff
@@ -43,24 +43,18 @@ const CHILD_INFO: ChildInfo<'static> = ChildInfo::new_default(b"anchor");
 #[derive(Encode, Decode, Default, Clone, PartialEq)]
 #[cfg_attr(feature = "std", derive(Debug))]
 pub struct PreCommitData<Hash, AccountId, BlockNumber> {
-    signing_root: Hash,
-    identity: AccountId,
-    expiration_block: BlockNumber,
+    pub signing_root: Hash,
+    pub identity: AccountId,
+    pub expiration_block: BlockNumber,
```
vedhavyas: why do we need to make this public? Are we using these anywhere ?

runtime/src/proofs.rs
```diff
@@ -6,8 +6,8 @@ use sp_std::vec::Vec;
 #[cfg_attr(not(feature = "std"), derive(sp_runtime::RuntimeDebug))]
 #[cfg_attr(feature = "std", derive(Debug))]
 pub struct Proof {
-    leaf_hash: H256,
-    sorted_hashes: Vec<H256>,
+    pub leaf_hash: H256,
+    pub sorted_hashes: Vec<H256>,
```
vedhavyas: why do we need to make `sorted_hashes` public?

runtime/src/anchor/mod.rs
```diff
@@ -43,24 +43,18 @@ const CHILD_INFO: ChildInfo<'static> = ChildInfo::new_default(b"anchor");
...
-    anchored_block: BlockNumber,
-}
-
-impl<Hash, BlockNumber> AnchorData<Hash, BlockNumber> {
-    pub fn get_doc_root(self) -> Hash {
-        self.doc_root
-    }
+    pub id: Hash,
+    pub doc_root: Hash,
+    pub anchored_block: BlockNumber,
```
vedhavyas: why do we need to make `id` and `anchored_block` public?

runtime/src/anchor/mod.rs
juniuszhou: Just make the definition consistent and ready for access in the future.

runtime/src/proofs.rs
juniuszhou: The same as last comment. It is strange to just set one member as public to me. I can't see any side effect here.

runtime/src/anchor/mod.rs
juniuszhou: The same as last comment.

juniuszhou:
If just set the two data members public, why not just expose two get methods as I implement in last commit? I can't understand here.

PR #212, opened 2020-03-23
reformat code using cargo fmt

ivanceras:
The configurations used in in `rustfmt.toml` file, all source files are reformatted to use tabs instead of mix of spaces and tabs
The max_width of a line is set to 100

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,4 @@
-use vergen::{ConstantsFlags, generate_cargo_keys};
+use vergen::{generate_cargo_keys, ConstantsFlags};
 
 const ERROR_MSG: &str = "Failed to generate metadata files";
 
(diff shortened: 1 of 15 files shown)


ivanceras:
@philipstanislaus  here is the link to [substrate's config](https://github.com/paritytech/substrate/blob/master/.editorconfig)
These are the only relevant configs that is specific to the `cargo-fmt`
```
indent_style=tab
indent_size=tab
max_line_length=100
```
The rest are editor specific such as
```
tab_width=4
```
which is used by the editor on how to represent tabs shown to the user.

ivanceras:
It looks like rustfmt will reorder the imports alphabetically by default [reorder_imports](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#reorder_imports)
I will set the `reorder_imports = false` to mirror that of substrate. It should not change a lot of things other than the tabs and max line width of 100.

philipstanislaus:
I'm not sure that will lead to both looking the same. 

I feel it's quite unproductive to spend too much time on `rustfmt`, what speaks against just ignoring upstream the files?

ivanceras:
I agree, cargo fmt is doing a lot of code movement here. I will close this for now.

PR #215, opened 2020-03-26
Run srtool for every release

vedhavyas:


vedhavyas:
> If we do this for every release, can we also upload binaries to GitHub automatically?
We could do that and that was my initial proposal but @lucasvo suggested we rebuild and independently, verify hash from the value in Travic CI and to runtime

PR #219, opened 2020-03-31
Add ci script to auto-fail PR if changes in runtime version is not met with the defined conditions

ivanceras:
Closes issue #200 

The ci-script behavior:

Any pull request to the `centrifuge-chain` will **auto-fail** if
- there are changes in the `runtime/` and PR is labeled as `breakapi` and `spec_version` is not changed in `runtime/src/lib.rs`.
- there are changes in the `runtime/` and PR is NOT labeled as `breakapi` and `impl_version` is not changed in `runtime/src/lib.rs`.

The script should ensure that changes in `runtime/` will have a appropriate change in the `version`

### DIFF
--- a/ci/check_pr.sh
+++ b/ci/check_pr.sh
@@ -79,8 +79,8 @@ LABEL_MARKER="breakapi"
 LABELED_MARKER_COUNT=$(echo -e "${pr_label}" | grep -w ${LABEL_MARKER} | wc -l)
 
 if [ $RUNTIME_FILE_CHANGED != "0" ]
-	then
-		echo "There are ${RUNTIME_FILE_CHANGED} files changed in runtime "
+then
+	echo "There are ${RUNTIME_FILE_CHANGED} files changed in runtime "
 
 	if [ "${LABELED_MARKER_COUNT}" != "0" ]
 	then
@@ -112,6 +112,9 @@ if [ $RUNTIME_FILE_CHANGED != "0" ]
 			echo "ERROR: impl_version should be changed in ${VERSIONS_FILE}"
 		fi
 	fi
+else
+	echo "OK: No changes in runtime, no need for version change"
+	exit 0
 fi
 
 exit 1
--- a/ci/check_pr.sh
+++ b/ci/check_pr.sh
@@ -9,6 +9,12 @@
 # get the commit sha for this PR
 CI_COMMIT_SHA=$(git rev-parse HEAD)
 
+
+boldprint () { printf "|\n| \033[1m${@}\033[0m\n|\n" ; }
+
+boldprint "latest 10 commits of ${CI_COMMIT_REF_NAME}"
+git log --graph --oneline --decorate=short -n 10
+
 # base the origin/master as the base commit
 BASE_COMMIT="origin/master"
 VERSIONS_FILE="runtime/src/lib.rs"
@@ -20,7 +26,6 @@ echo "Changed files $CHANGED_FILES"
 
 # count the number of files changed in runtime directory
 RUNTIME_FILE_CHANGED=$(echo "${CHANGED_FILES}" | grep -e ^runtime/ | wc -l)
-echo "There are ${RUNTIME_FILE_CHANGED} files changed in runtime "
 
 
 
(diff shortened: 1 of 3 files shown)


ci/check_pr.sh
```diff
@@ -0,0 +1,120 @@
...
+# that alters the version.
+
+# get the commit sha for this PR
+CI_COMMIT_SHA=$(git rev-parse HEAD)
+
+
+boldprint () { printf "|\n| \033[1m${@}\033[0m\n|\n" ; }
+
+boldprint "latest 10 commits of ${CI_COMMIT_REF_NAME}"
+git log --graph --oneline --decorate=short -n 10
```
vedhavyas, line 16: any reason to print this ?

ci/check_pr.sh
```diff
@@ -0,0 +1,120 @@
...
+
+# base the origin/master as the base commit
+BASE_COMMIT="origin/master"
+VERSIONS_FILE="runtime/src/lib.rs"
+
+# show the diff of origin/master and this PR sha
+CHANGED_FILES=$(git diff --name-only ${BASE_COMMIT}...${CI_COMMIT_SHA})
+
+echo "Changed files $CHANGED_FILES"
```
vedhavyas, line 25: this will be to much noise IMO

ci/check_pr.sh
```diff
@@ -0,0 +1,120 @@
...
+
+LABEL_MARKER="breakapi"
+
+LABELED_MARKER_COUNT=$(echo -e "${pr_label}" | grep -w ${LABEL_MARKER} | wc -l)
+
+if [ $RUNTIME_FILE_CHANGED != "0" ]
+then
+	echo "There are ${RUNTIME_FILE_CHANGED} files changed in runtime "
+
+	if [ "${LABELED_MARKER_COUNT}" != "0" ]
```
vedhavyas, line 85: we cannot expect the PR to have labels. instead, we should do all the required checks and add the  labels if required to the PR

ci/check_pr.sh
ivanceras: This is for debugging purposes, the dev who make the PR can see which commits could have cause the PR to fail, this is the same as the `check_runtime.sh`. The dev also see if this script is checking against the latest commit.

ci/check_pr.sh
ivanceras: So, the behavior would then be check if the necessary label is in the PR, if not the script will have to add it according to the changes seen in `runtime/src/lib.rs` version?

lucasvo: Let's just have it reject the PR and make the author add the label manually. That's good enough and avoids having to use a github API key and make the script a lot more complicated.

ivanceras:
>     * I think it makes sense to have two labels. 1 `changes-runtime` 2. `breaks-api`
> 
>     * changes-runtime if the runtime was changed, `breaks-api` if the spec_version is incremented

@vedhavyas I already added the in the script expecting these 2 labels.
- [X] If the `breaks-api` label is present, the script expects that the `spec_version` is changed.
- [X] If the `changes-runtime` label is present, the script expects that the `impl_version` is changed.

However, we don't cover further:
- [ ] If `changes-runtime` is present, the `spec_version` must remains the same.
- [ ] If `breaks-api` is present, `impl_version` should be reset to `0`
- [ ] If `breaks-api` and `changes-runtime` label is present, the `changes-rutime` is ignored.

ivanceras:
@vedhavyas as per discussion, we remove checking the PR labels and this is now the behavior of the scripts

 - if runtime files are changed:
	- Check if the spec_version is incremented
 	- If spec_version incremented, ensure impl_version is 0 and exit successfully
 	- if spec_version is NOT incremented, check if the impl_version is incremented
 		- if impl_version incremented, then exit successfully
 		- if impl_version is NOT incremented then fail the script
 - if runtime files are NOT changed, exit successfully

ci/check_pr.sh
```diff
@@ -0,0 +1,103 @@
...
+
+OK="${green}${block}OK${nc}"
+ERROR="${red}${block}ERROR${nc}"
+FATAL="${red}${block}FATAL${nc}"
+
+
+# show the diff of origin/master and this PR sha
+CHANGED_FILES=$(git diff --name-only ${BASE_COMMIT}...${PR_COMMIT})
+
+#echo "Changed files: ${CHANGED_FILES}"
```
vedhavyas, line 35: remove the comment

Issue #222, opened 2020-04-09
PR commit hash in ci/check_runtime_changes.sh is wrong causing the script to pass instead of failing


ivanceras:
The command `git rev-parse HEAD` is returning a commit hash of the PR when merged to the master instead of the last commit hash of the PR prior to merging.

This causes the `git diff --name-only origin/master..${PR_COMMIT}` to return an error message, since the `${PR_COMMIT}` hash is not yet on this repo.

This causes the entire script to succeed since the next scripts just regex match `^runtime/` which returns 0.

The script must be revised to: 
- use the correct commit hash, rather than the PR commit hash. 
      - try using `git rev-parse HEAD^1` when it is a pull request, `git rev-parse HEAD` otherwise
- fail when `git diff` fails.

ivanceras:
The default branch set in github affects the travis-ci build. 
The version `git rev-parse HEAD` is indeed returning the correct hash, which is the PR commit hash.
This is also equivalent to just using HEAD which is the latest commit of the PR.

PR #223, opened 2020-04-10
fix: fail the entire script when git fails

ivanceras:
Resolves Issue #222

### DIFF
--- a/ci/check_runtime_changes.sh
+++ b/ci/check_runtime_changes.sh
@@ -12,10 +12,15 @@
 
 # origin/master as the base commit
 BASE_COMMIT="origin/master"
+PR_COMMIT="HEAD"
 VERSIONS_FILE="runtime/src/lib.rs"
 
-PR_COMMIT=$(git rev-parse HEAD)
-echo "commit: ${PR_COMMIT}"
+echo "PR commit: ${PR_COMMIT}"
+echo "BASE commit: ${BASE_COMMIT}"
+
+PR_BRANCH=${TRAVIS_PULL_REQUEST_BRANCH}
+
+echo "PR branch ${PR_BRANCH}"
 
 # use color in echo for indicating success or fail
 red='\033[0;31m'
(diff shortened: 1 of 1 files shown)


ci/check_runtime_changes.sh
```diff
@@ -30,7 +35,13 @@ FATAL="${red}${block}FATAL${nc}"
...
 
 # show the diff of origin/master and this PR sha
-CHANGED_FILES=$(git diff --name-only ${BASE_COMMIT}...${PR_COMMIT})
+CHANGED_FILES=$(git diff --name-only ${BASE_COMMIT} ${PR_COMMIT} 2>&1 )
+GIT_STATUS=$?
+if (( $GIT_STATUS != 0 ))
+then
+	echo -e "${red}${bold}GIT ERROR${nc}: $CHANGED_FILES"
+	exit 1
+fi
```
philipstanislaus, line 44: Can you point me to the explanation of what the status codes of git diff mean? I played around with them locally, and the status code always seems to be `0`

ivanceras: The `git diff` failed awhile back when I tested on my fork.  
This only happens when you set the default branch repo in github to some other branch other than `master`

Full details:
When I created a pull-request (with changes to `runtime/`, but no changes to `spec_version` and `impl_version`), I expected the whole script to fail, however it passed for the wrong reason.

`git diff` failed` because I set the default branch of my fork repo to some branch other than `master`, to which `travis-ci` pulls by default. So, when `git diff origin/master..{PR_COMMIT}`, it couldn't find `origin/master` in the git tree, therefore the command fails with exit 1 (which we didn't check). In the script, we only capture the output string (which returns the error messages instead of files that changed in `runtime/` directory.
We match which files are changed in the `runtime/` (empty at this point) and check the spec version (no version changed) so the whole script passed.

philipstanislaus: Makes sense, thanks!

PR #227, opened 2020-04-17
Created an annotated mainnet spec version

lucasvo:


philipstanislaus:
A couple of projects allow json with comments though, e. g. Visual Studio Code. https://github.com/Microsoft/node-jsonc-parser

I agree though that the current approach is problematic since it suggests the spec file can be imported, would it (probably) can’t (have not tested, maybe it works).

The easiest fix is just to rename it so the format is clearly different, I think `.json-with-comments`, `.jsonc` and `.yaml` are all fine.

lucasvo:
I changed the file ending so fancy IDEs don't complain about invalid json anymore.

PR #228, opened 2020-04-21
Initial code for pallet_bridge based on chainsafe/chainbridge example

ivanceras:
This is based on [ChainSafe/chainbridge](https://github.com/ChainSafe/chainbridge-substrate) example pallet.

## Request for comment on:
- rename `bridge` module as `pallet_bridge` to have a clear distinction from `chainbridge`
 
- needs enhancements on the doc comment on the module functions
    - `transfer_hash`
     - `transfer_native`
     - `transfer` 
     - `remark`

- Erc721 `parameter_types` is not included.
```rust
parameter_types!{
     pub const Erc721Id: bridge::ResourceId = bridge::derive_resource_id(1, &blake2_128(b"NFT"));
}
```
- No `GenesisConfig` defined in `chain_spec`.
     - in the chainsafe example `bridge_pallet` has no defined `decl_storage!`
     - `bridge_pallet` inherits on `chainbridge::Trait` where a `decl_storage` is defined.
         - will this have an entry into the `GenesisConfig` in `chain_spec`?
```rust
pub trait Trait: system::Trait + chainbridge::Trait {
...
}
```

### DIFF
--- a/runtime/src/bridge.rs
+++ b/runtime/src/bridge.rs
@@ -45,7 +45,8 @@ decl_module! {
             T::Currency::transfer(&source, &bridge_id, amount.into(), AllowDeath)?;
 
             let resource_id = T::NativeTokenId::get();
-            <chainbridge::Module<T>>::transfer_fungible(dest_id, resource_id, recipient, amount)
+            <chainbridge::Module<T>>::transfer_fungible(dest_id, resource_id, recipient, amount)?;
+			Ok(())
         }
 
 
--- a/runtime/src/bridge.rs
+++ b/runtime/src/bridge.rs
@@ -67,6 +67,17 @@ decl_module! {
             Ok(())
         }
 
+		pub fn transfer_generic(
+			origin,
+			dest_id: chainbridge::ChainId,
+			resource_id: ResourceId,
+			metadata: Vec<u8>,
+		) -> DispatchResult {
+			T::BridgeOrigin::ensure_origin(origin)?;
+			<chainbridge::Module<T>>::transfer_generic(dest_id, resource_id, metadata)?;
+			Ok(())
+		}
+
     }
 }
 
--- a/runtime/src/bridge.rs
+++ b/runtime/src/bridge.rs
@@ -68,12 +68,11 @@ decl_module! {
         }
 
 		pub fn transfer_generic(
-			origin,
+			_origin,
 			dest_id: chainbridge::ChainId,
 			resource_id: ResourceId,
 			metadata: Vec<u8>,
 		) -> DispatchResult {
-			T::BridgeOrigin::ensure_origin(origin)?;
 			<chainbridge::Module<T>>::transfer_generic(dest_id, resource_id, metadata)?;
 			Ok(())
 		}
--- a/runtime/src/bridge.rs
+++ b/runtime/src/bridge.rs
@@ -1,0 +1,477 @@
+use frame_support::traits::{Currency, ExistenceRequirement::AllowDeath, Get};
+use frame_support::{decl_error, decl_event, decl_module, dispatch::DispatchResult, ensure};
+use frame_system::{self as system, ensure_signed};
+use sp_runtime::traits::EnsureOrigin;
+use sp_std::prelude::*;
+
+type ResourceId = chainbridge::ResourceId;
+
+pub trait Trait: system::Trait + chainbridge::Trait {
+    type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
+    /// Specifies the origin check provided by the chainbridge for calls that can only be called by the chainbridge pallet
+    type BridgeOrigin: EnsureOrigin<Self::Origin, Success = Self::AccountId>;
+
+    /// Ids can be defined by the runtime and passed in, perhaps from blake2b_128 hashes.
+    type HashId: Get<ResourceId>;
+    type NativeTokenId: Get<ResourceId>;
+}
+
+decl_event! {
+    pub enum Event<T> where
+        <T as frame_system::Trait>::Hash,
+    {
+        Remark(Hash),
+    }
+}
+
+decl_error! {
+    pub enum Error for Module<T: Trait>{
+        InvalidTransfer,
+    }
+}
+
+decl_module! {
+    pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+        const HashId: ResourceId = T::HashId::get();
+        const NativeTokenId: ResourceId = T::NativeTokenId::get();
+
+        fn deposit_event() = default;
+
+        //
+        // Initiation calls. These start a chainbridge transfer.
+        //
+
+        /// Transfers an arbitrary hash to a (whitelisted) destination chain.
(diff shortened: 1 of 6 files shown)


runtime/src/lib.rs
```diff
@@ -538,6 +543,31 @@ impl pallet_identity::Trait for Runtime {
...
+impl bridge::Trait for Runtime {
+	type Event = Event;
+	type BridgeOrigin = EnsureSigned<AccountId>;
+	type HashId = HashId;
+	type NativeTokenId = NativeTokenId;
+}
+
+
+parameter_types! {
+    pub const ChainId: u8 = 0;
```
mikiquantum, line 561: I believe centrifuge-chain should be 1.

runtime/src/bridge.rs
```diff
@@ -0,0 +1,477 @@
...
+        const NativeTokenId: ResourceId = T::NativeTokenId::get();
+
+        fn deposit_event() = default;
+
+        //
+        // Initiation calls. These start a chainbridge transfer.
+        //
+
+        /// Transfers an arbitrary hash to a (whitelisted) destination chain.
+        pub fn transfer_hash(origin, hash: T::Hash, dest_id: chainbridge::ChainId) -> DispatchResult {
```
mikiquantum, line 45: This function will be wrapped in the NFT validate_mint, we can remove it from here

runtime/src/bridge.rs
```diff
@@ -0,0 +1,477 @@
...
+decl_error! {
+    pub enum Error for Module<T: Trait>{
+        InvalidTransfer,
+    }
+}
+
+decl_module! {
+    pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+        const HashId: ResourceId = T::HashId::get();
+        const NativeTokenId: ResourceId = T::NativeTokenId::get();
```
mikiquantum, line 36: How do we get the actual resourceID so we can whitelist it in the solidity side?

runtime/src/lib.rs
```diff
@@ -538,6 +543,31 @@ impl pallet_identity::Trait for Runtime {
     type ForceOrigin = EnsureProportionMoreThan<_1, _2, AccountId, CouncilCollective>;
 }
 
+
+parameter_types! {
+    pub const HashId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &blake2_128(b"hash"));
```
mikiquantum, line 548: I would rename this to cent_nft_hash
```suggestion
    pub const HashId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &blake2_128(b"cent_nft_hash"));
```

runtime/src/bridge.rs
```diff
@@ -0,0 +1,477 @@
...
+        pub fn transfer_hash(origin, hash: T::Hash, dest_id: chainbridge::ChainId) -> DispatchResult {
+            ensure_signed(origin)?;
+
+            let resource_id = T::HashId::get();
+            let metadata: Vec<u8> = hash.as_ref().to_vec();
+            <chainbridge::Module<T>>::transfer_generic(dest_id, resource_id, metadata)
+        }
+
+        /// Transfers some amount of the native token to some recipient on a (whitelisted) destination chain.
+        pub fn transfer_native(origin, amount: u32, recipient: Vec<u8>, dest_id: chainbridge::ChainId) -> DispatchResult {
```
mikiquantum, line 54: How are the weights/fees defined here? I think we need to define them here.

ivanceras: I have no idea how weights/fees would be here.

mikiquantum: @philipstanislaus this is the entrypoint for a user in the centchain to move tokens to ethereum. Do you have an idea of what weight amount should be set? Prob a bit more expensive than a regular balance transfer?

lucasvo: shouldn't it be `amount: u64`?

lucasvo: we'll have to discuss weights, @mikiquantum I don't think @philipstanislaus did any calculations but let's fix the weights at a later point.

mikiquantum: Technically a Balance in substrate is u128. The bridge pallet only supports u32 at the moment. That would be a problem if the bridge uses minimum units for any chain interaction.
> we'll have to discuss weights, @mikiquantum I don't think @philipstanislaus did any calculations but let's fix the weights at a later point.
Can we add at least now the same as any balance transfer weight?

ivanceras: > shouldn't it be amount: u64

I think this should be using Balance for the type.

ivanceras: https://github.com/ChainSafe/chainbridge-substrate/issues/44

runtime/src/bridge.rs
mikiquantum: We can get it from the metadata constants as soon as it is deployed

mikiquantum:
@ivanceras I think we can start adding here:
- new param to nft_validate_mint `dest_id: bridge::ChainId`
- pump the impl_version
We can add the amount precision a bit later once is fixed on the bridge pallet side

ivanceras:
> * pump the impl_version

Shouldn't we bump up the `spec_version` instead, since we add a new argument to `nfts::validate_mint`

ivanceras:
Added new commits just solving compilation errors.

Need to set the values for:
```rust
let resource_id = [0;32]; //TODO: where to get this?
let metadata = Vec::with_capacity(0); // TODO: what's the content of metadata
```

mikiquantum:
I see, I didnt notice that the spec_version has changed when adding the identity pallet. Since we didnt release I would stick with the same 227.
But we need to pump the impl_version otherwise the ci doesnt pass. Which I would add the version check into its own parallel check with the other two steps, so all three run in parallel and we the version check doesn't block the rest of the tests, but will still report failure as a whole.

mikiquantum:
For resource_id: `HashId: ResourceId = T::HashId::get();`
For metadata: `bundled_hash`

runtime/src/nfts.rs
```diff
@@ -39,15 +40,17 @@ decl_module! {
 
             // get the bundled hash
             let bundled_hash = Self::get_bundled_hash(pfs, deposit_address);
-
             Self::deposit_event(RawEvent::DepositAsset(bundled_hash));
 
+			let metadata = bundled_hash.as_ref().to_vec();
+			let resource_id = <T as pallet_bridge::Trait>::HashId::get();
+			<chainbridge::Module<T>>::transfer_generic(dest_id, resource_id, metadata)?;
             Ok(())
```
mikiquantum, line 48: Lets remove the `Ok()` here, since `<chainbridge::Module<T>>::transfer_generic(dest_id, resource_id, metadata)?;` already returns that.

ivanceras: This is a standard practice, and makes the code more extensible, such as: whenever you need to add more statement calls before the Ok(()), it doesn't need to be a function that returns a `Result`.

mikiquantum: I am fine by following the most standard practice, then lets follow the same thing everywhere, like: https://github.com/centrifuge/centrifuge-chain/pull/228/files#diff-ca12cf69a0ed4856af5c862837c21d31R48

PR #232, opened 2020-04-28
Update Bridge version + dev fixture data

mikiquantum:
This includes latest changes on bridge pallet:
- Wrapped admin operations into private functions
- Enable admin operations to use council voting

Allow genesis setup on dev/local/fulvous to whitelist bridge data.

### DIFF
--- a/runtime/src/bridge.rs
+++ b/runtime/src/bridge.rs
@@ -88,16 +88,19 @@ decl_module! {
 }
 
 impl<T: Trait> Module<T> {
+	/// Its called as part of genesis step to initialize some dev parameters
 	fn initialize_test(chains: &[u8], relayers: &[T::AccountId], resources: &[ResourceId]) {
 		chains.into_iter().for_each(|c| {
-			<chainbridge::Module<T>>::whitelist(*c);
+			<chainbridge::Module<T>>::whitelist(*c).unwrap_or_default();
 		});
 		relayers.into_iter().for_each(|rs| {
-			<chainbridge::Module<T>>::register_relayer(rs);
+			<chainbridge::Module<T>>::register_relayer(rs.clone()).unwrap_or_default();
 		});
-		<chainbridge::Module<T>>::set_relayer_threshold(relayers.len() as u32);
+		if !relayers.is_empty() {
+			<chainbridge::Module<T>>::set_relayer_threshold(relayers.len() as u32).unwrap_or_default();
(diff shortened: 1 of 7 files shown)


runtime/src/bridge.rs
```diff
@@ -20,6 +20,18 @@ pub trait Trait: system::Trait + chainbridge::Trait {
...
 
+decl_storage! {
+	trait Store for Module<T: Trait> as Bridge {}
+
+	add_extra_genesis {
+        config(chains): Vec<u8>;
+        config(relayers): Vec<T::AccountId>;
+        config(resources): Vec<ResourceId>;
+
+        build(|config| Module::<T>::initialize_test(&config.chains, &config.relayers, &config.resources))
```
philipstanislaus, line 31: Let’s call this just `initialize`, since it can also be used outside of tests

runtime/src/lib.rs
```diff
@@ -566,7 +566,8 @@ impl chainbridge::Trait for Runtime {
     type Event = Event;
     type Proposal = Call;
     type ChainId = ChainId;
-
+	/// A straight majority of the council can update bridge settings.
+	type AdminOrigin = pallet_collective::EnsureProportionAtLeast<_1, _2, AccountId, CouncilCollective>;
```
philipstanislaus, line 570: I would require >¾ here, for security purposes. The bridge holds a lot of value, better be safe than sorry ;)

runtime/src/nfts.rs
```diff
@@ -158,10 +159,15 @@ mod tests {
...
 
+    ord_parameter_types! {
+		pub const One: u64 = 1;
+	}
+
 	impl chainbridge::Trait for Test {
 		type Event = ();
 		type Proposal = Call;
 		type ChainId = TestChainId;
+        type AdminOrigin = EnsureSignedBy<One, u64>;
```
philipstanislaus, line 170: Indentation is somehow broken?

runtime/src/lib.rs
mikiquantum: makes sense.

runtime/src/nfts.rs
mikiquantum: I think is the classic tale of tabs vs spaces. My IDE transforms tabs in spaces, but I believe the initial code was with tabs. It looks good on an editor anyways.

PR #260, opened 2020-06-30
Include additional fees for nft_validate + native transfer

mikiquantum:


runtime/src/bridge.rs
```diff
@@ -13,7 +14,10 @@ type ResourceId = chainbridge::ResourceId;
 type BalanceOf<T> =
     <<T as Trait>::Currency as Currency<<T as frame_system::Trait>::AccountId>>::Balance;
 
-pub trait Trait: system::Trait + chainbridge::Trait {
+/// Additional Fee charged when moving native tokens to target chains (RAD)
+const TOKEN_FEE : u32 = 20;
```
mikiquantum, line 18: Wondering if I should make this a storage value, to modify it faster if needed.

runtime/src/bridge.rs
```diff
@@ -58,9 +62,21 @@ decl_module! {
...
+        /// Adds additional fee to compensate the current cost of target chains
         #[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
         pub fn transfer_native(origin, amount: BalanceOf<T>, recipient: Vec<u8>, dest_id: chainbridge::ChainId) -> DispatchResult {
             let source = ensure_signed(origin)?;
+			let token_fee = <T as pallet_balances::Trait>::Balance::from(TOKEN_FEE);
+			let total_amount = U256::from(amount.saturated_into()) + U256::from(token_fee.saturated_into());
+
+			// Ensure account has enough balance for both fee and transfer
+			// Check to avoid balance errors down the line that leave balance storage in an inconsistent state
+			let current_balance = T::Currency::free_balance(&source);
```
mikiquantum, line 74: Main reason of this check is to avoid burning tokens from the account and not being able to succeed bridging the token. Only applicable to errors related to balance.

runtime/src/nfts.rs
```diff
@@ -7,7 +7,10 @@ use sp_core::H256;
 use sp_std::vec::Vec;
 use crate::bridge as pallet_bridge;
 
-pub trait Trait: anchor::Trait + pallet_bridge::Trait {
+/// Additional Fee charged to validate NFT proofs (RAD)
+const NFT_FEE : u32 = 10;
```
mikiquantum, line 11: Same here, regarding the storage.

runtime/src/bridge.rs
```diff
@@ -58,9 +62,21 @@ decl_module! {
         fn deposit_event() = default;
 
         /// Transfers some amount of the native token to some recipient on a (whitelisted) destination chain.
+        /// Adds additional fee to compensate the current cost of target chains
         #[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
         pub fn transfer_native(origin, amount: BalanceOf<T>, recipient: Vec<u8>, dest_id: chainbridge::ChainId) -> DispatchResult {
             let source = ensure_signed(origin)?;
+			let token_fee = <T as pallet_balances::Trait>::Balance::from(TOKEN_FEE);
+			let total_amount = U256::from(amount.saturated_into()) + U256::from(token_fee.saturated_into());
```
philipstanislaus, line 70: Can we use `saturating_add` here? This could overflow.

runtime/src/bridge.rs
philipstanislaus: Agree that it is not strictly needed since we don't modify state before doing the burn_fee and the transfer.

runtime/src/bridge.rs
```diff
@@ -58,9 +62,21 @@ decl_module! {
...
+
+			// Ensure account has enough balance for both fee and transfer
+			// Check to avoid balance errors down the line that leave balance storage in an inconsistent state
+			let current_balance = T::Currency::free_balance(&source);
+			ensure!(U256::from(current_balance.saturated_into()) >= total_amount, "Insufficient Balance");
+
+            // Burn additional fees
+            <fees::Module<T>>::burn_fee(&source, token_fee)?;
```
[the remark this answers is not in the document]
philipstanislaus: Agree, good point @vedhavyas . 

Please also move the burning of fees under the chain_whitelisted check, so all checks are done before modifying state.

runtime/src/bridge.rs
```diff
@@ -299,6 +326,37 @@ mod tests{
...
+					amount.clone(),
+					recipient.clone(),
+					dest_chain,
+				),
+				"Insufficient Balance"
+			);
+
+			// Account balance should be reverted to original balance
+			account_current_balance = <pallet_balances::Module<Test>>::free_balance(RELAYER_B);
+			assert_eq!(account_current_balance, 100);
```
philipstanislaus, line 357: 👍

runtime/src/lib.rs
```diff
@@ -87,7 +87,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // implementation changes and behavior does not, then leave spec_version as
     // is and increment impl_version.
     spec_version: 228,
-    impl_version: 5,
+    impl_version: 6,
```
philipstanislaus, line 90: I'd argue that this is not just an implementation change (such as a refactor), but rather a change in runtime behavior. Can we implement spec_version instead?

runtime/src/nfts.rs
```diff
@@ -44,6 +48,11 @@ decl_module! {
 
 			let metadata = bundled_hash.as_ref().to_vec();
 			let resource_id = <T as pallet_bridge::Trait>::HashId::get();
+
+			// Burn additional fees
+            let nft_fee = <T as pallet_balances::Trait>::Balance::from(NFT_FEE);
+            <fees::Module<T>>::burn_fee(&who, nft_fee)?;
```
philipstanislaus, line 54: As @vedhavyas mentioned above, we should probably give those to the author as well.

runtime/src/bridge.rs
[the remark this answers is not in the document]
lucasvo: No, this was intentionally decided to burn. We don't want to give the bridge fees to the block author but be able to pay for the gas spent on bridging this to Ethereum.

runtime/src/bridge.rs
mikiquantum: The state is changed and the operations within are not reverted. So for example, if the user has enough balance for the fee but not for the transfer, indeed it will error out in the second one, but the first fee will be lost already.

runtime/src/lib.rs
mikiquantum: We havent released this yet, we will package it with this version.

runtime/src/bridge.rs
mikiquantum: I think in this case is an overkill to use the fee storage module. Since I will need to store as well the key separately, that would be two items to store. The key to use and the fee. 
I would rather just store the RAD amount.

philipstanislaus: Either is fine for me, but I guess you are right – we can always make fees adjustable in an on-chain upgrade. If we don't plan to change them anytime soon and if there's no economic risk of being attacked (i. e. fees are way too low), I'd suggest to keep the fees static for now. Any thoughts @lucasvo ?

runtime/src/bridge.rs
philipstanislaus: Yes, but I think we should follow Substrate's principle to not optimize for UX in the runtime. Sending a tx costs tx fees, even if it aborts/reverts. Loosing those fees would not be much different from that.

The UI can check for the balance before sending the tx.

runtime/src/bridge.rs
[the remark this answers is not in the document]
philipstanislaus: Got it, wasn't aware of that 👍

runtime/src/lib.rs
philipstanislaus: 👍

runtime/src/bridge.rs
vedhavyas: IMO, It would not be a lot storage since we will not have many keys. This can also be a single place for all configurable ones instead of having new API for each value. Either way, your call

runtime/src/bridge.rs
vedhavyas: Can we move all the checks before the state is changed?

mikiquantum: If the fee was relatively small I would agree with it, but we are talking at 20 RAD. I can move all the checks before the state change, I thought I did, but I missed the whitelist one, thanks for pointing it out.

runtime/src/bridge.rs
mikiquantum: The main reason for me to add just this fees storage is in case that the amount is too big/small that we need to react fast. My idea was to only add storage for the fee of 10 and 20 RAD ones for nft and native transfer, so we could do a quick storage update if needed. 
But maybe we are overengineering it, and in a few months once the bridge cost goes down, we can remove this fees altogether

PR #262, opened 2020-08-03
Fix wrong fee precision

mikiquantum:
Fee precision was set incorrectly when moving assets from the source chain.
We want to protect the bridge cost by increasing the fees to interact with the two main functions that move assets:
- nft_validate: 10 RAD
- native_transfer: 20 RAD

### DIFF
--- a/runtime/src/bridge.rs
+++ b/runtime/src/bridge.rs
@@ -1,3 +1,4 @@
+use crate::{fees};
 use frame_support::traits::{Currency, ExistenceRequirement::AllowDeath, Get};
 use frame_support::{
 	decl_error, decl_event, decl_module, decl_storage, dispatch::DispatchResult, ensure,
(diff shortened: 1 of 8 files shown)


runtime/src/lib.rs
```diff
@@ -87,7 +87,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // implementation changes and behavior does not, then leave spec_version as
     // is and increment impl_version.
     spec_version: 229,
-    impl_version: 0,
+    impl_version: 1,
```
philipstanislaus, line 90: I would argue that this changes the behavior (if a consumers application works now, it might stop working if we change fees).

runtime/src/bridge.rs
```diff
@@ -14,8 +14,8 @@ type ResourceId = chainbridge::ResourceId;
 type BalanceOf<T> =
     <<T as Trait>::Currency as Currency<<T as frame_system::Trait>::AccountId>>::Balance;
 
-/// Additional Fee charged when moving native tokens to target chains (RAD)
-const TOKEN_FEE : u32 = 20;
+/// Additional Fee charged when moving native tokens to target chains (20 RAD)
+const TOKEN_FEE: u128 = 20_000_000_000_000_000_000;
```
philipstanislaus, line 18: Can we use `20 * RAD` for simplicity here and in all the other places?

PR #266, opened 2020-08-17
RC6 + update module weights

vedhavyas:
# RC6 Changes from RC3:
- `TargetedFeeAdjustment` is removed and we are using the default one provided with RC6
- WeightInfo can be passed to core pallets. While some pallets doesn't use them, pallets like `democracy`, `balances` make full use of this feature. I have added the default weight values taken from the substrate node.
- `SlashDefferedTime` must be less than the `UnbondingTime`.
	- Reason: If a validator gets slashed, the slash is deferred for 7day. In the meantime, they can start unbonding and transfer funds since the bonding time is also 7. Hence made `SlashDefferedTime` to 6 days instead

# Weight Changes
With RC3, weights in substrate pallets have changed and so does the WeightToFee calculations. We would need to adjust our own module and our pallet weights accordingly.

## How are new weights calculated:
With the new weight changes, the extrinsic bytes fee will be significantly higher since we charge 1 Micro rad per byte.
So for a balance transfer of 1 rad, the extrinsic payload would be around 97 bytes i.e 97 Micro rad just for Byte fee. With adding the remaining BaseFee, WeightFee, we would be charging some around 300 Micro RAD. Right now, we charge 100 Micro RAD.

We can reduce the weight fee coefficient. Making it 0, means, ignoring the weight completely, would still lead to 97 Micro Rad. and that defeats the who purpose of having weights to extrinsic.

Most of the substrate's internal weights are set as per Reads and Writes to DB, and total processing time as well. In order to charge the extrinsic correctly and keeping the new fees in par with the current fees we charge, I have updated 
- TransactionByteFee -> 0.01 Micro RAD
- WeightFee Coefficient -> 315000

These changes will give us more control over how we charge the extrinsic based on their weights rather than extrinsic payload bytes.
With these changes, we will charge around 101.641 Micro RAD for 1 RAD balance transfer. But as the RAD number increases, so does the number of bytes in the extrinsic, which leads to slight increase in fees.

Using the `balance.transfer` as the base, I have calculated our module weights, multi-account, and chain bridge weights accordingly so that we are charging around the same fee we charging right now on Mainnet.

Anchor:

| Module      | Old Fee | New Fee|
| ----------- | ----------- | ----------- | 
| anchors.commit   | 100 Micro RAD | 100.627 |
| anchors.precommit   | 100 Micro RAD | 101.612 |

Since anchors commit and pre-commit, have the same extrinsic payload bytes, the fee will be similar. but extrinsic that contain huge bytes payload will change much higher.

An example would be `system.SetCode`. Seems like there are no fees charged for this extrinsic when I check the recent runtime upgrade we made from subscan(correct me if I'm wrong here). With this new weights, we will charging around `632.358 milli RAD`.


runtime/src/impls.rs
```diff
@@ -30,6 +33,30 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
...
+/// node's balance type.
+///
+/// This should typically create a mapping between the following ranges:
+///   - [0, frame_system::MaximumBlockWeight]
+///   - [Balance::min, Balance::max]
+///
+/// Yet, it can be used for any other sort of change to weight-fee. Some examples being:
+///   - Setting it to `0` will essentially disable the weight fee.
+///   - Setting it to `1` will cause the literal `#[weight = x]` values to be charged.
+pub struct WeightToFee;
```
mikiquantum, line 46: I think it would be useful to have here a sample fee conversion with the formula, input and output.
Let's say, if weight is `100_000_000` and applying `polynomial()` function breakdown then the result is -> `FeeResult`

runtime/src/lib.rs
```diff
@@ -120,17 +122,20 @@ impl Filter<Call> for BaseFilter {
 pub struct IsCallable;
 frame_support::impl_filter_stack!(IsCallable, BaseFilter, Call, is_callable);
 
+const AVERAGE_ON_INITIALIZE_WEIGHT: Perbill = Perbill::from_percent(10);
 parameter_types! {
     pub const BlockHashCount: BlockNumber = 250;
-    pub const MaximumBlockWeight: Weight = 1_000_000_000;
+    pub const MaximumBlockWeight: Weight = 2 * WEIGHT_PER_SECOND;
```
mikiquantum, line 128: Is this equivalent to 1/3 of the max block weight, based on a 6 secs block time right?

vedhavyas: yeap. This means 2 seconds of Compute time for target 6 seconds block time

mikiquantum:
@vedhavyas following this:

>WeightInfo can be passed to core pallets. While some pallets doesn't use them, pallets like democracy, balances make full use of this feature. I have added the default weight values taken from the substrate node.

Is it mandatory to pass this type to the runtime pallet? Wouldnt it use the default one if not passed?
```
type SystemWeightInfo = weights::frame_system::WeightInfo;
```

vedhavyas:
@mikiquantum true! Updated now

runtime/Cargo.toml
```diff
@@ -10,70 +10,72 @@ repository = "https://github.com/centrifuge/centrifuge-chain"
 
 [dependencies]
 # external pallet
-substrate-pallet-multi-account = { version = "2.0.0-rc3-0", git = "https://github.com/centrifuge/substrate-pallet-multi-account", rev = "557b4da832a1d74ed286c3a3a4b20b7d80f47e05", default-features = false }
-chainbridge = { version = "0.0.1", git = "https://github.com/chainsafe/chainbridge-substrate.git", rev = "59fbb767de890b6dcb781fb9e6c4c84a967693ff" , default-features = false}
+substrate-pallet-multi-account = { version = "2.0.0-rc6", git = "https://github.com/centrifuge/substrate-pallet-multi-account", rev = "5b8d12bd0260551f18223001b2c5091ea92fc7d8", default-features = false }
+chainbridge = { version = "0.0.1", git = "https://github.com/vedhavyas/chainbridge-substrate.git", rev = "19df3f21889e60be286f532cd785d5d17ff61c4c" , default-features = false}
```
vedhavyas, line 14: Will point to chain bridge once https://github.com/ChainSafe/chainbridge-substrate/pull/64 is merged

runtime/src/anchor/mod.rs
```diff
@@ -118,7 +118,7 @@ decl_module! {
         /// # <weight>
         /// minimal logic, also needs to be consume less block capacity + cheaper to make the pre-commits viable.
         /// # </weight>
-        #[weight = 500_000]
+        #[weight = 193_000_000]
```
philipstanislaus, line 121: Instead of constants, what do you think about using benchmarking as Substrate does to determine the time complexity, and using `T::DbWeight::get().reads_writes(1, 1)` for storage access? See https://centrifuge.hackmd.io/Mb8IAGhHQ32DEl1GIoBkCQ#Action-items for details.

runtime/src/impls.rs
```diff
@@ -30,89 +31,102 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
...
+/// 	LengthFee = ExtrinsicBytes * TransactionByteFee = 920000000000
+/// 	WeightFee = Weight * WeightCoefficient = 61425000000000
+/// 	Fee = BaseFee + LengthFee + WeightFee = 101720000000000 => 101.72 Micro RAD
+pub struct WeightToFee;
+impl WeightToFeePolynomial for WeightToFee {
+	type Balance = Balance;
+
+	fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
+		smallvec!(WeightToFeeCoefficient {
+			coeff_integer: 315000,
```
philipstanislaus, line 64: I noticed that you came to a different weight to fee constant (315000) from my calculation (625000) – are you sure your calculation is correct? I haven’t looked into it yet, just want to make sure that our fees will not suddenly be much cheaper than before.

runtime/src/lib.rs
```diff
@@ -94,10 +105,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // and set impl_version to 0. If only runtime
     // implementation changes and behavior does not, then leave spec_version as
     // is and increment impl_version.
-    spec_version: 232,
+    spec_version: 233,
     impl_version: 0,
     apis: RUNTIME_API_VERSIONS,
-    transaction_version: 0,
+    transaction_version: 1,
```
philipstanislaus, line 111: Why did this change?

philipstanislaus (changes requested):
Thanks @vedhavyas , just left a few comments regarding weights. 👍

runtime/src/anchor/mod.rs
vedhavyas: `T::DbWeight::get().reads_writes(1, 1)` implies we are doing only one read and one write for which is incorrect. Hence, I would like to keep our weights which were constant before as constants now to imply we are not considering the reads and writes but rather a chosen fixed weight @philipstanislaus

runtime/src/impls.rs
vedhavyas: Do verify this locally to ensure the fees are not lowered or made much cheaper than we are charging right now in mainnet @philipstanislaus

runtime/src/lib.rs
vedhavyas: It seems like this is the convention I saw in the node-template. i.e starting the transaction_verison from 1 - https://github.com/paritytech/substrate/blob/master/bin/node-template/runtime/src/lib.rs#L101 @philipstanislaus

runtime/src/anchor/mod.rs
philipstanislaus: @vedhavyas keeping the weights constant is a bad idea, since we most likely missprice our extrinsics compared to substrate. The issue with that is an attack vector – if you find a severely underpriced extrinsic, you could just send a load of them and the block can't be executed within the time limit (Not sure what happens then though?). 

Could you create a ticket to run benchmarks and use the actual reads/writes? Maybe link to that issue: https://github.com/paritytech/substrate/issues/6168

runtime/src/impls.rs
philipstanislaus: @vedhavyas done, I confirmed locally that anchor commits and balance transfers stay roughly the same. Thanks!

philipstanislaus (approved):
Thanks :+1:

PR #267, opened 2020-08-18
Disable multiaccount creation + rc3 prep

mikiquantum:


runtime/src/lib.rs
```diff
@@ -94,7 +94,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // and set impl_version to 0. If only runtime
     // implementation changes and behavior does not, then leave spec_version as
     // is and increment impl_version.
-    spec_version: 231,
```
vedhavyas, line 97: since we didn't touch runtime, we do not need to increment this. I was holding off incrementing this in my anchor prefix migration. WDYT ?

vedhavyas (approved):
LGTM
Small suggestions

runtime/src/lib.rs
mikiquantum: Unfortunately we need to increase this everytime due to the travis runtime version check. In any case since this is a behavioral change (create multiaccount is disabled) I would increase it.

PR #270, opened 2020-09-09
v2.0.0-RC6

vedhavyas:
Forked substrate to add our specific migrations. Taken a few migrations from Edgeware.

Migrations: https://github.com/centrifuge/substrate/tree/feat/migrations
GSRPC changes: https://github.com/centrifuge/go-substrate-rpc-client/pull/95

What are missing now:
- [x] Need to make a list of Accounts that need to be migrated. Have to reach out to folks how they did it. Maybe we can use subscan for that. Any thoughts ?
- [x] Anchor migrations
- [x] remove hardcoded addresses from the lib after adding the accounts
- [x] proxy pallet
- [x] multisig pallet
- [x] 7 day council term


runtime/src/impls.rs
```diff
@@ -72,324 +72,3 @@ impl WeightToFeePolynomial for WeightToFee {
...
-			.into_iter()
-			.for_each(|i| {
-				run_with_system_weight(i, || {
-					let fm = runtime_multiplier_update(max_fm);
-					// won't grow. The convert saturates everything.
-					assert_eq!(fm, max_fm);
-				})
-			});
-	}
-}
```
philipstanislaus, line 395: Why remove all these tests?

runtime/src/lib.rs
```diff
@@ -177,7 +177,8 @@ impl frame_system::Trait for Runtime {
 	/// Data to be associated with an account (other than nonce/transaction counter, which this
 	/// module does regardless).
 	type AccountData = pallet_balances::AccountData<Balance>;
-	/// Handler for when a new account has just been created.
+    type MigrateAccount = (Balances, Identity, Democracy, Elections, ImOnline, Staking, Session);
```
philipstanislaus, line 180: Please add a comment what this is for

runtime/src/lib.rs
```diff
@@ -793,8 +794,54 @@ pub type UncheckedExtrinsic = generic::UncheckedExtrinsic<Address, Call, Signatu
...
+                hex!["a23153e26c377a172c803e35711257c638e6944ad0c0627db9e3fc63d8503639"].into(),
+                hex!["a23153e26c377a172c803e35711257c638e6944ad0c0627db9e3fc63d8503639"].into(),
+
+                hex!["42a6fcd852ef2fe2205de2a3d555e076353b711800c6b59aef67c7c7c1acf04d"].into(),
+                hex!["42a6fcd852ef2fe2205de2a3d555e076353b711800c6b59aef67c7c7c1acf04d"].into(),
+                hex!["be1ce959980b786c35e521eebece9d4fe55c41385637d117aa492211eeca7c3d"].into(),
+                hex!["42a6fcd852ef2fe2205de2a3d555e076353b711800c6b59aef67c7c7c1acf04d"].into(),
+                hex!["42a6fcd852ef2fe2205de2a3d555e076353b711800c6b59aef67c7c7c1acf04d"].into(),
+                hex!["42a6fcd852ef2fe2205de2a3d555e076353b711800c6b59aef67c7c7c1acf04d"].into(),
+            ]
```
philipstanislaus, line 838: Please add a comment why there are hardcoded addresses here. Thanks!

runtime/src/impls.rs
vedhavyas: We are not maintaining the code for this. I would delegate this to parity for the Fee multiplier testing. These are orphaned tests

runtime/src/lib.rs
```diff
@@ -811,35 +812,14 @@ mod custom_migration {
             let accounts: Vec<AccountId> = Self::get_accounts();
             staking_upgrade::<Runtime>();
             accounts_upgrade::<Runtime>(accounts);
+            Anchor::migrate_anchors();
```
vedhavyas, line 815: Since this doesn't require any migration dependencies, let move this to `on_runtime_upgrade` in the anchor module ?

vedhavyas (commented):
LGTM!

mikiquantum (approved):
LGTM! I will do spend some time trying to get v2.0.0 release in, otherwise this seems to be the final version.

runtime/src/lib.rs
```diff
@@ -188,12 +191,74 @@ impl frame_system::Trait for Runtime {
...
 	// One storage item; value is size 4+4+16+32 bytes = 56 bytes.
-	pub const MultisigDepositBase: Balance = 30 * CENTI_RAD;
+	pub const DepositBase: Balance = 30 * CENTI_RAD;
 	// Additional storage item size of 32 bytes.
-	pub const MultisigDepositFactor: Balance = 5 * CENTI_RAD;
+	pub const DepositFactor: Balance = 5 * CENTI_RAD;
 	pub const MaxSignatories: u16 = 100;
 }
 
+impl pallet_multisig::Trait for Runtime {
```
mikiquantum, line 200: @vedhavyas added MultiSig Pallet, since in newer releases it was moved out of the utility one.

runtime/src/lib.rs
```diff
@@ -188,12 +191,74 @@ impl frame_system::Trait for Runtime {
...
+	// One storage item; value is size 4+4+16+32 bytes = 56 bytes.
+	pub const ProxyDepositBase: Balance = 30 * CENTI_RAD;
+	// Additional storage item size of 32 bytes.
+	pub const ProxyDepositFactor: Balance = 5 * CENTI_RAD;
+	pub const MaxProxies: u16 = 32;
+}
+
+/// The type used to represent the kinds of proxying allowed.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, RuntimeDebug)]
+pub enum ProxyType {
```
mikiquantum, line 220: @vedhavyas Added proxy type, since we are disabling the multiaccount creation, so we do not lose functionality.

PR #274, opened 2020-10-15
NFT and Registry modules

jaybutera:
Integrating NFT and VARegistry modules.
Supporting transfer, create_registry, and mint endpoints.


runtime/src/bridge.rs
```diff
@@ -86,6 +91,36 @@ decl_module! {
             Ok(())
         }
 
+        /// Transfer an nft to a whitelisted destination chain. Source nft is locked in bridge account
+        /// rather than being burned.
+        #[weight = 195_000_000]
+        pub fn transfer_asset(origin,
```
mikiquantum, line 97: Is there any reason why not naming this function `transfer_nft` ?

runtime/src/nft/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+decl_module! {
+    pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+        type Error = Error<T>;
+        fn deposit_event() = default;
+
+        /// Transfer a asset to a new owner.
+        ///
+        /// The dispatch origin for this call must be the asset owner.
+        ///
+        /// This function will throw an error if the new owner already owns the maximum
```
mikiquantum, line 94: Is this comment still relevant?

runtime/src/registry/mod.rs
```diff
@@ -0,0 +1,264 @@
+//! This substrate pallet defines a Verifiable Attributes Registry
```
mikiquantum, line 1: Should we rename this Pallet, since `registry` is pretty generic. Maybe nft_registry?

runtime/src/registry/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+        InvalidMintingValues,
+    }
+}
+
+decl_module! {
+    pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+        type Error = Error<T>;
+        fn deposit_event() = default;
+
+        #[weight = 10_000]
```
mikiquantum, line 92: We should review all these weights before merging.

runtime/src/registry/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    fn create_new_registry_id() -> Result<RegistryId, dispatch::DispatchError> {
+        let id = <RegistryNonce>::get();
+
+        // TODO: Make a U160 type for RegistryId with the uint crate.
+        // Passing through U256 is inefficient and H160 is unneeded.
+        let mut res = Vec::<u8>::with_capacity(32);
+        unsafe { res.set_len(32); }
+        // U256 > H160 so no need for a checked_add
+        U256::from_big_endian(id.as_bytes())
```
mikiquantum, line 151: So the registry id would be formed as:
```
id = [1,0,0,0,0,0,...]
nplus1 = U256(id).add(1) = [2,0,0,0,0,...]
```
so it would be a byte array nonce, right?

runtime/src/registry/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    fn mint(caller: &<T as frame_system::Trait>::AccountId,
+            owner_account: &<T as frame_system::Trait>::AccountId,
+            asset_id: &Self::AssetId,
+            asset_info: T::AssetInfo,
+            mint_info: MintInfo<<T as frame_system::Trait>::Hash, H256>,
+    ) -> Result<(), dispatch::DispatchError> {
+        let (registry_id, token_id) = AssetIdRef::from(asset_id).destruct();
+        let registry_info = Registries::get(registry_id);
```
mikiquantum, line 197: Wouldn't this be None if registry_id doesnt exist? In which case the below check can be replace by is_ok_or(RegistryDoesNotExist).

runtime/src/registry/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+            Error::<T>::RegistryDoesNotExist
+        );
+
+        // --------------------------
+        // Type checking the document
+
+        // The registry field must be a proof with its value as the token id.
+        // If not, the document provided may not contain the data and would
+        // be invalid. The registry field is always in the last place.
+        let registry_prop = &mint_info.proofs[ mint_info.proofs.len()-1 ].property;
```
mikiquantum, line 211: To be more precise, the registry property is the last one after the custom properties already defined in the registry. We currently have disabled the ability to pass more proofs, but if we always ensure that the registry_property is always the next one after the custom ones, we do not have to modify this logic if we add more proofs later on.

runtime/src/registry/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+        ensure!(
+            // First check that property is long enough
+            registry_prop.len()                      >= 28 &&
```
mikiquantum, line 214: Since you have stored the registry property as a required attribute in the registry, why dont compare as a whole?
`registry_prop == reg_info.fields[0]`

runtime/src/registry/types.rs
```diff
@@ -0,0 +1,151 @@
...
+pub const NFTS_PREFIX: &'static [u8] = &[20, 0, 0, 0, 0, 0, 0, 1];
+
+/// A vector of bytes, conveniently named like it is in Solidity.
+pub type Bytes = Vec<u8>;
+
+/// Registries are identified using a nonce in storage.
+pub type RegistryId = H160;
+
+/// A cryptographic salt to be combined with a value before hashing.
+pub type Salt = Bytes;
```
mikiquantum, line 17: Salt is Byte32 array, not sure if you want to constrain it here.

runtime/src/bridge.rs
```diff
@@ -86,6 +91,36 @@ decl_module! {
...
+
+            // Lock asset by transfering to bridge account
+            let bridge_id = <chainbridge::Module<T>>::account_id();
+            let asset_id = AssetId(from_registry, token_id);
+            <nft::Module<T> as Unique>::transfer(&source, &bridge_id, &asset_id)?;
+
+            // Transfer instructions for relayer
+            let tid: &mut [u8] = &mut[0; 32];
+            token_id.to_big_endian(tid);
+            <chainbridge::Module<T>>::transfer_nonfungible(dest_id, resource_id, tid.to_vec(), recipient, vec![]/*assetinfo.metadata*/)
```
mikiquantum, line 121: As mentioned offline, I wanted to throw out there, if we needed to add a list of supported registries per chain_id, so there is more control on where to move the nft. Since as it stands, as soon as we whitelist another nft registry in ethereum, we would be able to "accidentally" use the resource_id of that registry instead of the correspondent one.
I believe that might be duty of the resource_id mapping that the bridge is responsible for, but lets put some thought into it.

mikiquantum (commented):
Overall is great, clean, following composibility and flexible to iterate from it. Some clarification comments only.
Great Job @jaybutera

runtime/src/bridge.rs
jaybutera: Asset over nft is the naming convention throughout the project. Inspired from Dan Forbes' naming convention to try to be more general. That being said I have no problem in changing this, let me know.

runtime/src/registry/mod.rs
jaybutera: How about va_registry?

runtime/src/registry/mod.rs
jaybutera: That's correct. The RegistryId is still an H160. I wanted to make a U160 type for efficiency and ease of use, but had version conflicts with Substrate when importing the shared crate, uint. May take a while to resolve and probably isn't critical.

runtime/src/registry/mod.rs
jaybutera: I've been meaning to look deeper into this. For some reason, the StorageMap interface does not provide a try_get() - only get - [frame_support/storage/trait.StorageMap](https://substrate.dev/rustdocs/v2.0.0-rc6/frame_support/storage/trait.StorageMap.html).

I believe if a value doesn't exist, the default is returned (which is why a value must impl Default). Though I'm having trouble finding proof beyond [this point in the knowledge base](https://substrate.dev/docs/en/knowledgebase/runtime/storage#default-values).

It may be faster to compare the result to the default value but `contains_key` feels safer until I know for sure about the default values.

runtime/src/registry/mod.rs
jaybutera: Good point. Is there any reason to worry about the padding here like there is in the solidity version?

Is it always an exact match or should it be
```registry_prop[..28] == reg_info.fields[n]```

File: runtime/src/nft/mod.rs (hunk 2 above)
Comment (jaybutera) [reply runtime/src/nft/mod.rs]: Nope, removed 👍

File: runtime/src/registry/mod.rs (hunk 7 above)
Comment (jaybutera) [reply runtime/src/registry/mod.rs]: Updated 👍

File: runtime/src/registry/types.rs (hunk 9 above)
Comment (jaybutera) [reply runtime/src/registry/types.rs]: Updated 👍

File: runtime/src/bridge.rs (hunk 10 above)
Comment (jaybutera) [reply runtime/src/bridge.rs:121]: Ok from my understanding, we need a mapping `m: C -> E`
```
E = set of Tinlake contracts (on Ethereum)
C = set of registries on Centrifuge
```
Not every registry necessarily maps to a contract in `E`, and its possible that multiple registries map to the same contract, so `m` is not injective. But every Tinlake contract in `E` has a corresponding registry in `C` so `m` is surjective.

So a substrate StorageMap `ResourceId => ResourceId` to represent `m` would be sufficient.

We can put this into the centrifuge `src/bridge.rs` and allow access through the bridge admin account. I'm not sure how the proposal process works in the parent chainbridge-pallet and if that can be integrated to?

File: runtime/src/bridge.rs [hunk 11]
```diff
@@ -86,6 +91,36 @@ decl_module! {
...
+                              recipient: Vec<u8>,
+                              from_registry: RegistryId,
+                              token_id: TokenId,
+                              resource_id: ResourceId,
+                              dest_id: chainbridge::ChainId,
+        ) -> DispatchResult {
+            let source = ensure_signed(origin)?;
+
+            // Chain must be whitelisted
+            ensure!(<chainbridge::Module<T>>::chain_whitelisted(dest_id), Error::<T>::InvalidTransfer);
```
mikiquantum, line 107: This check is already done in the `transfer_nonfungible` function call below.

runtime/src/bridge.rs
mikiquantum: Yes that is correct. TInlake is just a consumer of the NFTs, so we can simplify to:
```
E = set of nft registries on Ethereum
C = set of registries on Centrifuge
```
I believe that it should be duty of the bridge contracts/pallets admin'd by the relayers (chainbridge-pallet and erc721-handler) that is whitelisted by the admin operators. Since eventually, the bridge will keep track of where the assets are moving to/from, sounds like a first step towards it.

A Centrifuge Asset can only be minted in the centrifuge protocol, so lets call it the NFT source of truth, so any move will always originate in the cent chain first `C -> E`. Since we always lock the asset before sending, when the NFT comes back `E -> C` (if it does), a check can be performed to verify that. the token_id exists in the desired registry, and owned by the bridge address, but that sounds like the duty of the centrifuge receiver function in `src/bridge.rs`. 

So a whitelist storage map in both sides `centrifuge-chain` and `ethereum`, and that means that we would need to modify the chainbridge-substrate pallet and the chainbridge erc721-handler contract.
That would mean that we need a new deployment on ethereum, and therefore a cascade of changes around most of the layers, which is a bit unfortunate.
There are some alternative solutions, and I would like put more thought. Let sync offline about it.

runtime/src/lib.rs
```diff
@@ -795,6 +804,46 @@ impl chainbridge::Trait for Runtime {
...
+
+impl nft::Trait for Runtime {
+    type Event = Event;
+    type AssetInfo = va_registry::types::AssetInfo;
+}
+
+impl bridge_names::Trait for Runtime {
+    type ResourceId = bridge::ResourceId;
+    type Address = bridge::Address;
+    type Admin = frame_system::EnsureRoot<Self::AccountId>;
```
mikiquantum, line 819: I would add here root or collective with 3/4 majority. See:
- https://github.com/centrifuge/centrifuge-chain/blob/master/runtime/src/lib.rs#L807
- https://github.com/centrifuge/centrifuge-chain/blob/master/runtime/src/rad_claims.rs#L174

runtime/src/lib.rs
```diff
@@ -834,6 +883,10 @@ construct_runtime!(
         Scheduler: pallet_scheduler::{Module, Call, Storage, Event<T>},
         Proxy: pallet_proxy::{Module, Call, Storage, Event<T>},
 		Multisig: pallet_multisig::{Module, Call, Storage, Event<T>},
+		Registry: va_registry::{Module, Call, Storage, Event<T>},
+		Nft: nft::{Module, Call, Storage, Event<T>},
```
mikiquantum, line 887: Please add the new pallets always at the end, otherwise it can screw up the metadata indices

runtime/src/nft/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn transfer(
+        caller: &T::AccountId,
+        dest_account: &T::AccountId,
+        asset_id: &AssetId,
+    ) -> dispatch::DispatchResult {
+        let owner = Self::owner_of(asset_id);
+        let (registry_id, token_id) = AssetIdRef::from(asset_id).destruct();
+
+        // Check that owner account exists
+        ensure!(owner != T::AccountId::default(),
```
mikiquantum, line 128: Does this check if the destination account exists as of has the existential deposit?

pallets/bridge-names/src/lib.rs
```diff
@@ -0,0 +1,100 @@
+//! # Bridge Access Control List Pallet
```
mikiquantum, line 1: Maybe rename to `bridge-mappings`

mikiquantum (approved):
Overall looks good!

runtime/src/nft/mod.rs
jaybutera: `AccountId::default()` is the return value if no owner exists. The approach is confusing so I switched to using an Option type check and removed this line.

PR #277, opened 2020-11-04
Add RAD Claims and Vesting Pallet

mikiquantum:


src/chain_spec.rs
```diff
@@ -267,6 +267,7 @@ pub fn testnet_genesis(
                 2_365_296_803_653,
             )],
         }),
+		pallet_vesting: Some(Default::default()),
```
mikiquantum, line 270: @vedhavyas Do you know if I need to modify the `*-raw-spec.json` files to add the empty definition of the `pallet_vesting` like above?

runtime/src/rad_claims.rs
```diff
@@ -0,0 +1,494 @@
...
+        /// The extrinsic is validated by the custom `validate_unsigned` function below
+        ///
+        /// # <weight>
+        /// - Based on hashes length
+        /// # </weight>
+        #[weight = 195_000_000]
+        pub fn claim(origin,
+                     account_id: T::AccountId,
+                     amount: T::Balance,
+                     sorted_hashes: Vec<T::Hash>,
```
vedhavyas, line 93: I believe it is important to verify the proofs on each node. With the current flow, only the propagator node will verify that.

runtime/src/rad_claims.rs
```diff
@@ -0,0 +1,494 @@
...
+                .fold(leaf_hash, |acc, hash| Self::sorted_hash_of(&acc, hash));
+
+            // Initial runs might only have trees of single leaves,
+            // in this case leaf_hash is as well root_hash
+            if sorted_hashes.len() == 0 {
+                root_hash = leaf_hash;
+            }
+
+            // Check that root exists in root hash storage
+            if Self::get_root_hash(root_hash) == true {
```
vedhavyas, line 217:
```suggestion
            if Self::get_root_hash(root_hash) {
```

runtime/src/rad_claims.rs
```diff
@@ -0,0 +1,494 @@
...
+
+pub trait Trait: frame_system::Trait + pallet_balances::Trait {
+    type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
+
+    /// An expected duration of the session.
+    ///
+    /// This parameter is used to determine the longevity of `heartbeat` transaction
+    /// and a rough time when we should start considering sending heartbeats,
+    /// since the workers avoids sending them at the very beginning of the session, assuming
+    /// there is a chance the authority will produce a block and they won't be necessary.
```
vedhavyas, line 29: Seems like the comment needs updating

runtime/src/rad_claims.rs
```diff
@@ -0,0 +1,494 @@
...
+
+            // Set account balance to amount
+            AccountBalances::<T>::insert(account_id.clone(), amount);
+
+            // Transfer payout amount
+            <pallet_balances::Module<T> as Currency<_>>::transfer(
+                &source,
+                &account_id,
+                payout,
+                AllowDeath,
```
vedhavyas, line 120: Not sure we want to `AllowDeath` for the source. Some dust amount might be lost due to this

runtime/src/rad_claims.rs
```diff
@@ -0,0 +1,494 @@
...
+
+            Ok(())
+        }
+
+    }
+}
+
+impl<T: Trait> Module<T> {
+    /// Hash a:b if a < b, else b:a. Uses the runtime module's hasher.
+    pub fn sorted_hash_of(a: &T::Hash, b: &T::Hash) -> T::Hash {
```
vedhavyas, line 164: We should repurpose the proofs module for verification.

runtime/src/rad_claims.rs
```diff
@@ -0,0 +1,494 @@
...
+            // in this case leaf_hash is as well root_hash
+            if sorted_hashes.len() == 0 {
+                root_hash = leaf_hash;
+            }
+
+            // Check that root exists in root hash storage
+            if Self::get_root_hash(root_hash) == true {
+                return ValidTransaction::with_tag_prefix("RADclaim")
+                    .priority(T::UnsignedPriority::get())
+                    .longevity(TryInto::<u64>::try_into(
```
vedhavyas, line 220: Does this mean the transaction is valid for `SessionDuration/2` ?

vedhavyas (commented):
Few questions.

The previous substrate commit has migrations that need to be removed. Can you update the substrate commit to rc6 without migrations?

runtime/src/rad_claims.rs
jaybutera: I agree, it would be nice to have a function to generate a document root. Right now, building the tree is coupled with verifying it in the proofs module.

For the `Proofs::sorted_hash_of`, it uses H256, but rad claims merkle tree will be encoded with the runtime hash function. They are both blake2 but H256 can't be elevated to T::Hash associated type. It seems cleaner to just redefine the function rather than change the proofs module and coerce down to H256.

runtime/src/rad_claims.rs
mikiquantum: Yep since it is the keyless account, I think it makes sense to KeepAlive. I would have to add a check above to include the existential_deposit amount in the balance check above.

runtime/src/rad_claims.rs
mikiquantum: Yes, I was thinking that it might be too long, I would like to specify a lifetime that is closer to a signed extrinsic one (which seems that is around 64 blocks)

jaybutera:
https://github.com/centrifuge/centrifuge-chain/blob/23d883bcf691bb66202a32d2889d25ce555f457e/runtime/src/rad_claims.rs#L110-L121

By transferring before setting `AccountBalances` we can remove InsufficientBalance check as its done by transfer.

mikiquantum:
> https://github.com/centrifuge/centrifuge-chain/blob/23d883bcf691bb66202a32d2889d25ce555f457e/runtime/src/rad_claims.rs#L110-L121
> 
> By transferring before setting `AccountBalances` we can remove InsufficientBalance check as its done by transfer.

Yep, that is correct. I initially wanted to be sure that the transfer is not done before all the account is set, but the:
`AccountBalances::<T>::insert(account_id.clone(), amount);`
Is very rare to fail, so I will swap it and remove the check 👍

runtime/src/rad_claims.rs
mikiquantum: Adding the check, wont harm anyways.

mikiquantum:
@vedhavyas & @jaybutera addressed comments. 
Only left to do is to use the latest RC6 from the centrifuge fork. Will add this today.

mikiquantum:
@vedhavyas Just removed all the migrations and now we are using the centrifuge substrate fork. 
One question, should we remove as well the child.rs and the custom host functions?

vedhavyas:
@mikiquantum just using the original rc6 fork is good enough. The host functions are required for syncing from the scratch. We could get rid of them once we do the parachain migration if possible

vedhavyas (approved):

PR #278, opened 2020-11-16
feat: add treasury pallet

manuelpolzhofer:


runtime/src/lib.rs
```diff
@@ -109,7 +109,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // implementation changes and behavior does not, then leave spec_version as
     // is and increment impl_version.
     spec_version: 237,
-    impl_version: 0,
+    impl_version: 1,
```
mikiquantum, line 112: Arguably would increment here the `spec_version`. Since adding a new pallet, changes the metadata and supported types by clients.

runtime/src/lib.rs
```diff
@@ -767,6 +767,39 @@ impl substrate_pallet_multi_account::Trait for Runtime {
     type MaxSignatories = MultiAccountMaxSignatories;
 }
 
+parameter_types! {
+    pub const TreasuryModuleId: ModuleId = ModuleId(*b"cen/trsy");
```
mikiquantum, line 771: What are `kusama` settings for all these? I would take them as a reference.

runtime/src/lib.rs
manuelpolzhofer: changed 👍

runtime/src/lib.rs
manuelpolzhofer: I reused the default values from kusama. Should we discuss the values with Cassidy? 
I think for example the TipFindersFee shouldn't be 20%.

vedhavyas:
@manuelpolzhofer We need to finalise the ways to fill the treasury through slashing etc...
Polkadot source maybe a good start on how the treasury is used

runtime/src/lib.rs
manuelpolzhofer: I created a hackMD to discuss all config parameter

runtime/src/lib.rs
```diff
@@ -767,6 +767,53 @@ impl substrate_pallet_multi_account::Trait for Runtime {
...
+	type ProposalBond = ProposalBond;
+	type ProposalBondMinimum = ProposalBondMinimum;
+	type SpendPeriod = SpendPeriod;
+	type Burn = Burn;
+	type Tippers = Elections;
+	type TipCountdown = TipCountdown;
+	type TipFindersFee = TipFindersFee;
+	type TipReportDepositBase = TipReportDepositBase;
+	type TipReportDepositPerByte = TipReportDepositPerByte;
+	type BurnDestination = ();
```
mikiquantum, line 813: What is the BurnDestination used for? And what is the default value.

runtime/src/lib.rs
```diff
@@ -11,19 +11,19 @@ use frame_support::{
         Weight,
         constants::{BlockExecutionWeight, ExtrinsicBaseWeight, RocksDbWeight, WEIGHT_PER_SECOND},
     },
-    traits::{Currency, KeyOwnerProofSystem, Randomness, LockIdentifier, InstanceFilter},
+    traits::{Currency, KeyOwnerProofSystem, Randomness, LockIdentifier, InstanceFilter },
```
mikiquantum, line 14: remove extra space plz

runtime/src/lib.rs
```diff
@@ -767,6 +767,53 @@ impl substrate_pallet_multi_account::Trait for Runtime {
     type MaxSignatories = MultiAccountMaxSignatories;
 }
 
+parameter_types! {
+	pub const SpendPeriod: BlockNumber = 6 * DAYS;
+	pub const Burn: Permill = Permill::from_perthousand(0);
+	pub const TreasuryModuleId: ModuleId = ModuleId(*b"py/trsry");
+
+	pub const TipCountdown: BlockNumber = 1 * DAYS;
```
mikiquantum, line 775: What happens after the day is over? is the tip lost?

runtime/src/lib.rs
```diff
@@ -767,6 +767,53 @@ impl substrate_pallet_multi_account::Trait for Runtime {
     type MaxSignatories = MultiAccountMaxSignatories;
 }
 
+parameter_types! {
+	pub const SpendPeriod: BlockNumber = 6 * DAYS;
```
mikiquantum, line 771: What does the SpendPeriod do? Is it how often they can be issued?

runtime/src/lib.rs
manuelpolzhofer: A percentage of the treasury funds can burned to have some kind of inflation mechanism. 
Cassidy suggested not to use it for now.

The `BurnDestination` would be handler to change were the burned amount should go. 
Our inflation is currently zero percentage, so it doesn't matter.

runtime/src/lib.rs
manuelpolzhofer: After a tip receives the threshold of tippers, currently (3/5 council) the tip countdowns starts.
It allows the remaining council members to place a tip as well which would be considered in the average.

runtime/src/lib.rs
manuelpolzhofer: Yes, it is the period between successive spends.

runtime/src/lib.rs
manuelpolzhofer: Yes, it is the period between successive spends.

NunoAlexandre:
Hi @manuelpolzhofer 👋 

This PR is stale for over a year, closing it for now.

manuelpolzhofer:
@NunoAlexandre all good.

PR #279, opened 2020-11-20
Fix hashing + tx validity tag

mikiquantum:


runtime/src/rad_claims.rs
```diff
@@ -167,7 +170,7 @@ impl<T: Trait> Module<T> {
             h.extend_from_slice(a.as_ref());
         }
 
-        T::Hashing::hash_of(&h).into()
+        T::Hashing::hash(&h).into()
```
mikiquantum, line 173: `T::Hashing::hash_of` scale encodes the payload before hashing. Since these are proofs generated out of a merkle tree, using plain concatenation and hashing, scale encoding breaks the verification.

runtime/src/rad_claims.rs
```diff
@@ -74,6 +74,9 @@ decl_event! {
 
 decl_module! {
     pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+
+        const RadClaimsAccountId: T::AccountId = MODULE_ID.into_account();
```
mikiquantum, line 78: Easy way to know the RadClaims AccountID

runtime/src/rad_claims.rs
```diff
@@ -215,8 +218,9 @@ impl<T: Trait> frame_support::unsigned::ValidateUnsigned for Module<T> {
         if let Call::claim(account_id, amount, sorted_hashes) = call {
             // Check that proofs are valid with a root that exists in the root hash storage
             if Self::verify_proofs(account_id, amount, sorted_hashes.into()) {
-                return ValidTransaction::with_tag_prefix("RADclaim")
+                return ValidTransaction::with_tag_prefix("RadClaims")
                     .priority(T::UnsignedPriority::get())
+                    .and_provides((account_id, amount, sorted_hashes))
```
mikiquantum, line 223: If we do not provide these tags, then other nodes will not be able to identify the correct tx to propagate.

vedhavyas (approved):

PR #290, opened 2021-01-21
runtime/Cargo.toml: fix wasm-builder-runner version

asymmetric:
The `wasm-builder-package` at the specified commit has version 1.0.6:

https://github.com/centrifuge/substrate/blob/e271ae7fda72a7ad55ff3c216b8b2abae0ae31fd/utils/wasm-builder-runner/Cargo.toml#L3

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -77,7 +77,7 @@ pallet-vesting = { version = "2.0.0-rc6", git = "https://github.com/centrifuge/s
 frame-benchmarking = { version = "2.0.0-rc6", git = "https://github.com/centrifuge/substrate.git", rev = "e271ae7fda72a7ad55ff3c216b8b2abae0ae31fd", default-features = false, optional = true }
 
 [build-dependencies]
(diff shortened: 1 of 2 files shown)


vedhavyas:
@asymmetric shall we close this in favor of #291 ?

asymmetric:
@vedhavyas sure, but do you agree that the version of `wasm-builder-runner` at that commit is 1.0.6? If so, then I'm not entirely sure why things compile when specifying 1.0.5. But they seem to, so :man_shrugging:

vedhavyas:
@asymmetric agree with you, I pulled 1.0.5 from the node template(https://github.com/centrifuge/substrate/blob/master/bin/node-template/runtime/Cargo.toml#L46), used to create and upgrade versions, which is still at 1.0.5.

we can use 1.0.6 as well. Could you please increment `impl_version` so that tests pass and we can merge this one up

asymmetric:
@vedhavyas the Docker Hub error seems unrelated. Is this good to go?

PR #291, opened 2021-02-14
Added nixos flake for binary and container

lucasvo:


runtime/Cargo.toml
```diff
@@ -77,7 +77,7 @@ pallet-vesting = { version = "2.0.0-rc6", git = "https://github.com/centrifuge/s
 frame-benchmarking = { version = "2.0.0-rc6", git = "https://github.com/centrifuge/substrate.git", rev = "e271ae7fda72a7ad55ff3c216b8b2abae0ae31fd", default-features = false, optional = true }
 
 [build-dependencies]
-wasm-builder-runner = { version = "1.0.5", package = "substrate-wasm-builder-runner", git = "https://github.com/centrifuge/substrate.git", rev = "e271ae7fda72a7ad55ff3c216b8b2abae0ae31fd" }
+wasm-builder-runner = { version = "1.0.6", package = "substrate-wasm-builder-runner", git = "https://github.com/centrifuge/substrate.git", rev = "e271ae7fda72a7ad55ff3c216b8b2abae0ae31fd" }
```
vedhavyas, line 80: anything specific about this version or just a regular upgrade?
Travis is failing because this is a runtime file and any runtime change should accompany with spec and/or impl version incremented.

lucasvo: The previous version didn't properly build in nixos. I'm not entirely sure why. Maybe @asymmetric can say more why we needed to do this?

lucasvo: @vedhavyas how would we go about bumping the impl/spec version? Perhaps we can bundle this with a few other runtime changes?

lucasvo: this is running on our mainnet validator just fine btw

vedhavyas: > @vedhavyas how would we go about bumping the impl/spec version? Perhaps we can bundle this with a few other runtime changes?

That would too much of a hassle. Just increment the impl_version in lib.rs to 1 and should be fine - https://github.com/centrifuge/centrifuge-chain/blob/master/runtime/src/lib.rs#L121

vedhavyas: that implies nothing on the runtime changed logically. next runtime changes will increment spec_version accordingly

asymmetric: Can't really reproduce the issue I was having. @lucasvo this [built](https://github.com/asymmetric/centrifuge-chain/runs/1918868346?check_suite_focus=true) for me with 1.0.5, so the bump looks unnecessary.

vedhavyas (approved):
LGTM!

PR #299, opened 2021-03-22
Runtime Upgrade 4

mikiquantum:
This upgrade bundles:

- Removes migration code from previous upgrade
- Renaming RAD -> CFG
- Create new ProxyType Vesting
- Make Bridge Fungible Fee configurable


runtime/src/lib.rs
```diff
@@ -231,19 +231,38 @@ pub enum ProxyType {
...
+				Call::Elections(..) |
+				Call::Utility(..)
 			),
-            ProxyType::Staking => matches!(c, Call::Staking(..)),
+            ProxyType::Staking => matches!(c,
+                Call::Staking(..) |
+                Call::Session(..) |
+				Call::Utility(..)
+            ),
+            ProxyType::Vesting => matches!(c,
```
mikiquantum, line 256: @lucasvo this is the new ProxyType that will be used for vesting

runtime/src/nft/mod.rs
```diff
@@ -159,7 +160,12 @@ impl<T: Trait>
 
         // Insert into storage
         AccountForAsset::<T>::insert(registry_id, token_id, owner_account);
-        Assets::<T>::insert(registry_id, token_id, asset_info);
+
+        //TODO(mig) Remove once we have storage fee for metadata
+        let empty_info: <T as Trait>::AssetInfo = AssetInfo {
```
mikiquantum, line 165: @vedhavyas if you have a better idea to create an empty value, let me know.

vedhavyas: You can use `default` since its already a trait implemented for AssetInfo trait

runtime/src/bridge.rs
```diff
@@ -365,6 +386,7 @@ mod tests{
 
 	parameter_types! {
 		pub HashId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &blake2_128(b"hash"));
+		//TODO rename xRAD to xCFG and create new mapping
 		pub NativeTokenId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &blake2_128(b"xRAD"));
```
vedhavyas, line 390: What is the process to change the resourceID @mikiquantum ?
Does it require another Upgrade?

runtime/src/lib.rs
```diff
@@ -231,19 +231,38 @@ pub enum ProxyType {
...
+                Call::Staking(..) |
+                Call::Session(..) |
+                Call::Democracy(..) |
+				Call::Council(..) |
+				Call::Elections(..) |
+				Call::Vesting(pallet_vesting::Call::vest(..)) |
+				Call::Vesting(pallet_vesting::Call::vest_other(..)) |
+				Call::Utility(..)
```
vedhavyas, line 264: @mikiquantum so any transfer call will be filtered out if it's part of the utility batch?

runtime/src/nft/mock.rs
```diff
@@ -53,6 +54,12 @@ impl system::Trait for Test {
 #[derive(Clone, Eq, PartialEq)]
 pub struct Test;
 
+impl From<AssetInfo> for Vec<u8> {
```
vedhavyas, line 57: we do not require this since we can go with `Default` as mentioned below

runtime/src/nft/mod.rs
```diff
@@ -18,6 +18,7 @@ use frame_support::{
     traits::Get,
     Hashable,
 };
+use sp_std::prelude::*;
```
vedhavyas, line 21: can we import specifics instead?

vedhavyas (commented):
Overall lgtm
I can `xRAD` Todos. Do we require another upgrade to change those

runtime/src/bridge.rs
mikiquantum: Yes, but that will require updates on bridge mappings, so for now I wouldn't change it and do it when we upgrade bridge code.

runtime/src/nft/mod.rs
mikiquantum: thanks I will try it out

runtime/src/nft/mod.rs
mikiquantum: might not even need to if Default works

runtime/src/lib.rs
mikiquantum: @vedhavyas actually looking at the code now, it seems that is only checking the parent call, so that would be a way to bypass a balance transfer. 
I will remove it to be safe.

mikiquantum: That means is something that can be bypassed in Kusama as well then.

vedhavyas: I think balance transfer would error out since the un-vested balances are still locked but since its the utility call, any call will filter through. 
Makes sense to remove it 👍🏼 @mikiquantum

PR #301, opened 2021-03-26
Allow arguments to be passed to Docker image

asymmetric:
This uses `ENTRYPOINT` instead of `CMD`, which is the correct way to set the
image's "base" command, since it doesn't get overwritten by arguments passed to
`docker run`. 
See [here](https://aws.amazon.com/blogs/opensource/demystifying-entrypoint-cmd-docker/) for an explanation.

Additionally, this PR:

- moves the Docker-related parts of the workflow file behind a check, ensuring they only run on the upstream repo and not on forks
- fixes the Docker tag, adding hours, minutes and seconds to the timestamp
- prepends `/bin` to the `PATH`, so that Kubernetes `Deployement`s don't have to be updated.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,24 +1,13 @@
 {
   description = "Nix package for centrifuge-chain";
 
-  inputs = {
-    nixpkgs.url = github:NixOS/nixpkgs/nixos-20.09;
-    gitignore-nix = {
-      url = github:hercules-ci/gitignore.nix;
-      flake = false;
-    };
-  };
+  inputs.nixpkgs.url = github:NixOS/nixpkgs/nixos-20.09;
 
   outputs = inputs:
     let
       name = "centrifuge-chain";
       version = "2.0.0";
       pkgs = inputs.nixpkgs.legacyPackages.x86_64-linux;
-      gitignore = (
-        import inputs.gitignore-nix {
-          inherit (inputs.nixpkgs.legacyPackages.x86_64-linux) lib;
-        }
-      ).gitignoreSource;
     in
     {
       packages.x86_64-linux.centrifuge-chain =
@@ -26,7 +15,7 @@
           pname = name;
           version = version;
 
-          src = gitignore ./.;
+          src = inputs.self;
 
           cargoSha256 = "sha256-52CN7N9FQiJSODloo0VZGPNw4P5XsaWfaQxEf6Nm2gI=";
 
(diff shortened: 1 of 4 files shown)


asymmetric:
@lucasvo @mikiquantum

asymmetric:
Note though that `docker run -it --rm centrifugeio/centrifuge-chain --dev` errors with:

```
Thread 'main' panicked at 'Development wasm binary is not available. This means the client is built with `BUILD_DUMMY_WASM_BINARY` flag and it is only usable for production chains. Please rebuild with the flag disabled.', runtime/src/lib.rs:106

This is a bug. Please report it at:

	centrifuge.io
```

This is a separate issue which has to do with how the Nix package is written, so let's create a separate issue if it's a concern.

asymmetric:
OT: It would be good to gate operations that will certainly fail on external PRs, so builds don't fail needlessly.

flake.nix
```diff
@@ -1,53 +1,69 @@
...
+      packages.x86_64-linux.dockerContainer =
         pkgs.dockerTools.buildImage {
           name = "centrifugeio/${name}";
           tag = "latest";
 
+          contents = inputs.self.defaultPackage.x86_64-linux;
+
           config = {
+            Env = [
+              "PATH=/bin/centrifuge-chain"
```
mikiquantum, line 54: I believe that here the PATH needs to point to `/bin` right? Otherwise it will try to load `centrifuge-chain` as a folder and is the actual executable.

flake.nix
```diff
@@ -1,53 +1,69 @@
...
+              "30333/tcp" = { };
+              "9933/tcp" = { };
+              "9944/tcp" = { };
             };
             Volumes = {
-                "/data" = {};
+              "/data" = { };
             };
-            Cmd = [ "${self.defaultPackage.x86_64-linux}/bin/centrifuge-chain" ];
+            Entrypoint = [ "/bin/centrifuge-chain" ];
```
mikiquantum, line 64: Can we use here the binary name only without the `/bin`, since it would be in the path after the above is adjusted.
```suggestion
            Entrypoint = [ "centrifuge-chain" ];
```

flake.nix
asymmetric: You are of course right :)

flake.nix
asymmetric: Done.

PR #304, opened 2021-03-31
Fix docker actions

mikiquantum:
Ensure that image is built and pushed only on commits to master.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,43 @@
...
+      - uses: cachix/cachix-action@v8
+        with:
+          name: centrifuge-chain
+          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
+      - run: nix build -L .#dockerContainer
+      - run: docker load -i result
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF##*/})"
+        id: extract_branch
```
asymmetric, line 28: Are we using this for something?

mikiquantum: Not anymore, we can get rid of it.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
+name: docker
+on:
+  push:
+    branches: [master]
```
asymmetric, line 4: @mikiquantum should we instead do this only when tags are pushed? We don't necessarily want to push a Docker tag on every commit that is merged into master (think of a PR only changing CI-related files for examples).

asymmetric: We've decided not to do this.

PR #311, opened 2021-04-01
CI: skip building when source hasn't changed

asymmetric:
This should ensure we don't trigger nix builds when not necessary, giving us much faster CI feedback.

We achieve this by ignoring the files listed in the `srcFilter` function, meaning that:

- they won't end up in the nix package
- a change to one of these files won't trigger a new nix build
- the respective docker image generation will be much faster

/cc @lucasvo

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,21 +1,5 @@
 {
   "nodes": {
-    "gitignore-nix": {
-      "flake": false,
-      "locked": {
-        "lastModified": 1611672876,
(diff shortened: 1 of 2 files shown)


vedhavyas (approved):
LGTM. Out of curiosity, how much CI time are we saving from earlier?

asymmetric:
@vedhavyas ~1hr :)

vedhavyas:
@asymmetric is there any more work to be done on this or are we good to merge this?

asymmetric:
@vedhavyas this is good to go. We might end up using something like [this](https://github.com/numtide/nix-filter) later on, but for now this is OK.

PR #338, opened 2021-04-22
First Implementation of the Reward Pallet

mustermeiszer:
The first implementation of the reward pallet must be tested with the claim module. Therefore the PR. 

Current state:
- Init of the pallet via root
- Reject of rewards, when min_vest of vesting pallet is not reached
- Funds are locked in the module
- Vesting period is defined via pallets GenesisConfig (globally)
- Conversion ratio from relay chain balance to native balance is defined via pallets GenesisConfig (globally)

The current state is covered by unit tests.


pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+// Ensure we're `no_std` when compiling for Wasm.
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use frame_support::{decl_module, decl_storage, decl_event, decl_error, weights::{Weight}, Parameter, ensure};
+use frame_system::{ensure_root, RawOrigin};
+use sp_runtime::traits::{Member, MaybeSerialize, Convert};
+use frame_support::sp_runtime::traits::{AtLeast32BitUnsigned, MaybeSerializeDeserialize};
+use frame_support::dispatch::{Codec, DispatchError};
+use frame_support::dispatch::fmt::Debug;
+use frame_support::traits::{Get, Currency, ExistenceRequirement::{AllowDeath }};
```
vedhavyas, line 46:
```suggestion
use frame_support::traits::{Get, Currency, ExistenceRequirement::{AllowDeath}};
```

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+const MODULE_ID: ModuleId = ModuleId(*b"rewards ");
+
+/// Callable functions (i.e. transaction) weight trait
+pub trait WeightInfo {
+    fn initialize() -> Weight;
+    fn reward() -> Weight;
+}
+
+// Define transaction weights to be used when testing
+pub struct TestWeightInfo;
```
vedhavyas, line 86: lets move this under mocks or tests

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+            let now = <frame_system::Module<T>>::block_number();
+            ensure!(now > <NextInit<T>>::get(), Error::<T>::OngoingLease);
+
+            let target = &MODULE_ID.into_account();
+            // TODO: Should we choose: KeepAlive or AllowDeath here?
+            T::Currency::transfer(
+                &source,
+                &target,
+                amount,
+                AllowDeath,
```
vedhavyas, line 146: ideal to keep it alive so that we don't lose dust and possibly re-use

runtime/src/lib.rs
```diff
@@ -237,7 +237,7 @@ impl frame_system::Config for Runtime {
 	/// A function that is invoked when an account has been determined to be dead.
 	///
 	/// All resources should be cleaned up associated with the given account.
-	type OnKilledAccount = ();
+	type OnKilledAccount = (); // TODO: Might implement something here?
```
vedhavyas, line 240: what do we want to cleanup when the account is killed ?

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+            T::Origin::from(RawOrigin::Signed(from)),
+            target,
+            schedule
+        ).map_err(|vest_err| {
+            match vest_err {
+                DispatchError::Module { index: _, error: _, message } => {
+                    message.map_or(vest_err, |msg| {
+                        match msg {
+                            "InsufficientBalance" => Into::<DispatchError>::into(Error::<T>::NotEnoughFunds),
+                            "AmountLow" => { vest_err
```
vedhavyas, line 188: Amount low is the minimum amount target gets, right? If we want to enforce the minimum amount, shouldn't we do the check before we reach here?

vedhavyas (commented):
question on this:
> No additional funding of module in ongoing lease period possible
The additional contribution you mean? or the balance on the module account?

runtime/src/lib.rs
mustermeiszer: Sorry. This was just a reminder for me, as I wanted to ask you if we want to do something there.

pallets/cf-reward/src/lib.rs
mustermeiszer: No, its the minimum amount that is necessary to create a vesting schedule. Currently this [1000*CFG](https://github.com/centrifuge/centrifuge-chain/blob/20dd92afbb64ef062543f0f8767dfe120ad6f96c/runtime/src/lib.rs#L841)

pallets/cf-reward/src/lib.rs
mustermeiszer: Definitely, yes.

pallets/cf-reward/src/lib.rs
mustermeiszer: I updated the current state. I first thought that we will be able to transfer any additional funds to the module. But with the module account being stored in the module's storage, this should be straightforward.

mustermeiszer:
I updated the current state. I first thought that we won't be able to transfer any additional funds to the module. But with the module account being stored in the module's storage, this should be straightforward.

runtime/src/lib.rs
vedhavyas: ahh cool. We don't kill any module accounts so we can ignore it

pallets/cf-reward/src/lib.rs
vedhavyas: I'm confused between `InsufficientBalance` and `AmountLow` now. Doesn't both represent same now ?

mustermeiszer: The first one refers to an error triggered by the balances pallet, and the second one to the vesting pallet. I am translating the `InsufficientBalance` error to a reward pallet specific one `NotEnoughFunds`. Thought this gives us more information.

vedhavyas:
@mustermeiszer can you rebase again?

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+        /// This dispatchable function must be called from a root account, so that other accounts
+        /// can not block the module via a random initialization. Furthermore it is important, that
+        /// the module has enough funds, as a later funding is not possible.
+        // TODO: Add correct weight function
+        #[weight = 10_000]
+        fn initialize(origin, source: T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+            ensure_root(origin)?;
+
+            let now = <frame_system::Module<T>>::block_number();
+            ensure!(now > <NextInit<T>>::get(), Error::<T>::OngoingLease);
```
mikiquantum, line 138: wouldn't this always fail in the first initialization? Since NextInit would be 0.

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+
+            let target = &MODULE_ID.into_account();
+            // TODO: Should we choose: KeepAlive or AllowDeath here?
+            T::Currency::transfer(
```
mikiquantum, line 142: I am wondering if we need to transfer the funds in this function or just send a tx to the module_id account?

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+
+decl_module! {
+    pub struct Module<T: Config> for enum Call where origin: <T as frame_system::Config>::Origin {
+        // Activate errors
+        type Error = Error<T>;
+
+        // Activate events
+        fn deposit_event() = default;
```
mikiquantum, line 126: Can we declare within this module the accountId for the reward module_id:
```
const RewardAccountId: T::AccountId = MODULE_ID.into_account();
```

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl<T: Config> Reward for Module<T> {
+    type ParachainAccountId = T::AccountId;
+    type ContributionAmount = T::RelayChainBalance;
+
+    fn reward(who: Self::ParachainAccountId, contribution: Self::ContributionAmount) -> DispatchResult {
+        let schedule = pallet_vesting::VestingInfo {
+            locked: Self::convert_to_native(contribution),
+            per_block: <<T as pallet_vesting::Config>::BlockNumberToBalance>::convert(T::VestingPeriod::get()),
```
mikiquantum, line 171: PerBlock here means the amount of tokens unlocked per block. Therefore that depends on each contribution amount and for how long are we vesting (LeasePeriod). 
So we need to change that to `per_block = contribution / n_total_blocks` and `n_total_blocks = LeasePeriod` I believe.

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+            schedule
+        ).map_err(|vest_err| {
+            match vest_err {
+                DispatchError::Module { index: _, error: _, message } => {
+                    message.map_or(vest_err, |msg| {
+                        match msg {
+                            "InsufficientBalance" => Into::<DispatchError>::into(Error::<T>::NotEnoughFunds),
+                            "AmountLow" => { vest_err
+                                // TODO: Should we do this? Or should we set MinVestAmount = 1?
+                                /*T::Currency::transfer(
```
mikiquantum, line 190: Something I am missing is the % transferred now vs the % transferred as a vesting schedule. Are we not doing that anymore?

pallets/cf-reward/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+authors = ['Centrifuge GmbH']
+description = 'The reward pallet that rewards contributors in bids for parachain slots for the centrifuge chain.'
+edition = '2018'
+license = "LGPL-3.0"
+name = 'cf-reward'
```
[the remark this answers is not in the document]
mustermeiszer: Done.

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+//! contributors of the relay chain crowdloan. In order to provide this functionality the pallet
+//! implements the `Reward` Trait from the `Claim` Pallet.
+//! Before any rewards can be provided to contributors the pallet MUST be initialized, so that the
+//! modules account holds the necessary funds to reward.
+//! All rewards are payed in form of a `vested_transfer` with a fixed vesting time. The vesting time
+//! is defined via the modules `Config` trait.
+//!
+//! ## Callable Functions
+//!
+//! - `initialize` - Initializes the module by transfering founds to the modules account and activating an init-lock
```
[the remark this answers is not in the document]
mustermeiszer: Done.

pallets/cf-reward/src/lib.rs
mustermeiszer: The only possibility I see would be before genesis. Depends on whether the system returns zero or one during this state.

pallets/cf-reward/src/lib.rs
mustermeiszer: Not sure what you mean?

pallets/cf-reward/src/lib.rs
mustermeiszer: Yes. Good idea.

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+        MaybeSerializeDeserialize + Debug +
+        Into<<<Self as pallet_vesting::Config>::Currency as Currency<<Self as frame_system::Config>::AccountId>>::Balance>;
+    /// AccountId of the relay chain
+    type RelayChainAccountId: Parameter + Member + MaybeSerializeDeserialize + Debug + MaybeSerialize + Ord
+        + Default;
+}
+
+type BalanceOf<T> = <<T as pallet_vesting::Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
+
+const MODULE_ID: ModuleId = ModuleId(*b"rewards ");
```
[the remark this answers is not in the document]
mustermeiszer: Done.

pallets/cf-reward/Cargo.toml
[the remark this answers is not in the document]
tankofzion: I will suggest that all our pallets being aligned with Polkadot/Substrate way of labelling pallets, namely `pallet-[name here]`, like for instance, `pallet-crowdloan-claim`, `pallet-crowdloan-reward`, and so on.

Cargo.toml
```diff
@@ -114,6 +115,10 @@ polkadot-cli = {git = 'https://github.com/paritytech/polkadot', branch = 'rococo
 node-runtime = { package = "centrifuge-chain-runtime", path = "runtime" }
 node-primitives = { branch = "rococo-v1", git = "https://github.com/paritytech/substrate.git" }
 
+# local dev dependencies
+cf-reward = { path = "pallets/cf-reward" }
+crowdloan-claim = { path = "pallets/crowdloan-claim" }
+
```
tankofzion, line 121: As this rewarding process is meant for crowdloan campaigns, I would suggest to rename the dependency as `crowdloan-reward` and the pallet's folder renamed `pallets/crowdloan-reward`.

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+//! ## Implementations
+//!
+//! - [`Reward`](crowdloan_claim::reward::Reward) : Rewarding functionality for contributors in a relay
+//!   chain crowdloan.
+
+// Ensure we're `no_std` when compiling for Wasm.
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use frame_support::{decl_module, decl_storage, decl_event, decl_error, weights::{Weight}, Parameter, ensure};
+use frame_system::{ensure_root, RawOrigin};
```
tankofzion, line 41: Seems easier to read when packages are in alphabetic order and no need for braces when we import only one package. Moreover, it seems also more readable when  not using one line only.

```rust
use frame_support::{
  decl_module, decl_storage, decl_event, decl_error, 
  ensure,
  Parameter, 
  weights::Weight, 
};
```

pallets/cf-reward/src/lib.rs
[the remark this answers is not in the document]
tankofzion: Concerning the naming of `ModuleId`s, perhaps would be nice to have a convention for our way of naming them.

pallets/cf-reward/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+        // Activate events
+        fn deposit_event() = default;
+
+        /// Initialize the module.
+        /// This dispatchable function must be called from a root account, so that other accounts
+        /// can not block the module via a random initialization. Furthermore it is important, that
+        /// the module has enough funds, as a later funding is not possible.
+        // TODO: Add correct weight function
+        #[weight = 10_000]
+        fn initialize(origin, source: T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
```
tankofzion, line 134: @mustermeiszer I'll add benchmarking features to calculate the weights when integrating with `pallet-crowdloan-claim` pallet.

pallets/cf-reward/src/lib.rs
mustermeiszer: I think ordering them alphabetically is too big of an overhead. Splitting it into multi-lines is fine.

mustermeiszer:
The code has been adjusted in many places. Hence, a new review is necessary.

vedhavyas:
@mustermeiszer can you rebase ?

pallets/vesting-reward/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+        fn deposit_event() = default;
+
+        #[weight = <T as Config>::WeightInfo::set_vesting_start()]
+        fn set_vesting_start(origin, start: T::BlockNumber) -> DispatchResult {
+            T::AdminOrigin::try_origin(origin)
+                .map(|_| ())
+                .or_else(ensure_root)?;
+
+            ensure!(
+                start >= <frame_system::Module<T>>::block_number(),
```
mikiquantum, line 160: This check would avoid retroactively create a vesting schedule right? I believe that might have to support that case and the vesting pallet supports creating schedules in the past.

pallets/vesting-reward/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+                Error::<T>::ElapsedTime
+            );
+            <VestingStart<T>>::put(start);
+
+            Self::deposit_event(RawEvent::UpadteVestingStart(start));
```
mikiquantum, line 165:
```suggestion
            Self::deposit_event(RawEvent::UpdateVestingStart(start));
```

pallets/vesting-reward/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+        /// [vesting_start, vesting_period, conversion_rate, direct_payout_ratio]
+        RewardModuleInitialized(BlockNumber, BlockNumber, Perbill, Perbill),
+        /// Direct payout ratio for contributors has been updated
+        UpdateDirectPayoutRatio(Perbill),
+        /// Conversion rate from relay to native token has been updated
+        UpdateConversionRate(Perbill),
+        /// Vesting period has been updated
+        UpdateVestingPeriod(BlockNumber),
+        /// Start of vesting has been updated
+        UpadteVestingStart(BlockNumber),
```
mikiquantum, line 113:
```suggestion
        UpdateVestingStart(BlockNumber),
```

pallets/vesting-reward/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+use frame_support::dispatch::fmt::Debug;
+use frame_support::traits::{Get, Currency, ExistenceRequirement::{KeepAlive}, EnsureOrigin};
+use frame_support::dispatch::DispatchResult;
+use crowdloan_claim::reward::Reward;
+use sp_runtime::{ModuleId, traits::{AccountIdConversion, StaticLookup}, Perbill};
+pub use weight::WeightInfo;
+use sp_runtime::traits::Zero;
+
+pub trait Config: frame_system::Config + pallet_vesting::Config {
+    /// This module emits events, and hence, depends on the runtime's definition of event
```
tankofzion, line 70: I would suggest that we try, as much as we can, when two specific pallets interact together, that the latter being not tightly coupled, especially when interacting with "native" Substrate pallets (as the code base is moving fast). A loosely-coupled regime seems better, using an associated type instead. It is how, for instance, [ChangeMember](https://substrate.dev/rustdocs/v3.0.0/frame_support/traits/trait.ChangeMembers.html) or [Currency](https://substrate.dev/rustdocs/v3.0.0/frame_support/traits/trait.Currency.html) traits work. Moreover, I would suggest that we put all traits in a specific `traits` folder. What is your feeling?

vedhavyas: not sure I follow what you mean @tankofzion. Do you want to move all the trait definitions to a common module? I think extending the vesting trait here make sense. Am I missing something?

pallets/vesting-reward/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+    trait Store for Module<T: Config> as Reward {
+        /// The conversion rate between relay chain and native chain balances.
+        pub ConversionRate get(fn conversion_rate) build(|config: &GenesisConfig| Perbill::from_percent(config.conversion)): Perbill;
+        /// Which ratio of the rewards are payed directly. The rest is transferred via a vesting schedule.
+        pub DirectPayoutRatio get(fn direct_payout_ratio) build(|config: &GenesisConfig| Perbill::from_percent(config.direct_payout)): Perbill;
+        /// Over which period are the contributions vested.
+        pub VestingPeriod get(fn vesting_period) : T::BlockNumber;
+        /// At which block number does the vesting start .
+        pub VestingStart get(fn vesting_start) : T::BlockNumber;
+    } add_extra_genesis {
```
vedhavyas, line 127: do we need to derive from genisis here since they don't exist? Cant we have a default value with the possibility to change by the root extrinsic

vedhavyas (commented):
LGTM! 
Good Job @mustermeiszer

pallets/vesting-reward/src/lib.rs
tankofzion: > not sure I follow what you mean @tankofzion. Do you want to move all the trait definitions to a common module? I think extending the vesting trait here make sense. Am I missing something?

Hello Ved. Yes, having traits in a module (when traits are rather "common"). In this, you're probably right, it's more an extension, rather than interaction between two pallets. But in the case it's for interacting with two pallets, loosely coupled regime by means of an associated type in the conf seems better.

vedhavyas: @tankofzion Since this is not bidirectional communication, I would suggest we not go the route of placing the traits under common module. In case of traits depending on each other, then they should belong to same modules. I always felt having common trait module a bit of a code smell :-D.
I'll leave it to @mustermeiszer to make that decision :-D

pallets/vesting-reward/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+        type Error = Error<T>;
+
+        const MODULE_ACCOUNT_ID: <T as frame_system::Config>::AccountId = MODULE_ID.into_account();
+
+        // Activate events
+        fn deposit_event() = default;
+
+        #[weight = <T as Config>::WeightInfo::set_vesting_start()]
+        fn set_vesting_start(origin, start: T::BlockNumber) -> DispatchResult {
+            T::AdminOrigin::try_origin(origin)
```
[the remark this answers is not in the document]
mustermeiszer: Noted. Will be done by the end of the week.

pallets/vesting-reward/src/lib.rs
mustermeiszer: Yes. I will check the code in the vesting pallet and then remove this ensure!.

mustermeiszer: Noted. Will be done by the end of the week.

pallets/vesting-reward/src/lib.rs
tankofzion: @vedhavyas I was thinking about the way FRAME support organizes [traits](https://github.com/paritytech/substrate/blob/master/frame/support/src/traits.rs). But for tightly related traits, inside a pallet, it make sense to keep them together. Freddy does not need to change anything for now. It was mostly a thought on how we can organize our code base (e.g. that pallets are not defined in the runtime, like radclaims, but in a separate `pallets` folder, so that to be able to play with different runtimes, if necessary). Just thoughts :)

pallets/vesting-reward/src/lib.rs
mustermeiszer: Noted. Will be done by the end of the week.

pallets/vesting-reward/src/lib.rs
mustermeiszer: Noted. Will be done by the end of the week.

pallets/vesting-reward/src/lib.rs
mustermeiszer: We need to tightly couple here in order to call the `vested_transfer` as it is implemented for `T as pallet_vesting::Config`. But I might be missing something here. 

Regarding the second thing: I think I don't understand the advantage here. If I get it right @tankofzion you are referring to creating an associated type, who has the traitbound of the specific pallet you want to interact with? I don't know if this is possible, to be honest. But could indeed be a benefit.

pallets/cf-reward/src/lib.rs
tankofzion: > I think ordering them alphabetically is too big of an overhead. Splitting it into multi-lines is fine.

We should use a Linter for that.

pallets/vesting-reward/src/lib.rs
tankofzion: > I get it right @tankofzion you are referring to creating an associated type, who has the traitbound of the specific pallet you

Yes. Look at https://substrate.dev/recipes/pallet-coupling.html for instance.

tankofzion: > We need to tightly couple here in order to call the `vested_transfer` as it is implemented for `T as pallet_vesting::Config`. But I might be missing something here.

I would say that you're right. My comment was more for pallets interacting together (dispatch extrinsics or call function).

tankofzion: > Regarding the second thing: I think I don't understand the advantage of having a module of traits. Personally I like to have them split.

It is more in the vein of what Substrate does, at least for common primitives, where all traits are reexported from a single module. Easier for later imports.

mustermeiszer:
I am closing this, due to the new PR #342.

PR #340, opened 2021-04-30
Update to latest rococo-v1

branan:
This performs a full "cargo update", since most of our dependencies are pallets. Code-wise, the changes were largely due to the XCM refactor that removed XcmHandler.

At this point, it should at least be representative of the shape that the new XCM handling code will take, based on what's in cumulus#rococo-v1. Besides actually formatting my changes, I don't expect landing any major adjustments that would cause conflicts, so this is nearly usable as a base for other work.

Still To-do before this is production ready:
* [x] Figure out why I'm seeing zero balances in my genesis block - is genesis broken, or is it a JS apps / API issue? (blocker)
* [x] ORML is not yet up-to-date with latest rococo-v1, so this currently relies on pointing to the PR branch for that update. We need to wait for that PR to be merged. We could also resolve this by ripping out the orml-xtokens stuff for now. See https://github.com/open-web3-stack/open-runtime-module-library/pull/454 (blocker)
* [x] Manually format things that need it so they don't look awful, or declare formatting bankruptcy and push up a commit that runs `cargo fmt` on the whole repo (Not technically a blocker, but I've got standards)
* [x] Undo the extra reformat on files that did get formatted
* [x] Remove unused imports and clean up any other newly introduced warnings (Also not technically a blocker, I guess)
* [ ] Validate that XCM is actually working in this configuration. Requires setting up a second test chain that understands our token. (could/should be a separate PR)
* [x] Figure out XCM security and fees - right now it's configured to allow free messages from any origin. (could/should be a separate PR)
* [x] Make sure that any migrations are accounted for (Could be a separate PR, but will block deployment if there are any)

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -97,10 +97,6 @@ pallet-vesting = { branch = "rococo-v1", git = "https://github.com/paritytech/su
 pallet-sudo = { branch = "rococo-v1", git = "https://github.com/paritytech/substrate.git",  default-features = false }
 frame-benchmarking = { branch = "rococo-v1", git = "https://github.com/paritytech/substrate.git",  default-features = false, optional = true }
 
-# Orml
-orml-xtokens = { branch = "stanly-update-cumulus", git = "https://github.com/stanly-johnson/open-runtime-module-library", default-features = false }
-orml-xcm-support = { branch = "stanly-update-cumulus", git = "https://github.com/stanly-johnson/open-runtime-module-library", default-features = false }
-
 [build-dependencies]
 #wasm-builder-runner = { version = "1.0.5", package = "substrate-wasm-builder-runner", git = "https://github.com/centrifuge/substrate.git", rev = "e271ae7fda72a7ad55ff3c216b8b2abae0ae31fd" }
 substrate-wasm-builder = "3.0.0"
@@ -169,8 +165,6 @@ std = [
     "xcm-executor/std",
     "xcm-builder/std",
     "cumulus-pallet-xcm/std",
-    "orml-xtokens/std",
-    "orml-xcm-support/std",
     #"cumulus-message-broker/std",
 ]
 runtime-benchmarks = ["frame-benchmarking"]
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -32,12 +32,16 @@ parachain-info = { git = "https://github.com/paritytech/cumulus.git", default-fe
 #cumulus-parachain-upgrade = { git = "https://github.com/paritytech/cumulus.git", default-features = false, branch = "rococo-v1" }
 cumulus-pallet-parachain-system = { git = "https://github.com/paritytech/cumulus.git", default-features = false, branch = "rococo-v1" }
 cumulus-primitives-core = {git = 'https://github.com/paritytech/cumulus', default-features = false, branch = "rococo-v1"}
+cumulus-primitives-utility = {git = 'https://github.com/paritytech/cumulus', default-features = false, branch = "rococo-v1"}
 #cumulus-message-broker = {git = 'https://github.com/paritytech/cumulus', default-features = false, branch = 'rococo-v1'}
-cumulus-pallet-xcm-handler = {git = 'https://github.com/paritytech/cumulus', default-features = false, branch = 'rococo-v1'}
+cumulus-pallet-xcm = {git = 'https://github.com/paritytech/cumulus', default-features = false, branch = 'rococo-v1'}
+cumulus-pallet-xcmp-queue = {git = 'https://github.com/paritytech/cumulus', default-features = false, branch = 'rococo-v1'}
+
 
 # polkadot dependencies
 polkadot-parachain = { git = "https://github.com/paritytech/polkadot", default-features = false , branch = "rococo-v1" }
 xcm = { git = "https://github.com/paritytech/polkadot", default-features = false , branch = "rococo-v1" }
+pallet-xcm = { git = "https://github.com/paritytech/polkadot", default-features = false , branch = "rococo-v1" }
 xcm-executor = { git = "https://github.com/paritytech/polkadot", default-features = false , branch = "rococo-v1" }
 xcm-builder = { git = "https://github.com/paritytech/polkadot", default-features = false , branch = "rococo-v1" }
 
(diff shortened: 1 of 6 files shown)


runtime/src/lib.rs
```diff
@@ -365,15 +418,55 @@ impl Convert<Balance, RelayChainBalance> for NativeToRelay {
...
+
+pub struct HandleXcm;
+impl orml_xcm_support::XcmHandler<AccountId, Call> for HandleXcm {
+    fn execute_xcm(origin: AccountId, xcm: xcm::v0::Xcm<Call>) -> sp_runtime::DispatchResult {
+        use xcm::v0::ExecuteXcm;
+		let xcm_origin = X1(Junction::AccountId32 {
+			network: xcm::v0::NetworkId::Any,
+			id: AccountId32Convert::convert(origin),
+		});
+		XcmExecutor::<XcmConfig>::execute_xcm(xcm_origin, xcm, 1000);
```
branan, line 453: Need to catch the Result here and fire an event indicating success / failure

runtime/Cargo.toml
```diff
@@ -94,8 +98,8 @@ pallet-sudo = { branch = "rococo-v1", git = "https://github.com/paritytech/subst
 frame-benchmarking = { branch = "rococo-v1", git = "https://github.com/paritytech/substrate.git",  default-features = false, optional = true }
 
 # Orml
-orml-xtokens = { branch = "master", git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false }
-orml-xcm-support = { branch = "master", git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false }
+orml-xtokens = { branch = "stanly-update-cumulus", git = "https://github.com/stanly-johnson/open-runtime-module-library", default-features = false }
```
mikiquantum, line 101: How did you find this fork? Is it a version of the ORML that has the latest rococo commits?

runtime/src/lib.rs
```diff
@@ -321,31 +360,45 @@ type LocalAssetTransactor = xcm_builder::CurrencyAdapter<
...
 impl Config for XcmConfig {
     type Call = Call;
-    type XcmSender = XcmHandler;
+    type XcmSender = XcmRouter;
     // How to withdraw and deposit an asset.
     type AssetTransactor = LocalAssetTransactor;
     type OriginConverter = LocalOriginConverter;
     type IsReserve = NativeAsset;
-    type IsTeleporter = ();
+    type IsTeleporter = (); // Don't allow any teleports
```
mikiquantum, line 386: Last I checked teleport will require hmp, which is currently worked on. Maybe we can enable it later on we do not have to pass through the relay chain

runtime/src/lib.rs
```diff
@@ -321,31 +360,45 @@ type LocalAssetTransactor = xcm_builder::CurrencyAdapter<
...
 	RelayChainAsNative<RelayChainOrigin, Origin>,
-	SiblingParachainAsNative<cumulus_pallet_xcm_handler::Origin, Origin>,
+	SiblingParachainAsNative<cumulus_pallet_xcm::Origin, Origin>,
 	SignedAccountId32AsNative<CentrifugeNetwork, Origin>,
 );
 
+pub type Barrier = (
+	TakeWeightCredit,
+	AllowTopLevelPaidExecutionFrom<All<MultiLocation>>,
+	AllowUnpaidExecutionFrom<All<MultiLocation>>, // TODO: Fees?
```
mikiquantum, line 370: Is this the place where we would add which calls are allowed to be free and which ones with fees? If so do you have some example to look at and see how that would look like? I think this is still worked on on the parity side.

runtime/src/lib.rs
mikiquantum: Somehow I feel that this should be in its own pallet and make the lib hold minimum logic. WDYT?

mikiquantum: What does the 1000 mean in this context?

mikiquantum (commented):
This is great @branan!

mikiquantum:
>  Figure out why I'm seeing zero balances in my genesis block - is genesis broken, or is it a JS apps / API issue? (blocker)

It could be js types issue, did you use this portal? https://polkadot.js.org/apps if so might have to add some custom types for now, have a look at https://github.com/polkadot-js/api/blob/master/packages/types-known/src/spec/rococo.ts and see if adding them it works. Otherwise to be sure, you can prob use gsrpc to check if there is any balance in the dev accounts.

> ORML is not yet up-to-date with latest rococo-v1, so this currently relies on pointing to the PR branch for that update. We need to wait for that PR to be merged. We could also resolve this by ripping out the orml-xtokens stuff for now. See open-web3-stack/open-runtime-module-library#454

I think we can get this one in without waiting on the ORML xtokens piece yet, we can then start working on a different branch based on this one

> Validate that XCM is actually working in this configuration. Requires setting up a second test chain that understands our token. (could/should be a separate PR)

try setting it up with https://github.com/paritytech/polkadot-launch

runtime/Cargo.toml
branan: It's from this WIP pull request: https://github.com/open-web3-stack/open-runtime-module-library/pull/454

runtime/src/lib.rs
branan: I tried to look into fees but AFAICT it is indeed still being worked on from the parity side. Not a lot there yet

mikiquantum:
> OK, I'll push up a second commit removing xtokens for now, and we can add that back once ORML is updated. Easy enough. I'm already subscribed to that PR, so I'll be able to jump on it when it's ready.

Do not worry too much about removing the xtokens & ORML, as long as it doesnt affect the rest of the code, its ok to leave it there as is and then we can continue from there.

This example of GSRPC and reading data into the AccountInfo should be enough to tell if it is a genesis or display issue:
https://github.com/centrifuge/go-substrate-rpc-client/blob/c38f4745089beae0d8bb8cc4aa4b364c452b81f8/teste2e/author_submit_extrinsic_test.go#L53-L89

branan:
Balances issue appears to be related to `MultiAddress` not round-tripping through the frontend correctly, causing it to query nonexistent accounts.

branan:
More details on that: Our chain uses `type Address = <Indices as StaticLookup>::Source`. The JS apps really like to use `MultiAddress` as the address now. I've been able to at least query balances, but obviously our JS app is going to need some cleanup / updating for Substrate V3

branan:
Alright, I think this is ready to go

runtime/src/lib.rs
branan: Yeah, it should be easy enough to enable as part of actually getting xtokens functional

PR #341, opened 2021-05-05
Adds the NonProxy Type to Proxies

mustermeiszer:
This commit removes the Vesting type in favor of the NonProxy type.
The new proxy type forbids an account of this type to execute
any proxy pallet related operations, besides executing as a different
call as a proxy himself.

## Questions
* [SOLVED] Is it possible to call a forbidden proxy extrinsic as a parameter of the `pallet_proxy::Call::proxy(..)`? 
   UPDATE: [This](https://github.com/paritytech/substrate/blob/e5437efefa82bd8eb567f1245f0a7443ac4e4fe7/frame/proxy/src/lib.rs#L748) should block this.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -231,7 +231,7 @@ pub enum ProxyType {
     NonTransfer,
     Governance,
     Staking,
-    Vesting,
+    NonProxy,
 }
 impl Default for ProxyType { fn default() -> Self { Self::Any } }
 impl InstanceFilter<Call> for ProxyType {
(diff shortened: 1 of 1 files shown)


runtime/src/lib.rs
```diff
@@ -253,15 +253,17 @@ impl InstanceFilter<Call> for ProxyType {
...
-                Call::Staking(..) |
-                Call::Session(..) |
-                Call::Democracy(..) |
-				Call::Council(..) |
-				Call::Elections(..) |
-				Call::Vesting(pallet_vesting::Call::vest(..)) |
-				Call::Vesting(pallet_vesting::Call::vest_other(..))
-            ),
+            ProxyType::NonProxy => !matches!(c,
+                Call::Proxy(pallet_proxy::Call::add_proxy(..)) |
```
mikiquantum, line 257: Can we just simplify this to `Call::Proxy(...)`

mikiquantum: So basically no operations on proxy account.

mustermeiszer: But doesn't the `restricted` account need the capability to, for example, transfer tokens from the anonymous account? This would be done via the `pallet_proxy::Call::proxy()` or am I confusing something?

mikiquantum: ah yes the proxy() call was not included, I oversighted it. 👍

branan:
This is superseded by #451

PR #342, opened 2021-05-06
Feature/crowdloan/reward

mustermeiszer:
This PR replaces PR #338 due the massive changes in the code with the usage of frame/v2. 

Additionally, this PR contains the new configuration of the pallet. Where we omit the usage of the `GenesisConfig` and configure the pallet with defaults in storage and provide setter via extrinsics. 

The formatting of the code is currently wrong. In order to allow a reasonable code review the PR will be updated with the correct indentation when it has passed review.


pallets/crowdloan-reward/src/lib.rs
```diff
@@ -607,13 +580,26 @@ impl<T: Config> Reward for Pallet<T>
...
     <pallet_vesting::Module<T>>::vested_transfer(
       T::Origin::from(RawOrigin::Signed(from.clone())),
       to,
       schedule
-    )?;
+    ).map_err(|err| {
+      T::Currency::transfer(&who, &from, direct_reward, AllowDeath)
+          .err()
+          .unwrap_or_else(|| err)
+    })?;
```
mustermeiszer, line 601: I am currently not sure if there is a better way to ensure both tx succeed or both fail.

pallets/crowdloan-reward/src/lib.rs
```diff
@@ -322,38 +326,26 @@ pub mod pallet {
   ///
   /// It allows to build genesis storage.
   #[pallet::genesis_config]
-	pub struct GenesisConfig {
-    #[doc = "Conversion rate between relay chain and parachain tokens."]
-    pub conversion: u32,
-    #[doc = "Direct reward payout ratio."]
-    pub direct_payout: u32
+  pub struct GenesisConfig {
```
vedhavyas, line 329: Wee can remove these instead of keeping them empty. WDYT ?

pallets/crowdloan-reward/src/lib.rs
```diff
@@ -607,13 +580,26 @@ impl<T: Config> Reward for Pallet<T>
...
+    // the direct payout back to the module.
+    //
+    // NOTE: This procedure does change the state...
     <pallet_vesting::Module<T>>::vested_transfer(
       T::Origin::from(RawOrigin::Signed(from.clone())),
       to,
       schedule
-    )?;
+    ).map_err(|err| {
+      T::Currency::transfer(&who, &from, direct_reward, AllowDeath)
```
vedhavyas, line 598: `KepAlive` instead

pallets/crowdloan-reward/src/lib.rs
mustermeiszer: I thought so too. But the compiler is complaining about a missing `GenesisConfig`. I am currently unsure why. Maybe it is because of the tight coupling with the vesting pallet.

pallets/crowdloan-reward/src/lib.rs
mustermeiszer: I think we need to allow death here. The account of the contributor is possibly a fresh one and if we do not allow death, we always loose a small amount upon errors of the vested transfer call.

pallets/crowdloan-reward/src/lib.rs
vedhavyas: ahh cool. if not possible then ls leave it then

pallets/crowdloan-reward/src/lib.rs
vedhavyas: true but this transfer doesn't involve any fees so we pretty negated the action. So if there are any dust, it pretty belongs to the `who`, no ?

mustermeiszer: Yes, of course. But in the case of a newly created account, the rollback of the previous transfer will fail of we use `KeepAlive` due to this [code line](https://github.com/paritytech/substrate/blob/75957575080a6eacfbdf8995b63f2693ee33efe6/frame/balances/src/lib.rs#L1351), leaving the direct payout at `who`.  As the claim module will register this as a failed claim, it will allow the contributor to claim again. If he does so with another newly created account he could get the direct payout a second time, and so forth.

Basically what I mean here is: If the account is above the existential deposit solely due to our previous transfer, we must be able to purge him, otherwise, we don't get our funds back.

Another thing that could be problematic is this [code line](https://github.com/paritytech/substrate/blob/75957575080a6eacfbdf8995b63f2693ee33efe6/frame/balances/src/lib.rs#L1350):
`!system::Pallet::<T>::is_provider_required(transactor)`, which checks if the account has any other module being dependent on it.

Do you know if there is the possibility for an account to have such a `consumers` reference while having less than the existential deposit as a balance? Then again, the scenario from above would apply.

mustermeiszer: I updated the code. It now checks in advance, if a vesting schedule exists. Currently, this and the amount to low error, are the only reason a vested transfer might fail. As we already check upfront if the module has enough funds for this reward payout. 

Generally, I would say that there is no reason that one of the transfers should fail now. We could therefore think about removing the action upon failure of the `vested_transfer`. But maybe better be safe than sorry? Of course, we then still have to decide, the above case.

pallets/crowdloan-reward/src/tests.rs
```diff
@@ -97,49 +109,49 @@ fn setters_ok() {
...
-    .execute_with(|| {
-      assert_noop!(
-        CrowdloanReward::set_vesting_start(Origin::root(), 2),
-        CrowdloanRewardError::<MockRuntime>::ElapsedTime
-      );
-    });
-}
-
 #[test]
 fn reward_participant() {
```
mikiquantum, line 132: Can we create a test that verifies 0% upfront reward?

pallets/crowdloan-reward/src/lib.rs
```diff
@@ -111,56 +98,61 @@ use frame_support::{
...
+// Re-export in crate namespace (for runtime construction)
+pub use pallet::*;
 // Claim reward trait to be implemented
 use pallet_crowdloan_claim::traits::Reward;
 
 // Extrinsics weight information
 pub use crate::traits::WeightInfo;
 
+// Mock runtime and unit test cases
+mod mock;
```
mikiquantum, line 140: Do you know if the tests and mocks are skipped when building the runtime blob? Should we create its own file outside of lib?

pallets/crowdloan-reward/src/lib.rs
mikiquantum: Since the `keepAlive|AllowDeath` are only applicable to the from, then is up to us to always have enough funds to cover the existential deposit, and allowing death for a module_id account is prob not a good idea.
Maybe we can add the `existential_deposit` to the total amount available check above, so:
```
ensure!( free_balance > reward + T::ExistentialDeposit::get(), Error::<T>::NotEnoughFunds );
```
Then we cover the case of keeping alive, right? 

For the transactor is_provider check then we do not need to care about it since the account will stay alive.

pallets/crowdloan-reward/src/tests.rs
mustermeiszer: Yes. Will be done.

pallets/crowdloan-reward/src/lib.rs
mustermeiszer: The #[cfg(test)] is missing. Then it will be skipped. Will add this to mock and test module.

pallets/crowdloan-reward/src/lib.rs
mustermeiszer: > Since the `keepAlive|AllowDeath` are only applicable to the from, then is up to us to always have enough funds to cover the existential deposit, and allowing death for a module_id account is prob not a good idea.
Why is it only applicable to the `from`? In case, we want to revert the previous transfer, it is applied to the `who` (recipient account).

> Maybe we can add the `existential_deposit` to the total amount available check above, so:
This check is done at line 560. The module also transfers with `KeepAlive` to the recipient. So, we should be good on this side.

> For the transactor is_provider check then we do not need to care about it since the account will stay alive.
But this will also mean, that the transfer fails and we can not revert the previous transfer. Although, I think we should be safe, as we are checking everything upfront and now.

pallets/crowdloan-reward/src/benchmarking.rs
```diff
@@ -15,7 +15,7 @@
 // along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
 
 
-// # Crowdloan reward pallet's transactions benchmarking
+// # Crowdlan reward pallet's transactions benchmarking
```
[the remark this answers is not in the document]
mustermeiszer: Done.

mustermeiszer:
@mikiquantum and @vedhavyas, do you still have objections? Otherwise I would merge this into the feature branch, as I will continue with the claim pallet from marc and need this branch.

vedhavyas (approved):

pallets/crowdloan-reward/src/tests.rs
mikiquantum: let's add this later in ur next pr

mikiquantum (approved):
LGTM, just some minor comments about the event names

pallets/crowdloan-reward/src/lib.rs
```diff
@@ -190,304 +184,296 @@ type BalanceOf<T> = <<T as pallet_vesting::Config>::Currency as Currency<<T as f
...
-        Error::<T>::ElapsedTime
-      );
+        /// Event triggered when the reward module is ready to reward contributors
+        /// \[vesting_start, vesting_period, conversion_rate, direct_payout_ratio\]
+        RewardModuleInitialized(T::BlockNumber, T::BlockNumber, BalanceOf<T>, Perbill),
 
-      <VestingStart<T>>::put(start);
+        /// Direct payout ratio for contributors has been updated
+        /// \[payout_ratio\]
+        UpdateDirectPayoutRatio(Perbill),
```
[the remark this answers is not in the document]
mustermeiszer: Done.

pallets/crowdloan-reward/src/lib.rs
```diff
@@ -539,84 +522,94 @@ impl<T: Config> Pallet<T> {
...
-    let ratio = Perbill::from_percent(direct_payout_ratio);
-    <DirectPayoutRatio<T>>::put(ratio);
+        // Ensure transfer will go through and we want to keep the module account alive.
+        let free_balance = <T as pallet_vesting::Config>::Currency::free_balance(&from)
+            .checked_sub(&<T as pallet_vesting::Config>::Currency::minimum_balance()).unwrap_or(Zero::zero());
+        ensure!( free_balance > reward, Error::<T>::NotEnoughFunds );
 
-    let rate = Perbill::from_percent(conversion_rate);
-    <ConversionRate<T>>::set(rate);
+        let direct_reward = Self::direct_payout_ratio() * reward;
```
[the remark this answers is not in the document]
mustermeiszer: We don't need `checked_mul` here as the `direct_payout_ratio()` is of type `Perbill` which is always <= 1.

mustermeiszer:
New changes solely due to rebasing and rust-fmt. Will merge this one in order to proceed with implementation.

PR #351, opened 2021-05-12
Update Charcoal genesis spec

branan:
Two commits in this one:
* Bump the runtime to include #346, which is needed to make the JS apps play nice
* Update genesis accounts to match Amber.

The first is probably impossible to review. The second should be a smaller diff (at least for the non-raw spec).

We may want to make some slight changes to the genesis distributions (removing folks that have left?) but if we don't get around to that this should still provide a good baseline. @mikiquantum gets root since he'll likely be dealing with deployment.


mikiquantum (approved):
LGTM!

branan:
Merging this now since I'll want to do another runtime upgrade for the switch to the polkadot-v0.9.1 branches. Any further tweaks to genesis should be straightforward.

PR #364, opened 2021-05-17
add github actions

vedhavyas:
Note: tests are not run in this one. Will be another once we migrate our pallets


.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
+on:
+  push:
+    branches: [master]
```
mikiquantum, line 3: Since this github actions commit will only stay in the parachain branch, I think it makes sense to add the parachain branch as well as trigger.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,40 @@
...
+      - run: echo "BRANCH=$(echo $GITHUB_REF | cut -d'/' -f 3)" >> $GITHUB_ENV
+      - name: Tag image
+        if: env.BRANCH == 'master'
+        run: docker tag centrifugeio/centrifuge-chain:latest "centrifugeio/centrifuge-chain:$(date -u +%Y%m%d%H%M%S)-$(git rev-parse --short HEAD)"
+      - name: Tag image
+        if: env.BRANCH == 'parachain'
+        run: docker tag centrifugeio/centrifuge-chain:latest "centrifugeio/centrifuge-chain:parachain-$(date -u +%Y%m%d%H%M%S)-$(git rev-parse --short HEAD)"
+      - name: List images
+        run: docker images
+      - name: Push image to Docker Hub
```
mikiquantum, line 39: Does it mean that we will push `latest` even if it is a parachain build? If so, might want to avoid doing that, to minimize community confusion.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  tests:
+    name: ${{ matrix.target }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        target: [runtime-test]
+    runs-on: ${{ matrix.os }}
+    env:
+      RUST_TOOLCHAIN: "nightly-2021-03-15"
```
mikiquantum, line 15: And use `nightly` here?

.github/workflows/build.yml
```diff
@@ -0,0 +1,29 @@
+on:
+  push:
+    branches: [master]
```
mikiquantum, line 3: Since this code will only stay in the parachain branch, should we add a trigger for it?

mikiquantum (approved):
Minor comments!

.github/workflows/tests.yml
vedhavyas: we cant use nightly since srtool latest image is above

vedhavyas:
@mikiquantum applying your review changes #365

PR #370, opened 2021-05-18
Allow specifying parachain ID for genesis from the CLI

branan:
This is necesary for compatibility with `polkadot-launch`

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -63,6 +63,10 @@ pub struct ExportGenesisStateCommand {
 	#[structopt(short, long)]
 	pub raw: bool,
 
+	/// Id of the parachain this state is for.
+	#[structopt(long)]
+	pub parachain_id: Option<u32>,
+
 	/// The name of the chain for that the genesis state should be exported.
 	#[structopt(long)]
 	pub chain: Option<String>,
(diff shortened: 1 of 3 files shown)


vedhavyas:
@branan that change already went in in the Previous PR. Did I miss something ?

src/cli.rs
```diff
@@ -63,6 +63,10 @@ pub struct ExportGenesisStateCommand {
 	#[structopt(short, long)]
 	pub raw: bool,
 
+	/// Id of the parachain this state is for.
+	#[structopt(long)]
+	pub parachain_id: Option<u32>,
```
vedhavyas, line 68: I have removed this since we dont need it anymore for Genesis State

branan: Unfortunately polkadot-launch still expects it to be there - https://github.com/paritytech/polkadot-launch/blob/master/src/spawn.ts#L125

And generally I think it makes sense for our CLI to follow the conventions of other cumulus projects

vedhavyas: That was older projects. I think they are trying to be backward compatible. This would be dead code on our end since we upto latest

vedhavyas:
@branan could you remove the cmd and get this in? anything else we need to do here ?

vedhavyas (approved):

PR #381, opened 2021-06-02
Adding Rumba and Charcoal Specs

mikiquantum:
This PR addresses:

- Adds chain specs for Rumba
- Updates both Rumba and Charcoal genesis files
- Removes relay chain reference from name, now that we have defined the supported environments
- Adds convenient script to generate specs and parachain files


branan (approved):
LGTM

PR #385, opened 2021-06-07
Crowdloan Claim Pallet V1 

mustermeiszer:
The test `test_valid_claim` is not working, as the proofs of contribution do not work. In order to examine the problem. I create this intermediate PR. 

Current state: 
* Proofs are generated from a local setup
* The generated nodes are equally for all three contributors of the local crowdloan (which is weird in my opinion). 
* The prefixed storage key and the keys themself inside this child storage are correct
     -> I can extract the contributions from the storage via these keys. (Allthough, the decoding function of scale, does not decode the stored `memo` correctly (See `test_valid_claim` line 234-241), but when done manually, the hex-string does represent the correct message. NOTE: The hex string here is retrieved from the local-chain-storage via the UI)
* The root hash is found in the provided proofs.
   -> Easiest is to debug through the code and set breakpoints in the `triedb.rs` line 70 (check of root existence) and go through the `triedb/lookup.rs` function `look_up` line 41-ff.

Appreciate any thoughts on this

### DIFF
--- a/pallets/crowdloan-claim/src/lib.rs
+++ b/pallets/crowdloan-claim/src/lib.rs
@@ -119,6 +119,7 @@ pub use pallet::*;
 
 // Extrinsics weight information
 pub use crate::traits::WeightInfo as PalletWeightInfo;
+use frame_support::dispatch::DispatchError;
 
 // Mock runtime and unit test cases
 #[cfg(test)]
@@ -188,6 +189,7 @@ pub mod pallet {
     use frame_system::pallet_prelude::*;
 
     use super::*;
+    use frame_support::sp_runtime::Perbill;
     use pallet_crowdloan_claim_reward::Reward;
 
     // Crowdloan claim pallet type declaration.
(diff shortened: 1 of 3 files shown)


mustermeiszer:
There a two things I would like to discuss: 

* The current reward interface couples the initialization of the claim-pallet, with the initialization of the reward-pallet. Do we want to keep this? Personally, I would decouple as this is a very specific coupling (namely providing conversion rates, vesting periods, etc.)

* The claim-pallet can only be initialized once. I would prefer to change this to initializing the pallet with a start-block and a lease-period (all Blocknumbers). The pallet can not be reinitialized until the lease-period is over.
NOTE: The `ProcessedClaims` storage, i.e. the relay-chain accounts that have already claimed can hereby either: 
   * be deleted upon new-init -> Loose state of previous crowdloan, but less storage usage
   * be extended to `(RelayChainAccount, Index)`, where the index is a counter that is incremented for every initialization -> State of previous crowdloans is kept, but more storage is used

PR #391, opened 2021-06-09
Crowdloan Claim Pallet V2 -Alternative to  #385

mustermeiszer:
This PR is an alternative to #385. 
The difference here are that the pallets need to be initialized separately. Hence, the `Reward` trait has a smaller interface and only provides a rewarding functionality. 
Furthermore, the claim pallet is changed, in order to provide the possibility to initialize it multiple times.

### DIFF
--- a/pallets/crowdloan-claim/src/lib.rs
+++ b/pallets/crowdloan-claim/src/lib.rs
@@ -97,7 +97,7 @@ use codec::{Decode, Encode};
 use frame_support::{
     dispatch::{fmt::Debug, Codec, DispatchResult},
     ensure,
-    sp_runtime::traits::{AtLeast32BitUnsigned, MaybeSerialize},
+    sp_runtime::traits::{AtLeast32BitUnsigned, MaybeSerialize, Saturating},
     traits::{EnsureOrigin, Get},
     weights::Weight,
 };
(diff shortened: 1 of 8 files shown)


pallets/crowdloan-claim/src/lib.rs
```diff
@@ -320,14 +344,48 @@ pub mod pallet {
...
+        StorageMap<_, Blake2_128Concat, (T::RelayChainAccountId, Index), bool>;
+
+    #[pallet::type_value]
+    pub fn OnIndexEmpty() -> Index {
+        0
+    }
+
+    #[pallet::storage]
+    #[pallet::getter(fn curr_index)]
+    pub type CurrIndex<T: Config> = StorageValue<_, Index, ValueQuery, OnIndexEmpty>;
```
vedhavyas, line 356: is it not possible to give the default here like pallet V1 ?

pallets/crowdloan-claim/src/mock.rs
```diff
@@ -63,6 +63,26 @@ impl WeightInfo for MockWeightInfo {
     fn claim_reward() -> Weight {
         0 as Weight
     }
+
+    fn set_lease_start() -> u64 {
+        0 as Weight
```
vedhavyas, line 68: so benchmarking is done in a different PR ?

vedhavyas (approved):
Some questions but LGTM otherwise.

Thanks!

pallets/crowdloan-claim/src/mock.rs
mustermeiszer: Yes. Once this is through I will add a PR for this pallet in parachain, where I integrate the benchmark. Wanted to close this first, finally...

pallets/crowdloan-claim/src/lib.rs
mustermeiszer: To be honest, I don't know. Thought, that this replaces the default, as you can also put arbitrary logic in the function like with the build closure.

vedhavyas: hmm. okay. That one seems much simple.

PR #392, opened 2021-06-09
Altair

vedhavyas:
This PR updates our runtime to Altair
The following pallets are added:
- System pallets
- Monetary pallets
	- Balances
	- Transaction payment
- Authoring
	- Aura
- Substrate pallets
	- Multisig, proxy .....
- our pallets
	- Fees
	- Anchors
	- Claims
- Admin
	- Sudo

Points to note:
- Only required extrisnics are allowed
- Staking and related pallets are not added yet. Might not go in the first runtime too
- All the pallets are numbered so that metadata could be maintained for signers without breaking them during upgrades
- Benchmarking is done for Fees. Rest needs to come along with the pallet integrations
- At the moment, `common.rs` is still part of runtime but with multi runtime, we can move them a common crate.


runtime/src/lib.rs
```diff
@@ -116,17 +71,13 @@ impl_opaque_keys! {
...
-    spec_name: create_runtime_str!("centrifuge-chain"),
-    impl_name: create_runtime_str!("centrifuge-chain"),
-    authoring_version: 10,
-    // Per convention: if the runtime behavior changes, increment spec_version
-    // and set impl_version to 0. If only runtime
-    // implementation changes and behavior does not, then leave spec_version as
-    // is and increment impl_version.
-    spec_version: 1000,
-    impl_version: 0,
+    spec_name: create_runtime_str!("altair"),
```
mikiquantum, line 74: We will need to add explicit types for this runtime in polkadot-js/api/apps regardless. Should we start with 1000? https://github.com/polkadot-js/api/blob/f97a6d041827e64c96e6274bc1e2c3ea275fcd5f/packages/types-known/src/spec/centrifuge-chain.ts#L72-L90

We should do something like this on the polka-js side: https://github.com/AcalaNetwork/acala.js/tree/master/packages/type-definitions/src/spec

runtime/src/lib.rs
```diff
@@ -468,8 +346,25 @@ impl InstanceFilter<Call> for ProxyType {
 				Call::Balances(..)
 			),
             ProxyType::Governance => matches!(c,
-				Call::Democracy(..) | Call::Council(..) | Call::Elections(..)
-			)
+				// Call::Democracy(..) |
```
mikiquantum, line 349: Should we add this already? https://github.com/centrifuge/centrifuge-chain/pull/341/files

mikiquantum (commented):
Look great! Is your plan to test this out first and then strip down the following pallets before we cut altair?:
- Balances
- Democracy
- Council/Elections
- Fees/Anchors/Claims

mustermeiszer (approved):
LGTM. Thanks!

mustermeiszer:
> Look great! Is your plan to test this out first and then strip down the following pallets before we cut altair?:
> 
> * Balances
> * Democracy
> * Council/Elections
> * Fees/Anchors/Claims

What do you mean by stripping down the pallets?

runtime/src/lib.rs
vedhavyas: can come in a different PR else this will get big

runtime/src/lib.rs
vedhavyas: for altair, we would need a different one but regarding the spec_version, I have no strong opinions about any number.

vedhavyas:
> Look great! Is your plan to test this out first and then strip down the following pallets before we cut altair?:
>
> Balances
> Democracy
> Council/Elections
> Fees/Anchors/Claims

We wont strip any. We might add more in the first runtime as per our discussion last week

tankofzion (commented):
Great job Ved. Thanks a lot Ioda Master :)

tankofzion (approved):

mustermeiszer (approved):
LGTM!

PR #393, opened 2021-06-11
Update to polkadot-v0.9.4-rc?

branan:
This is based on the WIP polkadot v0.9.4-rc2 branch, which hasn't actually
been tagged as rc2 yet. It's somewhere in between rc1 and what will be
rc2, but shouldn't have anything extraneous (as an update to master
might).


mikiquantum:
@branan what is left here, code-base wise? Can you generate as well the new specs for charcoal?
Essentially running this: https://github.com/centrifuge/centrifuge-chain/blob/parachain/scripts/export_parachain_files.sh with args 
```
charcoal-staging 2010 true
```
and then renaming to `charcoal-spec.json` and `charcoal-spec-raw.json` 

I am ok getting this in, if local works, and push 0.9.5 later.

branan:
@mikiquantum nothing left but to wait for Parity to actually tag 0.9.4 (or 0.9.5, or whatever they tag next). I'm happy merging this as-is (once we get the spec updated), since it gets us up to the closest thing to a tag that they seem to have published?

PR #395, opened 2021-06-15
Multi-runtime support

branan:


branan:
The networks that are mapped as "altair" vs. "charcoal" right now are definitely just plain wrong.

I think we want rumba/charcoal/altair to all be the "altair" runtime, with a new codename for the old runtime we're still developing/testing against. I'm open to suggestions for what to name that.

branan:
Maybe "delorean" since it's the runtime from the future? :rofl:

branan:
Since the `altair` runtime ended up containing most pallets, the only thing missing from the development runtime is the XCM stuff. I don't think that needs to block getting this in.

branan:
Will still need to regenerate a bunch of chain spec JSON files to go with this

tankofzion (approved):
LGTM

Cargo.toml
```diff
@@ -30,12 +30,14 @@ members = [
...
 	"pallets/fees",
 	"pallets/anchors",
-    "pallets/claims",
+	"pallets/claims",
 	"pallets/crowdloan-claim",
 	"pallets/crowdloan-reward",
 	"libs/trait-crowdloan-reward",
 	"libs/proofs",
-	"runtime",
+        "runtime/altair",
```
vedhavyas, line 38: indentation ?

runtime/centrifuge/src/impls.rs
```diff
@@ -0,0 +1,59 @@
+//! Some configurable implementations as associated type for the substrate runtime.
```
vedhavyas, line 1: we should move this to common pallet as well.

mikiquantum, line 1: Do you think is possible to move this to common?

src/command.rs
```diff
@@ -34,23 +34,67 @@ use sp_core::hexdisplay::HexDisplay;
...
 use std::{io::Write, net::SocketAddr};
 
+enum ChainIdentity {
+	Altair,
+	Centrifuge,
+	Development,
+}
+
+trait IdentifyChain {
+	fn identity(&self) -> ChainIdentity;
```
vedhavyas, line 44:
```suggestion
	fn identify(&self) -> ChainIdentity;
```

runtime/development/src/constants.rs
```diff
@@ -0,0 +1,15 @@
+//! A set of constant values used in altair runtime
```
mikiquantum, line 1: Since the value conversion should be the same among runtimes, do you think we can just create an alias  from CFG?

mikiquantum (approved):
LGTM!

vedhavyas (approved):
Some suggestions but looks great!

mustermeiszer:
> Maybe "delorean" since it's the runtime from the future? 🤣

Actually, I like it! Haha

mustermeiszer (approved):
Nice! LGTM! Thanks Branan!

vedhavyas (approved):

PR #396, opened 2021-06-18
add session pallet to inject collator session keys

vedhavyas:
This PR adds session pallets to inject session keys for collators during genesis and also makes it easier to migrate to staking later


mustermeiszer (approved):
LGTM!

PR #397, opened 2021-06-21
0.9.5

vedhavyas:
- 0.9.5
- Update benchmark script
- Re-run benchmark for fees.
- remove `build-runtime` from build if its a PR since build-node should build the runtime as well. Instead we run the build-runtime for merged branch. This is would reduce the build time for pr by good amount


scripts/run_benchmark.sh
```diff
@@ -22,3 +22,5 @@ cargo run --release --features runtime-benchmarks -- benchmark \
   --heap-pages=4096 \
   --output="${output}" \
   --template=./scripts/frame-weight-template.hbs
+
+cargo fmt
```
branan, line 26: Why are we adding `cargo fmt` to the benchmark script?

vedhavyas: it generates the weight.rs file that may or may not cargo fmted. 
Given its a generated file, we can completely avoid lint on that file and many more to come. But I'd rather have us cargo fmt them so that we dont polllute

branan: Cool, sounds good

branan (approved):
This LGTM once CI goes green. Changes for 0.9.5 look minor as expected, and the other fixes included seem straightforward enough.

tankofzion (approved):
LGTM

PR #399, opened 2021-06-21
[WIP] Migration Pallet - Controls migrations

mustermeiszer:
This is a WIP, but as I wanted to discuss on the actual handling of our migrations as early as possible, I already put it here. 

My current approach is the following:

- create migration data as: patriciakey-value pairs. Both are scale encoded and can inserted directly into the database via the 
`sp_io::storage::set()` or via the `frame_support::unhashed::put_raw()` API. 

NOTE: Here we can decide now: 
- adjust frame_system, and other substrate based pallets and their the `on_runtime_upgrade` logic
- use the proposed `migration` pallet, which simply handles the inserts into the raw storage and possibly other calls to other pallets. (If this works for all pallets, e.g. vesting, I must research further).


vedhavyas:
@mustermeiszer For the runtime upgrade, We should use this 
https://github.com/centrifuge/centrifuge-chain/blob/parachain/runtime/src/lib.rs#L712

that way each runtime can have its own migrations, avoids us to have another pallet migration that may be common to multiple runtimes at a time.

vedhavyas:
> If this works for all pallets, e.g. vesting, I must research further

for vesting example, I think its better use the pallet_vesting pallet for migration since it would be better to let it handle the locks rather we doing a raw key,value store.

mustermeiszer:
> @mustermeiszer For the runtime upgrade, We should use this
> https://github.com/centrifuge/centrifuge-chain/blob/parachain/runtime/src/lib.rs#L712
> 
> that way each runtime can have its own migrations, avoids us to have another pallet migration that may be common to multiple runtimes at a time.

You mean a struct implementing the `OnRuntimeUpgrade` trait? I thought about this too. Might be the better option. I wanted to use the pallet, as I thought about hacking a multi-block-upgrade into our system, by "miss-using" the `on_initialize` of the pallet like the following: 
- If runtime_upgrade started -> then always return the `max_block` weight in the `on_initialize` of the migration pallet until all storage is migrated. 


But yeah, not sure if it is worth it. In this way, we could prevent waiting too long for all the upgrades to happen. And not sure, if the relay-chain can handle a "upgrade-all-at-once" wasm blob.

vedhavyas:
> as I thought about hacking a multi-block-upgrade into our system, by "miss-using" the on_initialize of the

I think this is worth a try. maybe we can use the pallet for what u meant but get the data from the OnRuntimeUpgrade through lib.rs. This way, pallet can be generic over data. WDYT ?

mustermeiszer:
> > as I thought about hacking a multi-block-upgrade into our system, by "miss-using" the on_initialize of the
> 
> I think this is worth a try. maybe we can use the pallet for what u meant but get the data from the OnRuntimeUpgrade through lib.rs. This way, pallet can be generic over data. WDYT ?

Yes, this could work. Then we only have a boolean-indicator in the Migration pallet, for each pallet we want to migrate and use the `on_runtime_upgrade` for the logic. This could also solve my concerns of diverging permanently from the substrate base with out costum pallets.

PR #404, opened 2021-06-23
Added licencse templates

mustermeiszer:
Here are the templates for the file headers.


apache-template.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2021 Centrifuge Foundation (centrifuge.io).
```
vedhavyas, line 1: do we need them to be a rust file? can we do away with simple text file?

vedhavyas (commented):
I think the we need to add apache license as well. It is missing at the moment.

apache-template.rs
mustermeiszer: No need to be rust. Can be txt. Will change this.

vedhavyas (approved):
LGTM

PR #449, opened 2021-07-16
Runtime 5

branan:
This runtime bump includes two changes:
* An increase to the number of nominators that will be rewarded for a given validator, from 64 to 128
* An increase to the proposal lifetime for chainbridge transactions, from 100 blocks (10 minutes) to 1000 (100 minutes)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ branches:
   only:
   - master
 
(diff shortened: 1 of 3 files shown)


mikiquantum (approved):
LGTM, thanks!

PR #451, opened 2021-07-27
Replace Vesting proxy with NonProxy proxy

branan:
This removes the `Vesting` proxy type, and replaces it with a `NonProxy` proxy type. This new proxy type forbids all proxy operations on the proxied account **except** for calling `proxy::proxy(..)` to invoke an account which is proxied by the proxied account.


branan:
My changes from what @mustermeiszer had in his version are pretty straightforward - let me know if there are any questions.

vedhavyas (approved):

PR #452, opened 2021-07-27
Use `ensure_can_withdraw` in chain bridge transfers

branan:
Previously, the chain bridge transfer checked that the total transfer
amount + fee was greater than the "free" balance in the account. This
does not take into account various types of locks that can be on an
account.

This PR changes to use the `ensure_can_withdraw` function from the
Currency trait. This function checks all locks and fails the
transaction if there is not enough available balance in the account.

### DIFF
--- a/runtime/src/bridge.rs
+++ b/runtime/src/bridge.rs
@@ -8,7 +8,7 @@ use crate::{fees, constants::currency};
 use frame_support::traits::{Currency, ExistenceRequirement::AllowDeath, Get, WithdrawReasons};
 use frame_support::{
     decl_error, decl_event, decl_module, decl_storage, dispatch::DispatchResult, ensure,
-    traits::EnsureOrigin,
+    traits::EnsureOrigin, transactional
 };
 use frame_system::{self as system, ensure_signed, ensure_root};
 use sp_core::U256;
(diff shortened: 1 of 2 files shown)


branan:
Regarding making this transactional: It looks like while there is a `transactional` macro available, it transforms the function in a way that breaks that `?` operator - I don't think I have enough time this evening to track this down fully, but I'll try. Worst case, this PR does at least fix the known breakage as-is.

vedhavyas:
> It looks like while there is a transactional macro available, it transforms the function in a way that breaks that ? operator 

How much work do you anticipate to get this working ?

vedhavyas (approved):

runtime/src/bridge.rs
```diff
@@ -257,13 +213,75 @@ impl<T: Trait> Module<T> {
 			.or_else(ensure_root)?;
 		Ok(())
 	}
+
+        fn do_transfer_native(origin: T::Origin, amount: BalanceOf<T>, recipient: Vec<u8>, dest_id: chainbridge::ChainId) -> DispatchResult {
+            let source = ensure_signed(origin)?;
+
+            let token_fee: T::Balance = TokenTransferFee::<T>::get();
+            let currency_token_fee = TryInto::<u128>::try_into(token_fee).map_err(|_| "Token fee not convertible to currency")?
+                                                     .try_into().map_err(|_| "Token fee not convertible to currency")?;
```
mikiquantum, line 222: Is this `try_into` duplicate?

mikiquantum (approved):
Minor comment only, very cool that you could get the `transactional` macro in!

runtime/src/bridge.rs
branan: Alas not redundant. We first convert the T::Balance into a u128, and then from that into a Currency::Balance. I could not figure out how to get the compiler to believe that those two balances were the same type, so we have this unfortunate dance.

PR #454, opened 2021-08-18
New Proofs and Proof-Verification for Crowdloan-Claims

mustermeiszer:
This PR includes a new Proof, our costume base-2 Merkel tree, for claims of crowdloan contributors. Furthermore, of course, tests and verification of proofs are adjusted.


pallets/crowdloan-claim/src/lib.rs
```diff
@@ -751,6 +756,22 @@ impl<T: Config> Pallet<T> {
 		T::PalletId::get().into_account()
 	}
 
+	/// Build a sorted hash of two given hash values.
+	///
+	/// Hash a:b if a < b, else b:a. Uses the runtime module's hasher.
+	pub fn sorted_hash_of(a: &T::Hash, b: &T::Hash) -> T::Hash {
```
vedhavyas, line 762: we have the proofs library for proof verification. Can you use that ?

mustermeiszer: Isn't that part of the pallet claims?

mustermeiszer: Nevermind I found it.

pallets/crowdloan-claim/src/lib.rs
```diff
@@ -790,42 +811,38 @@ impl<T: Config> Pallet<T> {
...
+
+		// Generate root hash
+		let leaf_hash = T::Hashing::hash(&v);
+		let mut root_hash = proof
+			.iter()
+			.fold(leaf_hash, |acc, hash| Self::sorted_hash_of(&acc, hash));
+
+		// Initial runs might only have trees of single leaves,
+		// in this case leaf_hash is as well root_hash
+		if proof.len() == 0 {
```
mikiquantum, line 840: I don't think that would be the case for us, since we will build the whole tree at once.

pallets/crowdloan-claim/src/lib.rs
mustermeiszer: @vedhavyas and @mikiquantum I would prefer to have this duplicate code. 
I would a Hashing algorithm to implement `Hasher` trait, as we are not using the substrate I can not use the system hashing here. Would be kind overkill in my opinion just for using this one function. 

WDYT?

vedhavyas: This is thee change I was suggesting https://github.com/centrifuge/centrifuge-chain/blob/parachain/libs/proofs/src/mock.rs#L15

Not sure I completely follow you

mustermeiszer:
Closing thi

PR #455, opened 2021-08-18
Integrate migration pallet into Development Runtime

mustermeiszer:


runtime/development/src/lib.rs
```diff
@@ -597,6 +590,22 @@ impl pallet_claims::Config for Runtime {
...
+}
+
+// Implement the migration manager pallet
+// The actual associated type, which executes the migration can be found in the migration folder
+impl pallet_migration_manager::Config for Runtime {
+	type MigrationMaxAccounts = MigrationMaxAccounts;
+	type MigrationMaxVestings = MigrationMaxVestings;
+	type MigrationMaxProxies = MigrationMaxProxies;
+	type Event = Event;
+	type WeightInfo = ();
```
vedhavyas, line 606: can you point the standard weight implmentation instead of default ?

vedhavyas (approved):
LGTM. Use standard weight implelemtation

mikiquantum (approved):
👍

runtime/development/src/lib.rs
mustermeiszer: Done. Also added this one for altair.

PR #456, opened 2021-08-18
Integrates crowdloan pallets (claim and reward)

mustermeiszer:
WARN: This is based upon the changes that happen in the crowdloan-claim-pallet. We can do both together here in my opinion.


pallets/crowdloan-claim/src/lib.rs
```diff
@@ -751,6 +756,22 @@ impl<T: Config> Pallet<T> {
 		T::PalletId::get().into_account()
 	}
 
+	/// Build a sorted hash of two given hash values.
+	///
+	/// Hash a:b if a < b, else b:a. Uses the runtime module's hasher.
+	pub fn sorted_hash_of(a: &T::Hash, b: &T::Hash) -> T::Hash {
```
vedhavyas, line 762: We should use proofs for verifying the merkle proofs. Here is a mock on how to use it https://github.com/centrifuge/centrifuge-chain/blob/90fca013acaa449a887afcd5763cdfac7dc886e2/libs/proofs/src/mock.rs#L15

Also, blake256 is our  hashing function for our merkle trees. While its the same for T::Hashing, it would make it easier for us to use proofs library where the hash function is in our control. 
We would need to change it claims as well at some point

vedhavyas (commented):
Assuming benchamrks are done in different PR. Woulld prefer re/using `proofs` library for any sort of proof verification chain

pallets/crowdloan-claim/src/lib.rs
mustermeiszer: But we would need to remove the `Hasher` trait of the lib here and use the rust specific definition. Otherwise, we would need to manually implement `Hasher` for blake256. 
WDYT?

libs/proofs/src/lib.rs
```diff
@@ -1,6 +1,10 @@
 //! # Optimized Merkle proof verifier and bundle hasher
 //!
 //! This pallet provides functionality of verifying optimized merkle proofs and bundle hasher.
+
+// Ensure we're `no_std` when compiling for WebAssembly.
+#![cfg_attr(not(feature = "std"), no_std)]
```
vedhavyas, line 6: nice!

vedhavyas (commented):
LGTM. Didn't look benchmarks for crowdloan

pallets/crowdloan-reward/Cargo.toml
```diff
@@ -41,9 +42,11 @@ default = ['std']
 runtime-benchmarks = ["frame-benchmarking"]
 std = [
   'codec/std',
+  'sp-std/std',
+  'trait-crowdloan-reward/std',
   'frame-support/std',
   'frame-system/std',
-  'log/std',
+  'log',
```
vedhavyas, line 49: why do we need this crate ?

vedhavyas (commented):
LGTM. Question on log crate

pallets/crowdloan-reward/Cargo.toml
mustermeiszer: I am not sure. It was there when I took over and I kept it. I believe it is in the substrate template. Never tested if it is there for a reason of some macro using it.

PR #457, opened 2021-08-19
Testbed used to execute Centrifuge Chain behind Parity `substrate-api-sidecar`

tankofzion:
This PR implements the following features:
- A Docker-based testbed infrastructure for the Centrifuge Chain (built from source).
- Expose the Centrifuge Chain features behind the Parity Substrate API Sidecar.
- Provide Docker Compose file and Makefile to ease the bootstrap of the testbed infrastructure.

### DIFF
--- a/tools/testbed/docker/Dockerfile
+++ b/tools/testbed/docker/Dockerfile
@@ -1,0 +1,119 @@
+
+###############################################################################
+# Centrifuge Chain                                                            #
+# Cash on Steroids                                                            #
+#                                                                             #
+# tools/docker/Dockerfile                                                     #
+#                                                                             #
+# Handcrafted since 2020 by Centrifuge Foundation                             #
+# All rights reserved                                                         #
+#                                                                             #
+#                                                                             #
+# Description: Centrifuge chain's Docker image definition.                    #
+###############################################################################
+
+
+# -----------------------------------------------------------------------------
+# Sandbox (or compilation playground) stage
+# -----------------------------------------------------------------------------
+
+FROM rustlang/rust:nightly AS sandbox
+
+ARG ARG_RUST_TOOLCHAIN="nightly-2020-08-16"
+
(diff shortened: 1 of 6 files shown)


tools/testbed/docker/Dockerfile
```diff
@@ -0,0 +1,119 @@
...
+###############################################################################
+
+
+# -----------------------------------------------------------------------------
+# Sandbox (or compilation playground) stage
+# -----------------------------------------------------------------------------
+
+FROM rustlang/rust:nightly AS sandbox
+
+ARG ARG_RUST_TOOLCHAIN="nightly-2020-08-16"
```
mustermeiszer, line 22: When this is hardcoded here, what effect does changing the compiler version in the settings have? Wouldn't the version be (probably) not available then?

mustermeiszer (commented):
I am not very familiar with make. But I have a few questions regarding the capabilities of this extension:
* Does it spin-up a normal stand-alone chain?
* How does substrate-api-sidecar come into play here? (i.e. what is it good for and how does it work)
* Is it possible to create a parachain-testbed with this script?

tools/testbed/docker/Dockerfile
tankofzion: It's a default value. It is the value set in the "settings.mk" file which is the one used. I prefer to have a default value in case the Dockerfile is used outside the Makefile, in a manual Docker image build command (i.e. `docker image build -t repo/image_name, --target ... --file Dockerfile`).

tankofzion: The design rationale behind this simple Testbed infrastructure is to provide Kraken with an easy to use Centrifuge Chain deployment script, so that they can test Centrifuge Chain transactions through the sidecar (it's a type of proxy). There's no need for the parachain.

mustermeiszer (approved):

PR #466, opened 2021-09-09
Lower minimum vesting amount to 10 CFG/AIR

branan:
The previous lower bound of 1000 tokens has caused issues with
crowdloan rewards and other small vesting quantities.


mikiquantum (approved):
👍

PR #471, opened 2021-09-17
Add Epoch Validation

mikiquantum:


pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -560,13 +614,79 @@ pub mod pallet {
...
-			_reserve: T::Balance,
-			_target: &[(T::Balance, T::Balance)],
-			_solution: &[(Perquintill, Perquintill)],
-		) -> bool {
-			// TODO: Implement this
-			true
+		pub fn is_epoch_valid(
+			reserve: T::Balance,
+			max_reserve: T::Balance,
+			target: &[(T::Balance, T::Balance)],
```
mikiquantum, line 620: We can probably bundle the `current_tranche_value` as par of the target tuple.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -477,8 +511,28 @@ pub mod pallet {
...
+				// todo Define/Calculate current tranche values
+				let current_tranche_values = pool
+					.tranches
+					.iter()
+					.map(|_| Zero::zero())
+					.collect::<Vec<_>>();
+
 				ensure!(
-					Self::is_epoch_valid(epoch_reserve, &target, &solution),
+					Self::is_epoch_valid(
```
branan, line 528: The solution needs to be passed in here somewhere?

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -560,13 +614,79 @@ pub mod pallet {
...
+			reserve: T::Balance,
+			max_reserve: T::Balance,
+			target: &[(T::Balance, T::Balance)],
+			tranche_ratios: &[Perquintill],
+			current_tranche_values: &[T::Balance],
+		) -> DispatchResult {
+			let acc_supply: T::Balance = target
+				.iter()
+				.fold(Zero::zero(), |sum: T::Balance, tranche| {
+					sum.checked_add(&tranche.0).unwrap()
```
branan, line 627: This (and redeem below it) should be multiplying the target `Balance` with the solution `Perquintill` to get the actual Balance amounts for this epoch solution. Otherwise we're just checking the targets over and over with each solution we pass.

branan (changes requested):
A couple of small things noted, but the big one is that we're missing the solution being passed in to the validate function. This means we're only checking the constraints as if the full epoch was to be executed, and partial solutions won't be correctly validated.

pallets/tinlake-investor-pool/src/lib.rs
mikiquantum: I was accounting on the targets to take care of having that baked in already. But yeah, we either do it within this function or by the invoker. I am fine either way, so I will consume here the solution ratios.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -477,8 +511,28 @@ pub mod pallet {
...
 				ensure!(
-					Self::is_epoch_valid(epoch_reserve, &target, &solution),
+					Self::is_epoch_valid(
+						epoch_reserve,
+						pool.max_reserve,
+						&target,
+						&tranche_ratios,
+						&current_tranche_values
+					)
+					.is_ok(),
```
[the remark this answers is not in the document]
mikiquantum: Yes exactly, I didnt want to touch this part until you jumped on the scoring part, but I can make that swtich here.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -451,7 +462,30 @@ pub mod pallet {
...
+					.tranches
+					.iter()
+					.map(|_| Zero::zero())
+					.collect::<Vec<_>>();
+
+				if Self::is_epoch_valid(
+					epoch_reserve,
+					pool.max_reserve,
+					&epoch_targets,
+					&tranche_ratios,
```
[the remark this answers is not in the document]
mikiquantum: I can do that

pallets/tinlake-investor-pool/Cargo.toml
```diff
@@ -29,7 +29,8 @@ orml-traits = { git = "https://github.com/centrifuge/open-runtime-module-library
 [dev-dependencies]
 sp-core = { git = "https://github.com/centrifuge/substrate", default-features = false , branch = "master" }
 sp-io = { git = "https://github.com/centrifuge/substrate", default-features = false , branch = "master" }
-
+runtime-common = { path = "../../runtime/common"}
```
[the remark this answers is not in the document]
mikiquantum: Ok!

branan (approved):

PR #472, opened 2021-09-27
Ultimate refactoring of bridge, nft and registry pallets

tankofzion:
This PR aims at closing pallets FRAME v2 refactoring.

PR #473, opened 2021-09-28
Update to Polkadot v0.9.10

branan:


branan:
The tests for `pallet-bridge` and `pallet-registry` build with 0.9.10 but are failing. Since those tests aren't currently enabled and they aren't used in the Altair runtime, I'm going to say that getting those tests working is part of #472.

I did go ahead and enable the tests for a few other pallets, though, since they build+pass as-is and were likely not enabled as an oversight.

PR #474, opened 2021-09-30
Fix: Foreign claims and inhance error of validate unsigned

mustermeiszer:
The changes to the validate unsigned are here to reflect more
detailed error messages at failure with the `Costum(u8)` error
instead of the misleading `BadProof'`return value.

Furthermore, this fixes an issue where somebody could have a
signature of any parachain-relay-chain account combination and then use
a proof of any other relay-chain contribution to gain those
rewards.


pallets/crowdloan-claim/src/lib.rs
```diff
@@ -810,6 +830,11 @@ impl<T: Config> Pallet<T> {
 			Error::<T>::InvalidProofOfContribution
 		);
 
+		ensure!(
```
mikiquantum, line 833: Yes, the leaf always has to be generated or checked based on the raw data, thank you for catching it.

pallets/crowdloan-claim/src/lib.rs
```diff
@@ -704,35 +707,52 @@ pub mod pallet {
...
+										)
+										.unwrap_or(64_u64),
+									)
+									.propagate(true)
+									.build();
+							} else {
+								return InvalidTransaction::Custom(3).into();
+							}
+						} else {
+							return InvalidTransaction::Custom(2).into();
```
mikiquantum, line 749: I wouldn't give more hints to a potential attacker of where the code failed, I would stick to BadOrigin for all.

mikiquantum (approved):
LGTM, just one comment around the validate_unsigned error. In addition, do you think we can add a test case to cover this scenario of identity not linked to contribution?

PR #475, opened 2021-09-30
Dynamic Call-Filter via Migration-Pallet

mustermeiszer:
I think this could be a possible way to remove the call filter based on extriniscs after we migrated the data correctly.

Pallet still needs an extriniscs to actually change the storage with a sudo call.

## Tasks

- [x] ~~Needs bench on cloud machine again before merge~~ -> moved to #582

### DIFF
--- a/pallets/migration/src/lib.rs
+++ b/pallets/migration/src/lib.rs
@@ -6,10 +6,13 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use codec::{Decode, Encode};
-use frame_support::traits::{Contains, Currency};
-use scale_info::TypeInfo;
-
+use frame_support::{
+	dispatch::DispatchResult,
+	ensure,
+	traits::{Contains, Currency},
+};
 pub use pallet::*;
+use scale_info::TypeInfo;
 pub use weights::*;
 
 #[cfg(test)]
@@ -30,7 +33,7 @@ type BalanceOf<T> = <<T as pallet_vesting::Config>::Currency as Currency<
 >>::Balance;
 
 #[derive(Encode, Decode, PartialEq, Clone, TypeInfo)]
-pub enum Status {
+pub enum MigrationStatus {
 	Inactive,
 	Ongoing,
 	Complete,
@@ -100,13 +103,13 @@ pub mod pallet {
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
 
 	#[pallet::type_value]
-	pub fn OnCompletedEmpty() -> Status {
-		Status::Inactive
+	pub fn OnStatusEmpty() -> MigrationStatus {
+		MigrationStatus::Inactive
 	}
 
 	#[pallet::storage]
-	#[pallet::getter(fn completed)]
-	pub(super) type Completed<T: Config> = StorageValue<_, Status, ValueQuery, OnCompletedEmpty>;
+	#[pallet::getter(fn status)]
+	pub(super) type Status<T: Config> = StorageValue<_, MigrationStatus, ValueQuery, OnStatusEmpty>;
 
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
@@ -194,17 +197,7 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			ensure_root(origin)?;
 
-			let mut status = <Completed<T>>::get();
-
-			if status == Status::Inactive {
-				<Completed<T>>::set(Status::Ongoing);
-				status = Status::Ongoing;
-			}
-
-			ensure!(
-				status == Status::Ongoing,
-				Error::<T>::MigrationAlreadyCompleted
-			);
+			Self::activate_migration()?;
 
 			let num_accounts = accounts.len();
 			ensure!(
@@ -243,17 +236,7 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			ensure_root(origin)?;
 
-			let mut status = <Completed<T>>::get();
-
-			if status == Status::Inactive {
-				<Completed<T>>::set(Status::Ongoing);
-				status = Status::Ongoing;
-			}
-
-			ensure!(
-				status == Status::Ongoing,
-				Error::<T>::MigrationAlreadyCompleted
-			);
+			Self::activate_migration()?;
 
 			let current_issuance = pallet_balances::Pallet::<T>::total_issuance();
 			let total_issuance = current_issuance.saturating_add(additional_issuance);
@@ -282,17 +265,7 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			ensure_root(origin)?;
 
-			let mut status = <Completed<T>>::get();
-
-			if status == Status::Inactive {
-				<Completed<T>>::set(Status::Ongoing);
-				status = Status::Ongoing;
-			}
-
-			ensure!(
-				status == Status::Ongoing,
-				Error::<T>::MigrationAlreadyCompleted
-			);
+			Self::activate_migration()?;
 
 			ensure!(
 				vestings.len()
@@ -364,17 +337,7 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			ensure_root(origin)?;
 
-			let mut status = <Completed<T>>::get();
-
-			if status == Status::Inactive {
-				<Completed<T>>::set(Status::Ongoing);
-				status = Status::Ongoing;
-			}
-
-			ensure!(
-				status == Status::Ongoing,
-				Error::<T>::MigrationAlreadyCompleted
-			);
+			Self::activate_migration()?;
 
 			ensure!(
 				proxies.len()
@@ -427,11 +390,11 @@ pub mod pallet {
 			ensure_root(origin)?;
 
 			ensure!(
-				<Completed<T>>::get() == Status::Ongoing,
+				<Status<T>>::get() == MigrationStatus::Ongoing,
 				Error::<T>::OnlyFinalizeOngoing
 			);
 
-			<Completed<T>>::put(Status::Complete);
+			<Status<T>>::set(MigrationStatus::Complete);
 
 			Self::deposit_event(Event::<T>::MigrationFinished);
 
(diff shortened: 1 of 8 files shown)


mustermeiszer:
Works and is good for review

pallets/migration/src/lib.rs
```diff
@@ -351,5 +388,34 @@ pub mod pallet {
...
+			Self::deposit_event(Event::<T>::MigrationFinished);
+
+			Ok(())
+		}
+	}
+}
+
+impl<T: Config> Contains<<T as frame_system::Config>::Call> for Pallet<T> {
+	fn contains(c: &<T as frame_system::Config>::Call) -> bool {
+		if !pallet::Completed::<T>::get() {
```
vedhavyas, line 415: why not just add this filter to the BaseFilter ?
I'm not sure I fully understand.

mustermeiszer: Wdym?

mustermeiszer: This pallet is the filter of `system::config`. Just want to have the ability to remove the filter via an extrinisic.

vedhavyas: so this would not allow any calls including the one's we have enabled when the migration is underway ?

vedhavyas: nvm. Should work

mustermeiszer: This allows every call we provide the migration pallet filter type with. I pass our normal `BaseFilter` here. So its the same filter as we normally use during the start off the chain on Altair for example. Once we are through with this, we can call the `finalize` extrinsic and afterward all calls are enabled.

mustermeiszer:
@mikiquantum Just one more thing: Do we want to enable all calls or do we want to limit the allowed calls for some time after the migration? If second, we could add a second `Filter` that is used after migration is finalized.

pallets/migration/src/lib.rs
```diff
@@ -175,6 +192,11 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			ensure_root(origin)?;
 
+			ensure!(
+				<Completed<T>>::get() == false,
```
NunoAlexandre, line 196: `nit`: It is more idiomatic to use `! expr` than `expr == false` 
```suggestion
				!<Completed<T>>::get(),
```

pallets/migration/src/lib.rs
```diff
@@ -351,5 +388,34 @@ pub mod pallet {
...
+	}
+}
+
+impl<T: Config> Contains<<T as frame_system::Config>::Call> for Pallet<T> {
+	fn contains(c: &<T as frame_system::Config>::Call) -> bool {
+		if !pallet::Completed::<T>::get() {
+			T::Filter::contains(c)
+		} else {
+			true
+		}
```
NunoAlexandre, line 419: We can do some clean up here:

```suggestion		
		return pallet::Completed::<T>::get() || T::Filter::contains(c)
```

pallets/migration/src/lib.rs
```diff
@@ -76,11 +76,22 @@ pub mod pallet {
...
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
 
+	#[pallet::type_value]
+	pub fn OnCompletedEmpty() -> bool {
+		false
+	}
+
+	#[pallet::storage]
+	#[pallet::getter(fn completed)]
+	pub(super) type Completed<T: Config> = StorageValue<_, bool, ValueQuery, OnCompletedEmpty>;
```
NunoAlexandre, line 93: So, now we are calling this storage entry after a boolean status (completed or not). There's nothing wrong with that, but as I came across in the code below, [the boolean blindness ](https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/) becomes easily in the way.

My _suggestion_ here would be to rename this storage entry to `Status` (as in, _Migration status_), which would hold a value of type `Status`:

``` rust
 enum Status {
    Inactive,
    Ongoing, // If we need to cover this intermediate state
    Complete,
}
```

Then the `#[pallet::getter(fn completed)]` would become `#[pallet::getter(fn status)]`, and in the `ensure` expressions below we would have 

``` rust
ensure!(
  Self::status() != Status.Completed,
	Error::<T>::MigrationAlreadyCompleted
);
```

pallets/migration/src/lib.rs
```diff
@@ -76,11 +76,22 @@ pub mod pallet {
...
 		/// WeightInfo
 		type WeightInfo: WeightInfo;
+
+		type Filter: Contains<<Self as frame_system::Config>::Call>;
 	}
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
 
+	#[pallet::type_value]
```
NunoAlexandre, line 86: For my education, what does this mean?

NunoAlexandre (commented):
Sick work! Left some ideas and questions 🏄‍♂️

pallets/migration/src/lib.rs
mustermeiszer: I likem will change. Although with the new enum.

pallets/migration/src/lib.rs
mustermeiszer: I like. Will adapt with new filter enum.

pallets/migration/src/lib.rs
mustermeiszer: I like. Will change that. I also don't like booleans, when I am not too lazy.

pallets/migration/src/lib.rs
mustermeiszer: Nor sure. Would bet it automatically implements some trait so that we can use this function when our storage code does not exist yet and fails to decode in sp_io.

NunoAlexandre: TIL: https://crates.parity.io/frame_support/attr.pallet.html#type-value-pallettype_value-optional

pallets/migration/src/lib.rs
```diff
@@ -76,28 +85,28 @@ pub mod pallet {
...
-
-	#[cfg(feature = "std")]
-	impl Default for GenesisConfig {
-		fn default() -> Self {
-			Self {}
-		}
+	#[pallet::type_value]
+	pub fn OnCompletedEmpty() -> Status {
+		Status::Inactive
 	}
```
NunoAlexandre, line 105: Ok so this is actually a "default" value for this type. You could implement the `Default` trait. Leave it up to you.
But we do need to update this:

```suggestion
	pub fn DefaultMigrationStatus() -> Status {
		MigrationStatus::Inactive
	}
```

pallets/migration/src/lib.rs
```diff
@@ -351,5 +418,35 @@ pub mod pallet {
 				.into(),
 			)
 		}
+
+		/// This extrinsic disables the call-filter. After this has been called the chain will accept
+		/// all calls again.
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::finalize())]
```
NunoAlexandre, line 424: For my education, when do we know that we can call this extrinsic and flag the migration as finished?

pallets/migration/src/mock.rs
```diff
@@ -248,6 +269,15 @@ impl TestExternalitiesBuilder {
 	}
 }
 
+pub(crate) fn get_account() -> AccountId {
```
NunoAlexandre, line 272: Might be worth leaving a comment clarifying what account this is.

NunoAlexandre (commented):
Thanks a lot for the work put in!

The new revision is now a mix of the old approach using `Completed` and the new one using `Status`. We have things like:

``` rust
<Completed<T>>::set(Status::Ongoing);
```

which, unless you have written it, is super unclear for newcomers imo.

---

My suggestion to get this clean and polish is:

1. Rename `enum Status` to `enum MigrationStatus` (to be able to have a storage named `Status`

2. Rename `Completed<T: Config>` to `Status<T: Config>`

3. Rename `#[pallet::getter(fn completed)]` to `#[pallet::getter(fn status)]`

4. Check the storage state with `Self::status()`

5. Update the storage state with `<Status<T>>::set(..)`

pallets/migration/src/lib.rs
mustermeiszer: No, I want it explicitly to be OnEmpty. As there is no default, just a storage default in case storage is empty.

NunoAlexandre: Yes that's a good point 👍

pallets/migration/src/lib.rs
```diff
@@ -175,6 +194,18 @@ pub mod pallet {
...
+			if status == Status::Inactive {
+				<Completed<T>>::set(Status::Ongoing);
+				status = Status::Ongoing;
+			}
+
+			ensure!(
+				status == Status::Ongoing,
+				Error::<T>::MigrationAlreadyCompleted
+			);
+
```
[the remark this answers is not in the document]
mustermeiszer: No, this is not how storage works. Regarding storage. You will always get a copy of a storage value, never a reference. Although, some interfaces take care of updating the underlying storage. 

But as accessing storage is costly, as you need to interact with the underlying patricia-tree you want to reduce `get` and `set` calls. So I am getting it once, update my local copy and the storage value. If you wanna know more about the storage internals you can expand the `sp_io` crate and the `externalities` crate and then spend a day or two in the rabbit whole.

pallets/migration/src/lib.rs
mustermeiszer: You can query the state of the chain.

pallets/migration/src/mock.rs
mustermeiszer: It's just a dummy account. Any kind of 32 bytes.

mustermeiszer:
I will rename the enum and storage element. That is really confusing this way.

NunoAlexandre (approved):
Thanks for bearing with me and for the explanations 🙌🏻

vedhavyas (approved):
Nice work!

mustermeiszer:
Don't merge until we have clearance if this will go into Altair RTU3 or Not.

mustermeiszer:
This can be integrated and should go into the next PRTU for Centrifuge mainnet.

mustermeiszer:
closes #516

PR #480, opened 2021-10-05
Pallet: fixes and cleanup

vedhavyas:
This PR brings the following:
- Cleanup pallets 
- Add some fixes to registry pallets for tocc usages
- Clippy suggestions for Proofs lib

### DIFF
--- a/pallets/registry/src/lib.rs
+++ b/pallets/registry/src/lib.rs
@@ -126,9 +126,8 @@ use frame_support::{dispatch::DispatchError, ensure};
 
 use frame_system::ensure_signed;
 
-use proofs::Verifier;
+use proofs::{Proof, Verifier};
 
-use sp_core::H256;
 use sp_runtime::traits::Hash;
 
 use common_traits::BigEndian;
@@ -263,7 +262,7 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			let caller = ensure_signed(origin)?;
 
-			let registry_id = Self::create_new_registry(caller, info)?;
+			let registry_id = Self::create_new_registry(caller, info);
 
 			Self::deposit_event(Event::<T>::RegistryCreated(registry_id));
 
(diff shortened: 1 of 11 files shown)


pallets/registry/src/types.rs
```diff
@@ -69,30 +66,28 @@ pub struct CompleteProof<Hash> {
...
 /// Generates the leaf hash from underlying data, other hashes remain the same.
-///
-/// The Keccak is the hasher compatible with Ethereum.
 impl From<CompleteProof<H256>> for proofs::Proof<H256> {
 	fn from(mut proof: CompleteProof<H256>) -> Self {
 		// Generate leaf hash from property ++ value ++ salt
 		proof.property.extend(proof.value);
 		proof.property.extend(&proof.salt);
-		let leaf_hash = sp_io::hashing::keccak_256(&proof.property).into();
+		let leaf_hash = blake2_256(&proof.property).into();
```
branan, line 74: Where/how is this hash consumed? Do other bits of code need to be updated for the changed algorithm?

pallets/registry/src/lib.rs
```diff
@@ -361,16 +353,16 @@ impl<T: Config> VerifierRegistry for Pallet<T> {
...
-		caller: Self::AccountId,
-		owner_account: Self::AccountId,
-		asset_id: AssetId<Self::RegistryId, T::TokenId>,
-		asset_info: Self::AssetInfo,
-		mint_info: Self::MintInfo,
+		caller: T::AccountId,
+		owner_account: T::AccountId,
+		asset_id: AssetId<T::RegistryId, T::TokenId>,
+		asset_info: T::AssetInfo,
+		mint_info: MintInfo<T::Hash, T::Hash>,
```
tankofzion, line 365: Perhaps use the alias `SystemHashOf<T>` so that to be coherent with line 280

tankofzion (approved):
LGTM

pallets/registry/src/types.rs
vedhavyas: no just the proof verification that will be affected. completely independent

pallets/registry/src/lib.rs
vedhavyas: I would also sugest we remove SystemHashOf since its mostly just T:Hash. 
I would suggest we use alias for something that is not super straight forward to understand

pallets/registry/src/types.rs
vedhavyas: more context: we dont use kecckak excpet when we are bridging to eth. Only code that bridge to eth is BundledHasher, which we are going to remote soon

pallets/registry/src/lib.rs
vedhavyas: lets keep that for our nft studio refactor 😉

PR #520, opened 2021-11-09
WIP: Fix/try increasing proposal proportion

mustermeiszer:
Just a try to fix altiar.


mustermeiszer:
This branch has no more means, as we did only use it for pushing through the heavy block during the migration. Will close.

PR #521, opened 2021-11-11
Write off groups

vedhavyas:
Spec: https://centrifuge.hackmd.io/SKpjZILDQlCV0ORshhYAjw#Write-Off

### DIFF
--- a/pallets/loan/src/lib.rs
+++ b/pallets/loan/src/lib.rs
@@ -374,6 +374,9 @@ pub mod pallet {
 
 		/// Emits when there is no valid write off groups associated with given index
 		ErrInvalidWriteOffGroupIndex,
+
+		/// Emits when new write off group is invalid
+		ErrInvalidWriteOffGroup,
 	}
 
 	#[pallet::call]
@@ -463,16 +466,16 @@ pub mod pallet {
 				Error::<T>::ErrLoanIsActive
 			);
 
-			// calculate ceiling
-			let ceiling = loan_type.ceiling().ok_or(Error::<T>::ErrLoanTypeInvalid)?;
-			ensure!(ceiling > Zero::zero(), Error::<T>::ErrLoanValueInvalid);
+			// ensure loan_type is valid
+			let now = Self::time_now()?;
+			ensure!(loan_type.is_valid(now), Error::<T>::ErrLoanValueInvalid);
 
 			// ensure rate_per_sec >= one
 			ensure!(rate_per_sec >= One::one(), Error::<T>::ErrLoanValueInvalid);
 
-			// ensure loan_type is valid
-			let now = Self::time_now()?;
-			ensure!(loan_type.is_valid(now), Error::<T>::ErrLoanValueInvalid);
+			// calculate ceiling
+			let ceiling = loan_type.ceiling().ok_or(Error::<T>::ErrLoanTypeInvalid)?;
+			ensure!(ceiling > Zero::zero(), Error::<T>::ErrLoanValueInvalid);
 
 			// update the loan info
 			LoanInfo::<T>::mutate(pool_id, loan_id, |maybe_loan_info| {
@@ -515,6 +518,12 @@ pub mod pallet {
 
 			// check if the pool exists
 			pallet_pool::Pallet::<T>::check_pool(pool_id)?;
+
+			// ensure write off percentage is not more than 100
+			ensure!(
+				group.percentage <= One::one(),
+				Error::<T>::ErrInvalidWriteOffGroup
+			);
 
 			// append new group
 			let index = PoolWriteOffGroups::<T>::mutate(pool_id, |write_off_groups| -> u32 {
@@ -688,10 +697,26 @@ impl<T: Config> Pallet<T> {
 
 		// ensure debt is all paid
 		// we just need to ensure principal debt is zero
-		ensure!(
+		// if not, we check if the loan is written of 100%
+		match (
 			loan_info.principal_debt == Zero::zero(),
-			Error::<T>::ErrLoanNotRepaid
-		);
+			loan_info.write_off_index,
+		) {
+			// debt is cleared
+			(true, _) => Ok(()),
+			// debt not cleared and loan not written off
+			(_, None) => Err(Error::<T>::ErrLoanNotRepaid),
+			// debt not cleared but loan is written off
+			// if written off completely, then we can close it
+			(_, Some(write_off_index)) => {
+				let groups = PoolWriteOffGroups::<T>::get(pool_id);
+				let group = groups
+					.get(write_off_index as usize)
+					.ok_or(Error::<T>::ErrInvalidWriteOffGroupIndex)?;
+				ensure!(group.percentage == One::one(), Error::<T>::ErrLoanNotRepaid);
+				Ok(())
+			}
+		}?;
 
 		// transfer asset to owner
 		let asset = loan_info.asset_id;
--- a/pallets/loan/src/lib.rs
+++ b/pallets/loan/src/lib.rs
@@ -146,6 +146,28 @@ where
 			_ => None,
 		}
 	}
+
+	/// returns the present value of the loan adjusted to the write off group assigned to the loan if any
+	// pv = pv*(1 - write_off_percentage)
+	fn present_value_with_write_off(
+		&self,
+		write_off_groups: Vec<WriteOffGroup<Rate>>,
+	) -> Option<Amount> {
+		let maybe_present_value = self.present_value();
+		match self.write_off_index {
+			None => maybe_present_value,
+			Some(index) => maybe_present_value.and_then(|pv| {
+				write_off_groups
+					.get(index as usize)
+					// convert rate to amount
+					.and_then(|group| math::convert::<Rate, Amount>(group.percentage))
+					// calculate write off amount
+					.and_then(|write_off_percentage| pv.checked_mul(&write_off_percentage))
+					// calculate adjusted present value
+					.and_then(|write_off_amount| pv.checked_sub(&write_off_amount))
+			}),
+		}
+	}
 }
 
 pub type RegistryIdOf<T> = <T as pallet_nft::Config>::RegistryId;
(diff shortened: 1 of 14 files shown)


pallets/loan/src/lib.rs
```diff
@@ -819,6 +977,96 @@ impl<T: Config> Pallet<T> {
...
+					Error::<T>::ErrLoanNotActive
+				);
+
+				let maturity_date = loan_data
+					.loan_type
+					.maturity_date()
+					.ok_or(Error::<T>::ErrLoanTypeInvalid)?;
+				let now = Self::time_now()?;
+
+				// ensure loan's maturity date has passed
```
offerijns, line 1009: Ideally, write-offs should even be possible pre maturity date. in Tinlake on Ethereum this wasn't allowed, to simplify, but it would be really nice if we don't have to have this restriction in TOCC. I think the way the spec was written, this should not be an issue, this check can simply be removed.

pallets/loan/src/loan_type.rs
```diff
@@ -121,6 +121,8 @@ where
 			self.expected_loss_over_asset_maturity.is_positive(),
 			// maturity date should always be in future where now is at this instant
 			self.maturity_date > now,
+			// advance rate should not be more than 1
```
offerijns, line 124: Actually, in theory we could allow someone to loan more than the value at some point (i.e. undercollateralized loans).

offerijns (commented):
Looks great! Commented on just two minor things I spotted.

pallets/loan/src/loan_type.rs
vedhavyas: we can always remove this check when we are there. or is this something we are looking from the start ?

pallets/loan/src/lib.rs
vedhavyas: do you mean when admin write off the loan?

offerijns: Yeah indeed, it should ideally be possible for an admin to write off a loan before the maturity date.

pallets/loan/src/lib.rs
```diff
@@ -86,6 +98,14 @@ pub struct LoanData<Rate, Amount, AssetId> {
...
 	status: LoanStatus,
 	loan_type: LoanType<Rate, Amount>,
+
+	// whether the loan written off by admin
+	// if so, we wont update the write off group on this loan further from permission less call
+	admin_written_off: bool,
+	// write off group index in the vec of write off groups
+	// none, the loan is not written off yet
+	// some(index), loan is written off and write off details are found under the given index
+	write_off_index: Option<u32>,
```
NunoAlexandre, line 108: Is `write_off_index` dependant on `admin_written_off`? As in, can `write_off_index` be `Some(index)` while `admin_writter_off` being `false`?

pallets/loan/src/lib.rs
```diff
@@ -126,6 +146,28 @@ where
 			_ => None,
 		}
 	}
+
+	/// returns the present value of the loan adjusted to the write off group assigned to the loan if any
```
NunoAlexandre, line 150:
```suggestion
	/// returns the present value of the loan adjusted to the write off group assigned to the loan (if any)```

File: pallets/loan/src/lib.rs [hunk 5]
```diff
@@ -434,6 +503,74 @@ pub mod pallet {
...
+			// ensure this is coming from an admin origin
+			<T as Config>::AdminOrigin::ensure_origin(origin)?;
+
+			// check if the pool exists
+			pallet_pool::Pallet::<T>::check_pool(pool_id)?;
+
+			// ensure write off percentage is not more than 100
+			ensure!(
+				group.percentage <= One::one(),
+				Error::<T>::ErrInvalidWriteOffGroup
```
NunoAlexandre, line 525: I would find it helpful to be more specific in the naming of this variant and make it `ErrInvalidWriteOffGroupPercentage`.

pallets/loan/src/lib.rs
```diff
@@ -553,15 +692,31 @@ impl<T: Config> Pallet<T> {
...
 		ensure!(
 			loan_info.status == LoanStatus::Active,
-			Error::<T>::ErrLoanIsInActive
+			Error::<T>::ErrLoanNotActive
 		);
 
 		// ensure debt is all paid
 		// we just need to ensure principal debt is zero
-		ensure!(
+		// if not, we check if the loan is written of 100%
```
NunoAlexandre, line 700:
```suggestion
		// if not, we check if the loan is 100% written-off
```

pallets/loan/src/lib.rs
```diff
@@ -819,6 +977,97 @@ impl<T: Config> Pallet<T> {
...
+							// not written off by admin, and non admin trying to write off, then
+							// fetch the best write group available for this loan
+							math::valid_write_off_group(
+								maturity_date,
+								now,
+								write_off_groups.clone(),
+							)
+							.ok_or(Error::<T>::ErrNoValidWriteOffGroup)
+						}
+					}?;
```
NunoAlexandre, line 1041: Note that we are matching on the tuple `(loan_data.admin_written_off, override_write_off_index)` but you just case on the index.


```suggestion
					match override_write_off_index {
						// admin is trying to write off
						Some(index) => {
							// check if the write off group exists
							write_off_groups
								.get(index as usize)
								.ok_or(Error::<T>::ErrInvalidWriteOffGroupIndex)?;
							loan_data.admin_written_off = true;
							Ok(index)
						}
						
						None => {
							// non-admin is trying to write off but admin already did
							if loan_data.admin_written_off {
								return Err(Error::<T>::ErrLoanWrittenOffByAdmin.into());
							}

							let maturity_date = loan_data
								.loan_type
								.maturity_date()
								.ok_or(Error::<T>::ErrLoanTypeInvalid)?;

							// ensure loan's maturity date has passed
							ensure!(now > maturity_date, Error::<T>::ErrLoanHealthy);

							// not written off by admin, and non admin trying to write off, then
							// fetch the best write group available for this loan
							math::valid_write_off_group(
								maturity_date,
								now,
								write_off_groups.clone(),
							)
							.ok_or(Error::<T>::ErrNoValidWriteOffGroup)
						}
					}?;
```

pallets/loan/src/math.rs
```diff
@@ -52,10 +53,17 @@ pub fn calculate_principal_debt<Amount: FixedPointNumber, Rate: FixedPointNumber
...
+pub(crate) fn seconds_per_day() -> u64 {
+	3600 * 24
+}
+
 /// returns the seconds in a given normal year(365 days)
 /// https://docs.centrifuge.io/learn/interest-rate-methodology/
+#[inline]
 pub(crate) fn seconds_per_year() -> u64 {
-	3600 * 24 * 365
+	seconds_per_day() * 365
```
NunoAlexandre, line 66: I like this 👌

pallets/loan/src/math.rs
```diff
@@ -125,6 +133,29 @@ where
 		.and_then(|d| expected_cash_flow.checked_div(&d))
 }
 
+/// returns the valid write off group give the maturity date and current time
```
NunoAlexandre, line 136:
```suggestion
/// returns the valid write off group given the maturity date and current time
```

pallets/loan/src/math.rs
```diff
@@ -125,6 +133,29 @@ where
...
+/// since the write off groups are not guaranteed to be in a sorted order and
+/// we want to preserve the index of the group,
+/// we also pick the group that has the highest overdue days found in the vector
+pub(crate) fn valid_write_off_group<Rate>(
```
NunoAlexandre, line 140: A suggestion here would be to first sort the vec by `overdue_days` DESC and then find the first with a valid offset. That would remove all the procedural and mutable code. Leave it up to you!

pallets/loan/src/math.rs
```diff
@@ -267,3 +298,62 @@ mod tests {
 		)
 	}
 }
+
+#[test]
+fn valid_write_off_groups() {
```
NunoAlexandre, line 303: It also be interesting to test for invalid write off groups

pallets/loan/src/tests.rs
```diff
@@ -109,7 +109,7 @@ fn issue_loan() {
 	TestExternalitiesBuilder::default()
 		.build()
 		.execute_with(|| {
-			let owner: u64 = 1;
+			let owner: u64 = 10;
```
NunoAlexandre, line 112: For my education, could you explain this change? :)

pallets/loan/src/tests.rs
```diff
@@ -853,5 +855,483 @@ fn test_pool_nav() {
...
+					pool_id,
+					WriteOffGroup {
+						percentage: Rate::saturating_from_rational(group.1, 100),
+						overdue_days: group.0,
+					},
+				);
+				assert_ok!(res);
+			}
+
+			// index
```
NunoAlexandre, line 1153: Could we make this comment more descriptive please

NunoAlexandre (commented):
Great work Ved!

This has been a very educational review, TIL on write-offs.

Left some comments with some questions and suggestions for small improvements.

pallets/loan/src/lib.rs
vedhavyas: yes it is possible

pallets/loan/src/lib.rs
vedhavyas: We do not want to be too specific about it IMO

pallets/loan/src/tests.rs
vedhavyas: 1 is admin account. 10 is non admin

pallets/loan/src/math.rs
vedhavyas: we can't sort them, then out indexs will change and we will end picking the wrong write off group.

pallets/loan/src/tests.rs
vedhavyas: sure

pallets/loan/src/math.rs
vedhavyas: it is already tested down below

NunoAlexandre (approved):
LGTM 🚀

PR #528, opened 2021-11-26
Rt3

mustermeiszer:
PR for the RT3 of Altair. This upgrade deals with migrating from static keys for collators provided via `pallet_session` to collators provided by `collator_selection`.

A mean focus in reviews should lie on the migration. The spec for this can be found [here](https://centrifuge.hackmd.io/FlQ3jymZRVCC5Np2u8ouZw?view).

## Tasks
- [ ] Test locally
- [ ] Test Chachacha
- [ ] Define actually accounts for new invulnerables (probably our current nodes)


mustermeiszer:
close in favor of #534

Issue #530, opened 2021-11-30
Design a custom collator whitelisting pallet


NunoAlexandre:
Right now, we run a PoA setup where our validators/collators are listed in genesis and selected through the session pallet to go on producing blocks.

We want to build a pallet that allows us to whitelist specific collator candidates so that we can, also through this pallet, implement a custom `ValidatorRegistration` impl that checks both for our whitelisting flag as well as for the default `Session` impl of said trait.

Worth noting that this is not meant to be our long-term approach to collator selection but rather a temporary solution that allow us to have more control over the selected collators until we have picked the right staking solution for us, and to avoid a FCFS setup.

We can break this work in small deliverable PRs

- [x] 1. Implement and test the pallet itself with the expected use cases and requirements (https://github.com/centrifuge/centrifuge-chain/pull/535)
- [x] 2. Add benchmarks and auto-generate the right weights + add the pallet to the altair and development runtimes (https://github.com/centrifuge/centrifuge-chain/pull/542)

NunoAlexandre:
@mustermeiszer regarding (*A), if I got it right, this extrinsic is a sudo-only call. 

Could you also point me to the right labels I should add to this issue pls?

mustermeiszer:
@NunoAlexandre can you base your work upon https://github.com/centrifuge/centrifuge-chain/tree/rt4?

NunoAlexandre:
Done ✅

PR #535, opened 2021-12-01
pallet: Collator allowlist

NunoAlexandre:
Step 1 of #530 

> 1. Implement and test the pallet itself with the expected use cases and requirements

We add a new pallet, `pallet-collator-allowlist` which offers two extrinsics:

1. `add`, which adds a given collator to the allowlist
2. `remove`, which removes a given collator to the allowlist 

Intentionally, I am leaving tests, mocks, and benchmarking out of scope of this PR to move in smaller steps and first have us focused on the core functionality we want from this pallet.

**To Do**

- [x] Add tests 
- [x] Clean up list of dependencies


pallets/collator-allowlist/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+			ensure_root(origin)?;
+
+			ensure!(
+				!<Allowlist<T>>::contains_key(&collator_id),
+				Error::<T>::CollatorAlreadyAllowed
+			);
+			<Allowlist<T>>::insert(collator_id.clone(), ());
+			Self::deposit_event(Event::CollatorAdded(collator_id));
+			Ok(())
+		}
```
mustermeiszer, line 113: We should add a check here, that checks if the collator has already uploaded its keys to the session pallet. 
Maybe via wiring in the Session pallet as a `ValidatorRegistration` type.

NunoAlexandre: Hm. What if the collator had uploaded its keys to the session pallet, we verify that here and thus allowlist them, and afterwards they unload their keys?

Point being, if we don't have a way to reactively add and remove collators from the allow list based on the state of their keys in the session pallet, it's not very meaningful to add that check here.

mikiquantum: I like the idea of duplicating a bit the session key check. The main reason is that we enforce whoever we want to allow as collator to have the session keys before whitelisting (and we do not need to do any manual check). 
As @NunoAlexandre said they could remove the keys in between, but at least we have an initial check that would hopefully remove 90% of the readiness of a collator when onboarding.
In the case that they do remove keys in between whitelisting and registering, registering will fail (assuming empty `0x` is not a successful `isSome`). And in the case that they remove the keys after registration, then collator will not produce blocks in next session, and every session change the collator-selection logic will kick out those that did not produce blocks.
We of course keep the session key check as well in the `is_registered`.

mikiquantum (commented):
@NunoAlexandre Are you planing on adding tests after in the PR or a different one? Since some of the discussion below requires testing that an empty key `0x` would be effectively the same as _no keys_, I would add tests to this PR,  now that I think we are more or less aligned.

pallets/collator-allowlist/src/lib.rs
NunoAlexandre: Ok, that sounds reasonable to me and put in practise it will be handy, indeed. I will check whether there is a way to hook on `purge_keys` or so to keep this storage in sync with the `session` pallet, but that is something we can do afterwards.

pallets/collator-allowlist/src/weights.rs
```diff
@@ -0,0 +1,63 @@
...
+use sp_std::marker::PhantomData;
+
+/// Weight functions needed for pallet_collator_allowlist.
+pub trait WeightInfo {
+	fn allow() -> Weight;
+	fn remove() -> Weight;
+}
+
+/// The weight associated with allowlist a collator
+const ALLOWLIST_WEIGHT: Weight = 42 as Weight;
```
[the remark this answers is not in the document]
NunoAlexandre: Yeah for sure! I suggest we leave that for the follow up PR.

NunoAlexandre:
> @NunoAlexandre Are you planing on adding tests after in the PR or a different one? 

My intention is to avoid ending up with a PR that has a lot of parallel discussions and where import details can go easily unnoticed. I generally prefer to include tests together with a feature; this time I was thinking of first agreeing on the functionality and design of this pallet before testing it (as it could change a fair bit). I am happy to work on the tests and include them in this PR after we have the first round of reviews covered. How does that sound?

> Since some of the discussion below requires testing that an empty key `0x` would be effectively the same as _no keys_, I would add tests to this PR, now that I think we are more or less aligned.

For my education, could you expand on that? I am not sure I get how and where an empty key `0x` is a thing 🤓  Help is much appreciated!

mikiquantum:
Yep, I would add the tests to this PR after we are in agreement.
> For my education, could you expand on that? I am not sure I get how and where an empty key 0x is a thing 🤓 Help is much appreciated!

Since the user can either `purge_keys` or `set_keys` potentially with `0x` as the value, I just want to make sure either substrate code behaves as expected and/or our code takes into account this posiblity.

Example of `isSome` https://github.com/paritytech/substrate/blob/8bf9836f3a709f220ff74be23a4004146480eb79/frame/session/src/lib.rs#L896

pallets/collator-allowlist/src/lib.rs
```diff
@@ -98,17 +101,25 @@ pub mod pallet {
...
 		///   - `collator_id` is already part of the allowlist
 		#[pallet::weight(<T as pallet::Config>::WeightInfo::allow())]
 		pub fn add(origin: OriginFor<T>, collator_id: T::ValidatorId) -> DispatchResult {
 			ensure_root(origin)?;
 
+			ensure!(
+				pallet_session::Pallet::<T>::is_registered(&collator_id),
+				Error::<T>::CollatorKeysNotLoaded
+			);
+
```
mustermeiszer, line 114: Do we need to tightly-couple this pallet, with pallet-session? 
I would prefer to have an associated type that is bounded to `ValidatorRegistration` and wire the session pallet at runtime in.

I know, that this is mostly cosmetic at this point, but in my opinion using this loose-coupling hast two advantages and hence I would prefer to use this style as often as possible: 
* Testing can be mocked and does not depend on other pallets
* Refactoring is a lot easier in the future. 

Arguing against my case, that we specifically just want to check if the keys are loaded here, then I would prefer to not use the `ValidatorRegistration` interface but just check the storage directly as this makes it clear, what we want.

NunoAlexandre: @mustermeiszer both of your points are very good. So, first I did not appreciate designing this with such a thigh dependency style but ended up following the example I found at `pallet_anchors`, which seemed fair enough for this allowlist pallet given it's expected lifetime and given that it only makes sense - in practise - if used "on top" of the `pallet-session`.

> Testing can be mocked and does not depend on other pallets

Loosely coupling objects is indeed a major advantage when testing.

> Refactoring is a lot easier in the future.

As I pointed out, we don't expect a long lifetime for this pallet so I wouldn't put too much weight on that point in this case.

> Arguing against my case, that we specifically just want to check if the keys are loaded here, then I would prefer to not use the ValidatorRegistration interface but just check the storage directly as this makes it clear, what we want.

I appreciate the thought flexibility here. And I follow your sentiment that calling `Session::is_registered` is somewhat obscure for the intent at hand.

### Next steps

Given that it can be a good exercise for me to play with the loosely coupled pallet approach - which I just read through in the substrate docs `TIL` - I will give that a try and see what comes out. _If_ going that route introduces more indirection and less clarity than a tighter style, and given we only want this pallet for a short time period, I will keep it tight and just access storage directly to make the code more objective.

How does that sound?

NunoAlexandre: @mustermeiszer I have implemented the loose approach in https://github.com/centrifuge/centrifuge-chain/pull/535/commits/d90f7aec7693840c7f41b97d8e1dbc67736b55dd.

I like it. It's a trade-off and it was a good learning exercise for me, but overall I think this is good. I added a custom fn to deal with the unnatural language of "is_registered".

Let me know what you think, I am flexible.

pallets/collator-allowlist/src/lib.rs
```diff
@@ -0,0 +1,168 @@
+//! Collator Allowlist Pallet
+//!
+//! This pallet provides two extrinsics, one that allows sudo to
+//! add collator ids to an allowlist, and another one that allows
+//! sudo remove them.
```
mustermeiszer, line 5: sudo to remove them

pallets/collator-allowlist/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+//!
+//! We have this pallet implementing `ValidatorRegistration`, which,
+//! in addition to the default `Session` pallet implementation, also
+//! checks for the presence of a collator id in this allowlist.
+//!
+//! We do that to have tigher control over which collators get selected
```
mustermeiszer, line 11: tighter

pallets/collator-allowlist/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Add the given `collator_id` to the allowlist.
+		/// Fails if
+		///   - `origin` fails the `ensure_root` check
+		///   - `collator_id` did not yet load their keys into the session pallet
+		///   - `collator_id` is already part of the allowlist
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::allow())]
+		pub fn add(origin: OriginFor<T>, collator_id: T::ValidatorId) -> DispatchResult {
```
mustermeiszer, line 112: maybe sync `WeightInfo::allow()` and extrinsic `add()` to the same name? Or is there some reasoning behind this?

pallets/collator-allowlist/src/weights.rs
```diff
@@ -0,0 +1,63 @@
...
+
+/// The weight associated with removing a collator
+const REMOVE_WEIGHT: Weight = 24 as Weight;
```
mustermeiszer, line 40: Is this auto-generated? Looks like they updated benchmarking then.

mustermeiszer (approved):
Super neat!! Just some really minor spelling errors and one remark regarding naming of weight method and extrinsic method.

mustermeiszer:
@NunoAlexandre as we are merging this into rt4 would you mind opening another PR for integrating this pallet into altair and development runtime?

pallets/collator-allowlist/src/lib.rs
NunoAlexandre: Thanks for the catch 🌟

pallets/collator-allowlist/src/lib.rs
NunoAlexandre: Thanks for the catch again :D

pallets/collator-allowlist/src/lib.rs
NunoAlexandre: Good catch! The reason for the misalignment was that first we went from `whitelist` to `allowlist` but then having an extrinsic `allowlist` didn't read well so I change that to `add` and forgot to update the weight. Thanks for catching this, it's now in sync 😎

pallets/collator-allowlist/src/weights.rs
NunoAlexandre: No, it's manual. Sorry, I didn't realise that this was meant to be auto-generated, I just copied the one from the fees pallets to get started. I was planning on working on the benchmarking side of things and make these numbers realistic in a follow up PR anyway.

NunoAlexandre:
> @NunoAlexandre as we are merging this into rt4 would you mind opening another PR for integrating this pallet into altair and development runtime?

Sure thing!

pallets/collator-allowlist/src/weights.rs
NunoAlexandre: To be clear, I used `42` and `24` as random placeholder values here until the right benchmarking weight outputs values were calculated.

NunoAlexandre:
Merging this and will follow up with 2) adding the benchmarks and proper weights and 3) adding this pallet to the altair and the development runtime.

PR #539, opened 2021-12-02
Pool Roles and actions

vedhavyas:
Spec: https://centrifuge.hackmd.io/Lga8cxI9Tia-u0ZPXrt84g


pallets/loan/src/mock.rs
```diff
@@ -224,6 +231,14 @@ impl Default for TestExternalitiesBuilder {
 	}
 }
 
+parameter_types! {
+	pub const PoolAdmin: u64 = 1;
+	pub const Borrower: u64 = 2;
+	pub const RiskAdmin: u64 = 3;
+	pub const DropInvestor: u64 = 4;
```
offerijns, line 238: We're not going to be using the DROP/TIN token names anymore for TOCC as we'll have multiple tranches. It's probably fine in the mocks/tests, but ideally we'd just call it senior/junior investors.

libs/common-traits/src/lib.rs
```diff
@@ -103,26 +106,37 @@ pub trait PoolNAV<PoolId, Amount> {
...
-/// A trait that support pool reserve operations such as borrow and repay
-pub trait PoolReserve<Origin, AccountId> {
+/// Role can hold any type of role specific functions a user can do on a given pool.
+#[derive(Encode, Decode, Clone, Copy, PartialEq)]
+#[cfg_attr(any(feature = "std", feature = "runtime-benchmarks"), derive(Debug))]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum Role {
+	PoolAdmin,
+	Borrower,
+	PricingAdmin,
```
offerijns, line 116: This has a `PricingAdmin` while the spec doesn't, is that to separate issue from activate/price permissions?

offerijns (commented):
That was quick! 🔥

libs/common-traits/src/lib.rs
vedhavyas: yeap. It would better to keep them seperate though

pallets/loan/src/mock.rs
vedhavyas: true. Will change this

libs/common-traits/src/lib.rs
```diff
@@ -103,26 +106,37 @@ pub trait PoolNAV<PoolId, Amount> {
 	fn update_nav(pool_id: PoolId) -> Result<Amount, DispatchError>;
 }
 
-/// A trait that support pool reserve operations such as borrow and repay
-pub trait PoolReserve<Origin, AccountId> {
+/// Role can hold any type of role specific functions a user can do on a given pool.
+#[derive(Encode, Decode, Clone, Copy, PartialEq)]
+#[cfg_attr(any(feature = "std", feature = "runtime-benchmarks"), derive(Debug))]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum Role {
```
mustermeiszer, line 113: Maybe name `PoolRole` or similiar? Or are we expecting `Role` to be have roles not associated with a pool?

mustermeiszer (approved):
LGTM!

libs/common-traits/src/lib.rs
vedhavyas: Make sense!

PR #540, opened 2021-12-03
Fix migration benchmarking

NunoAlexandre:
Currently, building the benchmarks fails to compile with the following error:

``` rust
error[E0308]: mismatched types
  --> pallets/migration/src/benchmarking.rs:95:5
   |
93 |   ...   let storage_vesting_info: VestingInfo<<<T as pallet_vesting::Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>...
   |  _________________________________-
94 | | ...   T::BlockNumber> =
   | |_____________________- expected due to this
95 |   ...       pallet_vesting::Vesting::<T>::try_get(id).unwrap();
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `VestingInfo`, found struct `frame_support::BoundedVec`
```

It seems that the Api has changed, which we fix in this PR. Please check carefully whether my _fix_ is fair.

PR #541, opened 2021-12-03
Quick broom 🧹 

NunoAlexandre:
Back-porting the migration/benchmarking compilation error fix (https://github.com/centrifuge/centrifuge-chain/pull/540) and cleaning unused imports and unused code from `crowdloan-claim/src/benchmarking.rs`.


mustermeiszer (approved):
High five for that!

PR #542, opened 2021-12-03
collator_allowlist: Benchmark + include it in the altair & dev runtimes

NunoAlexandre:
Part 2 of #530.

- [x]  add benchmarks to the `collator_allowlist` pallet,
- [x] update its respective `weights` generated from the benchmarks
- [x] add the pallet to the `altair` and `development` runtimes.


pallets/collator-allowlist/Cargo.toml
```diff
@@ -14,6 +14,7 @@ targets = ['x86_64-unknown-linux-gnu']
 codec = { package = 'parity-scale-codec', version = '2.0.0', features = ['derive'] , default-features = false }
 scale-info = { version = "1.0", default-features = false, features = ["derive"] }
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.12" }
+frame-benchmarking = { git = "https://github.com/paritytech/substrate",  default-features = false , optional = true , branch = "polkadot-v0.9.12" }
```
vedhavyas, line 17: lets move all the optional deps seperately.

pallets/collator-allowlist/src/benchmarking.rs
```diff
@@ -0,0 +1,37 @@
+#![cfg(feature = "runtime-benchmarks")]
+use super::*;
+use frame_benchmarking::{benchmarks, impl_benchmark_test_suite};
+use frame_system::RawOrigin;
+
+benchmarks! {
+	// Add a collator to the allowlist
+	add {
+		let collator_id = get_account::<T>();
```
vedhavyas, line 9: you would need to whitelist the account so that benchmarking wouldn't consider the key underneath for change

pallets/collator-allowlist/src/benchmarking.rs
```diff
@@ -0,0 +1,37 @@
...
+	}: add(RawOrigin::Root, collator_id.clone())
```
vedhavyas, line 10:
```suggestion
	}: _(RawOrigin::Root, collator_id.clone())
```

pallets/collator-allowlist/src/benchmarking.rs
```diff
@@ -0,0 +1,37 @@
...
+		assert!(<Allowlist<T>>::get(collator_id).is_none(), "Collator should have been removed");
+	}
+}
+
+impl_benchmark_test_suite!(Pallet, crate::mock::new_test_ext(), crate::mock::Test,);
+
+// Return an account that will be included as part of the initial
+// state of the pallet_session in the runtime used for benchmarking.
+fn get_account<T: Config>() -> T::ValidatorId {
+	let pub_key: [u8; 32] = [
```
vedhavyas, line 31: create a new account using benchmarking utils instead of static ones.

pallets/collator-allowlist/src/lib.rs
```diff
@@ -47,7 +49,10 @@ pub mod pallet {
 		type WeightInfo: WeightInfo;
 
 		/// The Validator Id type
-		type ValidatorId: Member + Parameter + MaybeSerializeDeserialize;
+		type ValidatorId: Member
+			+ Parameter
+			+ MaybeSerializeDeserialize
+			+ sp_std::convert::TryFrom<Self::AccountId>;
```
vedhavyas, line 55: what does ValidatorId repesent here ?
Why can't we use `IsType` if the they are of the same types?

pallets/collator-allowlist/src/weights.rs
```diff
@@ -1,21 +1,21 @@
...
-//! DATE: 2021-06-18, STEPS: `[50, ]`, REPEAT: 100, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
+//! DATE: 2021-12-05, STEPS: `100`, REPEAT: 200, LOW RANGE: `[]`, HIGH RANGE: `[]`
 //! EXECUTION: Some(Wasm), WASM-EXECUTION: Compiled, CHAIN: Some("altair-dev"), DB CACHE: 128
 
 // Executed Command:
 // target/release/centrifuge-chain
 // benchmark
 // --chain=altair-dev
-// --steps=50
```
vedhavyas, line 11: I would suggest keeping it to 50 and 100 since the higher the steps are the more the weight saturation leading to smaller weights

vedhavyas (commented):
Looks good!

pallets/collator-allowlist/Cargo.toml
NunoAlexandre: That doesn't seem possible unfortunately. 
If I move this to `[dev-dependencies]`, I need to drop the `optional = true` flag, or else it fails with:

```
dev-dependencies are not allowed to be optional: `frame-benchmarking`
```

But by keeping it, we get:

```
feature `runtime-benchmarks` includes `frame-benchmarking`, but `frame-benchmarking` is not an optional dependency
A non-optional dependency of the same name is defined; consider adding `optional = true` to its definition.
```

pallets/collator-allowlist/src/benchmarking.rs
NunoAlexandre: I followed the example of other pallet's benchmarking and in this case we do need to get this specific account. Could you also expand on what do you refer to as "benchmarking utils"?

pallets/collator-allowlist/src/lib.rs
NunoAlexandre: `ValidatorId` is any type that implements those traits. I guess we _could_ do that, but we wouldn't want our design to be unnecessarily restrictive but instead follow what the pallet_session, which we sort of extend, does: https://paritytech.github.io/substrate/master/src/pallet_session/lib.rs.html#380-384

pallets/collator-allowlist/src/weights.rs
NunoAlexandre: Note that I didn't set these numbers, this is auto-generated. Why did the command running on my machine bump those numbers or how do you keep them at bay?

pallets/collator-allowlist/src/benchmarking.rs
NunoAlexandre: I am not sure what you mean, sorry. Are you suggesting adding a new use case where the collator had already been added to the allowlist and thus we verify that we get an error in such case?

mustermeiszer: Wdytm @vedhavyas? The whitelisting. Not sure I get it...

pallets/collator-allowlist/Cargo.toml
vedhavyas: I meant moving to different group within dependcies so its easy to understand what are hard deps and what are benchmarking deps. Something like this - https://github.com/centrifuge/centrifuge-chain/blob/tocc/pallets/loan/Cargo.toml#L24

pallets/collator-allowlist/src/weights.rs
mustermeiszer: > I would suggest keeping it to 50 and 100 since the higher the steps are the more the weight saturation leading to smaller weights

What does weight saturation mean here? Isn't this to have more information about possible outliers in the measurements?

mustermeiszer (approved):
LGTM! Thanks for the work!

pallets/collator-allowlist/src/benchmarking.rs
vedhavyas: Something like this - https://github.com/centrifuge/centrifuge-chain/blob/tocc/pallets/loan/src/benchmarking.rs#L88
Account needs to be whitelisted so that any changes to this account's balance etc are not tracked by the benchmark pallets unless we want that to be tracked. 
https://github.com/centrifuge/centrifuge-chain/blob/tocc/runtime/development/src/lib.rs#L962

pallets/collator-allowlist/src/weights.rs
NunoAlexandre: Anyway, those numbers are set here, so it seems off-scope to change them for this pallet's benchmarking:
https://github.com/centrifuge/centrifuge-chain/blob/rt4/scripts/run_benchmark.sh#L16-L17

pallets/collator-allowlist/src/benchmarking.rs
vedhavyas: https://github.com/centrifuge/centrifuge-chain/blob/tocc/pallets/loan/src/benchmarking.rs#L133
Here is how you can generate an account coming from benchmarking utils

pallets/collator-allowlist/src/benchmarking.rs
NunoAlexandre: Thanks for the links, Ved. The advantage being that the benchmark doesn't take that collateral work into account when calculating the weights?

pallets/collator-allowlist/src/weights.rs
vedhavyas: > What does weight saturation mean here? Isn't this to have more information about possible outliers in the measurements?
This means, when you run benchmarks over multiple times, your host does all sorts of caches that would make this runt faster. In the real world, we can predict the which code path the wasm will take since we cant predict the kind of transactions tht would come in. 
The more number of time you run the same code in benchmarks, the efficient the host if caching it and giving much smaller weights than it would ideally take on real world

mustermeiszer: Thanks! Yes, makes sense.

pallets/collator-allowlist/src/benchmarking.rs
vedhavyas: yes

pallets/collator-allowlist/src/lib.rs
vedhavyas: Not sure I like this new type :-D
Since ValidatorRegistration has parametric type, we can directly set T::AccountId since the I see no use for having a seperate associate type.

I would like to move away from `TryFrom` since this introduces another Optional success which would ideally not be present since the types are same. If we want to follow pallet_session, then we can move to having a conversion type like they have it `ValidatorIdOf` so that mapping would be easier if the types are not same

NunoAlexandre: By directly setting `T::AccountId` as the type parameter of `ValidatorRegistration`, we make it impossible for this pallet to be used in runtimes where those two types are not the same. I will see what I can do 👍

pallets/collator-allowlist/src/benchmarking.rs
NunoAlexandre: @vedhavyas that looks interesting but I am not able to generate the exact same pub key that we need to make this work. I have tried `frame_benchmarking::account::<T::AccountId>("/Alice", 0, 0);` and `frame_benchmarking::account::<T::AccountId>("//Alice", 0, 0);` but both fail to match with the key used in the runtime we benchmark against.

pallets/collator-allowlist/src/lib.rs
vedhavyas: I dont like the tryFrom there but I'll leave this to you.

NunoAlexandre: I think we don't need that specific trait so we may be able do drop it, just verifying that 👍

pallets/collator-allowlist/src/benchmarking.rs
vedhavyas: hm I would have to check but I wasn't aware of the requirement for it to be predefined account.

pallets/collator-allowlist/src/benchmarking.rs
NunoAlexandre: Looking through this, it seems very useful to know about this but it doesn't seem quite as useful in this specific pallet's benchmarking. We don't play around with balances here and adding that extra code would add significant noise for a marginal - if any - difference or so it seems to me. what do you think, Ved?

vedhavyas (approved):

PR #546, opened 2021-12-07
Extend iohandler with default rpcs 

mustermeiszer:
This PR extends all nodes (dev, altair, cfg) with the default rpc methods of `PaymentApi`, `SystemApi`.

### DIFF
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -1,0 +1,67 @@
+// Copyright 2021 Centrifuge Foundation (centrifuge.io).
+//
+// This file is part of the Centrifuge chain project.
+// Centrifuge is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version (see http://www.gnu.org/licenses).
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+//! Centrifuge specific rpc endpoints (common endpoints across all environments)
+
+use pallet_transaction_payment_rpc::{TransactionPayment, TransactionPaymentApi};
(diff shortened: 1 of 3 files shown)


src/rpc.rs
```diff
@@ -0,0 +1,67 @@
...
+
+	io.extend_with(SystemApi::to_delegate(FullSystem::new(
+		client.clone(),
+		pool,
+		deny_unsafe,
+	)));
+	io.extend_with(TransactionPaymentApi::to_delegate(TransactionPayment::new(
+		client.clone(),
+	)));
+
```
vedhavyas, line 65: do we keep the anchor api seperate ? I see adding anchor might not work for centrifuge runtime since it doesn't implement it yet I think. But altair and dev does alreaady right ?

NunoAlexandre (commented):
Some thoughts:

a. `FullDeps` seems unnecessary. It is only used to pass its fields to `create_full`, where the type is deconstructed. I would drop it and just pass the values directly, like `create_full(client, pool, deny_unsafe)`

b. We could DRY this block of code, now duplicated in 3 different places

``` rust
|client, pool, deny_unsafe| {
	let mut io = crate::rpc::create_full(FullDeps {
		client: client.clone(),
		pool,
		deny_unsafe,
	});
	io.extend_with(AnchorApi::to_delegate(Anchor::new(client)));
	Ok(io)
}
```

c. Would it be an idea to add `io.extend_with(AnchorApi::to_delegate(Anchor::new(client)));` to the `create_full`, since we already do `io.extend_with` for a couple of other bits there and we do it in the 3 places where we set up `io`?
  _Edit_: I see the answer to this last point here https://github.com/centrifuge/centrifuge-chain/pull/546#discussion_r764147927 👍

mustermeiszer:
WDYT about removing `FullDeps` but keeping the rest? It's redundant but probably change in the near future.

NunoAlexandre:
> WDYT about removing `FullDeps` but keeping the rest? It's redundant but probably change in the near future.

That sounds wise yeah, that's why I was asking 👍

NunoAlexandre (approved):
Sweet 🐝  🍯

PR #548, opened 2021-12-07
Prepares centrifuge runtime for first release

### LINKED ISSUE #547: Prepare Centrifuge Runtime for First Release
We will mimic the state of the Altair runtime during our parachain launch for centrifuge as close as possible. 
So the runtime for centrifuge should be as close as possible to the one we had for Altair at `v3.0.0-alpha2`.

mustermeiszer:
Mimic v3.0.0-aplha2 wrt to altair matching cfg runtime. 

closes #547


runtime/centrifuge/src/lib.rs
```diff
@@ -348,45 +347,38 @@ pub enum ProxyType {
 	Any,
 	NonTransfer,
 	Governance,
-	// Staking,
-	// Vesting,
+	_Staking, // Deprecated ProxyType, that we are keeping due to the migration
+	NonProxy,
```
NunoAlexandre, line 351:
```suggestion
	_Staking, 
	// Deprecated ProxyType, that we are keeping due to the migration
	NonProxy,
```

Did you mean like this?

NunoAlexandre (approved):
LGTM

runtime/centrifuge/src/lib.rs
mustermeiszer: No, it refers to staking. Do you prefer it above?

NunoAlexandre: yeah, I find it best to have the comment above the thing it targets for consistency and clarity

mustermeiszer: Alright, will change.

runtime/centrifuge/src/lib.rs
```diff
@@ -605,7 +598,7 @@ impl pallet_vesting::Config for Runtime {
 	type BlockNumberToBalance = ConvertInto;
 	type MinVestedTransfer = MinVestedTransfer;
 	type WeightInfo = pallet_vesting::weights::SubstrateWeight<Self>;
-	const MAX_VESTING_SCHEDULES: u32 = 28;
+	const MAX_VESTING_SCHEDULES: u32 = 3;
```
mikiquantum, line 601: Why 3 only? I guess we will need to change the migration auxiliary code to change the destination format when migrating vesting schedules from one chain to another.

mikiquantum (commented):
We need to change the `SS58Prefix`, we thought about keeping 36 (same as current standalone). Since there will not be much to do until we do the data migration, we can do another runtime to upgrade in case we cannot reuse.
Once we change that, looks good!

runtime/centrifuge/src/lib.rs
mustermeiszer: 3 is from Kusama I think. We are iterating in vesting pallet. Not sure about the number though.

mustermeiszer: We don't even need to change the migration from our off-chain side. Tested this already.

mustermeiszer:
> We need to change the `SS58Prefix`, we thought about keeping 36 (same as current standalone). Since there will not be much to do until we do the data migration, we can do another runtime to upgrade in case we cannot reuse.
> Once we change that, looks good!

It's set to 136 as I copied a lot from Altair. Will Change on Monday. To 36 sounds good then we don't need to change much in the front end and so on. And users will recognize their accounts.

mustermeiszer:
ss58 is set to 36 now!

mikiquantum (approved):
LGTM

PR #552, opened 2021-12-08
NAV Loan types: V1

vedhavyas:
Spec: https://centrifuge.hackmd.io/uJ3AXBUoQCijSIH9He-NxA#V1

This PR adds
- Credit line loan type
- Credit line with maturity loan type


pallets/loan/src/types.rs
```diff
@@ -110,12 +112,35 @@ where
...
+	) -> Option<Amount> {
 		// calculate current debt and present value
-		math::debt(self.principal_debt, self.accumulated_rate).and_then(|debt| {
-			self.loan_type
-				.present_value(debt, self.last_updated, self.rate_per_sec)
-		})
+		math::debt(self.principal_debt, self.accumulated_rate)
+			.and_then(|debt| {
+				// if the debt is written off, write off accordingly
+				match self.write_off_index {
```
branan, line 123: This should probably just be a `self.write_off_index.map_or(debt, |index| ....)`

pallets/loan/src/benchmarking.rs
```diff
@@ -224,11 +225,13 @@ fn activate_test_loan_with_defaults<T: Config>(
 	<T as LoanConfig>::Rate: From<Rate>,
 	<T as LoanConfig>::Amount: From<Amount>,
 {
-	let loan_type = LoanType::BulletLoan(BulletLoan::new(
+	let loan_type = LoanType::CreditLineWithMaturity(CreditLineWithMaturity::new(
```
branan, line 228: We should probably bench all three types, take the max, then set up the extrinsics to refund the difference based on the loan type actually used.

I think more complex benchmarking and making sure we don't use more block weight than our actual extrinisc times can be a future thing to work on, though

vedhavyas: true. Right now i took the one with more computation and storage combined. We have to fine tune these benchmarks down the line. Not a major difference between each types for prioritise this for V1 atleast.

pallets/loan/src/types.rs
vedhavyas: I like it

PR #554, opened 2021-12-09
Tinlake V1

vedhavyas:


branan (approved):
As far as the merge itself goes, this looks clean to me.

We just need to coordinate this with Polkadot updates for the runtime.

pallets/loan/src/loan_type.rs
```diff
@@ -0,0 +1,328 @@
...
+			LoanType::CreditLine(_) => None,
+			LoanType::CreditLineWithMaturity(clm) => Some(clm.maturity_date),
+		}
+	}
+
+	pub(crate) fn is_valid(&self, now: u64) -> bool {
+		match self {
+			LoanType::BulletLoan(bl) => bl.is_valid(now),
+			LoanType::CreditLine(cl) => cl.is_valid(),
+			LoanType::CreditLineWithMaturity(clm) => clm.is_valid(now),
```
mustermeiszer, line 46: Maybe introduce a `LoanType` trait that boundles `is_valid`, `present_value` etc. together to an interface or are the arguments needed for the different loan-type to heterogenous for that?

primitives/tokens/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+	() => {
+		pub struct TrancheToken<T>(core::marker::PhantomData<T>);
+
+		impl<T> pallet_tinlake_investor_pool::TrancheToken<T> for TrancheToken<T>
+		where
+			T: Config,
+			<T as Config>::PoolId: Into<u64>,
+			<T as Config>::TrancheId: Into<u8>,
+			<T as Config>::CurrencyId: From<CurrencyId>,
+		{
```
mustermeiszer, line 27: `From<CurrencyId>` refers to type of `CurrencyId` that we import in the context of this trait being used, right? I think we should have hidden imports here, that make it clear, to which `CurrencyId` type the `From` refers to!

runtime/common/src/currency.rs
```diff
@@ -0,0 +1,29 @@
...
+	<T as Config>::TrancheId: Into<u8>,
+	<T as Config>::CurrencyId: From<CurrencyId>,
+{
+	fn tranche_token(
+		pool: <T as Config>::PoolId,
+		tranche: <T as Config>::TrancheId,
+	) -> <T as Config>::CurrencyId {
+		CurrencyId::Tranche(pool.into(), tranche.into()).into()
+	}
+}
```
mustermeiszer, line 29: Shouldn't we use our own marco here?

runtime/common/src/lib.rs
```diff
@@ -117,6 +115,34 @@ pub mod types {
...
+		Copy,
+		Clone,
+		PartialEq,
+		Eq,
+		codec::CompactAs,
+		Debug,
+		TypeInfo,
+	)]
+	#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+	pub struct InstanceId(pub u128);
```
mustermeiszer, line 145: Do we wanna have a common `TrancheId` and `EpochId` to, if the rest is here?

mustermeiszer (approved):
Nice work, just minor remarks!

pallets/loan/src/lib.rs
```diff
@@ -0,0 +1,558 @@
...
+			+ TypeInfo
+			+ IsType<ClassIdOf<Self>>;
+
+		/// The LoanId/InstanceId type
+		type LoanId: Parameter
+			+ Member
+			+ MaybeSerializeDeserialize
+			+ Copy
+			+ TypeInfo
+			+ From<u128>
```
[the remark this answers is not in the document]
vedhavyas: yes. the we dont expect to offer 2^128 loans anyway 😉

pallets/loan/src/loan_type.rs
vedhavyas: no, make it a trait doen't help us with different func definition and args we need for speicifc loan types. 
That IMO would be too much abstraction

primitives/tokens/src/lib.rs
vedhavyas: I'm not sure if we want to keep this here. I would prefer to bundle this as part of pool pallet. thoughts @branan ?

runtime/common/src/currency.rs
vedhavyas: true! see the above. 
Also, macro doesn't seem of much use if the same impl is referred everywhere instead if generating one though macro
We can remove this macro TBH

pallets/loan/src/loan_type.rs
mustermeiszer: Yeah, reasonable.

runtime/common/src/lib.rs
vedhavyas: yes for sure. Maybe a seperate ticket?

libs/common-traits/Cargo.toml
```diff
@@ -21,6 +22,7 @@ sp-std = { git = "https://github.com/paritytech/substrate", default-features = f
 [features]
 default = ['std']
 std = [
+    "serde/std",
```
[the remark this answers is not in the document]
vedhavyas: I think so. I'm using this when its std, so I bundled serde in the same feature set.

PR #559, opened 2021-12-14
Use 27-bit fixed point for tranche interest rates

### LINKED ISSUE #558: Use 27 decimals for interest rates on lender side

branan:
This also stores the interest rate in a different fixed format which
avoids some redudant conversionss for every interest calculation

Fixes #558

### DIFF
--- a/pallets/tinlake-investor-pool/src/tests.rs
+++ b/pallets/tinlake-investor-pool/src/tests.rs
@@ -316,7 +316,7 @@ fn epoch() {
 		let pool = TinlakeInvestorPool::pool(0).unwrap();
 		assert_eq!(
 			pool.tranches[0].interest_per_sec,
-			Perquintill::from_parts(000000003170979198)
+			Rate::from_inner(1_000000003170979198376458650)
 		);
 		assert_eq!(pool.tranches[0].debt, 0);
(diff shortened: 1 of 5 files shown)


pallets/tinlake-investor-pool/src/tests.rs
```diff
@@ -316,7 +316,7 @@ fn epoch() {
 		let pool = TinlakeInvestorPool::pool(0).unwrap();
 		assert_eq!(
 			pool.tranches[0].interest_per_sec,
-			Perquintill::from_parts(000000003170979198)
+			Rate::from_inner(1_000000003170979198376458650)
```
mustermeiszer, line 319: Where is the number coming from?

mustermeiszer (approved):
LGTM!

pallets/tinlake-investor-pool/src/tests.rs
branan: This is 10% annual interest rate divided by (365 days * 24 hours * 60 minutes * 60 seconds).

The changes from the old test here are:
* Nine more digits of precision
* Adding a one to the front in storage, since previously the first thing we did when reading from storage was add 1

PR #560, opened 2021-12-15
Adds type to ensure root is admin

mustermeiszer:
This PR adds a type of `type RootOr<O>  = EnsureOneOf<EnsureRoot, O>`. When using this as the origin type for an admin Origin of `0`, this ensures, that root is always also able to call as admin.

I think as a general rule we should always use this type for any kind of admin origin. This ensures that democracy can always act as an admin.

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -45,7 +45,7 @@ pub mod types {
 	use sp_runtime::traits::{BlakeTwo256, IdentifyAccount, Verify};
 	use sp_std::vec::Vec;
 
-	pub type RootOr<O> = EnsureOneOf<AccountId, EnsureRoot<AccountId>, O>;
+	pub type EnsureRootOr<O> = EnsureOneOf<AccountId, EnsureRoot<AccountId>, O>;
 
 	/// An index to a block.
 	pub type BlockNumber = u32;
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -37,7 +37,6 @@ pub mod apis {
 
 /// Common types for all runtimes
 pub mod types {
-	use frame_support::traits::EnsureOrigin;
 	use frame_system::{EnsureOneOf, EnsureRoot};
 	use scale_info::TypeInfo;
 	#[cfg(feature = "std")]
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -37,12 +37,16 @@ pub mod apis {
 
 /// Common types for all runtimes
 pub mod types {
+	use frame_support::traits::EnsureOrigin;
+	use frame_system::{EnsureOneOf, EnsureRoot};
 	use scale_info::TypeInfo;
 	#[cfg(feature = "std")]
 	use serde::{Deserialize, Serialize};
 	use sp_core::{H160, U256};
 	use sp_runtime::traits::{BlakeTwo256, IdentifyAccount, Verify};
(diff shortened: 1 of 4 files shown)


mustermeiszer:
I still would like to get this in and ensure we always use the `RootOr<O>` type for admin origins.

mustermeiszer:
Would like to get this merged today. @vedhavyas and @NunoAlexandre if you have time 😃

runtime/common/src/lib.rs
```diff
@@ -37,13 +37,16 @@ pub mod apis {
...
 pub mod types {
+	use frame_system::{EnsureOneOf, EnsureRoot};
 	use scale_info::TypeInfo;
 	#[cfg(feature = "std")]
 	use serde::{Deserialize, Serialize};
 	use sp_core::{H160, U256};
 	use sp_runtime::traits::{BlakeTwo256, IdentifyAccount, Verify};
 	use sp_std::vec::Vec;
 
+	pub type RootOr<O> = EnsureOneOf<AccountId, EnsureRoot<AccountId>, O>;
```
NunoAlexandre, line 48: We better follow the `Ensure_` convention here.

```suggestion
	pub type EnsureRootOr<O> = EnsureOneOf<AccountId, EnsureRoot<AccountId>, O>;
```

NunoAlexandre (approved):
Looks good 🚀

mikiquantum (commented):
I am not sure we should change the behavior on `Ensure` checks that are not already exposing admin execution by root. It would mean diverting from Polkadot/Kusama on how is that being implemented for some core pallets (identity & democracy), for our custom ones I am fine enabling root.
On the other hand I like the idea of consolidating into that trait the usage of rootOr for those traits/operations that already require so.
Might need to update Development/runtime as well.

mustermeiszer:
@mikiquantum As far as I can see, Kusama and Polkadot already do this. 
- See definition: https://github.com/paritytech/polkadot/blob/8ea1a359ac30372e4abfd5dc585d33e7b31606de/runtime/kusama/src/lib.rs#L151-L154 
- And usage: https://github.com/paritytech/polkadot/blob/8ea1a359ac30372e4abfd5dc585d33e7b31606de/runtime/kusama/src/lib.rs#L754-L769

In my opinion, this also makes a lot of sense. Root should always be able to change stuff. Otherwise, we are excluding democracy from fixing things. (i.e. we will always need a runtime upgrade to fix things otherwise.)

mikiquantum:
Yes I agree, my point is more towards the items that are not already as such in the pokadot/kusama case, like: https://github.com/paritytech/polkadot/blob/8ea1a359ac30372e4abfd5dc585d33e7b31606de/runtime/kusama/src/lib.rs#L610

Our custom ones and the core ones that are already allowing root then lets do it

mustermeiszer:
Mmh. You are right, good point. But this only concern democracy origins that are defined to come from the council? I.e. these [here](https://github.com/paritytech/polkadot/blob/8ea1a359ac30372e4abfd5dc585d33e7b31606de/runtime/kusama/src/lib.rs#L599-L613)

WDYT about switching `ExternalOrigin`, `ExternalMajorityOrigin` and `ExternalDefaultOrigin` to only by council. This also makes sense, as `External` should not come from democracy itself, which root basically implies when we remove sudo.

mustermeiszer:
@mikiquantum WDYT about the current state?

mustermeiszer:
@mikiquantum are you fine with the current state?

mikiquantum (approved):

PR #567, opened 2021-12-21
Rename specs to genesis 

### LINKED ISSUE #566: Rename Genesis Spec - Altair & Centrifuge
Rename the genesis specs of Altair and Centrifuge. Both chains will have:
- Initial authorities (that take care of block production)
- Initial root account
- No initial `TotalIssuance`

mustermeiszer:
Rename the existing specs to `*-genesis-spec-*.json` to make it clear that those are specs that should not be changed. 

Closes #566


NunoAlexandre:
What do you think of having a `res/genesis` directory instead, with a Readme file re-enforcing the idea that the specs within that dir are final? That would help us by having what's final and what's volatile in separate directories.

mustermeiszer:
WDYT about the current state @NunoAlexandre ?

NunoAlexandre:
> WDYT about the current state @NunoAlexandre ?

LGTM, altho I was thinking of then dropping the `-genesis-` bit from the spec file names. But I also don't mind keeping it, as it keeps context which might be helpful if we copy paste this files around.

src/chain_spec.rs
```diff
@@ -324,7 +329,7 @@ pub fn charcoal_dev(para_id: ParaId) -> AltairChainSpec {
 
 pub fn development(para_id: ParaId) -> DevelopmentChainSpec {
 	let mut properties = Properties::new();
-	properties.insert("tokenSymbol".into(), "DAIR".into());
+	properties.insert("tokenSymbol".into(), "DEVEL".into());
```
NunoAlexandre, line 332: "Devel" 👿

mustermeiszer:
I would like to keep them for the reason you were mentioning and make using them less ambiguous.

NunoAlexandre (approved):
🏄‍♂️

PR #570, opened 2021-12-27
Collect pool orders

offerijns:
This PR adds an extrinsic to the pool pallet to collect executed orders (tranche tokens for investment orders and currency for redemption orders). The implementation uses a separate `calculate_collect` method. This is based on the original implementation in Solidity: https://github.com/centrifuge/tinlake/blob/develop/src/lender/tranche.sol#L204.

The differences are minor, the main one being that this accepts a `collect_n_epochs` parameter instead of `end_epoch`, as this can then be used for the weight calculation.

It also adds a metadata attribute to pools and the ability to update this. This is a draft to simplify UI integration of this pallet. In the future, this should be a BoundedVec (it's restricted to IPFS hashes) and use deposits.


pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -300,10 +316,23 @@ pub mod pallet {
...
+		OrdersCollected(
+			T::PoolId,
+			T::TrancheId,
+			T::EpochId,
+			T::AccountId,
+			T::Balance,
+			T::Balance,
+			T::Balance,
+			T::Balance,
+		),
```
NunoAlexandre, line 335: It seems that we could make use of the `OutstandingCollections` type here to simplify and make it neater.
 
```suggestion
		/// Fulfilled orders collected [pool, tranche, end_epoch, user, outstanding_collections]
		OrdersCollected(
			T::PoolId,
			T::TrancheId,
			T::EpochId,
			T::AccountId,
			OutstandingCollections<T::Balance>
		),
```

runtime/development/src/lib.rs
```diff
@@ -771,7 +771,7 @@ impl pallet_crowdloan_claim::Config for Runtime {
 
 parameter_types! {
 	pub const LoanPalletId: PalletId = PalletId(*b"pal/loan");
-	pub const MaxLoansPerPool: u64 = 200;
+	pub const MaxLoansPerPool: u64 = 10;
```
NunoAlexandre, line 774: For my education, what triggers this change?

scripts/init.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 cmd=$1
-parachain="${PARA_CHAIN_SPEC:-altair-dev}"
+parachain="${PARA_CHAIN_SPEC:-development}"
```
NunoAlexandre, line 6: this change seems off-scope

pallets/tinlake-investor-pool/src/tests.rs
```diff
@@ -287,14 +291,21 @@ fn epoch() {
 			CurrencyId::Usd,
 			10_000 * CURRENCY
 		));
+		assert_ok!(TinlakeInvestorPool::set_pool_metadata(
+			pool_owner.clone(),
+			0,
+			"QmUTwA6RTUb1FbJCeM1D4G4JaMHAbPehK6WwCfykJixjm3"
```
NunoAlexandre, line 297: Is it maybe worth leaving a comment explaining how you generated this value?

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
...
+				pool.last_epoch_executed
+			} else {
+				end_epoch
+			};
+
+			let mut payout_currency_amount: T::Balance = Zero::zero();
+			let mut payout_token_amount: T::Balance = Zero::zero();
+			let mut remaining_supply_currency = order.supply;
+			let mut remaining_redeem_token = order.redeem;
+			while epoch_idx <= parse_until_epoch
```
NunoAlexandre, line 895:
```suggestion
			
			while epoch_idx <= parse_until_epoch
```

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
 			};
 		}
 
+		pub(crate) fn calculate_collect(
```
NunoAlexandre, line 866: general comment for this function: this is _a lot_ of code and logic in a row. Do you see a way of maybe extracting some parts into self-sustained that can be composed and make this function more readable and easier to follow for non-experts?

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -187,6 +199,10 @@ pub mod pallet {
 		type Time: UnixTime;
 	}
 
+	/// The maximum length of pool metadata
+	// #[pallet::constant]
+	// type PoolMetadataLimit: Get<u32>;
+
```
NunoAlexandre, line 205: to be deleted?

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -508,6 +560,77 @@ pub mod pallet {
...
+				Self::deposit_event(Event::OrdersCollected(
+					pool_id,
+					tranche_id,
+					end_epoch,
+					who.clone(),
+					collections.payout_currency_amount,
+					collections.payout_token_amount,
+					collections.remaining_supply_currency,
+					collections.remaining_redeem_token,
+				));
```
mustermeiszer, line 629: Move this out of the `try_mutate` maybe?

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -508,6 +560,77 @@ pub mod pallet {
...
+				Ok(())
+			})
+		}
```
mustermeiszer, line 632: Currently, the `Query` type of `Order` is `ValueQuery`. Two scenarios here: 

a. Order for user is existent -> We work on this order
b. Order for user is non-existent -> We work on a default value or `UserOrder {supply: 0, redeem: 0, epoch: 0}`

It feels like **b.** is a bad thing. As after this call we will have an `OrdersCollected` event in place of a user that has no orders and we have a 
```rust 
UserOrder {
   supply: collections.remaining_supply_currency,
   redeem: collections.remaining_redeem_token,
   epoch: end_epoch + One::one()
}
``` 
struct in storage that should not be there.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
...
+		pub(crate) fn calculate_collect(
+			loc: TrancheLocator<T::PoolId, T::TrancheId>,
+			order: UserOrder<T::Balance, T::EpochId>,
+			pool: PoolDetails<T::AccountId, T::CurrencyId, T::EpochId, T::Balance>,
+			end_epoch: T::EpochId,
+		) -> OutstandingCollections<T::Balance> {
+			let mut epoch_idx = order.epoch;
+
+			// No collect possible in this epoch
+			if epoch_idx == pool.current_epoch {
```
mustermeiszer, line 875: Could we have a `<=` check here? Or why only early return when we have `==`?

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
...
+					remaining_redeem_token: order.redeem,
+				};
+			}
+
+			// It is only possible to collect epochs which are already over
+			let parse_until_epoch = if end_epoch > pool.last_epoch_executed {
+				pool.last_epoch_executed
+			} else {
+				end_epoch
+			};
```
mustermeiszer, line 889: `max(end_epoch, pool.last_epoch_executed)`

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
...
+
+			let mut payout_currency_amount: T::Balance = Zero::zero();
+			let mut payout_token_amount: T::Balance = Zero::zero();
+			let mut remaining_supply_currency = order.supply;
+			let mut remaining_redeem_token = order.redeem;
+			while epoch_idx <= parse_until_epoch
+				&& (remaining_supply_currency != Zero::zero()
+					|| remaining_redeem_token != Zero::zero())
+			{
+				if remaining_supply_currency != Zero::zero() {
```
mustermeiszer, line 899: Are the two `if`-clauses related in anyway? If not we could split them into two separate functions. Makes the `&& (remaining_supply_currency != Zero::zero() || remaining_redeem_token != Zero::zero())` go away and overall more readable.

mustermeiszer: I would parse the `epoch` as a parameter, as each `if`-clause currently accesses storage separately which is costly.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
...
+							payout_token_amount.checked_add(&amount_token).unwrap();
+						remaining_supply_currency =
+							remaining_supply_currency.checked_sub(&amount).unwrap();
+					}
+				}
+
+				if remaining_redeem_token != Zero::zero() {
+					let epoch = Epoch::<T>::try_get(&loc, epoch_idx)
+						.map_err(|_| Error::<T>::NoSuchPool)
+						.unwrap();
```
mustermeiszer, line 930: When could this get call fail?

mustermeiszer (commented):
Nice work! 
But I got some overall questions regarding the logic of the code. My hesitation mostly concerns the interaction with storage. Not sure, if my remarks are reasonable but would like to clarify them first.

pallets/tinlake-investor-pool/src/lib.rs
offerijns: That's neat! Didn't know that was possible.

runtime/development/src/lib.rs
offerijns: This is actually mainly a change for testing, probably shouldn't be committed. The weight of the `updateNav` extrinsic is based on this constant. It turns out, the weight of updating the NAV of 200 loans is over the block limit, at least in the development runtime, hence I had to lower it to be able to call the extrinsic.

pallets/tinlake-investor-pool/src/tests.rs
offerijns: It's just a test IPFS hash, it doesn't really matter what's entered here. I'll add a comment explaining this.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -508,6 +560,77 @@ pub mod pallet {
...
+				order.epoch <= pool.last_epoch_executed,
+				Error::<T>::EpochNotExecutedYet
+			);
+
+			let n_epochs: T::EpochId = collect_n_epochs.into();
+			let end_epoch: T::EpochId = if order.epoch + n_epochs > pool.last_epoch_executed {
+				pool.last_epoch_executed
+			} else {
+				(order.epoch + n_epochs).into()
+			};
```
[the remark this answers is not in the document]
offerijns: Should be min, so `min(order.epoch.checked_add(n_epochs).ok_or(Error::Overflow)?, pool.last_epoch_executed)`, I think. But yeah, that's definitely cleaner than the if-else.

pallets/tinlake-investor-pool/src/lib.rs
offerijns: I'm not sure if there are any considerations with this besides code cleanliness, but I put it here because I noticed the `deposit_event` is put within `try_mutate` in several other places in the code already.

pallets/tinlake-investor-pool/src/lib.rs
offerijns: Returning early here when `==` is because if you submitted a new order in the current epoch, then you can't collect yet; the epoch needs to be executed first. `<=` shouldn't be possible, as you can't submit a new order without collecting first (which reminds me, we need to implement a check for this).

pallets/tinlake-investor-pool/src/lib.rs
offerijns: It shouldn't, that would mean something went wrong somewhere else.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -508,6 +560,77 @@ pub mod pallet {
...
+				tranche_id,
+			};
+			let order = Order::<T>::try_get(&loc, &who).unwrap_or(UserOrder {
+				supply: Zero::zero(),
+				redeem: Zero::zero(),
+				epoch: Zero::zero(),
+			});
+			let pool = Pool::<T>::try_get(pool_id)
+				.map_err(|_| Error::<T>::NoSuchPool)
+				.unwrap();
```
mustermeiszer: Unwrapping here is not what we want. Use early return via ? instead 

```rust
let pool = Pool::<T>::try_get(pool_id)
				.map_err(|_| Error::<T>::NoSuchPool)?
```

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
...
+						.unwrap();
+
+					// TODO: this should round down, in favor of the system
+					let amount = T::BalanceRatio::checked_from_rational(
+						epoch.supply_fulfillment.deconstruct(),
+						Perquintill::ACCURACY,
+					)
+					.unwrap()
+					.checked_mul_int(remaining_supply_currency)
+					.unwrap();
```
mustermeiszer: Is it safe to `unwrap()` here or are we expecting some errors? I would like to remove all unwraps and rather return errors.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
...
+			let mut remaining_supply_currency = order.supply;
+			let mut remaining_redeem_token = order.redeem;
+			while epoch_idx <= parse_until_epoch
+				&& (remaining_supply_currency != Zero::zero()
+					|| remaining_redeem_token != Zero::zero())
+			{
+				if remaining_supply_currency != Zero::zero() {
+					let epoch = Epoch::<T>::try_get(&loc, epoch_idx)
+						.map_err(|_| Error::<T>::NoSuchPool)
+						.unwrap();
```
mustermeiszer, line 913: early return via ?. See comment above

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
...
+
+					// TODO: this should round down, in favor of the system
+					let amount = T::BalanceRatio::checked_from_rational(
+						epoch.redeem_fulfillment.deconstruct(),
+						Perquintill::ACCURACY,
+					)
+					.unwrap()
+					.checked_mul_int(remaining_redeem_token)
+					.unwrap();
```
mustermeiszer: Remove unwrap unless we are sure it never fails.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -740,6 +863,106 @@ pub mod pallet {
...
+
+					if amount != Zero::zero() {
+						let amount_currency = epoch
+							.token_price
+							.checked_mul_int(amount)
+							.unwrap_or(Zero::zero());
+
+						payout_currency_amount = payout_currency_amount
+							.checked_add(&amount_currency)
+							.unwrap();
```
mustermeiszer, line 950: Remove `unwrap()` and replace with  something like overflow error or so.

mustermeiszer (commented):
Some notes regarding the removal of unwraps. Some of them should definitely be replaced by early returns, some might be safe.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -751,6 +882,127 @@ pub mod pallet {
...
+				return Ok(OutstandingCollections {
+					payout_currency_amount: Zero::zero(),
+					payout_token_amount: Zero::zero(),
+					remaining_supply_currency: order.supply,
+					remaining_redeem_token: order.redeem,
+				});
+			}
+
+			// It is only possible to collect epochs which are already over
+			let parse_until_epoch = end_epoch.min(pool.last_epoch_executed);
```
NunoAlexandre, line 902: This value is only used on line 914 so I would move it just before that block.

Also, the name is misleading since it's not parsing anything, it just represents the upper epoch bound we are interested in:

 ```suggestion
			// It is only possible to collect epochs which are already over
			let end_epoch = end_epoch.min(pool.last_epoch_executed);
```

File: pallets/tinlake-investor-pool/src/lib.rs [hunk 21]
```diff
@@ -751,6 +882,127 @@ pub mod pallet {
...
+			let mut outstanding = OutstandingCollections {
+				payout_currency_amount: Zero::zero(),
+				payout_token_amount: Zero::zero(),
+				remaining_supply_currency: order.supply,
+				remaining_redeem_token: order.redeem,
+			};
+
+			// Parse remaining_supply_currency into payout_token_amount
+			outstanding =
+				Self::parse_invest_executions(&loc, outstanding, order.epoch, parse_until_epoch)?;
```
NunoAlexandre, line 914: This is great, thanks for breaking things into self-suficient functions!

Now we are passing a mutable value, mutate it, and re-assign it. Comparing to the alternative of just passing a mutable reference, I might like this more since it's more "functional"-like and it's easier to test. But I wonder if the reference approach is more efficient, considering that the mutations occur within a loop.

I would say, leave it as is and once we implement benchmarking for this we go over that. It would be a good idea to leave a `TODO` to remind us of that.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -751,6 +882,127 @@ pub mod pallet {
...
+					payout_token_amount: Zero::zero(),
+					remaining_supply_currency: order.supply,
+					remaining_redeem_token: order.redeem,
+				});
+			}
+
+			// It is only possible to collect epochs which are already over
+			let parse_until_epoch = end_epoch.min(pool.last_epoch_executed);
+
+			// We initialize the outstanding collections to the ordered amounts, and then fill the payouts based on executions in the accumulated epochs
```
NunoAlexandre, line 904:
```suggestion
			// We initialize the outstanding collections to the ordered amounts, 
			// and then fill the payouts based on executions in the accumulated epochs.
```

NunoAlexandre (commented):
Looking great, just a few notes!

pallets/tinlake-investor-pool/src/lib.rs
offerijns: Great idea, just added the `TODO`. Thanks!

NunoAlexandre: Pleasure!

PR #572, opened 2022-01-01
Preparing Antares Testnet

mikiquantum:


scripts/prepare-new-env-keys.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ "$#" -ne 1 ]; then
+	echo "Please provide the number of initial collators!"
+	exit 1
+fi
+
+NETWORK=${NETWORK:-altair}
```
mustermeiszer, line 10: I know its littel, but can we start making Centrifuge the default for all the scripts instead of Altair?

src/command.rs
```diff
@@ -50,7 +50,10 @@ impl IdentifyChain for dyn sc_service::ChainSpec {
 	fn identify(&self) -> ChainIdentity {
 		if self.id().starts_with("centrifuge") {
 			ChainIdentity::Centrifuge
-		} else if self.id().starts_with("altair") || self.id().starts_with("charcoal") {
+		} else if self.id().starts_with("altair")
+			|| self.id().starts_with("charcoal")
+			|| self.id().starts_with("antares")
+		{
```
mustermeiszer, line 56: Can we put these into development identiy?

mustermeiszer (approved):
LGTM! Thanks for the quick action @mikiquantum . 

One remark regarding `identity` of the chain. I don't like putting our dev envs all as `Altair` identity. But rather have a common `StagingDev` or make the `Development.`

src/command.rs
mikiquantum: There will be a development environment for centrifuge soon, before we go live in March. Charcoal will most likely disappear from this list. This two envs `Antares` & `Axis/Catalyst` should use the same as their parent network. `Development` is for non-long-term environments or internal only.

branan: This is only used to guess the right `ChainSpec` to use for custom JSON chain specifications - see https://github.com/centrifuge/centrifuge-chain/pull/572/files#diff-de6b9cfd1a256da7900d35e26aa7279bd713110e68e91b52616befb7397848dfR90-R101

PR #575, opened 2022-01-03
wip: Emulate statemine's XCM config

NunoAlexandre:
Our first use case for XCM is to allow bridging NFTs from Statemine into our parachain.

Here, we start by emulating (aka copy-paste) Statemine's XCM config.


NunoAlexandre:
Supersed by `nuno/xcm-playground`

PR #573, opened 2022-01-03
Tinlake Naming updates

branan:
Resolves #561

PR #576, opened 2022-01-04
Permissions as its own pallet

mustermeiszer:
This implements a pretty "generic" permissions pallet. Where the actual types and checks are implemented by our types in the common sections of our runtimes. 

Although, this is WIP it would like to have a review already on this, in order for us to sync if we want to proceed with this approach.

## Tasks
- [x] Implement tests
- [x] Implement mock
- [x] ~Benchmark~ -> moved to #577 
- [x] ~Docs~ -> moved to #578


runtime/common/src/impls.rs
```diff
@@ -163,3 +165,56 @@ impl From<u128> for InstanceId {
 		Self(v)
 	}
 }
+
+impl Default for PermissionRoles {
+	fn default() -> Self {
+		Self {
+			admin: AdminPoolRoles::empty(),
+			tranches: Vec::new(),
```
branan, line 173: If we wanted to be clever, we could probably also store tranches as a u32 bitfield, since we'll at some point want to set an upper limit on tranches allowed

pallets/permissions/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+#[cfg(test)]
+mod tests;
+
+#[cfg(feature = "runtime-benchmarks")]
+mod benchmarking;
+
+use frame_support::{dispatch::DispatchResult, pallet_prelude::*};
+use frame_system::pallet_prelude::*;
+
+pub trait Permissions<AccountId> {
```
branan, line 31: Something I've noticed about this API - right now it is up to the calling pallet to ensure the origin is an admin. I'd suggest we take the origin as a parameter and do that checking here, so that it's centralized.

If we want pallets to be able to invoke add/rm without the admin origin, we can potentially allow `root` as an admin origin for everything. This would also allow democracy to make changes in the case that the admin account is lost.

runtime/development/src/lib.rs
```diff
@@ -784,10 +785,19 @@ impl pallet_loan::Config for Runtime {
...
 	type WeightInfo = pallet_loan::weights::SubstrateWeight<Self>;
 	type MaxLoansPerPool = MaxLoansPerPool;
 }
 
+impl pallet_permissions::Config for Runtime {
+	type Event = Event;
+	type Location = PoolId;
+	type Role = pallet_loan::PoolRole;
+	type Storage = PermissionRoles;
+	type AdminOrigin = EnsureProportionAtLeast<_1, _2, AccountId, CouncilCollective>;
```
branan, line 798: Might make more sense to require Root here, and the council can only fast-track? I'm really unsure of the reasons why we'd want something to be deomcracy-able vs council-able

pallets/permissions/src/lib.rs
branan: Sorry, in this case I mean the _pool_ admin, not the democracy/root "AdminOrigin"

branan: The other option would be to move `approve_role_for` and `revoke_role_for` into the permissions pallet, then the API here only needs to be checking if an account is authorized?

The overall goal in my mind is
1) make it foolproof for other pallets to interact with this one, by not relying on external code to do permissions checks
2) ensure we have a council/democracy safety valve in case a pool admin loses their key

runtime/common/src/impls.rs
mustermeiszer: We can definitely do this. My reasoning here was, that due to the u8 we will have at most 256 distinct tranches per pool. And iterating over this range seemed okish. But starting with 32 distinct tranches is fine for me.

runtime/development/src/lib.rs
mustermeiszer: Ah. Yes definitely. This was just to please the compiler as fast as possible. I have an open PR already, that I will merge that takes care of ensuring we always have root in place. I would prefer to go the kusama/polkadot route, that democracy should be able to do anything as admin as long as it is not distinct to be a council call, like the specific origin (forgot the name) that is defined in democracy.

pallets/permissions/src/lib.rs
mustermeiszer: I am not sure I follow your idea. 

* Do you want the trait to contain a caller, i.e. the origin, that this trait method is called from? 

I would like to not do this. My idea was, that the pallet-permissions provides root (i.e. the `AdminOrigin`) access to the storage. In this way root (if we set the `AdminOrigin` correctly) will always be able to fix permissions. 

Other pallets using the `trait Permissions` should not have to deal with carrying around an Origin in order to call this. I think it is fair to assume, that pallets should be able to just add and remove permissions from `AccountId`s.

mustermeiszer: Regarding the `approve_role_for` and the `revoke_role_for` I think we can remove them. I just did not wanted to remove as much of the API as I was not sure about the reasoning here. 
I guess in case of pool-creation, the pallet-pool should take care of inserting an `PoolAdmin` into the permissions pallet.

mustermeiszer:
This is ready for review. Missing docs and missing benches, but I would move this to a later PR so we can get this in for audit!

branan:
@mustermeiszer This is updated to be able to handle zero-indexed tranches now. It does rely on some manual bit-twiddling, but it's tucked away in an impl.

pallets/permissions/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+	type Location;
+	type Role;
+	type Error;
+
+	fn clearance(location: Self::Location, who: AccountId, role: Self::Role) -> bool;
```
offerijns, line 38: I think the previous naming was simpler, "clearance" is less explicit about what it means.

```suggestion
	fn has_role(location: Self::Location, who: AccountId, role: Self::Role) -> bool;
```

runtime/common/src/impls.rs
```diff
@@ -163,3 +165,56 @@ impl From<u128> for InstanceId {
 		Self(v)
 	}
 }
+
+impl Default for PermissionRoles {
+	fn default() -> Self {
+		Self {
+			admin: AdminRoles::empty(),
+			tranches: TrancheInvestors::empty(),
```
offerijns, line 173: Could we rename this? Right now, this causes the usage to be `self.tranches.contains(id.into())`, which implies you're checking if this id is a tranche, instead of a tranche investor. 

```suggestion
			admins: AdminRoles::empty(),
			trancheInvestors: TrancheInvestors::empty(),
```

pallets/permissions/src/lib.rs
mustermeiszer: I think `clearance` describes it good with respect to the trait definition. But I understand your point. I don`t like `has_role`. What about `has_permission` to make it more obvious in contrast to `clearance`?

offerijns: Yeah I think `has_permission` is also good!

mustermeiszer:
@branan should we bring in your https://github.com/centrifuge/centrifuge-chain/pull/573 PR? Otherwise, I would rebase now and merge.

runtime/common/src/lib.rs
```diff
@@ -146,6 +152,30 @@ pub mod types {
...
+	bitflags::bitflags! {
+		/// The current admin roles we support
+		#[derive(codec::Encode, codec::Decode,  TypeInfo)]
+		pub struct AdminRoles: u32 {
+			const POOL_ADMIN = 0b00000001;
+			const BORROWER  = 0b00000010;
+			const PRICING_ADMIN = 0b00000100;
+			const LIQUIDITY_ADMIN = 0b00001000;
+			const MEMBER_LIST_ADMIN = 0b00010000;
+			const RISK_ADMIN = 0b00100000;
```
vedhavyas, line 165: does this mean we cannot have more than 8 roles defined. or we can we extend to 16bit without any migration later in case the situation arises ?

runtime/common/src/lib.rs
```diff
@@ -146,6 +152,30 @@ pub mod types {
...
+		}
+	}
+
+	#[derive(codec::Encode, codec::Decode, TypeInfo, Debug, Clone, Eq, PartialEq)]
+	pub struct TrancheInvestors(pub(crate) u32);
+
+	/// The structure that we store in the pallet-permissions storage
+	/// This here implements trait Properties.
+	#[derive(codec::Encode, codec::Decode, TypeInfo, Clone, Eq, PartialEq, Debug)]
+	pub struct PermissionRoles {
```
vedhavyas, line 175: I would like if we move these to either common libs or bring the PoolRole here. Seems two tightly connected with each other

mustermeiszer: I have thought about this too. I don't want this in the common-libs, as it clearly is a runtime-type. And I did not want to move the `PoolRole` here, as we then have dependency in our pallets on `runtime-common`. 😄 

But I also think you got a valid point here. I am fine with leaving it as it is, but if we want to change something I would move `PoolRole` here.

runtime/common/src/lib.rs
mustermeiszer: The `AdminRoles`supports 32 roles without another storage migration.

runtime/common/src/impls.rs
```diff
@@ -163,3 +165,90 @@ impl From<u128> for InstanceId {
 		Self(v)
 	}
 }
+
+impl Default for PermissionRoles {
+	fn default() -> Self {
+		Self {
+			admin: AdminRoles::empty(),
+			tranche_investor: TrancheInvestors::empty(),
```
vedhavyas, line 173: Tranche investors are the memebers that can hold restricted tokens with specific valid time. This cannot be part of this Role as this will ideally come from restricted token pallet. 

Pool pallet will check if the at the time of supply if the the investor is member that can hold the restricted tokens.
cc: @offerijns

runtime/common/src/lib.rs
mustermeiszer: Probably my main reason against having it in the common-libs is that it does not belong into the traits module. But maybe its time to open a types module. Then we can move it there. WDYT?

runtime/common/src/impls.rs
vedhavyas: With memeberlist, there will be minumum valid time for the member per pool/tranche. That Minumum valid time ideally be present in the  restricted tokens. When a MemebrlistAdmin tries to add a new tranche investor or update the existing tranche investor, they shoukd ideally add that through the restricted tokens pallet. 

If the permission pallet role contains the tranche investor, then it should ideally get the MinumumValid time for that tranche in a pool from Resricted tokens. So this means, permissions pallets should know what a Restricted token is. I feel that defeats the purpose of the pallet being generic

mustermeiszer: But this would be possible. Either we adapt the TrancheInvestor to be of type 
```rust
TrancheInvestor(u32, Since)
```
where since simply stores the moment in time or a block number when this was added.

Or we store the moment in time, when the restricted-token pallets adds an tranche investor in the pallet-restricted-tokens and it can check it there.

mustermeiszer: Ah, we would need to go with the second approach, as the actual type in storage is never exposed in the trait interface.

runtime/common/src/lib.rs
vedhavyas: either is fine by me as long the crates dont inter depend on each other

runtime/common/src/impls.rs
vedhavyas: The requirements for the restricted tokens are 
- to check if the member can hold the tokens at this instant
- Ability to add or modify a member with a minimum holding time by a Memberlist Admin
- (Good to have) to be able to update minimum holding time for each pool/tranche instead of default 7 days


> Can't you use the Permissions trait in the restricted tokens pallet and add/remove users there?

Yes but there would be two extrinsics that does the same. One in Permissions and other in restricted tokens. If we decide to not have that in restricted tokens, then permissions pallets should support variable admin origins where only a Memberlist admin can add/update the members, and also further fetches the minimum delay for the given tranche.

mustermeiszer: The pallet-permissions allows root basically to add and remove any permission. Which I think is fine. And users can purge their permissions if wanted. 

Other pallets, that need additional checks, should do themselves. At least this is my opinion.

vedhavyas: A pool owner should be able to update the admin level permissions for the given pool. Making this a democracy process is not required since its always up to the pool owner or manager.

> And users can purge their permissions if wanted.

This is not always the case 😉 

> Other pallets, that need additional checks, should do themselves. At least this is my opinion.

True but the permissions pallet should ideally not provide the these extrinsics by itself then. It should only act like a storage layer and provide internal calls for the pallets. 
Then pool will ideally have the extrinsincs that allow its pool owner to make these changes
and restricted tokens will be able to make changes after verifying the membership admin role.

mustermeiszer: So restricted tokens could do like: 

```rust
fn OnMinHoldingEmpty<T:Config> -> T::MinHolding { T::MinHolding::get() }

pub type MinHolding<T: Config> = StorageDoubleMap<
		_,
		Blake2_128Concat,
		T::PoolId,
		Blake2_128Concat,
		T::TrancheId,
		T::MinHolding,
                ValueQuery,
                OnMinHoldingEmpty
	>;
```

for the holding period and update this on a pool-tranche-basis if needed or use the default value.

mustermeiszer: Or are you worried that in another place of the codebase we change the permissions of `TrancheInvestor`s?

vedhavyas: We just need to doublemap (pool and tranche) to a moment on the restricted tokens. This will be checked along with MemberlistAdmin role before adding/removing permisssions on the permissions pallet

> Or are you worried that in another place of the codebase we change the permissions of TrancheInvestors?

yes, there should be no other way this permissions will be changed through externisics apart from restricted tokens

runtime/common/src/impls.rs
```diff
@@ -163,3 +166,139 @@ impl From<u128> for InstanceId {
...
 }
+
+#[derive(codec::Encode, codec::Decode, TypeInfo, Debug, Clone, Eq, PartialEq)]
+pub(crate) struct TrancheInvestorInfo {
+	pub(crate) tranche_id: TrancheId,
+	pub(crate) permissioned_at: Moment,
+}
+
+impl<MaxTranches: Get<TrancheId>, MaxHold, MinDelay> Default
+	for TrancheInvestors<MaxHold, MinDelay, MaxTranches>
```
vedhavyas, line 177: We dont need to MaxHold. Just Mindelay so that Investor has time to do some redeem if required. For some pools, we may have the max value of the moment

runtime/common/src/impls.rs
```diff
@@ -163,3 +166,139 @@ impl From<u128> for InstanceId {
...
+	pub fn contains(&self, tranche: TrancheId, now: Moment) -> bool {
+		if tranche >= self.max_tranches {
+			return false;
+		}
+
+		self.info
+			.iter()
+			.position(|info| {
+				info.tranche_id == tranche
+					&& info.permissioned_at.saturating_add(now) <= MaxHold::get()
```
vedhavyas, line 271: we just need to ensure if the permissioned_at is after `now`

runtime/common/src/impls.rs
```diff
@@ -163,3 +166,139 @@ impl From<u128> for InstanceId {
...
+	}
+
+	pub fn remove(&mut self, tranche: TrancheId, now: Moment) {
+		if tranche >= self.max_tranches {
+			return;
+		}
+
+		if let Some(index) = self.info.iter().position(|info| {
+			info.tranche_id == tranche
+				&& info.permissioned_at.saturating_add(now) >= MinDelay::get()
```
vedhavyas, line 283: we should not remove the role, instead we take new `permissioned_at` that is atleast `now + MaxDelay` and update it.

We can remove the expired role through purge call if we want to save storage

runtime/common/src/impls.rs
```diff
@@ -163,3 +166,139 @@ impl From<u128> for InstanceId {
...
+	}
+
+	pub fn insert(&mut self, tranche: TrancheId, now: Moment) {
+		if tranche >= self.max_tranches {
+			return;
+		}
+
+		if let Some(index) = self.info.iter().position(|info| info.tranche_id == tranche) {
+			let info = &mut self.info[index];
+			info.permissioned_at = now;
```
vedhavyas, line 296: I would suggest we take a permissioned_at value and check if the value is `now+minDelay`.
Remove and Insert here would do the same thing

runtime/development/src/lib.rs
vedhavyas: it should ideally be Council with fallback to Root.
For council, lets use same proportion we use for fast track in their respective runtimes

pallets/permissions/src/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+		/// Because this pallet emits events, it depends on the runtime's definition of an event.
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		type Location: Member + Parameter;
+
+		type Role: Member + Parameter;
+
+		type Storage: Member + Parameter + Properties<Property = Self::Role> + Default;
+
+		type Editors: Contains<(Self::AccountId, Self::Role)>;
```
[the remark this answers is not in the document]
mustermeiszer: Not sure I follow. Do you propose a new Trait? I would not use the interface of `pallet-permissions` for different use-cases. I think this one should be pretty general and only be used by admins. 

As you suggested, you wanted to be safe, that only a specific pallet can change the permissions. This will be checked now in the `Permissions` trait.

pallets/permissions/src/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(100)]
+		pub fn add_permission(
+			origin: OriginFor<T>,
+			to: T::AccountId,
+			location: T::Location,
+			role: T::Role,
+		) -> DispatchResult {
+			Self::ensure_admin(origin)?;
```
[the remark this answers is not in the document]
mustermeiszer: Editors should use the trait interface.

runtime/common/src/impls.rs
mustermeiszer: I thought we defined it as

So the check `has_permission(pool_id, who, PoolRole::TrancheInvestor(tranche_id)` must check if the whitelisting could be expired.
And the check `rm_permission(pool_id, who, PoolRole::TrancheInvestor(tranche_id)` must check if the minimum delay has been passed.

`MaxHold`represents the first one, `MinDelay` the last one

runtime/common/src/impls.rs
mustermeiszer: If we never remove a role, we don`t need to implement this function.

runtime/common/src/impls.rs
mustermeiszer: Not sure I follow. This function basically updates the `permissioned_at` to `now` in case the person already was whitelisted.

runtime/common/src/impls.rs
vedhavyas: ideally, yes, hence this will mostly be an alias for insert

pallets/permissions/src/lib.rs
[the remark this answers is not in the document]
vedhavyas: >  I would not use the interface of pallet-permissions for different use-cases. I think this one should be pretty general and only be used by admins.

I do not follow this. Editors are admins
For example as mentioned here - https://centrifuge.hackmd.io/Lga8cxI9Tia-u0ZPXrt84g#Notes
A poolAdmin and MemberlistAdmin can add new tranche investors for a given pool.
In that case, we want to ensure, they alone can edit them either through permissions extrinsics or through permissions trait used by other pallets. So that other pallets use pemission traits mostly yo check permissions rather than add new ones

pallets/permissions/src/lib.rs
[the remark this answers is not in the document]
vedhavyas: I disagree. This will make the extrinsics just used by the Democracy which is not the case 90% of the time. It is mostly, PoolAdmins and MemberlistAdmin who make these changes for a given pool. Letting council or root doing this is just a added benefit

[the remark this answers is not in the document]
vedhavyas: Else, Pool Pallet will have to add new extrinsics for pool Admin to add new admins to the pool and restricted tokens will have to implement another extrincisc to add new investors

runtime/common/src/impls.rs
vedhavyas: HasPermission should ensure the this instant a investor can invest and hold tokens. Which can be simple check of `permissioned_at > now`

RmPermission needs to ensure the new permissioned_at moment is atleast more than MinDelay so that investor has enough time to redeem or make other investing decisions

runtime/common/src/impls.rs
vedhavyas: new permissioned_at should be atleast now+min_delay. Since we have a use case where the permissioned_at could be the Max value of u64(Moment) so that they are indefinitely allow to invest

pallets/permissions/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Oh, this was misleading. With admin I meant the `AdminOrigin` of the pallet, which will be root, democracy or council. The Editors are pallets or whatever we put into Contians call. 

Hance, the pool-pallet can edit pool roles besides `TranchInvestors` and the pallet-restricted-tokens can edit only `TranchInvestors`

runtime/common/src/lib.rs
```diff
@@ -172,6 +175,11 @@ pub mod constants {
 	/// Milliseconds per day
 	pub const MILLISECS_PER_DAY: u64 = 86400000;
```
offerijns, line 176: Since you added `SECONDS_PER_DAY` etc now, should we redefine this as `SECONDS_PER_DAY * 1000`?

libs/common-types/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+			return false;
+		}
+
+		self.info
+			.iter()
+			.position(|info| info.tranche_id == tranche && info.permissioned_till >= Now::now())
+			.is_some()
+	}
+
+	pub fn remove(&mut self, tranche: TrancheId, delta: Moment) {
```
offerijns, line 207: Do we need both insert and remove? I think it could just be `set/update`, and just always do the `MinDelay` check.

libs/common-types/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+		if let Some(index) = self.info.iter().position(|info| info.tranche_id == tranche) {
+			let valid_till = &self.info[index].permissioned_till;
+			let now = Now::now();
+
+			if *valid_till <= now {
+				// The account is already invalid. Hence no more grace period
+				return;
+			} else {
+				// Ensure that permissioned_till is at least now + min_delay.
+				self.info[index].permissioned_till = self.validity(delta);
```
offerijns, line 221: Why is this argument called `delta`? If I'm not mistaken, it's the new "permissioned until timestamp". `delta` implies that it's the change in the value, as in, I would understand that to be how much we're adding/removing to the current value (i.e. if you pass the number of seconds in a day, that it moves up by one day).

offerijns: Also, if someone attempts to set it to less than `now + min_delay,` could we throw an error here instead of setting it to exactly `now + min_delay`? I think that's better to block than to implicitly have different behaviour.

runtime/common/src/lib.rs
mustermeiszer: Sounds good.

libs/common-types/src/lib.rs
mustermeiszer: mmh. Probably not. Its actually already doing what you are mentioning. Do you want this to be renamed?

libs/common-types/src/lib.rs
mustermeiszer: It is currently as you are understanding it. It moves the permission up by max(now + delta , now + min_delay). I don't have an opinion here. I can change it to be `permissioned_until_timestamp`. Should this be the default for every update (i.e. insert also). 

Erroring out upon not respecting the `MinDelay` is a bigger change. I understand your reasoning. The implicit behavior is not good. 

I will add this.

offerijns: Hm, on second thought, I think you're right that using `delta` here makes a lot of sense. The user would indeed whitelist someone "for a year", not "until day x".  Let's keep it with the delta variable 👍 

Changing it to error out would be great.

libs/common-types/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+	pub fn empty() -> Self {
+		Self::default()
+	}
+
+	pub fn is_empty(&self) -> bool {
+		self.info.is_empty()
+	}
+
+	fn validity(&self, delta: Moment) -> Moment {
+		let now = Now::now();
```
vedhavyas, line 190: This will be in Milliseconds. Use UnixTime trait and get duration and handy as_sec for seconds

libs/common-types/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+				// The account is already invalid. Hence no more grace period
+				return;
+			} else {
+				// Ensure that permissioned_till is at least now + min_delay.
+				self.info[index].permissioned_till = self.validity(delta);
+			}
+		};
+	}
+
+	pub fn insert(&mut self, tranche: TrancheId, delta: Moment) {
```
vedhavyas, line 226: since this is doing the upsert, I would lean towards removing the remove and rename this to upsert.
And error out like @offerijns suggested instead of setting the min_delay + now

libs/common-types/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+
+impl<T, Moment> Time for TimeProvider<T, Moment>
+where
+	T: Time<Moment = Moment>,
+	Moment: AtLeast32Bit + Parameter + Default + Copy,
+{
+	type Moment = Moment;
+
+	fn now() -> Self::Moment {
+		<T as Time>::now()
```
vedhavyas, line 257: This should be UnixTime instead of Time for getting seconds easily

vedhavyas (commented):
Looks great!
Amazing work! ❤️ 

Some suggestions on Time vs UnixTime traits

libs/common-types/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+// Ensure we're `no_std` when compiling for WebAssembly.
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use codec::{Decode, Encode};
+use common_traits::Properties;
+use frame_support::scale_info::build::Fields;
+use frame_support::scale_info::Path;
+use frame_support::scale_info::Type;
+use frame_support::sp_runtime::traits::Saturating;
+use frame_support::traits::{Get, Time};
```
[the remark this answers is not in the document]
NunoAlexandre: @vedhavyas for my own education, I'm curious about this suggestion. Could you explain why? :)

libs/common-types/src/lib.rs
mustermeiszer: Ah okay. Will Change that. Do we default to using UnixTime?

libs/common-types/src/lib.rs
mustermeiszer: We can not really remove the remove as the remove also checks if the permission has already passed and errors out then. 

I changed it to the following (But can of course further adapt): 

**Insert**:
* delta + now < min_delay -> Error
* delta + now > min_delay -> update current permission or insert new object if not found. 
   I Think here we could discuss to only allow the increase of permissions_till and error out otherwise.
   
**Remove**:
* permissioned_till < now -> Error out
* delta + now < min_delay -> Error out
* Update permissioned_till otherwise

mustermeiszer:
No idea why the test build fails. The node builds just fine on my machine. Awwwwwwww.

libs/common-types/src/lib.rs
vedhavyas: yeah since we only deal with seconds everywhere to be consistent across

libs/common-types/src/lib.rs
vedhavyas: Input wont be delta but rather a moment in the future. This will allow us to set what ever time we want without thinking about any time diffs.

It would look like this IMO

 ```
Insert:

provided_time < now + min_delay -> Error
provided_time >= now + min_delay -> update current permission or insert new object if not found.

Remove:
same as above
```

libs/common-types/src/lib.rs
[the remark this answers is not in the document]
vedhavyas: https://github.com/centrifuge/centrifuge-chain/pull/576#discussion_r784668945 @NunoAlexandre

libs/common-types/src/lib.rs
vedhavyas: @mustermeiszer any concerns on the above ?

vedhavyas (approved):

PR #580, opened 2022-01-10
Loan pallet fixes

vedhavyas:
This PR:
- Moves from Millisecs to seconds to be consistent with pool pallet
- Update tests to set milliseconds since timestamp pallets expects the Milliseconds in each block
- Ensure NAV is set to default when pool is initialised
- Deal with rounding error when trying to repay in the same block as borrow.
	- **I leaned towards allowing repay starting from the next block. any concerns ?**


pallets/loan/src/lib.rs
```diff
@@ -240,8 +239,8 @@ pub mod pallet {
 		/// Emits when operation is done on an inactive loan
 		ErrLoanNotActive,
 
-		/// Emits when epoch time is overflowed
-		ErrEpochTimeOverflow,
+		// Emits when borrow and repay doesn't happen in the same block
```
offerijns, line 242:
```suggestion
		// Emits when borrow and repay happens in the same block
```

mustermeiszer:
I think lending and repaying in one block should not have a real use-case besides gaming the system in a unwanted way, right? So I think leaning to delaying it to the next block seems reasonable.

pallets/loan/src/functions.rs
```diff
@@ -330,6 +330,16 @@ impl<T: Config> Pallet<T> {
...
+				let now: u64 = Self::time_now();
+
+				// ensure current time is more than origination time
+				// this is mainly to deal with how we calculate debt while trying to repay
+				// therefore we do not let users repay at same instant origination happened
+				ensure!(
+					now > loan_info.origination_date,
+					Error::<T>::ErrRepayTooEarly
+				);
+
```
mustermeiszer, line 342: Can `Self::time_now()` differ in one block?

mustermeiszer (approved):
LGTM! 

Just a question, regarding if `T::Time::now()` can differ in a block.

pallets/loan/src/functions.rs
mustermeiszer: Nevermind, its set on a per block basis.

PR #583, opened 2022-01-12
Min epoch time, max NAV age, pool updates

offerijns:
- Add a min epoch time flag
- Add a max NAV age flag
- Add method for liquidity admins to update max reserve
- Add method for pool admins to update min epoch time and max nav age (and soon, challenge time)
- Add method for pool admins to update tranche structure
- Reverse tranche order by default


pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -471,6 +490,48 @@ pub mod pallet {
 			})
 		}
 
+		#[pallet::weight(100)]
+		pub fn set_max_reserve(
```
offerijns, line 494: This is separate from `update_pool` as the required permission is different.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -443,6 +460,8 @@ pub mod pallet {
 					max_reserve,
 					available_reserve: Zero::zero(),
 					total_reserve: Zero::zero(),
+					min_epoch_time: T::DefaultMinEpochTime::get(),
```
offerijns, line 463: We could also add this as an argument. However, I think this is an area where it's good to enforce some conventions (i.e. a default min epoch time and max nav age per runtime), but still allow pool owners to change them if needed.

pallets/tinlake-investor-pool/src/lib.rs
```diff
@@ -471,6 +487,70 @@ pub mod pallet {
...
+				Self::has_role_in_pool(pool_id, PoolRole::PoolAdmin, &who),
+				Error::<T>::NoPermission
+			);
+
+			Pool::<T>::try_mutate(pool_id, |pool| -> DispatchResult {
+				let pool = pool.as_mut().ok_or(Error::<T>::NoSuchPool)?;
+
+				pool.min_epoch_time = min_epoch_time;
+				pool.max_nav_age = max_nav_age;
+				Self::deposit_event(Event::PoolUpdated(pool_id));
```
mustermeiszer, line 528: Do we want to make these Update events a little more expressive? 

Like:
`PoolUpdate(T::PoolId, Reserve, ...)` so that one can see the new values without querying the storage.

mustermeiszer (commented):
Quick first overview.

pallets/tinlake-investor-pool/src/lib.rs
offerijns: I didn't add it here because these values can be trivially queried, and @vedhavyas made the point to me recently that if this is the case, then it's probably not worth the additional storage of the event parameters.

vedhavyas: As long as client can fetch the reserve in o(1) using PoolID, I would lean towards using events hinting and client fetching the changes from state. Also makes it easy on storage

mustermeiszer: Ah yes. Good idea, when they already got the pool I'd.

offerijns:
Replaced by https://github.com/centrifuge/centrifuge-chain/pull/588

PR #584, opened 2022-01-14
Feature/pallet restricted tokens

mustermeiszer:
Just a draft. It we don`t need it, we can discard it anyways.


mustermeiszer:
closing, will create a new one

PR #587, opened 2022-01-14
Remove unnescesary unstable feature

mustermeiszer:
The tests for previously required an unstable feature which made compiling the pallets or runtimes without a nighty compiler version impossible. 

This was not needed and this PR removes the unstable feature.

### DIFF
--- a/libs/common-traits/src/lib.rs
+++ b/libs/common-traits/src/lib.rs
@@ -17,10 +17,6 @@
 //! # A common trait for centrifuge
 //!
 //! This crate provides some common traits used by centrifuge.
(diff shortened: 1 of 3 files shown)


NunoAlexandre (approved):
This PR not only remove the unnecessary unstable feature but also changes time-related stuff in the test suite. Is this intended? It's good practise to separate unrelated concerns

mustermeiszer:
Yes. The reason for why I had this feature enabled was my usage of a method call to initialize the `static now`. I realized that this is pretty unnecessary and hence changed it to this approach.

PR #592, opened 2022-01-17
Pool Metadata with BoundedVev

mustermeiszer:
Use a `BoundedVec` for the metadata of a pool instead of a simple `Vec`

### DIFF
--- a/runtime/development/src/lib.rs
+++ b/runtime/development/src/lib.rs
@@ -721,6 +721,8 @@ parameter_types! {
 	pub const DefaultChallengeTime: u64 = 2 * 60; // 2 minutes
 	pub const DefaultMaxNAVAge: u64 = 1 * 60; // 1 minute
 	pub const PoolPalletId: frame_support::PalletId = frame_support::PalletId(*b"roc/pool");
+	#[derive(scale_info::TypeInfo, Eq, PartialEq, Debug, Clone, Copy )]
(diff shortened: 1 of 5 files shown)


runtime/development/src/lib.rs
```diff
@@ -721,6 +721,8 @@ parameter_types! {
 	pub const DefaultChallengeTime: u64 = 2 * 60; // 2 minutes
 	pub const DefaultMaxNAVAge: u64 = 1 * 60; // 1 minute
 	pub const PoolPalletId: frame_support::PalletId = frame_support::PalletId(*b"roc/pool");
+	#[derive(scale_info::TypeInfo, Eq, PartialEq, Debug, Clone, Copy )]
+	pub const MaxSizeMetadata: u32 = 100;
```
branan, line 725: Would be good to have a comment about where this comes from (maximum length of an IPFS hash)

mustermeiszer: Ah, I just put in a random value here TBH. What is the maximum size of an IPFS hash?

branan: 46 bytes, apparently? Or fewer if we want it in binary instead of base58 https://docs.ipfs.io/concepts/content-addressing/#identifier-formats

offerijns (approved):

PR #593, opened 2022-01-17
Submission scoring

offerijns:


pallets/pools/src/lib.rs
```diff
@@ -907,10 +931,93 @@ pub mod pallet {
...
+					.iter()
+					.zip(epoch.tranches)
+					.zip(Self::get_tranche_weights(&pool))
+					.fold(
+						Some(Zero::zero()),
+						|sum: Option<T::Balance>,
+						 ((solution, epoch_execution), (invest_weight, redeem_weight))| {
+							sum.and_then(|sum| {
+								solution
+									.invest_fulfillment
```
offerijns, line 945: It feels like there must be a better way to do this.. But I couldn't find one.

pallets/pools/src/lib.rs
```diff
@@ -140,27 +217,197 @@ pub struct EpochDetails<BalanceRatio> {
...
+		// If there are no differences in risk buffer scores, we look at the reserve improvement score.
+		if self.reserve_improvement_score > other.reserve_improvement_score {
+			return Some(Ordering::Greater);
+		} else if self.reserve_improvement_score < other.reserve_improvement_score {
+			return Some(Ordering::Less);
+		}
+
+		Some(Ordering::Equal)
+	}
+}
```
mustermeiszer, line 381: This is not what we expect from the code. We are zipping it into: `Some(vec[...], vec[...])` in case BOTH solutions are a `MinRiskBufferViolation` and `None` if one or both of them doesn`t. The above will return `Ordering::Greater` in the second case, although, the solution with no `MinRiskBufferViolation` is actually better. 

Hence, I propose the following: 
* Check if both have a `MinRiskBufferViolated` -> Compare the two with simple `<` as rust allows to compare vectors even when wrapped in an `Option` (see reference below).
   * If both scores are equal do the same logic for `MaxReserveViolated`
* If only one of them contains a `MinRiskBufferViolated` the one that doesn`t contain one is greater.

For reference:
```rust
let vec_1 = Some(vec![1,2,3,4,5]);
    let vec_2 = Some(vec![1,3,0,0]);
    let vec_3 = None;
    let vec_4 = None;
    
    assert_eq!(vec_1 > vec_2, false);
    assert_eq!(vec_2 > vec_1, true);
    assert_eq!(vec_1 > vec_3, true);
    assert_eq!(vec_3 == vec_4, true);
    assert_eq!(vec_2 >= vec_4, true);
    
    let val_1 = Some(10u32);
    let val_2 = Some(11u32);
    let val_3 = None;
    let val_4 = None;
    
    assert_eq!(val_1 > val_2, false);
    assert_eq!(val_2 > val_1, true);
    assert_eq!(val_1 > val_3, true);
    assert_eq!(val_3 == val_4, true);
    assert_eq!(val_2 >= val_4, true);
```

pallets/pools/src/lib.rs
```diff
@@ -838,38 +1088,75 @@ pub mod pallet {
...
+						.iter()
+						.map(|_| TrancheSolution {
+							invest_fulfillment: Perquintill::zero(),
+							redeem_fulfillment: Perquintill::zero(),
+						})
+						.collect::<Vec<_>>();
+
+					let existing_state_solution =
+						Self::score_solution(&pool_id, &epoch, &no_execution_solution)?;
+					epoch.best_submission = Some(existing_state_solution);
```
mustermeiszer, line 1159: This approach won`t work with the current `PartialOrd` as this solution is a "healthy" one. In general, tranche-filfillments of 0 will result in a `SolutionHealthy` from the score solution. I guess this should be an extra `UnhealtyState` so that we can compare it with other unhealthy ones.

mustermeiszer (commented):
The current approach won`t work as intended (if I understood everything correctly).

pallets/pools/src/lib.rs
```diff
@@ -51,22 +52,48 @@ pub trait TrancheToken<T: Config> {
...
+}
+
+/// A type alias for the Tranche weight calculation
+type NumTranches = u32;
+
+impl<Balance, Rate> TrancheWeigher for Tranche<Balance, Rate>
+where
+	Balance: From<u128>,
+{
+	type Weight = (Balance, Balance);
```
branan, line 80: Should this just be `(u128, u128)`, since we're not holding any sort of Balance here?

pallets/pools/src/lib.rs
```diff
@@ -140,27 +160,56 @@ pub struct EpochDetails<BalanceRatio> {
...
-	pub remaining_invest_currency: Balance,
-	pub remaining_redeem_token: Balance,
+impl<Balance, BalanceRatio> TrancheWeigher for EpochExecutionTranche<Balance, BalanceRatio>
+where
+	Balance: From<u128>,
+{
+	type Weight = (Balance, Balance);
+	type External = NumTranches;
+
+	fn calculate_weight(&self, input: Self::External) -> Self::Weight {
```
branan, line 178: could `input` have a more semantic name?

branan:
I've given this a first pass and :see_no_evil:, but this is significantly more code than I was expecting for this. It's nowhere near this complex in the solidity impl.

I'd like to spend some more time on this today in much more in-depth review trying to understand where the complexity came from and if there's anything we can do about it.

pallets/pools/src/lib.rs
mustermeiszer: Yeah, I don't really like this neither. But i am also not a friend of raw types. But I would be fine with this optionX

pallets/pools/src/lib.rs
mustermeiszer: It basically represent external input. Maybe external or something?

mustermeiszer:
I think it would be good if you give this a dedicated go through. I and also jeroen decided to diverge from the solidity code as the solution was not really well defined with respect to healthiness.

pallets/pools/src/lib.rs
```diff
@@ -838,38 +890,75 @@ pub mod pallet {
...
 					.zip(&epoch_tranche_prices)
-					.map(|(((invest, redeem), value), price)| EpochExecutionTranche {
-						value: *value,
-						price: *price,
-						invest: *invest,
-						redeem: *redeem,
-					})
+					.zip(&seniorities)
+					.zip(&tranche_min_risk_buffs)
+					.map(
```
branan, line 918: Now that we're passing in several fields of the tranche, I'm not sure it makes sense to precollect things into vecs?. We should be able to just pass in the tranche (untested, written in github comment):


```
let epoch_tranches = orders
    .iter()
    .zip(&tranches)
    .zip(&epoch_tranche_prices)
    .map(|(i((nvest, redeem), tranche), price) {
        let supply = tranche.debt.checked_add(&tranche.reserve)?;
        Some(EpochExecutionTranche {
		supply,
		price: *price,
		invest: *invest,
		redeem: *redeem,
		seniority: tranche.seniority,
		min_risk_buffer: tranche.min_risk_buffer,
	})
    })
    .collect::<Option<Vec<_>>>()
    .ok_or(Error::<T>::Overflow)?;
```

pallets/pools/src/solution.rs
```diff
@@ -0,0 +1,649 @@
...
+		}
+	}
+}
+
+impl<Balance> PartialEq for EpochSolution<Balance>
+where
+	Balance: PartialEq,
+{
+	fn eq(&self, other: &Self) -> bool {
+		match self {
```
branan, line 108:
```
match (self, other)` {
    (EpochSolution::Healthy(s_1), EpochSolution::Healthy(s_2)) => s_1 == s_2,
    (EpochSolution::Unhealthy(s_1), EpochSolution::Unhealthy(s_2)) => s_1 == s_2,
    _ => false,
}
```

branan: Actually, I think this can just be a `#[derive]`? I don't see any special logic

pallets/pools/src/solution.rs
```diff
@@ -0,0 +1,649 @@
...
+
+#[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+pub enum UnhealthyState {
+	MaxReserveViolated,
+	MinRiskBufferViolated,
+}
+
+/// The solutions struct for epoch solution
+#[derive(Encode, Decode, Clone, Eq, RuntimeDebug, TypeInfo)]
+pub enum EpochSolution<Balance> {
```
branan, line 79: If the `Unhealthy` variant was first, we could just `#[derive(PartialOrd)]`

pallets/pools/src/solution.rs
```diff
@@ -0,0 +1,649 @@
...
+		}
+	}
+}
+
+impl<Balance> PartialOrd for EpochSolution<Balance>
+where
+	Balance: PartialOrd,
+{
+	fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+		match self {
```
branan, line 126: Since the `Healthy` variant comes first we need to be explicit here, but we could list `Unhealthy` first and just use derive: https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#derivable

If we do want to keep being explicit, it would be nice to collapse this to a single `match (self, other)`

pallets/pools/src/solution.rs
```diff
@@ -0,0 +1,649 @@
...
+
+impl<Balance> PartialOrd for HealthySolution<Balance>
+where
+	Balance: PartialOrd,
+{
+	fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+		let score_self = &self.score;
+		let score_other = &other.score;
+
+		Some(if score_self > score_other {
```
branan, line 153: Is this just `self.score.partial_cmp(&other.score)`?

pallets/pools/src/lib.rs
branan: It's `n_tranches` in the other impl, Should it be the same here?

branan:
This makes sense to me. The more explicit ordering and handling of constraint-violating solutions all seems sane, and I agree the extra code is worth it to have all of that logic laid out clearly in front of us :+1:

I think everything I've left comments on is style, not substance.

I still don't like that we use `Balance` for this and would like to see a new type, either just using u128 internally or creating a new `Score` Config parameter.

pallets/pools/src/lib.rs
mustermeiszer: I am on your side here. But I would keep this refactoring for another PR in order to keep the scope as is. I would suggest we do this when we decide on a refactor for the pool and how we handle the tranches.

pallets/pools/src/solution.rs
mustermeiszer: I will change that to be derive.

pallets/pools/src/solution.rs
mustermeiszer: I want to keep being explicit about the `PartialOrd` just for the sake of readability and understandability. But I will collapse into a single match.

pallets/pools/src/solution.rs
mustermeiszer: Probably, will change and see if tests go through.

mustermeiszer (commented):
I mostly adapt your remarks @branan. Only want to be explicit about `ParitalOrd` impl.

mustermeiszer:
> I still don't like that we use `Balance` for this and would like to see a new type, either just using u128 internally or creating a new `Score` Config parameter.

I also don`t like this one. I will go with another Config parameter that maybe is bounded to `Convert<Self::Balance>`.

vedhavyas (approved):
Overall LGTM. I wait for @branan to give a 👍🏼 considering the changes
Makes sense to move the refactor to another PR.

PR #595, opened 2022-01-18
Fix: Out of bounds error with user defined `TrancheId`

mustermeiszer:
This PR fixes https://github.com/centrifuge/centrifuge-chain-internal/issues/2. 

Furthermore, it refactors the extrinsics code a bit.

### DIFF
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -629,7 +629,6 @@ pub mod pallet {
 			})?;
 
 			Self::deposit_event(Event::InvestOrderUpdated(pool_id, who));
-
 			Ok(())
 		}
 
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -420,6 +420,10 @@ pub mod pallet {
 		InvalidTrancheSeniority,
 		/// Invalid metadata passed
 		BadMetadata,
+		/// Invalid TrancheId passed. In most cases out-of-bound index
+		InvalidTrancheId,
+		/// Indicates that the new passed order equals the old-order
+		NoNewOrder,
 	}
 
 	#[pallet::call]
@@ -969,7 +973,7 @@ pub mod pallet {
 			let mut outstanding = &mut pool
 				.tranches
 				.get_mut(tranche_id.into())
-				.ok_or(Error::<T>::InvalidData)?
+				.ok_or(Error::<T>::InvalidTrancheId)?
 				.outstanding_invest_orders;
 			let pool_account = PoolLocator { pool_id }.into_account();
 
@@ -997,7 +1001,7 @@ pub mod pallet {
 			let mut outstanding = &mut pool
 				.tranches
 				.get_mut(tranche_id.into())
-				.ok_or(Error::<T>::InvalidData)?
+				.ok_or(Error::<T>::InvalidTrancheId)?
 				.outstanding_redeem_orders;
 			let pool_account = PoolLocator { pool_id }.into_account();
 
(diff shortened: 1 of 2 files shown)


pallets/pools/src/lib.rs
```diff
@@ -160,6 +169,17 @@ type LookUpSource<T> = <<T as frame_system::Config>::Lookup as StaticLookup>::So
 // Type that indicates a point in time
 type Moment = u64;
 
+// Types to ease function signatures
+type PoolDetailsOf<T> = PoolDetails<
```
branan, line 173: Yay, I was going to make a pass at doing this in one of my PRs soon, updating all the generics everywhere was getting to be such a pain

pallets/pools/src/lib.rs
```diff
@@ -994,6 +958,97 @@ pub mod pallet {
...
+			order: &mut UserOrderOf<T>,
+			amount: T::Balance,
+			pool_id: T::PoolId,
+			tranche_id: T::TrancheId,
+		) -> DispatchResult {
+			let currency = T::TrancheToken::tranche_token(pool_id, tranche_id);
+			let mut outstanding = &mut pool
+				.tranches
+				.get_mut(tranche_id.into())
+				.ok_or(Error::<T>::InvalidData)?
```
branan, line 1000: Might be nice to put a specific error in here instead of a generic one?

pallets/pools/src/lib.rs
```diff
@@ -994,6 +958,97 @@ pub mod pallet {
...
+		}
+
+		fn update_order_amount<'a>(
+			who: &'a T::AccountId,
+			pool: &'a T::AccountId,
+			order: &mut T::Balance,
+			update_order: T::Balance,
+			pool_orders: &mut T::Balance,
+		) -> Result<(&'a T::AccountId, &'a T::AccountId, T::Balance), DispatchError> {
+			ensure!(*order != update_order, Error::<T>::InvalidData);
```
branan, line 1023: Is there a reason to fail here instead of just making this a no-op with an OK result? Either way no state got updated, but when I think about us maybe wanting to integrate pools with other logic in the future, idempotence is a nice property

branan (approved):
Couple of small things that might be worth updating, but I'm generally happy with this. I'm not gonna block it landing as-is

pallets/pools/src/lib.rs
```diff
@@ -994,6 +958,97 @@ pub mod pallet {
...
+
+			order.epoch = pool.current_epoch;
+			T::Tokens::transfer(currency, send, recv, transfer_amount)
+		}
+
+		fn update_order_amount<'a>(
+			who: &'a T::AccountId,
+			pool: &'a T::AccountId,
+			order: &mut T::Balance,
+			update_order: T::Balance,
```
offerijns, line 1020: Maybe this is slightly more explicit:

```suggestion
			old_order: &mut T::Balance,
			new_order: T::Balance,
```

pallets/pools/src/lib.rs
```diff
@@ -994,6 +958,97 @@ pub mod pallet {
...
+				&mut order.redeem,
+				amount,
+				&mut outstanding,
+			)?;
+
+			order.epoch = pool.current_epoch;
+			T::Tokens::transfer(currency, send, recv, transfer_amount)
+		}
+
+		fn update_order_amount<'a>(
```
offerijns, line 1016: What is this `'a` syntax? :)

pallets/pools/src/lib.rs
```diff
@@ -994,6 +958,97 @@ pub mod pallet {
...
+			order: &mut T::Balance,
+			update_order: T::Balance,
+			pool_orders: &mut T::Balance,
+		) -> Result<(&'a T::AccountId, &'a T::AccountId, T::Balance), DispatchError> {
+			ensure!(*order != update_order, Error::<T>::InvalidData);
+
+			Ok(if update_order > *order {
+				let transfer_amount = update_order
+					.checked_sub(order)
+					.expect("Updated order larger old order. qed.");
```
offerijns, line 1028: Is there an advantage of using expect here instead of also throwing `Error::<T>::Overflow`?

offerijns (approved):

pallets/pools/src/lib.rs
mustermeiszer: Will add.

pallets/pools/src/lib.rs
mustermeiszer: My reasoning was, that I don`t want events triggered although nothing was updated. @offerijns WDYT? I am happy with both approaches.

pallets/pools/src/lib.rs
mustermeiszer: Will change.

pallets/pools/src/lib.rs
mustermeiszer: It`s a lifetime for the references. The compiler takes this in order to verify how long the references of the returned data are valid and throws a compile error if they would be dangling pointers.

pallets/pools/src/lib.rs
mustermeiszer: It's impossible to have an overflow here as we check the bounds in the if-clause. I like to be explicit about the fact, that this is a safe way to basically `unwrap` here.

mustermeiszer: We could also not use `checked_sub` TBH. But I got used to it and it somehow triggers me in the runtime code if its not saturating or checked. Haha.

pallets/pools/src/lib.rs
mustermeiszer: I changed it to no-op. Was the previous state anyways. We can change it later on if we think the event is an issue.

mustermeiszer: It's a bit tricky to change, as we rely on the fact, that a transfer of zero is a no-op. Which it is, but It looks a bit weird. Will do anyways.

mustermeiszer: So changing this to a no-op then.

mustermeiszer: Actually, from a state perspective, this is not a no-op. As the root of the state changes. I think I will stay with erroring out here.

PR #596, opened 2022-01-19
reverse tranche order

vedhavyas:
This PR ensures that tranche order is from most junior to senior.

For easier calculations for rebalancing, price calculations etc... we reverse the order to go from most senior to junior. The reverse is applied on the results if necessary

### DIFF
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -51,7 +51,6 @@ pub trait TrancheToken<T: Config> {
 pub struct TrancheInput<Rate> {
 	pub interest_per_sec: Option<Rate>,
 	pub min_risk_buffer: Option<Perquintill>,
-	pub seniority: Option<u32>,
 }
 
 #[derive(Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug, Default, TypeInfo)]
@@ -388,7 +387,7 @@ pub mod pallet {
 		InSubmissionPeriod,
 		/// Attempted to close an epoch with an out of date NAV
 		NAVTooOld,
-		/// An arithmetic overflow occured
+		/// An arithmetic overflow occurred
 		Overflow,
 		/// A Tranche ID cannot be converted to an address
 		TrancheId,
@@ -416,8 +415,6 @@ pub mod pallet {
 		CollectRequired,
 		/// Adding & removing tranches is not supported
 		CannotAddOrRemoveTranches,
-		/// Invalid tranche seniority value
-		InvalidTrancheSeniority,
 		/// Invalid metadata passed
 		BadMetadata,
 		/// Invalid TrancheId passed. In most cases out-of-bound index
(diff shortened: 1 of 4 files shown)


mustermeiszer (approved):
LGTM!

pallets/pools/src/lib.rs
```diff
@@ -1190,9 +1180,13 @@ pub mod pallet {
 			let total_assets = epoch_nav.checked_add(&epoch_reserve).unwrap();
 			let mut remaining_assets = total_assets;
 			let pool_is_zero = total_assets == Zero::zero();
+			// we are gonna reverse the order
+			// such that prices are calculated from most senior to junior
+			// there by all the remaining assets are given to the most junior tranche
 			let junior_tranche_id = tranches.len() - 1;
 			tranches
 				.iter_mut()
+				.rev()
```
branan, line 1189: I believe we want to be enumerating before the reverse? We use that as a `tranche_id`, so if the goal is to have stable IDs for more-junior tranches this doesn't solve that

pallets/pools/src/lib.rs
```diff
@@ -444,18 +441,14 @@ pub mod pallet {
...
 				.map(|(tranche_id, tranche)| Tranche {
 					interest_per_sec: tranche.interest_per_sec.unwrap_or(One::one()),
 					min_risk_buffer: tranche.min_risk_buffer.unwrap_or(Perquintill::zero()),
-					seniority: tranche.seniority.unwrap_or(
-						n_tranches
-							.saturating_sub(1)
-							.saturating_sub(tranche_id as u32),
-					),
+					// seniority increases as index since the order is from junior to senior
+					seniority: (tranche_id as u32),
```
branan, line 451: Should still respect the manually specified seniority if it's specified?

pallets/pools/src/lib.rs
branan: This would force `junior_tranche_id` to 0, which is also great because it no longer has to be calculated :tada:

pallets/pools/src/lib.rs
offerijns: Yeah it should take `tranche.seniority` unless it's `None`.

pallets/pools/src/lib.rs
vedhavyas: yeap

pallets/pools/src/lib.rs
vedhavyas: Is seniority not incremental like the index? What do seniority here means @offerijns ?

offerijns: By default, seniority will be the tranche id, meaning it will be 0 for junior, 1 for the senior, in a 2 tranche structure. This is used for the submission scoring weights: a higher seniority tranche should have higher weights, to give the tranches preference in epoch execution. However, I purposefully already included the ability to override the seniority property, to support tranches with the same seniority. As in, it will be useful (at some point) to make it possible to have multiple tranches with the same seniority.

vedhavyas: can we bring that feature of allowing custom seniority down the line? It will be a runtime change and should not touch the storage ?

PR #597, opened 2022-01-20
use safe math while finding a valid writeoff group

vedhavyas:


pallets/loans/src/functions.rs
```diff
@@ -253,7 +254,7 @@ impl<T: Config> Pallet<T> {
 				let new_borrowed_amount = loan_info
 					.borrowed_amount
 					.checked_add(&amount)
-					.ok_or(Error::<T>::AddAmountOverflow)?;
+					.ok_or(DispatchError::Arithmetic(ArithmeticError::Overflow))?;
```
offerijns, line 257: I think it's a good idea to use the standard error types for this. Should we use this everywhere, i.e. also in the pools pallet? (doesn't need to be in this PR, we can just create a ticket for this)

offerijns (approved):
Quick turn-around, nice 🔥

pallets/loans/src/functions.rs
vedhavyas: I think we can proactively do it

branan: I didn't realize ArithmeticError existed. We should switch to it :+1:

branan, line 257: `DispatchError` implements `From<ArithmeticError>`. We should be able to just do `ok_or(ArithmeticError::Overflow)?` here and the `?` operator will handle the conversioe

PR #598, opened 2022-01-20
Fix collator migration for RTU4

mikiquantum:


runtime/altair/src/lib.rs
```diff
@@ -272,30 +133,20 @@ where
...
-			consumed +=
-				IntegrateCollatorSelection::<T>::inject_invulnerables(invulnerables.as_slice());
-			consumed +=
-				IntegrateCollatorSelection::<T>::inject_desired_candidates(DESIRED_CANDIDATES);
-			consumed += IntegrateCollatorSelection::<T>::inject_candidates(
-				candidates.as_slice(),
-				CANDIDACY_BOND.into(),
-			);
-			consumed +=
-				IntegrateCollatorSelection::<T>::inject_candidacy_bond(CANDIDACY_BOND.into());
```
mustermeiszer, line 298: We still should inject `CandidacyBond` and `DesiredCandidates`. Or why do you want remove them? Not sure, If they are set correctly if we don`t go through genesis init.

mustermeiszer (approved):
LGTM! 

But I would set `CandidacyBond`, so that it is not zero and we must not set it manually. But we can also leave it to zero.

mikiquantum:
@mustermeiszer We can set both parameters later with the council/democracy. We havent discussed them internally yet so not sure what to set there at the moment.

Issue #605, opened 2022-01-24
Remove `try-get` where possible


mustermeiszer:
Currently, some places use the `Storage::try_get().is_some()` in order to verify if a key is available or not. We should use `Storage::contains_key()` instead.

PR #606, opened 2022-01-25
Use contains_key instead of try_get().is_some()

NunoAlexandre:
closes https://github.com/centrifuge/centrifuge-chain/issues/605


mustermeiszer (approved):
Thanks for that! LGTM

PR #607, opened 2022-01-25
add benchmarks for permissions

vedhavyas:


pallets/permissions/src/lib.rs
```diff
@@ -233,10 +247,11 @@ impl<T: Config> Pallet<T> {
 					roles.rm(role).map_err(|_| Error::<T>::WrongParameters)?;
 
 					if roles.empty() {
-						Ok(Permission::<T>::remove(who, location))
+						Permission::<T>::remove(who, location);
 					} else {
-						Ok(Permission::<T>::insert(who, location, roles))
+						Permission::<T>::insert(who, location, roles);
 					}
+					Ok(())
```
mustermeiszer, line 254: Why this change?

pallets/permissions/src/mock.rs
```diff
@@ -116,11 +116,12 @@ impl Properties for Storage {
 					self.org.remove(OrgStorage::HEAD_OF_SAUBERMACHING)
 				}
 			},
-		})
+		};
+		Ok(())
```
mustermeiszer, line 120: Why this change?

pallets/permissions/src/mock.rs
```diff
@@ -116,11 +116,12 @@ impl Properties for Storage {
...
 	}
 
 	fn add(&mut self, property: Self::Property) -> Result<(), ()> {
-		Ok(match property {
```
mustermeiszer, line 123: See above.

pallets/permissions/src/lib.rs
```diff
@@ -167,28 +185,26 @@ impl<T: Config> Pallet<T> {
...
+			});
 
 			if T::Editors::contains(&(who, has_role, location, role)) {
-				return Ok(());
+				return Ok(Who::Editor);
 			}
 		}
 
-		Self::ensure_admin(origin)
+		Self::ensure_admin(origin).map(|_| Who::Root)
```
mustermeiszer, line 203: I think with the approach of the weights, we should early return here in any case. As a `Origin::signed(who)`, where `who` is not contained in the `Editors` will get the weight for the `Who::Root`, which is less and not correct. 

Or we move the admin check to the top.

mustermeiszer (approved):
LGTM! Just one thing were I think the weights are calculated incorrectly. And some minor remarks.

pallets/permissions/src/mock.rs
vedhavyas: https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg

pallets/permissions/src/mock.rs
vedhavyas: https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg

pallets/permissions/src/lib.rs
vedhavyas: rust-lang.github.io/rust-clippy/master/index.html#unit_arg

pallets/permissions/src/lib.rs
mustermeiszer: Nope, you are right. Then nevermind.

mustermeiszer: And what if you are Admin but not editor? Can a signed extrinsic pass the admin check? Yes, right? Then this edge case would get the wrong weight.

vedhavyas: I have refactored it a lil bit. take another look ?

mustermeiszer (approved):
LGTM!

PR #609, opened 2022-01-26
Add benchmarking and docs for pool pallet

branan:
* Benchmark pools pallet
* Add docs to pools extrinsics
* Add a maximum number of tranches to pools (as necessary for benchmarking)


libs/common-traits/src/lib.rs
```diff
@@ -100,9 +100,13 @@ pub trait BigEndian<T> {
 
 /// A trait that can be used to fetch the nav and update nav for a given pool
 pub trait PoolNAV<PoolId, Amount> {
+	type ClassId;
+	type Origin;
 	// nav returns the nav and the last time it was calculated
 	fn nav(pool_id: PoolId) -> Option<(Amount, u64)>;
 	fn update_nav(pool_id: PoolId) -> Result<Amount, DispatchError>;
+	fn initialise(origin: Self::Origin, pool_id: PoolId, class_id: Self::ClassId)
```
branan, line 108: I really don't like this, but right now we're in the situation where the `loans` pallet already depends on `pools`, so in order to work with the `loans` pallet from the `pools` benchmarks, we need to go through a trait to break the dependency cycle.

We could make a `NAVBenchmarking` trait that has this method and is only implemented when benchmarks are enabled, but that feels even clunkier and more error-prone.

branan:
Weights were computed on my laptop, and will need to be regenerated on reference hardware

pallets/pools/src/lib.rs
```diff
@@ -559,7 +590,15 @@ pub mod pallet {
 			Ok(())
 		}
 
-		#[pallet::weight(100)]
+		/// Update per-pool configuration settings.
+		///
+		/// This sets the minimum epoch length, epoch solution challenge
+		/// time, and maximum NAV age will be set to chain-wide
+		/// defaults.
```
offerijns, line 597:
```suggestion
		/// time, and maximum NAV age.
```

pallets/pools/src/lib.rs
```diff
@@ -815,12 +910,33 @@ pub mod pallet {
...
+		/// next epoch.
+		///
+		/// If all orders can be executed without violating any
+		/// pool constraints - which include maximum reserve and
+		/// the tranche risk buffers - the execution will also be
+		/// done. See `execute_epoch` for details on epoch
+		/// execution.
+		///
+		/// If pool constraints would be violated by executing all
+		/// orders, the pool enters a challenge period. During a
```
offerijns, line 931:
```suggestion
		/// orders, the pool enters a submission period. During a
```

pallets/pools/src/lib.rs
```diff
@@ -815,12 +910,33 @@ pub mod pallet {
...
+		/// challenge period, partial executions can be submitted
```
offerijns, line 932:
```suggestion
		/// submission period, partial executions can be submitted
```

offerijns (approved):
Looks great! 🔥

branan:
> LGTM! Nice!
> 
> I would suggest, we keep the current solution for the circular dependency and decide on another solution later on. WDYT? And creta a ticket for that.

Yeah, I think this is good enough for now (nobody is likely to misuse that `initialise` method just yet).

At some point I want to do a pass on everything so we have cleaner dependency layering (primitives -> pallets -> runtime) and we'll revisit this solution then.

PR #613, opened 2022-02-01
Integration Tests - PoC

mustermeiszer:
I am not sure if this makes sense. But I wanted to test how the token-prices behave in the full-system. 
Its pretty dirty code currently, but good enough for a PoC. 

Would appreciate an opinion regarding, if we want those kind of "integration" tests with the full runtime. @NunoAlexandre will most likely add something like this for xcm anyways.


mustermeiszer:
Closing in favor of new PR, as we already have integration-test folder for xcm and detangling this is not worth the time.

PR #614, opened 2022-02-01
Pool Refactor - Part 1 - Move tranche related structures

mustermeiszer:
I will split this refactoring into multiple smaller chunks. Although, this will duplicate a little work, I hope that the reviewing process will be better and we can follow more on the changes and discuss them. 

This part does:
* Move all tranche related structures into the tranche.rs file
* Wrap the two `Vec<Tranche>` & `Vec<EpochExecutionTranche>` structs into new types
* Implement some logic like the accumulation of supplies of all pools in methods of the new types
* Remove the role approval logic from the pool api
* Adapt tests and mocks accordingly
* Makes `Order` an `OptionQuery`

### DIFF
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -19,7 +19,7 @@ mod tranche;
 mod weights;
 
 use codec::HasCompact;
-use common_traits::{Permissions, TrancheWeigher};
+use common_traits::Permissions;
 use common_traits::{PoolInspect, PoolNAV, PoolReserve};
 use common_types::PoolRole;
 use core::{
@@ -31,7 +31,6 @@ use frame_support::transactional;
 use frame_support::{dispatch::DispatchResult, pallet_prelude::*, traits::UnixTime, BoundedVec};
 use frame_system::pallet_prelude::*;
 use scale_info::TypeInfo;
-use sp_runtime::traits::StaticLookup;
 use sp_runtime::{
 	traits::{
 		AccountIdConversion, AtLeast32BitUnsigned, CheckedAdd, CheckedMul, CheckedSub, One,
@@ -49,7 +48,7 @@ where
 {
 	pub owner: AccountId,
 	pub currency: CurrencyId,
-	pub tranches: Vec<Tranche<Balance, Rate, Weight>>, // ordered junior => senior
+	pub tranches: Tranches<Balance, Rate, Weight>, // ordered junior => senior
 	pub current_epoch: EpochId,
 	pub last_epoch_closed: Moment,
 	pub last_epoch_executed: EpochId,
@@ -95,7 +94,7 @@ pub struct EpochExecutionInfo<Balance, BalanceRatio, EpochId, Weight> {
 	nav: Balance,
 	reserve: Balance,
 	max_reserve: Balance,
-	tranches: Vec<EpochExecutionTranche<Balance, BalanceRatio, Weight>>,
+	tranches: EpochExecutionTranches<Balance, BalanceRatio, Weight>,
 	best_submission: Option<EpochSolution<Balance>>,
 	challenge_period_end: Option<Moment>,
 }
@@ -109,14 +108,8 @@ pub struct OutstandingCollections<Balance> {
 	pub remaining_redeem_token: Balance,
 }
 
-// type alias for StaticLookup source that resolves to account
-type LookUpSource<T> = <<T as frame_system::Config>::Lookup as StaticLookup>::Source;
-
 // Type that indicates a point in time
 type Moment = u64;
-
-// Type that indicates the senority of a tranche
-type Seniority = u32;
 
 // Types to ease function signatures
 type PoolDetailsOf<T> = PoolDetails<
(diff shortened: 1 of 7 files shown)


pallets/pools/src/tranche.rs
```diff
@@ -0,0 +1,386 @@
...
+					redeem_starts
+						.checked_mul(10u128.pow(tranche.seniority.saturating_add(1)).into())
+						.unwrap_or(u128::MAX)
+						.into(),
+				)
+			})
+			.collect()
+	}
+
+	pub fn risk_buffers(&self) -> Vec<Perquintill> {
```
offerijns, line 232: This should be called `min_risk_buffers`

pallets/pools/src/tranche.rs
```diff
@@ -0,0 +1,386 @@
...
+	}
+
+	pub fn redemptions(&self) -> Vec<Balance> {
+		self.0
+			.iter()
+			.map(|tranche| tranche.outstanding_redeem_orders)
+			.collect()
+	}
+
+	pub fn acc_redemptions(&self) -> Result<Balance, DispatchError> {
```
offerijns, line 198: Can we rename this to `total_outstanding_redeem_orders`?

pallets/pools/src/tranche.rs
```diff
@@ -0,0 +1,386 @@
...
+	}
+
+	pub fn investments(&self) -> Vec<Balance> {
+		self.0
+			.iter()
+			.map(|tranche| tranche.outstanding_invest_orders)
+			.collect()
+	}
+
+	pub fn acc_investments(&self) -> Result<Balance, DispatchError> {
```
offerijns, line 182: Can we rename this to `total_outstanding_invest_orders`?

pallets/pools/src/tranche.rs
```diff
@@ -0,0 +1,386 @@
...
+
+	pub fn supplies(&self) -> Result<Vec<Balance>, DispatchError> {
+		self.0
+			.iter()
+			.map(|tranche| tranche.debt.checked_add(&tranche.reserve))
+			.collect::<Option<Vec<_>>>()
+			.ok_or(ArithmeticError::Overflow.into())
+	}
+
+	pub fn acc_supply(&self) -> Result<Balance, DispatchError> {
```
offerijns, line 163: Can we rename this to `total_supply`?

pallets/pools/src/tranche.rs
```diff
@@ -0,0 +1,386 @@
...
+					.checked_add(&solution.invest_fulfillment.mul_floor(tranche.invest))
+					.and_then(|value| {
+						value.checked_sub(&solution.redeem_fulfillment.mul_floor(tranche.redeem))
+					})
+			})
+			.collect::<Option<Vec<_>>>()
+			.ok_or(ArithmeticError::Overflow.into())
+	}
+
+	pub fn acc_supply_with_fulfillment(
```
offerijns, line 292: Can we also rename the `acc_` methods here? I realize you've somewhat copied this from the previous implementation. But I don't think `acc_` is clear, I wouldn't immediately know what it means.

pallets/pools/src/tranche.rs
```diff
@@ -0,0 +1,386 @@
...
+					redeem_starts
+						.checked_mul(10u128.pow(tranche.seniority.saturating_add(1)).into())
+						.unwrap_or(u128::MAX)
+						.into(),
+				)
+			})
+			.collect()
+	}
+
+	pub fn risk_buffers(&self) -> Vec<Perquintill> {
```
offerijns, line 368: Should be called `min_risk_buffers`

offerijns (commented):
My background is in Solidity, where best practice is to write self-contained smart contracts that have everything inline, minimizing imports, and minimizing abstractions. So this is going away further from that, but when trying to understand the pre-refactoring code, I definitely feel that having one big file with little imports doesn't work very well with Rust/Substrate. Thus, adding abstractions to some degree I think makes sense.

I would say, one benefit of separating it out like this is that it makes it easier to test individual parts, so I think we should add significant unit tests to `tranche.rs`.

pallets/pools/src/lib.rs
```diff
@@ -241,13 +123,6 @@ type EpochExecutionInfoOf<T> = EpochExecutionInfo<
...
-type EpochExecutionTrancheOf<T> = EpochExecutionTranche<
-	<T as Config>::Balance,
-	<T as Config>::BalanceRatio,
-	<T as Config>::TrancheWeight,
->;
-type TrancheOf<T> =
-	Tranche<<T as Config>::Balance, <T as Config>::InterestRate, <T as Config>::TrancheWeight>;
 
 #[frame_support::pallet]
 pub mod pallet {
```
offerijns, line 128: I can't seem to comment it in the right place, but the `RoleApproved` and `RoleRevoked` events can also be removed now.

pallets/pools/src/lib.rs
```diff
@@ -1212,72 +1018,55 @@ pub mod pallet {
...
+				.contains(&UnhealthyState::MinRiskBufferViolated)
+			{
+				let new_tranche_supplies: Vec<_> = tranches.supplies_with_fulfillment(solution)?;
+				let risk_buffers =
+					Self::calculate_risk_buffers(&new_tranche_supplies, &tranches.prices())?;
+
+				// Score: 1 / (min risk buffer - risk buffer)
+				// A higher score means the distance to the min risk buffer is smaller
+				let non_junior_tranches =
+					tranches.as_tranche_slice().split_first().unwrap().1.iter();
```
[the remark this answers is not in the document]
mustermeiszer: Yes, I wanted to see in the next part of the refactor what additional methods make sense. Will add the above.

pallets/pools/src/tranche.rs
```diff
@@ -0,0 +1,386 @@
...
+
+/// Types alias for Tranches
+pub(super) type TranchesOf<T> =
+	Tranches<<T as Config>::Balance, <T as Config>::InterestRate, <T as Config>::TrancheWeight>;
+
+/// Types alias for Tranche
+pub(super) type TrancheOf<T> =
+	Tranche<<T as Config>::Balance, <T as Config>::InterestRate, <T as Config>::TrancheWeight>;
+
+/// Type that indicates the senority of a tranche
```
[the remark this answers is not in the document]
mustermeiszer: Done.

pallets/pools/src/tranche.rs
mustermeiszer: Done.

pallets/pools/src/lib.rs
mustermeiszer: Done.

pallets/pools/src/tranche.rs
mustermeiszer: Done.

pallets/pools/src/tranche.rs
mustermeiszer: What is the difference between the `outstanding_invest` of the normal `Tranche` and the `invest` of the `EpochExecutionTranche` fields? From a business logic perspective.

mustermeiszer:
As this is blocking me I am merging this without review. The refactoring branch will be reviewed anyways before being merged into parchain.

Issue #618, opened 2022-02-04
Pallet-pools/collect: Error out instead of implicit using `last_epoch_closed`


mustermeiszer:
Instead of the current behaviour, where we chose `min(order.epoch + collect_n_epoch, pool.last_epoch_executed)` we want to error out if the `order.epoch + collect_n_epoch` ranges beyond the last executed epoch.

branan:
Why error here?

In theory we'd want the `collect` extrinsic to just collect everything. We need some `n` for our weight function, so we take a parameter to bound the collection. The point of that parameter is solely to set a bound for the weight, so it's fine if it's above what is actually required. The extrinsic will refund the weight difference in the end.

mustermeiszer:
Yes, but this implicit behaviour is not good in the backend. At least from my point of view. We will have an extrinsic in storage that does collect `n` epochs and is successful and we have an event that signals `min(n, last_epoch_executed)`. The front-end will take care of bounding a user to only valid inputs, but from my point of view, the chain should only accept valid inputs and not make any implicit behaviour.

mustermeiszer:
> In theory we'd want the collect extrinsic to just collect everything. We need some n for our weight function, so we take a parameter to bound the collection. The point of that parameter is solely to set a bound for the weight, so it's fine if it's above what is actually required. The extrinsic will refund the weight difference in the end.

Although, upon second thought. If we wanna keep the existing approach I would suggest, we remove the parameter `collect_n_epochs` and use an associated type `MaxCollectEpochs` or something and use this as an upper bound what one can collect in a single collect call? WDYT? 

I thought at the beginning that we wanted to have the ability to collect especially user defined n-epochs. But if it should be an overall just collect, I think the current solution with the upper bound would be a goog middleground.

Issue #599, opened 2022-01-21
Revert to separate reserve changes calculation


offerijns:
With a low reserve, any single tranche having more redemptions than investments could cause this to underflow now and we'd flag an error. That seems more likely than the total of investments or the total of redemptions individually overflowing, and is why I precalculated those totals before.

_Originally posted by @branan in https://github.com/centrifuge/centrifuge-chain/pull/588#discussion_r786147725_

mustermeiszer:
@branan Regarding this. If we have an underflow in the `do_execute_epoch` doesn`t this mean we missed something early on? This basically means, we have an epoch execution that wipes out the pool. So Reverting to separate calculations is not really helpful here.

mustermeiszer:
If the above is the case, I would close this one here.

branan:
The underflow isn't due to actual constraint violations, but due to order of operations. Originally we did something like this:

`reserve = reserve + total_investments - total_redemptions`

Now we do this:

```
for each tranche:
    reserve = reserve + tranche_investments - tranche_redemptions
```

The issue here is that a single tranche with a large number of redemptions can cause an arithmetic underflow, even if we have the investments to balance things out in another tranche that we haven't gotten to yet.

mustermeiszer:
Ah, I see. Yes, this is not right. I changed the behaviour in the refactoring and close this here therefore.

mustermeiszer:
I will reopen so we cose this once the main PR goes in.

PR #619, opened 2022-02-04
Pool Refactor - Part 2

mustermeiszer:
I will split this refactoring into multiple smaller chunks. Although this will duplicate a little work, I hope that the reviewing process will be better and we can follow more on the changes and discuss them.

This part does:
* Remove implicit behaviour for `pool::collect` calls, where `last_epoch_executed` is used as a maximum for the n_epoch that are collected over. Instead now it errors out. #618
* Enforce sane tranche structure #631
* Wrap bookkeeping into methods or trait #633
* Unifiy interface of reserve trait #642
* Adjust calculation of total reserve #599

### DIFF
--- a/pallets/loans/src/tests.rs
+++ b/pallets/loans/src/tests.rs
@@ -939,7 +939,7 @@ macro_rules! test_pool_nav {
 					assert_err!(res, Error::<MockRuntime>::LoanCeilingReached);
 
 					// payback 50 and borrow more later
-					let repay_amount = Amount::from_inner(200 * USD);
+					let repay_amount = Amount::from_inner(50 * USD);
 					let res =
 						Loans::repay(Origin::signed(borrower), pool_id, loan_id, repay_amount);
 					assert_ok!(res);
--- a/pallets/loans/src/tests.rs
+++ b/pallets/loans/src/tests.rs
@@ -939,7 +939,7 @@ macro_rules! test_pool_nav {
 					assert_err!(res, Error::<MockRuntime>::LoanCeilingReached);
 
 					// payback 50 and borrow more later
-					let repay_amount = Amount::from_inner(50 * USD);
+					let repay_amount = Amount::from_inner(200 * USD);
 					let res =
 						Loans::repay(Origin::signed(borrower), pool_id, loan_id, repay_amount);
 					assert_ok!(res);
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -1660,7 +1660,7 @@ pub mod pallet {
 				for tranche in pool.tranches.senior_to_junior_slice_mut() {
 					tranche.accrue(now)?;
 
-					let tranche_amount = if tranche.tranche_type == TrancheType::Residual {
+					let tranche_amount = if tranche.tranche_type != TrancheType::Residual {
 						tranche.ratio.mul_ceil(amount)
 					} else {
 						remaining_amount
@@ -1672,13 +1672,18 @@ pub mod pallet {
 						tranche_amount
 					};
 
-					tranche.debt -= tranche_amount;
+					// NOTE: we ensure this is never underflowing. But better be safe than sorry.
+					tranche.debt = tranche.debt.saturating_sub(tranche_amount);
 					tranche.reserve = tranche
 						.reserve
 						.checked_add(&tranche_amount)
 						.ok_or(Error::<T>::Overflow)?;
 
-					remaining_amount -= tranche_amount;
+					// NOTE: In case there is an error in the ratios this might be critical. Hence,
+					//       we check here and error out
+					remaining_amount = remaining_amount
+						.checked_sub(&tranche_amount)
+						.ok_or(ArithmeticError::Underflow)?;
 				}
 
 				T::Tokens::transfer(pool.currency, &who, &pool_account, amount, false)?;
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -457,12 +457,7 @@ pub mod pallet {
 			// A single pool ID can only be used by one owner.
 			ensure!(!Pool::<T>::contains_key(pool_id), Error::<T>::PoolInUse);
 
-			Self::is_valid_tranche_change(
-				&Tranches {
-					tranches: Vec::new(),
-				},
-				&tranches,
-			)?;
+			Self::is_valid_tranche_change(&Tranches::new(Vec::new()), &tranches)?;
 
 			let now = Self::now();
 			let tranches = Tranches::from_input::<T::PoolId, T::TrancheId, T::TrancheToken>(
@@ -1358,7 +1353,7 @@ pub mod pallet {
 		) -> DispatchResult {
 			let mut outstanding = &mut pool
 				.tranches
-				.as_mut_tranche_slice()
+				.junior_to_senior_slice_mut()
 				.get_mut(tranche_id.into())
 				.ok_or(Error::<T>::InvalidTrancheId)?
 				.outstanding_invest_orders;
@@ -1387,7 +1382,7 @@ pub mod pallet {
 			let currency = T::TrancheToken::tranche_token(pool_id, tranche_id);
 			let mut outstanding = &mut pool
 				.tranches
-				.as_mut_tranche_slice()
+				.junior_to_senior_slice_mut()
 				.get_mut(tranche_id.into())
 				.ok_or(Error::<T>::InvalidTrancheId)?
 				.outstanding_redeem_orders;
@@ -1644,7 +1639,7 @@ pub mod pallet {
 
 			let min_risk_buffers = pool_details
 				.tranches
-				.as_tranche_slice()
+				.junior_to_senior_slice()
 				.iter()
 				.map(|tranche| tranche.min_risk_buffer)
 				.collect::<Vec<_>>();
@@ -1792,7 +1787,7 @@ pub mod pallet {
 			// Update tranche orders and add epoch solution state
 			for ((((tranche_id, tranche), solution), executed_amounts), epoch_tranche) in pool
 				.tranches
-				.as_mut_tranche_slice()
+				.junior_to_senior_slice_mut()
 				.iter_mut()
 				.enumerate()
 				.zip(solution.iter().copied())
@@ -1899,7 +1894,7 @@ pub mod pallet {
 			let junior_tranche_id = 0;
 			let tranches_senior_to_junior = pool
 				.tranches
-				.as_mut_tranche_slice()
+				.junior_to_senior_slice_mut()
 				.iter_mut()
 				.enumerate()
 				.rev();
@@ -1992,9 +1987,7 @@ pub mod pallet {
 					.ok_or(Error::<T>::Overflow)?;
 
 				let mut remaining_amount = amount;
-				let tranches_senior_to_junior =
-					pool.tranches.as_mut_tranche_slice().iter_mut().rev();
-				for tranche in tranches_senior_to_junior {
+				for tranche in pool.tranches.senior_to_junior_slice_mut() {
 					tranche.accrue(now)?;
 
 					let tranche_amount = if tranche.interest_per_sec != One::one() {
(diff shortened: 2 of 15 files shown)


mustermeiszer:
Two tests in the loans pallet are failing. The worked previously because the debt calculations in the tranches was not correct (the junior tranche did not receive any debt, did not check why though).

I would like to merge this without fixing this issue as we merge this into the refactoring branch and not parachain. 

#649 will close the bug subsequently. But this PR is already way too large and I want the bug to be closed and reviewed more cleanly.

pallets/loans/src/tests.rs
```diff
@@ -939,7 +939,7 @@ macro_rules! test_pool_nav {
 					assert_err!(res, Error::<MockRuntime>::LoanCeilingReached);
 
 					// payback 50 and borrow more later
-					let repay_amount = Amount::from_inner(50 * USD);
+					let repay_amount = Amount::from_inner(200 * USD);
```
branan, line 942: Why this change? (and it's not out of sync with the comment)

mustermeiszer: Removed it. Forget about that. Haha. I wanted to check how we handle over-repayment. This makes the tests work again. So forget about my comment below....

branan (changes requested):
First pass just reading `lib.rs` - I've still got another 1500 lines or so in `tranche.rs`, `solution.rs`, and `tests.rs` to go through. And I'm sure more will come up in `lib.rs` as I piece all the shuffling together

pallets/pools/src/lib.rs
```diff
@@ -737,16 +773,17 @@ pub mod pallet {
 			let order =
 				Order::<T>::try_get(&loc, &who).map_err(|_| Error::<T>::NoOutstandingOrder)?;
 
+			let end_epoch: T::EpochId = collect_n_epochs
+				.checked_sub(&One::one())
+				.ok_or(Error::<T>::CollectsNoEpochs)?
+				.checked_add(&order.epoch)
+				.ok_or(DispatchError::from(ArithmeticError::Overflow))?;
```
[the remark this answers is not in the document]
mustermeiszer: Nope, the method will then collect the latest epoch, the previous implicit behaviour just hid that. As mentioned in #618 I think the implicit behaviour is bad and can relate to exploits, especially when inconsistent extriniscs and events are triggered. 

This logical change is simply due to the fact, that `n-epochs` describes the number of epochs I want to collect. As one basically means, collect the current epoch, we need to subtract it.

If we want to have it as a gas-parameter we should use an associated type that defines a maxium.

pallets/pools/src/lib.rs
```diff
@@ -42,22 +42,54 @@ use sp_std::vec::Vec;
...
+		self.current_epoch += One::one();
+		self.last_epoch_closed = now;
+		// TODO: Remove and set state rather to EpochClosing or similar
+		// Set available reserve to 0 to disable originations while the epoch is closed but not executed
+		self.reserve.available_reserve = Zero::zero();
+
+		Ok(())
+	}
+
+	fn start_epoch(&mut self, _now: Moment) -> DispatchResult {
```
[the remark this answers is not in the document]
mustermeiszer: Left-over from me trying to abstract it in a trait. Will remove.

pallets/pools/src/lib.rs
```diff
@@ -42,22 +42,54 @@ use sp_std::vec::Vec;
 pub struct PoolDetails<AccountId, CurrencyId, EpochId, Balance, Rate, MetaSize, Weight>
 where
 	MetaSize: Get<u32> + Copy,
+	Rate: FixedPointNumber<Inner = Balance>,
+	Balance: FixedPointOperand,
```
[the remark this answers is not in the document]
mustermeiszer: Done.

pallets/pools/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Done

pallets/pools/src/lib.rs
```diff
@@ -42,22 +42,54 @@ use sp_std::vec::Vec;
...
+		self.current_epoch += One::one();
+		self.last_epoch_closed = now;
+		// TODO: Remove and set state rather to EpochClosing or similar
+		// Set available reserve to 0 to disable originations while the epoch is closed but not executed
+		self.reserve.available_reserve = Zero::zero();
+
+		Ok(())
+	}
+
+	fn start_epoch(&mut self) -> DispatchResult {
```
offerijns, line 86: I think this naming is confusing. The next epoch starts immediately when `current_epoch` is incremented. New orders can already come in for this new epoch while the previous epoch is still in a submission period, as well as repayments (just not originations as those are blocked by the `available_reserve` being 0). I'm not sure what the best name for this is though, the best I came come up with is something like `process_executed_epoch`.

pallets/pools/src/tranche.rs
```diff
@@ -172,31 +559,31 @@ where
...
+	pub fn oustanding_redemptions(&self) -> Vec<Balance> {
+		self.tranches
 			.iter()
 			.map(|tranche| tranche.outstanding_redeem_orders)
 			.collect()
 	}
 
-	pub fn acc_redemptions(&self) -> Result<Balance, DispatchError> {
-		self.0
+	pub fn acc_oustanding_redemptions(&self) -> Result<Balance, DispatchError> {
```
offerijns, line 585:
```suggestion
	pub fn acc_outstanding_redemptions(&self) -> Result<Balance, DispatchError> {
```

pallets/pools/src/tranche.rs
```diff
@@ -172,31 +559,31 @@ where
...
 			.iter()
 			.fold(Some(Balance::zero()), |sum, tranche| {
 				sum.and_then(|acc| acc.checked_add(&tranche.outstanding_invest_orders))
 			})
 			.ok_or(ArithmeticError::Overflow.into())
 	}
 
-	pub fn redemptions(&self) -> Vec<Balance> {
-		self.0
+	pub fn oustanding_redemptions(&self) -> Vec<Balance> {
```
offerijns, line 578:
```suggestion
	pub fn outstanding_redemptions(&self) -> Vec<Balance> {
```

pallets/pools/src/lib.rs
```diff
@@ -913,66 +909,93 @@ pub mod pallet {
 					Error::<T>::WipedOut
 				);
 
-				// Redeem orders are denominated in tranche tokens, not in the pool currency.
-				// Convert redeem orders to the pool currency and return a list of (invest, redeem) pairs.
-				let orders =
-					Self::convert_orders_to_currency(&epoch_tranche_prices, &pool.tranches)
-						.ok_or(Error::<T>::Overflow)?;
+				if pool.tranches.acc_outstanding_investments()?.is_zero()
+					&& pool.tranches.acc_oustanding_redemptions()?.is_zero()
```
offerijns, line 913:
```suggestion
					&& pool.tranches.acc_outstanding_redemptions()?.is_zero()
```

pallets/pools/src/lib.rs
```diff
@@ -913,66 +909,93 @@ pub mod pallet {
...
-				let tranche_supply = pool.tranches.supplies()?;
+					Self::deposit_event(Event::EpochExecuted(pool_id, submission_period_epoch));
 
-				let seniorities = pool
-					.tranches
-					.as_tranche_slice()
-					.iter()
-					.map(|tranche| tranche.seniority)
-					.collect::<Vec<_>>();
+					return Ok(Some(T::WeightInfo::close_epoch_no_investments(
```
offerijns, line 941: I think this name should be:

```suggestion
					return Ok(Some(T::WeightInfo::close_epoch_no_orders(
```

pallets/pools/src/lib.rs
```diff
@@ -1602,223 +1452,59 @@ pub mod pallet {
...
-			for (risk_buffer, min_risk_buffer) in risk_buffers
-				.iter()
-				.rev()
-				.zip(min_risk_buffers.iter().copied().rev())
-			{
-				if risk_buffer < &min_risk_buffer {
-					state.add_unhealthy(UnhealthyState::MinRiskBufferViolated);
-				}
+			// In case we are not setting up a new pool (i.e. a tranche setup already exists) we check
+			// wether the changes are valid with respect to the existing setup.
```
offerijns, line 1498:
```suggestion
			// whether the changes are valid with respect to the existing setup.
```

pallets/pools/src/tranche.rs
```diff
@@ -86,14 +72,52 @@ impl<PoolId, TrancheId> TrancheLocator<PoolId, TrancheId> {
...
-pub struct Tranche<Balance, Rate, Weight> {
-	pub(super) interest_per_sec: Rate,
-	pub(super) min_risk_buffer: Perquintill,
+impl<Rate> TrancheType<Rate>
+where
+	Rate: PartialOrd + PartialEq,
+{
+	/// Compares tranches with the following schema:
+	///
+	/// * (Residual, Residual) => true
```
offerijns, line 90: Maybe I misunderstand this code, but this would imply we can have multiple residual tranches. We don't support this properly right now (although we might support it in the future), as we'd need to modify the tranche debt and tranche token price calculations.

pallets/pools/src/tranche.rs
```diff
@@ -106,62 +130,425 @@ pub struct Tranche<Balance, Rate, Weight> {
...
+			let total_issuance = Tokens::total_issuance(tranche.currency);
+
+			if pool_is_zero || total_issuance == Zero::zero() {
+				Ok(One::one())
+			} else if tranche.tranche_type == TrancheType::Residual {
+				BalanceRatio::checked_from_rational(remaining_assets, total_issuance)
+					.ok_or(ArithmeticError::Overflow.into())
+			} else {
+				tranche.accrue(now)?;
+				let tranche_value = tranche.balance()?;
```
offerijns, line 455: We should probably be consistent in balance vs value naming. I'd say `value = balance * price`, so in this case `tranche.balance()` is just the `tranche_balance`.

pallets/pools/src/tranche.rs
```diff
@@ -205,10 +592,10 @@ where
 	}
 
 	pub fn calculate_weights(&self) -> Vec<(Weight, Weight)> {
-		let n_tranches: u32 = self.0.len().try_into().expect("MaxTranches is u32");
+		let n_tranches: u32 = self.tranches.len().try_into().expect("MaxTranches is u32");
 		let redeem_starts = 10u128.checked_pow(n_tranches).unwrap_or(u128::MAX);
 
-		self.0
+		self.tranches
```
offerijns, line 598: We should add a comment explaining this logic. Something like

```suggestion
		// The desired order priority is:
		// - Senior redemptions
		// - Junior redemptions
		// - Junior investments
		// - Senior investments
		// We ensure this by having a higher base weight for redemptions,
		// increasing the redemption weights by seniority,
		// and decreasing the investment weight by seniority.
		self.tranches
```

pallets/pools/src/tranche.rs
```diff
@@ -230,63 +617,357 @@ where
...
 	pub fn min_risk_buffers(&self) -> Vec<Perquintill> {
-		self.0
+		self.tranches
 			.iter()
-			.map(|tranche| tranche.min_risk_buffer)
+			.map(|tranche| tranche.min_risk_buffer())
 			.collect()
 	}
+
+	pub fn senorities(&self) -> Vec<Seniority> {
```
offerijns, line 626:
```suggestion
	pub fn seniorities(&self) -> Vec<Seniority> {
```

offerijns (approved):
Finally found some time to look at this. This is a huge improvement in the structure of the pools pallet code, huge 🔥 on getting this done @mustermeiszer! I left a few questions, and a bunch of tiny naming changes, but nothing major.

pallets/pools/src/tranche.rs
mustermeiszer: Yes, this function does support it. But one can not create a pool with multiple residual tranches see: https://github.com/centrifuge/centrifuge-chain/blob/166e155c6f502a4c1aa1de92513b8b0de2684850/pallets/pools/src/lib.rs#L1465-L1473

pallets/pools/src/tranche.rs
mustermeiszer: Done.

PR #620, opened 2022-02-04
Pool Refactoring - Main PR

### LINKED ISSUE #591: Make `Order` storage a `OptionQuery`
`Order` is of `ValueQuery`. Meaning the `try_get` will always get a

```rust
Order {
   invest: 0,
   redeem: 0,
   epoch: 0
}
``` 

Can we make `Order` storage a `OptionQuery` instead? Then this here would actually do, what I think we intend to do.

_Originally posted by @mustermeiszer in https://github.com/centrifuge/centrifuge-chain/pull/588#discussion_r785782105_

### LINKED ISSUE #599: Revert to separate reserve changes calculation

### LINKED ISSUE #618: Pallet-pools/collect: Error out instead of implicit using `last_epoch_closed`

### LINKED ISSUE #631: Enforce "sane" tranche structure
The current implementation of `is_valid_tranche_change` does not enforce a specific tranche structure.
The implementation should ensure the following two properties for a tranche defined as the following, starting from the most senior to the most junior tranche.

* Interested rate tranche A <= interested rate tranche B
* Only one junior tranche that receives all left assets

### LINKED ISSUE #633: Wrap accounting of Reserve and Tranches into Methods
Currently, we do the accounting/bookeeping manaully for tranches and and reserve structures. 
This should rather be wrapped into methods. abstracted by either a trait or manually. 

## Impl proposal
Implement the following trait for the PoolDetails itself, and then for the Reserve and the tranches. So that PoolDetails only passes down to the reserve and tranches
```rust
trait Accountant {
	fn income()

	fn expense()

	fn balance()

	fn free_balance()

	fn accrue()
	
	fn next_period()

	fn debt()
}
```

### LINKED ISSUE #642: Standardize pool interface on withdraw/deposit
Remove/rename borrow/payback

mustermeiszer:
The main PR for the pool refactoring. Tracks closing of issues.

# Closes
Closes #591
Closes #618 
Closes #631
Closes #633
Closes #642
Closes #599

### DIFF
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -5,6 +5,8 @@
 /// <https://substrate.dev/docs/en/knowledgebase/runtime/frame>
 pub use pallet::*;
 pub use solution::*;
+pub use tranche::*;
+pub use weights::*;
 
 #[cfg(feature = "runtime-benchmarks")]
 mod benchmarking;
@@ -13,23 +15,19 @@ mod mock;
 mod solution;
 #[cfg(test)]
 mod tests;
+mod tranche;
 mod weights;
-pub use weights::*;
 
 use codec::HasCompact;
-use common_traits::{Permissions, TrancheWeigher};
+use common_traits::Permissions;
 use common_traits::{PoolInspect, PoolNAV, PoolReserve};
 use common_types::PoolRole;
-use core::{
-	convert::TryFrom,
-	ops::{AddAssign, Sub},
-};
+use core::{convert::TryFrom, ops::AddAssign};
 use frame_support::traits::fungibles::{Inspect, Mutate, Transfer};
 use frame_support::transactional;
 use frame_support::{dispatch::DispatchResult, pallet_prelude::*, traits::UnixTime, BoundedVec};
 use frame_system::pallet_prelude::*;
 use scale_info::TypeInfo;
-use sp_runtime::traits::StaticLookup;
 use sp_runtime::{
 	traits::{
 		AccountIdConversion, AtLeast32BitUnsigned, CheckedAdd, CheckedMul, CheckedSub, One,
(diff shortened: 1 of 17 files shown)


mustermeiszer:
This is now rebased upon latest state of parachain. It is mostly already reviewed. 
BUT locations to check as they had conflicts: 
* lib.rs
    * do_collect()
    * do_withdraw()
    * do_borrow()
    * close_epoch()

mustermeiszer:
I would like to get this in ASAP. As otherwise rebasing will be pain again ^^

pallets/pools/src/lib.rs
```diff
@@ -1491,22 +1234,23 @@ pub mod pallet {
...
+
+			let end_epoch: T::EpochId = collect_n_epochs
+				.checked_sub(&One::one())
+				.ok_or(Error::<T>::CollectsNoEpochs)?
+				.checked_add(&order.epoch)
+				.ok_or(DispatchError::from(ArithmeticError::Overflow))?;
+
 			ensure!(
-				order.epoch <= pool.last_epoch_executed,
+				end_epoch <= pool.last_epoch_executed,
```
offerijns, line 1245: Why was this changed to `end_epoch`? I think this should remain `order.epoch`.

offerijns (approved):
One comment, otherwise LGTM!

pallets/pools/src/lib.rs
mustermeiszer: We are checking here if the last epoch a user wants to collect is already executed. Or am I missing something here?

offerijns: If they submitted an order in epoch 2, the current `last_epoch_executed` is 2, and they want to collect 1 epoch, then `end_epoch` is 3. Now, `end_epoch <= pool.last_epoch_executed` is false, but this should be possible, while `order.epoch <= pool.last_epoch_executed` is correctly true.

offerijns: Ahh, but https://github.com/centrifuge/tinlake/blob/develop/src/lender/tranche.sol#L210 meant that the Solidity implementation had the same effect, but was more implicit. I think you're right that by changing this to `end_epoch`, the code has the right behaviour.

offerijns: Took another look at this.

Say an investor submitted orders in epoch 2 and 3, and epoch 2 was executed, epoch 3 wasn't.
In this case, collect 1 epoch should work, collect 2 epochs should fail.

```
order.epoch = 2
last_epoch_executed = 2

collect_n_epochs = 1
order.epoch <= last_epoch_executed = 2 <= 2 = true
end_epoch = 1 - 1 + 2 = 2
end_epoch <= last_epoch_executed = 2 <= 2 = true

collect_n_epochs = 2
order.epoch <= last_epoch_executed = 2 <= 2 = true
end_epoch = 2 - 1 + 2 = 3
end_epoch <= last_epoch_executed = 3 <= 2 = false
```

So the new implementation using `end_epoch <= last_epoch_executed` is indeed correct. The Solidity implementation worked because `end_epoch` would be set to `last_epoch_executed` on https://github.com/centrifuge/tinlake/blob/develop/src/lender/tranche.sol#L212.

mustermeiszer:
Are we comftable with merging this one now?

mustermeiszer:
@branan If you approve this one here, feel free to hit the button.

PR #621, opened 2022-02-04
add loop for write group weights

vedhavyas:
Compute the weight incurred for looping through write off groups

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -136,6 +136,10 @@ pub mod pallet {
 		/// This is a soft limit for maximum loans we can expect in a pool.
 		/// this is mainly used to calculate estimated weight for NAV calculation.
 		type MaxLoansPerPool: Get<u64>;
+
+		/// This is a soft limit for maximum number of write off groups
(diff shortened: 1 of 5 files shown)


pallets/loans/src/lib.rs
```diff
@@ -509,19 +513,34 @@ pub mod pallet {
...
+			// total loops = index+1
+
+			// this is soft limit, so runtime do not have enough in the deposit.
+			// charge the maximum we have deposited
+			let count = index + 1;
+			if count >= T::MaxWriteOffGroups::get() {
+				return Ok(().into());
+			}
+
+			Ok(Some(Self::write_off_group_weight(count as u64)).into())
```
mustermeiszer, line 543: There is no way to determine the number of write-off-groups ahead?

mustermeiszer (approved):
LGTM! 
Just wondered if we could error out, if the write-off group is too big. But this might introduce all sorts of other problems I am not aware off. I think you thought about this already thoroughly @vedhavyas. So was just my first second idea^^.

pallets/loans/src/lib.rs
vedhavyas: not without a storage call unfortunately

PR #629, opened 2022-02-11
xcm: add xcm config, e2e tests and artefacts

NunoAlexandre:
This PR gathers the initial work for integrating XCM into our parachains.

We start by adding an XCM config to the `development` runtime and test token transfers from said parachain into a sibling parachain.

As of now, we are able to transfer both the native currency and our `Usd` token.

Some config values need to be calculated properly later on but there is no use case for that yet so I believe we can wait on those.


runtime/integration-tests/Cargo.toml
```diff
@@ -0,0 +1,89 @@
...
+
+# Polkadot dependencies
+polkadot-parachain = {git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.16" }
+polkadot-primitives = {git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.16" }
+xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.16" }
+xcm-executor = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.16" }
+xcm-builder = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.16" }
+pallet-xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.16" }
+kusama-runtime = {git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.16" }
+
```
NunoAlexandre, line 54:
```suggestion
```

scripts/init.sh
```diff
@@ -42,6 +42,35 @@ start-parachain)
     --rpc-cors all \
     --ws-external \
     --rpc-methods=Unsafe \
+    --state-cache-size 0 \
+    --log="main,debug" \
+  ;;
+
+start-parachain-sibling)
```
mikiquantum, line 49: The only reason to have this duplicated section is due to the ports, instead of duplicating I would let pass 3 ports, or even better add the para_id to the base ports. Like 30333 + 2000 or whatever paraid is passed.

mikiquantum (approved):
Overall looks good to me. Only a few comments:
- Usually we set the parameters to be consumed by a pallet configuration in the runtime, before the actual runtime. Just nitpicking on consistency.
- Since we might want to deploy various development parachains with different para_ids, might make sense to allow configuration for another para_id 3000 in the runtime config?

The e2e integration testing setup is really great, thanks for laying the ground to test more complex flows later on!

NunoAlexandre:
@mikiquantum thanks a lot for the review and for your feedback!

> Usually we set the parameters to be consumed by a pallet configuration in the runtime, before the actual runtime. Just nitpicking on consistency.

I am not totally sure what you mean here. Are you referring to the `parameter_types`, as in that should have them beforehand their usage? I am ok with doing that for consistency although I am a fan of the opposite: sorting code by the order of reference. So if the config needs a parameter type, the type comes after. That brings context to the code. But again, happy to keep it consistent!


> Since we might want to deploy various development parachains with different para_ids, might make sense to allow configuration for another para_id 3000 in the runtime config?

Yeah I am a bit puzzled by this, since the location of a token needs to be canonical. We don't have any hard use case requiring, say, transferring the development native token from parachain 3000 to 2000, as we already test the inverse. And when handling other tokens (say, kUsd), that will also not be necessary.

> The e2e integration testing setup is really great, thanks for laying the ground to test more complex flows later on!

Awesome, I am also really happy to have this in place! 🏄‍♂️

scripts/init.sh
NunoAlexandre: Great suggestion, done 👍 Thank you!

NunoAlexandre:
@mikiquantum

- [x] _Usually we set the parameters to be consumed by a pallet configuration in the runtime, before the actual runtime. Just nitpicking on consistency._

Have a look at my comment above for context. For now, I have organised the code to be sorted by order of importance and reference which - as I said - feels like a much more natural and readable way. As for consistency, we can also open a PR later making that approach consistent across the other runtimes.

- [x] Drop `start-parachain-sibling` and make `start-parachain` reusable for other para ids

- [x] Since we might want to deploy various development parachains with different para_ids, might make sense to allow configuration for another para_id 3000 in the runtime config?

Also have a look above. I will see what we can do here although it's not a blocker in my perspective 👍

PR #636, opened 2022-02-18
Delete orphan runtime/common/src/currency.rs file

NunoAlexandre:
Looks like this felt was re-introduced, probably in a rebase gone wrong.


mustermeiszer (approved):
LGTM!

PR #637, opened 2022-02-18
pallet: Nft Sales

NunoAlexandre:
``` rust
//! NFT Sales pallet
//!
//! This pallet provides a place for digital art creators and owners to offer their
//! NFTs for sale and for potential buyers to browse and buy them.
//!
//! To sell NFTs, users will call `add`. Doing so will have the NFT being transferred
//! from the seller to this pallet's account, to grant us the own
//!
//! To remove an NFT from sale and thus reclaim its ownership, sellers will call `remove`.
//!
//! To buy an NFT, users will call `buy`.
//!
```

### Scope

See [the spec](https://centrifuge.hackmd.io/_crWQv3jQjKDKLr0HXJaPA?edit)

For this first iteration, we include the core functionality + unit testing.
On a follow up iteration, we will set proper weights and benchmarking and other improvements we may find.

### DIFF
--- a/pallets/nft-marketplace/src/lib.rs
+++ b/pallets/nft-marketplace/src/lib.rs
@@ -1,0 +1,165 @@
+//! NFT Marketplace pallet
+//!
+//! This pallet provides a marketplace for digital art creators and owners
+//! to enlist their NFTs for sale and for potential buyers to browse and
+//! buy NFTs.
+//!
+//! // TODO(nuno): Explain more, including how the NFTs will be locked once set for sale.
+//!
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use frame_support::dispatch::DispatchResult;
+use frame_system::ensure_signed;
+use sp_runtime::traits::AtLeast32BitUnsigned;
+
+pub use pallet::*;
+
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+	use frame_support::pallet_prelude::*;
+	use frame_system::pallet_prelude::*;
+
+	#[pallet::pallet]
+	#[pallet::generate_store(pub (super) trait Store)]
+	#[pallet::without_storage_info]
+	pub struct Pallet<T>(_);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		/// The overarching event type.
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		/// The NFT class id
+		type ClassId: Parameter + Member + MaybeSerializeDeserialize + Copy + Default + TypeInfo;
+
+		/// The NFT instance id
+		type InstanceId: Parameter + Member + MaybeSerializeDeserialize + Copy + Default + TypeInfo;
+
+		/// The supported currencies that NFTs can be sold in
+		type CurrencyId: Parameter + Member + MaybeSerializeDeserialize + Copy + Default + TypeInfo;
+
+		/// The type for the asking price amount
+		type Balance: Parameter
+			+ Member
+			+ AtLeast32BitUnsigned
+			+ Default
+			+ Copy
+			+ MaybeSerializeDeserialize
+			+ MaxEncodedLen;
+
+		//TODO(nuno): we also need an impl of fungibles to move balances when having the buyer
+		// paying the asking price
+	}
+
+	// The genesis config type.
+	#[pallet::genesis_config]
+	pub struct GenesisConfig<T: Config> {
+		//TODO(nuno): define this type appropriately later
+		pub initial_state: Option<T::Balance>,
+	}
+
+	// The default value for the genesis config type.
+	#[cfg(feature = "std")]
+	impl<T: Config> Default for GenesisConfig<T> {
+		fn default() -> Self {
+			Self {
+				initial_state: None,
+			}
+		}
+	}
+
+	// The build of genesis for the pallet.
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
+		fn build(&self) {
+			//TODO(nuno): prefill the gallery with the initial state once that's ready.
+		}
+	}
+
+	/// The gallery of NFTs currently for sale
+	#[pallet::storage]
+	#[pallet::getter(fn get_allowlisted)]
+	pub(super) type Gallery<T: Config> = StorageDoubleMap<
+		_,
+		// The hasher for the first key
+		Blake2_128Concat,
+		// The first key, the nft class Id
+		T::ClassId,
+		// The hasher for the second key
+		Blake2_128Concat,
+		// The second key, the nft instance id
+		T::InstanceId,
+		// The asking price
+		(T::CurrencyId, T::Balance),
+	>;
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	//TODO(nuno): given these meaningful params
+	pub enum Event<T: Config> {
+		/// An NFT has been added to the gallery and is now for sale
+		ForSale,
+
+		/// An NFT has been removed from the gallery and is no longer for sale
+		Unlisted,
+
+		/// An NFT has been sold
+		Sold,
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// The origin is not the owner of an NFT
+		NotOwner,
+
+		/// A seller has attempted to list an NFT that is already for sale
+		AlreadyForSale,
+
+		/// A buyer attempted to buy an NFT that is not for sale
+		NotForSale,
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Put the given NFT for sale
+		/// Fails if
+		///   - `origin` is not the owner of the nft
+		///   - TODO(nuno)
+		#[pallet::weight(10_000_000)]
+		pub fn add(origin: OriginFor<T>) -> DispatchResult {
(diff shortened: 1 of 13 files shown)


NunoAlexandre:
@vedhavyas @offerijns 

As of now, the core features for this NFT Marketplace are implemented and tested. However, we face two edge cases that we need to handle.

## The 2 Edge cases
	
### 1. Lost permission

A seller puts an NFT for sale, which requires our pallet to be its freezer & admin.
AFTERWARDS they change the freezer / admin to another account.

This leaves us with an asset for sale that we cannot actually sell anymore

**Outcomes**:

1. A buyer attempts to purchase the NFT and we find that it is not managed by us anymore
      Remove said NFT when we come across this case

 2. No one ever tries to buy it and it lingers around forever, taking storage space pointlessly


### 2. Parallel sell

A seller puts an NFT for sale and THEN proceeds with selling it on the side.
This leaves us with an active sale initiated by an account (A) that no longer owns it and only the new owner (B) can manage.

**Outcomes**: 

1. Another account C buys the NFT from account B (which may or may have not be aware that the NFT is open for sale)

## Possible approaches


1. Lock the NFT admin & freezer, not letting the owner change those roles
 This appears to be impossible since the owner has the ultimate privilege
	- 👍  This would be the most UX-friendly approach since the NFT still shows up in the users' wallet while giving us reliable access to manage the sale


2. Have our pallet become the owner of the NFT
	- On `remove`, we check whether the user is the original owner
	- 👎  This is not a great solution UX wise since NFTs for sale won’t show up in the user’s wallet
	- 👍  Cleanest and most reliable solution since it patches all edge cases and is an incentive-by-default model to keep the storage neat

3. Hooks
	- Every n blocks, clean up the gallery, **removing** sales that:
		- Sell an NFT which we no longer manage
		- Were initiated by an account that is no longer the owner of the nft

	-  👎 
			- Seems rather expensive to transverse the whole gallery [O(n)] every n seconds
			- The outcome 1.1 can still happen

	-  👍:
			- Clean storage, no undesired scenarios like 2.1
			 	

4. Reactively detect an NFT ownership or team change and update the storage accordingly
	- I have no idea if or how this is possible in Substrate

pallets/nft-marketplace/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+		///
+		/// Fails if
+		///   - the NFT is not for sale
+		///   - `origin` is the seller of the NFT
+		///   - `origin` does not have enough balance of the currency the nft is being sold in
+		///   - transferring the asking price from the buyer to the seller fails
+		///   - transferring the nft to the buyer fails
+		#[pallet::weight(10_000_000)]
+		#[transactional]
+		pub fn buy(
```
branan, line 269: Either the (maximum) buy price needs to be included here, or listing should be updated to have some sort of unique ID.

Right now there's a "fun" scam you can pull with this pallet

1. List NFT for low price
2. Somebody guys to buy
3. Scammer removes then re-lists at a higher price before the buy transaction is included in a block
4. User gets charged a much higher price than planned

Anyone operating a full node that accepts transactions (the various providers for polkadot.js.org) as well as collators have the knowledge to do this sort of front-running.

pallets/nft-marketplace/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+
+			// Have the buyer pay the seller for the NFT
+			T::Fungibles::transfer(
+				sale.price.currency,
+				&buyer,
+				&sale.seller,
+				sale.price.amount,
+				true,
+			)
+			.map_err(|_| Error::<T>::PaymentFailed)?;
```
branan, line 293: We probably don't want to `map_err` here - it's better to expose the full detailed error of the underlying currency pallet, IMO

pallets/nft-marketplace/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+
+			// Transfer the NFT to the buyer
+			T::NonFungibles::transfer(
+				// Self::origin(),
+				&class_id.into(),
+				&instance_id.into(),
+				&buyer.clone(),
+			)
+			.map_err(|_| Error::<T>::FailedNftTransfer)?;
```
branan, line 302: As with the currency transfer, IMO it's better to let the underlying error be what we return so there's as much debugging information as possible in the block

pallets/nft-marketplace/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+			instance_id: T::InstanceId,
+		) -> DispatchResult {
+			let buyer = ensure_signed(origin.clone())?;
+			let sale = <Gallery<T>>::get(class_id, instance_id).ok_or(Error::<T>::NotForSale)?;
+
+			// Ensure that the buyer is not the seller of the NFT
+			ensure!(buyer != sale.seller, Error::<T>::IsSeller);
+
+			// Make sure the buyer can pay for the NFT
+			T::Fungibles::can_withdraw(sale.price.currency, &buyer, sale.price.amount)
```
branan, line 281: This check isn't technically necessary - the `transfer` should fail with a descriptive error. Once that detailed error is exposed instead of a generic error we don't need to do this.

I would pretty much only ever use `can_withdraw` in a non-transactional extrinsic where I want to check the balance early and skip a bunch of on-chain work that would be done before the final transfer, or something like that. Otherwise just let the transfer fail - the extra check costs more weight.

In addition, `InsufficientBalance` is pretty generic - it could be that the user has enough tokens but they're locked, for example. The underlying error of a failed transfer will include more details than we can provide by just querying if withdrawal is possible.

pallets/nft-marketplace/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+
+			// Check that the nft is not yet for sale
+			ensure!(
+				!<Gallery<T>>::contains_key(class_id, instance_id),
+				Error::<T>::AlreadyForSale
+			);
+
+			// Transfer the NFT to the parachain account
+			T::NonFungibles::transfer(&class_id.into(), &instance_id.into(), &Self::account())
+				.map_err(|_| Error::<T>::FailedNftTransfer)?;
```
branan, line 211: As in other places, it's better to just return the error from the NFT pallet rather than hiding the details.

pallets/nft-marketplace/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+		pub fn buy(
+			origin: OriginFor<T>,
+			class_id: T::ClassId,
+			instance_id: T::InstanceId,
+		) -> DispatchResult {
+			let buyer = ensure_signed(origin.clone())?;
+			let sale = <Gallery<T>>::get(class_id, instance_id).ok_or(Error::<T>::NotForSale)?;
+
+			// Ensure that the buyer is not the seller of the NFT
+			ensure!(buyer != sale.seller, Error::<T>::IsSeller);
```
branan, line 278: What's the reason for this? It's easy enough to generate a new account for self-dealing, so it seems like a bit of a meaningless check that just costs an extra branch in the common case

branan (changes requested):
The only thing in this set of comments I consider a blocker is the `buy` front-running opportunity. Everything else is perhaps more stylistic/opinion-based

pallets/nft-marketplace/src/lib.rs
branan: The best fix here is probably just to include a `max_price` in this call, and error if `max_price` is less than `sale_price`.

That way if there is a bit of a race where the seller drops the price just as someone is buying, the buy transaction will still succeed (for the lower price, even!).

If a buyer sets a higher `max_price` than asking and the seller is in a position to increase their price to match, well, at that point it's not really an attack anymore :)

pallets/nft-marketplace/src/lib.rs
NunoAlexandre: The idea here really was to avoid having a seller buy their own NFT, as that wouldn't make much sense and I'd see that as something we could prevent. But you have a point and this is also something that the UI could be user-friendly about and not necessarily something the "protocol" needs to block 👍

pallets/nft-marketplace/src/lib.rs
NunoAlexandre: Thanks for pointing this out! I oversaw how errors worked in here. I assumed the `Error` type of `DispatchResult` would be our own `Error` type and thus we needed to return one of our variants. I tried to use `thiserror` to wrap the underlying error in all of these variants but Substrate doesn't allow for that. But I now understand that we can just return the underlying error directly which is definitely much better :)

pallets/nft-marketplace/src/lib.rs
NunoAlexandre: Thank you for taking the time for educating me on this, I _really_ appreciate it 🙌

pallets/nft-marketplace/src/lib.rs
NunoAlexandre: This is so cool, thanks a lot @branan for spotting this and proposing solutions 💯  I went ahead with the max_offer solution since it requires no extra data to be stored on chain.

NunoAlexandre:
> Nothing new jumping out at me, this looks solid.

Awesome!


> It does need benchmarks still

We are adding benchmarks in a follow up PR. The idea is to get this out in the test environment ASAP so that we can start testing it from the UI/Apps side.

branan (approved):

mikiquantum (approved):
LGTM, since we are not pumping the version when deploying to development env we will need to clean the env. Please double check with Apps if there is anything in the current development env that they are using to test POCC UI. If so, might want to pump the version.

PR #639, opened 2022-02-21
feat: add runtime level pool parameter bounds

offerijns:
Also replaces custom `Overflow` error in pools pallet with standard arithmetic errors.


mustermeiszer (commented):
Its basically fine, but currently upon creation we insert defaults. Which might violate the bounds. I would propose to just use the bounds. Wdyt?

If we wanna keep both defaults and bounds I would suggest we add more explicit documentation and/or do a `min(DefaultMaxNAVAge::get(), MaxNaveAgeUpperBound())` respectively a `max(DefaultChallengeTime::get(), ChallengeTimeLowerBound::get())`

mustermeiszer (approved):
LGTM!

PR #638, opened 2022-02-20
Version in flake

asymmetric:
This PR adds a substring of the commit hash to a few places:

- output of `centrifuge-chain --version`
- docker image tag
- nix package version

The substring is formaed by the first 6 characters of the `HEAD`, or by the word `dirty`, in case there are uncommitted changes.

Examples:

```
❯ result/bin/centrifuge-chain --version
centrifuge-chain 2.0.0-dirty-x86_64-linux-gnu
```

PR #645, opened 2022-02-22
Flakes on parachain

asymmetric:
Closes #638 - rebasing that branch (which had started on top of master) on top of parachain proved too compliacted.


.github/workflows/nix-build.yml
```diff
@@ -0,0 +1,21 @@
+name: nix-build
+on:
+  pull_request:
+  push:
+    branches: [master]
```
mikiquantum, line 5: parachain for now

flake.nix
```diff
@@ -0,0 +1,115 @@
...
+      url = github:hercules-ci/gitignore.nix;
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+
+  };
+
+  outputs = inputs:
+    let
+      name = "centrifuge-chain";
+      major = "2.0.0";
```
mikiquantum, line 16: Please match the node version https://github.com/centrifuge/centrifuge-chain/blob/parachain/Cargo.toml#L3

.github/workflows/nix-build.yml
```diff
@@ -0,0 +1,21 @@
...
+          # Nix Flakes doesn't work on shallow clones
+          fetch-depth: 0
+      - uses: cachix/install-nix-action@v16
+        with:
+          install_url: https://releases.nixos.org/nix/nix-2.6.0/install
+      - uses: cachix/cachix-action@v10
+        with:
+          name: centrifuge-chain
+          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
+      - run: nix build -L
```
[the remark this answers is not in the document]
asymmetric: We can, but it's also being built (and pushed) by the `docker.yml` workflow, so we should be careful not to push from two places.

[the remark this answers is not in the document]
asymmetric: @mikiquantum I replaced the previous Docker building code with one based on Nix.

asymmetric:
Note that we don't build the WASM runtime here. Doing that would be part of a separate PR.

flake.nix
```diff
@@ -0,0 +1,124 @@
...
+
+  outputs = inputs:
+    let
+      name = "centrifuge-chain";
+      major = "0.10.9";
+      version = "${major}-${commit-substr}";
+
+      # This evaluates to the first 6 digits of the git hash of this repo's HEAD
+      # commit, or to "dirty" if there are uncommitted changes.
+      commit-substr = builtins.substring 0 6 (inputs.self.rev or "dirty");
```
mikiquantum, line 25: The git short is usually 7 digits, not sure if `builtins.substring` is upperbound inclusive or not, but would love to keep using 7.

mikiquantum: In addition, since we currently add a time component to the image (easier to filter by), can we keep the same here?

flake.nix
```diff
@@ -0,0 +1,124 @@
...
+
+      # This could be made into a list, to support multiple platforms
+      system = "x86_64-linux";
+
+      pkgs = inputs.nixpkgs.legacyPackages.${system};
+
+      nightlyRustPlatform = pkgs.makeRustPlatform {
```
mikiquantum, line 32: As discussed offline, we may need a session around using specific rust toolchains and update of dependencies. AKA, how to make a link between `nightly-2021-11-07` and the proper version of the nix dependency.

mikiquantum (commented):
Overall looks good, before merging we should understand which nightly version is used and adapt accordingly to our latest supported one `nightly-2021-11-07`

flake.nix
asymmetric: What do you mean with time component? The date the image was created? If so, that's possible, but it makes the image impure (since it depends on creation date), and therefore not 100% reproducible.

asymmetric: Changed hash length to 7.

asymmetric: I've also pinned nightly to the version 2021-11-07.

asymmetric:
@mikiquantum I've pinned nightly to the version 2021-11-07.

asymmetric:
@mikiquantum the CI error seems like something you guys should fix on your end? Maybe update the references to 0.9.16 in the `Cargo.lock` everywhere?

flake.nix
asymmetric: @mikiquantum I've added a commit that fixes the image creation date. Compare the first and the second line:

```
❯ podman images
REPOSITORY                            TAG         IMAGE ID      CREATED       SIZE
localhost/centrifugeio/go-centrifuge  2.0-dirty   c3a8eb0b74ce  3 days ago    70.5 MB
localhost/centrifugeio/go-centrifuge  2.0-5ae4cd  829162b1650d  52 years ago  70.5 MB
```

asymmetric:
@mikiquantum this is good to on my end

mikiquantum:
Closing in favor of https://github.com/centrifuge/centrifuge-chain/pull/893 Thanks buddy!

PR #647, opened 2022-02-23
Add `collect_for` extrinsic

branan:
This new extrinsic allows another account to cause a collection in a
permissionless way.

It is not benchmarked separately from `collect`


offerijns (approved):
LGTM

mustermeiszer (approved):
LGTM

PR #650, opened 2022-02-24
add local docker compose with latest by default

mikiquantum:


branan (approved):

docker-compose-local-chain.yml
```diff
@@ -0,0 +1,32 @@
...
+      - "30355:30333"
+      - "9936:9933"
+      - "9946:9944"
+    volumes:
+      - type: bind
+        source: ./res/rococo-local.json
+        target: /chainspec.json
+        read_only: true
+    command: >
+      --chain="${CC_CHAIN_ID:-development-local}"
```
NunoAlexandre, line 16: Should we merge naming for the runtime env var? Here we call it `CC_CHAIN_ID`, in the `init.sh` we call it `PARA_CHAIN_SPEC`

docker-compose-local-chain.yml
```diff
@@ -0,0 +1,32 @@
...
+      - "9946:9944"
+    volumes:
+      - type: bind
+        source: ./res/rococo-local.json
+        target: /chainspec.json
+        read_only: true
+    command: >
+      --chain="${CC_CHAIN_ID:-development-local}" 
+      --alice
+      --parachain-id="2000"
```
NunoAlexandre, line 18:
```suggestion
      --parachain-id="${CC_PARA_ID:-2000}"
```

docker-compose-local-relay.yml
```diff
@@ -14,6 +14,7 @@ services:
         read_only: true
     command: >
       --chain=/chainspec.json
+      --node-key=e5be9a5092b81bca64be81d212e7f2f9eba183bb7a90954f7b76361f6edb5c0a
```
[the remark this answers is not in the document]
mikiquantum: This are just static so the node p2p do not change between restarts. `subkey inspect //Alice` and the equivalent for `bob`

docker-compose-local-chain.yml
mikiquantum: I opted to not do that since ports would need to be updated accordingly too

docker-compose-local-relay.yml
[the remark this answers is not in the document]
NunoAlexandre: ah cool. I would leave a comment documenting that for future clarity

docker-compose-local-chain.yml
NunoAlexandre: ok 👍

NunoAlexandre (approved):
🔆

PR #652, opened 2022-02-25
nft-sales: add benchmarking

NunoAlexandre:
We add proper benchmarks to the `nft-sales` pallet.

Disclaimer: my brain twitched getting the `benchmarking.rs` to compile and be sound so an attentive look is much appreciated.


pallets/nft-sales/Cargo.toml
```diff
@@ -18,8 +18,16 @@ frame-support = { git = "https://github.com/paritytech/substrate", default-featu
 frame-system = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.16" }
 sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.16" }
 
+# Orml crates
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.16" }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.16" }
```
branan, line 23: These should probably be optional and included only when benchmarking is enabled?

pallets/nft-sales/Cargo.toml
```diff
@@ -18,8 +18,16 @@ frame-support = { git = "https://github.com/paritytech/substrate", default-featu
...
+
 # optional dependencies for benchmarking
 frame-benchmarking = { git = "https://github.com/paritytech/substrate",  default-features = false , optional = true , branch = "polkadot-v0.9.16" }
+common-types = { path = "../../libs/common-types", default-features = false }
+pallet-balances = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.16" }
+pallet-uniques = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.16" }
```
branan, line 29: Same with these

pallets/nft-sales/Cargo.toml
```diff
@@ -18,8 +18,16 @@ frame-support = { git = "https://github.com/paritytech/substrate", default-featu
...
+runtime-common = { path = "../../runtime/common", default-features = false }
```
branan, line 30: I really don't like the pattern we've gotten into of including this in pallet crates, as we may in the future end up needing to depend on arbitrary pallets in the runtime-common crate. Ideally we'd move anything that doesn't depend on pallets into a primitives crate (or I guess what we call common-types)

For now I'm not gonna ask you to change it, just something to keep in mind for future.

pallets/nft-sales/Cargo.toml
NunoAlexandre: So I'm pretty sure I tried to set these or another pallet only used for benchmarking as optional but that didn't work. Maybe I forgot something else?

pallets/nft-sales/Cargo.toml
NunoAlexandre: Absolutely, I am also pretty bummed that we are going about it this way. 

The alternative seemed to be to extend the pallet type params to include more traits to perform certain operations in the benchmarks (e.g add + Mutate for the NonFungibles type in nft-sales to be able to mint an nft), which doesn't seem optimal either.

What is your idea to tackle this?

NunoAlexandre: Maybe we could make those trait constraints only needed for benchmarking be under the benchmarks flags? That way they are not included in the main build but are available for benchmarks?

pallets/nft-sales/Cargo.toml
NunoAlexandre: Got it 🕺  thanks!

NunoAlexandre:
@branan would you like to have another look?

branan (approved):

PR #654, opened 2022-02-26
Error handling for impl PoolReserve in pallet-pools

### LINKED ISSUE #646: Error handling for impl PoolReserve  in pallet-pools
Currently, we will face an `ArithmeticError::Underflow` error in case we withdraw more from a pool than possible.

The impl of the `PoolReserve` trait should instead handle errors with something like `NotEnoughFunds`.

stanly-johnson:
resolves #646

### DIFF
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -35,7 +35,7 @@ use sp_runtime::{
 		AccountIdConversion, AtLeast32BitUnsigned, CheckedAdd, CheckedMul, CheckedSub, One,
 		Saturating, Zero,
 	},
-	FixedPointNumber, FixedPointOperand, Perquintill, TypeId,
+	FixedPointNumber, FixedPointOperand, Perquintill, TokenError, TypeId,
 };
 use sp_std::cmp::Ordering;
 use sp_std::vec::Vec;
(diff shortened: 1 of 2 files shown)


pallets/pools/src/lib.rs
```diff
@@ -521,6 +521,8 @@ pub mod pallet {
 		NoSolutionAvailable,
 		/// One of the runtime-level pool parameter bounds was violated
 		PoolParameterBoundViolated,
+		/// Not enough funds in pool
+		NotEnoughFundsInPool,
```
NunoAlexandre, line 525:
```suggestion
		InsufficientReserveFunds,
```

branan:
I wonder if it makes more sense to use one of the built-in [`TokenError`](https://docs.substrate.io/rustdocs/latest/sp_runtime/enum.TokenError.html) variants instead of creating our own error?

mustermeiszer:
@branan the `TokenError` enum is not very descriptive in my opinion. The only roughly matching variant would be `TokenError::NoFunds`, that translates into  "Funds are unavailable" as the error message. I would keep it as an PoolErorr so one knows the reason directly.

mustermeiszer (changes requested):
Thanks for the contribution @stanly-johnson! 

Referring to what @branan said, I think it makes sense to use the `TokenError::NoFunds` variant here instead of using a costume error message.

branan (approved):
Looks good to me. Once tests have run I believe we can hit the button.

Thanks for your contribution!

NunoAlexandre:
@stanly-johnson 👏

PR #655, opened 2022-02-26
Restrict pools.create() origin

### LINKED ISSUE #644: Restrict `pools.create()` origin
Calls to the `pools.create()` extrinsic should be restricted by origin, defined per runtime. E.g. in the development runtime all origins should be allowed, while in the altair/centrifuge-chain runtimes only `Root` would be allowed.

stanly-johnson:
resolves #644

### DIFF
--- a/pallets/loans/src/mock.rs
+++ b/pallets/loans/src/mock.rs
@@ -24,7 +24,7 @@ use frame_support::{
 	traits::{GenesisBuild, SortedMembers},
 	PalletId,
 };
-use frame_system::EnsureSignedBy;
+use frame_system::{EnsureSigned, EnsureSignedBy};
 use orml_traits::parameter_type_with_key;
 use pallet_pools::PoolLocator;
 use runtime_common::{
@@ -180,6 +180,7 @@ impl pallet_pools::Config for MockRuntime {
 	type MaxNAVAgeUpperBound = MaxNAVAgeUpperBound;
 	type PalletId = PoolPalletId;
 	type Permission = Permissions;
+	type PoolCreateOrigin = EnsureSigned<u64>;
 	type MaxSizeMetadata = MaxSizeMetadata;
 	type MaxTranches = MaxTranches;
 	type WeightInfo = ();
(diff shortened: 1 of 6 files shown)


NunoAlexandre (approved):
🏄‍♂️ this looks great! 

thanks a lot for contributing 🙌

branan:
Thanks for your contribution!

I think that since we want to allow `Root` to be an origin we'll need some more work here:

* `Root` is not a signed origin, so any calls to `ensure_signed` will fail with Root as the origin
* `Root` doesn't have an account ID, so trying to set the pool owner based on it also won't work.

Those points also apply to Council origins.

We could possibly store the Origin as the owner of the pool, but I think that would break with Council origins, which are different depending on the number of council members and how many voted yes. We may be OK with this and just want to support `Root` and `Signed` origins.

I think if we want to properly support Council origins as well we'll need a custom `Owner` enum and a way of converting from various Origins to that owner enum. This will need to work across pools, loans, and permissions pallets.

/cc @offerijns What are your thoughts on the product requirements here? Should we be trying to support Council?

branan:
I think all that is also complicated by the fact that we don't necessarily want Democracy to be required to make changes to all the pool settings, so maybe the real answer is to just set an Admin account as a parameter of the create call and call it a day?

offerijns:
> /cc @offerijns What are your thoughts on the product requirements here? Should we be trying to support Council?

No we don't need to support Council. We just need to support requiring democracy votes (`Root`) for the centrifuge-chain and altair runtimes, and any origin for the development runtime.

> I think all that is also complicated by the fact that we don't necessarily want Democracy to be required to make changes to all the pool settings, so maybe the real answer is to just set an Admin account as a parameter of the create call and call it a day?

This is a great point, indeed we should add a parameter `pool_admin` and set this rather than the origin.

branan:
I think then the necessary changes should be:

* Take an admin account explicitly in `create`
* Remove the `ensure_signed` in `create`, as that is replace by the `PoolCreateOrigin::ensure_origin` call
* Remove the entire `owner` field from the pool details, since we don't use it

And then if we ever decide to implement pool deletion we'll figure out the exact logic to either use `PoolCreateOrigin` or the admin account, depending on the exact rules we want to have there.

mustermeiszer (approved):
LGTM! Thanks for the contribution @stanly-johnson. If you wanna work on something else feel free to grap another issue or ping us.

PR #659, opened 2022-02-28
Upgrade to polkadot v0.9.17

NunoAlexandre:


runtime/integration-tests/src/xcm_transfers.rs
```diff
@@ -27,6 +27,7 @@ use development_runtime::{
 };
 use runtime_common::Balance;
 
+#[ignore]
```
branan, line 30: Is there a ticket opened to fix and un-ignore these tests?

branan (approved):

runtime/integration-tests/src/xcm_transfers.rs
mikiquantum: Nope, but will be fixed in the next day or so

mikiquantum:
Regarding the polkadot-launch compatibility, I haven't check there in a while. If you @branan or someone else can give it a try would be great.

mikiquantum (approved):

PR #663, opened 2022-03-01
Pool update restrictions

offerijns:
Fixes the issue where interest rate updates apply retroactively.

Also introduces a way to schedule updates. This will have two parameters:
- A minimum delay
- A requirement to only execute the update when all redemption orders are fulfilled.

The goal is that for a pool on Centrifuge Chain, these might be set to 7/14 days and the requirement for redemptions is enabled. This means that once an issuer of a pool schedules an update, this can be shown in the UI, and investors can decide to submit redemption orders if they disagree with the update, which would mean they could get out of the pool before the update is executed.


pallets/pools/src/lib.rs
```diff
@@ -77,6 +77,16 @@ pub struct Tranche<Balance, Rate, Weight> {
...
 }
 
+#[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, Default, TypeInfo)]
+pub struct PoolUpdate<Rate> {
+	pub tranches: Option<Vec<TrancheInput<Rate>>>,
+	pub min_epoch_time: Option<Moment>,
+	pub challenge_time: Option<Moment>,
+	pub max_nav_age: Option<Moment>,
+	pub min_update_delay: Option<Moment>,
+	pub require_redeem_fulfillments_before_updates: Option<bool>,
```
mustermeiszer, line 87: I would suggest we rename this to requirements and create an enum. 

```rust
enum Requirements {
   None, 
   RedeemFilfillments
}
```

pallets/pools/src/lib.rs
```diff
@@ -120,6 +130,10 @@ where
 	pub min_epoch_time: Moment,
 	pub challenge_time: Moment,
 	pub max_nav_age: Moment,
+	pub min_update_delay: Moment,
+	pub require_redeem_fulfillments_before_updates: bool,
+	pub scheduled_update: Option<PoolUpdate<Rate>>,
+	pub scheduled_update_executed_after: Option<Moment>,
```
mustermeiszer, line 136: I would prefer to have a seperate struct for this if not change the logic to have a separate storage for this. 
So wrapping the new update fields into a `struct UpdateDetails`

mustermeiszer: I think we can split this. Having the `min_update_delay` and the requirements as parts of an `UpdateDetails` struct and the scheduled infos as a separate storage entry. I don`t wanna bloat the `PoolDetails` more.

pallets/pools/src/lib.rs
```diff
@@ -619,9 +630,7 @@ pub mod pallet {
 		pub fn update(
 			origin: OriginFor<T>,
 			pool_id: T::PoolId,
-			min_epoch_time: u64,
-			challenge_time: u64,
-			max_nav_age: u64,
+			changes: PoolUpdate<T::InterestRate>,
```
mustermeiszer, line 633: Instead of having single update function, I would split it into multiple small update functions. Cheaper and easier to extend and we don`t have the changes struct.

pallets/pools/src/lib.rs
```diff
@@ -2121,6 +2171,42 @@ pub mod pallet {
 			pool.available_reserve = pool.total_reserve;
 
 			Self::rebalance_tranches(pool, &epoch, &executed_amounts)?;
+			Self::execute_scheduled_update(pool_id, pool, Some(epoch), Some(solution))?;
```
mustermeiszer, line 2174: I would not add this to the already big epoch execution, but rather enforce anybody to call an `execute_scheduled_update` extrinsic.

mustermeiszer (commented):
I think we should split the update into multiple smaller  extriniscs and move the update execution into its own extrinsic.

pallets/pools/src/lib.rs
offerijns: I think this would complicate the logic quite a lot. Would it be a `Vec` of multiple potential changes, with separate timestamps? I think the use case is that issuers won't do updates like this often (maybe a few times per year), and will likely combine multiple changes.

mustermeiszer:
@offerijns can you make this draft PR?

pallets/pools/src/lib.rs
mustermeiszer: I am not sure about this though. Anytime, we add some updatable settings to the pools, the API of this changes. Splitting it into multiple smaller update-extrinsics is relatively stable. 

But if we think, the updatable settings will be stable, I would be fine with this.

mustermeiszer: We could think abou the `Updates` storage holding a `Vec<PoolUpdates>` where `PoolUpdates` is something like:

```rust
enum PoolUpdates {
    MinEpochTime(Moment),
    ....
}
```

Or we have separate storage elements for each updatable setting in the pool.

Both solutions allow us to extend these settings without a storage migration (assuming if we only append the enum and do not remove stuff)

offerijns:
Replaced by https://github.com/centrifuge/centrifuge-chain/pull/759

PR #664, opened 2022-03-01
nft-sales: Add `NftsBySeller` lookup storage

### LINKED ISSUE #662: Allow querying NFT sale listings by account
The NFT Studio has an Account page where a user can see the NFTs that they own. When listing an NFT for sale, its ownership is transferred (to the nftSales pallet I presume). This means the NFT won't be returned anymore when querying `uniques.account()` for an address, but we'd still like to display these NFTs on the account page. The only way to get these NFTs at the moment would be to query all listings across all collections and then filter them by address, but this is a potentially inordinate amount. If possible, allow querying NFT sale listings by account, similar to `uniques.account()`.

NunoAlexandre:
Fixes #662 

Different suggestions on naming are most welcome.


branan:
I believe `Vec::push` and `Vec::retain` are O(n) (sorta). If that's correct, we need to add a new config parameter that is the maximum number of NFTs a single account can be selling. This will be used to cap the maximum weight of an extrinsic that needs to iterate over that vector. Substrate provides a [`BoundedVec`](https://crates.parity.io/frame_support/storage/bounded_vec/struct.BoundedVec.html) that we can use here to help enforce that limit.

The benchmarks will need to be updated to be worst-case as well (listing an NFT when you're already selling `MAX-1`; buying or removing when there are `MAX` NFTs for sale)

Once we've got it benchmarked with a maximum N, we may also want to make sure we can refund weight for smaller numbers of NFTs for sale. That's a bit more complex benchmarking- and extrinsic-wise and can come in a separate PR.

(*)`retain` is always O(n). `push` is whenever it reallocates. I'm unsure if substrate leaves room on the end of a vec when loaded from storage to make `push` O(1), or if it's going to force a reallocation and copy of the data.

branan:
@NunoAlexandre Definitely black magic, but given how storage maps work it makes sense we can just iterate over sub-keys of a nested map like that.

I'm happy with this approach

branan (approved):

Issue #509, opened 2021-10-22
Runtime Specific Weights for Pallets


mustermeiszer:
In order to have a reasonable bench-bot, each runtime should have its own weights for every pallet included. See https://github.com/paritytech/polkadot/tree/master/runtime/polkadot/src/weights.

## Tasks
- [ ] ~~Integrate ALL (e.g. also balances, system, etc.) pallets we ware using in a runtime into the benchmark section of our runtimes, not just our costume ones~~
- [ ] ~~Bench Bot (to have a reasonable possibility to bench all pallets for a runtime upon changes See #510~~
- [ ] Runtime specific weights folder.

mustermeiszer:
@mikiquantum and @NunoAlexandre. I think we should start working on this sooner than later! We are currently using the default weights for the pallets, but as we start to actually change some associated types with costume pallets (see collator selection) the weights of the extriniscs do not represent what we are actually doing. 

This will be a dangerous pitfall in the future.

mustermeiszer:
Relates to the following issues and if done will close them:

- #582
- #578

PR #669, opened 2022-03-03
Runtime specific weights

mustermeiszer:
This PR adds a script that runs the benchmarks for all pallets the respective runtime exposes. This happens automatically. Hence, the script can be used for automatic CI or a bot. 

NOTE: The weights must be manually used in the actually `pallet::Config` implementation for the runtime. 

NOTE: The current state of the runtimes includes only benchmarks that work. 

Closes: #509


mustermeiszer:
@mikiquantum  and @gpmayorga: Do you think the given script does work to be used for CI?

branan (approved):
This looks good to me, assuming we're happy with the script for CI.

I think we'll want to also add the parity pallets to our benchmarks at some point?

mustermeiszer:
Yes, but in order to keep the PR small I only added the working benches that we already had in the runtimes. There is a tracking issue for each runtime that tracks which pallets are already in. So we can split the work and keep PRs small.

NunoAlexandre (approved):
Sweet 👌

mustermeiszer:
@gpmayorga and @mikiquantum pinging you again for your input regarding the script ^^

mustermeiszer:
I will merge once all tests after the rebase are clean and wait for a response from @mikiquantum  and @gpmayorga afterward and adapt later on, as this blocks integration of further pallets.

PR #676, opened 2022-03-03
Adding ChainBridge Pallet to Development runtime 

mikiquantum:
Solves #677


branan:
I'm going to need to spend a bit of time untangling the relationship between bridge-mapping, nft, and registry to make sure this looks sane with the migration to `uniques`. High level the integration looks good, and removes a bunch of stuff we ought not to need

mikiquantum:
Thanks @branan, merging now

PR #682, opened 2022-03-08
fix onboarding parachain

mikiquantum:


scripts/js/onboard/index.js
```diff
@@ -0,0 +1,69 @@
+const { ApiPromise, WsProvider, Keyring } = require('@polkadot/api');
+const fs = require('fs')
+
+const run = async () => {
+  try {
+    const wsProvider = new WsProvider('ws://0.0.0.0:9944');
```
mikiquantum, line 6: I would like this to be passed as well, since we might extend this script to onboard or upgrade programatically against the development environment

scripts/init.sh
```diff
@@ -62,18 +62,21 @@ start-parachain)
...
+  wasm_location="${PWD}/${parachain}-${para_id}.wasm"
+  ./target/release/centrifuge-chain export-genesis-wasm --chain="${parachain}" > $wasm_location
 
   echo "Parachain Id:" $para_id
   echo "Genesis state:" $genesis
-  echo "WASM path:" "./target/release/wbuild/${runtime_id}-runtime/${runtime_id}_runtime.compact.wasm"
+  echo "WASM path:" "${parachain}-${para_id}.wasm"
+
+  cd scripts/js/onboard
+  yarn && yarn execute "//Alice" ${para_id} "${genesis}" $wasm_location
```
NunoAlexandre, line 76:
```suggestion
  yarn && yarn execute "//Alice" ${para_id} "${genesis}" $wasm_location
  cd ../../../
```

scripts/js/onboard/index.js
NunoAlexandre: I think this is fine but it's also very easy to just pass this from `init.sh`

NunoAlexandre (commented):
This is really nice 💯 

The only observation I have is that we don't need to place the new script within a `js` directory?

mustermeiszer:
@mikiquantum could you also adapt the default para-id in the init.sh script to our current default of `10001`?

scripts/init.sh
mikiquantum: No need, this doesn't affect the parent shell.

mikiquantum:
@NunoAlexandre my idea was that we might have other auxiliary scripts in other languages, so this makes it clear which.
@mustermeiszer we are autogenerating this value to accommodate spinning up various parachains. We have some xcm configs already that are using 2000 & 3000, I think we should stick with that, and change the default on the code.

scripts/js/onboard/index.js
mikiquantum: Yeah I am already passing a few values, why not.

NunoAlexandre (approved):
🙏

PR #683, opened 2022-03-09
Pallet-balances benchmarks and weights for all Runtimers

mustermeiszer:
Adds the pallet-balances benchmarks to all runtimes and adds the respective weights via the script. 

NOTE: The amount of changes result from the fact, that the weights are always completely newly calculated. One MUST ignore the automatically generated files in the respective weights folder of the runtimes.


branan (approved):
Straightforward. Maybe we should revert all the changed weights when adding new benchmarks in the future, instead of committing, just to keep the change set easier to review?

mustermeiszer:
Ah, yes! Good idea! 

We will have on big overwrite in the end from the cloud machine anyway.

PR #684, opened 2022-03-09
Fix runtime-integration xcm tests

NunoAlexandre:
After the update to Polkadot v0.9.17, we broke these tests.
Here, we fix that by reverting some of the changes that appeared correct when performing  the update and [use a fork](https://github.com/shaunxw/xcm-simulator/pull/25) to fix compilation errors caused by duplicated dependencies.


mikiquantum (approved):
Thanks!

PR #686, opened 2022-03-09
Cap number of roles per location in permissions pallet

branan:
Previously this was unbounded, which is a potential DOS vector against
the chain.

Rather than taking a deposit per-assigned-role, we instead set a
"sufficiently large" limit on the roles per location and expect that
the creation of that location will handle taking a deposit for the
entire thing.

In our current usage, a "location" is a pool, so this deposit needs to
be taken by the pool pallet.

This also refactors the permissions pallet slightly to avoid using a
`map_or` with a complex default. Instead we use the built-in substrate
`try_mutate` and take a more procedural code approach to the
permissions manipulation.

### DIFF
--- a/pallets/permissions/src/lib.rs
+++ b/pallets/permissions/src/lib.rs
@@ -63,6 +63,8 @@ pub mod pallet {
 
 		type AdminOrigin: EnsureOrigin<Self::Origin>;
 
+		type MaxRolesPerLocation: Get<u32>;
+
 		type WeightInfo: WeightInfo;
 	}
 
@@ -82,6 +84,10 @@ pub mod pallet {
 		T::Storage,
 	>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn permission_count)]
+	pub type PermissionCount<T: Config> = StorageMap<_, Blake2_128Concat, T::Location, u32>;
+
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
@@ -98,6 +104,7 @@ pub mod pallet {
 		NoRoles,
 		NoEditor,
 		WrongParameters,
+		TooManyRoles,
 	}
 
 	#[pallet::call]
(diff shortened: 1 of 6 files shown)


branan:
This will merge cleanly with the combined rebase+parachain branch once those have been merged (and should cherry-pick cleanly between them, if we want to retarget it)

pallets/permissions/src/lib.rs
```diff
@@ -216,51 +223,56 @@ impl<T: Config> Pallet<T> {
...
+			Permission::<T>::try_mutate(who.clone(), location.clone(), |maybe_roles| {
+				let mut roles = maybe_roles.take().unwrap_or_default();
+				if roles.exists(role.clone()) {
 					Err(Error::<T>::RoleAlreadyGiven.into())
+				} else {
+					roles
+						.add(role.clone())
+						.map_err(|_| Error::<T>::WrongParameters)?;
+					*maybe_roles = Some(roles);
+					Ok(())
```
mustermeiszer, line 242: Nice change! 

Could we add a test that checks that the `num_permissions` is correctly increased and does not exists if we error out early?

pallets/permissions/src/lib.rs
```diff
@@ -216,51 +223,56 @@ impl<T: Config> Pallet<T> {
...
+				let mut roles = maybe_roles.take().ok_or(Error::<T>::NoRoles)?;
 				if roles.exists(role.clone()) {
 					roles.rm(role).map_err(|_| Error::<T>::WrongParameters)?;
-
 					if roles.empty() {
-						Permission::<T>::remove(who, location);
+						*maybe_roles = None
 					} else {
-						Permission::<T>::insert(who, location, roles);
+						*maybe_roles = Some(roles)
```
mustermeiszer, line 268: As above. Could we add a test that checks if the `num_permissions` is decreased correctly and both counts and roles are removed from storage ?

mustermeiszer (approved):
LGTM! But could we add two or three more test checking whether altering the numbers of permissions is correctly and everything is cleaned up correctly in storage upon roles are empty and counter is zero?

branan:
@mustermeiszer I added a test to track the lifecycle of the permissions count storage through addition/removal of roles

mustermeiszer:
Thanks! @branan

mustermeiszer (approved):
LGTM!

offerijns (approved):

PR #695, opened 2022-03-15
Introduce PoolStatus

### LINKED ISSUE #632: Provide State to Pool
Currently, we keep track of the state of a pool rather implicitly. Meaning, the struct `PoolDetails` is not aware of the current state of the pool. So, when a pool is in a submission-phase, we set the available balance to zero in order to disallow withdraws.

This should be handled explicitly by a state enum or similar.

## Implementation Proposal
```rust
enum PoolState {
     Open,
     InSubmissionPhase,
     Freezed,
     Closed
}
```

### LINKED ISSUE #722: Detailed `PoolStatus` Handling
We want to support the following cases:
* Tranche token price == 0 ->  blocks:
   * investments in any tranche, 
   * redemptions in the wiped out tranches 
   * originations (withdraws from the pool acting as a reserve)
* `PoolStatus` of the pool should indicate the state is weird. Maybe even noting which tranches are currently wiped out.
* If Tranche token price != 0 again after some time -> unblock the above

### Notable areas that are affected
* Solution calculations -> wiped out tranches MUST NOT accept any currently existing orders. New orders will be blocked anyways, but existing once could go though currently
* Detecting recovery from the wiped out state

mustermeiszer:
Introduces `PoolStatus` enum to the pool. 
Closes #632.
Closes #722

Introduces changes for core logic. Especially for rebalancing: 
* Spec for rebalancing: https://centrifuge.hackmd.io/w4MrddApTQykzBA8xh18RQ?both


pallets/pools/src/lib.rs
```diff
@@ -55,6 +55,17 @@ where
...
 	pub max_nav_age: Moment,
+
+	pub status: PoolStatus,
+}
+
+#[derive(Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+pub enum PoolStatus {
+	Open,
+	InSubmissionPeriod,
+	/// Repayments possible, else not possible
```
NunoAlexandre, line 66: What does this mean?

pallets/pools/src/lib.rs
```diff
@@ -55,6 +55,17 @@ where
 	pub min_epoch_time: Moment,
 	pub challenge_time: Moment,
 	pub max_nav_age: Moment,
+
```
NunoAlexandre, line 58:
```suggestion
```

?

pallets/pools/src/lib.rs
mustermeiszer: I am not sure what else is not possible. Need to check this back with @offerijns.

offerijns: I'm not sure if this comment is referring to the `InSubmissionPeriod` or the `Frozen` state?

In case of `InSubmissionPeriod`, here just originations should be blocked, nothing else.

In case of `Frozen`, I'm assuming this is the state when 1 tranche was wiped out? If so, the spec I discussed in the past with Dennis was:
```
Once a tranche is wiped out: originations should be disallowed, repayments should be allowed, redemptions from the non wiped out tranches should be allowed, investments shouldn’t be allowed in any tranche.
```

We don't need to implement this now though.

mustermeiszer: I think besides the "per-tranche-base" actions the PR does it implement it that way.

offerijns (commented):
Finally got a chance to look at this more closely. I think the behaviour of blocking orders is not right: this means that anyone who did already submit an order, will still have the chance of getting this fulfilled. With Tinlake on Ethereum, once a pool was in a wiped out state (junior price was 0.0), then no more investment fulfillments were possible in the solutions. See https://github.com/centrifuge/tinlake/blob/develop/src/lender/coordinator.sol#L528 (it doens't allow any solutions with non-zero `supply` orders if the pool is "closing", which meant a tranche was wiped out. I think this is what we should also implement for Pools. Ideally per tranche, but can start on a pool level. 

I'm not really sure it makes sense to differentiate between `Frozen` and `Closed` state. These seem redundant to me. I think the states are:
1) `Open`
2) `InSubmissionPeriod` (no originations allowed)
3) `Closed` . This is an active action (either by the pool admin or by governance) to disable further originations and also blocks investments in any tranche.

I think we should separately store `WipedOutTranches = Vec<TrancheId>` or something like that, as in there's 1 or more tranches that are wiped out. This is automatically triggered, and blocks investments in any tranche, redemptions in the wiped out tranches, and originations. I'd store this separately as once it's recovered (the tranches are not wiped out anymore), then the pool should go back into the `Open` or `Closed` state that it was in before. If we add this as another state, we lose this information.

offerijns:
> * Pools work "normally" for all tranches but the "wiped-out" tranches in case one is wiped out?
>   * Do we have any tests or experience how recovery from wiped out looks like?

We don't have any tests for this yet, no. The basic flow would be that an asset was previously written off, but some of the underlying asset is recovered and more is repaid than what was written off. This will cause the NAV to go back up again, which will cause the token prices to increase. We definitely don't need to implement this yet though, for now I think it's fine to just say that once a token price goes to 0.0, it's wiped out. And maybe support that root can undo this.

> * Closing a pool -> Allow to redeem and deposit but anything else is blocked

I'd say in a closing state, originations and new investments are blocked.

> * Switching from `Closed` to `Open` possible?

No need to support this now, but yes I think this should be an option at some point.

mustermeiszer:
@offerijns and @branan I merged the latest parachain just now.

I think if we are fine with #781, best would be to cherry-pick the diff between this here and the #781 and use this PR overall. 
I just did not wanted to pollute the important changes in the rebalancing logic with the changes made here for the actuall status of a pool. 

Just noting my thoughts here, if I am already off, when we have agreed on #781.

mustermeiszer:
@offerijns commit 2c5e11ebfc03ccb8772e5ad4b02a4549c80c35c1 includes the changes from #781 in a slightly changed manner.

mustermeiszer:
@offerijns and @branan: This one is my "final" version of the rebalancing. I think in order to make the most sense of is to review the pool-status logic separately from the rebalancing-changes. 

### Commits relevant for rebalancing
Those should make sense when reviewed separately. Each is an incrementation on the previous
All commits follwoing
[2c5e11e](https://github.com/centrifuge/centrifuge-chain/pull/695/commits/2c5e11ebfc03ccb8772e5ad4b02a4549c80c35c1)

### Commits relevant for pool-status
All commits before [2c5e11e](https://github.com/centrifuge/centrifuge-chain/pull/695/commits/2c5e11ebfc03ccb8772e5ad4b02a4549c80c35c1)

mustermeiszer:
If we are not sure, if we wanna include this, we can just add the `loss` field to the tranche and add this later. Then we don't need a migration.

mustermeiszer:
Closing. This will need to be a new PR.

PR #697, opened 2022-03-15
Use fast storage hasher

### LINKED ISSUE #696: Use fast storage hashes when it is not a security concern

mustermeiszer:
Closes #696

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -173,7 +173,7 @@ pub mod pallet {
 		_,
 		Blake2_128Concat,
 		PoolIdOf<T>,
-		Blake2_128Concat,
+		Twox64Concat,
 		T::LoanId,
 		LoanData<T::Rate, T::Amount, AssetOf<T>>,
 		OptionQuery,
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -324,7 +324,7 @@ pub mod pallet {
 		_,
 		Blake2_128Concat,
 		TrancheLocator<T::PoolId, T::TrancheId>,
-		Blake2_128Concat,
+		Twox64Concat,
 		T::EpochId,
 		EpochDetails<T::BalanceRatio>,
 	>;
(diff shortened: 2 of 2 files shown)


mustermeiszer:
I did not change anything in the integration tests. Why does this fail now? ^^

```
error: failed to get `xcm-emulator` as a dependency of package `runtime-integration-tests v0.1.0 (/home/runner/work/centrifuge-chain/centrifuge-chain/runtime/integration-tests)`
```

@NunoAlexandre Do you know if this is the xcm-emulator from shawn from parity?

NunoAlexandre:
> I did not change anything in the integration tests. Why does this fail now? ^^
> 
> ```
> error: failed to get `xcm-emulator` as a dependency of package `runtime-integration-tests v0.1.0 (/home/runner/work/centrifuge-chain/centrifuge-chain/runtime/integration-tests)`
> ```
> 
> @NunoAlexandre Do you know if this is the xcm-emulator from shawn from parity?

Maybe because my PR of that fork was merged? I find that weird tho. In any case, we can just move forward by merging this #698

mustermeiszer:
Regarding this I was wondering if it actually makes sense to use the hashers for the sake of speed. My main concern is that we have a shorter less random hash, which for the patricia-tri will lead to more branches in the trie itself, which makes storage accesses slower. 

@branan I am not sure if my understanding of the short-and-less-random-affecting-number-of-shared-letters in a patricia-path is correct. What is your feeling for this?

mustermeiszer:
ping @branan regarding the above

branan:
I think it depends mostly on how many keys we're ultimately expecting. For something with few storage items but lots of lookups, the faster hash saving time in the runtime could be worth it compared to the node time accessing the slightly worse trie. For something with many many storage items, we really want to be sure the trie itself is as optimal as possible.

I'm curious if benchmarking this shows any big differences in the runtime performance?

mustermeiszer:
The benchmarking will only show the difference in the hashing as the trie is in-memory purely and not very broad in testing. 
Also db-write/reads are static currently. Which is bad...

mustermeiszer:
But would be interesting to see, how the hashing itself is affecting this. Maybe I will run a bench run

mustermeiszer:
I thing it is not worth the speed to actually use the xx-hash here. From my feeling storage access, i.e. keeping the trie deep and not broad will be more important. Although, I have no data nor anything else, just my feeling. But I will close this.

PR #698, opened 2022-03-15
Move back to `shaunxw/xcm-simulator`

NunoAlexandre:
My pr updating that dependency to polkadot v17 has been merged so we can
move back to use the official project.


mustermeiszer (approved):
Sweeeet!

PR #703, opened 2022-03-16
Fix infinite loop in collect when epochs were partially executed

branan:


branan:
As far as I can tell, we don't test partial epochs in the test suite right now, but we DO in benchmarks (since we need to have the collection across N epochs)

mustermeiszer (approved):
Thanks for catching this. My bad introducing this in the refactor...

PR #704, opened 2022-03-17
Support exporting state & wasm from docker image

mikiquantum:
This helps go-centrifuge testworld to run without the need of having to compile our chain.


NunoAlexandre (approved):
LGTM!

PR #707, opened 2022-03-17
Add "unique" ids to tranches

### LINKED ISSUE #681: Refactor tranche ids to be unique rather than based on order id
Current Ids are generated by the order a tranche has at its creation. 
E.g.: Tranches starting from Residual to Non-Residual once
```
Tranche A -> Id 0
Tranche B -> Id 1
Tranche C -> Id 2
...
```

This makes adding and removing tranches quite annoying. Also the `CurrencyId` of a tranche is derived from this by the type `TrancheToken(PoolId, TranchId)`. 

## Proposal 
Change tranche id`s to a unique identifier 

* A: A combination between `PoolId` and block of creation of the tranche
* B: ??? 

**NOTE:** Currently `TrancheId` is `u8`, which is nice, as it only uses little storage. Maybe a full hash would be exaggerated. We need something pseudo-random. 

## Tasks 
- [ ] Get rid of all usages in the pool where ids are assumed to be in the above-described order
- [ ] Check if `TrancheLocator` can still be used. If not, remove it
- [ ] Check need to store `TrancheId` in the tranche struct
- [ ] Check compatibility with `Permissions` we currently define in runtimes
- [ ] Check compatibility with `CurrencyId::TrancheToken(PoolId, TrancheId)`

mustermeiszer:
This PR does implement a way to actually insert, remove, append tranches arbitrarily. 

Closes #681

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6157,6 +6157,7 @@ dependencies = [
  "pallet-restricted-tokens",
  "pallet-timestamp",
  "parity-scale-codec",
+ "rand 0.8.5",
  "rev_slice",
  "runtime-common",
  "scale-info",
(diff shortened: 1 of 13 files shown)


mustermeiszer:
Closing in favor of #716

PR #712, opened 2022-03-20
Interest accrual pallet

offerijns:
Spec: https://centrifuge.hackmd.io/bN29g0qIT3qJRFLNe9ntjw?view


pallets/interest-accrual/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+					.checked_mul(&rate)
+					.and_then(|debt| Self::convert::<T::NormalizedDebt, T::Amount>(debt))
+			})
+		}
+
+		fn calculate_cumulative_rate<Rate: FixedPointNumber>(
+			interest_rate_per_sec: Rate,
+			cumulative_rate: Rate,
+			last_updated: Moment,
+		) -> Option<Rate> {
```
mustermeiszer, line 199: I would make this `DispatchResult`, we can either have `Underflows`, first substracting or `Overflows`, checked_pow and checked_mul, here. 

Hence, I would simply propagate the result to the calley.

```rust
let time_difference_secs = Self::now().checked_sub(&last_updated).ok_or(ArithemticError::Underflow)?;

checked_pow(interest_rate_per_sec, time_difference_secs as usize)
    .ok_or(ArithmeticError::Overflow)?
    .checked_mul(&cumulative_rate)
    .ok_or(ArithmeticError::Overflow.into())
```

pallets/interest-accrual/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+
+		/// converts a fixed point from A precision to B precision
+		/// we don't convert from un-signed to signed or vice-verse
+		fn convert<A: FixedPointNumber, B: FixedPointNumber>(a: A) -> Option<B> {
+			if A::SIGNED != B::SIGNED {
+				return None;
+			}
+
+			B::checked_from_rational(a.into_inner(), A::accuracy())
+		}
```
mustermeiszer, line 214: We also have this in the loans? It might makes sense to move this to commons. Or have dedicated implementation for two generics that are implementing `FixedPointNumber` and use this everywhere. Or will we have a usecase for converting signed-to-un-sigend?

pallets/interest-accrual/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+					.and_then(|rate| {
+						// Apply adjustment to debt
+						match adjustment {
+							Adjustment::Increase(amount) => debt.checked_add(&amount),
+							Adjustment::Decrease(amount) => debt.checked_sub(&amount),
+						}
+						// Calculate normalized debt = debt / cumulative_rate
+						.and_then(|debt| debt.checked_div_int(rate))
+					})
+					.ok_or(Error::<T>::DebtAdjustmentFailed)?;
```
mustermeiszer, line 177: To prevent all the nested `and_then` as a proposal, but this is more gusto of style:

```rust
Self::convert::<T::InterestRate, T::Amount>(rate.cumulative_rate)
                                          .ok_or(Error::<T>::DebtAdjustmentFailed)
					.map(|rate| {
						// Apply adjustment to debt
						match adjustment {
							Adjustment::Increase(amount) => debt.checked_add(&amount),
							Adjustment::Decrease(amount) => debt.checked_sub(&amount),
						}
					})
					.ok_or(Error::<T>::DebtAdjustmentFailed)?
                                          .checked_div_int(rate)
                                          .ok_or(Error::<T>::DebtAdjustmentFailed)?;
```

pallets/interest-accrual/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+					Rates::<T>::insert(interest_rate_per_sec, &new_rate);
+					new_rate.cumulative_rate
+				}
+				Ok(rate) => {
+					let new_cumulative_rate = Self::calculate_cumulative_rate(
+						interest_rate_per_sec,
+						rate.cumulative_rate,
+						rate.last_updated,
+					)
+					.ok_or(Error::<T>::DebtCalculationFailed)?;
```
mustermeiszer, line 143: I am missing an update of ther underlying `RateDetails` here, or in the `calculate_cumulative_rate`. I would prefer the later.

pallets/interest-accrual/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+			let new_normalized_debt =
+				Self::convert::<T::InterestRate, T::Amount>(rate.cumulative_rate)
+					.and_then(|rate| {
+						// Apply adjustment to debt
+						match adjustment {
+							Adjustment::Increase(amount) => debt.checked_add(&amount),
+							Adjustment::Decrease(amount) => debt.checked_sub(&amount),
+						}
+						// Calculate normalized debt = debt / cumulative_rate
+						.and_then(|debt| debt.checked_div_int(rate))
```
mustermeiszer, line 175: Not sure if this is a problem. But shouldn't we round here in the division?

pallets/interest-accrual/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+			T::Time::now().as_secs()
+		}
+	}
+}
+
+impl<T: Config> InterestAccrual<T::InterestRate, T::Amount> for Pallet<T> {
+	type NormalizedDebt = T::NormalizedDebt;
+	type Adjustment = Adjustment<T::Amount>;
+
+	fn get_current_debt(
```
mustermeiszer, line 226: WDYT about simply naming this `current_debt`? There are some calculations involved and get seems like a normal field fetch call. Also I am not a fan of get as a name, so I might be biased here.

mustermeiszer (commented):
This is REALLY NICE! Will simplify the pools accuracy also a ton! 

Just a first quick scan. Added a few notes and I think we are currently missing updating the `cumulative_rate` in the storage. It will always be one if I am not missing something. 

### Regarding the flow of the logic for a given loan
Would it look like the following: 
1. create loan at some point in time (no calls to this here, normalized debt = 0 in the loans storage)
2. borrow at some later point
   2.1 Call `adjust_normalized_debt `, store new normalized debt in loan pallet
3. repay at some later point
   3.1 Call `get_current_debt` in order to retrieve current debt in the loan pallet

offerijns:
> Would it look like the following:
> 
> 1. create loan at some point in time (no calls to this here, normalized debt = 0 in the loans storage)
> 2. borrow at some later point
>    2.1 Call `adjust_normalized_debt `, store new normalized debt in loan pallet
> 3. repay at some later point
>    3.1 Call `get_current_debt` in order to retrieve current debt in the loan pallet

Yes this is exactly right 👍

pallets/interest-accrual/src/lib.rs
branan: This will go away after the Amount -> Balance refactor

pallets/interest-accrual/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+						)
+						.map_err(|_| Error::<T>::DebtCalculationFailed)?;
+
+						rate.accumulated_rate = new_accumulated_rate;
+						rate.last_updated = Self::now();
+
+						rate
+					} else {
+						*rate_details = Some(RateDetails {
+							accumulated_rate: T::InterestRate::saturating_from_rational(100, 100)
```
branan, line 187: We should just use `One::one()` here

branan:
I'm leaving some notes for myself as comments on this code, feel free to ignore

branan:
Moving forward, I think we ideally only have two types: `Rate` & `Balance`, and two conversion functions: `calculate_rate`, which takes a (normalized) Rate and outputs `normalized_rate * accumulated_interest` and `calculate_balance` which takes a (normalized) Balance and outputs `normalized_balance * accumulated_interest`.

We may want to give these nicer type aliases, but we should do so in a way where the conversions are free (just using `.into()` in a way the compiler can prove it is a noop)

Generally, debt amounts will be stored as a normalized Balance (on the loans side), and token prices will be stored as a normalized Rate (on the pools side)

pallets/interest-accrual/src/lib.rs
offerijns: That would be great :)

branan:
Closing in favor of #789, which also includes the fixed-point unification

PR #714, opened 2022-03-21
altair: Update benchmark computed in the cloud

NunoAlexandre:
All the changes are automated output from running `run_benchmarks` in our cloud instance


runtime/altair/src/weights/pallet_nft_sales.rs
```diff
@@ -32,17 +32,17 @@ use sp_std::marker::PhantomData;
 pub struct SubstrateWeight<T>(PhantomData<T>);
 impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	fn add() -> Weight {
-		(32_000_000 as Weight)
+		(79_141_000 as Weight)
```
NunoAlexandre, line 35: FYI, the previous revision of this file was a copy of `pallets/nft-sales/src/weights.rs`, obtained by running the benchmarks on my machine. Thus the significant difference comparing to the value now computed in the cloud.

mikiquantum (approved):
Thanks!

Issue #718, opened 2022-03-23
Pallet-pools: Pool parameters as blocks, not as time


mustermeiszer:

mustermeiszer:
We will choose to use time everywhere and will come back later to this if needed.

PR #719, opened 2022-03-23
Fixes possible deadlock upon epoch close

mustermeiszer:
The following PR closes https://github.com/centrifuge/centrifuge-chain-internal/issues/11

NOTE: Most of the changes are tests ;-)

### DIFF
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -54,6 +54,7 @@ where
 	pub min_epoch_time: Moment,
 	pub challenge_time: Moment,
 	pub max_nav_age: Moment,
+	pub min_submission_time: Moment,
 }
 
 #[derive(Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug, TypeInfo)]
@@ -137,6 +138,7 @@ pub struct EpochExecutionInfo<Balance, BalanceRatio, EpochId, Weight> {
 	max_reserve: Balance,
 	tranches: EpochExecutionTranches<Balance, BalanceRatio, Weight>,
 	best_submission: Option<EpochSolution<Balance>>,
+	min_submission_period_end: Moment,
 	challenge_period_end: Option<Moment>,
 }
 
@@ -275,6 +277,14 @@ pub mod pallet {
 		/// Default max NAV age
 		type DefaultMaxNAVAge: Get<u64>;
 
+		/// Default min submission time.
+		///
+		/// The submission time is used
+		/// to determine after which period a
+		/// pool does accept a zero-fulfillment
+		/// solution as valid.
+		type DefaultMinSubmissionTime: Get<u64>;
+
 		/// Min epoch time lower bound
 		type MinEpochTimeLowerBound: Get<u64>;
 
@@ -283,6 +293,9 @@ pub mod pallet {
 
 		/// Max NAV age upper bound
 		type MaxNAVAgeUpperBound: Get<u64>;
+
+		/// Min submission time lower bound
+		type MinSubmissionTimeLowerBound: Get<u64>;
 
 		/// Max size of Metadata
 		type MaxSizeMetadata: Get<u32> + Copy + Member + scale_info::TypeInfo;
@@ -493,6 +506,10 @@ pub mod pallet {
 						T::DefaultMaxNAVAge::get(),
 						T::MaxNAVAgeUpperBound::get(),
 					),
+					min_submission_time: sp_std::cmp::max(
+						T::DefaultMinSubmissionTime::get(),
+						T::MinSubmissionTimeLowerBound::get(),
+					),
 					reserve: ReserveDetails {
 						max_reserve,
 						available_reserve: Zero::zero(),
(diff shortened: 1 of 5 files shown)


mustermeiszer:
There is an exiting issue currently in the codebase with how we handle bounds of pool parameters. I did not wanted to include this here. This issue here https://github.com/centrifuge/centrifuge-chain/issues/718 describes the needed changes and we should address this before altair launch.

offerijns:
I'm not really sure I follow why we need the min submission time. Why can't we just immediately accept a no execution solution when the epoch goes into a submission period? This is how it also worked with Tinlake on Ethereum: https://github.com/centrifuge/tinlake/blob/develop/src/lender/test/coordinator-submitEpoch.sol#L250

mustermeiszer:
@offerijns WDYT? Am I missing something. We could also allow to always check for solutions to be `new_solution >= prev_solution`. This would mean, that one can always start the `challenge_period` by submitting a zero-fulfillment-solution. 

My feeling is, that this should be okayish, as the only attack vector would be to prevent/make-it-harder-for fulfillments.

offerijns:
We could also allow to always check for solutions to be `new_solution >= prev_solution`. This would mean, that one can always start the `challenge_period` by submitting a zero-fulfillment-solution.
> 
> My feeling is, that this should be okayish, as the only attack vector would be to prevent/make-it-harder-for fulfillments.

I'll try to take another more detailed look at this tomorrow, but I think this is indeed how it worked with Tinlake on Ethereum, because of the `>=` here: https://github.com/centrifuge/tinlake/blob/develop/src/lender/coordinator.sol#L416

branan:
I think the `>=` implementation is cleaner, but I don't have _super_ strong feelings

mustermeiszer:
I would be fine with both. We just must agree upon that we allow zero-solutions to trigger the challenge time. Wdyt? Can we agree on this here? Then I can close this next week.

offerijns:
> I would be fine with both. We just must agree upon that we allow zero-solutions to trigger the challenge time. Wdyt? Can 
we agree on this here? Then I can close this next week.

Thought about this some more, I think allowing zero solutions to solve this deadlock makes sense. The goal of the challenge period is to prevent suboptimal solutions from being accepted, by giving interested actors a reasonable chance to submit a better solution. As long as the challenge period is sufficiently long, then a zero-solution does not change this, anyone will have time to submit a better solution.

mustermeiszer:
@branan, @offerijns I adapted the PR to the above. If you give green lights for this I will merge.

pallets/pools/src/lib.rs
```diff
@@ -1027,7 +1027,7 @@ pub mod pallet {
 				let new_solution = Self::score_solution(&pool, &epoch, &solution)?;
 				if let Some(ref previous_solution) = epoch.best_submission {
 					ensure!(
-						&new_solution > previous_solution,
+						&new_solution >= previous_solution,
```
offerijns, line 1030: I wonder about one thing: if there are multiple solutions with the same score for an epoch, this makes it possible to keep overriding the best solution. I'm not sure if that's an issue. I don't think so, but maybe worth a little bit of thought.

A potential solution would be to only allow `>=` once.

pallets/pools/src/tests.rs
```diff
@@ -803,24 +803,21 @@ fn submission_period() {
...
-					},
-					TrancheSolution {
-						invest_fulfillment: Perquintill::one(),
-						redeem_fulfillment: Perquintill::one(),
-					}
-				]
-			),
-			Error::<Test>::NotNewBestSubmission
-		);
+		// Can submit the same solution twice
```
offerijns, line 806: Maybe you could add a test case specifically for the deadlock scenario? Basically copying https://github.com/centrifuge/tinlake/blob/develop/src/lender/test/coordinator-submitEpoch.sol#L250 and/or https://github.com/centrifuge/tinlake/blob/develop/src/lender/test/coordinator-submitEpoch.sol#L290

pallets/pools/src/lib.rs
offerijns: Never mind, I don't think this is an issue.

offerijns (approved):

pallets/pools/src/tests.rs
mustermeiszer: Not sure I am getting, that this is the deadlock scenario. With the current `>=` there can not be a deadlock as assured by the test above. But I can add a test that allows to submit a zero-solution as the first one, to trigger the `ChallangePeriod`. WDYT?

mustermeiszer:
@offerijns and @branan  I changed the `execute_epoch` error resolution slightly. Are you fine with this? Then you can hit the green button.

offerijns:
> @offerijns and @branan I changed the `execute_epoch` error resolution slightly. Are you fine with this? Then you can hit the green button.

Yeah good idea, this is more explicit 👍

pallets/pools/src/tests.rs
mustermeiszer: Ah, yes. That makes total sense. I will append this scenario to this pr.

Issue #720, opened 2022-03-23
Transfer asset NFTs into pool account


offerijns:
Right now, asset NFTs are transferred into the loans pallet. This will get rather confusing in the UI, as a single account will own all NFTs across pools. We should change this to transfer the asset NFTs to the pool account.

Once we add https://github.com/centrifuge/centrifuge-chain/issues/701, this will also mean that the pool identity will be linked to all the asset NFTs and all the currency in the pool.

Issue #717, opened 2022-03-22
Minor renamings & cleanups in loans & pools


offerijns:
- [x] Standardize on `loan_info` / `loan_data` in the `loans` pallet
- [x] Move `Moment` type to `common-types`
- [x] `rate_per_sec` => `interest_rate_per_sec` in `pools` pallet
- [x] Rename `borrowed_amount` to `total_borrowed` and `repaid_amount` to `total_repaid`
- [ ] Rename `latest_nav` to `latest`
- [x] Remove `_reserve` postfix from `ReserveDetails` struct
- [ ] Change `origination_date` to `Option<Moment>`

PR #721, opened 2022-03-23
Loans pallet cleanup

offerijns:
Resolves https://github.com/centrifuge/centrifuge-chain/issues/720 and part of https://github.com/centrifuge/centrifuge-chain/issues/717


pallets/loans/src/functions.rs
```diff
@@ -70,20 +71,16 @@ impl<T: Config> Pallet<T> {
 		);
 
 		// create new loan nft
-		let loan_pallet_account: T::AccountId = T::LoansPalletId::get().into_account();
```
mustermeiszer, line 73: Is the `LoansPalletId` used anywhere else to hold something or can we remove it?

pallets/loans/src/functions.rs
```diff
@@ -224,27 +213,24 @@ impl<T: Config> Pallet<T> {
...
 		Self::check_loan_owner(pool_id, loan_id, owner.clone())?;
 
-		LoanInfo::<T>::try_mutate(
+		Loan::<T>::try_mutate(
 			pool_id,
 			loan_id,
-			|maybe_loan_info| -> Result<bool, DispatchError> {
-				let mut loan_info = maybe_loan_info.take().ok_or(Error::<T>::MissingLoan)?;
+			|maybe_loan| -> Result<bool, DispatchError> {
+				let mut loan = maybe_loan.take().ok_or(Error::<T>::MissingLoan)?;
```
mustermeiszer, line 220: I would love to get rid of all the `let mut loan = maybe_loan.take().ok_or(Error::<T>::MissingLoan)?;` and rather work on the references instead via `let mut loan = maybe_loan.as_ref().ok_or(Error::<T>::MissingLoan)?;`.

Replacing the given Option with a None feels bad.

mustermeiszer (approved):
LGTM! 

Just one nity bit, that I don`t like the logic that we currently have here in most of the `try_mutate` calls, where we are taking the possible `Some(T)` and then storing it at the end of the logic again in the `Option<t>`

pallets/loans/src/functions.rs
branan: I agree, I much prefer the `as_mut()` pattern we use in the `Pools` pallet

pallets/loans/src/functions.rs
offerijns: It's used here: https://github.com/centrifuge/centrifuge-chain/blob/loans-cleanup/pallets/loans/src/functions.rs#L38 Currently the loan NFTs are transferred here, and it's also used by the tests to find the pallet account. We can probably consider removing this, but I'd rather do that in a separate PR.

pallets/loans/src/functions.rs
offerijns: Just changed this @mustermeiszer @branan

mustermeiszer (approved):
Nice! LGTM!

branan:
I'm happy with this now, as well. Will hit the button when CI completes

PR #723, opened 2022-03-24
scripts: improve by-product locations

NunoAlexandre:
We move all the `./scripts/init.sh` by-products to `/tmp/centrifuge-chain/<sub_dir>` to have a more organised structure for the different commands.

This also fixes the latest `onboard-parachain` writting the wasm file to the project dir and thus causing an unwanted git diff.


mustermeiszer (approved):
LGTM!

PR #726, opened 2022-03-28
Switch TrancheId generator hash-algorithm

mustermeiszer:
Currently we are planning to use `Twox128` in order to generate `TrancheIds`.  I think this is a problem. It's seems that it is trivially to generate collisions, especially for the input data that we have, which are deterministic. 
As we only generate `TranchIds` at the setup of pools and then sparsly from time to time when a tranche is replaced or added I think it is okay to use the slower `Blake_128`

### Attack scenario
* Off-chain: generate hashes with available pool-ids and increasing tranche-counters
* Compare the generated hashes with existing tranche-ids
* If one matches: create the respective pool and replace the senior-tranche with new tranches until the `TrancheId` matches the found collision

-> We then have two pools, one normal one, one controlled by the attacker, where we have a colliding `TranchId`. 

### Note
No migration needed as we don't have yet any pools.

### DIFF
--- a/pallets/loans/src/test_utils.rs
+++ b/pallets/loans/src/test_utils.rs
@@ -22,7 +22,7 @@ use common_types::PoolRole;
 use frame_support::sp_runtime::traits::One;
 use frame_support::traits::fungibles::Transfer;
 use frame_support::traits::tokens::nonfungibles::{Create, Inspect, Mutate};
-use frame_support::{assert_ok, parameter_types, StorageHasher, Twox128};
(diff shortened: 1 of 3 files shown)


mustermeiszer:
I am not sure if I am being paranoid here. But I would try this attack at least.

NunoAlexandre:
> I am not sure if I am being paranoid here. But I would try this attack at least.

If such an attack can happen, then you are not being paranoid. Better safe than attacked.

branan:
I think a cryptographic hash is the right choice here, given our input data :+1:

PR #731, opened 2022-03-29
Pools improvements

offerijns:
- Restructure `PoolDetails`
- Rename `PoolDetails`  impl methods


branan (changes requested):
I believe that both borrowers and lenders will also need `restricted_tokens` access in order to transfer tokens in and out of their accounts.

The refactors look good in this, though, if we want to split the PR?

offerijns:
> The refactors look good in this, though, if we want to split the PR?

This makes sense. I've just removed the proxy types, will submit a new PR for that.

mustermeiszer (approved):
LGTM!

PR #738, opened 2022-03-30
fix: pricing active but non borrowed against loans should be possible

### LINKED ISSUE #736: Enable pricing of active but non-financed loans
Once a loan is priced, it cannot be priced again. In the future we want to enable pricing of active loans, but this is complex. However, we should already support re-pricing active loans which have not been borrowed against, since this would solve issues with accidental mispricing.

Hence, we should change the logic from only allowing pricing for loans with the `Created` status, to allow pricing for loans with the `Created` status or the `Active` status and `total_borrowed == 0`.

offerijns:
Closes #736

### DIFF
--- a/pallets/loans/src/functions.rs
+++ b/pallets/loans/src/functions.rs
@@ -124,8 +124,12 @@ impl<T: Config> Pallet<T> {
 		Loan::<T>::try_mutate(pool_id, loan_id, |loan| -> DispatchResult {
 			let loan = loan.as_mut().ok_or(Error::<T>::MissingLoan)?;
 
-			// ensure loan is created
-			ensure!(loan.status == LoanStatus::Created, Error::<T>::LoanIsActive);
+			// ensure loan is created or priced but not yet borrowed against
+			ensure!(
+				loan.status == LoanStatus::Created
+					|| loan.status == LoanStatus::Active && loan.total_borrowed == Zero::zero(),
+				Error::<T>::LoanIsActive
+			);
 
 			// ensure loan_type is valid
 			let now = Self::now();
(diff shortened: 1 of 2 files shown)


pallets/loans/src/functions.rs
```diff
@@ -124,8 +124,12 @@ impl<T: Config> Pallet<T> {
...
 			let loan = loan.as_mut().ok_or(Error::<T>::MissingLoan)?;
 
-			// ensure loan is created
-			ensure!(loan.status == LoanStatus::Created, Error::<T>::LoanIsActive);
+			// ensure loan is created or priced but not yet borrowed against
+			ensure!(
+				loan.status == LoanStatus::Created
+					|| loan.status == LoanStatus::Active && loan.total_borrowed == Zero::zero(),
+				Error::<T>::LoanIsActive
+			);
```
mustermeiszer, line 132: Just thought about this a little. What happens when I have repaid everything (i.e. `total_borrowed = 0`, `status = Active` and I then call this function? 

Could this result in weird behavior?

mustermeiszer (commented):
LGTM! But I need to understand the implications of allowing this better.

pallets/loans/src/functions.rs
offerijns: Hm, maybe we need to rename `total_borrowed`. But this is an aggregated value, so it's not decreased on repayment (on purpose, because with a bullet loan, you shouldn't be able to borrow again after you've repaid the full value of the loan).

mustermeiszer: Then I am fine with this!

mustermeiszer (approved):
LGTM! Again, fast as hell! 🚀

PR #739, opened 2022-03-31
Integration tests

mustermeiszer:
First integration tests for ROCC.

## Focus
* This is currently just the "sceleton" for the integration tests for ROCC. In order to not let the PR get to big, I will add test-cases and more utility methods in upcoming PRs.


NunoAlexandre (approved):
Neat 👌

runtime/integration-tests/src/rocc/pool_starts.rs
```diff
@@ -9,7 +9,7 @@
 // but WITHOUT ANY WARRANTY; without even the implied warranty of
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU General Public License for more details.
-use crate::chain::centrifuge::PARA_ID;
+use crate::chain::centrifuge::{Runtime as CentrifugeRuntime, PARA_ID};
```
NunoAlexandre, line 12: Really not a big fan of renaming imports and it would read super nice to refer to these items as `centrifuge::Runtime` and `centrifuge::PARA_ID`.

mustermeiszer: I will rename it in the afterwards PR. Wdyt? Named it like this as I thought at the beginning I will also need the relay runtime but I think this won't be necessary.

NunoAlexandre: No biggy, we can always iterate on this later on, let's get the 🥩 on the table first yes :)

PR #740, opened 2022-03-31
centrifuge: Add chainbridge pallet

NunoAlexandre:
Along with it, we also add `pallet_nfts` and `pallet_bridge` as they are both dependencies of chainbridge.

### DIFF
--- a/runtime/centrifuge/src/lib.rs
+++ b/runtime/centrifuge/src/lib.rs
@@ -654,6 +654,62 @@ impl pallet_anchors::Config for Runtime {
 	type WeightInfo = ();
 }
 
+parameter_types! {
+	pub const NftProofValidationFee: u128 = NFT_PROOF_VALIDATION_FEE;
+}
+
+impl pallet_nft::Config for Runtime {
+	type Event = Event;
+	type ChainId = chainbridge::ChainId;
+	type ResourceId = chainbridge::ResourceId;
+	type HashId = HashId;
+	type NftProofValidationFee = NftProofValidationFee;
+	type WeightInfo = ();
+}
+
+parameter_types! {
+	pub const BridgePalletId: PalletId = PalletId(*b"c/bridge");
+	pub HashId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &sp_io::hashing::blake2_128(b"cent_nft_hash"));
+	//TODO create new mapping (< copied from 'development', need to figure out what this means)
+	pub NativeTokenId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &sp_io::hashing::blake2_128(b"xCFG"));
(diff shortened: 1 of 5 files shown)


runtime/centrifuge/src/lib.rs
```diff
@@ -654,6 +654,62 @@ impl pallet_anchors::Config for Runtime {
...
+	type HashId = HashId;
+	type NftProofValidationFee = NftProofValidationFee;
+	type WeightInfo = ();
+}
+
+parameter_types! {
+	pub const BridgePalletId: PalletId = PalletId(*b"c/bridge");
+	pub HashId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &sp_io::hashing::blake2_128(b"cent_nft_hash"));
+	//TODO create new mapping (< copied from 'development', need to figure out what this means)
+	pub NativeTokenId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &sp_io::hashing::blake2_128(b"xCFG"));
```
NunoAlexandre, line 674: 🚨  NOTE FOR REVIEWERS

Do you know what this means?

src/chain_spec.rs
```diff
@@ -768,6 +768,30 @@ fn centrifuge_genesis(
...
 		democracy: Default::default(),
 		parachain_system: Default::default(),
+		bridge: centrifuge_runtime::BridgeConfig {
+			// Whitelist chains Ethereum - 0
+			chains: vec![0],
+			// Register resourceIDs
+			resources: vec![
+				// xCFG ResourceID to PalletBridge.transfer method (for incoming txs)
+				(
+					// TODO(all): Adapt these values
```
mikiquantum, line 778: This can remain the same, we will need to update it in the runtime manually, but this works fine for testing development flows.

src/chain_spec.rs
```diff
@@ -768,6 +768,30 @@ fn centrifuge_genesis(
...
+					hex!["00000000000000000000000000000009e974040e705c10fb4de576d6cc261900"],
+					hex!["50616c6c65744272696467652e7472616e73666572"]
+						.iter()
+						.cloned()
+						.collect(),
+				),
+			],
+			// TODO(all): Adapt these values
```
mikiquantum, line 786: Same as above, no need to do anything else here.

mikiquantum (approved):
👍

runtime/centrifuge/src/lib.rs
NunoAlexandre: thanks @mikiquantum. so this code is good, we just need to make sure that on the Ethereum side we are ready to handle `xCFG`?

src/chain_spec.rs
NunoAlexandre, line 778:
```suggestion
```

src/chain_spec.rs
NunoAlexandre, line 786:
```suggestion
```

Issue #625, opened 2022-02-11
PRTU1 - Include Migration Pallet + Support changes


mikiquantum:
This is the first Runtime Upgrade on the new Centrifuge Parachain. It should cover most of the support code needed to make enabling transfers and other extrinsicis seamless.
Its scope:

- [x]  migration pallet: Remove automatic filtering when migration finishes https://github.com/centrifuge/centrifuge-chain/pull/741
- [x]  Add Migration Pallet to the Centrifuge Runtime https://github.com/centrifuge/centrifuge-chain/pull/742
- [x]  Add ChainBridge Pallet: https://github.com/centrifuge/centrifuge-chain/issues/625
- [x]  Adapt Claim Pallets to remove unsigned extrinsic: https://github.com/centrifuge/centrifuge-chain/pull/744
- [x] Change Minimum democracy deposit to 1000 CFG https://github.com/centrifuge/centrifuge-chain/pull/743
- [x] Update benchmarks
- [x] Bump runtime version

PR #741, opened 2022-03-31
pallets-migration: Drop filtering logic

NunoAlexandre:
Part of https://github.com/centrifuge/centrifuge-chain/issues/625

In a follow up PR, I will add this pallet to the `centrifuge-runtime`

### DIFF
--- a/runtime/development/src/lib.rs
+++ b/runtime/development/src/lib.rs
@@ -142,7 +142,6 @@ parameter_types! {
 
 // system support impls
 impl frame_system::Config for Runtime {
-	//TODO(reviews): Are we fine with Everything on dev, despite a possible migration taking place?
 	type BaseCallFilter = Everything;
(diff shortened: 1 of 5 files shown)


runtime/development/src/lib.rs
```diff
@@ -142,7 +142,8 @@ parameter_types! {
 
 // system support impls
 impl frame_system::Config for Runtime {
-	type BaseCallFilter = Migration;
+	//TODO(reviews): Are we fine with Everything on dev, despite a possible migration taking place?
```
mikiquantum, line 145: Yep its fine, its essentially how it works today

mikiquantum (approved):
👍

runtime/development/src/lib.rs
NunoAlexandre, line 145:
```suggestion
```

PR #742, opened 2022-03-31
centrifuge: Add migrations pallet

NunoAlexandre:
Part of https://github.com/centrifuge/centrifuge-chain/issues/625

We will run the benchmarks once everything in the parent issue is done.


mikiquantum (approved):
👍

PR #758, opened 2022-04-01
Integration Utilities and Testing Env

mustermeiszer:
A lot of utilities we are going to need and testing if the provided env works.

### DIFF
--- a/runtime/integration-tests/src/pools/utils/accounts.rs
+++ b/runtime/integration-tests/src/pools/utils/accounts.rs
@@ -1,0 +1,199 @@
+// Copyright 2021 Centrifuge Foundation (centrifuge.io).
+//
+// This file is part of the Centrifuge chain project.
+// Centrifuge is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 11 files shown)


runtime/integration-tests/src/pools/utils/accounts.rs
```diff
@@ -0,0 +1,199 @@
...
+pub enum Keyring {
+	Admin,
+	TrancheInvestor(u32),
+	Alice,
+	Bob,
+	Charlie,
+	Dave,
+	Eve,
+	Ferdie,
+	Costume(&'static str),
```
offerijns, line 33: What does this refer to? Should it be `Custom`?

offerijns (approved):
One question, otherwise LGTM

runtime/integration-tests/src/pools/utils/accounts.rs
mustermeiszer: ^^ Yes. I will change that.

mustermeiszer: I wil merge it an rename it different PR as otherwise the CI runs again...

PR #759, opened 2022-04-02
Scheduled pool updates & pool parameter refactor

offerijns:
### Challenge time
Changes challenge time to be a runtime-level parameter, as this really is a system parameter and there's no need for this to be set per pool. Also changes it to be defined in blocks.

### Min epoch time upper bound
Adds a min epoch time upper bound, to prevent pool admins from setting arbitrarily high min epoch times which would block any liquidity.

### Scheduled pool updates
Also introduces a way to schedule updates. This will have two checks:
- A minimum delay
- A requirement to only execute the update when all redemption orders are fulfilled.

The goal is that for a pool on Centrifuge Chain, these might be set to ~7 days and the requirement for redemptions is enabled. This means that once an issuer of a pool schedules an update, this can be shown in the UI, and investors can decide to submit redemption orders if they disagree with the update, which would mean they could get out of the pool before the update is executed.

Fixes the issue where interest rate updates apply retroactively.


pallets/pools/src/lib.rs
```diff
@@ -1107,7 +1160,7 @@ pub mod pallet {
 					epoch
 						.challenge_period_end
 						.expect("Challenge period is some. qed.")
-						<= Self::now(),
+						<= <frame_system::Pallet<T>>::block_number(),
```
mustermeiszer, line 1163: Can we make this a pallet call like `Self::now()`?

pallets/loans/src/mock.rs
```diff
@@ -172,11 +175,13 @@ impl pallet_pools::Config for MockRuntime {
 	type NAV = Loans;
 	type TrancheToken = TrancheToken<MockRuntime>;
 	type Time = Timestamp;
+	type ChallengeTime = ChallengeTime;
+	type MinUpdateDelay = MinUpdateDelay;
+	type RequireRedeemFulfillmentsBeforeUpdates = RequireRedeemFulfillmentsBeforeUpdates;
```
mustermeiszer, line 180: I would prefer this to be a trait like 

```rust
trait UpdateGuard {
   released(pool: &PoolDetails, update: &ScheduledUpdateDetails) -> bool;
}
```

Then we could implement the release checking logic on a per runtime basis and also be more flexible with respect to changing the logic of when we allow/dissallow updates to be released.

pallets/pools/src/lib.rs
```diff
@@ -447,6 +464,12 @@ pub mod pallet {
 		NoSolutionAvailable,
 		/// One of the runtime-level pool parameter bounds was violated
 		PoolParameterBoundViolated,
+		/// No update for the pool is scheduled
+		NoScheduledUpdate,
+		/// Scheduled time for this update is in the future
+		ScheduledTimeHasNotPassed,
+		/// In the last executed epoch, not all redemptions were fulfilled
+		RedemptionsHaveNotBeenFulfilled,
```
mustermeiszer, line 472: Would make this more generic and say `UpdatePrerequesitesNotFulfilled` (In case we are going with my idea from above)

pallets/pools/src/lib.rs
```diff
@@ -545,21 +570,93 @@ pub mod pallet {
...
+				);
+
+				ensure!(
+					pool.tranches.acc_outstanding_redemptions()?.is_zero(),
+					Error::<T>::RedemptionsHaveNotBeenFulfilled
+				);
+
+				Self::do_update_pool(&pool_id, &update.changes)?;
+			} else {
+				Self::do_update_pool(&pool_id, &update.changes)?;
```
mustermeiszer, line 656: What about getting the `PoolDetails` beforehand and erroring out if the pool does not exists and pass it then down the `do_update_pool` logic?

pallets/pools/src/lib.rs
```diff
@@ -1248,6 +1301,47 @@ pub mod pallet {
...
+				}
+
+				if let Change::NewValue(max_nav_age) = changes.max_nav_age {
+					pool.parameters.max_nav_age = max_nav_age;
+				}
+
+				if let Change::NewValue(tranches) = &changes.tranches {
+					let now = Self::now();
+
+					pool.tranches.combine_with_mut_non_residual_top(
```
mustermeiszer, line 1322: Aren't the tranches in the `Vec<TranchInput>` sorted from residual-to-non-residual? Then `combine_with_mut_residual_top`

mustermeiszer (commented):
This looks really good!

I left a comment that I would prefer to use a trait instead of a boolean and leave the actual logic that releases an update up to the specific runtime. This is no blocker just wanted to discuss the idea and if we wanna go with it. 

And regarding the block vs. seconds (just wanna understand our reasoning here):
* Why use blocks only for challenge time? 
* `MinUpdateDelay`: Wouldn't it make sense to make this in blocks to?

pallets/pools/src/lib.rs
offerijns: Yeah good point. I guess it was wrong in the old `update_tranches` implementation then, I copied it from there: https://github.com/centrifuge/centrifuge-chain/blob/parachain/pallets/pools/src/lib.rs#L651

branan: `combine_with_mut_non_residual_top` was bugged until a recent change, and was equivalent to the residual-top version. I fixed one incorrect usage in the benchmarks PR. I'm not surprised to see another one here

pallets/loans/src/mock.rs
offerijns: This is done now.

pallets/pools/src/lib.rs
offerijns: This would also mean moving `try_mutate` to the places where this is called, which I think is adding rather than reducing code.

pallets/pools/src/lib.rs
offerijns: Fixed.

pallets/pools/src/lib.rs
```diff
@@ -587,18 +619,22 @@ pub mod pallet {
...
+		#[pallet::weight(T::WeightInfo::update_no_execution()
+			.max(T::WeightInfo::update_and_execute()))]
 		pub fn update(
 			origin: OriginFor<T>,
 			pool_id: T::PoolId,
-			min_epoch_time: u64,
-			challenge_time: u64,
-			max_nav_age: u64,
+			changes: PoolChanges<T::InterestRate>,
 		) -> DispatchResult {
```
branan, line 638: This should be updated to use a `DispatchResultWithPostInfo` and the return value updated from `()` to the appropriate weight for each code paths. Otherwise users will always be paying for execution even if we defer it.

pallets/pools/src/lib.rs
```diff
@@ -611,21 +647,92 @@ pub mod pallet {
...
+
+			let now = Self::now();
+
+			let update = ScheduledUpdateDetails {
+				changes: changes.clone(),
+				scheduled_time: now.saturating_add(T::MinUpdateDelay::get()),
+			};
+
+			if T::MinUpdateDelay::get() == 0 && T::UpdateGuard::released(&pool, &update, now) {
+				Self::do_update_pool(&pool_id, &changes)
```
branan, line 696: Here (and the other branch) is where we need to return the actual weight consumed

branan (changes requested):
One important change. Otherwise this is generally looking good :+1:

branan (approved):

PR #760, opened 2022-04-04
Fixup benchmarks and enable them as tests

branan:


branan:
Two pallets are not currently running their benchmarks as tests:

* crowdloan claim. The mock runtime for tests uses different account types, and trying to fixup all the craziness that this pallet does didn't seem worth it
* restricted tokens. This one we should fix. It's mostly an issue of the mock runtime using a custom permissions implementation and currency ID - fixing the unit tests to just use `pallet_permissions` should then let the benchmarks also run on the mock runtime.

ci/script.sh
```diff
@@ -25,7 +25,7 @@ case $TARGET in
     ;;
 
   tests)
-    cargo test --workspace --release
+    RUST_MIN_STACK=8388608 cargo test --workspace --release --features runtime-benchmarks
```
branan, line 28: secondary threads on linux normally start with a slightly smaller stack than the main thread. This was necessary on my laptop to get benchmark tests to run without forcing tests into single-threaded mode.

branan:
I'd recommend reviewing this commit-by-commit, as they're each pretty well cleaned up and squashed, and keep related changes better together.

branan:
This may fail on the integration tests - I haven't fully tracked down that issue and it's been sporadic for me (yay for heisenbugs). If it does fail, is that something you can look into @mustermeiszer? It'd be another heavy lift for me to learn how Fudge works right now, and this work has already taken _so long_

mustermeiszer (approved):
Wow, that was an awful lot of changes needed! Thanks @branan for taking care of this. 

Regarding the crowdloan-claim pallet:
* Think its fine to not fix this. There won't be any changes.

Regarding the integration-tests crate: 
* I fixed this in the next release of fudge and will include those changes into my PR that's already up.

mustermeiszer:
@branan the fix for the issues with the integration tests is here https://github.com/branan/centrifuge-chain/pull/6/files

mustermeiszer (approved):
LGTM!

PR #761, opened 2022-04-05
Adding restrictions check for pool-currency

mustermeiszer:
As we want to control which currencies are allowed as base-currencies for a pool this PR implements a check for this.


runtime/development/src/lib.rs
```diff
@@ -856,6 +856,18 @@ impl pallet_pools::Config for Runtime {
...
 	type TrancheWeight = TrancheWeight;
+	type PoolCurrency = PoolCurrencyChecker;
+}
+
+struct PoolCurrencyChecker;
+impl PoolCurrency for PoolCurrencyChecker {
+	type CurrencyId = CurrencyId;
+	fn base(id: Self::CurrencyId) -> bool {
+		match id {
+			CurrencyId::Tranche(_, _) | CurrencyId::Native => false,
```
offerijns, line 867: Here it's fine, but we should remember to re-enable `Native` as supported on `altair`, since the first pool on Altair will launch with AIR as currency.

pallets/pools/src/lib.rs
```diff
@@ -447,6 +450,9 @@ pub mod pallet {
 		NoSolutionAvailable,
 		/// One of the runtime-level pool parameter bounds was violated
 		PoolParameterBoundViolated,
+		/// Indicates that a pool base currency that is NOT whitelisted was used
+		/// for creating a pool
+		NotAPoolBaseCurrency,
```
offerijns, line 455: I'm nitpicking, but I don't think "pool base currency" is a thing, it's just the pool currency (there are no other types of pool currencies).
```suggestion
		NotAValidPoolCurrency,
```

libs/common-traits/src/lib.rs
```diff
@@ -222,3 +222,24 @@ pub trait TokenMetadata {
...
+
+/// Trait for checking which currencies are "whitelisted" as pool-currencies.
+/// Runtimes need to define which one they want to allow to be
+/// a base currency for a pool.
+pub trait PoolCurrency {
+	type CurrencyId;
+
+	/// Checks whether the given currency is allowed as a
+	/// base currency of a pool
+	fn base(currency: Self::CurrencyId) -> bool;
```
offerijns, line 234: Maybe it's clearer to call this method any of `is_valid` / `is_supported` / `is_allowed`?

libs/common-traits/src/lib.rs
```diff
@@ -222,3 +222,24 @@ pub trait TokenMetadata {
 pub trait TrancheToken<PoolId, TrancheId, CurrencyId> {
 	fn tranche_token(pool: PoolId, tranche: TrancheId) -> CurrencyId;
 }
+
+/// Trait for checking which currencies are "whitelisted" as pool-currencies.
```
offerijns, line 226:
```suggestion
/// Trait for checking which currencies are allowed as pool-currencies.
```

pallets/pools/src/lib.rs
```diff
@@ -447,6 +450,9 @@ pub mod pallet {
 		NoSolutionAvailable,
 		/// One of the runtime-level pool parameter bounds was violated
 		PoolParameterBoundViolated,
+		/// Indicates that a pool base currency that is NOT whitelisted was used
```
offerijns, line 453:
```suggestion
		/// Indicates that a pool currency that is NOT allowed was used
```

offerijns (approved):
That was really quick! Implementation looks good, have a few comments on naming though.

libs/common-traits/src/lib.rs
mustermeiszer: moved to `allowed`. Does this work for you?

libs/common-traits/src/lib.rs
mustermeiszer: Done

pallets/pools/src/lib.rs
mustermeiszer: Done

pallets/pools/src/lib.rs
mustermeiszer: Done

libs/common-traits/src/lib.rs
offerijns: Yep!

pallets/pools/src/lib.rs
NunoAlexandre: "InvalidCurrency"

mustermeiszer: I like that.

NunoAlexandre (commented):
This is fine but I'd fine it more reusable and simpler to just use the `Contains` trait.

We would:

- Change the pools type param to `PoolCurrency: Contains<Self::CurrencyId>`
- Delete the `PoolCurrency` trait
- Replace `()` with `Everything`
-  Rename `PoolCurrencyChecker` with just `PoolCurrency`

mustermeiszer:
@NunoAlexandre cleaned-up as suggested ❤️

NunoAlexandre (approved):
Awesome :) thanks!!!

pallets/pools/src/mock.rs
```diff
@@ -302,6 +302,17 @@ impl Config for Test {
...
 	type WeightInfo = ();
 	type TrancheWeight = TrancheWeight;
+	type PoolCurrency = PoolCurrency;
+}
+
+pub struct PoolCurrency;
+impl Contains<CurrencyId> for PoolCurrency {
+	fn contains(id: &CurrencyId) -> bool {
+		match id {
+			CurrencyId::Tranche(_, _) | CurrencyId::Native => false,
```
branan, line 312: Are we sure we won't want to allow Native, especially on altair where we might be financing crypto-native assets?

branan: Whoops, Jeroen made the same comment below. All set :)

PR #764, opened 2022-04-06
Tests for Pools - Part 1

mustermeiszer:
First tests for actual pool logic


mustermeiszer:
This is mostly again utilities. But writing the test themselves should be straightforward now.

runtime/integration-tests/Cargo.toml
```diff
@@ -9,7 +9,8 @@ repository = "https://github.com/centrifuge/centrifuge-chain"
 
 
 [dependencies]
-fudge = { git = "https://github.com/centrifuge/fudge", tag = "v0.0.7-polkadot-v0.9.17"}
+#fudge = { git = "https://github.com/centrifuge/fudge", tag = "v0.0.5-polkadot-v0.9.17"}
```
offerijns, line 12:
```suggestion
```

runtime/integration-tests/src/pools/utils/loans.rs
```diff
@@ -0,0 +1,245 @@
...
+
+/// Issues a default loan with the following properties
+/// * 15% APR
+/// * value with amount
+/// * maturity as given
+/// * Type: BulletLoan
+/// 	* advance_rate: 90%,
+///     * probability_of_default: 5%,
+///     * loss_given_default: 50%,
+/// 	* discount_rate: 4% ,
```
offerijns, line 116: Just an FYI that a more realistic example is that `discount_rate` is slightly below the interest rate. E.g. a discount rate of 3.5% for an interest rate of 4%. If `interest rate == discount rate`, then the debt basically grows at the same rate as the present value (if we exclude risk adjustments from PD & LGD), which isn't a practical example as future cashflows should generally be discounted at present.

runtime/integration-tests/src/pools/utils/pools.rs
```diff
@@ -0,0 +1,450 @@
...
+/// * Whitelistings
+/// 	* Keyring::TrancheInvestor(index) accounts with index 0 - 9 for tranche with id 0
+///  	* Keyring::TrancheInvestor(index) accounts with index 10 - 19 for tranche with id 1
+/// 	* Keyring::TrancheInvestor(index) accounts with index 20 - 29 for tranche with id 2
+/// 	* Keyring::TrancheInvestor(index) accounts with index 30 - 39 for tranche with id 3
+/// 	* Keyring::TrancheInvestor(index) accounts with index 40 - 49 for tranche with id 4
+/// * Currency: CurrencyId::Usd,
+/// * MaxReserve: 100_000 Usd
+pub fn default_pool_calls(admin: AccountId, pool_id: PoolId, nfts: &mut NftManager) -> Vec<Call> {
+	pool_setup_calls(
```
offerijns, line 115: Great to have all these utilities 🔥

runtime/integration-tests/src/pools/utils/time.rs
```diff
@@ -0,0 +1,36 @@
...
+
+//! Time constants and utlities around time
+
+/// Start date used for timestamps in test-enviornments
+/// Sat Jan 01 2022 00:00:00 GMT+0000
+pub const START_DATE: u64 = 1640995200u64;
+
+pub mod secs {
+	pub const SECOND: u64 = 1000u64;
+	pub const SECONDS_PER_MINUTE: u64 = 60 * SECOND;
```
offerijns, line 21: Could these be re-used from https://github.com/centrifuge/centrifuge-chain/blob/parachain/runtime/common/src/lib.rs#L195?

offerijns (approved):

runtime/integration-tests/Cargo.toml
mustermeiszer: Wil adapt this once I release the latest fudge version.

runtime/integration-tests/src/pools/utils/loans.rs
mustermeiszer: Thanks! Makes sense. And for the assumed interest rate we take the interest rate of the top-non-residual tranche (i.e. the most senior one)?

runtime/integration-tests/src/pools/utils/time.rs
mustermeiszer: Will take a look!

mustermeiszer: I think I keep them although being a duplicate. This makes it more easy to integrate as `time::blocks` and `time::secs`

runtime/integration-tests/src/pools/utils/loans.rs
offerijns: The loan interest rate should be slightly higher than the interest rate of the top-non-residual tranche. Otherwise the residual tranche makes a loss or makes the same yield as the top-non-residual tranche, while instead it should get a higher yield to compensate for the higher risk.

offerijns: As a practical example, the REIF pool currently has a senior (non residual) interest rate of 5%, an average financing fee (loan interest rate) of 9.5% and a discount rate for all loans of 7% (that's probably a bit too low actually). This leads to junior (residual) getting a yield of ~15%.

mustermeiszer:
@branan Can this go in or do you wanna take another look?

runtime/integration-tests/src/pools/utils/accounts.rs
```diff
@@ -127,25 +239,185 @@ impl std::str::FromStr for Keyring {
...
+/// * Keyring::TrancheInvestor(42)
+/// * Keyring::TrancheInvestor(43)
+/// * Keyring::TrancheInvestor(44)
+/// * Keyring::TrancheInvestor(45)
+/// * Keyring::TrancheInvestor(46)
+/// * Keyring::TrancheInvestor(47)
+/// * Keyring::TrancheInvestor(48)
+/// * Keyring::TrancheInvestor(49)
+/// * Keyring::TrancheInvestor(50)
+pub fn default_investors() -> Vec<Keyring> {
```
branan, line 369: how about (0..50).iter().map(|idx| Keyring::TrancheInvestor(idx+1)).collect()

branan: (though actually, keeping Investor 0 might also be fine?)

runtime/integration-tests/src/pools/utils/time.rs
branan: `runtime_common` should _really_ only be used by the three main runtimes. If we want to DRY this up we should move it into `lib/common_something`

runtime/integration-tests/src/pools/utils/accounts.rs
```diff
@@ -12,12 +12,124 @@
...
 pub use sp_core::sr25519;
 use sp_core::{
 	sr25519::{Pair, Public, Signature},
 	Pair as PairT,
 };
 use sp_runtime::AccountId32;
+use std::collections::HashMap;
+
+/// Struct that takes care of handling nonces for accounts
+pub struct NonceManager {
```
branan, line 31: All the nonce stuff feels like pretty core functionality that maybe should be in fudge instead of in the test utils?

branan (approved):
I could find some more nitpicks, but I think this is a solid increment of what we have. Feel free to ignore my other comments if you think they're the wrong direction.

runtime/integration-tests/src/pools/utils/accounts.rs
mustermeiszer: Nice, yes will do that in the follow up PR! And you are right, I will keep the 0 investor. Already changed that in the follow up, but wanna close this first.

mustermeiszer:
I won't introduce the suggested changes into this PR but in the follow Up as I want to close this here first. 

Regarding the nonce stuff: 
I think it should not be part of fudge as it makes pretty strong assumptions 
* Private keys are passed to fudge here
* nonces are always managed by fudge
* Signing must normally happen before submitting an extriniscs -> I just made the interface accept the `Call` enum to abstract away the signing process and nonce management. As this otherwise gets really messy. 

Hope you are fine with this, otherwise we can of course discuss further.

PR #765, opened 2022-04-08
integration-tests: Adapt structure

NunoAlexandre:
With the latest changes to the integration tests, the new structure had us refer to whichever selected runtime (`altair`/`development`/`centrifuge`) as `::centrifuge::` which is confusing.

The main change is is going from:

``` rust
mod chain {
  mod centrifuge {
     // import actually selected runtime `altair`/`development`/`centrifuge`
  }

  mod relay {
     // import the right relay for the selected runtime `altair`/`development`/`centrifuge`
  }
}
```

to

``` rust
mod parachain {
  // import actually selected runtime `altair`/`development`/`centrifuge`
}

mod relay {
  // import the right relay for the selected runtime `altair`/`development`/`centrifuge`
}
```

The `chain::` prefix mod didn't seem to add any palpable value and this way we now refer to the selected parachain runtime as `::parachain::`, which is fitting to all usages.

### DIFF
--- a/runtime/integration-tests/src/lib.rs
+++ b/runtime/integration-tests/src/lib.rs
@@ -16,43 +16,41 @@
 mod pools;
 mod xcm;
 
-/// Re-exports the correct runtimes that we run the integration tests with
-/// This allows all other modules to use the import of `crate::chain::{...}`
-/// in order to get the right stuff from the respective runtime.
-mod chain {
+/// Re-exports the correct runtimes that we run the integration tests with.
+/// This allows all other modules to use the import of `crate::parachain::{...}` and
+/// `crate::relay::{...}` in order to get the right stuff from the respective runtime.
+pub mod parachain {
+	#[cfg(feature = "runtime-altair")]
+	pub use altair::*;
+	#[cfg(feature = "runtime-centrifuge")]
+	pub use centrifuge::*;
+	#[cfg(feature = "runtime-development")]
+	pub use development::*;
+
+	#[cfg(feature = "runtime-centrifuge")]
 	pub mod centrifuge {
-		#[cfg(feature = "runtime-altair")]
-		pub use altair::*;
-		#[cfg(feature = "runtime-centrifuge")]
-		pub use centrifuge::*;
-		#[cfg(feature = "runtime-development")]
-		pub use development::*;
-
-		#[cfg(feature = "runtime-centrifuge")]
-		pub mod centrifuge {
-			pub use centrifuge_runtime::*;
-			pub const PARA_ID: u32 = 2031;
-		}
-
-		#[cfg(feature = "runtime-altair")]
-		pub mod altair {
-			pub use altair_runtime::*;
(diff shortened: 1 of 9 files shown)


mustermeiszer:
@NunoAlexandre I would like to keep it as is. When we integrate acala here for example this will be more straightforward like `chain::acala` and `chain::centrifuge`.

NunoAlexandre:
@mustermeiszer this current setup does not work, sorry. Currently this "says" that we are using centrifuge for the xcm tests which is just wrong and misleading.

To support other external runtimes, we can always have another mod.

mustermeiszer:
> this current setup does not work, sorry.
Working and being misleading are two different things. 

How would the naming be in the case of another external runtime?

NunoAlexandre:
We are getting lost in semantics. Naming a reference to Altair as "Centrifuge" is a no-go.

Regarding naming for external runtimes, I don't know. We could have all the runtimes under a "parachain" module as suggested here, and we would pick the specific one based on the need; for example, for the xcm tests that now focus on Altair, we would have "parachain::altair" and in the future also "parachain::karura". I don't know yet how/ if this is possible with cargo.

Or we wrap them within their respective network, and we would have "kusama:: altair" and "kusama::karura", respectively, as well as "kusama::relay". That way we don't have the issue of grouping all runtimes under a company name and fall into the name collision issue.

mustermeiszer:
I disagree here. We are naming the project and the used runtime is implicit. Like our repo is named `centrifuge-chain`, while it contains multiple runtimes (`Altair`, `Centrifuge`). 

I think we won't come to a consensus here. What do @offerijns, @mikiquantum  and @branan think?

NunoAlexandre:
> Like our repo is named `centrifuge-chain`, while it contains multiple runtimes (`Altair`, `Centrifuge`)

That is not a fair analogy tho, as we never refer to all of those different runtimes as "centrifuge". Instead, we must be explicit regarding the runtime we want to use.

mustermeiszer:
True, it's not really a fair analogy and I really see the point that is bugging you.

But I just don't see how we could integrate later with other runtimes with your proposed wording, without losing the capability of writing tests without importing the "chains" always with feature guards.

NunoAlexandre:
> But I just don't see how we could integrate later with other runtimes with your proposed wording, without losing the capability of writing tests without importing the "chains" always with feature guards.

Fair enough. I also don't know for sure. Specially when considering the case of being able, in the future, of referring to multiple runtimes cross-network to, for example, test USDT transfers from Altair to Centrifuge or something alike.

If we could flag a `*.rs` to only be included for a specific flag, than we can follow the suggestion I left above and have an explicit setup where we could do `kusama::altair`, `kusama::karura`, `kusama::relay` or `polkadot::centrifuge` or `rococo::development`, etc.

I will see what's possible. I doubt that IDE's support this smoothly but the compiler should handle this just fine.

NunoAlexandre:
Closing this for now, this will need some more thorough thinking

PR #766, opened 2022-04-09
Introduce Role enum and generic permissioned currency variant

offerijns:
- Introduces a new `Role` enum, and moves the existing roles to the `PoolRole` variant, so we can have non-pool specific roles.
- Adds a `Permissioned` currency type, and corresponding roles. This can be used to create permissioned currencies other than the tranche tokens.
- Renames permission `Location` to `Scope`.

In a future PR, the `Tranche` currency will be rewritten to be a variant of the `Permissioned` currency type, so the permissioning logic is re-used.


libs/common-types/src/lib.rs
```diff
@@ -63,9 +100,28 @@ bitflags::bitflags! {
 		const LIQUIDITY_ADMIN = 0b00001000;
```
offerijns, line 100: @mustermeiszer wdyt, should we change this to `u64` already to accommodate more roles in the future?

mustermeiszer: It's twice the storage. Do you think we will have more than 32 roles on a pool level? We can add those bitflags for other role locations too. But if we think we will have more than 32 roles, better adapt now!

libs/common-types/src/lib.rs
```diff
@@ -53,6 +56,40 @@ pub enum PoolRole<TrancheId = [u8; 16], Moment = u64> {
...
+	/// This role can mint/burn tokens
+	Issuer,
+}
+
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, TypeInfo, Debug)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum Role<TrancheId = [u8; 16], Moment = u64> {
+	PoolRole(PoolRole<TrancheId, Moment>),
+	PermissionedCurrencyRole(PermissionedCurrencyRole<Moment>),
+}
```
NunoAlexandre, line 75: Could you leave a type-level comment explaining how this type is used overall and within its specific variants?

libs/common-types/src/lib.rs
```diff
@@ -130,47 +197,166 @@ where
...
 		match property {
-			PoolRole::Borrower => Ok(self.admin.remove(AdminRoles::BORROWER)),
-			PoolRole::LiquidityAdmin => Ok(self.admin.remove(AdminRoles::LIQUIDITY_ADMIN)),
-			PoolRole::PoolAdmin => Ok(self.admin.remove(AdminRoles::POOL_ADMIN)),
-			PoolRole::PricingAdmin => Ok(self.admin.remove(AdminRoles::PRICING_ADMIN)),
-			PoolRole::MemberListAdmin => Ok(self.admin.remove(AdminRoles::MEMBER_LIST_ADMIN)),
-			PoolRole::RiskAdmin => Ok(self.admin.remove(AdminRoles::RISK_ADMIN)),
-			PoolRole::TrancheInvestor(id, delta) => self.tranche_investor.remove(id, delta),
+			Role::PoolRole(pool_role) => match pool_role {
+				PoolRole::Borrower => Ok(self.admin.remove(AdminRoles::BORROWER)),
```
NunoAlexandre, line 240: I wonder if we could have a `TryInto` of sorts to try and convert `PoolRole` to `AdminRoles`. Having that would dry a lot of these duplicated match cases, reduce the noise, and give room for the exception cases.

libs/common-types/src/lib.rs
```diff
@@ -130,47 +197,166 @@ where
...
+	fn validity(&self, delta: Moment) -> Result<Moment, ()> {
+		let now: Moment = Now::now().as_secs().into();
+		let min_validity = now.saturating_add(MinDelay::get());
+		let req_validity = now.saturating_add(delta);
+
+		if req_validity < min_validity {
+			Err(())
+		} else {
+			Ok(req_validity)
+		}
```
NunoAlexandre, line 315: Doesn't matter but in many other places we and Substrate overall opts for this style instead:

```suggestion
		if req_validity < min_validity {
			return Err(());
		}
		
		Ok(req_validity)
```

libs/common-types/src/lib.rs
```diff
@@ -130,47 +197,166 @@ where
...
+			let now = Now::now().as_secs().into();
+
+			if *valid_till <= now {
+				// The account is already invalid. Hence no more grace period
+				Err(())
+			} else {
+				// Ensure that permissioned_till is at least now + min_delay.
+				Ok(self.info = Some(PermissionedCurrencyHolderInfo {
+					permissioned_till: self.validity(delta)?,
+				}))
```
NunoAlexandre, line 338: I find this quite confusing to read. How about this?

```suggestion
        let permissioned_till = self.validity(delta)?;
        self.info = Some(PermissionedCurrencyHolderInfo {
					permissioned_till
				});
				Ok(())
```

libs/common-types/src/lib.rs
```diff
@@ -130,47 +197,166 @@ where
...
+
+	pub fn contains(&self) -> bool {
+		if let Some(info) = &self.info {
+			info.permissioned_till >= Now::now().as_secs().into()
+		} else {
+			false
+		}
+	}
+
+	pub fn remove(&mut self, delta: Moment) -> Result<(), ()> {
```
NunoAlexandre, line 326: Are we sure we want the return type to be `Result<(), ()>`?

libs/common-types/src/lib.rs
```diff
@@ -130,47 +197,166 @@ where
...
+				Err(())
+			} else {
+				Ok(self.info = Some(PermissionedCurrencyHolderInfo {
+					permissioned_till: validity,
+				}))
+			}
+		} else {
+			Ok(self.info = Some(PermissionedCurrencyHolderInfo {
+				permissioned_till: validity,
+			}))
```
NunoAlexandre, line 359: We should simplify this:

```suggestion
    match &self.info {
      Some(info if info.permissioned_till > validity => Err(()),
      _ => Ok(self.info = Some(PermissionedCurrencyHolderInfo {
					permissioned_till: validity,
				}))
    }
```

libs/common-types/src/tokens.rs
```diff
@@ -7,13 +7,22 @@ use serde::{Deserialize, Serialize};
...
+}
+
 #[derive(
 	Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen,
 )]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 pub enum CurrencyId {
 	Native,
 	Usd,
+	Permissioned(PermissionedCurrency),
```
NunoAlexandre, line 25: Could you move this variant to the end of the list? The variants are encoded as numbers and for some xcm integrations in XTokens (so Native == 0/1, Usd = 1/2, etc) so it will be good to leave the existing variants what they were.

NunoAlexandre (commented):
The overall feature looks good to me. However, I missed some docs on some types and/or functions and also left some suggestions to clean up some code.

libs/common-types/src/tokens.rs
offerijns: Good point!

libs/common-types/src/lib.rs
offerijns: This is a bigger change I think, fine if I add a ticket for this instead?

libs/common-types/src/lib.rs
offerijns: Yeah you're right, 32 should be fine.

libs/common-types/src/lib.rs
NunoAlexandre: thanks a lot @offerijns 👍

NunoAlexandre (approved):
🏄‍♂️

libs/common-types/src/lib.rs
offerijns: https://github.com/centrifuge/centrifuge-chain/issues/780

libs/common-types/src/lib.rs
offerijns: This wasn't changed in this PR, but I think the goal is just to return whether it failed or not, there's no need to return any more information if it succeeded. What do you think should be returned otherwise?

libs/common-types/src/lib.rs
```diff
@@ -53,6 +56,45 @@ pub enum PoolRole<TrancheId = [u8; 16], Moment = u64> {
 	TrancheInvestor(TrancheId, Moment),
```
mustermeiszer, line 56: Shouldn`t this be removed? Or due to tests we want to do this later? ^^

libs/common-types/src/lib.rs
```diff
@@ -53,6 +56,45 @@ pub enum PoolRole<TrancheId = [u8; 16], Moment = u64> {
...
+}
+
+impl<PoolId, CurrencyId> Default for PermissionScope<PoolId, CurrencyId>
+where
+	PoolId: Default,
+{
+	fn default() -> Self {
+		Self::Pool(PoolId::default())
+	}
+}
```
mustermeiszer, line 96: Can we hide this behind an `#[cfg(test)]` guard?

libs/common-types/src/lib.rs
```diff
@@ -63,9 +105,28 @@ bitflags::bitflags! {
 		const LIQUIDITY_ADMIN = 0b00001000;
 		const MEMBER_LIST_ADMIN = 0b00010000;
 		const RISK_ADMIN = 0b00100000;
+		const PERMISSIONED_ASSET_MANAGER = 0b01000000;
+		const PERMISSIONED_ASSET_ISSUER = 0b10000000;
```
mustermeiszer, line 109: We can think about extending the `PermissionRoles` as 

```rust
struct PermissionRoles {
    pool_admin: PoolAdminRoles,
    curreny_admin: CurrencyAdminRoles,
    permissioned_asset_holder: PermissionedCurrencyHolders<Now, MinDelay, Moment>,
    tranche_investor: TrancheInvestors<Now, MinDelay, TrancheId, Moment>,
```

if we wanna split the admin roles further. But just an idea.

mustermeiszer (approved):
Really nice. Although, it gets complicated already. I am not sure, if my design is the best here... 

Anyways. LGTM! Sorry that it took so long to review. 

Just one note: Can we add a few tests that test the new `CurrencyRole` variant? This would be important for me.

libs/common-types/src/lib.rs
offerijns: Good point 👍

libs/common-types/src/lib.rs
offerijns: Hm yes, I think that's a good idea! Will explore this.

mustermeiszer (approved):
LGTM!

PR #767, opened 2022-04-11
xcm: Support KSM + KUSD

NunoAlexandre:
Highlights here:

- We add support and tests for bridging `KSM` and `KUSD`
- We fix and test the `Trader` rules
- We implement `ToTreasury` (not very important yet tho)

Notes:
- With the new integration-tests setup, we were locked up to use `rococo` paired with the `development` runtime, where XCM is placed now. However, we need to use `Kusama` instead, so I left a small tweak to allow for that until we add XCM to Altair (this week)

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -23,6 +23,8 @@ pub use types::*;
 mod fixed_point;
 mod impls;
 
+pub use common_types::CurrencyId;
+
 pub mod apis {
 	use node_primitives::{BlockNumber, Hash};
 	use pallet_anchors::AnchorData;
@@ -230,5 +232,52 @@ pub mod constants {
 
 	pub const fn deposit(items: u32, bytes: u32) -> Balance {
 		items as Balance * 15 * CENTI_CFG + (bytes as Balance) * 6 * CENTI_CFG
+	}
+}
+
+pub mod parachains {
+	pub mod karura {
+		pub const ID: u32 = 2000;
+		pub const KUSD_KEY: &[u8] = &[0, 129];
+	}
+}
+
+pub mod xcm_fees {
+	use common_traits::TokenMetadata;
+	use common_types::CurrencyId;
+	use frame_support::weights::constants::{ExtrinsicBaseWeight, WEIGHT_PER_SECOND};
+
+	use super::types::Balance;
+	use super::CENTI_CFG as CENTI_CURRENCY;
+
+	pub fn base_tx_in_air() -> Balance {
+		CENTI_CURRENCY / 10
+	}
+
+	// The fee cost per second for transferring the native token in cents.
+	pub fn native_per_second() -> Balance {
+		base_tx_per_second(CurrencyId::Native)
+	}
+
+	pub fn ksm_per_second() -> Balance {
+		base_tx_per_second(CurrencyId::KSM) / 50
+	}
+
+	fn base_tx_per_second(currency: CurrencyId) -> Balance {
+		let base_weight = Balance::from(ExtrinsicBaseWeight::get());
+		let base_tx_per_second = (WEIGHT_PER_SECOND as u128) / base_weight;
+		base_tx_per_second * base_tx(currency)
+	}
+
+	fn base_tx(currency: CurrencyId) -> Balance {
+		cent(currency) / 10
+	}
+
+	pub fn dollar(currency_id: common_types::CurrencyId) -> Balance {
+		10u128.saturating_pow(currency_id.decimals().into())
+	}
+
+	pub fn cent(currency_id: CurrencyId) -> Balance {
+		dollar(currency_id) / 100
 	}
 }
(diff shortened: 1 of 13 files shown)


NunoAlexandre:
@mustermeiszer the build is failing with this error but this seems to be a caching issue 🤔 

Screenshot 2022-04-11 at 12 39 27

I can build and test the repo successfully locally. Do you know what's flaring up?

runtime/development/src/lib.rs
```diff
@@ -1325,38 +1329,74 @@ impl xcm_executor::Config for XcmConfig {
...
+			id: Concrete(location),
+			fun: Fungible(amount),
+		} = revenue
+		{
+			if let Ok(currency_id) =
+				<CurrencyIdConvert as Convert<MultiLocation, CurrencyId>>::convert(location)
+			{
+				let _ = OrmlTokens::deposit(currency_id, &TreasuryAccount::get(), amount);
+			}
+		}
```
mustermeiszer, line 1355: Where is `take_revenue` called and what happens if the given `CurrencyId`-`Multilocation`-combination is not convertible?

runtime/common/src/lib.rs
```diff
@@ -232,3 +234,50 @@ pub mod constants {
 		items as Balance * 15 * CENTI_CFG + (bytes as Balance) * 6 * CENTI_CFG
 	}
 }
+
+pub mod parachains {
+	pub mod karura {
+		pub const ID: u32 = 2000;
+		pub const KUSD_KEY: &[u8] = &[0, 129];
```
mustermeiszer, line 241: What is they key used for?

runtime/development/src/lib.rs
```diff
@@ -1325,38 +1329,74 @@ impl xcm_executor::Config for XcmConfig {
...
+}
+
 parameter_types! {
+	pub KsmPerSecond: (AssetId, u128) = (MultiLocation::parent().into(), ksm_per_second());
+
 	pub NativePerSecond: (AssetId, u128) = (
 		MultiLocation::new(
 			1,
-			X2(Parachain(2000), GeneralKey(CurrencyId::Native.encode())),
+			X2(Parachain(2088), GeneralKey(CurrencyId::Native.encode())),
```
mustermeiszer, line 1365: I remember that in the docs for XCM they stated that the `GeneralKey` variant should be omitted if possible. But I have seen this quite a lot also in other projects. Is this currently the way to go? Is there a global registry or is the key just for us?

runtime/integration-tests/src/xcm/setup.rs
```diff
@@ -20,7 +20,9 @@ pub const ALICE: [u8; 32] = [4u8; 32];
 pub const BOB: [u8; 32] = [5u8; 32];
 
 /// Parachain Ids
+pub const PARA_ID_DEVELOPMENT: u32 = 2088;
```
mustermeiszer, line 23: The most straightforward would probably be to just import the `crate::chain::centrifuge::altair::PARA_ID`

mustermeiszer (approved):
THis looks fine from a first view. But I am pretty behind on the xcm stuff. Would be nice, if we could have a meeting explaining the current setup. WDYT @NunoAlexandre ?

runtime/integration-tests/src/xcm/setup.rs
NunoAlexandre: Unfortunately that doesn't work since these tests currently can only run within the `runtime-development` flag. Once we add XCM to altair (next step), we can make that switch. wdyt?

runtime/development/src/lib.rs
NunoAlexandre: Do you have a link to that resource? As far as I know, Multilocations are canonical, so in some places such as the traders and the `CurrencyIdConvert` the specific para id and general key need to be used.

runtime/common/src/lib.rs
NunoAlexandre: A token is identified as a pair of `(ParaId, GeneralKey)`. The `ID` is karura's Para Id in Kusama and `KUSD_KEY` is the general key value of KSUD. KSUD in the network is identified as `(200, [0, 129])`.

runtime/development/src/lib.rs
NunoAlexandre: `ToTreasury` is called by the `Trader` rules. Is the pair is not convertible, it should have failed before reaching that point, since that would mean that this chain didn't know about that token neither how to charge fees for it. Let me know if that makes sense!

NunoAlexandre:
> THis looks fine from a first view. But I am pretty behind on the xcm stuff. Would be nice, if we could have a meeting explaining the current setup. WDYT @NunoAlexandre ?

I think that's a great idea :)

runtime/development/src/lib.rs
mustermeiszer: https://github.com/paritytech/xcm-format -> search for GeneralKey

NunoAlexandre: Thanks for the link. Yeah, those docs are pretty abstract. This setup is similar to a lot of projects I have looked at, namely Acala.

>  Is there a global registry or is the key just for us?

There is no "global registry" per se, but rather a global knowledge of these keys. For instance, for Acala to support AIR, we need to tell them our para id and AIR's general key.

PR #769, opened 2022-04-12
centrifuge: Add Crowload Claim and Reward pallets

### LINKED ISSUE #627: PRTU3 - Add Crowdloan Claim Pallets
This Upgrade enabled Crowdloan Claiming.

Since after the snapshot the network will be populated and access to CFG is easier, we should remove the unsigned extrinsic logic and let the sender pay for the fees. Of course sender can still claim on behalf of anyone else as long as can verify its identity as before.

NunoAlexandre:
Closes #627


runtime/centrifuge/src/lib.rs
```diff
@@ -719,6 +719,38 @@ impl pallet_claims::Config for Runtime {
...
+	type Event = Event;
+	type PalletId = CrowdloanRewardPalletId;
+	type AdminOrigin = EnsureRootOr<HalfOfCouncil>;
+	type WeightInfo = weights::pallet_crowdloan_reward::SubstrateWeight<Self>;
+}
+
+// Parameterize crowdloan claim pallet
+parameter_types! {
+	pub const CrowdloanClaimPalletId: PalletId = PalletId(*b"cc/claim");
+	pub const ClaimTransactionPriority: TransactionPriority = TransactionPriority::max_value();
```
mikiquantum, line 738: I think this is not needed anymore?

runtime/centrifuge/src/lib.rs
```diff
@@ -719,6 +719,38 @@ impl pallet_claims::Config for Runtime {
...
+	pub const ClaimTransactionLongevity: u32 = 64;
```
mikiquantum, line 739: Same as above

mikiquantum (approved):
LGTM, just a minor comment.

runtime/centrifuge/src/lib.rs
NunoAlexandre: My bad, fixed! Thanks!

PR #770, opened 2022-04-12
Fix missing matching arms

NunoAlexandre:


pallets/restricted-tokens/src/benchmarking.rs
```diff
@@ -164,10 +164,7 @@ where
 	// NOTE: We do not want to be CurrencyId::Native to be used here
 	let max_variants = match CurrencyId::Native {
 		CurrencyId::Native => 2u32,
```
NunoAlexandre, line 166: @mustermeiszer that is this `2u32` value meant to represent? I'd think by the context that it is the number of `CurrencyId` variants but it is this value is `2` when we already had 3 variants 🤔

mustermeiszer:
@NunoAlexandre with the current fix, will you add the matching call for the other variants too? If not, I would propose that we remove this function and simply always return the same variant. WDYT?

PR #773, opened 2022-04-13
altair: Add XCM

NunoAlexandre:
We add XCM to Altair, supporting the transfer of `AIR`, `KSM`, and `KUSD`.

The configuration is a pruned copy of the one used at the. `development` runtime, with all the possible config impls placed in a `xcm.rs` file.

Given that token identification in XCM is canonical, we now have a `parachains` mod with both `altair` and `karura` IDs and respective key ids, which is also clearer than using `<currencyIdVariant>.encode()`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [master, parachain]
+    branches: [master, parachain, altair/xcm]
   pull_request:
 name: Tests
 jobs:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [master, parachain, altair/xcm]
+    branches: [master, parachain]
   pull_request:
 name: Tests
 jobs:
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -243,7 +243,7 @@ pub mod parachains {
 
 	pub mod altair {
 		pub const ID: u32 = 2088;
-		pub const AIR_KEY: &[u8] = &[0];
+		pub const AIR_KEY: &[u8] = &[0, 1];
 	}
 }
 
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -240,6 +240,11 @@ pub mod parachains {
 		pub const ID: u32 = 2000;
 		pub const KUSD_KEY: &[u8] = &[0, 129];
 	}
+
+	pub mod altair {
+		pub const ID: u32 = 2088;
+		pub const AIR_KEY: &[u8] = &[0];
+	}
 }
 
 pub mod xcm_fees {
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -235,6 +235,13 @@ pub mod constants {
 	}
 }
 
+/// Listing of parachains we integrate with.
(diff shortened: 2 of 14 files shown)


runtime/integration-tests/src/xcm/xcm_transfers.rs
```diff
@@ -410,36 +253,25 @@ fn currency_id_convert_air() {
...
-
-// The fee associated with transferring Native tokens
-fn usd_fee() -> Balance {
-	let (_asset, fee) = UsdPerSecond::get();
+	let (_asset, fee) = AirPerSecond::get();
 	// We divide the fee to align its unit and multiply by 4 as that seems to be the unit of
 	// time the transfers take.
 	// NOTE: it is possible that in different machines this value may differ. We shall see.
-	fee.div_euclid(10_000) * 4
+	fee.div_euclid(10_000) * 8
```
NunoAlexandre, line 260: This value doubled after the doubling of `UnitWeightCost` (was `100m` in development, now `200m` in altair)

NunoAlexandre (commented):
notes

runtime/integration-tests/Cargo.toml
```diff
@@ -82,6 +82,6 @@ xcm-executor = { git = "https://github.com/paritytech/polkadot", default-feature
 default = [ "runtime-development" ]
 runtime-benchmarks = [ "default", "development-runtime/runtime-benchmarks", "frame-benchmarking/runtime-benchmarks"]
 # TODO(nuno): Remove "kusama-runtime" from `runtime-development` once XCM is available on Altair
-runtime-development = ["rococo-runtime", "development-runtime", "kusama-runtime"]
+runtime-development = ["rococo-runtime", "development-runtime", "kusama-runtime", "altair-runtime"]
```
NunoAlexandre, line 85: Blocked by https://github.com/centrifuge/centrifuge-chain/issues/774

.github/workflows/tests.yml
```diff
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [master, parachain]
+    branches: [master, parachain, altair/xcm]
```
NunoAlexandre, line 3: Temporary change to have the CI build a docker image for this revision.

runtime/common/src/lib.rs
```diff
@@ -240,6 +240,11 @@ pub mod parachains {
 		pub const ID: u32 = 2000;
 		pub const KUSD_KEY: &[u8] = &[0, 129];
 	}
+
+	pub mod altair {
+		pub const ID: u32 = 2088;
+		pub const AIR_KEY: &[u8] = &[0, 1];
```
mikiquantum, line 246: What is the purpose of the AIR_KEY? It uniquely identifies a chain within the local knowledge base?

mikiquantum (approved):
LGTM, granted I am not super familiar with this yet, it looks good. 
I guess we will update the runtime version later?

runtime/common/src/lib.rs
NunoAlexandre: Good question! The idea is that chains and tokens IDs in XCM are canonical, so having those for our own chain as well helps with clarity by having it also listed alongside other chains we integrate with.

The AIR_KEY value itself is arbitrary but it must be immutable as it's used as the GeneralKey value when we represent the token in XCM MultiLocation. In development, we were creating that value by doing CurrencyId::Native::encode() but that may change silently if we change the order of variants in that enum type.

PR #775, opened 2022-04-14
Tests pools - Healthy pricing behaviour

mustermeiszer:
Tests for ensuring prices of tranches tokens are correct (in a healthy state without defaults)

### DIFF
--- a/pallets/pools/src/tranche.rs
+++ b/pallets/pools/src/tranche.rs
@@ -179,7 +179,7 @@ where
 	}
 
 	pub fn accrue(&mut self, now: Moment) -> DispatchResult {
-		let delta = now - self.last_updated_interest;
+		let delta = now.saturating_sub(self.last_updated_interest);
 		let interest = self.interest_rate_per_sec();
 		// NOTE: `checked_pow` can return 1 for 0^0 which is fine
 		//       for us, as we simply have the same debt if this happens
--- a/runtime/development/src/lib.rs
+++ b/runtime/development/src/lib.rs
@@ -53,7 +53,10 @@ use xcm_executor::XcmExecutor;
 use common_traits::Permissions as PermissionsT;
 use common_traits::PreConditions;
 pub use common_types::CurrencyId;
-use common_types::{PermissionRoles, PoolRole, TimeProvider, UNION};
+use common_types::{
+	PermissionRoles, PermissionScope, PermissionedCurrencyRole, PoolId, PoolRole, Role,
+	TimeProvider, UNION,
+};
 use pallet_anchors::AnchorData;
 use pallet_restricted_tokens::{
 	FungibleInspectPassthrough, FungiblesInspectPassthrough, TransferDetails,
@@ -855,7 +858,10 @@ pub struct PoolCurrency;
 impl Contains<CurrencyId> for PoolCurrency {
 	fn contains(id: &CurrencyId) -> bool {
 		match id {
-			CurrencyId::Tranche(_, _) | CurrencyId::Native | CurrencyId::KSM => false,
+			CurrencyId::Tranche(_, _)
+			| CurrencyId::Native
+			| CurrencyId::KSM
+			| CurrencyId::Permissioned(_) => false,
 			CurrencyId::Usd | CurrencyId::KUSD => true,
 		}
 	}
@@ -979,6 +985,7 @@ impl pallet_loans::Config for Runtime {
 	type Time = Timestamp;
 	type LoansPalletId = LoansPalletId;
 	type Pool = Pools;
+	type CurrencyId = CurrencyId;
 	type Permission = Permissions;
 	type WeightInfo = weights::pallet_loans::SubstrateWeight<Self>;
 	type MaxLoansPerPool = MaxLoansPerPool;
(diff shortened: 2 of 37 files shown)


runtime/development/src/lib.rs
```diff
@@ -836,7 +836,7 @@ impl pallet_pools::Config for Runtime {
 	type TrancheId = TrancheId;
 	type EpochId = u32;
 	type CurrencyId = CurrencyId;
-	type Tokens = OrmlTokens;
+	type Tokens = Tokens;
```
mustermeiszer, line 839: This should be tokens. Otherwise, one can not use a native currency in the pools.

mustermeiszer:
I am sure I won't be able to close the other defined test-cases. But I wanted to note them for the future. Is this alright for you? Otherwise, we could move them to issues.

offerijns:
> I am sure I won't be able to close the other defined test-cases. But I wanted to note them for the future. Is this alright for you? Otherwise, we could move them to issues.
I think it's good to already add them here as placeholders, and then we create a single issue to track actually implementing these.

runtime/integration-tests/src/pools/token_prices_normal.rs
```diff
@@ -0,0 +1,209 @@
...
+	let manager = env::task_manager(Handle::current());
+	let mut env = {
+		let mut genesis = Storage::default();
+		genesis::default_balances::<Runtime>(&mut genesis);
+		env::test_env_with_centrifuge_storage::<BLOCK_TIME>(&manager, genesis)
+	};
+
+	let mut nft_manager = NftManager::new();
+	let pool_id = 0u64;
+	let loan_amount = 10_000 * DECIMAL_BASE_12;
```
offerijns, line 55:
```suggestion
	let loan_value = 10_000 * DECIMAL_BASE_12;
```

runtime/integration-tests/src/pools/token_prices_normal.rs
```diff
@@ -0,0 +1,209 @@
...
+	let investment = 5_000 * DECIMAL_BASE_12;
+	let maturity = time::date(90 * SECONDS_PER_DAY);
+
+	env::run!(
+		env,
+		Chain::Para(PARA_ID),
+		Call,
+		ChainState::PoolEmpty,
+		Keyring::Admin => default_pool_calls(Keyring::Admin.into(), pool_id, &mut nft_manager),
+			issue_default_loan(
```
offerijns, line 68: I missed this in the other PR, but maybe we should rename this to 
```suggestion
			issue_bullet_loan(
```

Once we have many loan types, it won't be clear what the default one is.

runtime/integration-tests/src/pools/token_prices_normal.rs
```diff
@@ -0,0 +1,209 @@
...
+		Event::Pools(pallet_pools::Event::EpochClosed(id, ..)) if [id == pool_id],
+		Event::Loans(pallet_loans::Event::Borrowed(id, loan_id, amount))
+			if [id == pool_id && loan_id == loan_id && amount == borrow_amount],
+	);
+
+	env::pass_n(&mut env, 30 * time::blocks_per_day::<BLOCK_TIME>());
+
+	let token_prices = env
+		.with_state(Chain::Para(PARA_ID), || {
+			pools::with_ext::get_tranche_prices(pool_id)
```
offerijns, line 140: I didn't understand originally why the token price had already gone up here, but realized later on that you're calling `update_nav` within `get_tranche_prices`. Maybe overly explicit, but wdyt about calling this `update_nav_and_get_tranche_prices`?

runtime/integration-tests/src/pools/utils/loans.rs
```diff
@@ -243,3 +246,42 @@ pub fn mint_nft_call(class: ClassId, instance: InstanceId, owner: AccountId) ->
...
+
+/// Calculates the final debt for a given
+/// principal amount, rate and period.
+/// Compounding happens every second.
+///
+/// Rate should be in percent. E.g. 15 -> 15%APR
+///
+/// Logic: n = second-per-year
+/// * principal * (1 + r/n)^(end-start)
+pub fn calculate_debt(principal: Balance, rate: u64, start: Moment, end: Moment) -> Balance {
```
offerijns, line 262: I think we should try to keep using the same terminology everywhere. In the `loans` pallet we use this as naming now:

```suggestion
pub fn calculate_expected_repayment_amount(borrowed_amount: Balance, interest_rate_per_sec: u64, origination_date: Moment, maturity_date: Moment) -> Balance {
```

pallets/pools/src/tranche.rs
```diff
@@ -179,7 +179,7 @@ where
 	}
 
 	pub fn accrue(&mut self, now: Moment) -> DispatchResult {
-		let delta = now - self.last_updated_interest;
+		let delta = now.saturating_sub(self.last_updated_interest);
```
offerijns, line 182: 💯

runtime/integration-tests/src/pools/token_prices_normal.rs
mustermeiszer: Good thought! Will do.

runtime/integration-tests/src/pools/token_prices_normal.rs
mustermeiszer: Yes, agree! will do that.

mustermeiszer:
@NunoAlexandre Was there lately not a caching issue? Because the failing integration test runs through just fine on my machine...

offerijns (approved):

PR #776, opened 2022-04-16
Interface change pallet-permissions

mustermeiszer:
This PR is a proposal for changing two things from the pallet-permissions.
1. `add`/`remove` extrinsics take the `with_role` as an `Option<Role>`
2. Checks for whether an editor is allowed to edit permissions now also passes the `who` (i.e. to whom permissions shall be added/from whom permissions shall be taken)

### DIFF
--- a/pallets/permissions/src/lib.rs
+++ b/pallets/permissions/src/lib.rs
@@ -30,8 +30,16 @@ enum Who {
 	Editor,
 }
 
-use common_traits::{Permissions, Properties};
-use frame_support::traits::Contains;
+/// Informs preconditions about the edit parameters
+pub struct EditEffects<AccountId, Role, Location> {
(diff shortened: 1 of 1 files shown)


mustermeiszer:
@offerijns and @branan I had this lying around after we had a talk about the permissions pallet in a call with jeroen. I think it makes sense to especially include the account in the checks that receives permissions. 

Anyways. Its just a draft and I wanted to drop this from my local machine here. Feel also free to close it, if we don't wanna adapt to this. 

Tests and runtimes not yet adapted

mustermeiszer:
Thanks for checking @NunoAlexandre. I would like to bring it in. 

@offerijns Do you think it is worth bringing in? I would like this as a clean-up. But no strong opinions.

PR #778, opened 2022-04-18
Extend pool and loan pallet events

offerijns:
- Extend order events to include old & new order amounts
- Add event on tranche rebalancing
- Extend loan pricing event with pricing info

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -431,7 +431,12 @@ pub mod pallet {
 			// ensure sender has the pricing admin role in the pool
 			ensure_role!(pool_id, origin, PoolRole::PricingAdmin);
 			Self::price_loan(pool_id, loan_id, interest_rate_per_sec, loan_type)?;
-			Self::deposit_event(Event::<T>::Priced(pool_id, loan_id));
+			Self::deposit_event(Event::<T>::Priced(
+				pool_id,
+				loan_id,
+				interest_rate_per_sec,
+				loan_type,
+			));
 			Ok(())
 		}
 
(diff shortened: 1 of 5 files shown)


pallets/pools/src/lib.rs
```diff
@@ -701,12 +707,13 @@ pub mod pallet {
...
-			amount: T::Balance,
+			new_order: T::Balance,
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			let tranche_id =
-				Pool::<T>::try_mutate(pool_id, |pool| -> Result<T::TrancheId, DispatchError> {
+			let (tranche_id, old_order) = Pool::<T>::try_mutate(
+				pool_id,
+				|pool| -> Result<(T::TrancheId, T::Balance), DispatchError> {
```
NunoAlexandre, line 716: Do you think we could move these mutations to a separate function? That would keep the high-level extrinsic more declarative, easier to follow and less noisy.

offerijns: I think in this case, we already have sufficient separation of code. This method now calls into `do_update_invest_order`, which on its own calls into `update_order_amount`. I think adding another level will just make it harder to understand.

NunoAlexandre: Fair 💪🏻 I was also reviewing this on the train with a smaller screen so everything looked more packed and harder to follow.

pallets/loans/src/lib.rs
```diff
@@ -200,8 +200,13 @@ pub mod pallet {
...
 		Closed(PoolIdOf<T>, T::LoanId, AssetOf<T>),
-		/// A loan was priced. [pool, loan]
-		Priced(PoolIdOf<T>, T::LoanId),
+		/// A loan was priced. [pool, loan, interest_rate_per_sec, loan_type]
+		Priced(
+			PoolIdOf<T>,
+			T::LoanId,
+			T::Rate,
+			LoanType<T::Rate, T::Amount>,
+		),
```
mustermeiszer, line 209: Nice!

mustermeiszer (approved):
LGTM!

Issue #774, opened 2022-04-14
discussion: Rethink integration-tests setup


NunoAlexandre:
### Context

The new setup, introduced [here](https://github.com/centrifuge/centrifuge-chain/pull/739/), brought a structure that brings on a few assumptions:

1. Mutually exclusive runtimes - For all integration tests, only _one_ of our runtimes is to be used: either `altair` _or_ `centrifuge` _or_ `development`

2. All runtimes can be tested with interchangeably - The project should compile using any of the runtimes.

``` rust
/// Re-exports the correct runtimes that we run the integration tests with
/// This allows all other modules to use the import of `crate::chain::{...}`
/// in order to get the right stuff from the respective runtime.
mod chain {
	pub mod centrifuge {
		#[cfg(feature = "runtime-altair")]
		pub use altair::*;
		#[cfg(feature = "runtime-centrifuge")]
		pub use centrifuge::*;
		#[cfg(feature = "runtime-development")]
		pub use development::*;

		#[cfg(feature = "runtime-centrifuge")]
		pub mod centrifuge {
			pub use centrifuge_runtime::*;
			pub const PARA_ID: u32 = 2031;
		}

		#[cfg(feature = "runtime-altair")]
		pub mod altair {
			pub use altair_runtime::*;
			pub const PARA_ID: u32 = 2088;
		}

		#[cfg(feature = "runtime-development")]
		pub mod development {
			pub use development_runtime::*;
			pub const PARA_ID: u32 = 2000;
		}
	}

	pub mod relay {
		#[cfg(feature = "runtime-altair")]
		pub use kusama_runtime::*;
		#[cfg(feature = "runtime-centrifuge")]
		pub use polkadot_runtime::*;
		#[cfg(feature = "runtime-development")]
		pub use rococo_runtime::*;
	}
}
```

### Problems

I have been facing some problems with this new setup and the assumptions it forces upon:

1. In the future, we want the flexibility to tests cross-runtime XCM transfers. Assumption 1. above makes this impossible at the moment.

2. More important now, XCM tests are _NOT_ interchangeable - These tests require importing specific objects from a specific parachain and relaychain runtimes and they are meant to test a very specific runtime. For instance, `centrifuge-runtime` does NOT support XCM and thus any of the related imports, if compiled against the new `runtime-centrifuge` flag, will fail to resolve.

[...]

PR #779, opened 2022-04-19
integration-tests: Use empty feature flags

NunoAlexandre:
This is an alternative to proposal 1 in #774, where we keep the feature flags to allow us to keep the `chain` abstraction and thus be able to run the same tests against different runtimes while still being able to use any runtime. With this proposal, instead of following the proposal 1 by the letter, we also don't need to run the integration tests with all the different flags to make sure that all the tests are being run.


mustermeiszer:
> With this proposal, instead of following the proposal 1 by the letter, we also don't need to run the integration tests with all the different flags to make sure that all the tests are being run.

I don't get this?

NunoAlexandre:
> > With this proposal, instead of following the proposal 1 by the letter, we also don't need to run the integration tests with all the different flags to make sure that all the tests are being run.
> 
> I don't get this?

wow, what the hell did I write. Sorry! I meant to say that proposal 1 still locked us into having only one runtime available, whereas with this proposal where we can both keep the `chain` abstraction and have all the other runtimes available.

PR #781, opened 2022-04-19
pallet-pools: Fix-rebalancing and introduce losses and recovery of losses

mustermeiszer:
This PR is a draft-proposal that does
* Fix rebalancing issues when NAV drops
* Introduce losses 
* Introduce possibility to recover these losses and finally recover the pool.

Logic is based on this spec here: https://centrifuge.hackmd.io/w4MrddApTQykzBA8xh18RQ?both

### Notes
This PR does contain this logic but the pool will never reach this state as we currently simply block a pool once a tranche-price has gone to zero. Hence, this must be enabled afterwards. 

The pool-state PR  #695 was the PR that first brought up the problems with NAV drops and rebalancing and is a good starting-point for introducing the possibility for such states after this PR here.

### DIFF
--- a/pallets/pools/src/lib.rs
+++ b/pallets/pools/src/lib.rs
@@ -902,12 +902,70 @@ pub mod pallet {
 
 				pool.start_next_epoch(now)?;
 
+				let mut pool_value = total_assets;
+				// Update the tranches here and note losses and new expected value (i.e. debt)
+				pool.tranches.combine_mut_non_residual_top(|tranche| {
+					if tranche.tranche_type == TrancheType::Residual {
+						tranche.debt = pool_value;
+					} else {
+						tranche.accrue(now)?;
+
+						if tranche.expected_assets()? > pool_value {
+							// This arm matches if the pool_value can only partially fill the debt of
+							// a tranche
+							if tranche.debt > pool_value {
+								// Losses increase by what the diff between current value and expected value
+								tranche.loss = tranche
+									.loss
+									.saturating_add(tranche.debt.saturating_sub(pool_value));
+								// We assume the pool will still be able to generate enough value for
+								// satisfying the tranche. This boils down to the assumption of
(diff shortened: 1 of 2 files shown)


mustermeiszer:
This touches core calculations for a pool a needs EXTENSIVE review!

pallets/pools/src/tranche.rs
```diff
@@ -777,8 +799,10 @@ where
 		let mut prices = self.combine_mut_non_residual_top(|tranche| {
 			let total_issuance = Tokens::total_issuance(tranche.currency);
 
-			if pool_is_zero || total_issuance == Zero::zero() {
-				Ok(One::one())
+			if total_issuance == Zero::zero() {
+				Ok(BalanceRatio::one())
+			} else if pool_is_zero {
+				Ok(BalanceRatio::zero())
```
mustermeiszer, line 805: If a pool is zero after passing the issuance is not zero clause, this means we have lost all assets on a pool that had value. Hence, we should make the tokens-price zero.

pallets/pools/src/lib.rs
```diff
@@ -902,12 +902,71 @@ pub mod pallet {
...
+								tranche.debt = pool_value;
+								tranche.reserve = Zero::zero();
+								pool_value = Zero::zero();
+
+							// This arm matches if the new pool_value can satisfy the debt and the reserve of tranche
+							// This favors debt over reserve as we want as much value to accrue interest for a tranche
+							} else if tranche.debt.saturating_add(tranche.reserve) > pool_value {
+								let loss = pool_value
+									.saturating_sub(tranche.debt.saturating_add(tranche.reserve));
+								// Losses are partially re-covered
```
offerijns, line 937: I think this comment is misleading, in this case there's no recovery of losses, right?

```suggestion
								// Losses are partially increased```

mustermeiszer:
Closing in favor of #695

PR #784, opened 2022-04-20
Feature/pallet investments

mustermeiszer:
The investment palelt that should take care of:

- Providing orders for assets
- allowing users to collect these orders

### Implementation Note
* The `AssetManager` implementation should implement for a specific asset, while the `AssetId` in the pallet-investemtns can get an enum of 
```rust
pub enum AssetId<PoolId, TrancheId>{
    Pool(PoolId, TrancheLoc<TrancheId>)
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6252,6 +6252,34 @@ dependencies = [
 ]
 
 [[package]]
+name = "pallet-investments"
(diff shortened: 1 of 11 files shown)


mustermeiszer:
Closing in favor of #909

PR #785, opened 2022-04-21
Pass bootstrap council members

NunoAlexandre:


src/chain_spec.rs
```diff
@@ -15,6 +15,7 @@
 // along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
 
 use cumulus_primitives_core::ParaId;
+use frame_benchmarking::frame_support::metadata::StorageEntryModifier::Default;
```
mikiquantum, line 18: Is this the expected import?

mikiquantum (approved):
Just minor comment, LGTM

src/chain_spec.rs
NunoAlexandre: You are the best 👍

PR #796, opened 2022-04-27
Additional RPC-Methods and Runtime-Apis

### LINKED ISSUE #747: Additional Runtime APIs and RPC Methods
This is a tracking issue for additional runtime apis we want to provide and the respective rpc methods if those methods should be exposed to the outer world. 

In most cases, we will want both API & rpc as we have no real need for runtime apis that only the node internally uses currently. (Might change once we have ZKPs for which nodes need to access the runtime).

## APIs & RPC
- [x] #750 
- [x] #749 
- [ ] #748 
- [x] Pools Data: Retrieving tranche ids from indexes, 
- [x] #793

### LINKED ISSUE #748: EpochSolution Inspection RuntimeApi and RPC Method
The api should provide the possibilities to call a node whether a solution is valid for given pool and how the respective score of this solution would be.

## Implementation Proposal
```rust
#[rpc]
pub trait EpochSolutionApi {
	#[rpc(name = "pools_inspectSolution")]
	fn inspect_epoch_solution(&self, id: T::PoolId, solution: Vec<TrancheSolution>) -> Result<EpochSolution<Balance>>;
}

decl_runtime_apis! {
    pub trait EpochSolutionApi {
	 fn inspect_epoch_solution(id: Hash) -> Option<EpochSolution<Balance>>;
    }
}
```

### LINKED ISSUE #749: TokenPrice RuntimeApi and RPC Method
The api should provide the possibilities to call a node to retrieve the latest available token-price. This is a little tricky as calculating token-prices generally involves a "more" complex calculation. 
It would be nice to have this anyways and this feature will most likely need us to provide a separate method for the pools pallet to achieve it.

* Should NOT fail when NAV is too old

## Implementation Proposal
```rust
#[rpc]
pub trait TokenTranchePriceApi {
	#[rpc(name = "pools_tokenPrice")]
	fn tranche_token_price(&self, id: T::PoolId, tranche: TrancheLoc<T::TrancheId>) 
	    -> Result<BalanceRatio>;
}

decl_runtime_apis! {
    pub trait TokenTranchePriceApi {
	 fn tranche_token_price(id: T::PoolId, tranche: TrancheLoc<T::TrancheId>) -> Option<BalanceRatio>;
    }
}
```

### LINKED ISSUE #750: NAV RuntimeApi and RPC Method
The api should provide the possibilities to call a node to retrieve the latest available nav. 

* Should NOT fail when NAV is too old

## Implementation Proposal
```rust
#[rpc]
pub trait NavApi {
	#[rpc(name = "loans_nav")]
	fn nav(&self, id: T::PoolId) 
	    -> Result<Balance>;
}

decl_runtime_apis! {
    pub trait NavApi {
	 fn nav(id: T::PoolId) -> Option<Balance>;
    }
}
```

### LINKED ISSUE #793: RPC endpoint for max borrow amount

mustermeiszer:
Implements some needed runtime-apis and corresponding rpc-methods.

Closes #747 
Closes #749
Closes #748


mustermeiszer:
I have not yet tested the rpc endpoints with a local chain. Thought, this would be the easiest I can close and be off help.

runtime/development/src/lib.rs
```diff
@@ -1458,12 +1459,99 @@ impl_runtime_apis! {
...
+		fn pool_value(pool_id: PoolId) -> Option<Balance> {
+			let pool = pallet_pools::Pool::<Runtime>::get(pool_id)?;
+			let nav: Balance = pallet_loans::Pallet::<Runtime>::update_nav_of_pool(pool_id)
+				.ok()
+				.map(|(latest, _)| latest.into())?;
+
+			Some(pool.reserve.total.saturating_add(nav))
+		}
+
+		fn pool_currency(pool_id: PoolId) -> Option<CurrencyId>{
```
offerijns, line 1496: I was thinking we just add RPC methods for anything that we can't easily retrieve from storage, so this could be removed. Or was there a specific reason you also added this?

offerijns (commented):
This is really useful @mustermeiszer!

runtime/development/src/lib.rs
mustermeiszer: Yes. Was just in the Flow. Did not think about it. 

For the tranche currencies. They are also retrievable from storage, but I think it is more to complex with the ids etc. so I would keep those.

mustermeiszer: @offerijns I thought about it a little. And actually I would like to have more rpcs and keep the pool-currency here. I.e. in my opinion it boils down to, abstracting the storage structure from the front-end. Where ever possible. I am not 100% sure how storage fetches work in polkadot-js, but I guess they have a similar overhead. 

Having rpc-abstractions on top of storage would allow us to have a more stable interface for the UI, even if we change logic.

mustermeiszer:
Rpc endpoints seem to work. But apps should dedicatedly test this.

mustermeiszer:
The node currently panics when calling `pool_poolValue`. But I am not sure why. It is failing upon serializing a `u128`. So lets wait to merge this till https://github.com/paritytech/cumulus/issues/1224 is settled.

runtime/common/src/apis/pools.rs
```diff
@@ -0,0 +1,46 @@
...
+	pub trait PoolsApi<PoolId, TrancheId, Balance, Currency, BalanceRatio>
+	where
+		PoolId: Codec,
+		TrancheId: Codec,
+		Balance: Codec,
+		Currency: Codec,
+		BalanceRatio: Codec,
+	{
+		fn pool_value(pool_id: PoolId) -> Option<Balance>;
+
```
offerijns, line 31:
```suggestion
```

runtime/common/src/apis/pools.rs
```diff
@@ -0,0 +1,46 @@
...
+		fn pool_currency(pool_id: PoolId) -> Option<Currency>;
+
+		fn inspect_epoch_solution(pool_id: PoolId, solution: Vec<TrancheSolution>) -> Option<EpochSolution<Balance>>;
+
+		fn tranche_token_price(pool_id: PoolId, tranche: TrancheLoc<TrancheId>) -> Option<BalanceRatio>;
```
offerijns, line 36:
```suggestion
		fn tranche_token_price(pool_id: PoolId, tranche: TrancheId) -> Option<BalanceRatio>;
```

runtime/common/src/apis/pools.rs
```diff
@@ -0,0 +1,46 @@
...
+
+		fn tranche_token_prices(pool_id: PoolId) -> Option<Vec<BalanceRatio>>;
+
+		fn tranche_ids(pool_id: PoolId) -> Option<Vec<TrancheId>>;
+
+		fn tranche_id(pool_id: PoolId, tranche_index: TrancheIndex) -> Option<TrancheId>;
+
+		fn tranche_currency(pool_id: PoolId, tranche_loc: TrancheLoc<TrancheId>) -> Option<Currency>;
```
offerijns, line 44:
```suggestion
		fn tranche_currency(pool_id: PoolId, tranche_loc: TrancheId) -> Option<Currency>;
```

runtime/development/src/lib.rs
```diff
@@ -1513,12 +1515,99 @@ impl_runtime_apis! {
...
+
+		fn max_borrow_amount(id: PoolId, loan_id: InstanceId) -> Option<Balance> {
+			let now = <pallet_timestamp::Pallet::<Runtime> as UnixTime>::now().as_secs();
+			pallet_loans::Loan::<Runtime>::get(id, loan_id)
+				.map(|loan_details| loan_details.max_borrow_amount(now).into())
+		}
+	}
+
+	impl runtime_common::apis::PoolsApi<Block, PoolId, TrancheId, Balance, CurrencyId, Rate> for Runtime {
+		fn pool_value(pool_id: PoolId) -> Option<Balance> {
```
offerijns, line 1543: Should be removed.

runtime/common/src/apis/pools.rs
branan: This was intentional to keep the runtime API as close as possible to the underlying pallets, doing the client transformation at the RPC endpoint. I don't have super strong feelings one way or the other, though

runtime/common/src/apis/pools.rs
```diff
@@ -0,0 +1,46 @@
...
+	where
+		PoolId: Codec,
+		TrancheId: Codec,
+		Balance: Codec,
+		Currency: Codec,
+		BalanceRatio: Codec,
+	{
+		fn pool_value(pool_id: PoolId) -> Option<Balance>;
+
+		fn pool_currency(pool_id: PoolId) -> Option<Currency>;
```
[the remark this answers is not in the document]
branan: Good call, make sense :+1:

branan:
@offerijns addressed your feedback. I left the mapping of `TrancheId` to `TrancheLoc` at the RPC layer as we discussed in Slack. The other changes were straightforward enough.

src/rpc/pools.rs
```diff
@@ -0,0 +1,197 @@
...
+use runtime_common::apis::PoolsApi as PoolsRuntimeApi;
+use sp_api::ProvideRuntimeApi;
+use sp_blockchain::HeaderBackend;
+use sp_runtime::{generic::BlockId, traits::Block as BlockT};
+use std::fmt::Debug;
+use std::sync::Arc;
+
+#[rpc]
+pub trait PoolsApi<PoolId, TrancheId, Balance, Currency, BalanceRatio> {
+	#[rpc(name = "pools_poolCurrency")]
```
offerijns, line 14:
```suggestion
	#[rpc(name = "pools_currency")]
```

offerijns (approved):

PR #803, opened 2022-05-10
altair: Fix and test AIR incoming transfer

NunoAlexandre:
### The problem

This PR fixes an issue that made it impossible to handle incoming `AIR` tokens through XCM.

The issue was caused by the fact that we were expecting to handle AIR in its canonical `MultiLocation` (`parents: 1, X2(Parachain(2088), GeneralKey([0,1])))`) when we should actually be handling its relative representation (`parents: 0, X1(GeneralKey([0,1]))`).

Sender parachains specify AIR in its canonical representation but there is a re-write step that translates the canonical into the relative representation since it realizes that the `Parachain(2088)` value matches the one of the Altair.

### The fix

- we handle AIR in its relative representation, `parents: 0, X1(GeneralKey([0,1]))`
- we cover outgoing and incoming AIR token transfers in the integration tests thoroughly

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [master, parachain]
+    branches: [master, parachain, altair/xcm-air-fix]
   pull_request:
 name: Build
 jobs:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [master, parachain, altair/xcm-air-fix]
+    branches: [master, parachain]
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [master, parachain]
+    branches: [master, parachain, altair/xcm-air-fix]
```
NunoAlexandre, line 3: Remove this once ready

NunoAlexandre:
@mikiquantum this one is ready for your review :)

mikiquantum (approved):
LGTM!

PR #806, opened 2022-05-12
altair: Bump package to 0.10.13

NunoAlexandre:
I forgot to also bump the Altair package version when I bumped the altair spec version so we do we it here.


mikiquantum (approved):
👍

PR #807, opened 2022-05-12
[HOLD] Adapt tx fees on both altair and centrifuge by aprox. factor of 33 and transfer some % to the treasury.

### LINKED ISSUE #809: Implement custom fee logic to send a % of the transaction fees to the treasury
Following up on https://gov.centrifuge.io/t/rfc-increase-transaction-fees-on-centrifuge-chain-by-100x-and-use-the-fees-towards-a-token-burn/3723/13

[More details to come soon]

We can get some inspiration out of the following Polkadot [code](https://github.com/paritytech/polkadot/blob/3161228f2993a4cc30658bf31f00d603b4da75e5/runtime/common/src/impls.rs#L38)

mikiquantum:
Proposal to increment the tx fees:
https://gov.centrifuge.io/t/rfc-increase-transaction-fees-on-centrifuge-chain-by-100x-and-use-the-fees-towards-a-token-burn/3723?u=ash

Comparison of old vs new fee (around 33 times more):
 - Current balance transfer fee on a dev environment: **0.000099871597561621 CFG**
 - New balance transfer fee on same dev environment **0.003287559668446737 CFG** 

Pending before merging:
- [x] Implementation
- [x] Discuss the correct % of fees to the treasury: 80% treasury and 20% author.
- [x] Discuss the correct % to burn in the treasury: 0% by now
- [x] Unitary Tests.
- [x] ~~Integration Tests (blocked until [fudge](https://github.com/centrifuge/fudge) works with Polkadot 9.24).~~ (Can be updated in other PR)
- [x] Analyzing new fee values (Blocked until having the correct %).

Closes #809

### DIFF
--- a/runtime/altair/src/lib.rs
+++ b/runtime/altair/src/lib.rs
@@ -212,7 +212,7 @@ impl pallet_timestamp::Config for Runtime {
 
 // money stuff
 parameter_types! {
-	/// TransactionByteFee is set to 0.01 MicroCFG
+	/// TransactionByteFee is set to 0.01 MicroAIR
 	pub const TransactionByteFee: Balance = 1 * (MICRO_AIR / 100);
 	/// This value increases the priority of `Operational` transactions by adding
 	/// a "virtual tip" that's equal to the `OperationalFeeMultiplier * final_fee`.
--- a/runtime/altair/src/lib.rs
+++ b/runtime/altair/src/lib.rs
@@ -212,7 +212,7 @@ impl pallet_timestamp::Config for Runtime {
 
 // money stuff
 parameter_types! {
-	/// TransactionByteFee is set to 0.01 MicroAIR
+	/// TransactionByteFee is set to 0.01 MicroCFG
 	pub const TransactionByteFee: Balance = 1 * (MICRO_AIR / 100);
 	/// This value increases the priority of `Operational` transactions by adding
 	/// a "virtual tip" that's equal to the `OperationalFeeMultiplier * final_fee`.
--- a/runtime/common/Cargo.toml
+++ b/runtime/common/Cargo.toml
@@ -22,6 +22,7 @@ node-primitives = { git = "https://github.com/paritytech/substrate",  default-fe
 pallet-authorship = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.24" }
 pallet-balances = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.24" }
 pallet-collective = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.24" }
+pallet-treasury = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.24" }
 sp-api = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.24" }
 sp-std = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.24" }
 sp-arithmetic = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.24" }
--- a/runtime/common/Cargo.toml
+++ b/runtime/common/Cargo.toml
@@ -43,6 +43,9 @@ pallet-pools = { path = "../../pallets/pools", default-features = false }
 common-types = { path = "../../libs/common-types", default-features = false }
 pallet-permissions = { path = "../../pallets/permissions", default-features = false }
 
+[dev-dependencies]
+sp-io = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.24" }
+
 [features]
 default = ["std"]
 std = [
--- a/runtime/common/Cargo.toml
+++ b/runtime/common/Cargo.toml
@@ -58,6 +58,7 @@ std = [
     "pallet-authorship/std",
     "pallet-balances/std",
     "pallet-collective/std",
+    "pallet-treasury/std",
     "pallet-permissions/std",
     "serde",
     "sp-core/std",
--- a/runtime/common/src/impls.rs
+++ b/runtime/common/src/impls.rs
@@ -3,9 +3,8 @@
 use super::*;
 use codec::{Decode, Encode, MaxEncodedLen};
 use common_types::CurrencyId;
-use core::marker::PhantomData;
 use frame_support::sp_runtime::app_crypto::sp_core::U256;
-use frame_support::traits::{Currency, OnUnbalanced};
+use frame_support::traits::{Currency, Imbalance, OnUnbalanced};
 use frame_support::weights::{
 	constants::ExtrinsicBaseWeight, WeightToFeeCoefficient, WeightToFeeCoefficients,
 	WeightToFeePolynomial,
@@ -23,17 +22,22 @@ use sp_std::vec::Vec;
 
 common_types::impl_tranche_token!();
 
-pub struct DealWithFees<Config>(PhantomData<Config>);
-pub type NegativeImbalance<Config> =
-	<Balances<Config> as Currency<<Config as SystemConfig>::AccountId>>::NegativeImbalance;
-impl<Config> OnUnbalanced<NegativeImbalance<Config>> for DealWithFees<Config>
+pub type NegativeImbalance<R> =
+	<Balances<R> as Currency<<R as SystemConfig>::AccountId>>::NegativeImbalance;
+
+pub struct DealWithFees<R>(sp_std::marker::PhantomData<R>);
+impl<R> OnUnbalanced<NegativeImbalance<R>> for DealWithFees<R>
 where
-	Config: AuthorshipConfig + BalancesConfig + SystemConfig,
+	R: AuthorshipConfig + BalancesConfig + SystemConfig + pallet_treasury::Config,
+	pallet_treasury::Pallet<R>: OnUnbalanced<NegativeImbalance<R>>,
 {
-	fn on_nonzero_unbalanced(amount: NegativeImbalance<Config>) {
-		if let Some(who) = Authorship::<Config>::author() {
-			Balances::<Config>::resolve_creating(&who, amount);
+	fn on_nonzero_unbalanced(amount: NegativeImbalance<R>) {
+		let split = amount.ration(50, 50);
+		if let Some(who) = Authorship::<R>::author() {
+			Balances::<R>::resolve_creating(&who, split.0);
 		}
+
+		<pallet_treasury::Pallet<R> as OnUnbalanced<_>>::on_unbalanced(split.1);
 	}
 }
 
--- a/runtime/common/src/impls.rs
+++ b/runtime/common/src/impls.rs
@@ -7,7 +7,8 @@ use core::marker::PhantomData;
 use frame_support::sp_runtime::app_crypto::sp_core::U256;
 use frame_support::traits::{Currency, OnUnbalanced};
 use frame_support::weights::{
-	WeightToFeeCoefficient, WeightToFeeCoefficients, WeightToFeePolynomial,
+	constants::ExtrinsicBaseWeight, WeightToFeeCoefficient, WeightToFeeCoefficients,
+	WeightToFeePolynomial,
 };
 use frame_system::pallet::Config as SystemConfig;
 use pallet_authorship::{Config as AuthorshipConfig, Pallet as Authorship};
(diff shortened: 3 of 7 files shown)


runtime/common/src/impls.rs
```diff
@@ -48,46 +49,19 @@ where
...
-/// 	let weight_fee: Balance = weight * weight_coefficient; // 61425000000000
-/// 	let fee: Balance = base_fee + length_fee + weight_fee;
-/// 	assert_eq!(fee, 10172 * (MICRO_AIR / 100));
-/// ```
 pub struct WeightToFee;
 impl WeightToFeePolynomial for WeightToFee {
 	type Balance = Balance;
 
 	fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
+		let p = super::CENTI_CFG;
```
mikiquantum, line 57: Originally we manually set this value and we never got to update it accordingly along with the new substrate releases, so the actual fees were getting behind.

mikiquantum (commented):
Proposal to increment the tx fees: 
https://gov.centrifuge.io/t/rfc-increase-transaction-fees-on-centrifuge-chain-by-100x-and-use-the-fees-towards-a-token-burn/3723?u=ash

branan (approved):
Straightforward enough. Looks like fee calculations were just straight up wrong before?

mikiquantum:
@branan well, they were initially and manually calculated based on some rules that were supposed to work with lower weight sizes, some substrate upgrade changed the weight precision significantly and we never adjusted it.

NunoAlexandre:
@mikiquantum is it an idea to write a test verifying the new tx fees? That way we have a sanity check and will know if a new version of Substrate causes a change.

NunoAlexandre:
@mikiquantum just a heads up that there were some changes to the base fee value(s) introduced in Polkadot v0.9.23. Having tests in place sanity-checking the fee values for some operations will be very valuable at catching this stuff (I caught that change in .23 because of having such tests for the xcm fees)

lemunozm:
The last commit https://github.com/centrifuge/centrifuge-chain/pull/807/commits/a078cb4a72d8c6cc9bf0e1708d83371b8b89a6a3 is a proof of concept of how to split the fee into author and treasury destinations. Things to keep in mind:

1. Currently, the fees imply both the **fee** itself and the associated **tips**. Do we want to handle them as a whole, or should they be split? (i.e. Polkadot do not send to the treasury the tips, 100% of the tips are sent to the author)
2. The rate is hardcoded 50%/50%. We need to evaluate the correct values and how to configure them (maybe a constant like `FEE_TO_TREASURE_RATE` defined by the runtime.
3. Since this is done in `runtime/common`. Development, Altair, and Centrifuge runtimes will modify their behaviour if merged.
4. Not tested yet

mikiquantum:
@lemunozm thanks for the update.
1. Yes we should keep the **tip** as 100% for the block author. 
2. We need to discuss on the actual numbers, but yes I agree that a constant should be used.
3. Yep, since the Treasury now is added in all runtimes it should be applied for all 👍

runtime/altair/src/lib.rs
```diff
@@ -209,7 +209,7 @@ impl pallet_timestamp::Config for Runtime {
 
 // money stuff
 parameter_types! {
-	/// TransactionByteFee is set to 0.01 MicroAIR
```
mikiquantum, line 212: The reference to the token in the Altair runtime is AIR, so prob leave it as is.

runtime/common/src/impls.rs
```diff
@@ -23,16 +20,40 @@ use sp_std::vec::Vec;
 
 common_types::impl_tranche_token!();
```
lemunozm, line 21: I'm wondering why this sentence is a macro (since internally seems to not make use of any macro features). Could it be implemented in `common-types` directly without using `macro_rules!`, as plain Rust code, and later include it as `use common_types::TrancheToken` if needed?

Implementation: https://github.com/centrifuge/centrifuge-chain/blob/97f6ee328f5b0d1f2b08a010401672cbc80dea7b/libs/common-types/src/tokens.rs#L42

branan:
@mikiquantum IIRC that macro is trying to DRY some things up between the real runtimes and mocks, while avoiding Rust's orphan rules.

We could probably eliminate it by moving some things from runtime-common to a primitives crate that pallets can depend on. But that refactor has been on my TODO list for a long while now

runtime/common/src/impls.rs
```diff
@@ -23,16 +20,40 @@ use sp_std::vec::Vec;
...
+	R: pallet_balances::Config + pallet_treasury::Config + pallet_authorship::Config,
+	pallet_treasury::Pallet<R>: OnUnbalanced<NegativeImbalance<R>>,
 {
-	fn on_nonzero_unbalanced(amount: NegativeImbalance<Config>) {
-		if let Some(who) = Authorship::<Config>::author() {
-			Balances::<Config>::resolve_creating(&who, amount);
+	fn on_unbalanceds<B>(mut fees_then_tips: impl Iterator<Item = NegativeImbalance<R>>) {
+		if let Some(fees) = fees_then_tips.next() {
+			let trasury_amount = super::TREASURY_FEE_RATIO * fees.peek();
+			let mut split = fees.split(trasury_amount);
```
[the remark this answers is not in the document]
lemunozm: Changed using `ration`. Maybe longer but more readable of what is happening 👍🏻

runtime/common/src/impls.rs
lemunozm: https://github.com/centrifuge/centrifuge-chain/pull/807#issuecomment-1178485832

runtime/common/src/impls.rs
```diff
@@ -281,3 +278,160 @@ pub mod xcm {
...
+			.unwrap();
+		// We use default for brevity, but you can configure as desired if needed.
+		pallet_balances::GenesisConfig::<Test>::default()
+			.assimilate_storage(&mut t)
+			.unwrap();
+		t.into()
+	}
+
+	#[test]
+	fn test_fees_and_tip_split() {
```
lemunozm, line 417: Although the test itself is only this, the previous code (ported from Polkadot) is necessary to initialize a mock runtime with the required pallets to make the test works. Other `runtime/common` tests could reuse the same runtime (adding more pallets) to test more things.

runtime/common/src/impls.rs
```diff
@@ -44,8 +44,11 @@ where
 {
 	fn on_unbalanceds<B>(mut fees_then_tips: impl Iterator<Item = NegativeImbalance<R>>) {
 		if let Some(fees) = fees_then_tips.next() {
-			let trasury_amount = super::TREASURY_FEE_RATIO * fees.peek();
-			let mut split = fees.split(trasury_amount);
+			// for fees, split the destination
+			let mut split = fees.ration(
+				super::TREASURY_FEE_RATIO.deconstruct(),
```
NunoAlexandre, line 49: can we import `TREASURY_FEE_RATIO` above and avoid referencing it twice with `super::`?

runtime/common/src/impls.rs
```diff
@@ -275,3 +278,160 @@ pub mod xcm {
 		}
 	}
 }
+
+#[cfg(test)]
```
NunoAlexandre, line 282: It's awesome to bring this in but it would be better to have the mock runtime in a `mock.rs` and allow for tests needing/using it to import it, just like we do for example for all the pallets individually.

runtime/common/src/impls.rs
```diff
@@ -275,3 +278,160 @@ pub mod xcm {
...
+			let tip = Balances::issue(20);
+
+			assert_eq!(Balances::free_balance(Treasury::account_id()), 0);
+			assert_eq!(Balances::free_balance(TEST_ACCOUNT), 0);
+
+			DealWithFees::on_unbalanceds(vec![fee, tip].into_iter());
+
+			assert_eq!(
+				Balances::free_balance(Treasury::account_id()),
+				super::TREASURY_FEE_RATIO * 10
```
NunoAlexandre, line 429: Can we use the values defined above? So here:

```rust
super::TREASURY_FEE_RATIO * fee
```

runtime/common/src/impls.rs
```diff
@@ -275,3 +278,160 @@ pub mod xcm {
...
+			);
+			assert_eq!(
+				Balances::free_balance(TEST_ACCOUNT),
+				20 + (Perbill::one() - super::TREASURY_FEE_RATIO) * 10
```
NunoAlexandre, line 433: and here:

``` rust
tip + (Perbill::one() - super::TREASURY_FEE_RATIO) * fee
```

That would make the tests more readable and less error-prone.

runtime/common/Cargo.toml
```diff
@@ -42,6 +43,9 @@ pallet-pools = { path = "../../pallets/pools", default-features = false }
 common-types = { path = "../../libs/common-types", default-features = false }
 pallet-permissions = { path = "../../pallets/permissions", default-features = false }
 
+[dev-dependencies]
+sp-io = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.24" }
+
 [features]
```
NunoAlexandre, line 49: I believe we will need to add `pallet-treasury/std` here or else `runtime-common` won't compile when building altair/centrifuge when this branch is merged and the full CI will attempt to build the runtimes.

runtime/common/src/impls.rs
NunoAlexandre: Also, I think we could actually make this a const/function that we can then apply - as a sort of bimap - against a fee value. Doing that would also dry the duplicated logic we now have in the tests.

runtime/common/src/impls.rs
```diff
@@ -23,16 +20,43 @@ use sp_std::vec::Vec;
...
+pub struct DealWithFees<R>(sp_std::marker::PhantomData<R>);
+impl<R> OnUnbalanced<NegativeImbalance<R>> for DealWithFees<R>
+where
+	R: pallet_balances::Config + pallet_treasury::Config + pallet_authorship::Config,
+	pallet_treasury::Pallet<R>: OnUnbalanced<NegativeImbalance<R>>,
+{
+	fn on_unbalanceds<B>(mut fees_then_tips: impl Iterator<Item = NegativeImbalance<R>>) {
+		if let Some(fees) = fees_then_tips.next() {
+			// for fees, split the destination
+			let mut split = fees.ration(
```
NunoAlexandre, line 48: Does this need to be `mut`?

runtime/common/src/impls.rs
```diff
@@ -23,16 +20,43 @@ use sp_std::vec::Vec;
...
+				(Perbill::one() - super::TREASURY_FEE_RATIO).deconstruct(),
+			);
+			if let Some(tips) = fees_then_tips.next() {
+				// for tips, if any, 100% to author
+				tips.merge_into(&mut split.1);
+			}
+
+			use pallet_treasury::Pallet as Treasury;
+			<Treasury<R> as OnUnbalanced<_>>::on_unbalanced(split.0);
+			<ToAuthor<R> as OnUnbalanced<_>>::on_unbalanced(split.1);
```
NunoAlexandre, line 59: Would it be an idea to name the `split` 0 and 1 parts to make this more readable? Like 

```
let (treasury_amt, author_amt) =  ...

// and then here
```suggestion
			<Treasury<R> as OnUnbalanced<_>>::on_unbalanced(treasury_amt);
			<ToAuthor<R> as OnUnbalanced<_>>::on_unbalanced(author_amt);
```

runtime/common/src/impls.rs
NunoAlexandre: Ah wait, the `.ration(` function is the bimap I was hoping to have. It could still be helpful to dry this into a function and thus avoid the duplication in the tests but leave it up to you.

runtime/common/src/impls.rs
lemunozm: I understand your point, maybe we could make two new variables with the raw value, but I do not know if that makes more complicated the test. If you think it's worth it, I add them 😃

runtime/common/src/impls.rs
lemunozm: Ok, I agree with that 👍🏻

runtime/common/Cargo.toml
lemunozm: The `pallet-treasury/str` is already imported under the `[dependencies]` tag. I do not know if there is something more I'm not seeing regarding it...

runtime/common/src/impls.rs
lemunozm: It is mutated later in line 54

runtime/common/src/impls.rs
lemunozm: I left as Polkadot did, but I agree with that, good idea! 😃

runtime/common/src/impls.rs
lemunozm: Regarding the first point, I agree. In fact, I would avoid also including the `super::*` and import explicitly each constant used in this module, in order to identify better where they come from. Anyway, I do not know if that `super::*` for common types/constants is something used (recommended) by the framework.

lemunozm: Regarding the second point, I do not understand exactly where the duplication is. Anyway, If I use the same "logic" function in both places, code and test, I would never notice a possible error if somebody in the future changes that logic, since the test will still pass.

runtime/common/src/impls.rs
lemunozm: Thinking a little bit more, I think it's worth it, maybe as constants, so I will add them!

runtime/common/src/impls.rs
NunoAlexandre: Sorry if I am overseeing a lot here, but can't we just clone it? And how much work is it to support multiplication - at least in the tests - to reference the defined values instead of duplicating them as literals and thus loosing context and readability?

runtime/common/Cargo.toml
NunoAlexandre: Sorry, I messed up where I wanted to place this comment. I mean that we need to add `pallet-treasury/std` to the `std = ` list/

runtime/common/src/impls.rs
lemunozm: An Imbalance is not `Clone` or `Copy`. And it should be since it's only a "number", it's weird...

runtime/common/Cargo.toml
lemunozm: Ohh, ok that makes much more sense 😃. You're right!

runtime/common/src/impls.rs
lemunozm: Regarding the second point, and for the current code size. I think it could more messy if changed because in the non-test code you should need to call the extracted function and later transform each value with `.deconstruct()`, and the result seems less readable than what we're trying to archive.

runtime/common/src/impls.rs
NunoAlexandre: 🤯  but thanks for having a look @lemunozm!

runtime/common/src/impls.rs
NunoAlexandre: Oke, sounds fair 👍

NunoAlexandre (approved):
This looks really good, fine work! 👍

NunoAlexandre:
@mikiquantum @lemunozm I guess we need to bump the runtime and node versions?

mustermeiszer (commented):
I did not work on this, so I don't wanna approve. But everything looks alright to me. Only got one question where I did not understand what we are calculating.

Thanks, @lemunozm!

runtime/common/src/impls.rs
```diff
@@ -47,46 +72,19 @@ where
...
+		let q = 10 * Balance::from(ExtrinsicBaseWeight::get());
+
 		smallvec!(WeightToFeeCoefficient {
-			coeff_integer: 315000,
-			coeff_frac: Perbill::zero(),
-			negative: false,
 			degree: 1,
+			negative: false,
+			coeff_frac: Perbill::from_rational(p % q, q),
+			coeff_integer: p / q,
```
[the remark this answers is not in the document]
lemunozm: Thanks for the review!

I think the idea was to replicate the [polkadot behavior](https://github.com/paritytech/polkadot/blob/f000f3a992a1f0ae043feeb27b50751d4d124930/runtime/polkadot/constants/src/lib.rs#L84) for the weight to fee conversion in order to get a higher fee. Maybe @mikiquantum that was the author of that commit can explain more deeply.

NunoAlexandre (approved):

PR #808, opened 2022-05-15
Loan storage refactor

offerijns:
Changes:
- `Vec` does not scale well to large datasets, but is way more efficient to read (as it's one storage read) in one go than a `Map`. `update_nav` needs to iterate over all active loans. Therefore, we move the active loan details to a new `BoundedVec` of all active loans. We still store the closed loans in a separate map, mainly for the UI.
- We should bound the number of active loans and number of existing write-off groups.
- A loan can either be non written off, written off permissionlessly, or written off by an admin. When writing off permissionlessly, this should happen according to the predefined write-off schedule. When an admin is writing off an asset, they should be able to choose the write-off % on demand.
- We should add penalty interest rates to write off groups already, otherwise there's a lack of incentive to repay.


branan (changes requested):

branan (changes requested):

pallets/loans/src/functions.rs
```diff
@@ -72,14 +72,15 @@ impl<T: Config> Pallet<T> {
 		pool_id: PoolIdOf<T>,
 		loan_id: T::LoanId,
 		f: F,
-	) -> Result<R, DispatchError>
+	) -> Result<(u32, R), DispatchError>
```
branan, line 75: The point of this is to propagate the number of active loans up and to the top-level extrinsic fn for weight calculations (coming once we have benchmarking updated)

pallets/loans/src/functions.rs
```diff
@@ -422,62 +512,63 @@ impl<T: Config> Pallet<T> {
...
-					loan.interest_rate_per_sec,
-					loan.normalized_debt,
-				)?;
+		// TODO: this will do 1 storage read and up to 1 storage write.
+		// Could we optimize this? Or if not, should we return here whether a storage
+		// write was performed (i.e. the debt wasn't updated yet), so that the end
+		// of the update_nav extrinsic, we can only charge for the number of storage
+		// writes that were actually performed.
+		let debt = T::InterestAccrual::current_debt(
+			active_loan.interest_rate_per_sec,
```
offerijns, line 525: For other reviewers: this should actually use interest rate + penalty interest rate, but we will implement this later: https://github.com/centrifuge/centrifuge-chain/issues/850

pallets/loans/src/lib.rs
```diff
@@ -398,12 +423,15 @@ pub mod pallet {
...
-			let first_borrow = Self::borrow_amount(pool_id, loan_id, owner, amount)?;
+			let (active_count, first_borrow) =
+				Self::borrow_amount(pool_id, loan_id, owner, amount)?;
 			Self::deposit_event(Event::<T>::Borrowed(pool_id, loan_id, amount));
-			match first_borrow {
-				true => Ok(Some(T::WeightInfo::initial_borrow()).into()),
-				false => Ok(Some(T::WeightInfo::further_borrows()).into()),
-			}
+
+			Ok(if first_borrow {
```
NunoAlexandre, line 430: Not a fan of this syntax and `clippy`, which we would like to introduce some time soonish, will complain as well.
My preference here would be an `if first_borrow { ... } else { ... }` or just keep the same `match` approach used before.

pallets/loans/src/lib.rs
```diff
@@ -523,20 +543,20 @@ pub mod pallet {
...
-
-			// this is soft limit, so runtime do not have enough in the deposit.
-			// charge the maximum we have deposited
-			let count = index + 1;
-			if count >= T::MaxWriteOffGroups::get() {
-				return Ok(().into());
-			}
-
-			Ok(Some(Self::write_off_group_weight(count as u64)).into())
+			let count = index.expect("Write offs should always return a valid index") + 1;
```
NunoAlexandre, line 558: Do we really want to `.expect()` here? 🤔

NunoAlexandre: and what if it overflows? I don't want a stale chain 😢

pallets/loans/src/math.rs
```diff
@@ -145,8 +145,8 @@ pub(crate) fn valid_write_off_group<Rate>(
 	maturity_date: Moment,
 	now: Moment,
 	write_off_groups: &[WriteOffGroup<Rate>],
-) -> Result<Option<u32>, DispatchError> {
-	let mut index = None;
+) -> Result<Option<(u32, &WriteOffGroup<Rate>)>, DispatchError> {
```
NunoAlexandre, line 148: Could we have a type alias for `(u32, &WriteOffGroup<Rate>)`? Like `(Index, &WriteOffGroup<Rate>)` or a descriptive alias for the tuple itself.

pallets/loans/src/math.rs
```diff
@@ -356,54 +356,60 @@ mod tests {
...
-			(0, 5, Some(1)),
+			(0, 5, Some((1, &groups[1]))),
 			// now is 6 and less than 7 days, the index is valid
-			(0, 6, Some(2)),
+			(0, 6, Some((2, &groups[2]))),
 			// now is 7 and 8 and less than 9 days, the index is valid
-			(0, 7, Some(5)),
-			(0, 8, Some(5)),
+			(0, 7, Some((5, &groups[5]))),
+			(0, 8, Some((5, &groups[5]))),
```
NunoAlexandre, line 408: Is the index and the groups element always needing to be related or is that just a test convenience?

NunoAlexandre (commented):
Half way through it, leaving my comments so far

pallets/loans/src/lib.rs
```diff
@@ -185,7 +183,27 @@ pub mod pallet {
...
+	>;
+
+	#[pallet::storage]
+	pub(crate) type ClosedLoans<T: Config> = StorageDoubleMap<
+		_,
+		Blake2_128Concat,
+		PoolIdOf<T>,
+		Blake2_128Concat,
+		T::LoanId,
+		ActiveLoanDetailsOf<T>,
```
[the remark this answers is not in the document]
offerijns: We just discussed this, will rename it to `PricedLoanDetails`

pallets/loans/src/lib.rs
branan: Pretty sure clippy complains about match on bools. I can pull the `Ok` into the if, I was just being clever

pallets/loans/src/math.rs
branan: `valid_write_off_group` returns the tuple of `(idx, &group)`, so these do need to match in the tests

pallets/loans/src/lib.rs
branan: It's an expect for sure here. The internal `write_off_loan` is shared by both `admin_write_off` and the regularly `write_off` extrinsic (since most of the logic is identical). The non-admin case always returns an index, and the admin case always does not.

It also will not overflow since the index is capped by `T::MaxWriteOffGroups`

I'll add a better comment here explaining all that

pallets/loans/src/lib.rs
branan: I ended up going a slightly different way with this, but clippy seems to be happy. What do you think?

branan:
@NunoAlexandre @offerijns this is ready for another pass

branan:
UGGGGH the squash really messed up the cross-merging here. I'm gonna have to just flat-rebase this onto parachain

branan:
I'll have to just preemptively squash this to get it to rebase cleanly. When there are no more review comments I'll do so (I've already got the current squash prepared if we like the current state of things)

NunoAlexandre:
> I'll have to just preemptively squash this to get it to rebase cleanly. When there are no more review comments I'll do so (I've already got the current squash prepared if we like the current state of things)

Really recommend https://github.com/sheerun/git-squash for this

pallets/loans/src/lib.rs
```diff
@@ -517,20 +550,21 @@ pub mod pallet {
...
-			// this is soft limit, so runtime do not have enough in the deposit.
-			// charge the maximum we have deposited
-			let count = index + 1;
-			if count >= T::MaxWriteOffGroups::get() {
-				return Ok(().into());
-			}
-
-			Ok(Some(Self::write_off_group_weight(count as u64)).into())
+			// total loops = index+1. This cannot overflow since it is
+			// capped by `MaxWriteOffGroups`
```
NunoAlexandre, line 565: sweet 👌

PR #811, opened 2022-05-19
Drop CurrencyId::Usd in favour of AUSD

NunoAlexandre:
The `CurrencyId::Usd` variant served as a placeholder for a dollar stable coin. With the introduction of the `KUSD` variant, we drop the placeholder `Usd` and use `KUSD` instead.

### DIFF
--- a/libs/common-types/src/tokens.rs
+++ b/libs/common-types/src/tokens.rs
@@ -30,6 +30,9 @@ pub enum CurrencyId {
 
 	/// Karura Dollar
 	KUSD,
+	/// Acala Dollar
+	/// Note: KUSD and AUSD will be merged into a single token, AUSD.
+	AUSD,
 
 	Permissioned(PermissionedCurrency),
 }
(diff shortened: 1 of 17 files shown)


branan (approved):
This looks fine to me, nothing surprising

One thought, though: we'll need to change all this again when Acala finishes the aUSD/kUSD merger - maybe we should preemptively add the aUSD enum entry and use that in all our tests, keeping kUSD just for the altair/karura integration parts?

NunoAlexandre:
@branan thank you, that's a good observation.

Regarding the upcoming merge of kUSD and aUSD, could you or @mikiquantum help me understand how we would handle that in Altair? I assume that we need both the kUSD and aUSD variants and migrate all kUSD balances to aUSD and in the follow up Altair upgrade we could drop the kUSD variant?

If that's the procedure, I agree that we this PR could already introduce aUSD, avoid duplicated work in the near future and have us already half through to work on that migration.

branan:
@NunoAlexandre Yeah, presumably it will be something like that. Hopefully the Acala will have some guidance for us when they get closer to implementing it.

Even before the merger we'll also need aUSD for Centrifuge chain, though.

PR #816, opened 2022-05-23
Disallow updating orders during submission period

branan:
Prevents panics in epoch execution due to the invest/redeem amount
changing during the submission period.

This is a temporary measure until the investments pallet can be
fully integrated.


branan:
We fixed this in #816, though it's kind of hacky. The correct fix is still to integrate the investments pallet. I'm gonna close this for now, since we're tracking that work separately.

PR #817, opened 2022-05-24
Migrate pools pallet to named events

offerijns:


branan (approved):
Nice and straightforward change, and should make the API easier.

This might require frontend changes, we should be sure to sync up with that team.

PR #823, opened 2022-06-01
Fix next loan ID being per pool rather than global

offerijns:
Bug fix: loan IDs in the loans pallet are incremental (1,2,3…) and should be incremental per pool, but were accidentally incremental globally (e.g. if a first pool is created and it has 5 loans, then the first loan in the second pool will have loan ID 6).


NunoAlexandre (approved):
LGTM to me but do we have more context?

PR #830, opened 2022-06-10
claim root hashes value instead of map

mikiquantum:
Potential change (pending agreement) to change root hashes storage from a map to single value.

In addition adding try-runtime to verify upgrades against a live chain data. To execute:
```
cargo build --release --features=try-runtime

RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace ./target/release/centrifuge-chain try-runtime --execution Native --chain centrifuge-local --no-spec-name-check on-runtime-upgrade live --uri wss://fullnode.parachain.centrifuge.io:443
```

### DIFF
--- a/pallets/claims/src/lib.rs
+++ b/pallets/claims/src/lib.rs
@@ -120,6 +120,8 @@ pub use crate::traits::WeightInfo as PalletWeightInfo;
 // Re-export in crate namespace (for runtime construction)
 pub use pallet::*;
 
+mod migration;
+
 // ----------------------------------------------------------------------------
 // Traits and types declaration
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


pallets/claims/src/lib.rs
```diff
@@ -230,11 +230,10 @@ pub mod pallet {
 	pub(super) type ClaimedAmounts<T: Config> =
 		StorageMap<_, Blake2_128Concat, T::AccountId, T::Balance, ValueQuery>;
 
-	/// Map of root hashes that correspond to lists of reward claim amounts per account.
+	/// Root hash that correspond to lists of reward claim amounts per account.
 	#[pallet::storage]
 	#[pallet::getter(fn get_root_hash)]
-	pub(super) type RootHashes<T: Config> =
-		StorageMap<_, Blake2_128Concat, T::Hash, bool, ValueQuery>;
+	pub(super) type RootHash<T: Config> = StorageValue<_, T::Hash, OptionQuery>;
```
mikiquantum, line 236: The old storage is only of size 1, so I think a cleanup migration is an overkill

NunoAlexandre (approved):
Yes 100 times 💪🏻

pallets/claims/src/lib.rs
branan: If it was a difficult migration I'd agree, but since it's a one-liner to nuke the storage I think we should just do it

mikiquantum:
Please @branan @NunoAlexandre have another look with the migration added and the try-runtime support moving forward (for now only on the Centrifuge Runtime and Claims)

branan (approved):
This looks good to me. It ended up being a bit more boilerplate than a one-liner, but I'm glad all the infrastructure is in place now :+1:

Issue #836, opened 2022-06-16
discussion: kUSD -> aUSD merge


NunoAlexandre:
## What

Our first integration with Acala involved supporting XCM transfers of `kUSD` between Altair and Karura. Now, Acala is merging `kUSD` and `aUSD` to simply `aUSD` cross network.

### Adapt XCM config
Now, we convert `kUSD` from and to `MultiLocation` like so:

***currently***
- incoming: `MultiLocation { parents: 1, X2(Parachain(2000), GeneralKey([0, 129])) }` -> `CurrencyId::KUSD`
- outgoing: `CurrencyId::KUSD` -> `MultiLocation { parents: 1, X2(Parachain(2000), GeneralKey([0, 129])) }`

`aUSD` in `Kusama` will use the same `MultiLocation` that was used for `kUSD`, stated above. So, in theory, we just have to change the `CurrencyId` variant we convert said `MultiLocation` from and to:

***becomes***
- incoming: `MultiLocation { parents: 1, X2(Parachain(2000), GeneralKey([0, 129])) }` -> `CurrencyId::AUSD`
- outgoing: `CurrencyId::AUSD` -> `MultiLocation { parents: 1, X2(Parachain(2000), GeneralKey([0, 129])) }`

### Migrate balances

As the XCM config will move to use `AUSD` and not `KUSD`, users that hold balance for `CurrencyId::KUSD` will need to be able to see those funds as AUSD and be able to transfer them through XCM as such.

## Proposal

My idea initially when dropping [`CurrencyId::USD` in favour of `CurrencyId::AUSD`](https://github.com/centrifuge/centrifuge-chain/pull/811/files) as so that we would already have the `CurrencyId` variant we would need to transfer funds to and to migrate to at the XCM config level.

However, now with the introduction of `asset-registry`, we have a new option. So we can choose from:

### Option 1 (using `asset-registry`)

1. Drop `CurrencyId::AUSD`
2. Register `AUSD` as a `CurrencyId::ForeignAsset` in the asset-registry
3. Drop all `KUSD`-related XCM logic
    automatically, `AUSD` XCM transfers will start being handled by the new XCM logic that looks at the `asset-registry`
4. Migrate all `KUSD` balances 
   4.a from `CurrencyId::KUSD` to `CurrencyId::ForeignAsset(<ausd_foreign_asset_id>)`; OR
   4.b we announce the change to the community and let the few KUSD holders run that transfer themselves
[...]

PR #837, opened 2022-06-16
common-types: Test CurrencyId variants encoding

NunoAlexandre:
### Context

<blockquote>
🤯  **PLUS**, I jus realised this, if we choose this option I am pretty certain that we won't need to concern ourselves with the balances migration step because by dropping `CurrencyId::KUSD`, `CurrencyId::AUSD` will automatically be encoded to the value that `CurrencyId::KUSD` was, so the migration is essentially automatically done for free. I need to verify this tho!

In any case, we need to find a way of not breaking the balances when messing around with `CurrencyId` variants. Pehraps we need to implement a custom `Encode`/`Decode` impl for `CurrencyId`.
</blockquote>

_Originally posted by @NunoAlexandre in https://github.com/centrifuge/centrifuge-chain/issues/836#issuecomment-1157606927_

### Purpose

Whatever route we choose to we proceed with on the linked issue, it's important to have a sanity test verifying the values that each `CurrencyId` variant `codec::Encode` to.

This will alert us for any change we make to `CurrencyId` and avoids us from making a serious mistake which would, for instance, mess up with the balances indexing in `OrmlTokens` which I believe (*) encodes `CurrencyId`.

(*) busy verifying this

### DIFF
--- a/libs/common-types/src/tests.rs
+++ b/libs/common-types/src/tests.rs
@@ -200,3 +200,21 @@ fn permission_roles_work() {
 	assert!(!roles.exists(Role::PoolRole(PoolRole::LiquidityAdmin)));
 	assert!(!roles.exists(Role::PoolRole(PoolRole::MemberListAdmin)));
 }
+
+/// Sanity check for every CurrencyId variant's encoding value.
+/// This will stop us from accidentally moving or dropping variants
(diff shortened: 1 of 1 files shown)


offerijns (approved):
Great idea!

PR #845, opened 2022-06-18
Include fast-runtime feature flag with some defaults

mikiquantum:
**Context:**
The feedback loop of testing upgrades locally (or just some operations that require Council or Democracy) is slow. The current alternative to this is to modify the runtime code only in your dev branch to test your code but that comes with potential issues down the line, unwanted commits, extra code changes that shouldn't be required among others.

**Proposal:**
I propose to include this feature flag that we can use for certain constants that usually delay the development cycle. When this the code has been compiled with this flag, the `prod_or_fast` macro developed by parity allows to pass a test value and potentially a value read from the environment at the time of compilation (the latter is the one I used here but I do not see us using it too much)

This way when running the chain locally the `init.sh start-parachain` script builds with that flag, so we can test out altair or centrifuge runtime easily and quickly. 

In addition, I propose to build a docker image as well ready for local testing, without the need to compile again the code. This is particularly useful for other teams that just want to set up a local env quickly.

Two images tags will be pushed:
- `centrifugeio/centrifuge-chain:parachain-$date-$commit`
- `centrifugeio/centrifuge-chain:test-parachain-$date-$commit`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -192,3 +192,10 @@ try-runtime = [
 	"centrifuge-runtime/try-runtime",
 	"try-runtime-cli"
 ]
+
+fast-runtime = [
+	"altair-runtime/fast-runtime",
+	"centrifuge-runtime/fast-runtime"
+]
+
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -197,5 +197,3 @@ fast-runtime = [
 	"altair-runtime/fast-runtime",
 	"centrifuge-runtime/fast-runtime"
 ]
-
-
--- a/runtime/altair/src/lib.rs
+++ b/runtime/altair/src/lib.rs
@@ -29,7 +29,7 @@ use pallet_collective::{EnsureMember, EnsureProportionAtLeast, EnsureProportionM
 pub use pallet_timestamp::Call as TimestampCall;
 pub use pallet_transaction_payment::{CurrencyAdapter, Multiplier, TargetedFeeAdjustment};
 use pallet_transaction_payment_rpc_runtime_api::{FeeDetails, RuntimeDispatchInfo};
-use polkadot_runtime_common::{BlockHashCount, SlowAdjustingFeeUpdate, prod_or_fast};
+use polkadot_runtime_common::{prod_or_fast, BlockHashCount, SlowAdjustingFeeUpdate};
 use scale_info::TypeInfo;
 use sp_api::impl_runtime_apis;
 use sp_core::OpaqueMetadata;
--- a/runtime/altair/src/lib.rs
+++ b/runtime/altair/src/lib.rs
@@ -29,7 +29,7 @@ use pallet_collective::{EnsureMember, EnsureProportionAtLeast, EnsureProportionM
 pub use pallet_timestamp::Call as TimestampCall;
 pub use pallet_transaction_payment::{CurrencyAdapter, Multiplier, TargetedFeeAdjustment};
 use pallet_transaction_payment_rpc_runtime_api::{FeeDetails, RuntimeDispatchInfo};
-use polkadot_runtime_common::{BlockHashCount, SlowAdjustingFeeUpdate};
+use polkadot_runtime_common::{BlockHashCount, SlowAdjustingFeeUpdate, prod_or_fast};
 use scale_info::TypeInfo;
 use sp_api::impl_runtime_apis;
 use sp_core::OpaqueMetadata;
@@ -446,7 +446,7 @@ impl pallet_preimage::Config for Runtime {
 }
 
 parameter_types! {
-	pub const CouncilMotionDuration: BlockNumber = 5 * DAYS;
+	pub CouncilMotionDuration: BlockNumber = prod_or_fast!(5 * DAYS, 1 * MINUTES, "AIR_MOTION_DURATION");
 	pub const CouncilMaxProposals: u32 = 100;
 	pub const CouncilMaxMembers: u32 = 100;
 }
@@ -481,7 +481,7 @@ parameter_types! {
 	pub const CandidacyBond: Balance = 500 * AIR;
 	pub const VotingBond: Balance = 50 * CENTI_AIR;
 	pub const VotingBondBase: Balance = 50 * CENTI_AIR;
-	pub const TermDuration: BlockNumber = 7 * DAYS;
+	pub TermDuration: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "AIR_TERM_DURATION");
 	pub const DesiredMembers: u32 = 9;
 	pub const DesiredRunnersUp: u32 = 9;
 	pub const ElectionsPhragmenModuleId: LockIdentifier = *b"phrelect";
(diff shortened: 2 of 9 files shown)


branan (approved):
Maybe we should name this more generically in case there are other things we want to change based on whether we're in dev or prod? But otherwise it's a nice change and LGTM :+1:

runtime/altair/src/lib.rs
```diff
@@ -524,13 +524,13 @@ impl pallet_elections_phragmen::Config for Runtime {
 }
 
 parameter_types! {
-	pub const LaunchPeriod: BlockNumber = 7 * DAYS;
-	pub const VotingPeriod: BlockNumber = 7 * DAYS;
-	pub const FastTrackVotingPeriod: BlockNumber = 3 * HOURS;
+	pub LaunchPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "AIR_LAUNCH_PERIOD");
+	pub VotingPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "AIR_LAUNCH_PERIOD");
```
NunoAlexandre, line 528:
```suggestion
	pub VotingPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "AIR_VOTING_PERIOD");
```

runtime/altair/src/lib.rs
```diff
@@ -524,13 +524,13 @@ impl pallet_elections_phragmen::Config for Runtime {
...
+	pub FastTrackVotingPeriod: BlockNumber = prod_or_fast!(3 * HOURS, 1 * MINUTES, "AIR_FAST_TRACK_VOTING_PERIOD");
 	pub const InstantAllowed: bool = false;
 	pub const MinimumDeposit: Balance = 500 * AIR;
-	pub const EnactmentPeriod: BlockNumber = 8 * DAYS;
-	pub const CooloffPeriod: BlockNumber = 7 * DAYS;
+	pub EnactmentPeriod: BlockNumber = prod_or_fast!(8 * DAYS, 1 * MINUTES, "AIR_ENACTMENT_PERIOD");
+	pub CooloffPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "AIR_ENACTMENT_PERIOD");
```
NunoAlexandre, line 533:
```suggestion
	pub CooloffPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "AIR_COOLOFF_PERIOD");
```

runtime/centrifuge/src/lib.rs
```diff
@@ -514,13 +514,13 @@ impl pallet_elections_phragmen::Config for Runtime {
...
-	pub const FastTrackVotingPeriod: BlockNumber = 3 * HOURS;
+	pub  LaunchPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "CFG_LAUNCH_PERIOD");
+	pub  VotingPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "CFG_LAUNCH_PERIOD");
+	pub  FastTrackVotingPeriod: BlockNumber = prod_or_fast!(3 * HOURS, 1 * MINUTES, "CFG_FAST_TRACK_VOTING_PERIOD");
 	pub const InstantAllowed: bool = false;
 	pub const MinimumDeposit: Balance = 1000 * CFG;
-	pub const EnactmentPeriod: BlockNumber = 8 * DAYS;
-	pub const CooloffPeriod: BlockNumber = 7 * DAYS;
+	pub  EnactmentPeriod: BlockNumber = prod_or_fast!(8 * DAYS, 1 * MINUTES, "CFG_ENACTMENT_PERIOD");
+	pub  CooloffPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "CFG_ENACTMENT_PERIOD");
```
NunoAlexandre, line 523:
```suggestion
	pub  CooloffPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "CFG_COOLOFF_PERIOD");
```

runtime/centrifuge/src/lib.rs
```diff
@@ -514,13 +514,13 @@ impl pallet_elections_phragmen::Config for Runtime {
 }
 
 parameter_types! {
-	pub const LaunchPeriod: BlockNumber = 7 * DAYS;
-	pub const VotingPeriod: BlockNumber = 7 * DAYS;
-	pub const FastTrackVotingPeriod: BlockNumber = 3 * HOURS;
+	pub  LaunchPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "CFG_LAUNCH_PERIOD");
+	pub  VotingPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "CFG_LAUNCH_PERIOD");
```
NunoAlexandre, line 518:
```suggestion
	pub  VotingPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "CFG_VOTING_PERIOD");
```

Cargo.toml
```diff
@@ -192,3 +192,10 @@ try-runtime = [
 	"centrifuge-runtime/try-runtime",
 	"try-runtime-cli"
 ]
+
+fast-runtime = [
+	"altair-runtime/fast-runtime",
+	"centrifuge-runtime/fast-runtime"
+]
+
+
```
NunoAlexandre, line 201:
```suggestion
```

NunoAlexandre (changes requested):
This looks very neat 👌 TIL on `prod_or_fast`

NunoAlexandre:
@branan 

> Maybe we should name this more generically in case there are other things we want to change based on whether we're in dev or prod? But otherwise it's a nice change and LGTM 👍

The `prod_or_fast` macro operates with the `fast-runtime` flag: 

```
#[macro_export]
macro_rules! prod_or_fast {
	($prod:expr, $test:expr) => {
		if cfg!(feature = "fast-runtime") {
			$test
		} else {
			$prod
		}
	};
	($prod:expr, $test:expr, $env:expr) => {
		if cfg!(feature = "fast-runtime") {
			core::option_env!($env).map(|s| s.parse().ok()).flatten().unwrap_or($test)
		} else {
			$prod
		}
	};
}
```

NunoAlexandre (dismissed):

NunoAlexandre (commented):
Some thoughts here:

- I wouldn't name the build with `features=fast-runtime` enabled as `test` - it's confusing since we already have other things named `test` in the CI and in the code. I'd either call it `fast`, `fast-runtime`, or give it a funnier name like `speedy` or `turbo`.

- Given how long the CI already takes to run, introducing yet another build for this `fast-runtime`-enabled package is yet another thing making it slower. Can we maybe only build this image when the commit is prefixed with, say, `[speedy]` or something like that?

NunoAlexandre (approved):
> There are only two hard things in computer science: cache invalidation, naming things, and off-by-one errors.

Not blocking this 🍻

NunoAlexandre:
> I am concerned about anyone that pulls that image because they think that is a faster version of Centrifuge, lightclient and that can cause support issues down the line when running against the production chains. I would rename to something else if `test` is not agreed on happily but any of the suggestions still imply what my initial concern is based on. Maybe `dev-*` or `dev-only-*` would make more sense.

Yeah I thought the same originally thus the funny name suggestion. Also though of `dev-*` but that's also confusing since it overlaps with the `dev` runtime genesis.

Naming things is hard. I don't know anymore 🙃 Unless if @branan comes with a good suggestion I'd say go ahead with `test` and we iterate over it afterwards if a better name comes up.

PR #833, opened 2022-06-15
Update to Polkadot v0.9.24

NunoAlexandre:
### Forks used

- NunoAlexandre/xcm-simulator [pr](https://github.com/shaunxw/xcm-simulator/pull/32)
- NunoAlexandre/open-runtime [pr](https://github.com/open-web3-stack/open-runtime-module-library/pull/773)

### Important changes

- Move from `jsonrpc` to `jsonrpsee` which changes the way RPC apis/methods are built and composed
- `AccountIdConversion` trait methods renames
- pallet_uniques
   - tl;dr: "class" -> "collection", "instance" -> "item" 
  - ❗ Api breaks (For frontend's special attention)
    - Anything "class" becomes "collection"
    - Anything "instance" becomes "item"

PR #846, opened 2022-06-18
rpc: Ressurections

NunoAlexandre:
We resurrect [#747](https://github.com/centrifuge/centrifuge-chain/pull/796) which was git-reverted in #833 due to lots of conflicts and the need to quickly deliver on that upgrade.

### Changes

- resurrect the new RPC methods that had been added, as well as follow the restructure that it had introduced.
- migrate rpc::pools to be built with `jsonrpsee`
- abstract and clean error building boiler plate code

Note: In e1282d8348b261088786909ad89bd505dffac28d I drop the params being passed to every `invalid_params_error`.  That data is just printing the params passed by the user/client, so there is really no value nor need to bloat the code with all this redundant code as far as I see. Let me know if you feel differently about it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,7 +70,7 @@ checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
  "getrandom 0.2.7",
  "once_cell",
- "version_check 0.9.4",
+ "version_check",
 ]
 
 [[package]]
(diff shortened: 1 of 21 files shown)


NunoAlexandre:
@offerijns @branan can you please have a look at this one?

branan (dismissed):
LGTM :+1:

Thanks for taking this on

branan (approved):

PR #856, opened 2022-07-04
fixup! Fix CI build-runtime

NunoAlexandre:


branan (approved):
👍 at least it was an easy fix.

It's weird that stool builds with std, considering the final runtime is obviously no-std.

PR #861, opened 2022-07-12
Connectors POC

NunoAlexandre:
closes https://github.com/centrifuge/centrifuge-chain/issues/955

### To Do

- [x] Working version of `AddPool` on a testnet
- [x] Unit-test remote call encoding
- [x] Use `xcm_transactor::transact_through_sovereign`
      For our use case, we want the messages to be sent from the sovereign account but the account calling the outer connectors' extrinsic paying for the fees involved in the Transact execution. `transact_through_sovereign` seems to be the best fit here.
- [x] Extend `XcmDomain`
    The drawback here is that it makes it impossible to unit-test the remote call encoding since it would depend on `T::Currency` of sorts
- [x] Clean and simplify
- [ ] Impl `add_tranche` and `transfer` ToDo's
- [x] Run complete test on testnet


pallets/connectors/src/message.rs
```diff
@@ -0,0 +1,32 @@
+use crate::*;
+
+#[derive(Decode, Clone, PartialEq, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub enum Message<PoolId> {
+	AddPool { pool_id: PoolId }, // More to come...
```
offerijns, line 6: We need to make sure the indices of this enum matches the indices in the EVM implementation: https://github.com/centrifuge/connectors/blob/main/src/Messages.sol#L10 There it starts with an `Invalid` variant (copied from the Nomad sample implementation, my assumption is that the reason for this is that a message that is entirely zeros (0x000000...) is not considered a valid message. If the first message is `AddPool` and you add a pool ID of 0, then the full message would actually become 0x00000...).

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+	#[cfg_attr(feature = "std", derive(Debug))]
+	pub struct NomadRouter {
+		forwarding_contract: String, // TODO(nuno): make it a MultiLocation
+	}
+
+	#[derive(Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
+	#[cfg_attr(feature = "std", derive(Debug))]
+	pub struct XCMRouter {
+		multilocations: (), // TODO(nuno): make it a Map<Domain, MultiLocation>
+	}
```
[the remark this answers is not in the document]
NunoAlexandre: Done

pallets/connectors/src/message.rs
```diff
@@ -0,0 +1,277 @@
...
+
+impl<
+		Domain: Encode + Decode,
+		PoolId: Encode + Decode,
+		TrancheId: Encode + Decode,
+		Balance: Encode + Decode,
+		Rate: Encode + Decode,
+	> Encode for Message<Domain, PoolId, TrancheId, Balance, Rate>
+{
+	fn encode(&self) -> Vec<u8> {
```
mustermeiszer, line 76: I think we should only implement `using_encode` as the doc mention. Assuming, this is a value type, and I understand the definition of value vs. allocating type.

> /// Implementations should override `using_encoded` for value types and `encode_to` and `size_hint` for allocating types.
> /// Wrapper types should override all methods.

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,522 @@
...
+
+		/// Build the encoded `ethereum_xcm::transact(eth_tx)` call that should request to execute `contract_call`.
+		///
+		/// * `xcm_domain` - All the necessary info regarding the domain
+		/// where this `ethereum_xcm` call is to be executed
+		/// * `contract_call` - The encoded EVM call calling ConnectorsXcmRouter::handle(msg)
+		pub fn encoded_ethereum_xcm_call(
+			xcm_domain: XcmDomain<CurrencyIdOf<T>>,
+			evm_call: Vec<u8>,
+		) -> Result<Vec<u8>, ()> {
```
NunoAlexandre, line 456: No error here, so: 

```suggestion
		) -> Vec<u8> {
```

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,522 @@
...
+			Self::deposit_event(Event::MessageSent { message, domain });
+
+			Ok(())
+		}
+
+		/// Build the encoded `ethereum_xcm::transact(eth_tx)` call that should request to execute `contract_call`.
+		///
+		/// * `xcm_domain` - All the necessary info regarding the domain
+		/// where this `ethereum_xcm` call is to be executed
+		/// * `contract_call` - The encoded EVM call calling ConnectorsXcmRouter::handle(msg)
```
NunoAlexandre, line 452: Update this

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,522 @@
...
+				},
+			);
+
+			// The complete encoded call payload to be sent in a `Transact` XCM instruction,
+			// encoding an ethereum_xcm::transact call.
+			let mut encoded: Vec<u8> = Vec::new();
+			encoded.append(&mut xcm_domain.ethereum_xcm_transact_call_index.clone());
+			encoded.append(&mut eth_tx.encode());
+
+			Ok(encoded)
```
NunoAlexandre, line 474: Change the layout to be more declarative:

1. append(call_index)
2. append(eth_tx)
3. return

pallets/connectors/src/message.rs
```diff
@@ -0,0 +1,276 @@
...
+{
+	fn call_type(&self) -> u8 {
+		match self {
+			Self::Invalid => 0,
+			Self::AddPool { .. } => 1,
+			Self::AddTranche { .. } => 2,
+			Self::UpdateTokenPrice { .. } => 3,
+			Self::UpdateMember { .. } => 4,
+			Self::Transfer { .. } => 5,
+		}
```
NunoAlexandre, line 62: Consider a idiomatic way of doing this

runtime/development/src/lib.rs
```diff
@@ -1107,6 +1109,44 @@ impl pallet_collator_selection::Config for Runtime {
...
+	}
+}
+
+impl xcm_primitives::XcmTransact for DestTransactors {
+	fn destination(self) -> MultiLocation {
+		match self {
+			DestTransactors::Moonbeam => MultiLocation::new(1, X1(Parachain(2023))),
+		}
+	}
+}
```
NunoAlexandre, line 1129: This is actually not used by the extrinsic we use from `xcm_transactor` so clean this up a little

NunoAlexandre (commented):
Self review

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,522 @@
...
+pub use routers::*;
+
+#[derive(Encode, Decode, Clone, PartialEq, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub enum Domain {
+	Centrifuge,
+	Moonbeam,
+	Ethereum,
+	Avalanche,
+	Gnosis,
```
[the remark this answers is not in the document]
offerijns: > is it for incoming call executions?

Yes exactly. The domain indices here need to match those used in the EVM contracts and these need to pass the Centrifuge domain to send tranche tokens from the other domain here.

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,522 @@
...
+		/// The specified domain has no associated router
+		InvalidDomain,
+		/// Failed to send a message
+		SendFailure,
+		/// Token price is not set
+		MissingPrice,
+		/// The router does not exist
+		MissingRouter,
+		/// The selected domain is not yet supported
+		UnsupportedDomain,
```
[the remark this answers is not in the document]
offerijns: An edge case: in theory, at some point we could disable a router to some domain again. Then the Domain would still be included in the enum (since the indices need to remain consistent, similar to the CurrencyId enum), but it would throw this error.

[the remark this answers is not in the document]
NunoAlexandre: Ok cool, that's fair 👍

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: Sweet, thanks for the input.

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: I am adding the Domain variants back but keep this error variant deleted since we don't need to support this now.

pallets/connectors/src/message.rs
NunoAlexandre: @lemunozm @gruberb maybe your Rust skills can come to the rescue here? I tried using `#repr(u8)` here but that's not well supported for enum variants with fields. Do you suggest some other approach?

gruberb: Hm, what does `call_type` do? Maybe adding a field `call_type` to the `enum` variants itself with an `enum CallType { NameA = 0, NameB = 1, ...}`

pallets/connectors/src/message.rs
NunoAlexandre: @mustermeiszer thank for the input. `using_encoded` has a completely different api that doesn't seem to quite fit what we need here. We need to go from `Message` to `Vec<u8>`; that's the api that `encode` provides whereas `using_encoded` is `using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R`, i.e, it first converts the type to bytes and that applies the `FnOnce` passed as an argument to produce the output `R`.

pallets/connectors/src/message.rs
NunoAlexandre: `call_type` is just an `Message` variant id essentially, that needs to be stable as it's to be referenced cross-chain. Your suggestion doesn't help here since `call_type` is a value to be known at compilation-time, not at the runtime/value level. I think we can keep it as is, it's not much code and it's explicit. Thanks tho!

NunoAlexandre:
@offerijns @mikiquantum cc @mustermeiszer we just have a few todos to complete here, I am opening this PR for review to start addressing your reviews. Keep in mind the scope of this PR and that we are keeping it on development for now. Thanks 🏄

pallets/connectors/src/message.rs
mustermeiszer: But doesn't the default implementation of `Encode` allow us to use `encode()` when we implement `using_encoded`? 

TBH, I am not sure if this is important. Just was thinking that this is the case for how the default impl works.

NunoAlexandre (commented):
Leaving two notes on the 2 TODOs

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+#[cfg_attr(feature = "std", derive(Debug))]
+pub struct DomainAddress<Domain> {
+	pub domain: Domain,
+	pub address: [u8; 32],
+}
+
+impl<Domain> TypeId for DomainAddress<Domain> {
+	const TYPE_ID: [u8; 4] = *b"dadr";
+}
+
```
mustermeiszer, line 68: Move to common-types and ids respectively. Or do you want to keep them here for the time while connectors is still POC?

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+					PermissionScope::Pool(pool_id),
+					who.clone(),
+					Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, Self::now()))
+				) || T::Permission::has(
+					PermissionScope::Pool(pool_id),
+					who.clone(),
+					Role::PoolRole(PoolRole::MemberListAdmin)
+				),
+				BadOrigin
+			);
```
mustermeiszer, line 314: Does this mean anybody who is a tranche investor is allowed to add a domain as a tranche investor? This seems wrong to me.

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+			let who = ensure_signed(origin.clone())?;
+
+			// Check that the destination is a member of this tranche token
+			ensure!(
+				T::Permission::has(
+					PermissionScope::Pool(pool_id),
+					address.into_account_truncating(),
+					Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, Self::now()))
+				),
+				BadOrigin
```
mustermeiszer, line 354: I think we should add better error resolution to the permissions trait. The `BadOrigin` is getting really opaque if we have batched calls and so on. 

This is on my plate but I would appreciate some input on what you think is a good approach.

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+		) -> DispatchResult {
+			let Router::Xcm(xcm_domain) =
+				<DomainRouter<T>>::get(domain.clone()).ok_or(Error::<T>::MissingRouter)?;
+
+			let contract_call = Self::encoded_contract_call(message.encode());
+			let ethereum_xcm_call =
+				Self::encoded_ethereum_xcm_call(xcm_domain.clone(), contract_call);
+
+			pallet_xcm_transactor::Pallet::<T>::transact_through_sovereign(
+				T::Origin::root(),
```
mustermeiszer, line 406: Just curious. Why does it need to be root?

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+				// The destination to which the message should be sent
+				Box::new(xcm_domain.location),
+				fee_payer,
+				// The currency in which we want to pay fees
+				CurrencyPayment {
+					currency: Currency::AsCurrencyId(xcm_domain.fee_currency),
```
mustermeiszer, line 412: many currency... 

For my understanding, where does `Currency::AsCurrencyId` come from? And which currency enum are we referring to here? Ours?

pallets/connectors/src/message.rs
mustermeiszer: There is a nightly api if I remember correctly. Bu probably not the right way to go ^^

mustermeiszer (commented):
I am not sure, at which state we are with this. One thing I noticed is that the actual transfer to the local domain address is commented out. Is this intended even for the dev environment? 

The only thing I think is wrong is that we allow anybody who has the `PoolRole::TrancheInvestor` to add a `Domain` as a tranche investor to a pool. But maybe I am misunderstanding the logic or this is intended. Could you clarify that @NunoAlexandre 

if this is good, I would approve. ^^

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+			// T::Tokens::transfer(
+			// 	T::CurrencyId::Tranche(pool_id, tranche_id),
+			// 	&who,
+			// 	&DomainLocator {
+			// 		domain: address.domain,
+			// 	}
+			// 	.into_account_truncating(),
+			// 	amount,
+			// 	false,
+			// )?;
```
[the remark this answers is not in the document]
mustermeiszer: I am working on the currencyId enum anyways this week. We could make the `TrancheCurrency` a struct which connectors directly depends on and it will implement `Into<CurrencyId>`. 

Other solution, we extend the `PoolInspect` trait and let it return `TrancheCurrency` and `PoolCurrency` for a pool, if it exists. Then you could simply use the returned `Currency`.

pallets/connectors/src/lib.rs
NunoAlexandre: Good question! So, our current use case is one where we have the `ConnectorsXcmRouter` contract deployed on Moonbeam's EVM. When an origin calls a Connectors extrinsic targeting `Moonbeam`, they will essentially be sending an XCM to Moonbeam which will request to Transact an ethereum_xcm::transact call, which in turns targets the `ConnectorsXcmRouter::handle(msg)` function. In the contract, we need to check that `handle(msg)` was evoked by Centrifuge, since we can't trust any other source to send Connectors Messages.

For that authentication check, we need the evm tx `from` to be derived deterministically from the Centrifuge parachain account, independently the outer origin calling the Connectors extrinsic. That way, we have the caller paying for fees while having the Centrifuge parachain sovereign account as the origin of the XCM and thus the EVM tx executed on Moonbeam.

Let me know if that makes sense 👍

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: Ah yes that would be nice @mustermeiszer 😃

pallets/connectors/src/message.rs
```diff
@@ -0,0 +1,282 @@
...
+	Invalid,
+	AddPool {
+		pool_id: PoolId,
+	},
+	AddTranche {
+		pool_id: PoolId,
+		tranche_id: TrancheId,
+		token_name: [u8; 32],
+		token_symbol: [u8; 32],
+	},
```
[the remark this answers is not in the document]
NunoAlexandre: Yes, that logic is checked for at the Contract level in the respective domain.

pallets/connectors/src/lib.rs
NunoAlexandre: This was coded by @offerijns but my understanding is not that we are adding a Domain as a tranche investor but rather that we are sending an `UpdateMember` message to update the relative state on the target Domain. Does that make sense?

pallets/connectors/src/lib.rs
NunoAlexandre: I would keep them here while this is in POC to be honest, yeah 👍

pallets/connectors/src/lib.rs
offerijns: The idea was to use the permission pallet to bookkeep whitelisted addresses on other domains (chains), by storing a derived account from a `DomainAddress`, which represents an address on another domain.

The reasoning for having this `or` clause was that if you are already a whitelisted tranche investor on centrifuge chain, you might be able to whitelist your own address on another domain. But I've been going back and forth on whether this makes sense as well, so for now I think it would probably be cleaner to just only allow `MemberListAdmin`s to whitelist `DomainAddress`ses.

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+
+		/// Update a token price
+		#[pallet::weight(< T as Config >::WeightInfo::update_token_price())]
+		pub fn update_token_price(
+			origin: OriginFor<T>,
+			pool_id: PoolIdOf<T>,
+			tranche_id: TrancheIdOf<T>,
+			domain: Domain,
+		) -> DispatchResult {
+			let who = ensure_signed(origin.clone())?;
```
mikiquantum, line 267: Since we added a permissions check just below, that triggered me to ask who is supposed to be allowed to perform this operation? I can see in the spec `Only calleable by router.`, but is not clear what i means to me. I guess since it is still a POC for functional verification is not critical for now
The same question applies to all the functions that we will support in this first iteration.

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+			address: DomainAddress<Domain>,
+			amount: <T as pallet::Config>::Balance,
+		) -> DispatchResult {
+			let who = ensure_signed(origin.clone())?;
+
+			// Check that the destination is a member of this tranche token
+			ensure!(
+				T::Permission::has(
+					PermissionScope::Pool(pool_id),
+					address.into_account_truncating(),
```
mikiquantum, line 351: Why do we check here the target address against the permissions for that tranche token? Does it mean that a user that is willing to send some tranche tokens elsewhere they need to first add that address as a member of this token and then trigger the transfer? What is the reason we do it that way, out of curiosity?

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+			);
+
+			encoded
+		}
+
+		/// Return the encoded contract call, i.e, ConnectorsXcmRouter::handle(encoded_msg).
+		pub fn encoded_contract_call(encoded_msg: Vec<u8>) -> Bytes {
+			let contract = xcm_router_contract();
+			let encoded_contract_call = contract
+				.function("handle")
```
mikiquantum, line 466: I guess `handle` should always be the same, but wouldn't hurt to make it a constant?

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+/// since that's all we need to build the calls for remote EVM execution.
+pub fn xcm_router_contract() -> Contract {
+	use sp_std::collections::btree_map::BTreeMap;
+
+	let mut functions = BTreeMap::new();
+	#[allow(deprecated)]
+	functions.insert(
+		"handle".into(),
+		vec![ethabi::Function {
+			name: "handle".into(),
```
mikiquantum, line 487: And use the constant here as well

pallets/connectors/src/lib.rs
mustermeiszer: Is the whitelisted address from derived from the user account that we have here on-chain? I.e. can we link Centrifuge account x to Ethereum account y? Else, we would give users who are tranche investors `MemberListAdmin` rights on all other chains we connect to, right?

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Which one do you prefer? I am still in the process of making up my mind around the changes to `CurrencyId`. 

I have two options in my head: 

### First - structured currency types that are wrapped in enums
```rust
// Also implements `Into<CurrencyId`
struct TrancheCurrency<PoolId, TrancheId> {
   pool_id: PoolId, 
   tranche_id: TrancheId
}

impl<PoolId, TrancheId> TrancheCurrency<PoolId, TrancheId> {
   const DECIMALS: u128 = ???; // Need to get this from the registry
   const SYMBOL: &str =  ???; //Need to get this from the registry
}

// Also implements `Into<CurrencyId`
struct KSM;

impl KSM {
   const DECIMALS: u128 = primitives::currency_decimals::KSM;
   const SYMBOL: &str =  &"KSM";
}

// Also implements `Into<CurrencyId`
struct AUSD;

impl AUSD {
   const DECIMALS: u128 = primitives::currency_decimals::AUSD;
   const SYMBOL: &str =  &"AUSD";
}

// Currencies that can. beused in a pool as a base currency
pub enum PoolCurrency {
   KSM(KSM),
   AUSD(AUSD)
}

// The currencies one can invest into 
pub enum InvestmentId {
   Tranche(TrancheCurrency<PoolId, TrancheId>)
}

// The overarching CurrencyId enum that is only used by pallets
// that need knowledge of all currencies, e.g. orml_tokens
//
// MUST NOT be differentiate between runtimes, but could be
mod altair {
   enum CurrencyId {
     KSM(KSM),
     AUSD(AUSD),
     TrancheCurrency(TrancheCurrency)
   }
}
```

### Second - Do not use structured currencies but rather have the enums convert it

pallets/connectors/src/lib.rs
mustermeiszer: Totally! Thanks for the detailed answer!

pallets/connectors/src/lib.rs
mustermeiszer: 🤯 

Naturally...

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: @mustermeiszer do you mind creating an issue to discuss this? 🙏

pallets/connectors/src/lib.rs
NunoAlexandre: yes good idea

pallets/connectors/src/lib.rs
NunoAlexandre: Done

pallets/connectors/src/lib.rs
NunoAlexandre: Done

pallets/connectors/Cargo.toml
```diff
@@ -0,0 +1,102 @@
...
+sp-core = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.26" }
+sp-io = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.26" }
+pallet-balances = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.26" }
+pallet-timestamp = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.26" }
+pallet-uniques = { git = "https://github.com/paritytech/substrate",  default-features = false , branch = "polkadot-v0.9.26" }
+
+# Orml crates
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = true, branch = "polkadot-v0.9.26" }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.26" }
+
```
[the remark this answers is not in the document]
mustermeiszer: Building pallets in dev depends on those flags being true, as we are running in `std`. It works for all pallets being tested together, as flags are `or` but for single pallets its sometimes annoying. Generally speaking, we are in std so we should enable std.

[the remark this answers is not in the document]
NunoAlexandre: oh you are specifically referring to the `[dev-dependencies]`, now I get it 👍 Done

pallets/connectors/src/lib.rs
NunoAlexandre: Good point on the batch scenario. We can introduce a `InvalidTransferPermissions` error variant for this case at least for the time being. What do you think?

pallets/connectors/src/lib.rs
NunoAlexandre: @mikiquantum I think you meant to leave this comment in the wrong extrinsic? 🤔 I'm saying this because there is no permissions check below. Honestly I had no memory about that "Only calleable by router" bit of the spec and I also don't know what that's about. @offerijns?

Regardless, as this comment now targets `update_token_price`, anyone could be allowed to call this.

pallets/connectors/src/lib.rs
NunoAlexandre: cc @offerijns ^

pallets/connectors/src/lib.rs
offerijns: "Only calleable by router" is specific to the EVM side of Connectors (it applies to methods which should only be executed by messages coming from Centrifuge Chain, hence through the router, rather than directly called by users).

pallets/connectors/src/lib.rs
offerijns: Yeah this is exactly the problem. I think for now we should simplify this to only allowing `MemberListAdmin`s to add members, so removing the OR clause here (cc @NunoAlexandre). But maybe there's value in having links between Centrifuge accounts and other domain accounts, so let's have another discussion on this when I'm back.

NunoAlexandre: Sound good @offerijns, thanks!

NunoAlexandre:
@offerijns thanks a lot for the review 🚀  I will go through it later today!

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+mod contract;
+pub use contract::*;
+
+/// A Domain is a chain or network we can send a Connectors message to.
+/// The domain indices need to match those used in the EVM contracts and these
+/// need to pass the Centrifuge domain to send tranche tokens from the other
+/// domain here. Therefore, DO NOT remove or move variants around.
+#[derive(Encode, Decode, Clone, PartialEq, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub enum Domain {
```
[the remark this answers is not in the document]
NunoAlexandre: So the idea here was to differentiate between EVM and XCM-based domains? I'm asking since it's been a while since we spoke about this and as of now, for this POC, we only support Moonbeam, and I don't see the benefit of having to specify `Domain::Parachain(<id>))` instead of `Domain::Moonbeam`. Appreciate hearing back on this 👍

pallets/connectors/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+		#[pallet::weight(< T as Config >::WeightInfo::update_token_price())]
+		pub fn update_token_price(
+			origin: OriginFor<T>,
+			pool_id: PoolIdOf<T>,
+			tranche_id: TrancheIdOf<T>,
+			domain: Domain,
+		) -> DispatchResult {
+			let who = ensure_signed(origin.clone())?;
+
+			ensure!(
```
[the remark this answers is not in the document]
NunoAlexandre: Hm yeah, if we do that we should change the error `MissingPrice` to something more general, otherwise it will sound like only the price of an non-existing tranche is missing. wdyt?

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
offerijns: The problem is, we can't really migrate to a different domain encoding later, as this also requires changes on the Solidity side. So if we want to move to this other model, we should do this now, before launching. The main advantage of the other model was that we can rely on existing standards (EVM chain IDs from https://chainlist.org/ and parachain IDs) rather than have to define our own.

To be clear, we can also switch in a future PR, I don't think it needs to be in this one. But we should do it before launch if we want to do it at all.

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: Cool, thanks! @mustermeiszer I might ping you on this one :)

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: An issue I see with your suggestion as is, is that the standard in Polkadot for parachains is not to have (just) a `ParachainId` but to have a `MultiLocation`. We can just have `ParachainId` but that's assuming that it will translate always to `MultiLocation { parents: 1, X1(Parachain(ParachainId)) }`, which is going back to a sort of aliases self-made standard.

On the other hand, if we always have to specify the full multiLocation of an xcm-based domain, that's also annoying from a user and developer perspective, and it also feels _off_ to be able to specify _any_ multilocation when these messages should only be sent to a subset of existing domains that run a Centrifuge Connector.

The disadvantage of having an enum enumerating each supported domain is that it is used across networks, as in, we call a domain "Moonbeam" even if it's used in Moonbase Alpha where it's called Moonbase or in Kusama where it's called Moonriver.

I feel like I am going in circles in my head about this one 🙃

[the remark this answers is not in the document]
offerijns: What if we create our own `Parachain` enum, while stil using the EVM chain ID standard? So basically

```
enum ParachainId {
  Centrifuge
  Moonbeam
  Acala
}

type EVMChainId = u64;

enum Domain {
  EVM(EVMChainId),
  Parachain(ParachainId)
}
```

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
offerijns: Yeah that makes sense, not entirely sure what the ideal naming is though. Any suggestions?

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: I like you, @offerijns 🚀  great suggestion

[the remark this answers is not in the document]
NunoAlexandre: Done ✅  This solution still presents the same issues with the naming cross network that I mentioned but I think it's fine. If you disagree, we can always have the `ParachainId` type being passed as a Config type of the pallet itself, so that it can be set on a per-runtime basis.

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: So, we need to check whether the tranche token exists in `add_tranche` whilst not needing to fetch the tranche price. This means that we need to have an error variant for `TrancheNotFound` for this case. 

So I just renamed `MissingPrice` to `MissingTranchePrice` to be more specific it's Tranche-related and left a richer doc comment on what this error can mean.

NunoAlexandre:
@offerijns this is pretty much ready but now that we merged the 0.9.29 I need to update our forks of the Moonbeam repos and it's not looking pretty. On it 🏃

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
offerijns: @NunoAlexandre do we still want to address this TODO here? I think it should be possible now that https://github.com/centrifuge/centrifuge-chain/pull/995 was merged

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
offerijns: I think it works this way! 👍

offerijns (approved):
Just wondering if we want to address fixing the transfers in this PR. Besides that, everything is looking great! 🔥

PR #870, opened 2022-07-18
Move interest accrual to `on_initialize`

branan:
This nets us 60% more loans in each `update_nav`, at the cost of the extra weight in `on_initialize`. However, because that weight is better-constrained (each rate is guaranteed to be accrued more often), we don't have to be as conservative in how we benchmark the accrual step and can compute a reasonably-precise weight for the initialize.

This also fixes penalties, since the broken penalty rates were breaking my shiny new interest accrual work.


pallets/interest-accrual/src/lib.rs
```diff
@@ -132,82 +150,78 @@ pub mod pallet {
...
+			let mut weight = T::DbWeight::get().reads_writes(1, 1);
+			let then = LastUpdated::<T>::get();
+			LastUpdated::<T>::set(Self::now());
+			let delta = Self::now() - then;
+			let _bits = Moment::BITS - delta.leading_zeros();
+			Rate::<T>::translate(|per_sec, mut rate: RateDetailsOf<T>| {
+				let new_rate =
+					match Self::calculate_accumulated_rate(per_sec, rate.accumulated_rate, then) {
+						Ok(rate) => rate,
+						Err(e) => panic!("Failure of interest accrual for rate {:?}", per_sec),
```
branan, line 179: We obviously shouldn't panic here. My inclination is to just remove the rate if calculation fails - it's obviously "wrong" in some way (probably way too large) and any loans which depend on it will also likely be "wrong" and will need to be re-issued with the correct interest rate.

Removing rates here will cause any updates to those loans to fail with a `NoSuchRate` error, but of course before this change they would have failed with an overflow in calculating the updated rate anyway.

pallets/interest-accrual/src/lib.rs
```diff
@@ -244,7 +279,7 @@ pub mod pallet {
 			accumulated_rate.checked_mul_int(normalized_debt)
 		}
 
-		fn calculate_accumulated_rate<Rate: FixedPointNumber>(
+		pub fn calculate_accumulated_rate<Rate: FixedPointNumber>(
```
branan, line 282: I made this `pub` to make it accessible to tests in the loans pallet

pallets/loans/src/functions.rs
```diff
@@ -665,14 +690,34 @@ impl<T: Config> Pallet<T> {
...
+							previous_interest_rate,
 						) = match action {
 							WriteOffAction::WriteOffToCurrentGroup => {
 								// Loans that were already written off by an admin,
 								// cannot be written off to the current group anymore.
-								let is_written_off_by_admin = match active_loan.write_off_status {
-									WriteOffStatus::WrittenOffByAdmin { .. } => true,
-									_ => false,
-								};
+								let (previous_interest_rate, is_written_off_by_admin) =
```
branan, line 698: All of this mess is the big driver of the refactor I want to do around computed rates with penalties... This is unacceptably complex code below here (it was already pretty nasty)

pallets/interest-accrual/src/lib.rs
```diff
@@ -132,82 +160,70 @@ pub mod pallet {
...
-					let debt = Self::calculate_debt(normalized_debt, rate.accumulated_rate)
-						.ok_or(Error::<T>::DebtCalculationFailed)?;
-					Ok(debt)
-				},
-			)
+			if let Some(rate) = Rate::<T>::get(interest_rate_per_sec) {
+				Self::calculate_debt(normalized_debt, rate.accumulated_rate)
+					.ok_or(Error::<T>::DebtCalculationFailed.into())
+			} else {
+				Err(Error::<T>::NoSuchRate.into())
```
branan, line 201: This if-let-else could probably just be an `ok_or`. Not sure why I wrote it this way

pallets/interest-accrual/src/benchmarking.rs
```diff
@@ -0,0 +1,51 @@
...
+fn interest_rate_per_sec<Rate: FixedPointNumber>(rate_per_annum: Rate) -> Option<Rate> {
+	rate_per_annum
+		.checked_div(&Rate::saturating_from_integer(seconds_per_year() as u128))
+		.and_then(|res| res.checked_add(&Rate::one()))
+}
+
+benchmarks! {
+	calculate_accumulated_rate {
+		let n in 1..25;
+		let now: Moment = (1 << n) - 1;
```
branan, line 40: I should leave a comment explaining how this works. The short version is "logarithms"

runtime/altair/src/lib.rs
```diff
@@ -1245,8 +1246,26 @@ pub type Executive = frame_executive::Executive<
...
+	use frame_support::{traits::OnRuntimeUpgrade, weights::Weight};
+
+	pub struct Upgrade;
+	impl OnRuntimeUpgrade for Upgrade {
+		fn on_runtime_upgrade() -> Weight {
+			let mut weight = 0;
+			weight += InterestAccrual::upgrade_to_v1();
+			weight += Loans::reference_active_rates();
+			weight += InterestAccrual::remove_unused_rates();
+			weight
```
gruberb, line 1264: Since we start at `0`, we could do: 
```
InterestAccrual::upgrade_to_v1() + Loans::reference_active_rates() + InterestAccrual::remove_unused_rates()
```

 in the body of `on_runtime_upgrade` ?

pallets/interest-accrual/src/lib.rs
```diff
@@ -244,25 +281,77 @@ pub mod pallet {
...
+				weight += migration::v1::migrate::<T>();
+			}
+			StorageVersion::<T>::set(1);
+			weight
+		}
+
+		pub fn remove_unused_rates() -> Weight {
+			let mut weight = 0;
+			Rate::<T>::translate(|_, rate: RateDetailsOf<T>| {
+				weight += T::DbWeight::get().reads_writes(1, 1);
```
gruberb, line 346: Why does this line has to be inside the `Rate::<T>::translate` function? It seems independent of it. Or is there substrate magic somewhere?

gruberb (commented):
I can't really comment on the business logic just yet, but I pointed out some parts of the code I didn't understand with the hope I am learning something!

pallets/interest-accrual/src/migration.rs
```diff
@@ -0,0 +1,105 @@
...
+pub mod v1 {
+	use super::*;
+	pub fn migrate<T: Config>() -> Weight {
+		let mut weight = T::DbWeight::get().reads_writes(1, 1);
+		let now = Pallet::<T>::now();
+		LastUpdated::<T>::set(now);
+		Rate::<T>::translate(|per_sec, rate: RateDetailsV0Of<T>| {
+			let delta = now - rate.last_updated;
+			let bits = Moment::BITS - delta.leading_zeros();
+			weight += T::DbWeight::get().reads_writes(1, 1);
```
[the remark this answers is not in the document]
branan: Yup, it's intentional. We do one read and one write globally (getting the current timestamp and writing that as the last updated time), then one read and one write per interest rate we accrue (reading then writing the rate information).

runtime/altair/src/lib.rs
branan: We could, but I wanted the structure to emphasize that we're doing a sequence of actions (that each just happen to return a weight).

pallets/interest-accrual/src/lib.rs
branan: It could probably be global for the reads&writes, just as was done in the `on_initialize`.

pallets/interest-accrual/src/lib.rs
```diff
@@ -291,4 +380,20 @@ impl<T: Config> InterestAccrual<T::InterestRate, T::Balance, Adjustment<T::Balan
...
+
+	fn renormalize_debt(
+		old_interest_rate: T::InterestRate,
+		new_interest_rate: T::InterestRate,
+		normalized_debt: Self::NormalizedDebt,
+	) -> Result<Self::NormalizedDebt, DispatchError> {
+		Pallet::<T>::do_renormalize_debt(old_interest_rate, new_interest_rate, normalized_debt)
+	}
+
+	fn reference_rate(interest_rate_per_sec: T::InterestRate) {
```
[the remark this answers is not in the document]
branan: I'm not sure why this pallet was written that way, I just followed the existing style. AFAIK there's no particular reason we need to do it this way

pallets/interest-accrual/src/lib.rs
```diff
@@ -132,82 +160,70 @@ pub mod pallet {
...
+					.ok_or(ArithmeticError::Overflow)?;
+				let past_rate = rate
+					.accumulated_rate
+					.checked_div(&rate_adjustment)
+					.ok_or(ArithmeticError::Underflow)?;
+				Self::calculate_debt(normalized_debt, past_rate)
+					.ok_or(Error::<T>::DebtCalculationFailed.into())
+			} else {
+				Err(Error::<T>::NoSuchRate.into())
+			}
```
mustermeiszer, line 226: As I mentioned in the top-level comment. I think we should note it explicitly that `past_debt` will not be correct 
`if when < last_adjusted_normalized_debt`

pallets/interest-accrual/src/lib.rs
```diff
@@ -244,25 +281,77 @@ pub mod pallet {
...
+			let mut weight = T::DbWeight::get().reads_writes(1, 1);
+			let version = Pallet::<T>::storage_version();
+			if version < 1 {
+				weight += migration::v1::migrate::<T>();
+			}
+			StorageVersion::<T>::set(1);
+			weight
+		}
+
+		pub fn remove_unused_rates() -> Weight {
```
mustermeiszer, line 343: When is this used? 

I am not a big fan of translate as it is super slow.

pallets/interest-accrual/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: A lot of pallets use this style in substrate.

pallets/interest-accrual/src/lib.rs
```diff
@@ -132,82 +160,70 @@ pub mod pallet {
...
+						rate.accumulated_rate = new_rate;
+						rate
+					})
+			});
+			T::DbWeight::get().reads_writes(2, 1)
+				+ count
+					* (T::DbWeight::get().reads_writes(1, 1)
+						+ T::Weights::calculate_accumulated_rate(bits))
+		}
+	}
```
mustermeiszer, line 190: We should call `upgrade_to_v1` here or we should provide a costume migration struct that altair can use in order to upgrade. 

And if we have a `StorageVersion` we should provide a on genesis builder that uses the latest `StorageVersion`. Hence again, would strive for enum as `StorageVersion`.

mustermeiszer (approved):
LGTM. Really, really nice to have this functionality now! Such a nice way to calculate interests!

A few points: 
* Would love to see `StorageVersion` as an enum and some thoughts on our side, how we handle versioning of storage objects in the future. I might think about some macro to handle this, if possible
* The migration itself is never called. Is this coming later?
* **IMPORTANT:** The API of `previous_debt` should document the issue I mentioned in the comments.

pallets/interest-accrual/src/lib.rs
```diff
@@ -65,22 +65,40 @@ use sp_runtime::{
...
 // Type aliases
-type RateDetailsOf<T> = RateDetails<<T as Config>::InterestRate, Moment>;
+type RateDetailsOf<T> = RateDetails<<T as Config>::InterestRate>;
+type RateDetailsV0Of<T> = RateDetailsV0<<T as Config>::InterestRate, Moment>;
 
 // Storage types
 #[derive(Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
 #[cfg_attr(feature = "std", derive(Debug))]
-pub struct RateDetails<InterestRate, Moment> {
+pub struct RateDetailsV0<InterestRate, Moment> {
```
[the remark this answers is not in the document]
branan: I copied this style from Parity. We could also namespace it (`v0::RateDetails`). Or I'm happy for other suggestions.

Ultimately we'll delete the old version once we release with the migration (we don't need to keep old migrations around like Parity does). So it won't be around long

pallets/interest-accrual/src/lib.rs
branan: You're right, we do need to add that to the genesis. I'll take care of that with the rest of the updates

pallets/interest-accrual/src/lib.rs
branan: It's used as part of the migration. Since we only expect the migration to be run on Altair against a very small pool, I'm not too worried.

pallets/interest-accrual/src/lib.rs
```diff
@@ -114,13 +132,23 @@ pub mod pallet {
...
 	pub(super) type Rate<T: Config> =
 		StorageMap<_, Blake2_128Concat, T::InterestRate, RateDetailsOf<T>, OptionQuery>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn last_updated)]
+	pub(super) type LastUpdated<T: Config> = StorageValue<_, Moment, ValueQuery>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn storage_version)]
+	pub(super) type StorageVersion<T: Config> = StorageValue<_, u32, ValueQuery>;
```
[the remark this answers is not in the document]
branan: Yeah, we can use an enum. I was just being lazy :)

pallets/interest-accrual/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Yes, I am not worried for this one here. Just for when we move more quickly and with more devs, that we have a save way of not messing up migrations. 

Just wanted to mention it here, so we can think about something in the future.

pallets/loans/src/weights.rs
```diff
@@ -48,147 +48,147 @@ pub trait WeightInfo {
 pub struct SubstrateWeight<T>(PhantomData<T>);
 impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	fn initialise_pool() -> Weight {
-		(49_422_000 as Weight)
+		(50_478_000 as Weight)
```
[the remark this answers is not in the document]
branan: These files are autogenerated by the benchmarking utility. There's not a lot of value in making the template more complex (though we _could_ have it spit out consts if we really wanted it to?)

runtime/altair/src/lib.rs
```diff
@@ -1245,8 +1246,26 @@ pub type Executive = frame_executive::Executive<
...
 >;
 
+/// Runtime upgrade logic
+mod upgrade {
+	use super::*;
+	use frame_support::{traits::OnRuntimeUpgrade, weights::Weight};
+
+	pub struct Upgrade;
+	impl OnRuntimeUpgrade for Upgrade {
+		fn on_runtime_upgrade() -> Weight {
```
branan, line 1259: @mustermeiszer The migration is handled here in the runtime, since we need to coordinate some back-and-forth between the loans and interest-accrual pallets.

branan:
This should have feedback addressed now

pallets/interest-accrual/src/lib.rs
```diff
@@ -95,6 +95,14 @@ pub struct RateDetails<InterestRate> {
...
 }
 
+#[derive(Encode, Decode, Default, TypeInfo, PartialEq)]
+#[repr(u32)]
+pub enum Release {
+	#[default]
+	V0,
+	V1,
+}
+
```
mustermeiszer, line 105: Not blocking, but WDYT about moving this to common-types in the future and if somebody has a new version of its storage, we simply create another variant?

mustermeiszer (approved):
LGTM! 🎉

pallets/interest-accrual/src/lib.rs
branan: I think it's better to track the allowed versions for each crate separately, just to avoid any potential for accidentally putting the wrong versions into storage

mustermeiszer (approved):
Still good to go.

Issue #871, opened 2022-07-18
Add field names to pallet `Event` variants


offerijns:
Per https://github.com/paritytech/substrate/issues/9903

- [x] pools pallet
- [x] loans pallet (https://github.com/centrifuge/centrifuge-chain/pull/863)
- [x] claims pallet (https://github.com/centrifuge/centrifuge-chain/pull/875)
- [x] collator-allowlist pallet (https://github.com/centrifuge/centrifuge-chain/pull/874)
- [x] keystore pallet (https://github.com/centrifuge/centrifuge-chain/pull/873)
- [x] permissions pallet (https://github.com/centrifuge/centrifuge-chain/pull/872)

mustermeiszer:
This is done.

mustermeiszer:
> I have added a PR per pallet because I thought it would imply more changes. Maybe only one PR would have been easier to review, sorry!

I actually like that. I should do smaller PRs more often.

PR #873, opened 2022-07-19
Used struct events in keystore pallet

lemunozm:
Related to: https://github.com/centrifuge/centrifuge-chain/issues/871

### DIFF
--- a/pallets/keystore/src/lib.rs
+++ b/pallets/keystore/src/lib.rs
@@ -129,11 +129,20 @@ pub mod pallet {
 	#[pallet::generate_deposit(pub (super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// A key was added.
-		KeyAdded(T::AccountId, T::Hash, KeyPurpose, KeyType),
+		KeyAdded {
+			account_id: T::AccountId,
+			key: T::Hash,
+			purpose: KeyPurpose,
+			key_type: KeyType,
+		},
 		/// A key was revoked.
-		KeyRevoked(T::AccountId, T::Hash, T::BlockNumber),
+		KeyRevoked {
+			account_id: T::AccountId,
+			key: T::Hash,
+			block_number: T::BlockNumber,
+		},
 		/// A deposit was set.
-		DepositSet(T::Balance),
+		DepositSet { new_deposit: T::Balance },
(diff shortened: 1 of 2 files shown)


pallets/keystore/src/lib.rs
```diff
@@ -281,11 +290,11 @@ pub mod pallet {
 							let block_number = <frame_system::Pallet<T>>::block_number();
 							storage_key.revoked_at = Some(block_number.clone());
 
-							Self::deposit_event(Event::KeyRevoked(
-								account_id.clone(),
```
lemunozm, line 285: I have removed here the clones because they can be passed by move

offerijns (approved):

pallets/keystore/src/lib.rs
```diff
@@ -129,11 +129,20 @@ pub mod pallet {
 	#[pallet::generate_deposit(pub (super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// A key was added.
-		KeyAdded(T::AccountId, T::Hash, KeyPurpose, KeyType),
+		KeyAdded {
+			account_id: T::AccountId,
```
NunoAlexandre, line 133: maybe we can name this more meaningfully? I assume the account here is the key `owner`?

NunoAlexandre (approved):
LGTM :)

pallets/keystore/src/lib.rs
lemunozm: No idea. Maybe someone else can confirm it?. To be honest, I wrote `account_id` because it was the name it was referring to when the previous enum was deconstructed.

NunoAlexandre: cc @cdamian

cdamian: `owner` makes sense. 

To give some more context - these keys are generated by the p2p node and then added for every account that is created on the node.

lemunozm: Maybe @offerijns can confirm what Nuno's say? 😃

offerijns (dismissed):

NunoAlexandre (dismissed):
💯

offerijns (approved):

Issue #865, opened 2022-07-13
Add pre_commit deposit


lemunozm:
- [x] Discuss if the fees for a `pre_commit` should go to the Treasury or Author.
  - There is not fees, the solution will use a deposit.
- [ ] Implementation and testing: https://github.com/centrifuge/centrifuge-chain/pull/879

Closes: https://github.com/centrifuge/centrifuge-chain-internal/issues/14

PR #879, opened 2022-07-21
[WIP] pallet-anchors: add precommit deposit

lemunozm:
This is a WIP PR, I have some pending questions:

- [ ] 1. Should we do the same (adding a deposit) for `commit()` and `evict_anchors()` (which removes part of the data that `commit` adds)?
- [ ] 2. Should `commit()` be called even if a `recommit` has not been made before? Currently exists this possibility.
- [ ] 3. Should we configure the deposit amount as a constant or should it be set by a pallet call (similar to the keystore pallet with its `set_deposit`)? I understand that the purpose of how keystore works is to update the amount without needing to update the runtime. Do we want the same?
- [ ] 4. `precommit` error handling: See [issue](https://github.com/centrifuge/centrifuge-chain-internal/issues/22)

Notes: 
- I have not added a refund in the `commit()` call for the `precommit` resources because a `commit` does not remove the `recommit` data; only the `evit_pre_commits()` call does.

Closes https://github.com/centrifuge/centrifuge-chain/issues/865

### DIFF
--- a/pallets/anchors/src/lib.rs
+++ b/pallets/anchors/src/lib.rs
@@ -19,8 +19,10 @@ use codec::{Decode, Encode};
 use frame_support::{
 	dispatch::{DispatchError, DispatchResult},
 	storage::child,
+	traits::Currency,
 	RuntimeDebug, StateVersion,
 };
+
 use scale_info::TypeInfo;
 use sp_arithmetic::traits::{CheckedAdd, CheckedMul};
 use sp_runtime::{traits::Hash, ArithmeticError};
@@ -40,6 +42,9 @@ mod mock;
 mod tests;
 
 mod common;
+
+type BalanceOf<T> =
+	<<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
 
 /// Expiration duration in blocks of a pre-commit
 /// This is the maximum expected time for document consensus to take place between a pre-commit of an anchor and a
@@ -83,6 +88,7 @@ pub mod pallet {
 	// Import various types used to declare pallet in scope.
 	use super::*;
 	use frame_support::pallet_prelude::*;
+	use frame_support::traits::ReservableCurrency;
 	use frame_system::pallet_prelude::*;
 	use sp_runtime::traits::Hash;
 	use sp_std::vec::Vec;
@@ -100,6 +106,12 @@ pub mod pallet {
 	{
 		/// Type representing the weight of this pallet
 		type WeightInfo: WeightInfo;
+
+		/// Currency as viewed from this pallet
+		type Currency: ReservableCurrency<Self::AccountId>;
+
+		#[pallet::constant]
+		type PrecommitDepositAmount: Get<BalanceOf<Self>>;
 	}
 
 	/// PreCommits store the map of anchor Id to the pre-commit, which is a lock on an anchor id to be committed later
@@ -223,6 +235,8 @@ pub mod pallet {
 				Error::<T>::PreCommitAlreadyExists
 			);
 
+			<T as pallet::Config>::Currency::reserve(&who, T::PrecommitDepositAmount::get())?;
+
 			let expiration_block = <frame_system::Pallet<T>>::block_number()
 				.checked_add(&T::BlockNumber::from(PRE_COMMIT_EXPIRATION_DURATION_BLOCKS))
 				.ok_or(ArithmeticError::Overflow)?;
@@ -236,6 +250,7 @@ pub mod pallet {
 			);
 
 			Self::put_pre_commit_into_eviction_bucket(anchor_id, expiration_block)?;
+
 			Ok(())
 		}
 
@@ -338,7 +353,7 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			evict_bucket: T::BlockNumber,
 		) -> DispatchResult {
-			ensure_signed(origin)?;
+			let who = ensure_signed(origin)?;
 			ensure!(
 				<frame_system::Pallet<T>>::block_number() >= evict_bucket,
 				Error::<T>::EvictionNotPossible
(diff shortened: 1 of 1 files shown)


pallets/anchors/src/lib.rs
```diff
@@ -516,15 +533,15 @@ impl<T: Config> Pallet<T> {
 			.take(MAX_LOOP_IN_TX as usize)
 			// get eviction date of the anchor given by index
 			.filter_map(|idx| {
-				let anchor_index = <AnchorIndexes<T>>::get(idx)?;
-				let eviction_date = <AnchorEvictDates<T>>::get(anchor_index).unwrap_or_default();
-				Some((idx, anchor_index, eviction_date))
+				let anchor_id = <AnchorIndexes<T>>::get(idx)?;
```
lemunozm, line 536: I have taken the liberty of renaming this variable because it refers to the id of the anchor, not the index.

NunoAlexandre: Nice!

runtime/centrifuge/src/lib.rs
```diff
@@ -767,8 +767,14 @@ impl pallet_fees::Config for Runtime {
 	type WeightInfo = weights::pallet_fees::SubstrateWeight<Self>;
 }
 
+parameter_types! {
+	pub const PreCommitDeposit: Balance = 100 * CFG;
```
NunoAlexandre, line 771: I this value final? This is _a lot_ 🤔

NunoAlexandre (commented):
Code looks good!

runtime/centrifuge/src/lib.rs
lemunozm: I have used the same used in the keystore pallet, but it is a value that definitely we need to discuss. I add it to the question list 😛

pallets/anchors/src/lib.rs
```diff
@@ -356,6 +372,8 @@ pub mod pallet {
 
 				<PreCommitEvictionBuckets<T>>::remove((evict_bucket, idx));
 
+				<T as pallet::Config>::Currency::unreserve(&who, T::PreCommitDeposit::get());
```
mikiquantum, line 375: We cannot do this like this here, the `who` origin can be triggered by anyone, so we cannot give the deposit to the invoker. We need to get the identity from the PreCommit storage instead here, and unreserve on it.

mikiquantum (commented):
I am still on the fence about the deposit vs fee here bc the archiving node nature. In any case, I can be leaned towards either one. 
A few comments:
- I would add the unreserve as well in the commit call, when there is an existing precommit. The reason being that for most of the flows that call precommit before will call commit a few minutes later max (most likely seconds). 
- The issue of not storing the PreCommitDeposit in the precommit itself is that if we unreserve after that amount has been changed by a runtime upgrade, then the user would get less/more (depending on the lock) of what it was reserved. If we go with this approach we need to store the precommit balance in the storage.

pallets/anchors/src/lib.rs
```diff
@@ -100,25 +106,26 @@ pub mod pallet {
...
 		type WeightInfo: WeightInfo;
+
+		/// Currency as viewed from this pallet
+		type Currency: ReservableCurrency<Self::AccountId>;
+
+		/// Amount of funds reserved in a [`Pallet::pre_commit()`] call.
+		/// These funds will be unreserved once the user make the [`commit()`] succesfully
+		/// or call [`Pallet::evict_pre_commits()`]
+		#[pallet::constant]
+		type PreCommitDeposit: Get<BalanceOf<Self>>;
```
lemunozm, line 117: Probably this type can be removed once `pallet_fees` be updated with a storage key-enum based.

runtime/centrifuge/src/lib.rs
lemunozm: Finally we've chosen 1 CFG :)

pallets/anchors/src/benchmarking.rs
```diff
@@ -0,0 +1,119 @@
...
+			.expect("resulting BoundedVec is equal to EVICT_PRE_COMMIT_LIST_SIZE");
+
+		frame_system::Pallet::<T>::set_block_number(PRE_COMMIT_EXPIRATION_DURATION_BLOCKS.into());
+
+	}: _(RawOrigin::Signed(caller), anchor_ids)
+	verify {
+		assert_eq!(<PreCommits<T>>::iter_values().count(), 0);
+	}
+
+	evict_anchors {
```
lemunozm, line 95: It is really difficult to accuracy benchmark this. It has two big parameters that affect the computation of this call. First, the `MAX_LOOP_IN_TX`, and second the number of days from the last `evict_anchors` call. The first one is handled. But the second one has no limit, and I do not know how to deal with it. Anyway, I don't think it is an issue, since this call is not idempotent. I mean, only the first call could have a big amount of computation, the next ones not.

pallets/anchors/src/lib.rs
lemunozm: @mikiquantum Do we want to use a key from the pallet key for the pre-commit deposit?

mikiquantum: Yes, I would say so, since we already have the setup for it.

pallets/anchors/src/benchmarking.rs
mikiquantum: I think we should change the evict_anchor logic and simplify it to be called externally per anchor_id like we do with the evict_precommit. Lets chat more about it offline.

mikiquantum (approved):

libs/common-types/src/lib.rs
```diff
@@ -504,6 +504,7 @@ pub enum Adjustment<Amount> {
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 pub enum FeeKey {
 	CommitAnchor,
+	PreCommitDeposit,
```
offerijns, line 507: I should have looked at this earlier, my bad, but wouldn't it make more sense to call these keys something like
```
AnchorCommitDeposit
AnchorPreCommitDeposit
```

Since there could possibly be other pre-commits in the future not related to anchors? And for consistency.

lemunozm: Good idea! Even we can add a sub-enum per pallet I think. Something like:
```rust
FeeKey::Anchor(Anchor::CommitDeposit);
```

PR #881, opened 2022-07-21
Add `CurrencyPrice` trait and implement for tranche token prices

offerijns:
Implements a generic trait that return a price for a pair of two `CurrencyId`s. For tranche tokens specifically, we almost always want to get the price denominated in the pool currency, hence to simplify the API and not require clients to know the pool currency id, we make this optional.

This is a requirement for https://github.com/centrifuge/centrifuge-chain/pull/861

In the future, the trait can also be implemented for other currencies, for example referring to an oracle pallet to get stablecoin prices.

### DIFF
--- a/pallets/pools/src/impls.rs
+++ b/pallets/pools/src/impls.rs
@@ -11,7 +11,6 @@
 // GNU General Public License for more details.
 
 use common_traits::{CurrencyPair, PriceValue};
-use common_types::CurrencyId;
 
 use super::*;
 
@@ -39,38 +38,31 @@ impl<T: Config> PoolInspect<T::AccountId, T::CurrencyId> for Pallet<T> {
 		let mut pool = Pool::<T>::get(pool_id)?;
 
 		// Get cached nav as calculating current nav would be too computationally expensive
-		if let Some((nav, nav_last_updated)) = T::NAV::nav(pool_id) {
-			let total_assets = pool.reserve.total.saturating_add(nav);
+		let (nav, nav_last_updated) = T::NAV::nav(pool_id)?;
+		let total_assets = pool.reserve.total.saturating_add(nav);
 
-			let tranche_index: usize = pool
-				.tranches
-				.tranche_index(&TrancheLoc::Id(tranche_id))?
-				.try_into()
-				.ok()?;
-			let prices = pool
-				.tranches
-				.calculate_prices::<T::BalanceRatio, T::Tokens, _>(total_assets, now)
-				.ok()?;
+		let tranche_index: usize = pool
+			.tranches
+			.tranche_index(&TrancheLoc::Id(tranche_id))?
+			.try_into()
+			.ok()?;
+		let prices = pool
+			.tranches
+			.calculate_prices::<T::BalanceRatio, T::Tokens, _>(total_assets, now)
+			.ok()?;
 
-			let base = pool.tranches.tranche_currency(TrancheLoc::Id(tranche_id))?;
+		let base = pool.tranches.tranche_currency(TrancheLoc::Id(tranche_id))?;
 
-			match prices
-				.get(tranche_index)
-				.map(|rate: &T::BalanceRatio| rate.clone())
-			{
-				Some(price) => Some(PriceValue {
-					pair: CurrencyPair {
-						base,
-						quote: pool.currency,
-					},
-					price,
-					last_updated: nav_last_updated,
-				}),
-				None => None,
-			}
-		} else {
-			None
-		}
+		let price = prices.get(tranche_index).cloned()?;
+
+		Some(PriceValue {
+			pair: CurrencyPair {
+				base,
+				quote: pool.currency,
+			},
+			price,
+			last_updated: nav_last_updated,
+		})
 	}
 }
 
--- a/pallets/pools/src/impls.rs
+++ b/pallets/pools/src/impls.rs
@@ -1,0 +1,87 @@
+// Copyright 2021 Centrifuge Foundation (centrifuge.io).
+//
+// This file is part of the Centrifuge chain project.
+// Centrifuge is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version (see http://www.gnu.org/licenses).
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+use common_traits::{CurrencyPair, PriceValue};
+use common_types::CurrencyId;
+
+use super::*;
+
+impl<T: Config> PoolInspect<T::AccountId, T::CurrencyId> for Pallet<T> {
+	type PoolId = T::PoolId;
+	type TrancheId = T::TrancheId;
+	type Rate = T::BalanceRatio;
+	type Moment = Moment;
+
+	fn pool_exists(pool_id: Self::PoolId) -> bool {
+		Pool::<T>::contains_key(pool_id)
+	}
+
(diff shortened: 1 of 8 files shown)


NunoAlexandre:
> For tranche tokens specifically, we almost always want to get the price denominated in the pool currency, hence to simplify the API and not require clients to know the pool currency id, we make this optional.

This makes sense and it can also be applied differently in different contexts; for example, on a context where USD would be a sensible default, passing `None` as the `quote` value would default to USD.

runtime/development/src/lib.rs
```diff
@@ -936,6 +936,56 @@ impl PoolUpdateGuard for UpdateGuard {
...
+	type Rate = Rate;
+	type Moment = Moment;
+
+	fn get_latest(
+		base: CurrencyId,
+		quote: Option<CurrencyId>,
+	) -> Option<PriceValue<CurrencyId, Self::Rate, Self::Moment>> {
+		match base {
+			CurrencyId::Tranche(pool_id, tranche_id) => {
+				let now = <pallet_timestamp::Pallet<Runtime> as UnixTime>::now().as_secs();
```
branan, line 950: This should probably return `None` if the requested `quote` parameter isn't `None` or the pool's currency?

runtime/development/src/lib.rs
```diff
@@ -936,6 +936,56 @@ impl PoolUpdateGuard for UpdateGuard {
...
+				let mut pool = pallet_pools::Pool::<Runtime>::get(pool_id)?;
```
branan, line 951: Should the body of this be moved into the pool pallet (a `get_tranche_token_price` method or something like that?)

runtime/development/src/lib.rs
```diff
@@ -936,6 +936,56 @@ impl PoolUpdateGuard for UpdateGuard {
 	}
 }
 
+pub struct CurrencyPriceSource;
+impl CurrencyPrice<CurrencyId> for CurrencyPriceSource {
+	type Rate = Rate;
+	type Moment = Moment;
+
+	fn get_latest(
```
branan, line 944: This is complex enough to warrant a test, either here or as broken up into an accessor on the pallet

branan (changes requested):

runtime/development/src/lib.rs
offerijns: I think that makes sense, yeah 👍 Will move it.

pallets/pools/src/impls.rs
```diff
@@ -0,0 +1,87 @@
...
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version (see http://www.gnu.org/licenses).
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+use common_traits::{CurrencyPair, PriceValue};
+use common_types::CurrencyId;
```
NunoAlexandre, line 14: This is unused

pallets/pools/src/impls.rs
```diff
@@ -0,0 +1,87 @@
...
+
+	fn get_tranche_token_price(
+		pool_id: Self::PoolId,
+		tranche_id: Self::TrancheId,
+	) -> Option<PriceValue<T::CurrencyId, T::BalanceRatio, Moment>> {
+		let now = Self::now();
+		let mut pool = Pool::<T>::get(pool_id)?;
+
+		// Get cached nav as calculating current nav would be too computationally expensive
+		if let Some((nav, nav_last_updated)) = T::NAV::nav(pool_id) {
```
NunoAlexandre, line 42: To simplify this block, I suggest:

1. Instead of having a `if <> { huge_block } else { None }`, do:
``` rust
let (nav, nav_last_updated) = T::NAV::nav(pool_id)?;
```

and thus flatten this whole block

pallets/pools/src/impls.rs
```diff
@@ -0,0 +1,87 @@
...
+				Some(price) => Some(PriceValue {
+					pair: CurrencyPair {
+						base,
+						quote: pool.currency,
+					},
+					price,
+					last_updated: nav_last_updated,
+				}),
+				None => None,
+			}
```
NunoAlexandre, line 70: 2. Clean this up as well:

``` rust
    let price = prices.get(tranche_index).cloned()?;

		Some(PriceValue {
			pair: CurrencyPair {
				base,
				quote: pool.currency,
			},
			price,
			last_updated: nav_last_updated,
		})
```

NunoAlexandre (commented):
Some code improvement suggestions. I will open a PR to make it easier to compare. The tests don't compile btw

runtime/development/src/lib.rs
offerijns: Yeah the rewriting to an accessor on the pallet already happened, test is still in progress :)

pallets/pools/src/tests.rs
```diff
@@ -650,6 +660,15 @@ fn epoch() {
 			pool.reserve.total + senior_epoch.token_price.saturating_mul_int(250 * CURRENCY),
 			1010 * CURRENCY
 		);
+
+		assert!(
+			<Pools as PoolInspect<
+				<Test as frame_system::Config>::AccountId,
+				<Test as Config>::CurrencyId,
+			>>::get_tranche_token_price(0, SeniorTrancheId::get())
```
NunoAlexandre, line 668: Maybe we can extract this to a test utils function to reduce noise

offerijns: I think this is too little to actually make sense to extract (it's just a function call). We definitely should extract a lot of other things in the pools tests to separate utils, but I'd consider that for a future PR.

offerijns: https://github.com/centrifuge/centrifuge-chain/issues/885

NunoAlexandre: I disagree. Sure there are bigger fish to fry but this takes 1 minute to implement; These minor but noisy implements easily creep up throughout the codebase and it's completely unnecessary.

offerijns: My point isn't that I don't want to spend time on changing this here (I'm just pointing to larger refactors as something to do later). It's that I don't think this should be moved into a separate test util at all.

This is already a single function call, if we start moving all of those into other separate functions, we just bloat the test code with utils.

In my opinion, a heuristic for whether something should be moved into a helper is that both of these apply:
- The same code is called multiple times => this is the case here.
- The code at least combines >1 function calls or otherwise has some complexity => this isn't the case here.

NunoAlexandre: Thanks for explaining where you stand on this. I find this approach inconsistent with, for example, how we have type aliases in our pallets (`AccountOf<T>`, `OriginOf<T>`, etc); it's all about reduce and make what's important easier to read/look at. 

In this case, if we could do `assert_eq(get_tranche_token_price(a, b), x)` or `assert_eq(PoolInspectOf<Pools>::get_tranche_token_price(a, b), x)` instead of of having all of this noisy lines of code duplicating type casting and references, that'd be a plus.

Leaving this one here tho 👍

runtime/development/src/lib.rs
offerijns: Misread the above, will add a check that `quote` parameter isn't `None` or the pool's currency

offerijns: Fixed now.

pallets/pools/src/tests.rs
branan: I think it's worth adding the type alias or a one-line helper, but not strongly enough to block this on it :shrug:

branan (approved):

pallets/pools/src/tests.rs
offerijns: I don't really agree with the one-line helper for a single function (since this would be adding a one-line helper for a one-line function, and if we do that consistently, we'll have a huge amount of helpers), but I hadn't considered the option of the type alias (we're not actually using any of these in the tests right now). I think that makes more sense. I'll still add this in another PR.

Issue #864, opened 2022-07-13
Support pricing ongoing loans


offerijns:
Currently, `loans.price` can only be called for loans with the `Created` status or loans that are `Active` but with `debt == 0`, by anyone with the `PricingAdmin` permission.

Going forward, we want to change this into two different scenarios:

### `Created` loans
If `loans.price` is called for a loan with the `Created` status:
  - Check whether the origin has the `PricingAdmin` permission.
  - Validate the interest rate and loan type.
  - Add to `ActiveLoans`.
  - Set the loan status to `Active`.

### `Active` loans
If `loans.price` is called for a loan with the `Active` status:
  - Check whether the origin has the `LoanAdmin` permission.
  - Validate the interest rate and loan type.
  - If `current debt > 0`, similar to what we do in borrow/repay:
    - Calculate the old present value.
    - Update the current debt.
    - Calculate the new present value.
    - `update_nav_with_updated_present_value`.

For the implementation of this, we should probably modify rename `price_loan` to `price_created_loan` and add a new method to `pallets/loans/functions.rs` called `price_active_loan`, to clearly delineate this logic.

offerijns:
I think we should remove that behaviour, it's overly complex and not really necessary. It's fine to restrict `PricingAdmin`s to new loans and `LoanAdmin`s to active loans, no matter the debt.

PR #884, opened 2022-07-22
Pricing ongoing loans

lemunozm:
Tasks:
- [x] Implementation
- [x] UTs update
- [x] Fix benchmark
- [x] Waiting for https://github.com/centrifuge/centrifuge-chain/pull/870 in order to fix https://github.com/centrifuge/centrifuge-chain/pull/884#discussion_r927544416

Closes https://github.com/centrifuge/centrifuge-chain/issues/864

### DIFF
--- a/pallets/loans/src/benchmarking.rs
+++ b/pallets/loans/src/benchmarking.rs
@@ -175,9 +175,7 @@ where
 		(500 * CURRENCY).into(),
 	);
 
-	// add borrower role and price admin and risk admin role
 	make_free_cfg_balance::<T>(borrower::<T>());
-	make_free_cfg_balance::<T>(risk_admin::<T>());
 	assert_ok!(<T as pallet_pools::Config>::Permission::add(
 		PermissionScope::Pool(pool_id.into()),
 		borrower::<T>(),
@@ -188,6 +186,13 @@ where
 		borrower::<T>(),
 		Role::PoolRole(PoolRole::PricingAdmin)
 	));
+	assert_ok!(<T as pallet_pools::Config>::Permission::add(
+		PermissionScope::Pool(pool_id.into()),
+		borrower::<T>(),
+		Role::PoolRole(PoolRole::LoanAdmin)
+	));
+
+	make_free_cfg_balance::<T>(risk_admin::<T>());
 	assert_ok!(<T as pallet_pools::Config>::Permission::add(
 		PermissionScope::Pool(pool_id.into()),
 		risk_admin::<T>(),
@@ -461,6 +466,7 @@ benchmarks! {
 			let loan_id = (idx + 1).into();
 			let (loan_owner, asset) = create_asset::<T>(loan_id);
 			LoansPallet::<T>::create(RawOrigin::Signed(loan_owner.clone()).into(), pool_id, asset).expect("loan issue should not fail");
+			// (Remove this comment) Maybe an issue here.
 			activate_test_loan_with_defaults::<T>(pool_id, loan_id, loan_owner);
 		}
 		let loan_owner = borrower::<T>();
--- a/pallets/loans/src/benchmarking.rs
+++ b/pallets/loans/src/benchmarking.rs
@@ -466,7 +466,6 @@ benchmarks! {
 			let loan_id = (idx + 1).into();
 			let (loan_owner, asset) = create_asset::<T>(loan_id);
 			LoansPallet::<T>::create(RawOrigin::Signed(loan_owner.clone()).into(), pool_id, asset).expect("loan issue should not fail");
-			// (Remove this comment) Maybe an issue here.
 			activate_test_loan_with_defaults::<T>(pool_id, loan_id, loan_owner);
 		}
 		let loan_owner = borrower::<T>();
--- a/pallets/loans/src/functions.rs
+++ b/pallets/loans/src/functions.rs
@@ -14,12 +14,22 @@
 //! Module provides loan related functions
 use super::*;
 use common_types::{Adjustment, PoolLocator};
-use sp_runtime::ArithmeticError;
+use sp_runtime::{traits::BadOrigin, ArithmeticError};
 
 impl<T: Config> Pallet<T> {
 	/// returns the account_id of the loan pallet
 	pub fn account_id() -> T::AccountId {
 		T::LoansPalletId::get().into_account_truncating()
+	}
+
+	pub fn ensure_role(
+		pool_id: PoolIdOf<T>,
+		sender: T::AccountId,
+		role: PoolRole,
+	) -> Result<(), BadOrigin> {
+		T::Permission::has(PermissionScope::Pool(pool_id), sender, Role::PoolRole(role))
+			.then_some(())
+			.ok_or(BadOrigin)
 	}
 
 	/// check if the given loan belongs to the owner provided
--- a/pallets/loans/src/functions.rs
+++ b/pallets/loans/src/functions.rs
@@ -219,10 +219,10 @@ impl<T: Config> Pallet<T> {
 					.ok_or(Error::<T>::LoanPresentValueFailed)?;
 
 				// calculate new normalized debt without amount
-				let normalized_debt = T::InterestAccrual::adjust_normalized_debt(
+				let normalized_debt = T::InterestAccrual::renormalize_debt(
+					active_loan.interest_rate_per_sec,
 					interest_rate_per_sec,
 					active_loan.normalized_debt,
-					Adjustment::Increase(Zero::zero()),
 				)?;
 
 				active_loan.loan_type = loan_type;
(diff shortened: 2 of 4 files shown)


pallets/loans/src/functions.rs
```diff
@@ -125,68 +125,99 @@ impl<T: Config> Pallet<T> {
...
+				let write_off_groups = PoolWriteOffGroups::<T>::get(pool_id);
+				let old_pv = active_loan
+					.present_value(old_debt, &write_off_groups, active_loan.last_updated)
+					.ok_or(Error::<T>::LoanPresentValueFailed)?;
+
+				// calculate new normalized debt without amount
+				let normalized_debt = T::InterestAccrual::adjust_normalized_debt(
+					interest_rate_per_sec,
+					active_loan.normalized_debt,
+					Adjustment::Increase(Zero::zero()),
```
lemunozm, line 198: I added here an amount of zero because I don't have an amount itself to reevaluate the normalized debt. I understand that if I'm only pricing it should be zero, but please, confirm this 😄

pallets/loans/src/functions.rs
```diff
@@ -125,68 +125,99 @@ impl<T: Config> Pallet<T> {
...
+		let active_loan = PricedLoanDetails {
+			loan_id,
+			loan_type,
+			interest_rate_per_sec,
+			origination_date: None,
+			normalized_debt: Zero::zero(),
+			total_borrowed: Zero::zero(),
+			total_repaid: Zero::zero(),
+			write_off_status: WriteOffStatus::None,
+			last_updated: now,
```
lemunozm, line 151: I have changed this line in order to use the same time that was previously checked instead of a call to `Self::now()` to avoid storing a time that has not been checked. I think this should be the expected behaviour, but I would thank a confirmation.

pallets/loans/src/lib.rs
```diff
@@ -681,16 +711,14 @@ impl<T: Config> TPoolNav<PoolIdOf<T>, T::Balance> for Pallet<T> {
 
 #[macro_export]
 macro_rules! ensure_role {
-	( $pool_id:expr, $origin:expr, $role:expr $(,)? ) => {{
-		let sender = ensure_signed($origin)?;
+	( $pool_id:expr, $sender:expr, $role:expr $(,)? ) => {{
```
lemunozm, line 714: I have modified this macro because it tied two concepts (ensuring the origin and ensuring the role) that are not at first related. For example, the new changes check the origin once but check for more than one role.

Also, maybe this macro no longer needs to be a macro and can be a normal function that returns an error, similar to `ensure_signed()?` works.

pallets/loans/src/functions.rs
```diff
@@ -125,68 +125,99 @@ impl<T: Config> Pallet<T> {
...
+	) -> Result<u32, DispatchError> {
+		let now = Self::now();
+		ensure!(loan_type.is_valid(now), Error::<T>::LoanValueInvalid);
+
+		ensure!(
+			interest_rate_per_sec >= One::one(),
+			Error::<T>::LoanValueInvalid
+		);
+
+		Self::try_mutate_active_loan(
```
lemunozm, line 178: In the issue description, it's not clear to me if we need to ensure that the current debt should be > 0 or if it should work for 0 too. The current implementation supposes the second option. Please check this 😝

pallets/loans/src/lib.rs
NunoAlexandre: I like that!

pallets/loans/src/functions.rs
offerijns: Yeah the second option is right 👍

pallets/loans/src/functions.rs
offerijns: This is a good question :) The interest rate here changes, so I actually think we need the new code from Branan's outstanding PR here: https://github.com/centrifuge/centrifuge-chain/pull/870/files#diff-71eb00873e4a9d42ca27d4e664714802c8331f5b7b648962b9ac892552665f1bR154 As in, this should be:

```
let normalized_debt = T::InterestAccrual::renormalize_debt(
  previous_interest_rate_per_sec, // we'd need to retrieve this this
  interest_rate_per_sec,
  active_loan.normalized_debt,
)?;
```

pallets/loans/src/lib.rs
```diff
@@ -519,16 +520,45 @@ pub mod pallet {
...
-			// ensure sender has the pricing admin role in the pool
-			ensure_role!(pool_id, origin, PoolRole::PricingAdmin);
+			let owner = ensure_signed(origin)?;
+
 			let active_count =
-				Self::price_loan(pool_id, loan_id, interest_rate_per_sec, loan_type)?;
+				Loan::<T>::try_mutate(pool_id, loan_id, |loan| -> Result<u32, DispatchError> {
+					let loan = loan.as_mut().ok_or(Error::<T>::MissingLoan)?;
+
+					match loan.status {
```
offerijns, line 529: I like how you've organized the code here, this is really easy to read 👍

pallets/loans/src/functions.rs
lemunozm: That fits much better!

pallets/loans/src/tests.rs
```diff
@@ -393,13 +425,6 @@ fn test_price_bullet_loan() {
 			let rp = Zero::zero();
 			let res = Loans::price(Origin::signed(borrower), pool_id, loan_id, rp, loan_type);
 			assert_err!(res, Error::<MockRuntime>::LoanValueInvalid);
-
```
lemunozm, line 396: Checked now in test_price_and_reprice_loan()

pallets/loans/src/tests.rs
```diff
@@ -554,17 +565,6 @@ macro_rules! test_borrow_loan {
 				let res = Loans::borrow(Origin::signed(borrower), pool_id, loan_id, borrow_amount);
 				assert_ok!(res);
 
-				// pricing an active loan with non zero borrowed amount should not be possible
```
lemunozm, line 557: Now it is possible. Correct behavior checked in test_price_and_reprice_loan

pallets/loans/src/benchmarking.rs
```diff
@@ -457,6 +462,7 @@ benchmarks! {
 			let loan_id = (idx + 1).into();
 			let (loan_owner, asset) = create_asset::<T>(loan_id);
 			LoansPallet::<T>::create(RawOrigin::Signed(loan_owner.clone()).into(), pool_id, asset).expect("loan issue should not fail");
+			// (Remove this comment) Maybe an issue here.
 			activate_test_loan_with_defaults::<T>(pool_id, loan_id, loan_owner);
```
lemunozm, line 466: Seems like there is an issue here. We suppose that the just created loan has the `loan_id` we have created above. It seems not to be always true because sometimes the loan is already priced at this point (which shouldn't), and I get an error because the `loan_owner` has not the `LoanAdmin` role for pricing on an active loan. I have "fixed" it by adding that role, but I think we should know why sometimes the loan is already priced at this point.

A noob question, is the pallet storage reset between benchmarks of the same `benchmarks!` section? That could be the problem because the same code is used in different benchmarks. The first benchmark will match the `loan_id` with the created loan, but the second does not and will reprice the first active loans.

pallets/loans/src/functions.rs
branan: #870 is finally merged, this is unblocked

pallets/loans/src/benchmarking.rs
offerijns: Is this still relevant, or was this fixed already?

pallets/loans/src/lib.rs
```diff
@@ -519,16 +519,45 @@ pub mod pallet {
...
+						LoanStatus::Active => {
+							Self::ensure_role(pool_id, owner, PoolRole::LoanAdmin)?;
+							Self::price_active_loan(
+								pool_id,
+								loan_id,
+								interest_rate_per_sec,
+								loan_type,
+							)
+						}
+						LoanStatus::Closed { .. } => Err(Error::<T>::LoanIsActive)?,
```
offerijns, line 550: Previously only created loans could be priced, now created and active loans. Hence, this error should be renamed to something like
```suggestion
						LoanStatus::Closed { .. } => Err(Error::<T>::LoanIsClosed)?,
```

offerijns (changes requested):
Looks great! Left just one suggestion.

pallets/loans/src/benchmarking.rs
lemunozm: @branan or @mustermeiszer I do not know if any of you can take a look at this before merging the PR 🤗

mustermeiszer (approved):
Besides Jeroens commet about the naming of the event, looks good!

pallets/loans/src/lib.rs
lemunozm: It makes total sense!

offerijns (approved):
LGTM!

offerijns (approved):

PR #886, opened 2022-07-27
Filter tranche transfers

### LINKED ISSUE #878: feat: Restrict `Tranche` token transfers through XCM
### Problem 

Centrifuge `Tranche` tokens can only be transferred to a whitelisted destination accounts / locations.

For local transfers, we already have the `restricted-tokens` pallet that wraps `orml-tokens` with an added layer of checks using the `permissions` pallet.

However, with XCM, once we make `Tranche` XCM-transferrable by supporting them on the `CurrencyIdConvert` / register them in the `asset-registry`, we would leave a backdoor for unauthorised transfers to take place. This could take place using `Xtokens` or by building a XCM message manually using the `polkadotXcm.send` extrinsic. 

### Steps

We need to block Tranche tokens from being transferred to non-whitelisted accounts through XCM. For that, we will have to:

- [ ] Add a Call filter for XTokens transfers on `CurrencyId::Tranche(_, _)`
- [ ] Add a Call filter for `PolkadotXcm` extrinsics that allow for sending custom messages
- [ ] Add `orml_xcm` pallet to have access to `PolkadotXcm::send` with an origin guard

### Dependencies

Blocks https://github.com/centrifuge/centrifuge-chain/issues/842

NunoAlexandre:
Closes #878 

`dev`

- [x] Add a Call filter for XTokens transfers on `CurrencyId::Tranche(_, _)`
- [x] Add a Call filter for `PolkadotXcm` extrinsics that allow for sending custom messages
- [x] Add `orml_xcm` pallet to have access to `PolkadotXcm::send` with an origin guard

`altair`

- [x] Repeat same steps
- [x] tests

`centrifuge`
- [x] Repeat same steps
- [x] tests

### DIFF
--- a/runtime/development/Cargo.toml
+++ b/runtime/development/Cargo.toml
@@ -102,6 +102,7 @@ pallet-treasury = { git = "https://github.com/paritytech/substrate", default-fea
 orml-asset-registry = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.26" }
 orml-tokens = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.26" }
 orml-traits = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.26" }
+orml-xcm = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.26" }
 orml-xcm-support = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.26" }
 orml-xtokens = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.26" }
 
--- a/runtime/development/src/lib.rs
+++ b/runtime/development/src/lib.rs
@@ -135,7 +135,7 @@ parameter_types! {
 
 // system support impls
 impl frame_system::Config for Runtime {
-	type BaseCallFilter = Everything;
+	type BaseCallFilter = BaseCallFilter;
 	type BlockWeights = RuntimeBlockWeights;
 	type BlockLength = RuntimeBlockLength;
 	/// The ubiquitous origin type.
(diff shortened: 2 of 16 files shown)


runtime/development/src/lib.rs
```diff
@@ -978,28 +978,54 @@ impl pallet_migration_manager::Config for Runtime {
...
+			Call::XTokens(method) => match method {
+				orml_xtokens::Call::transfer {
+					currency_id: CurrencyId::Tranche(_, _),
+					..
+				}
+				| orml_xtokens::Call::transfer_with_fee {
+					currency_id: CurrencyId::Tranche(_, _),
+					..
+				}
+				// We preemptively disable this as we haven't encounter a use case for it.
```
akhilesharora, line 1021:
```suggestion
				// We preemptively disable this as we haven't encountered an use case for it.
```

runtime/development/Cargo.toml
```diff
@@ -102,6 +102,7 @@ pallet-treasury = { git = "https://github.com/paritytech/substrate", default-fea
 orml-asset-registry = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.24" }
 orml-tokens = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.24" }
 orml-traits = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.24" }
+orml-xcm = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.24" }
```
akhilesharora, line 105:
```suggestion
orml-xcm = {  git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.26" }
```

akhilesharora (commented):
LGTM apart from few suggestions!

runtime/development/src/lib.rs
NunoAlexandre: Thanks!

runtime/development/Cargo.toml
NunoAlexandre: We are still on 0.9.26

runtime/development/src/lib.rs
```diff
@@ -978,28 +978,54 @@ impl pallet_migration_manager::Config for Runtime {
...
+					..
+				}
+				| orml_xtokens::Call::transfer_with_fee {
+					currency_id: CurrencyId::Tranche(_, _),
+					..
+				}
+				// We preemptively disable this as we haven't encountered an use case for it.
+				// Shall some user or use case require it, we will make it more fine-grained.
+				| orml_xtokens::Call::transfer_multicurrencies { .. } => false,
+				// Any other XTokens call is good to go
```
offerijns, line 1024: What about `transfer_multiasset`, this can't be used to send tranche tokens then?

offerijns (approved):
Makes sense to me! One question, probably just due to a lack of understanding on my point on how xtokens works, but wanted to double check anyways.

runtime/development/src/lib.rs
NunoAlexandre: Users could use that call to transfer tranche tokens, that's why we block it. As I left commented, we have never come across the need to use this call neither do I see the need for it in the foreseeable future so we block it.

offerijns: The blocked call is `transfer_multicurrencies` though, while my question is about `transfer_multiasset` :)

NunoAlexandre: Thank you, good catch! I missed a few more actually. Fixed 👍

NunoAlexandre:
@offerijns @branan I am trying to add tests verifying certain calls to be blocked but it's not failing as I'd expect 🤔 see the `polkadot_xcm_send` test for example. Is there some caveat I am not seeing here?

branan:
@NunoAlexandre I think the issue with your test is that you're invoking the pallet directly. The filter logic happens on the `Call`, so you'd need to do something with `Call::dispatch` in order for the filtering logic to work. Alas the `Call` type is one of the autogenerated runtime types, so it's not super well documented. I'd look into how other pallets which use `Call` invoke it, such as `proxy` or `sudo`?

NunoAlexandre:
@branan thanks for the valuable insight, I got it in place 👍 

@branan @offerijns could you give this a review?

runtime/integration-tests/src/xcm/kusama/restricted_calls.rs
```diff
@@ -0,0 +1,114 @@
...
+pub mod blocked {
+	use super::*;
+
+	#[test]
+	fn xtokens_transfer() {
+		// For now, Tranche tokens are not supported in the XCM config so
+		// we just safe-guard that trying to transfer a tranche token fails.
+		// Once Tranche tokens are supported, we need to tighten this test.
+		Altair::execute_with(|| {
+			assert!(XTokens::transfer(
```
branan, line 53: should this be an `assert_noop(...)` instead of `assert(...is_err())`?

runtime/integration-tests/src/xcm/polkadot/tests/restricted_calls.rs
```diff
@@ -0,0 +1,115 @@
...
+pub mod blocked {
+	use super::*;
+
+	#[test]
+	fn xtokens_transfer() {
+		// For now, Tranche tokens are not supported in the XCM config so
+		// we just safe-guard that trying to transfer a tranche token fails.
+		// Once Tranche tokens are supported, we need to tighten this test.
+		Centrifuge::execute_with(|| {
+			assert!(XTokens::transfer(
```
branan, line 54: Same question as in the Kusama test

branan (approved):
Just one question on the tests, otherwise this LGTM.

I think if we continue to see more duplicate tests, we should look into making some generic tests that we can instantiate for both runtimes. For now I think what's there is fine, and I'm happy to keep things simple until we need to make them complex

runtime/integration-tests/src/xcm/kusama/restricted_calls.rs
NunoAlexandre: The reason why I went with `assert(...is_err())` here instead of checking for a specific error is that in this case we don't care what the error is as long as it fails. But if you would prefer to go with `assert_noop` we can do that! let me know

branan (approved):

PR #888, opened 2022-07-27
Update Polkadot v0.9.26

NunoAlexandre:


mustermeiszer:
@NunoAlexandre could you update the nix hash once the last tests run through? I haven't set up this yet ^^

runtime/integration-tests/src/pools/pool_starts.rs
```diff
@@ -40,7 +40,7 @@ async fn create_init_and_price() {
 	let mut nft_manager = NftManager::new();
 	let pool_id = 0u64;
 	let loan_amount = 10_000 * DECIMAL_BASE_12;
-	let borrow_amount = Amount::from_inner(9_000 * DECIMAL_BASE_12);
+	let borrow_amount = Rate::from_inner(9_000 * DECIMAL_BASE_12);
```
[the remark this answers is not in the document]
mustermeiszer: It's actually not used. I removed it. But otherwise, you were right.

NunoAlexandre (commented):
LGTM but I also worked on this so not the best person to approve 👍

runtime/integration-tests/src/lib.rs
```diff
@@ -14,8 +14,7 @@
 #![cfg(test)]
 #![allow(unused)]
 
-//TODO(nuno): re-enable pools once fudge is compatible with polkadot 0.9.20
-// mod pools;
+mod pools;
```
mikiquantum, line 17: Nice

runtime/common/src/fixed_point.rs
```diff
@@ -134,6 +137,48 @@ macro_rules! implement_fixed {
 			fn into_inner(self) -> Self::Inner {
 				self.0
 			}
+
+			/// Creates `self` from a rational number. Equal to `n / d`.
+			///
+			/// Returns `None` if `d == 0` or `n / d` exceeds accuracy.
+			fn checked_from_rational<N: FixedPointOperand, D: FixedPointOperand>(
```
mikiquantum, line 144: Why did the upgrade to 0.9.26 forced us to implement this function?

runtime/altair/src/weights/pallet_treasury.rs
```diff
@@ -60,4 +60,8 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 			.saturating_add(RocksDbWeight::get().reads(1 as Weight))
 			.saturating_add(RocksDbWeight::get().writes(1 as Weight))
 	}
+
+	fn spend() -> Weight {
```
mikiquantum, line 64: I guess we need to recompute the runtime weights in general right? We can do that in a following PR too

src/command.rs
```diff
@@ -328,7 +328,7 @@ pub fn run() -> Result<()> {
 			)?;
 
 			let state_version = Cli::native_runtime_version(&chain_spec).state_version();
-			let block: Block = generate_genesis_block(&chain_spec, state_version)?;
+			let block: Block = generate_genesis_block(&**chain_spec, state_version)?;
```
mikiquantum, line 331: This dereference seems weird, doesnt it work with only one `&*chain_spec` if we change the above chain_spec declaration to:
```
let chain_spec = load_spec(
```

mikiquantum (approved):
LGTM! Some minor comments only

runtime/common/src/fixed_point.rs
```diff
@@ -134,6 +137,49 @@ macro_rules! implement_fixed {
...
+			fn checked_mul_int<N: FixedPointOperand>(self, n: N) -> Option<N> {
+				let lhs: I129 = self.into_inner().into();
+				let rhs: I129 = n.into();
+				let negative = lhs.negative != rhs.negative;
+
+				use sp_runtime::traits::UniqueSaturatedInto;
+				multiply_by_rational_with_rounding(
+					lhs.value,
+					rhs.value,
+					Self::DIV.unique_saturated_into(),
```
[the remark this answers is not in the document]
offerijns: Probably good if @branan confirms as he's spent more time on this, but I think that's a sane default. Should we still add a ticket to make rounding direction in all usages (at least in pools) explicit though? There are some cases where we'd want to round up (e.g. calculating debt for a loan).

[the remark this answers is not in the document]
mustermeiszer: @offerijns We can not decide on a per cal level now. IF we wann do that, we must add additional methods, that do this, which is simple and easy to do, but just to note that. 

These methods here are basically the defaults.

runtime/common/src/fixed_point.rs
mustermeiszer: Short: Parity

Long: Parity has changed the fixed point core logic that this trait default implements on the substrate side. They choose a rounding preference, which I think they choose, as it satisfies the older already existing tests before that change.
As a plus, they now have different rounding preferences, depending on which method you use. I already opened a ticket

The possibility to choose the rounding is great though, it's just a little bit weirdly implemented on their side IMO.

PR #889, opened 2022-07-27
Add update_token_price and update_member to connectors pallet

offerijns:
- Add `update_token_price` method
- Add `update_member` method
- Remove unused storage keys
- Add connectors pallet to development runtime
- Log event on message sending


NunoAlexandre (approved):
LGTM :)

PR #890, opened 2022-07-29
Pallet fees refactor

lemunozm:
Check pallet fees [specs](https://centrifuge.hackmd.io/gBwcUqaPREGe5QRLszKkOw)

Tasks:
- [x] Implementation.
- [x] Adapt tests.
- [x] Adapt benchmarks.
- [x] Storage migration (needed because the `StorageMap` has changed).


pallets/anchors/src/lib.rs
```diff
@@ -571,7 +572,7 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
-	fn fee_key() -> <T as frame_system::Config>::Hash {
+	fn commit_fee_key() -> <T as frame_system::Config>::Hash {
 		<T as frame_system::Config>::Hashing::hash_of(&0)
```
lemunozm, line 576: This `&0` should be changed to avoid collision between pallets. The initial purpose of this PR was to change this. 

@mikiquantum we were talking about creating an enum, which represents the numbers used in the hash, but thinking deeper into it I do not see a good place to define that enum without carrying a little knowledge of each pallet out of them. I think that maybe using a prefixed string for these cases could be simpler, the information is kept inside of each pallet. i.e. `hash_of("anchors::commit_fee")`, and there is no more collision between pallets. What do you think?

mikiquantum: @lemunozm The whole point is to have a central place where these keys are defined, so (1) developing the usage of a new fee is trivial and safe, and (2) these fee list (human readable - no hashes) are available to be read by a client (UI etc) so they can be included in tx_fee estimations. 
We could extend the fee storage to be indexed by an enum instead and that enum described the list of purposes (CommitAnchor, PreCommit, etc ...)

mikiquantum: Since the `pallet_fees` has to be imported by any other one that want to read a fee, then makes sense to add that enum list in that pallet or const variable.

mikiquantum (commented):
I am on the fence on doing this refactoring now, since we need to sit down and spec out the next iteration of a Fees Pallet and we might be coding something now that we throw away soon. Knowing that a migration is needed I would, personally, not do any changes to the storage if possible until we are more clear on the direction to go.

pallets/anchors/src/lib.rs
lemunozm: Ok, I see the point. I understood the `pallet_fees` as something isolated, without knowledge of how other pallets work (as an independent crate in the future). But yeah, definitely, the enum could be there, and that addition is simpler.

lemunozm:
> I would, personally, not do any changes to the storage if possible until we are more clear on the direction to go.

That makes sense! Are the rest of the non-storage changes ok to add now? If there is pending a new design for this pallet maybe it is better not to modify anything.

mustermeiszer:
Just am leaving my current thoughts about this here. No need to answer now. Just wanted to have this somewhere, so we can discuss next week: 

* I would strongly like to abstract this away from `pallet_fees::Config`, so that we don't need to tight couple pallets using fees
  E.g.:
   ```rust
   trait Fees {
      type AccountId;
      type Fee;
      
      fn fee_to_author(from: &Self::AccountId, fee: Self::Fee) -> DispatchResult;
      fn fee_to_burn(from: &Self::AccountId, fee: Self::Fee) -> DispatchResult;
      fn fee_to_treasury(from: &Self::AccountId, fee: Self::Fee) -> DispatchResult;
   } 
   ```
* I don't really see a use-case for adding fees dynamically via an extrinsic (adapting fees is a different thing) 
   -> Hence, I would suggest we add a fixed enum whose fees can be adapted. If we need additional variants, we do a rt-upgrade. This allows 
   * other pallets to pass variants
   * `set_fees` can use variants instead of hash
   * clients can simply query metadata for knowing the variant and query storage without hashing and knowing values upfront for this

libs/common-traits/src/lib.rs
```diff
@@ -278,3 +278,183 @@ impl<T> PreConditions<T> for Never {
...
+			Ok(())
+		}
+
+		fn fee_to_treasury(_: &Self::AccountId, _: Fee<Self::Balance, K>) -> DispatchResult {
+			Ok(())
+		}
+	}
+
+	#[cfg(feature = "std")]
+	pub mod test_util {
```
lemunozm, line 341: This is a utility to easily mock the Fees from the pallets. It has not production code.

pallets/nft/src/tests.rs
```diff
@@ -144,43 +152,9 @@ fn invalid_proof() {
 				),
 				Error::<MockRuntime>::InvalidProofs
 			);
-		})
-}
-
-#[test]
-fn insufficient_balance_to_mint() {
```
lemunozm, line 151: I remove the whole test because here we are **only** testing if the `fee_to_burn()` call inside the `NFT::validate_mind()` is working well. This is something already tested in pallet_fees. Also, because the pallet_fee is mocked at this test, no longer returns an `InsufficienteBalance` error.

pallets/fees/src/lib.rs
```diff
@@ -83,106 +102,82 @@ pub mod pallet {
...
-		pub fn set_fee(
-			origin: OriginFor<T>,
-			key: T::Hash,
-			new_price: BalanceOf<T>,
-		) -> DispatchResult {
-			Self::can_change_fee(origin)?;
-			Self::change_fee(key, new_price);
-			Self::deposit_event(Event::FeeChanged(key, new_price));
+		pub fn set_fee(origin: OriginFor<T>, key: T::FeeKey, fee: BalanceOf<T>) -> DispatchResult {
+			T::FeeChangeOrigin::try_origin(origin)
```
mikiquantum, line 128: I think we can simplify this like:
```suggestion
			T::FeeChangeOrigin::ensure_origin(origin)?;
```

pallets/fees/src/lib.rs
```diff
@@ -83,106 +102,82 @@ pub mod pallet {
...
-			&from,
-			fee,
-			WithdrawReasons::FEE.into(),
-			ExistenceRequirement::KeepAlive,
-		)?;
-
-		let author = <pallet_authorship::Pallet<T>>::author().ok_or(Error::<T>::InvalidAuthor)?;
-		T::Currency::resolve_creating(&author, value);
+	fn fee_to_author(from: &Self::AccountId, fee: Fee<BalanceOf<T>, K>) -> DispatchResult {
+		if let Some(author) = <pallet_authorship::Pallet<T>>::author() {
```
mikiquantum, line 148: Would it make sense to pass the pallet_authorship as trait?

libs/common-traits/src/lib.rs
```diff
@@ -278,3 +278,183 @@ impl<T> PreConditions<T> for Never {
...
 }
+
+pub mod fees {
+	use frame_support::dispatch::DispatchResult;
+	use frame_support::traits::tokens::Balance;
+
+	/// A way to identify a fee value.
+	pub enum Fee<Balance, FeeKey> {
+		/// The fee value itself.
+		Balance(Balance),
```
mikiquantum, line 289: I think once we have standardized the usage of the fees pallet across all consumer pallets, the balance fee might not be needed, since the fee value should be gathered from the pallet_fees storage.

mikiquantum (commented):
LGTM! 
For the migration, we will need to follow a few steps:
- create a new file like this one https://github.com/centrifuge/centrifuge-chain/blob/parachain/pallets/claims/src/migration.rs with the according logic
- Add hook logic like here: https://github.com/centrifuge/centrifuge-chain/pull/830/files#diff-675ef431f8b9d6d6def6186b961b26c700d05cd417f55243ccc1db72eddf2ac1R278

libs/common-traits/src/lib.rs
lemunozm: There is still some cases, as in the commit usage, where the fee is extracted from the storage, multiplied and then pay to the author where the Balance variant is needed today.

Other possible approach I have been evaluating is to add an optional map function argument to mutate the value before make the payment. In this way we force to use the fee storage for anybody who wants to pay something, but it's still flexible to change the value if needed.

pallets/fees/src/lib.rs
lemunozm: This idea is really cool, and I wanted to do the same, but the `pallet_authorship` do not exposes any trait for the `author()` method. I do not know if there is another approach to get this, if someone know, please shoot out!

pallets/fees/src/lib.rs
lemunozm: I'm not sure to get the same behaviour with the `?`. Currently it is calling to `ensure_root` in case the `try_origin` (or ensure_origin) fails, so I still need to make the `or_else`.

runtime/centrifuge/src/lib.rs
```diff
@@ -759,15 +760,22 @@ impl pallet_treasury::Config for Runtime {
...
+	type Treasury = pallet_treasury::Pallet<Self>;
 	type Event = Event;
-	/// A straight majority of the council can change the fees.
 	type FeeChangeOrigin = EnsureRootOr<HalfOfCouncil>;
+	type DefaultFeeValue = DefaultFeeValue;
 	type WeightInfo = weights::pallet_fees::SubstrateWeight<Self>;
 }
 
 impl pallet_anchors::Config for Runtime {
+	type Fees = pallet_fees::Pallet<Runtime>;
```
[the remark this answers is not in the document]
lemunozm: 👍🏻 done in all runtimes.

pallets/fees/src/lib.rs
mikiquantum: ah, I do not know if it is possible in the runtime definition to pass the function call instead? Otherwise nvm, I think is fine.

libs/common-traits/src/lib.rs
mikiquantum: ah right! the multiplier part of the commit fee 👍  I think is fine then, maybe later we can have a more dynamic approach like you suggests.

mikiquantum (approved):
LGTM!

pallets/fees/src/lib.rs
lemunozm: You need a trait that gives that call. And then implement that it for the real authorship pallet somehow. Maybe the implementation of this new trait for the authorship pallet could be done in runtime/common. I like the idea, let me think about it, and probably for other PR it could be done.

pallets/fees/src/lib.rs
lemunozm: Yeah, you're right! Changed 😃

pallets/anchors/src/lib.rs
```diff
@@ -95,9 +97,13 @@ pub mod pallet {
...
 	#[pallet::config]
-	pub trait Config:
-		frame_system::Config + pallet_timestamp::Config + pallet_fees::Config
-	{
+	pub trait Config: frame_system::Config + pallet_timestamp::Config {
+		/// Entity used to pay fees
+		type Fees: Fees<AccountId = Self::AccountId>;
+
+		/// Key used to retrieve the fee balances in the commit method.
+		type CommitAnchorFeeKey: Get<<Self::Fees as Fees>::FeeKey>;
```
mikiquantum, line 105: 👍

pallets/fees/src/migration.rs
```diff
@@ -0,0 +1,102 @@
...
+		ensure!(count != 0, "Fees storage items not found!");
+		log::info!(
+				target: "runtime::fees::pre-migrate",
+				"Pre Migrate check passed with count {:?}",
+				count,
+		);
+		Ok(())
+	}
+
+	pub fn migrate<T: Config>() -> Weight {
```
mikiquantum, line 55: I noticed that since the migration execution depends on each pallet hook implementation and is not easily visible anymore in the runtime file, then is easy to forget that the migration is still there once it has been executed. (Noticed with the Claims migration)
So this means that if we run a migration twice (accidentally) and the second time it fails then we can incur in a chain halt (worst case scenario)
In this scenario, I think the migrate code will delete every entry in the fees storage so I think we should add here a check that do not run the migration twice.
I think doing something like:
```
let on_chain_storage_version = <P as GetStorageVersion>::on_chain_storage_version();
if on_chain_storage_version < 1 {
  // run migration
  // StorageVersion::new(1).put::<P>();
} else {
  // skip with warn message0
}
```
Maybe there is a better way to do this? WDYT?

mikiquantum (commented):
@lemunozm Everything looks good, the only thing I would add is the migration double execution check in case we accidentally forget to remove the code between runtime upgrades.

pallets/fees/src/migration.rs
lemunozm: The `migration()` function is only removing the old storage and does nothing with the new (and valid) one. So if the migration is run twice, the second one no has an effect, you would try to remove the already removed storage. Please, tell me if I'm misunderstanding this and if there is some issue under the hood that I'm not seeing.

mikiquantum: Ah! For some reason I thought that the storage name/prefix was the same. 👍

mikiquantum (approved):
Thanks @lemunozm ! 
If you want and can run the try-runtime call:
```
cargo build --release --features=try-runtime

RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace ./target/release/centrifuge-chain try-runtime --execution Native --chain centrifuge-local --no-spec-name-check on-runtime-upgrade live --uri wss://fullnode.parachain.centrifuge.io:443
```

NunoAlexandre (approved):

PR #891, opened 2022-07-30
Pass metadata optionally on pool creation

offerijns:


pallets/pools/src/lib.rs
```diff
@@ -447,7 +447,7 @@ pub mod pallet {
 		/// Pool metadata was set.
 		MetadataSet {
 			pool_id: T::PoolId,
-			metadata: Vec<u8>,
+			metadata: BoundedVec<u8, T::MaxSizeMetadata>,
```
offerijns, line 450: A bit of a clean-up change, I can't think of any reason not to make this event param a bounded vec as well.

pallets/pools/src/lib.rs
```diff
@@ -615,6 +616,23 @@ pub mod pallet {
 			let now = Self::now();
 			let tranches = Tranches::from_input::<T::TrancheToken>(pool_id, tranches, now)?;
 
+			let checked_metadata: Option<BoundedVec<u8, T::MaxSizeMetadata>> = match metadata {
```
offerijns, line 619: There must be a better way to do this, open to suggestions!

pallets/pools/src/lib.rs
```diff
@@ -615,6 +616,23 @@ pub mod pallet {
...
+				Some(metadata_value) => {
+					let checked: BoundedVec<u8, T::MaxSizeMetadata> = metadata_value
+						.clone()
+						.try_into()
+						.map_err(|_| Error::<T>::BadMetadata)?;
+
+					Self::deposit_event(Event::MetadataSet {
```
NunoAlexandre, line 626: I am thinking whether this is a good idea. Nothing is actually _set_ here, I assume we just deposit this event at this point because here we know that `metadata` is `Some`.

Why do we want to deposit this event when creating a pool? At this point, I'd rather have the `Created` event containing a new field `metadata: Option<Vec<u8>>`

pallets/pools/src/lib.rs
NunoAlexandre: If we consider my point about the event, we can turn this into a `.map` or `.and_then` pipe and clean it up a little.

pallets/pools/src/lib.rs
offerijns: My reasoning was that a client that wants to know whenever the metadata changes, should only have to listen to the `MetadataSet` event. But I realized that this is actually inconsistent with the other `create` behaviour, we should then also submit `Updated`, `Rebalanced`, `MaxReserveSet` events. So I think it makes sense to remove it then.

NunoAlexandre (approved):

pallets/pools/src/lib.rs
offerijns: @branan Nuno and I both looked at this but couldn't figure it out, the `try_into` makes this rather difficult. Our conclusion:
> I'd say let it be because the only say I presume we can get it to work is by being explicit about the try_into conversion type and that will end up just bloat the code about as much

Let me know what you think (or if you have a suggestion for a better way of doing this)!

branan: Yeah, I think the compiler can't infer what we want here, and the turbofish is probably grosser than the match statement we've got. Let's keep this as-is, and we can always clean it up if a nicer way to write it appears.

PR #893, opened 2022-08-03
Wrap up Nix Build for Node Wrapper

mikiquantum:
This PR continues the work from @asymmetric (Thanks!) in this [PR](https://github.com/centrifuge/centrifuge-chain/pull/645)

One important thing to note, is that the build process skips building the WASM, this means that the binary and/or docker image needs to use an external spec to work. In later iterations we will add this support.

At the moment images are tagged with a `nix-do-not-use` suffix until we verify all is fine on production nodes.

A following PR, once we have WASM builds, will cover cleaning up and remove duplication of RUST_TOOLCHAIN in github actions.


flake.nix
```diff
@@ -0,0 +1,150 @@
...
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = inputs :
+    inputs.flake-utils.lib.eachDefaultSystem (system:
+        let
+          name = "centrifuge-chain";
+          # This is the program version.
+          version = "0.10.18";
```
branan, line 25: Is it possible to extract this from the source instead of needing to update our version in another place? (I need to learn more nix, obviously)

mikiquantum: As far as I know (and tried with my limited knowledge) is that, generally, arguments cannot be passed to flakes. There are a few discussions open around this topic: https://github.com/NixOS/nix/issues/5663

In any case, I would love to not have neither the version nor the nightly defined here. Maybe @mpontus have an idea around this?

gpmayorga (changes requested):
I think we should either remove the Dockerfile(s) and/or rename them and/or move them to a different directory to avoid confusion, as I understand it this replaces the Dockerfile process completely?

mikiquantum:
@gpmayorga the idea is to try this image out in live networks first and then replace all the docker code later. The other reason why we do not remove all that here, is because this binary and docker images doesnt build the wasm, so we cannot use that for our non-live environments. Once we fix that (will create a PR soon) we can do the cleanup.
In any case, I would like to get this merged independently.

flake.nix
```diff
@@ -0,0 +1,154 @@
...
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = inputs :
+    inputs.flake-utils.lib.eachDefaultSystem (system:
+        let
+          pkgs = inputs.nixpkgs.legacyPackages.${system};
+
+          cargoTOML = builtins.fromTOML (builtins.readFile ./Cargo.toml);
```
mikiquantum, line 25: @branan Now I read from file some params, I had to diver deeper on nix-how-to, but we figured it out.

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
```
mikiquantum, line 1: Once we are done with the cleanup step, this file should be the source of truth for the toolchain/s to use

mikiquantum:
@branan & @gpmayorga please have another look and give a thumbsup or ask more questions, so we can get this one in.

branan:
I don't really know Nix well enough to judge this properly, so if everybody else is happy I am. No need to block on me

flake.nix
```diff
@@ -0,0 +1,154 @@
...
+            # This applies the srcFilter function to the current directory, so
+            # we don't include unnecessary files in the package.
+            src = pkgs.lib.cleanSourceWith {
+              src = ./.;
+              filter = srcFilter ./.;
+              name = "${name}-source";
+            };
+
+            # This is a hash of all the Cargo dependencies, for reproducibility.
+            cargoSha256 = "sha256-hmXhJBjc4HuyKQbxtpiIIvaL/Kl/e70sMFgdNlw4E0o=";
```
gpmayorga, line 99: Does this change often? Would it be worth it to add a one-liner comment with the command on how to update the hash?

gpmayorga (approved):
Looks good to me. I like the idea of reading settings from file within the Nix code. I added a question but nothing critical, approving the merge from my end

flake.nix
mikiquantum: When running the nix build, it will complain and will show the new hash.

PR #899, opened 2022-08-08
All ids in common-types

mustermeiszer:
This PR is a small change I would like to bring in. Currently, we manually define all kinds of ids, that we then define accounts from. 
We already had a bug when migrating to Altair as one of theses ids was changed unnoticed in a PR. 

Hence, I would like to have ALL ids that are relevant in the `common_types::ids` mod, so that changes to theses ids are noticed directly. 

### Note:
* There are also a few more ids in the bridge section that identify "wCFG". I would propose we also move them in common_types.
* This might also be a useful place to locate all fee-ids, that can then be used by other pallets to pass into the pallet-fee for getting fees @mikiquantum

### DIFF
--- a/libs/common-types/src/ids.rs
+++ b/libs/common-types/src/ids.rs
@@ -10,25 +10,25 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU General Public License for more details.
 
-///! A module that contains all ids where we should be REALLY careful
-///! when changing them.
+//! A module that contains all ids where we should be REALLY careful
+//! when changing them.
 use crate::PoolLocator;
-use sp_runtime::TypeId;
 use frame_support::PalletId;
+use sp_runtime::TypeId;
 
 // The TypeId impl we derive pool-accounts from
 impl<PoolId> TypeId for PoolLocator<PoolId> {
-    const TYPE_ID: [u8; 4] = *b"pool";
+	const const_ID: [u8; 4] = *b"pool";
 }
 
 // Pallet-Ids that define pallets accounts
-pub type PoolsPalletId = PalletId(*b"roc/pool");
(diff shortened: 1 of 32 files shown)


offerijns (approved):
100% agreed, LGTM

libs/common-types/src/ids.rs
```diff
@@ -0,0 +1,38 @@
...
+pub const CLAIMS_PALLET_ID: PalletId = PalletId(*b"p/claims");
+pub const CROWDLOAN_REWARD_PALLET_ID: PalletId = PalletId(*b"cc/rewrd");
+pub const CROWDLOAN_CLAIM_PALLET_ID: PalletId = PalletId(*b"cc/claim");
+pub const TREASURY_PALLET_ID: PalletId = PalletId(*b"py/trsry");
+pub const NFT_SALES_PALLET_ID: PalletId = PalletId(*b"pal/nfts");
+pub const STAKE_POT_PALLET_ID: PalletId = PalletId(*b"PotStake");
+
+// Other ids
+pub const CHAIN_BRIDGE_HASH_ID: [u8; 13] = b"cent_nft_hash";
+pub const CHAIN_BRIDGE_NATIVE_TOKEN_ID: [u8; 4] = b"xCFG";
```
mustermeiszer, line 38: @mikiquantum I moved those here, too. If you want them back hardcoded in each runtime I can revert this.

mustermeiszer:
@branan I hope only the nix-build will fail now. Could you take care of this?

branan (approved):
As long as the nix build actually works, I'm happy for you to merge this during your day tomorrow.

mustermeiszer:
@mikiquantum if you are fine with moving the bridge ids, feel free to hit the green button.

mikiquantum (approved):
🙏

PR #900, opened 2022-08-08
At at-param to rpc calls

### LINKED ISSUE #877: Modify RPC endpoints to add `at?: BlockHash`
Starting with token price RPC method.

Required for calls from Subquery: https://doc.subquery.network/build/mapping.html#custom-substrate-chains

mustermeiszer:
This PR is a minor change. It adds the option to provide an "at" parameter, at which the state the runtime-api should be called. 

I have not tested, but the API should error-out with an "UnknownState", when the block is unknown to the node. Otherwise, we can also query the status before calling the API and return early.

Closes #877

## Tasks
- [x] Test return when calling at non-exisiting block

### DIFF
--- a/src/rpc/anchors.rs
+++ b/src/rpc/anchors.rs
@@ -1,6 +1,6 @@
 use jsonrpsee::{core::RpcResult, proc_macros::rpc};
 
-use node_primitives::{BlockNumber, Hash};
+use node_primitives::BlockNumber;
 use pallet_anchors::AnchorData;
 use sp_api::ProvideRuntimeApi;
 use sp_blockchain::HeaderBackend;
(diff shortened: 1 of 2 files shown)


mustermeiszer:
@offerijns not sure if Apps would rather just use a `BlockNumber` as a parameter. But using `BlockId` seemed more straightforward, as they can use both hash and number. LMK if I should change that.

offerijns:
> @offerijns not sure if Apps would rather just use a `BlockNumber` as a parameter. But using `BlockId` seemed more straightforward, as they can use both hash and number. LMK if I should change that.

@mustermeiszer the main use case for this right now is calling these RPC methods from subquery. It looks like most existing FRAME RPC methods use either `Option<BlockNumber>` or `Option<BlockHash>`, and this is what subql assumes right now: https://github.com/subquery/subql/pull/788 So we'd need to use either one of them if I see it correctly.

mustermeiszer:
Will adapt accordingly

mustermeiszer:
@gruberb if you want to spin up a local-env and play with the chain a bit. Testing the rpcs would be a good use of that. We can do it together, then we test this and I can show you the process.

offerijns (dismissed):
LGTM!

mustermeiszer:
@gruberb or @offerijns can you give me thumbs up after the rebase?

offerijns (approved):

Issue #842, opened 2022-06-17
Register tranche token in asset-registry


offerijns:
- [x] Add `create_asset_metadata(pool_currency, token_name, token_symbol): AssetMetadata` method to `tranche.rs`

Add `AssetRegistry.Inspect` trait to `pools` runtime parameters. Call `Inspect.metadata(pool.currency)` to retrieve the pool currency decimals.

Add `ParachainId`constant to `pools` runtime parameters.

This will accept pool currency, token name and token symbol as arguments and return the following:

```rust
asset_metadata = {
  decimals: pool_currency.decimals,
  name: token_name,
  symbol: token_symbol,
  existential_deposit: Zero::zero(),
  location: Some(MultiLocation::new(
    1,
      X2(
        Parachain(parachain_id),
        GeneralKey(tranche_id.to_vec()),
      ),
    );
  ),
  additional: {
    mintable: false,
    permissioned: false,
    pool_currency: false,
    xcm: {
      fee_per_second: None,
    }
  },
}
```

- [x] Register tranche tokens on pool creation

Add `MaxTokenNameLength` and `MaxTokenSymbolLength` runtime parameters to the `pools` pallet.

Adapt https://github.com/centrifuge/centrifuge-chain/blob/parachain/pallets/pools/src/tranche.rs#L58 to
```rust
pub struct TrancheMetadata<MaxTokenNameLength, MaxTokenSymbolLength> {
	pub token_name: BoundedVec<u8, MaxTokenNameLength>,
	pub token_symbol: BoundedVec<u8, MaxTokenSymbolLength>,
}

pub struct TrancheInput<Rate, MaxTokenNameLength, MaxTokenSymbolLength> {
	pub tranche_type: TrancheType<Rage>,
	pub seniority: Option<Seniority>,
	pub metadata: TrancheMetadata<MaxTokenNameLength, MaxTokenSymbolLength>,
}
```

Add `AssetRegistry.Mutate` trait to `pools` runtime parameters.

Call [`register_asset ` ](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/asset-registry/src/impls.rs#L199) for each tranche.

- [x] Update token metadata on `pools.update`

Adapt https://github.com/centrifuge/centrifuge-chain/blob/parachain/pallets/pools/src/lib.rs#L115 to:
```rust
pub struct PoolChanges<Rate, MaxTokenNameLength, MaxTokenSymbolLength> {
[...]

PR #901, opened 2022-08-08
Register tranche token

gruberb:
This addresses https://github.com/centrifuge/centrifuge-chain/issues/842


pallets/pools/src/lib.rs
```diff
@@ -1524,20 +1607,57 @@ pub mod pallet {
...
+							if let Some(new_seniority) = tranche_update.seniority {
 								tranche.seniority = new_seniority.clone();
 							}
+
 							Ok(())
 						},
 					)?;
 				}
 
+				// Checking if we have a new tranche?
```
gruberb, line 1625: I checked via the UI, and it seems that we just send over the updated or new tranches. Would need a hint on how a new tranche is marked or if we indeed get all current tranches in the pool (updated + new) from the call. Maybe @offerijns @branan @mustermeiszer know more?

pallets/pools/src/lib.rs
```diff
@@ -1524,20 +1607,57 @@ pub mod pallet {
...
+
+				//
+				// The case when Metadata AND the tranche changed, we don't allow for an or. Both have to be changed (for now)
+				//
+				if let Change::NewValue(metadata) = &changes.tranche_metadata {
```
gruberb, line 1630: That's my attempt to update metadata for each tranche. WhatI could not figure out yet how these two are associated. I assume we need the `tranche_id`, could not figure out yet how to get it, since the `tranche` from the parameters don't have the `tranche_id`.

pallets/pools/src/tranche.rs
```diff
@@ -1453,6 +1490,41 @@ where
 	Ok(risk_buffers)
 }
 
+pub(crate) fn create_asset_metadata<Balance, CurrencyId>(
```
gruberb, line 1493: Made this a standalone function since it doesn't really relate to any type.

pallets/pools/src/tranche.rs
```diff
@@ -1453,6 +1490,41 @@ where
...
+	token_name: Vec<u8>,
+	token_symbol: Vec<u8>,
+) -> Result<AssetMetadata<Balance, CustomMetadata>, DispatchError>
+where
+	Balance: Zero,
+	CurrencyId: Encode,
+	CustomMetadata: Parameter + Member + TypeInfo,
+{
+	// Is this the right ID?
+	let tranche_id = currency.encode();
```
gruberb, line 1506: Is this the right `tranche_id` @branan @mustermeiszer ?

mustermeiszer: I think it is correct, as the `CurrencyId::Tranche(PoolId, TrancheId)` is a unique Id in our system, while `tranche_id`'s are only unique per pool. 

cc: @NunoAlexandre for clarity, can you confirm that the `GeneralKey` we are using the id for should have this structure and properties?

pallets/pools/src/lib.rs
mustermeiszer: We currently do not support adding or removing of tranches. So the vector can be mapped one-to-one to the existing tranches.

If we wanna support adding, removing, inserting tranches, we need to change more.

pallets/pools/src/lib.rs
```diff
@@ -1524,20 +1607,57 @@ pub mod pallet {
...
+						// It seems like we need to associate metadata with a tranche? I can't seem to find the relation.
+						// The tranche_id in the location? But where does tranche_id come from?
+						for (_, updated_metadata) in tranches.iter().zip(metadata.iter()) {
+							let decimals =
+								T::AssetRegistry::metadata(&pool.currency).unwrap().decimals;
+							let parachain_id = T::ParachainId::get();
+
+							let m = create_asset_metadata(
+								decimals,
+								pool.currency,
```
mustermeiszer: This must be `tranche.currency`

pallets/pools/src/lib.rs
```diff
@@ -1524,20 +1607,57 @@ pub mod pallet {
...
+								updated_metadata.clone().token_symbol.to_vec(),
+							)?;
+
+							assert_ok!(T::AssetRegistry::update_asset(
+								pool.currency,
+								Some(m.decimals),
+								Some(m.name),
+								Some(m.symbol),
+								Some(m.existential_deposit),
+								Some(m.location),
```
mustermeiszer, line 1652: It should be the same location, but I am not sure, we should allow to change the location in our domain. WDYT @NunoAlexandre?

mustermeiszer (commented):
Just a first look. Looks good already. Tried to answer most questions, I think Nuno can clarify the open questions.

pallets/pools/src/lib.rs
NunoAlexandre: @gruberb @mustermeiszer yeah, changing the MultiLocation on a tranche token would cause some serious issue with other projects that would integrate with us. This location is essentially the id of the asset in the ecosystem, so if we change it everyone else needs to change it as well as move the balances associated with the old location/internal representation of the asset to the new one.

Cargo.toml
```diff
@@ -186,17 +187,19 @@ runtime-benchmarks = [
 	"centrifuge-runtime/runtime-benchmarks",
 	"development-runtime/runtime-benchmarks",
 	"runtime-common/runtime-benchmarks",
+
 ]
 test-benchmarks = [
 	"runtime-benchmarks",
 	"runtime-integration-tests/runtime-benchmarks",
+	"pallet-pools/runtime-benchmarks",
```
branan, line 195: I don't think we need this at this level - it should either be selected manually (by running `cargo test -p pallet-pools --features runtime-benchmarks`), or be transitive via features on the pallets which depend on `pallet-pools` (namely the runtimes)

pallets/pools/src/lib.rs
```diff
@@ -391,8 +427,12 @@ pub mod pallet {
 
 	#[pallet::storage]
 	#[pallet::getter(fn scheduled_update)]
-	pub type ScheduledUpdate<T: Config> =
-		StorageMap<_, Blake2_128Concat, T::PoolId, ScheduledUpdateDetails<T::InterestRate>>;
+	pub type ScheduledUpdate<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat,
+		T::PoolId,
+		ScheduledUpdateDetails<T::InterestRate, T::MaxTokenNameLength, T::MaxTokenSymbolLength>,
```
branan, line 434: Technically this requires a migration, though I believe as long as we coordinate with the active pool on Altair to ensure we don't have a scheduled pool update when we run the runtime upgrade, we can get away with it (as this map will be empty) /cc @mustermeiszer does that sound right to you?

pallets/pools/src/lib.rs
```diff
@@ -628,6 +684,42 @@ pub mod pallet {
...
+			for (tranche, tranche_input) in tranches.tranches.iter().zip(&tranche_inputs) {
+				let token_name: BoundedVec<u8, T::MaxTokenNameLength> = tranche_input
+					.clone()
+					.metadata
+					.token_name
+					.try_into()
+					.map_err(|_| Error::<T>::BadMetadata)?;
+
+				let token_symbol: BoundedVec<u8, T::MaxTokenSymbolLength> = tranche_input
+					.clone()
```
branan, line 696: we're cloning `tranche_input` twice here - does it make more sense to do `tranche_input.metadata.token_{name,symbol}.clone()` for each value instead, or to do `let tranche_input = tranche_input.clone()` then extract the values from it?

pallets/pools/src/lib.rs
```diff
@@ -628,6 +684,42 @@ pub mod pallet {
...
+					.metadata
+					.token_symbol
+					.try_into()
+					.map_err(|_| Error::<T>::BadMetadata)?;
+
+				let decimals = match T::AssetRegistry::metadata(&currency) {
+					Some(metadata) => metadata.decimals,
+					None => return Err(Error::<T>::BadMetadata.into()),
```
branan, line 704: I think we should make this a different error, since the user action for it is different (fixing the currency metadata vs. submitting correct tranche metadata)

branan (changes requested):
Left some comments - the meat of this looks great, and is easy to follow :+1:

I think it'd be good to add a couple of tests to the pools pallet verifying that metadata at creation and metadata update are all correctly propagated to the asset registry pallet. It doesn't have to be much more than a create then assert. I think

pallets/pools/src/lib.rs
```diff
@@ -365,6 +393,14 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxSizeMetadata: Get<u32> + Copy + Member + scale_info::TypeInfo;
 
+		/// Max length of token name
```
NunoAlexandre, line 396:
```suggestion
		/// Max length for a tranche token name
```

pallets/pools/src/lib.rs
```diff
@@ -365,6 +393,14 @@ pub mod pallet {
...
+		#[pallet::constant]
+		type MaxTokenNameLength: Get<u32> + Copy + Member + scale_info::TypeInfo;
+
+		/// Max length token symbol
```
NunoAlexandre, line 400:
```suggestion
		/// Max length of a tranche token symbol
```

pallets/pools/src/lib.rs
```diff
@@ -628,6 +684,42 @@ pub mod pallet {
 				None => None,
 			};
 
+			for (tranche, tranche_input) in tranches.tranches.iter().zip(&tranche_inputs) {
+				let token_name: BoundedVec<u8, T::MaxTokenNameLength> = tranche_input
+					.clone()
+					.metadata
+					.token_name
+					.try_into()
+					.map_err(|_| Error::<T>::BadMetadata)?;
```
NunoAlexandre, line 693: I would be more granular with this error. 

```suggestion
					.map_err(|_| Error::<T>::TrancheTokenNameTooLong)?;
```

pallets/pools/src/lib.rs
NunoAlexandre: Yes exactly ^

pallets/pools/src/lib.rs
```diff
@@ -628,6 +684,42 @@ pub mod pallet {
...
+					tranche.currency,
+					parachain_id,
+					token_name.to_vec(),
+					token_symbol.to_vec(),
+				)?;
+
+				assert_ok!(T::AssetRegistry::register_asset(
+					Some(tranche.currency),
+					metadata
+				));
```
NunoAlexandre, line 720: It's the first time I am seeing an `assert_ok` used in an extrinsic. This will panic if it fails, right? That should not happen, we need to handle the failure here. Something like:

```suggestion
				T::AssetRegistry::register_asset(
					Some(tranche.currency),
					metadata
				).map_err(|_| Error::<T>::FailedToRegisterTrancheMetadata)?;
```

pallets/pools/src/lib.rs
```diff
@@ -705,6 +797,15 @@ pub mod pallet {
...
 
+			// Both changes.tranches and changes.tranche_metadata
+			// have to be NoChange or Change, we don't allow to change either or
+			// ^ = XOR, !^ = negated XOR
+			ensure!(
+				!((changes.tranches == Change::NoChange)
+					^ (changes.tranche_metadata == Change::NoChange)),
+				Error::<T>::UpdatePrerequesitesNotFulfilled
+			);
+
```
NunoAlexandre, line 808: 👍 

Not sure about the `UpdatePrerequesitesNotFulfilled` error name tho. Maybe `UnmetUpdatePrerequesite`?

pallets/pools/src/lib.rs
```diff
@@ -1524,20 +1625,54 @@ pub mod pallet {
...
+							parachain_id,
+							updated_metadata.clone().token_name.to_vec(),
+							updated_metadata.clone().token_symbol.to_vec(),
+						)?;
+
+						assert_ok!(T::AssetRegistry::update_asset(
+							tranche.currency,
+							Some(m.decimals),
+							Some(m.name),
+							Some(m.symbol),
```
NunoAlexandre, line 1668: @offerijns are we really sure we want to allow for the decimals, name, and symbol to be updated? This would break cross-chain integrations.

pallets/pools/src/tranche.rs
```diff
@@ -55,7 +61,33 @@ pub(super) type TrancheOf<T> = Tranche<
...
-pub type TrancheInput<Rate> = (TrancheType<Rate>, Option<Seniority>);
+
+#[derive(Debug, Encode, PartialEq, Eq, Decode, Clone, TypeInfo)]
+pub struct TrancheMetadata<MaxTokenNameLength, MaxTokenSymbolLength>
+where
+	MaxTokenNameLength: Get<u32>,
+	MaxTokenSymbolLength: Get<u32>,
+{
+	pub token_name: BoundedVec<u8, MaxTokenNameLength>,
+	pub token_symbol: BoundedVec<u8, MaxTokenSymbolLength>,
```
NunoAlexandre, line 72: Given the context is sufficient here, I would drop the `token_` prefixes. Leaving it up to you tho

pallets/pools/src/tranche.rs
```diff
@@ -1453,6 +1490,40 @@ where
 	Ok(risk_buffers)
 }
 
+pub(crate) fn create_asset_metadata<Balance, CurrencyId>(
+	decimals: u32,
+	currency: CurrencyId,
+	parachain_id: ParaId,
+	token_name: Vec<u8>,
+	token_symbol: Vec<u8>,
+) -> Result<AssetMetadata<Balance, CustomMetadata>, DispatchError>
```
NunoAlexandre, line 1499: It seems like this never fails so this could return just `AssetMetadata<Balance, CustomMetadata>`, right?

pallets/pools/src/tranche.rs
```diff
@@ -1453,6 +1490,40 @@ where
...
+		name: token_name,
+		symbol: token_symbol,
+		existential_deposit: Zero::zero(),
+		location: Some(VersionedMultiLocation::V1(MultiLocation {
+			parents: 1,
+			interior: X2(Parachain(parachain_id.into()), GeneralKey(tranche_id)),
+		})),
+		additional: CustomMetadata {
+			mintable: false,
+			permissioned: false,
```
NunoAlexandre, line 1518:
```suggestion
			permissioned: true,
```

runtime/altair/src/lib.rs
```diff
@@ -1035,6 +1035,12 @@ parameter_types! {
 	#[derive(scale_info::TypeInfo, Eq, PartialEq, Debug, Clone, Copy )]
 	pub const MaxSizeMetadata: u32 = 46; // length of IPFS hash
 
+	#[derive(scale_info::TypeInfo, Eq, PartialEq, Debug, Clone, Copy )]
+	pub const MaxTokenNameLength: u32 = 128;
+
+	#[derive(scale_info::TypeInfo, Eq, PartialEq, Debug, Clone, Copy )]
+	pub const MaxTokenSymbolLength: u32 = 128;
```
NunoAlexandre, line 1042: The symbol can be much smaller. I haven't seen token symbols longer than 5 chars.

runtime/common/src/lib.rs
```diff
@@ -131,6 +131,10 @@ pub mod types {
 	/// Rate with 27 precision fixed point decimal
 	pub type Rate = crate::fixed_point::Rate;
 
+	pub type MaxTokenNameLength = u32;
```
NunoAlexandre, line 134: Could you document these type aliases? Out of context it's not trivial to know what they refer to.

NunoAlexandre (commented):
Good work so far @gruberb 🚀

pallets/pools/src/lib.rs
offerijns: Tranche token decimals should just copy the decimals of the pool currency, which in itself could in theory be updated but we probably would just not allow this on a governance level. I still need to take a look at the PR code to review how/whether this is now implemented this way.

pallets/pools/src/tranche.rs
offerijns: I don't think we should, since we will likely add a custom token permission check in restricted-tokens at some point, which applies to tokens which have this flag set to `true`. We will then also still have the tranche token specific permission checks. This would mean that we are checking the permissions for a tranche token twice.

runtime/altair/src/lib.rs
offerijns: In the UI we are allowing tranche token symbols up to 12 chars (that's the limit that Metamask allows). We'll definitely get past 5 chars, e.g. CGFF1-DROP was a requested symbol in the past.

pallets/pools/src/tranche.rs
NunoAlexandre: > We will then also still have the tranche token specific permission checks
I hear you, but those are not mutually exclusive properties, as we can still have a tranche-specific check. For example:

``` rust
if metadata.additional.permissioned {
  match currency {
    CurrencyId::Tranche(_, _) => specific_tranche_check(...),
    _ => default_check(...),
  }
}
```

This `permissioned` property was initially added to mitigate the requirement of flagging the permission requirements of tranche tokens, so if we now set this to `false` it sort of defeats the purpose. Which is fine if we no longer want it, but it seems off to me to have two sources of truth for permission flagging (one with this field, the other being implicit throughout the code). What are your thoughts?

pallets/pools/src/lib.rs
gruberb: I can change it. I just used the pre-existing error name `UpdatePrerequesitesNotFulfilled` since I thought it is getting confusing to have slightly different named errors.

pallets/pools/src/tranche.rs
mustermeiszer: But the context is not given, when we parametrized the pallet in the runtimes. So I would like to keep it.

mustermeiszer (approved):
I would like to clarify the issue with the update of decimals, name, and symbol that @NunoAlexandre brought up, before merging this. Otherwise, looks good. Nice one, @gruberb!

pallets/pools/src/lib.rs
NunoAlexandre: The decimals is a hard issue, the name and symbol is a soft issue; as in, if those are updated, other chains that may integrate with us and have tranche tokens will may display the outdated name and/or symbol which is a UX problem.

> Regarding the decimals, I think this could really be an issue and would like to make in non-updatable. To easy to miss and easy to add in the future.

💯 agree

pallets/pools/src/tranche.rs
NunoAlexandre: You are forgiven, my favourite German doughnut!

libs/common-types/src/lib.rs
```diff
@@ -43,6 +45,9 @@ pub type PoolId = u64;
 /// A representation of a tranche identifier
 pub type TrancheId = [u8; 16];
 
+/// Balance of an account.
+pub type Balance = u128;
```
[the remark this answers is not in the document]
NunoAlexandre: All in for that 👍

pallets/pools/src/lib.rs
```diff
@@ -628,6 +693,41 @@ pub mod pallet {
...
+				let token_symbol: BoundedVec<u8, T::MaxTokenSymbolLength> = tranche_input
+					.metadata
+					.token_symbol
+					.clone()
+					.try_into()
+					.map_err(|_| Error::<T>::TrancheTokenNameTooLong)?;
+
+				let decimals = match T::AssetRegistry::metadata(&currency) {
+					Some(metadata) => metadata.decimals,
+					None => return Err(Error::<T>::MetadataForCurrencyNoFound.into()),
```
offerijns, line 714:
```suggestion
					None => return Err(Error::<T>::MetadataForCurrencyNotFound.into()),
```

pallets/pools/src/lib.rs
```diff
@@ -705,6 +805,15 @@ pub mod pallet {
 				Error::<T>::InSubmissionPeriod
 			);
 
+			// Both changes.tranches and changes.tranche_metadata
+			// have to be NoChange or Change, we don't allow to change either or
+			// ^ = XOR, !^ = negated XOR
+			ensure!(
+				!((changes.tranches == Change::NoChange)
+					^ (changes.tranche_metadata == Change::NoChange)),
+				Error::<T>::UpdatePrerequesitesNotFulfilled
```
offerijns, line 814: `UpdatePrerequesitesNotFulfilled` has a different purpose: this is thrown when a scheduled update cannot be executed/fulfilled yet. We should use a custom error for this, something like `InvalidTrancheUpdate` probably?

offerijns (dismissed):
Two quick comments, but this is looking really good!

pallets/pools/src/lib.rs
mustermeiszer: Yes, I think it is reasonable to do it that way. 👍

pallets/pools/src/lib.rs
```diff
@@ -628,6 +696,41 @@ pub mod pallet {
...
+					.clone()
+					.try_into()
+					.map_err(|_| Error::<T>::TrancheTokenNameTooLong)?;
+
+				let token_symbol: BoundedVec<u8, T::MaxTokenSymbolLength> = tranche_input
+					.metadata
+					.token_symbol
+					.clone()
+					.try_into()
+					.map_err(|_| Error::<T>::TrancheTokenNameTooLong)?;
```
mustermeiszer, line 713: This should be `TrancheSymbolNameTooLong`

mustermeiszer (approved):
LGTM! Just one error returns the wrong variant. This should be fixed before merging.

mustermeiszer (approved):
💯

pallets/pools-registry/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+		/// invoke this extrinsic.
+		#[pallet::weight(T::WeightInfo::set_metadata(metadata.len().try_into().unwrap_or(u32::MAX)))]
+		pub fn set_metadata(
+			origin: OriginFor<T>,
+			pool_id: T::PoolId,
+			metadata: Vec<u8>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			ensure!(
+				T::Permission::has(
```
gruberb, line 180: What should the permissions be here? Is it still tightly coupled with `PoolAdmin`, should we introduce new permissions? @offerijns @mustermeiszer @branan

offerijns:
> * We have a `set_metadata` extrinsic in the new `pools-registry` pallet. It seems like we need to offer a `get_metadata` extrinsic as well?

Extrinsics are only needed whenever you want to mutate data on chain. The UI can just retrieve the `PoolMetadata` storage key directly. Similar to how there's no `get_pool` extrinsic on the pools pallet.

pallets/pools-registry/src/lib.rs
mustermeiszer: I went for something like

```rust
ensure!(
	T::PreConditions::check(OrderType::Investment {
		who: who.clone(),
		investment_id,
		amount
	}),
	BadOrigin
);
```

whenever I did not want to commit to anything specific in the pallet. Then the runtimes themselves can define what they want to do. I am not sure, if this is the best idea, but its a clean cut, as the registry must not now about roles.

pallets/pools/src/lib.rs
```diff
@@ -628,6 +698,41 @@ pub mod pallet {
...
+					tranche.currency,
+					parachain_id,
+					token_name.to_vec(),
+					token_symbol.to_vec(),
+				);
+
+				T::AssetRegistry::register_asset(Some(tranche.currency), metadata)
+					.map_err(|_| Error::<T>::FailedToRegisterTrancheMetadata)?;
+			}
+
```
mustermeiszer, line 735: It's not really possible to reduce this completely to a generic logic as we need `AssetRegistry` here. 

You could of course write something like `fn register_asset<Registry: orml_traits::registry::AssetRegistry()` but this feels a like it should be part of the pallet. 

What I think we can move to common is the boundaries check for vectors. With something like `fn respect_bounds<T, Bound: Get<u32>, E: Into<DispatchError>(vectors: Vec<T>) -> Result<BoundedVec<T, Bound>, DispatchError> {...}`. This would allow you to do the checks without the need to bind to `BoundedVec<T>` explicitly. 

Regarding the implementation. I think it would make sense to define a trait so that pallets can register pool info. We will most likely need this for the index-token too. And its light-coupling

gruberb:
Closing this now. Will reopen two new PRs:

- One PR which overs introducing `TrancheMetadata` to the codebase
- Another one which will introduce a new pallet, `pallet-pool-registry`

The current one blew up and touched many pallets and files. The constant rebasing and updating other PRs after they were merged was more work than the actual PR in the end.

Issue #897, opened 2022-08-07
Replace all `Vec`s with `BoundedVec`s


offerijns:
For [Weights V2](https://github.com/paritytech/substrate/pull/10918), we will (eventually) need to add `MaxEncodedLen` to all storage values. This means replacing all `Vec`s in storage with `BoundedVec`s and adding runtime parameters (or hardcoded consts when it makes more sense) for these bounds.

On first review, it looks like there's currently unbounded `Vec`s in these pallets:
- [x] anchors pallet: https://github.com/centrifuge/centrifuge-chain/pull/980
- [x] ~~bridge pallet~~
- [x] loans pallet: https://github.com/centrifuge/centrifuge-chain/pull/990
- [x] pools pallet: https://github.com/centrifuge/centrifuge-chain/pull/979

lemunozm:
`pallet-bridge` seems to not use any Vec in its storage, only in the `GenesisConfig`, but I think this is not related to WeightsV2.

lemunozm:
`pallet-loans` has a `Vec` in its storage, inside of the `WriteOffGroup` type. Seems like it's conceptually something unlimited. Anybody could add any number of write-offs using `add_write_off_group()`. 

@branan @mustermeiszer @gruberb Any recommendation to tackle this? We also need to give some kind of error in `add_write_off_group()` if we reach the new bound.

mustermeiszer:
@lemunozm we have not yet fully speced the changes to write-off groups. I think to unblock you here, the simplest would be to add a bounding associated type and simply error out with `TooManyWriteOfGroups`. We currently do not allow to remove them, so they would be stuck in this case, but we will refactor this code anyways heavily, so no need to spend much time on this now.

offerijns:
@lemunozm This limit already is implemented: https://github.com/centrifuge/centrifuge-chain/blob/parachain/pallets/loans/src/functions.rs#L637 The `BoundedVec` should just be bounded by `T:: MaxWriteOffGroups`.

lemunozm:
I must be blind, you're right. Thanks @offerijns!

PR #905, opened 2022-08-10
Migrate to Weights V2

akhilesharora:
As @offerijns already mentioned this [issue](https://github.com/centrifuge/centrifuge-chain/issues/897), storage migrations from Vec<T> to BoundedVec<T> for few of the remaining pallets as per [Weights V2](https://github.com/paritytech/substrate/pull/10918).


pallets/anchors/src/lib.rs
```diff
@@ -306,7 +310,7 @@ pub mod pallet {
 				.ok_or(ArithmeticError::Underflow)?;
 
 			let fee = base_fee
-				.checked_mul(&pallet_fees::BalanceOf::<T>::from(multiplier))
+				.checked_mul(&pallet_fees::BalanceOf::<T>::from(multiplier as u8))
```
mustermeiszer, line 313: Why is the as u8 needed now? I would like to use a `try_into().ok_or(ArithmethicError::Overflow)` here, as u8 is really small to cast to.

pallets/anchors/src/lib.rs
```diff
@@ -473,7 +477,7 @@ impl<T: Config> Pallet<T> {
 					.checked_add(expiration_horizon)
 					.ok_or(ArithmeticError::Overflow)
 			})
-			.and_then(|put_into_bucket| Ok(T::BlockNumber::from(put_into_bucket)))
+			.and_then(|put_into_bucket| Ok(T::BlockNumber::from(put_into_bucket as u8)))
```
mustermeiszer, line 480: See abve

pallets/anchors/src/lib.rs
```diff
@@ -490,8 +494,10 @@ impl<T: Config> Pallet<T> {
 			// store the root of child trie for the day on chain before eviction. Checks if it
 			// exists before hand to ensure that it doesn't overwrite a root.
 			.map(|(day, key)| {
+				let val: BoundedVec<u8, T::MaxBound> =
+					(child::root(&key, StateVersion::V0)).try_into().unwrap();
```
mustermeiszer, line 498: `unwrap()` should be handled

runtime/development/src/lib.rs
```diff
@@ -791,8 +791,14 @@ impl pallet_fees::Config for Runtime {
 	type WeightInfo = weights::pallet_fees::SubstrateWeight<Self>;
 }
 
+// Parameterize anchors pallet
+parameter_types! {
+	pub const MaxBound: u32 = 100;
```
mustermeiszer, line 796: I think the child root is a `H256`, so 256 bytes. The bound should allow this size if I am not mistaken. WDYT? @branan

mustermeiszer (changes requested):
One `unwrap()` and castings neeed to be handled. Also, I think the `MaxBound` is to small

pallets/anchors/src/lib.rs
akhilesharora: Ah sure, well while testing it was complaining about the types. But indeed, I was thinking along the same lines as `u8` is really small to cast to.

pallets/anchors/src/lib.rs
```diff
@@ -100,6 +100,9 @@ pub mod pallet {
 	{
 		/// Type representing the weight of this pallet
 		type WeightInfo: WeightInfo;
+
+		#[pallet::constant]
+		type MaxBound: Get<u32>;
```
lemunozm, line 105: Maybe this name can be renamed to something more relevant for the runtime configurator in order to know what is exactly this bound for. Something like `MaxRootOfChildTrieSize`.

Because our `Vec<u8>` is storing the child root (docs say that the hashing algorithm is defined by the Block), seems like its size is something known. I mean, instead of passing a new argument, could we extract it from the current hashing algorithm used?

lemunozm: @mustermeiszer I just see your comment below regarding the size, what do you think about this? Could it be possible?

mustermeiszer: Love the idea, but is is defined not by `T::Hash`, but by the `type Hash` of `trait Header`.  As the externalities hasher defines this hash type.

lemunozm: Done in https://github.com/centrifuge/centrifuge-chain/pull/980

lemunozm:
Closed in favor of https://github.com/centrifuge/centrifuge-chain/pull/980

PR #906, opened 2022-08-12
Removes the default `LoanType`

mustermeiszer:
Minor change: Just doesn't like default type for such a complex type and it was not used anywhere.


branan (approved):
Makes sense to me to get rid of it. We should never be using `default` for something like this.

This needs to be rebased or merged from parachain, then I'm happy to hit the button once CI runs again today

mustermeiszer:
@branan just rebased.

branan (approved):

PR #909, opened 2022-08-16
Pallet-Investments

mustermeiszer:
The investment pallet that should take care of:

Providing orders for assets
allowing users to collect these orders

## Implementation Note
The `AssetManager` implementation should implement for a specific asset, while the AssetId in the pallet-investments can get an enum of
pub enum AssetId<PoolId, TrancheId>{
    Pool(PoolId, TrancheLoc<TrancheId>)
}


mustermeiszer:
This PR solely contains the pallet-investments. Currently, no logic from pallet-pools is removed and replaced by the pallet to make this PR easier to be reviewed. 
A follow-up PR will integrate the pallet into our existing pool.

mustermeiszer:
This should be ready for review. Contains just the basic business logic for investing. No tests, benchmarks, weights. No integrations.

mustermeiszer:
@branan the Cargo.lock file is insanely changed. I am not sure why. Do you have an idea? Seems like cargo define a really specific version for many substrate devs, which I am not sure why. Bet I have a misconfig in the pallets Cargo.toml. 

^^

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,887 @@
...
+mod mock;
+#[cfg(test)]
+mod tests;
+
+type CurrencyOf<T> =
+	<<T as Config>::Tokens as Inspect<<T as frame_system::Config>::AccountId>>::AssetId;
+
+/// The outstanding collections for an account
+#[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+pub struct Collection<Balance> {
```
offerijns, line 46: The naming of these fields is much better than what we had before 👍

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,887 @@
...
+
+/// The enum we parse to `PreConditions` so the runtime
+/// can make an educated decision about this investment
+pub enum OrderType<AccountId, AssetId, Amount> {
+	Investment {
+		who: AccountId,
+		asset: AssetId,
+		amount: Amount,
+	},
+	Redeemption {
```
offerijns, line 91:
```suggestion
	Redemption {
```

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,887 @@
...
+
+	pub(crate) fn do_update_invest_order(
+		total_order: &mut TotalOrder<T::Amount>,
+		who: &T::AccountId,
+		asset_id: T::AssetId,
+		info: impl AssetProperties<T::AccountId, Currency = CurrencyOf<T>>,
+		order: &mut OrderOf<T>,
+		amount: T::Amount,
+	) -> DispatchResult {
+		let asset_account = AssetAccount { asset_id }.into_account_truncating();
```
offerijns, line 674: Currently, all tokens are stored in the `PoolId` derived account. I think we should keep it this way. The advantage of this is that (once Subscan properly supports token balances, which it should in the short term), you can easily see the entire balance of a pool on one account: the currency it holds, the tranche tokens, and all the collateral NFTs.

offerijns (commented):
Leaving a first set of minor comments, still trying to think through the approach more broadly.

libs/common-traits/src/lib.rs
```diff
@@ -278,3 +278,67 @@ impl<T> PreConditions<T> for Never {
 pub trait TrancheToken<PoolId, TrancheId, CurrencyId> {
 	fn tranche_token(pool: PoolId, tranche: TrancheId) -> CurrencyId;
 }
+
+pub trait InvestmentManager {
+	type Error;
+	type AssetId;
```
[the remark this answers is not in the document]
mustermeiszer: Yeah, definitely not a good one. What about `InvestmentId`?

pallets/investments/src/lib.rs
mustermeiszer: Makes Sense. I will adapt the asset management trait accordingly.

mustermeiszer: Wait. This is not for the pools to hold. It is basically the "in-between", when the pool has not yet signaled it will want to get investments. I think it makes more sense, to keep the "in-between" in this pallet. 

Wdyt?

mustermeiszer: Nah, you are right. THis should be pools. Although, it will eventually show inconsistent state then about "how much does the pool actually have". As it will contain orders that are not fulfilled, and might never be.

mustermeiszer: What about we transfer the tokens after the pool has signaled how much he wants to fulfill?

offerijns: Isn't that how it currently works? As in, my understanding would be: on epoch execution, the pool pallet mints the executed tranche tokens for investments into the pool account. This is the moment when the pool signals to the investments pallet what the fulfillment was, right?

branan: Right. The way it currently works is that the pool account holds tokens for the following three things:

* Tokens transferred _from_ investors as part of an invest or redeem
* Tokens yet to be claimed by investors after a successful investment / redemption
* Tokens that have been repaid by borrowers but are on-hold until an epoch close
* Tokens that are available for borrowers to borrow

The separation between those categories is handled entirely by the internal logic of the pools pallet (and is most of what the logic is about).

We could probably rejigger where some of that is stored so that it's clearer in tools like subscan what the state of the tokens is, but that adds extra transfer logic to lots of places where we currently just update a couple integers.

pallets/investments/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+scale-info = { version = "2.0", default-features = false, features = ["derive"] }
+
+frame-benchmarking = { git = "https://github.com/paritytech/substrate", default-features = false , optional = true , branch = "polkadot-v0.9.24" }
+frame-support = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.24" }
+frame-system = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.24" }
+pallet-timestamp = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.24" }
+sp-arithmetic = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.24" }
+sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.24" }
+sp-std = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.24" }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "master" }
```
branan, line 26: I think this is the line that dragged in a different version of polkadot. You want the `polkadot-v0.9.24` branch of ORML here.

pallets/investments/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+common-traits = { path = "../../libs/common-traits", default-features = false }
+common-types = { path = "../../libs/common-types", default-features = false }
+pallet-permissions = { path = "../../pallets/permissions", default-features = false}
+
+[dev-dependencies]
+sp-core = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.24" }
+sp-io = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.24" }
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "master" }
```
branan, line 34: As above for ORML versioning

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,887 @@
...
+		Blake2_128Concat,
+		T::AccountId,
+		Blake2_128Concat,
+		T::AssetId,
+		BoundedVec<Order<T::Amount, T::OrderId>, T::MaxOutstandingCollects>,
+		ValueQuery,
+	>;
+
+	#[pallet::type_value]
+	pub fn OnTotalOrderEmpty<T: Config>() -> TotalOrder<T::Amount>
```
branan, line 239: Why a custom value here instead of giving TotalOrder a `Default` impl for the `ValueQuery` to use?

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,887 @@
...
+				T::Tokens::transfer(
+					info.payment_currency(),
+					&asset_account,
+					&who,
+					collection.payout_asset_redeem,
+					false,
+				)?;
+
+				// drain the orders that have been collected
+				orders.retain(|order| !collected.contains(&order.id));
```
branan, line 579: Technically this is O(n^2) worst case, if we collect all orders.

I think the for loop above could be turned into the retain call, and we do the collection and removal all in one place?

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,887 @@
...
+
+				match collect_type {
+					CollectType::Overflowing => {
+						ActiveOrder::<T>::try_mutate(&asset_id, |total_order| -> DispatchResult {
+							if collection.remaining_asset_invest > T::Amount::zero() {
+								Orders::<T>::try_mutate(
```
branan, line 585: Aren't we already in a `try_mutate` block for orders here?

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,887 @@
...
+												.invest
+												.checked_add(&collection.remaining_asset_invest)
+												.ok_or(ArithmeticError::Overflow)?,
+										)
+									},
+								)?;
+							};
+
+							if collection.remaining_asset_redeem > T::Amount::zero() {
+								Orders::<T>::try_mutate(
```
branan, line 608: As above, this looks like a nested `try_mutate` of the same storage item

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,887 @@
...
+
+		/// Something that can handle payments and transfers of
+		/// currencies
+		type Tokens: Mutate<Self::AccountId>
+			+ Inspect<Self::AccountId, Balance = Self::Amount>
+			+ Transfer<Self::AccountId>;
+
+		/// A possible check if investors fulfill every condition to invest into a
+		/// given asset
+		type PreConditions: PreConditions<
```
branan, line 191: `PreConditions -> Preconditions`

branan (changes requested):
Some things that look like real bugs,so I'm flagging as `request changes` for now. But I'm broadly :+1: on this. It looks nice and clean and like a good separation of the investment logic from the current pools implementation to this.

pallets/investments/src/lib.rs
mustermeiszer: Why? ^^

mustermeiszer: Should we rename the trait to then?

pallets/investments/src/lib.rs
mustermeiszer: Good Idea!

pallets/investments/src/lib.rs
mustermeiszer: I logically made it like this. But because rust can not return early from these iterators it is still a for loop but O(2*n) now.

pallets/investments/src/lib.rs
mustermeiszer: Yes, corrected this!

pallets/investments/src/lib.rs
mustermeiszer: Good idea, done!

gruberb:
Seems to be nitpick, but can we use doc comments`/// `above `struct`s and `trait`s? Would help getting familiar with what the business implications etc are and how to use certain things!

Otherwise, have not much knowledge yet to comment on how it would affect the code base.

mustermeiszer:
> Seems to be nitpick, but can we use doc comments/// above structs and traits? Would help getting familiar with what the business implications etc are and how to use certain things!

Totally. I will add some docs in the next issues. Won't make it this week to this and don't wanna go through review again.

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+						.checked_mul_int(orders.redeem)
+						.ok_or(ArithmeticError::Overflow)?,
+				);
+				let asset_account = AssetAccount {
+					asset_id: asset_id.clone(),
+				}
+				.into_account_truncating();
+				let info = T::Accountant::info(asset_id.clone())?;
+
+				if invest >= redeem {
```
offerijns, line 841: Should this not be `invest > redeem`? If `invest == redeem`, then there's nothing to transfer.

pallets/investments/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+					// We subjectively decide here, that it is better not to error out if in the sequence of orders
+					// one has not yet been cleared.
+					//
+					// Instead we signal the collected orders via the vector if OrderId's in the event.
+					if let Ok(fulfillment) = ClearedOrders::<T>::try_get(asset_id.clone(), order.id)
+					{
+						Pallet::<T>::acc_payout_invest(&mut collection, &fulfillment, &order)?;
+						Pallet::<T>::acc_payout_redeem(&mut collection, &fulfillment, &order)?;
+						Pallet::<T>::acc_remaining_invest(&mut collection, &fulfillment, &order)?;
+						Pallet::<T>::acc_remaining_invest(&mut collection, &fulfillment, &order)?;
```
offerijns, line 536:
```suggestion
						Pallet::<T>::acc_remaining_redeem(&mut collection, &fulfillment, &order)?;
```

branan (approved):
Nothing new jumps out as weird/wrong to me, so I think this is ready to land

mustermeiszer:
Thanks a bunch for updating nix. I think it's getting time for me to set this up too...

mustermeiszer:
I will enable auto-merge then.

PR #926, opened 2022-08-25
readme: Document NIX builds

NunoAlexandre:


README.md
```diff
@@ -202,3 +202,24 @@ Tip: if you face compilation errors like "type X doesn't implement trait Y", and
...
+
+The CI runs a `nix-build` job that executes `nix build` on the entire repository. This job fails if the output 
+`cargoSha256` for the revision being built does not match the expected value defined in `flake.nix`.
+
+Whenever a dependency is added, removed, or updated, said `cargoSha256` value defined in `flake.nix` needs to be updated.
+To do that, you need:
+
+1. Install Nix
+   - [Guide for macos](https://www.mathiaspolligkeit.com/dev/exploring-nix-on-macos/)
+   - [ToDo Guide for Linux](...)
```
NunoAlexandre, line 217: Any suggestion for linux machines?

README.md
```diff
@@ -202,3 +202,24 @@ Tip: if you face compilation errors like "type X doesn't implement trait Y", and
...
+
+2. On your feature branch, run:
```
lemunozm, line 219: I'm not sure if before this step you need to update the `cargoSha256` value with 0s, or if any error value shows you later the correct one.

NunoAlexandre: I don't know what you mean with updating the value with 0s, how would that look like? FWIW, I fixed the build at #886 by updating the new `cargoSha256` that this command on step 2 outputted.

NunoAlexandre: @lemunozm an example of an output:

 ```
λ nix --extra-experimental-features "nix-command flakes" build
error: hash mismatch in fixed-output derivation '/nix/store/7pnkcpkxp6hzshw3ra772hk3ljn2incr-centrifuge-chain-0.10.18-vendor.tar.gz.drv':
         specified: sha256-pausA0+qcM6PEH1fS9/3Dn0GWC5LBnQXX+W8UVDbCj4=
            got:    sha256-wYfojyWdgMswa/pD5xxRcjILlDg74MfXErIk+08kL0c=
error: 1 dependencies of derivation '/nix/store/jvpks157rs9qg2w6mnmjx8047fkyzzqn-centrifuge-chain-0.10.18.drv' failed to build
```

lemunozm: Ohh, ok. I thought that you should substitute the hash by 0s in order to obtain the correct hash like it's specified here in this error example: https://github.com/centrifuge/centrifuge-chain/runs/8023256972?check_suite_focus=true but seems that it shows the correct one anyway. It's all ok then 😄

NunoAlexandre: ah yeah, I had to do that today actually :) all good 👍

README.md
lemunozm: What about linking the download official page? https://nixos.org/download.html

lemunozm (approved):

PR #930, opened 2022-08-31
Rustdocs

gruberb:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Create Index file
+        run: ./ci/create_index_for_rust_docs.sh
+
+      - name: Deploy Docs
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_branch: gh-pages
+          publish_dir: ./target/doc
+          cname: reference.centrifuge.io
```
gruberb, line 36: I don't actually know why the PR says "two new files". I added  the `cname` parameter.

gruberb: Ah, I should have opened a new branch. This shows the changes from last time as well. Hope we can still merge it like that?

NunoAlexandre: Maybe the base of this PR is not correct?

NunoAlexandre: Yeah it seems so

gruberb: Ahh of course, thank you!

offerijns (approved):

PR #933, opened 2022-08-31
Pallet bridge using new pallet fees

lemunozm:
**Tasks:**
- Update `pallet-bridge` to use the new `pallet-fees`.
- Clean unused dependencies and parameters from `pallet-bridge`.


pallets/bridge/src/lib.rs
```diff
@@ -285,12 +253,6 @@ pub mod pallet {
 	pub enum Error<T> {
 		/// Invalid transfer
 		InvalidTransfer,
-
-		/// Not enough means for performing a transfer
-		InsufficientBalance,
```
lemunozm, line 290: Removed because now is used the own `InsufficienteBalance` from `pallet-balance`.

pallets/bridge/src/lib.rs
```diff
@@ -167,12 +158,9 @@ pub mod pallet {
 	///
 	/// Associated types and constants are declared in this trait. If the pallet
 	/// depends on other super-traits, the latter must be added to this trait,
-	/// such as, in this case, [`pallet_nft::Config`] super-trait, for instance.
 	/// Note that [`frame_system::Config`] must always be included.
 	#[pallet::config]
-	pub trait Config:
-		frame_system::Config + chainbridge::Config + pallet_balances::Config + pallet_nft::Config
```
lemunozm, line 174: There is no need to be tight to `pallet_balances` and `pallet_nft`.

pallets/bridge/src/lib.rs
```diff
@@ -316,38 +278,13 @@ pub mod pallet {
...
-				.checked_add(&token_transfer_fee)
-				.ok_or(Error::<T>::TotalAmountOverflow)?;
-
-			// Ensure account has enough balance for both fee and transfer
-			// Check to avoid balance errors down the line that leave balance storage in an inconsistent state
-			let remaining_balance = <T as pallet::Config>::Currency::free_balance(&source)
-				.checked_sub(&total_transfer_amount)
-				.ok_or(Error::<T>::InsufficientBalance)?;
-
-			<T as pallet::Config>::Currency::ensure_can_withdraw(
```
lemunozm, line 333: This check is no longer needed because the call is marked as transactional.

runtime/centrifuge/src/lib.rs
```diff
@@ -806,22 +806,18 @@ impl pallet_nft::Config for Runtime {
 
 parameter_types! {
 	pub const BridgePalletId: PalletId = common_types::ids::BRIDGE_PALLET_ID;
-	pub HashId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &sp_io::hashing::blake2_128(&common_types::ids::CHAIN_BRIDGE_HASH_ID));
 	pub NativeTokenId: chainbridge::ResourceId = chainbridge::derive_resource_id(1, &sp_io::hashing::blake2_128(&common_types::ids::CHAIN_BRIDGE_NATIVE_TOKEN_ID));
-	pub const NativeTokenTransferFee: u128 = NATIVE_TOKEN_TRANSFER_FEE;
-	pub const NftTransferFee: u128 = NFT_TOKEN_TRANSFER_FEE;
+	pub const NativeTokenTransferFeeKey: FeeKey = FeeKey::NativeTokenTransfer;
```
lemunozm, line 810: **Careful!**: The new balance is 1 * CFG (The default value for the storage of `pallet-keys`), and the old value is `NATIVE_TOKEN_TRANSFER_FEE` which is 2000 * CFG. Is this a problem? Should we use some kind of runtime upgrade?

mikiquantum: I think as soon as it is enacted we can trigger a council motion to correct it. I dont think is a huge issue. Alternatively, what we can do is to split this PR in 2, one to add the Key enum, then we populate the new value and two to add the code changes.

lemunozm: It seems interesting the idea 👍🏻. Tomorrow we can talk deeply about it.

mikiquantum (approved):
lgtm!

mikiquantum (approved):
👍

PR #944, opened 2022-09-02
Rewards: pallet-rewards implementation

### LINKED ISSUE #960: Rewards: pallet-rewards implementation

### Description

Implement `pallet-rewards`. This implies:
- Implement the `Rewards` interface: https://github.com/centrifuge/centrifuge-chain/issues/1035
- Add Group/Currencies support
- Testing

### Research/based on
- See [specs](https://centrifuge.hackmd.io/9T9zAZaTRD6ZcHN-Gd_lzw?view)
- See [scalable-reward-changing](https://solmaz.io/2019/02/24/scalable-reward-changing/) paper
- See [these slides](https://docs.google.com/presentation/d/1xDodBX_u8zQg0SL5VaA4z0okb7u4_g08LHK2S6NEdbU/edit#slide=id.p) to implement the currency movement.

lemunozm:
# Description

Implementation of `pallet-rewards`.

Fixes #960

## Changes and Descriptions

Blocked by https://github.com/centrifuge/centrifuge-chain/pull/1036

- [x] Stake/Unstake/Claim/Distribute basic methods.
- [x] Several groups support.
- [x] Several currencies
    - [x] Implementation
    - [x] Transfers
    - [x] Movement
- [x] Unitary tests
- [x] Mock
- [x] Tests
- [x] Events
- [x] Docs

## Type of change

# How Has This Been Tested?
```rust
cargo test -p pallet-rewards
```

# Checklist:


mikiquantum (approved):
Looks accurate to me as the base implementation of the algorithm from the paper. 👍

pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+
+		#[pallet::constant]
+		type PalletId: Get<PalletId>;
+
+		#[pallet::constant]
+		type BlockPerEpoch: Get<Self::BlockNumber>;
+
+		type Currency: ReservableCurrency<Self::AccountId>;
+
+		type SignedBalance: From<BalanceOf<Self>>
```
mikiquantum, line 80: What is the purpose of a `Signed` Balance?

mikiquantum, line 80: What is the purpose of `Signed` Balance?

pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// Type used to initialize the first epoch with the correct block number
+pub struct FirstEpochDetails<P>(std::marker::PhantomData<P>);
+impl<P, N, B: Zero> Get<EpochDetails<N, B>> for FirstEpochDetails<P>
+where
+	P: BlockNumberProvider<BlockNumber = N>,
+{
+	fn get() -> EpochDetails<N, B> {
+		EpochDetails {
+			ends_on: P::current_block_number(),
```
mikiquantum, line 41: This means that the `on_initialize` will run in the same block when the code is enacted. Can we add a bit of space between there? Like static `100` blocks

pallets/rewards/src/lib.rs
lemunozm: It's a type that is expected to behave as a Balance but can be negative in some cases. It's used to hold the `reward_tally`.

pallets/rewards/src/lib.rs
lemunozm: This is the epoch that will end when the code is enacted. In the first `on_initialize` a new `EpochDetails` will be created using the correct `NextTotalReward`.
It can be understood as at the beginning of time, just a new epoch starts, and this code forces this to occur.

mikiquantum: I understand that, I am asking that we add a delay  of `n` blocks so it doesnt occur at the same time as the runtime upgrade first block.

lemunozm:
@mikiquantum In the last commit, I split the pallet into two levels: the **pallet's high-level logic** and the **types functionality** (`types.rs`). 

The intention is:
- Leave the pallet in the simplest way to the incoming things: deferent groups, currencies, etc...
- Divide testing: 
    - Tests inside of the `types.rs` module test the inner state of the operations in a more isolated/UT way.
    - Tests from the `test.rs` module are now simpler, focusing on performing pallet actions and balance changes (without looking at internal states). This ease the adoption of incoming groups/currencies.

libs/traits/src/rewards.rs
```diff
@@ -0,0 +1,180 @@
+// Copyright 2022 Centrifuge Foundation (centrifuge.io).
```
mikiquantum, line 1: I guess this file is already in.

pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		/// Identifier of this pallet used as an acount where stores the reward that is not claimed.
```
mikiquantum, line 84:
```suggestion
		/// Identifier of this pallet used as an account where stores the reward that is not claimed.
```

pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+
+			StakeAccounts::<T>::try_mutate(account_id, currency_id, |staked| {
+				staked.try_apply_rpt_tallies(currency.rpt_tallies())?;
+				let reward = staked.compute_reward(group.reward_per_token())?;
+
+				Ok(reward)
+			})
+		}
+
+		fn claim_reward(
```
mikiquantum, line 324: Would be great if we could add another function that claims for all the currencies that the account staked on. Not a hard requirement, but much better UX since it will only require one call instead of n depending on number of currencies staked. 
We can as well add it later.

pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+		CurrencyAttached {
+			currency_id: T::CurrencyId,
+			from: Option<T::GroupId>,
+			to: T::GroupId,
+		},
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		// Emits when a currency is used but it has no a related group.
```
mikiquantum, line 198:
```suggestion
		// Emits when a currency is used but it does not have a group associated to.
```

mikiquantum (commented):
LGTM! Thanks @lemunozm , the logic seems sound for the base implementation and the support for groups, currencies and movements.

libs/traits/src/rewards.rs
lemunozm: Yes, this branch was over the already merged branch. I need to rebase, thanks for the reminder!

pallets/rewards/src/lib.rs
lemunozm: Seems a good idea! Although we will need a new storage `Map<AccountId, vec<CurrencyId>>` to know which currencies are needed to check for a specific account.
Not sure if this logic should be in `pallet-rewards` or `pallet-liquidity-rewards`. I think the last one.

libs/traits/src/rewards.rs
lemunozm: Done

pallets/rewards/src/lib.rs
wischli: I have been following the development of this pallet and wanted to share my two cents:

IMO, such functionality should rather be handled in the Frontend than be part of the runtime. You can expose the list of staked (and claimable) currencies via a runtime API call and then claim these in a batch call. However, if it should be easily usable in Polkadot Js, an extrinsic would be helpful.

For the mapping of `AccountId` to staked `CurrencyId`s you can make use of [`iter_key_prefix(account_id)`](https://paritytech.github.io/substrate/master/frame_support/storage/trait.IterableStorageDoubleMap.html#tymethod.iter_key_prefix). IMO, it would be helpful to have a runtime API call exposing all staked currencies and another one filtering only for those for which you can claim.

lemunozm: Thanks for the comments @wischli 

> via a runtime API call and then claim these in a batch call

Good idea 🚀 

> you can make use of [StakeAccounts::<T>::iter_key_prefix(account_id)](https://paritytech.github.io/substrate/master/frame_support/storage/trait.IterableStorageDoubleMap.html#tymethod.iter_key_prefix)

I like it too. My only concern is that this pallet can be used for any other number of pallets. Some currencies can be used in one pallet, and others currencies in another pallet. Maybe the user is only interested in claiming the currencies of one of those pallets. Anyway, it seems like a very clean solution that at least we should add to the [`Rewards`](https://github.com/centrifuge/centrifuge-chain/pull/1036) API.

mustermeiszer: I.e. I would strongly lobby for single claim logic per currency and let the front combine as wished via batch calls.

pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+				)?;
+
+				Self::deposit_event(Event::GroupRewarded {
+					group_id,
+					amount: reward,
+				});
+
+				Ok(())
+			})
+		}
```
mustermeiszer, line 232: @lemunozm instead of default minting here, we could also refer to some external logic. 

```rust
                 type Balance = T::Balance;
		type GroupId = T::GroupId;
                 type Allocation = T::RewardAllocation;

		fn reward_group(group_id: Self::GroupId, reward: Self::Balance) -> DispatchResult {
			Groups::<T>::try_mutate(group_id, |group| {
				group.distribute_reward(reward)?;

				Allocation::allocate_for(
					group_id,
					&T::PalletId::get().into_account_truncating(),
					reward,
				)?;

				Self::deposit_event(Event::GroupRewarded {
					group_id,
					amount: reward,
				});

				Ok(())
			})
		}
```

where `T::RewardAllocation<T::RewardCurrency>` us bounded by a trait of `trait Allocation<Currency: Get<Self::CurrencyId>>`. We can then let the runtime decide if we wanna mint for some groups or use the pot for some others.

pallets/rewards/src/types.rs
```diff
@@ -0,0 +1,296 @@
...
+	EnsureAdd, EnsureAddAssign, EnsureFixedPointNumber, EnsureInto, EnsureSub, EnsureSubAssign,
+};
+use frame_support::pallet_prelude::*;
+use sp_runtime::{traits::Zero, ArithmeticError, FixedPointNumber, FixedPointOperand};
+
+/// Type that contains the stake properties of stake class
+#[derive(Encode, Decode, TypeInfo, MaxEncodedLen, RuntimeDebug)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct CurrencyInfo<Balance, Rate, GroupId, MaxMovements: Get<u32>> {
+	pub group_id: Option<GroupId>,
```
mustermeiszer, line 11: Why `Option<GroupId>`?

pallets/rewards/src/lib.rs
lemunozm: > we can provide a separate runtime-api and rpc-call for this purpose too

It has a lot of sense 👍🏻

[`StakeAccounts::::iter_key_prefix(account_id)`](https://paritytech.github.io/substrate/master/frame_support/storage/trait.IterableStorageDoubleMap.html#tymethod.iter_key_prefix) maybe has sense for the first iteration of CFG Rewards. With more time, we can make updates to make it by rpc calls once we use several pallets or our currencies start to grow in number.

pallets/rewards/src/types.rs
lemunozm: You need to configure which GroupId will be used for a currency. A non-attached Currency initially has a `None` group.

pallets/rewards/src/lib.rs
mikiquantum: Thanks @wischli for the comments. 
To add to what @mustermeiszer & @lemunozm mentioned. As a user I would like to:
- Get a list of currencies that I have staked in. (Runtime API)
- Have a way to see my **total** claimable rewards across all currencies I invested in. (Runtime API)
- Be able to claim rewards for one or many currencies. (Combination of Runtime API + extrinsic/batch extrinsic)

The first two I think are straight forward using [StakeAccounts::::iter_key_prefix(account_id)](https://paritytech.github.io/substrate/master/frame_support/storage/trait.IterableStorageDoubleMap.html#tymethod.iter_key_prefix) in the API level. 
For the third one, @mustermeiszer if we define a function that takes a Vec<CurrencyId, Max(20?)> then we can achieve both for most of the users. And if some users need to repeat the call because they staked in more than 20? currencies then use the Utility.Batch() call. The user/api would have to first call the `Runtime API, to get the total list of currencies` and then decide how to batch it, but at least for most of the users there will be no need to do a Batch call. 
Of course the benchmark here is important as always.

pallets/rewards/src/lib.rs
lemunozm: I've talked to @Miguel about this. Seems like by now, we will use the `PalletId` for minting inside the pallet. But I will keep this idea in mind to implement later if needed. Thanks @mustermeiszer for the hint!

lemunozm:
Changes from your last review @Miguel:
- Distribution tests moved from the pallet to the trait where the distribution is implemented.
- Minor simplifications in trait bounds for distribution methods.

pallets/rewards/src/lib.rs
mikiquantum: @mustermeiszer sounds good, lets go for that.

mikiquantum (approved):
LGTM, should we create a different PR for the following requirements?
- Get a list of currencies that I have staked in. (Runtime API)
- Have a way to see my total claimable rewards across all currencies I invested in. (Runtime API)
Or add them here? I'm ok either way.

PR #946, opened 2022-09-05
Polkadot v0.9.29

gruberb:
Upgrading to `polkadot-v0.9.28`


runtime/integration-tests/src/pools/utils/env.rs
```diff
@@ -338,10 +339,17 @@ pub struct TestEnv {
 	pub events: Arc<Mutex<EventsStorage>>,
 }
 
+pub type Block = sp_runtime::generic::Block<Header, UncheckedExtrinsic>;
+pub type UncheckedExtrinsic =
+	sp_runtime::generic::UncheckedExtrinsic<u32, <Runtime as frame_system::Config>::Call, u64, ()>;
+
```
mustermeiszer, line 345: Please import our Block type from `runtime_common`. Please use the `centrifuge::UncheckedExtrinsic` instead of generic one.

runtime/integration-tests/Cargo.toml
```diff
@@ -9,69 +9,69 @@ repository = "https://github.com/centrifuge/centrifuge-chain"
 
 
 [dependencies]
-fudge = { git = "https://github.com/centrifuge/fudge", branch = "polkadot-v0.9.26"}
+fudge = { git = "https://github.com/centrifuge/fudge", branch = "polkadot-v0.9.28"}
```
mustermeiszer, line 12: One problem we have is, that fudge depends on centrifuge. Currently the branch is pointing to this branch here, namely `centrifuge-runtime = {git = "https://github.com/gruberb/centrifuge-chain", branch = "polkadot-v0.9.28"}
`. 

So if you delete your branch we break our build, although only the tests as we have only dev-dependencies in fudge on centrifuge. I would suggest we merge this as is, then update the fudge branch to the rev of centriduge and crate a pr for centrifuge with the updated fudge version. 

WDYT?

mustermeiszer (changes requested):
Thanks a lot for pushing this fast. There are two things. One at the integration test imports that must be changed and one in the service where I think we need to change it.

docker-compose-local-relay.yml
```diff
@@ -4,7 +4,7 @@ version: '3'
 services:
   node_alice:
     container_name: alice
-    image: "parity/polkadot:v0.9.26"
+    image: "parity/polkadot:v0.9.29"
```
mikiquantum, line 7:
```suggestion
    image: "parity/polkadot:v0.9.29"
```

pallets/investments/src/lib.rs
```diff
@@ -23,7 +23,8 @@ use frame_support::{
 	traits::tokens::fungibles::{Inspect, Mutate, Transfer},
 };
 use frame_system::pallet_prelude::*;
-use sp_runtime::traits::{AccountIdConversion, CheckedSub};
+use sp_runtime::traits::AccountIdConversion;
```
mikiquantum, line 26: Just curious, why we cannot bundle them together?

runtime/altair/src/weights/pallet_uniques.rs
```diff
@@ -162,4 +162,19 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 			.saturating_add(T::DbWeight::get().reads(2 as Weight))
 			.saturating_add(T::DbWeight::get().writes(1 as Weight))
 	}
+	fn buy_item() -> Weight {
```
gruberb, line 165: I need help here @branan @mustermeiszer to figure out what the weights are or how I find/calculate... them!

pallets/nft-sales/src/tests.rs
```diff
@@ -391,8 +391,7 @@ fn buy_nft_respects_max_offer_currency() {
 fn prepared_nft(owner: &Origin) -> (u64, ItemId) {
 	let (collection_id, item_id) = (0, ItemId(1));
 
-	// Mint the nft in the uniques pallet
-	assert_ok!(Uniques::create(owner.clone(), collection_id, SELLER));
+	assert_ok!(Uniques::create(owner.clone(), SELLER));
```
mikiquantum, line 394: What is the reason for this test change to not pass a collection_id anymore?

runtime/altair/src/lib.rs
```diff
@@ -793,6 +796,7 @@ parameter_types! {
 	pub const PotId: PalletId = common_types::ids::STAKE_POT_PALLET_ID;
 	pub const MaxCandidates: u32 = 1000;
 	pub const MinCandidates: u32 = 5;
+	pub const MaxVoters: u32 = 10 * 1000;
```
mikiquantum, line 799: Seems that we should understand what this value means in case it affect block production indirectly. Is the value coming a live parity chain?

mikiquantum (commented):
I hope it was not too tedious! :)

pallets/investments/src/lib.rs
gruberb: Tested something and forgot to revert, my bad!

pallets/nft-sales/src/tests.rs
gruberb: Parity changed the function signature:
https://github.com/paritytech/substrate/blob/polkadot-v0.9.28/frame/uniques/src/lib.rs#L466

runtime/altair/src/lib.rs
gruberb: Thanks for the comment, I meant to ask the same question about a sane value here. I checked in Parities setup and this was the value I found: https://github.com/paritytech/substrate/blob/e24e9ad5855daef00f64a652ef70ac3848850d56/bin/node/runtime/src/lib.rs#L969

runtime/integration-tests/src/pools/pool_starts.rs
```diff
@@ -27,58 +27,64 @@ use runtime_common::{AccountId, Address, Balance, ItemId};
...
-		EventRange::All,
-		Event::System(frame_system::Event::ExtrinsicFailed{..}) if [count 0],
-		Event::Pools(pallet_pools::Event::Created { pool_id, .. }) if [pool_id == 0],
-		Event::Loans(pallet_loans::Event::PoolInitialised{pool_id}) if [pool_id == 0],
-		Event::Loans(pallet_loans::Event::Created{pool_id, loan_id, collateral})
-			if [pool_id == 0 && loan_id == ItemId(1) && collateral == Asset(4294967296, ItemId(1))],
-		Event::Loans(pallet_loans::Event::Priced{pool_id, loan_id, ..}) if [pool_id == 0 && loan_id == ItemId(1)],
-	);
-}
+// #[tokio::test]
```
mustermeiszer, line 36: Can you add a doc comment why we remove this for now? ^^

runtime/altair/src/lib.rs
mikiquantum: Sounds good, usually is more accurate to check on the runtime implementation (actual prod usage) https://github.com/paritytech/polkadot/blob/850d1f5a819de5a9bf52a149ffd8b7d584ba1076/runtime/polkadot/src/lib.rs#L753
And then looking at the meaning of the property: https://github.com/paritytech/substrate/blob/e24e9ad5855daef00f64a652ef70ac3848850d56/frame/elections-phragmen/src/lib.rs#L261-L266
It seems that is limiting the amount of votes taken when doing the onchain election, so if anything is protecting against  higher computation time. 👍

mikiquantum:
@gruberb please regenerate the res/rococo-local.json version to match the v0.9.28 version of polkadot.
You can run: 
```
docker run -it parity/polkadot:v0.9.28 build-spec --chain=rococo-local --disable-default-bootnode --raw > rococo-local.json
```
Make sure the file doesnt contain some non-json code :)

PR #947, opened 2022-09-05
Add TrancheMetadata

gruberb:
Solves https://github.com/centrifuge/centrifuge-chain/issues/842


mustermeiszer:
@gruberb ready for review?

libs/common-types/src/impls.rs
```diff
@@ -0,0 +1,49 @@
...
+// GNU General Public License for more details.
+use super::*;
+use frame_support::traits::{EnsureOrigin, EnsureOriginWithArg};
+use frame_system::RawOrigin;
+use sp_std::marker::PhantomData;
+
+type AccountId = u64;
+
+/// The OrmlAssetRegistry::AuthorityOrigin impl
+pub struct AuthorityOrigin<
```
mustermeiszer, line 20: Can we name this somehow that it describes its purpose?

pallets/pools/src/benchmarking.rs
```diff
@@ -11,7 +11,7 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU General Public License for more details.
 
-//! Module provides benchmarking for Loan Pallet
+//! Module provides benchmarking for the Pools Pallet
```
mustermeiszer, line 14: ❤️

pallets/pools/src/lib.rs
```diff
@@ -391,8 +426,12 @@ pub mod pallet {
 
 	#[pallet::storage]
 	#[pallet::getter(fn scheduled_update)]
-	pub type ScheduledUpdate<T: Config> =
-		StorageMap<_, Blake2_128Concat, T::PoolId, ScheduledUpdateDetails<T::InterestRate>>;
+	pub type ScheduledUpdate<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat,
+		T::PoolId,
+		ScheduledUpdateDetails<T::InterestRate, T::MaxTokenNameLength, T::MaxTokenSymbolLength>,
```
mustermeiszer, line 433: This will actually be a migration. There is no migration available, right? I am pretty sure, we do not have any scheduled updates in place. 

@branan and @offerijns , we could go without writing a migration here. I am a little hesitant, but think it would be okayish.

pallets/pools/src/tranche.rs
```diff
@@ -225,6 +260,41 @@ where
...
+		token_name: Vec<u8>,
+		token_symbol: Vec<u8>,
+	) -> AssetMetadata<Balance, CustomMetadata>
+	where
+		Balance: Zero,
+		Currency: Encode,
+		CustomMetadata: Parameter + Member + TypeInfo,
+	{
+		let tranche_id =
+			WeakBoundedVec::<u8, ConstU32<32>>::force_from(self.currency.encode(), None);
```
mustermeiszer, line 277: Not sure if we should go with it this way. We will never notice if this fails. At least we need a test to cover a failure of this, where we do a `from` instead of `force_from`.

also: 
* why is it bounded by length of `32`?

We could enforce the `CurrencyId` to implement some trait we define. Like `trait GenerelKey`. And then we define this manually for the `enum CurrencyId` we provide. 

@NunoAlexandre could this be useful for xcm in general?

pallets/pools/src/tranche.rs
```diff
@@ -225,6 +260,41 @@ where
...
+			name: token_name,
+			symbol: token_symbol,
+			existential_deposit: Zero::zero(),
+			location: Some(VersionedMultiLocation::V1(MultiLocation {
+				parents: 1,
+				interior: X2(Parachain(parachain_id.into()), GeneralKey(tranche_id)),
+			})),
+			additional: CustomMetadata {
+				mintable: false,
+				permissioned: false,
```
mustermeiszer, line 290: @offerijns aren't tranche currencies `permissioned = true`?

runtime/integration-tests/src/pools/env.rs
```diff
@@ -30,15 +30,15 @@ async fn env_works() {
 		.with_state(Chain::Para(PARA_ID), || {
 			frame_system::Pallet::<Runtime>::block_number()
 		})
-		.unwrap();
+		.expect("Cannot create block before");
```
mustermeiszer, line 33: ❤️

runtime/integration-tests/src/pools/env.rs
```diff
@@ -30,15 +30,15 @@ async fn env_works() {
...
 
 	env::pass_n(&mut env, num_blocks).unwrap();
 
 	let block_after = env
 		.with_state(Chain::Para(PARA_ID), || {
 			frame_system::Pallet::<Runtime>::block_number()
 		})
-		.unwrap();
+		.expect("Cannot create block after");
```
mustermeiszer, line 41: 💯

runtime/integration-tests/src/pools/pool_starts.rs
```diff
@@ -42,6 +42,22 @@ async fn create_init_and_price() {
...
+				symbol: "MOCK".as_bytes().to_vec(),
+				existential_deposit: 0,
+				location: None,
+				additional: common_types::CustomMetadata::default(),
+			},
+			Some(common_types::CurrencyId::AUSD),
+		)
+		.unwrap();
+	});
+
```
mustermeiszer, line 60: Didn't you go with the `GenesisConfig`? ^^

I will add this at a later point.

runtime/integration-tests/src/pools/pool_starts.rs
```diff
@@ -64,7 +80,7 @@ async fn create_init_and_price() {
 			Chain::Para(PARA_ID),
 			Event,
 			EventRange::All,
-			//Event::Pools(..) | Event::Loans(..) | Event::Uniques(..)
+			// Event::Pools(..) | Event::Loans(..) | Event::Uniques(..)
```
mustermeiszer, line 83: Remove whole `tracing::info` span

mustermeiszer (commented):
🎉  This looks really, really, good. 

The main reason I am not approving is that
* I am not sure if `permissioned` flag is set correctly in metadat
* I don't like the `force_from` for the generation of the `GeneralKey`. I would like to have the generation of the `GeneralKey` to be more explicit on our side.

runtime/integration-tests/src/pools/pool_starts.rs
gruberb: This is what holds the PR back. Currently my IDE tells me that `MockGenesisConfigAssetRegistry` doesn't implement `MaybeSerializeDeserialize` and I am in the process of figuring out what this means and how to add it :)

mustermeiszer: Think you can simply derive Encode, Decode. Just a blow in the wind though.

pallets/pools/src/lib.rs
offerijns: I think it's fine to not have a migration for this.

pallets/pools/src/tranche.rs
offerijns: We also started discussing this here: https://github.com/centrifuge/centrifuge-chain/pull/901#discussion_r954636495

I think @NunoAlexandre makes a good point: my initial reason for saying this should be `false`, is the idea that we will have 3 types of currencies:
1. Tranche tokens
2. Permissioned currencies
3. Non-permissioned currencies

But indeed, we can also say (arguably more logically) there are
1. Permissioned currencies
   - Tranche tokens
   - Non-tranche tokens
2. Non-permissioned currencies

And then in the token `PreConditions`, have a tranche token specific check or a non-tranche token permission check.

libs/common-types/src/impls.rs
gruberb: To my understanding, we can't. Because we overwrite the standard implementation. I will add additional code comments to make the purpose more clear!

pallets/pools/src/tranche.rs
NunoAlexandre: I need to go through this in more detail as I am just scanning it right now, but for xcm purposes all we need is to have a way of identifying a tranche token uniquely; if it's with `GeneralKey` or `GeneralIndex` or another `Junction`, it doesn't matter much; it just has to be unique and preferrably deterministically derived from the pool and/or tranche_id.

gruberb: We use `WeakBoundedVec<>::force_from` also in `runtime/common/impls.rs`, this is why I thought it is ok to use it here as well. Is this still a concern or can we merge this PR? @mustermeiszer @NunoAlexandre

libs/common-types/src/impls.rs
mustermeiszer: But aren't we implementing on our own struct? As long as we satisfy the bounds, I guess we can name it as we like.

gruberb: Yeah, we can. It's just that I want to be sure we stay as close to the real code as possible. We use the exact same naming, so if we want to reason about the logic, we don't run into a completely different name and have to wonder: Why did we use this in the first place?

I am open for a change, but for me, this doesn't make the logic nor the reasoning about it any simpler.

pallets/pools/src/tranche.rs
mustermeiszer: At least from what I see, we use this force for only [u8;2] arrays. But the encoding is arbitrarily large and we will not notice the changes.

mustermeiszer: But I will bring in a proposal for letting the currency define its general key itself, then we can use this method. At least for now, I would not block this anymore.

mustermeiszer (approved):
That was a big chunk of work, nice! Thanks @gruberb, let's get this in.

PR #961, opened 2022-09-07
add fix to use collator-selection correctly

lemunozm:


runtime/centrifuge/src/lib.rs
```diff
@@ -89,6 +93,78 @@ pub fn native_version() -> NativeVersion {
...
+			let mut current_validators = <pallet_session::Validators<T>>::get()
+				.into_iter()
+				.map(|who| who.into())
+				.collect::<Vec<_>>();
+
+			current_validators.truncate(T::MaxInvulnerables::get() as usize);
+
+			// SAFETY: The vector is truncate before to ensure the vector fit into the
+			// expected BoundedVec from pallet_collator_selection.
+			let current_validators = current_validators.try_into().unwrap();
```
lemunozm, line 157: I needed to add this section in addition to how it was on [Altair](https://github.com/centrifuge/centrifuge-chain/blob/v3.0.0-alpha7-2/runtime/altair/src/lib.rs#L94-L154) because now it is working with `Boundedvec`. Also the limit of immutable validators is 100. I hope it is enough

mikiquantum: we only have 3, so no problem since it will not be executed later in time.

PR #962, opened 2022-09-08
Adding templates for PRs and Issues

gruberb:
A first suggestion for using templates when opening new PRs or Issues. This is an attempt to remind developers to think about edge cases, thinking problems through, gathering feedback etc.

This shouldn't prevent PRs from being opened (quality gate), it is just so newcomers have an easier time contributing by having some guidelines, and to unify PR descriptions and issue descriptions.

Very happy to get feedback for adding/removing/rewording sections. That's just a first attempt!

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue_template.md
+++ b/.github/ISSUE_TEMPLATE/issue_template.md
@@ -1,14 +1,8 @@
 ### Which part of the code is the issue addressing?
 
-* [ ] Archicture 
-* [ ] Adressing a bug
+* [ ] Addressing a bug
 * [ ] Refactoring
-* [ ] pallet_xyz
-* [ ] pallet_xyz
-* [ ] Suggesting a new pallet
-* [ ] Runtime
-* [ ] Cross-Chain
-* [ ] Protocol
+* [ ] Feature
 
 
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/issue_template.md
```diff
@@ -0,0 +1,29 @@
...
+
+* [ ] Archicture 
+* [ ] Adressing a bug
+* [ ] Refactoring
+* [ ] pallet_xyz
+* [ ] pallet_xyz
+* [ ] Suggesting a new pallet
+* [ ] Runtime
+* [ ] Cross-Chain
+* [ ] Protocol
```
mustermeiszer, line 11: I would keep the list shorter. Remove the circles, as they are not constant and probably unknown to person not familiar with the company. We can manually add the labels for the circles. WDYT?

I would keep 
* Addressing a bug
* Refactoring
* Feature

mustermeiszer (commented):
I like this. What do the others think? Would make sense to have a more formalized way.

mustermeiszer (approved):
Let's give this a try and then we can iterate on the templates depending on how we use them

PR #969, opened 2022-09-12
limit iterations on evict_anchors

lemunozm:


runtime/altair/src/lib.rs
```diff
@@ -1338,6 +1338,7 @@ mod upgrade {
 			weight += InterestAccrual::upgrade_to_v1();
 			weight += Loans::reference_active_rates();
 			weight += InterestAccrual::remove_unused_rates();
+			weight += pallet_anchors::migration::fix_evict_date::migrate::<Runtime>();
```
lemunozm, line 1341: Hi @branan I've seen that you are made some specific runtime migrations in this way. I add mine too. Could you confirm if it's ok to add this line here? what I want is to ensure that the migration is done only for Altair.

mikiquantum: @lemunozm if we add the migration as hook in the pallet I think then you do not need to add it here. It will be executed automatically

mikiquantum: In fact, since you are not using the hooks, in the pallet then yes this is good here, but since the `OnRuntimeUpgrade` trait doesnt implement the pre and post then that part of the code will not be executed when running with `try-runtime`

mikiquantum (approved):
LGTM!

mikiquantum (approved):
LGTM

Issue #956, opened 2022-09-06
Create `pools-registry` pallet


gruberb:
We want to move `Metadata` related code out of the `pools-pallet` into its own pallet. 

- [x] Move https://github.com/centrifuge/centrifuge-chain/issues/842 into `pools-registry`
- [x] Moving `metadata` out of the `Pool` struct and create a new Storage item within `pools-registry` with `PoolsMetadata<PoolId, Metadata>`
- [x] Move `set_metadata` (https://github.com/centrifuge/centrifuge-chain/blob/parachain/pallets/pools/src/lib.rs#L796-L830) into `pools-registry`

@mustermeiszer and @offerijns  had some more suggestions what `pools-registry` should do.

offerijns:
One method we could add to `pools-registry` is `set_pool_name`, for https://github.com/centrifuge/centrifuge-chain/issues/701

mustermeiszer:
Not sure if we need this already in this version. But we should have a `UpdateGuard` at some point in the near future.

gruberb:
Going through the changes, I collected some "major" API changes:

- When we `create` a `Pool`, we send metadata. If we want to move all `metadata` related functionality into `pools-registry`, we have to always have two calls from now on and adjust the `create` paramaters.
- Same goes for `update` `Pool` (which has `TrancheMetadata` included in the `update` call)

offerijns:
> When we create a Pool, we send metadata. If we want to move all metadata related functionality into pools-registry, we have to always have two calls from now on and adjust the create paramaters.

For now, I think the metadata should still be passed to `pools.create`, but `pools-registry` should expose a trait method to update the method and this can then be called by `pools.create`. I think at some point, we can potentially consider entirely moving `create`/`update` to the registry, but this is definitely a much bigger change we should not start on now.

gruberb:
> but pools-registry should expose a trait method to update the method and this can then be called by pools.create

Hm, what would be the advantage? If I introduce a `trait`, which is just used for one other pallet, than we can implement it in the pallet right away! 

We currently use the `pools` storage for the metadata, so how I read this is:

Use the `pools-registry` storage for metadata, but still call the functionality to update this metadata inside the `create` and `update` extrinsic.

mustermeiszer:
> Hm, what would be the advantage? If I introduce a trait, which is just used for one other pallet, than we can implement it in the pallet right away!

what do you mean here? 

[...]

PR #974, opened 2022-09-14
Pallet pools registry

gruberb:
Closes https://github.com/centrifuge/centrifuge-chain/issues/956


NunoAlexandre:
@gruberb do we have a timeline to get this merged?

gruberb:
> @gruberb do we have a timeline to get this merged?

This will be a bigger change. In the first version, we are aiming to move out the `create`, `update` and `set_metadata` extrinsic of `pools` and move them to `pools-registry`. It's a bigger refactor, with possible migrations involved, so I hope in 2 weeks? I am still not up-to-speed on everything so takes a bit of time.

Do you have a pressing issue which requires this change to be done sooner?

mustermeiszer (changes requested):
Just a first look at formal stuff. Cargo and license. In general everywhere license on top ^^

mustermeiszer:
Are we planning on bringing this in in this state and iterate over it?

gruberb:
> Are we planning on bringing this in in this state and iterate over it?

It depends... the full re-write on bringing `pallet-pools` to state where we have `pools-registry` and `pools-system` takes some time. I ran against a few walls, and once the code is done, we have to re-write almost all tests, re-run benchmarks etc. My gut feeling tells me this takes longer than the 2 weeks we are eying to get these changes in.

So one idea from @offerijns was to have a first step, have the events and the first extrinsics in `pools-registry`, and then in the next iteration, do the slightly bigger change.

libs/types/src/pools.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use cfg_primitives::Moment;
+use codec::{Decode, Encode};
+use frame_support::{traits::Get, BoundedVec, RuntimeDebug};
+use orml_traits::Change;
+use scale_info::TypeInfo;
+
+use super::*;
+
+// #[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, TypeInfo)]
```
thea-leake, line 22: Is there a reason why we have this block here commented out?

libs/types/src/tranches.rs
```diff
@@ -0,0 +1,88 @@
...
+	/// Compares tranches with the following schema:
+	///
+	/// * (Residual, Residual) => true
+	/// * (Residual, NonResidual) => true,
+	/// * (NonResidual, Residual) => false,
+	/// * (NonResidual, NonResidual) =>
+	///         interest rate of next tranche must be smaller
+	///         equal to the interest rate of self.
+	///
+	pub fn valid_next_tranche(&self, next: &TrancheType<Rate>) -> bool {
```
thea-leake, line 41: Would this be something we would want to add tests for with the current tranche tests work?

pallets/pools/src/benchmarking.rs
```diff
@@ -49,76 +49,76 @@ benchmarks! {
...
-	})
-	verify {
-		// No redemption order was submitted and the MinUpdateDelay is 0 for benchmarks,
-		// so the update should have been executed immediately.
-		let pool = get_pool::<T>();
-		assert_update_tranches_match::<T>(pool.tranches.residual_top_slice(), &tranches);
-		assert_eq!(pool.parameters.min_epoch_time, SECS_PER_DAY);
-		assert_eq!(pool.parameters.max_nav_age, SECS_PER_HOUR);
-	}
+	// create {
```
thea-leake, line 52: I think it would be nice to have a comment explaining why this section is here but commented out, but I'm fine with it as is.

thea-leake (commented):
LGTM, I have a few questions, but nothing blocking :)

libs/types/src/pools.rs
gruberb: Thanks, I forgot to point that out! I wanted feedback on the idea @mustermeiszer had on converting the `PoolChanges` from a struct to an enum. If we do this, where in the frontend/App/wherever do we have to adjust or tell people?

NunoAlexandre:
@gruberb I would like to review this before it gets merged. Can I do that today or Monday?

pallets/pools-registry/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn get_pool_metadata)]
+	pub(super) type PoolMetadata<T: Config> =
+		StorageMap<_, Blake2_256, T::PoolId, PoolMetadataOf<T>>;
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// A pool creation was registered
+		/// and forwarded to the pool-systems pallet
```
offerijns, line 193:
```suggestion
		/// and forwarded to the pools-system pallet
```

pallets/pools-registry/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Invalid metadata passed
+		BadMetadata,
+		/// Pre-requirements for a TrancheUpdate are not met
+		/// for example: Tranche changed but not its metadata or vice versa
+		InvalidTrancheUpdate,
+		/// No metada for the given currency found
```
offerijns, line 213:
```suggestion
		/// No metadata for the given currency found
```

offerijns (commented):
First pass mostly with a bunch of small naming/comments changes

pallets/pools-registry/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+	#[pallet::getter(fn get_pool_metadata)]
+	pub(super) type PoolMetadata<T: Config> =
+		StorageMap<_, Blake2_256, T::PoolId, PoolMetadataOf<T>>;
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// A pool creation was registered
+		/// and forwarded to the pool-systems pallet
+		Registered { pool_id: T::PoolId },
```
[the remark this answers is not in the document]
gruberb: I think we want to already look-ahead with the events. Once a create is hitting `pools-registry`, it is not guaranteed yet that the creation will go through. We emit a `created` event inside `pool-systems`. 

I guess it's an in-between solution. I can see that we have to update Subgraph soon again because of all the new events we emit in both pallets. Therefore I am also open to change them to "as-is" and update them along the way.

Maybe @mustermeiszer can chime in as well?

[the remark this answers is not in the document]
offerijns: For events, we should try not to have in-between solutions. If we call this `Registered` now but change it to `Created` later once the code is fully ported, the subquery will forever have two event handlers. I think we should consider what the final event name would be, and just use that already now.

Cargo.toml
```diff
@@ -45,6 +45,8 @@ members = [
 	"pallets/migration",
 	"pallets/nft",
 	"pallets/nft-sales",
+	"pallets/pools",
+    "pallets/pool-registry",
```
mustermeiszer, line 49: fmt

flake.nix
```diff
@@ -93,7 +93,7 @@
             };
 
             # This is a hash of all the Cargo dependencies, for reproducibility.
-            cargoSha256 = "YysY+c0tr1+bGwkI7ncLMP1InefimXYAXAZjsxH5o3g=";
+            cargoSha256 = "Verp/cF0lbQQIpB54BCro0XWI5nXka56mo25TuKweJM=";
```
mustermeiszer, line 96: No more sha256 prefix?

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+		max_reserve: Balance,
+		metadata: Option<Vec<u8>>,
+	) -> DispatchResult;
+
+	fn update(
+		pool_id: PoolId,
+		changes: PoolChanges<Rate, MaxTokenNameLength, MaxTokenSymbolLength, MaxTranches>,
+	) -> DispatchResultWithPostInfo;
+
+	fn execute_update(pool_id: PoolId) -> DispatchResultWithPostInfo;
```
mustermeiszer, line 81: PoolSystem will implement this. Move to libs/traits.

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+				currency,
+				max_reserve,
+				metadata,
+			) {
+				Ok(_) => {
+					Self::deposit_event(Event::Created { pool_id, admin });
+					Ok(())
+				}
+				Err(e) => Err(e),
+			}
```
mustermeiszer, line 276: `map(|_| Self::deposit_event(Event::Created{pool_id, admin}))` instead of match. Or was this not possible?

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+			);
+
+			match T::ModifyPool::update(pool_id, changes) {
+				Ok(res) => {
+					Self::deposit_event(Event::Updated { pool_id });
+					Ok(res)
+				}
+				Err(e) => Err(e),
+			}
+		}
```
mustermeiszer, line 330: This is not correct. The update methods will also return success if the update is just stored, right? 

So we should have the trait `ModifyPool` to returning  an enum of something like 

```rust
pub enum UpdateState {
   Executed,
   Stored
}
```

Then we would trigger `Event::UpdateRegisterd {pool_id}` always, when we have an `Ok(UpdateState)` return value, and also trigger `Event::UpdateExecuted{pool_id}` if the state signals execution.

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// A pool was created.
+		Created {
+			pool_id: T::PoolId,
+			admin: T::AccountId,
+		},
+		/// A pool was updated.
+		Updated { pool_id: T::PoolId },
+		/// Pool metadata was set.
```
mustermeiszer, line 196: Split into: 

`UpdateRegisted` and `UpdateExecuted`

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+		) -> DispatchResultWithPostInfo {
+			ensure_signed(origin)?;
+
+			match T::ModifyPool::execute_update(pool_id) {
+				Ok(res) => {
+					Self::deposit_event(Event::Updated { pool_id });
+					Ok(res)
+				}
+				Err(e) => Err(e),
+			}
```
mustermeiszer, line 351: also just `map`

mustermeiszer (changes requested):
Left some comments where changes are needed. I would also not bring in the weight changes with this PR. We should have a separate PR for updating the weights of the runtimes. WDYT?

pallets/pool-registry/src/lib.rs
gruberb: We will get a circular dependency when moving this to `cfg-traits`, because we rely on `cfg-types` here because of `PoolChanges` and `TrancheInput`. Looking at this, it would make sense not to have any `impls` in `cfg-types` and move them to `cfg-traits`. So `cfg-types` doesn't rely on `cfg-traits`.

pallets/pool-registry/src/lib.rs
gruberb: How can we wrap or unbundle the enum inside a `DispatchResultWithPostInfo`? We need the weight of the `update` function inside `pools` as well.

pallets/pool-registry/src/lib.rs
gruberb: Hm... doesn't this swallow the `Error`?

mustermeiszer: No. See https://doc.rust-lang.org/src/core/result.rs.html#767

pallets/pool-registry/src/lib.rs
mustermeiszer: `Result<(UpdateState, PostDispatchInfo), DispatchErrorWithPostInfo<PostDispatchInfo>>` would be a solution.

libs/types/src/permissions.rs
```diff
@@ -130,10 +129,10 @@ pub struct TrancheInvestors<Now, MinDelay, TrancheId, Moment> {
...
 #[derive(Encode, Decode, TypeInfo, Clone, Eq, PartialEq, Debug)]
 pub struct PermissionRoles<Now, MinDelay, TrancheId, Moment = u64> {
-	pool_admin: PoolAdminRoles,
-	currency_admin: CurrencyAdminRoles,
-	permissioned_asset_holder: PermissionedCurrencyHolders<Now, MinDelay, Moment>,
-	tranche_investor: TrancheInvestors<Now, MinDelay, TrancheId, Moment>,
+	pub pool_admin: PoolAdminRoles,
+	pub currency_admin: CurrencyAdminRoles,
+	pub permissioned_asset_holder: PermissionedCurrencyHolders<Now, MinDelay, Moment>,
+	pub tranche_investor: TrancheInvestors<Now, MinDelay, TrancheId, Moment>,
```
mustermeiszer, line 135: Why do we need them public? Not against it, just curious and wondering if we should make it an API with methods on the struct rather than making them public.

pallets/pool-registry/src/lib.rs
gruberb: Oh sweet!

libs/types/src/permissions.rs
gruberb: It's all because we have `cfg-types` and `cfg-traits`. Inside `cfg-traits`, we depend on `cfg-types`. To prevent a circular dependency, we can't rely on `cfg-traits` inside `cfg-types`. Which moves the `impl traits` out of `types` and into `traits`. We need to test the `impl functions` inside `traits`.

gruberb: I don't know if my answer makes sense. I also don't like, but for how we currently structure everything, it's hard not to make them public (with testing and everything).

gruberb:
> Left some comments where changes are needed. I would also not bring in the weight changes with this PR. We should have a separate PR for updating the weights of the runtimes. WDYT?

Yeah, I test ran the weights to see if they are working. Reverted them back to the previous values 👍

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// A pool was created.
+		Created {
+			pool_id: T::PoolId,
+			admin: T::AccountId,
+		},
+		/// A pool update was registered.
+		UpdateRegisted { pool_id: T::PoolId },
```
offerijns, line 166:
```suggestion
		UpdateRegistered { pool_id: T::PoolId },
```

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+				),
+				BadOrigin
+			);
+
+			match T::ModifyPool::update(pool_id, changes) {
+				Ok((state, dispatch_info)) => {
+					Self::deposit_event(Event::UpdateRegisted { pool_id });
+
+					match state {
+						UpdateState::NotExecution => {
```
offerijns, line 300: I'm not sure if we need to emit an event in case of a no-op. I think the core goal of events is for clients to easily figure out what changed; if nothing changed, it doesn't seem like this requires an event. Wdyt?

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+					PermissionScope::Pool(pool_id),
+					who,
+					Role::PoolRole(PoolRole::PoolAdmin)
+				),
+				BadOrigin
+			);
+
+			match T::ModifyPool::update(pool_id, changes) {
+				Ok((state, dispatch_info)) => {
+					Self::deposit_event(Event::UpdateRegisted { pool_id });
```
offerijns, line 297:
```suggestion
					Self::deposit_event(Event::UpdateRegistered { pool_id });
```

pallets/pool-registry/src/mock.rs
```diff
@@ -0,0 +1,230 @@
...
+// Configure a mock runtime to test the pallet.
+frame_support::construct_runtime!(
+	pub enum Test
+	where
+		Block = Block,
+		NodeBlock = Block,
+		UncheckedExtrinsic = UncheckedExtrinsic,
+	{
+		System: frame_system::{Pallet, Call, Config, Storage, Event<T>},
+		PoolsRegistry: pallet_pool_registry::{Pallet, Call, Storage, Event<T>},
```
offerijns, line 162:
```suggestion
		PoolRegistry: pallet_pool_registry::{Pallet, Call, Storage, Event<T>},
```

pallets/pool-registry/src/tests.rs
```diff
@@ -0,0 +1,45 @@
...
+			let pool_owner = 9u64;
+			let pool_id = 0;
+			let changes = PoolChanges {
+				tranches: Change::NoChange,
+				tranche_metadata: Change::NoChange,
+				min_epoch_time: Change::NewValue(10),
+				max_nav_age: Change::NoChange,
+			};
+
+			assert_ok!(PoolsRegistry::update(
```
offerijns, line 21:
```suggestion
			assert_ok!(PoolRegistry::update(
```

pallets/pool-registry/src/tests.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[test]
+fn set_metadata() {
+	TestExternalitiesBuilder::default()
+		.build()
+		.execute_with(|| {
+			let pool_owner = 9u64;
+			let pool_id = 0;
+
+			assert_ok!(PoolsRegistry::set_metadata(
```
offerijns, line 37:
```suggestion
			assert_ok!(PoolRegistry::set_metadata(
```

runtime/development/src/lib.rs
```diff
@@ -1565,6 +1582,7 @@ construct_runtime!(
 		Investments: pallet_investments::{Pallet, Call, Storage, Event<T>} = 105,
 		Rewards: pallet_rewards::{Pallet, Storage, Event<T>} = 106,
 		LiquidityRewards: pallet_liquidity_rewards::{Pallet, Call, Storage, Event<T>} = 107,
+		PoolsRegistry: pallet_pool_registry::{Pallet, Call, Storage, Event<T>} = 108,
```
offerijns, line 1585:
```suggestion
		PoolRegistry: pallet_pool_registry::{Pallet, Call, Storage, Event<T>} = 108,
```

runtime/integration-tests/src/pools/utils/pools.rs
```diff
@@ -350,7 +353,7 @@ pub fn create_pool_call(
 	max_reserve: Balance,
 	tranche_inputs: Vec<TrancheInput<Rate, MaxTokenNameLength, MaxTokenSymbolLength>>,
 ) -> Call {
-	Call::Pools(PoolsCall::create {
+	Call::PoolsRegistry(PoolsRegistryCall::create {
```
offerijns, line 356:
```suggestion
	Call::PoolRegistry(PoolRegistryCall::create {
```

runtime/integration-tests/src/pools/utils/pools.rs
```diff
@@ -14,12 +14,15 @@
...
-use cfg_types::{CurrencyId, PermissionScope, PoolRole, Rate, Role};
+use cfg_types::{
+	CurrencyId, PermissionScope, PoolRole, Rate, Role, TrancheInput, TrancheMetadata, TrancheType,
+};
 use codec::Encode;
 use frame_support::{Blake2_128, StorageHasher};
 use fudge::primitives::Chain;
 use pallet_permissions::Call as PermissionsCall;
-use pallet_pools::{Call as PoolsCall, TrancheIndex, TrancheInput, TrancheMetadata, TrancheType};
+use pallet_pool_registry::Call as PoolsRegistryCall;
```
offerijns, line 24:
```suggestion
use pallet_pool_registry::Call as PoolRegistryCall;
```

libs/types/src/permissions.rs
mustermeiszer: At the current commit of main. There is solely a dev dependency that is not even used. I am not sure, where the issue came from. But I think we can agree, that `traits` is stand alone and types can depend on traits and implement them.

libs/traits/src/lib.rs
```diff
@@ -321,6 +368,23 @@ pub trait TrancheCurrency<PoolId, TrancheId> {
...
+
+	fn of_pool(&self) -> PoolId {
+		self.pool_id
+	}
+
+	fn of_tranche(&self) -> TrancheId {
+		self.tranche_id
+	}
+}
+
```
mustermeiszer, line 387: This must not be here.

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+				),
+				BadOrigin
+			);
+
+			match T::ModifyPool::update(pool_id, changes) {
+				Ok((state, dispatch_info)) => {
+					Self::deposit_event(Event::UpdateRegistered { pool_id });
+
+					match state {
+						UpdateState::NoExecution => (),
```
mustermeiszer, line 300: When would this happen? And what does this signal?

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+		) -> DispatchResultWithPostInfo {
+			ensure_signed(origin)?;
+
+			match T::ModifyPool::execute_update(pool_id) {
+				Ok(res) => {
+					Self::deposit_event(Event::UpdateExecuted { pool_id });
+					Ok(res)
+				}
+				Err(e) => Err(e),
+			}
```
mustermeiszer, line 332:
```rust
T::ModifyPool::execute_update(pool_id).map(|res| {
   Self::deposit_event(Event::UpdateExecuted { pool_id });
   res
})
```

pallets/pools/src/lib.rs
```diff
@@ -1739,4 +1399,246 @@ pub mod pallet {
 			Ok(())
 		}
 	}
+
+	impl<T: Config>
+		PoolMutate<
+			T::AccountId,
+			T::Balance,
```
mustermeiszer, line 1406: There is an `impl` file for trait impls.

mustermeiszer (changes requested):
We need to rethink about this restructuring made here. And I have the feeling this is too rushed to bring this in in this state.

pallets/pool-registry/src/lib.rs
gruberb: Well if we do a `Result<(UpdateState, PostDispatchInfo), DispatchErrorWithPostInfo>` from `fn update()`, we need this. We have three different valid returns in this function, `NoExecution` is one of them.

libs/traits/src/lib.rs
gruberb: The problem started when we moved `pub trait PoolMutate<...` into `cfg-traits`. This trait depends on `TrancheInput` and `PoolChanges`. Therefore, we need to depend on `cfg-types` inside `cfg-traits`.

pallets/pool-registry/src/lib.rs
branan: We really want `inspect`, but alas it's not stable yet https://github.com/rust-lang/rust/issues/91345

gruberb:
Decided that the last changes into `main` threw quite a few things off with this PR. So I will close this, and jerry-pick just the relevant changes related to `pallet-pool-registry` and open up a new PR.

PR #978, opened 2022-09-19
Revert "Revert "Pallet bridge using new pallet fees (#933)" (#963)"

lemunozm:
This change recovers the #933 PR


pallets/bridge/src/lib.rs
```diff
@@ -14,77 +14,6 @@
 //! # Bridge pallet
 //!
 //! This pallet implements a bridge between Chainbridge and Centrifuge Chain.
-//! - [`Config`]
-//! - [`Call`]
-//! - [`Pallet`]
-//!
-//! ## Overview
```
lemunozm, line 21: I've removed this documentation to ease the maintainability because rustdoc already builds it for you using the inline documentation.

pallets/bridge/src/lib.rs
```diff
@@ -385,92 +264,12 @@ pub mod pallet {
 		pub fn remark(
```
lemunozm, line 264: Is this method really useful?

mikiquantum: This function allows to trigger a bridge inbound call that is successful. Is not very used though, but not sure if we need to worry about it.

mikiquantum (dismissed):
Looks good, lets not forget to add the correspondent fees before we execute the next upgrade. 👍

NunoAlexandre (approved):

Issue #849, opened 2022-06-28
Add tests for interest-accrual pallet


offerijns:
_Originally posted by @offerijns in https://github.com/centrifuge/centrifuge-chain/pull/789#discussion_r902834994_

PR #983, opened 2022-09-20
Validate interest rates and limit the total count

branan:
# Description

Limit the valid set of interest rates to encourage re-use across loans. Also restrict the total count of rates to ensure a hard limit in possible on_initialize weight for interest accrual calculations.

## Changes and Descriptions

This limits the total number of rates in two ways: 1) A hard limit, currently set to 300
2) Limiting the actual range of rates that is considered valid.

A valid rate is anything with up to two decimal points in the range 00.01 to 99.99, inclusive. Some examples:
* 5.00 is valid
* 100.00 is invalid (too large)
* 12.999 is invalid (too many decimals)

In order to validate this range, the `price` extrinsic must now take a yearly rate.

API changes:
* Loans::price now takes a yearly rate instead of a secondly rate
* Loans::price can now fail with two new errors
  - `InterestAccrual::InvalidRate` when a rate does not match the requirements. Note that this includes some cases which would previously return a `Loans::LoanValidInvalid` error.
  - `InterestAccrual::TooManyRates` when the hard limit has been hit

Migrations:
* A migration has been added which counts the number of in-use interest rates so that the hard limit can be tracked.
## Type of change

# How Has This Been Tested?

Tested through existing unit & integration tests and benchmarking.

# Checklist:


branan:
Part two should be a much faster follow-on to this, but this is here if for some reason that doesn't get in.

Part two will be a bigger migration to a `BoundedVec` for storing rates, and with that a big change to NAV updates and their benchmarks.

branan:
@centrifuge/apps FYI, this contains a breaking API change for the Loans pallet. See the description above and the updated doc comment at https://github.com/centrifuge/centrifuge-chain/pull/983/files#diff-6f262dd703876877438056eed95093b7b37f23a7bc90a1f14f84eee09ef9b20aR523

An identical change to how rates are handled will be made to write-off groups as well, I just haven't gotten to it.

onnovisser:
@branan In the code it looks like there's max 100 rates with a precision of 4 decimals. Does that mean that there can only be 100 interest rates between 00.00% and 100.00% across all of the pools?

pallets/interest-accrual/src/lib.rs
```diff
@@ -383,6 +421,23 @@ pub mod pallet {
...
+			RateCount::<T>::set(count as u32);
+			T::DbWeight::get().reads(count as u64) + T::DbWeight::get().writes(1)
+		}
+
+		pub(crate) fn validate_rate(interest_rate_per_year: T::InterestRate) -> DispatchResult {
+			let four_decimals = T::InterestRate::saturating_from_integer(10000);
+			ensure!(
+				interest_rate_per_year < One::one()
+					&& interest_rate_per_year > Zero::zero()
+					&& (interest_rate_per_year * four_decimals).frac() == Zero::zero(),
```
mustermeiszer, line 436: `saturating_mul` although I know the booleans will fail before reaching this case.

pallets/loans/src/lib.rs
```diff
@@ -703,7 +709,7 @@ pub mod pallet {
 				for loan in active_loans.iter() {
 					weight += T::DbWeight::get().reads_writes(1, 1);
 					let rate = Self::rate_with_penalty(&loan, &write_off_groups);
-					T::InterestAccrual::reference_rate(rate);
+					T::InterestAccrual::reference_rate(rate).unwrap();
```
mustermeiszer, line 712: This migration is not versioned. So basically, this will run everytime we run a runtime upgrade.  
We should version. 

Not sure, but it probably ran once already, right? 

Also, I am on the fence of having migration code in the normal pallet.

pallets/loans/src/functions.rs
```diff
@@ -754,7 +752,7 @@ impl<T: Config> Pallet<T> {
 					.interest_rate_per_sec
 					.checked_add(&write_off_penalty_rate)
 					.ok_or(ArithmeticError::Overflow)?;
-				T::InterestAccrual::reference_rate(interest_rate_with_penalty);
+				T::InterestAccrual::reference_rate(interest_rate_with_penalty)?;
```
mustermeiszer, line 755: Adding write-off groups with penalty rates of e.g. `0.00001` is not checked, right? So this would break or invariant of only having four decimal interest rates in the system.

pallets/loans/src/functions.rs
```diff
@@ -227,6 +220,8 @@ impl<T: Config> Pallet<T> {
 					active_loan.normalized_debt,
 				)?;
 
+				T::InterestAccrual::unreference_rate(active_loan.interest_rate_per_sec)?;
```
mustermeiszer, line 223: Maybe do an 

`if active_loan.interest_rate_per_sec != interest_rate_per_sec` and then only update interest rate. Minor but safes one storage reads and writes.

pallets/loans/src/functions.rs
```diff
@@ -187,23 +183,20 @@ impl<T: Config> Pallet<T> {
 		let count = active_loans.len();
 		ActiveLoans::<T>::insert(pool_id, active_loans);
 
-		Ok(count.try_into().unwrap())
+		Ok((count.try_into().unwrap(), interest_rate_per_sec))
```
mustermeiszer, line 186: This will be a silent error in case we allow above `u32::max errors`. Probably never going to happen but I don't like unwrapping in the runtime.

pallets/interest-accrual/src/lib.rs
```diff
@@ -383,6 +421,23 @@ pub mod pallet {
 			});
 			T::DbWeight::get().reads_writes(count, count)
 		}
+
+		pub fn count_rates() -> Weight {
+			let count = Rate::<T>::iter_keys().count();
+			RateCount::<T>::set(count as u32);
+			T::DbWeight::get().reads(count as u64) + T::DbWeight::get().writes(1)
+		}
```
mustermeiszer, line 429: Is this migration-code? 

can we move this into a migration module. Othweise, its not simple to see which code can be removed or so in the future.

runtime/altair/src/lib.rs
```diff
@@ -1332,7 +1333,9 @@ mod upgrade {
 			weight += InterestAccrual::upgrade_to_v1();
 			weight += Loans::reference_active_rates();
 			weight += InterestAccrual::remove_unused_rates();
+			weight += InterestAccrual::count_rates();
```
mustermeiszer, line 1336: did this migration already run on the last runtime upgrade? 

At least the `InterestAccrual::removed_unused_rates()` and the `InterestAccrual::count_rates()` are running on each upgrade. Is this intended and what is their purpose? 

I guess my main point is, that I would like to have them behind some versioned logic or some docs on when they can be removed from the upgrade logic, so that we do not forget them.

mustermeiszer (commented):
This looks good. I am not approving because we will allow arbitrary interest rates to be added via the write-off groups. We should also check them and error out if they have too many decimals. WDYT? 

Otherwise just some minor nitpicks about using checked and saturating arithmetics. Furthermore, I would like to have the code solely used for migrations being located in the migration module as they do not seem to make sense in the lib.

pallets/interest-accrual/src/lib.rs
```diff
@@ -334,16 +347,41 @@ pub mod pallet {
...
+					ensure!(
+						rate_count < T::MaxRateCount::get(),
+						Error::<T>::TooManyRates
+					);
+
 					*rate = Some(RateDetailsOf::<T> {
 						accumulated_rate: One::one(),
 						reference_count: 1,
 					});
+					RateCount::<T>::set(rate_count + 1);
```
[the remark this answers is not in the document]
branan: Went for a checked add and erroring out. We should never hit this case either way (rate_count is limited by the configuration), but I'd rather not end up in a weird saturated-land when it comes to computing weights

pallets/loans/src/lib.rs
branan: This migration was a big of a mess due to needing to coordinate data between interest-accrual and the loans pallet. It's already been applied at this point, and just needs to be deleted in a future PR.

pallets/loans/src/functions.rs
branan: Changed this to an `expect` explaining why it can't fail. Unfortunately these conversions between `usize` and `u32` are _everywhere_ due to Parity using u32 in lots of APIs that then map to a Rust `usize`

pallets/interest-accrual/src/lib.rs
branan: Moved to hooks and version-checked

runtime/altair/src/lib.rs
branan: Moved this to a normal migration in the hooks of the pallet

pallets/loans/src/functions.rs
branan: That would require splitting up some of the logic of computing whether the rate is valid and then referencing it later. I kind of like that the API exposed by interest-accrual is currently atomic in that regard (except for penalty rates, of course)

pallets/interest-accrual/src/lib.rs
```diff
@@ -422,11 +485,27 @@ impl<T: Config> InterestAccrual<T::InterestRate, T::Balance, Adjustment<T::Balan
...
+
+	fn reference_rate(interest_rate_per_sec: T::InterestRate) -> Result<(), DispatchError> {
 		Pallet::<T>::reference_interest_rate(interest_rate_per_sec)
 	}
 
 	fn unreference_rate(interest_rate_per_sec: T::InterestRate) -> Result<(), DispatchError> {
 		Pallet::<T>::unreference_interest_rate(interest_rate_per_sec)
 	}
+
+	fn verify_penalty_rate(
```
mustermeiszer, line 502: Could we name it to something less connected to write-offs. Maybe just `verify_rate`

libs/traits/src/lib.rs
```diff
@@ -193,11 +193,23 @@ pub trait InterestAccrual<InterestRate, Balance, Adjustment> {
...
+
 	/// Indicate that a rate is in use
-	fn reference_rate(interest_rate_per_sec: InterestRate);
+	fn reference_rate(interest_rate_per_sec: InterestRate) -> DispatchResult;
 
 	/// Indicate that a rate is no longer in use
 	fn unreference_rate(interest_rate_per_sec: InterestRate) -> DispatchResult;
+
+	/// Verifies a yearly penalty rate and converts it to a per-second penalty rate
+	fn verify_penalty_rate(
```
mustermeiszer, line 210: naming to something less specifc. We can verify any rate with this. WDYT?

mustermeiszer (commented):
Don't wanna flag as request for changes as you will come online later and need to wait for me then, but I would definitely rename the `verify_penalty_rate` to something like `verify_yearly_rate`.

Otherwise, looks good to go.

pallets/interest-accrual/src/lib.rs
branan: Really, `verify` is the wrong verb, as it both verifies and converts to a per-second rate. The thing that makes it penalty-specific is that penalty rates are still of the form "0.xxxxxxx" instead of "1.xxxxx" for normal rates. Maybe we want to use a more generic term for "rates which can be used to manipulate other rates" instead of "rates that are meant to stand on their own as *real rates*? I'm not sure

runtime/altair/src/lib.rs
```diff
@@ -1160,6 +1160,7 @@ impl pallet_interest_accrual::Config for Runtime {
 	type Balance = Balance;
 	type Event = Event;
 	type InterestRate = Rate;
+	type MaxRateCount = MaxActiveLoansPerPool;
```
offerijns, line 1163: Why would this be equivalent?

branan: The current benchmarking code requires `MaxRateCount` to be greater or equal than `MaxActiveLoansPerPool` (it creates a unique rate for every loan).

The right values for MaxRateCount and MaxActiveLoansPerPool need to be udpated based on benchmarking, but I haven't done production benchmarking on this version of the code - I was gonna wait for Part 2 for that. I can add a TODO comment and a github ticket to adjust this based on benchmarking results.

branan: In other words, it's an expedient temporary value

offerijns: Makes sense. I think adding a TODO comment would be great 👍

pallets/interest-accrual/src/tests.rs
```diff
@@ -0,0 +1,37 @@
...
+	FixedPointNumber,
+};
+
+use crate::{
+	mock::{Rate, Test as TestConfig},
+	Pallet,
+};
+
+#[test]
+fn test_rate_validation() {
```
offerijns, line 25: Should we add a ticket to add some more tests to this pallet? @branan

offerijns (approved):

pallets/interest-accrual/src/tests.rs
branan: We already have one. #849.

Most of the interesting code paths are covered by the loans tests anyway, but it's be good to have dedicated tests (and good to make loans less fragile by removing it validating internal/intermediate values from i-i)

PR #989, opened 2022-09-21
Bump srtool build to Rust 1.62

branan:
At this point one of Parity's libraries requires 1.62, so we can't stay on 1.60 any longer.


branan:
I tested this with a temporary commit that enabled the runtime build in this PR and it went green. Auto-merge is on, somebody just needs to approve this

mikiquantum (approved):
👍

PR #994, opened 2022-09-22
Fix - remove dependencies to external `node_primitives`

mustermeiszer:
# Description
We currently still rely on the external library `node_primitives` to define our opaque node block type. Rather we must rely on our internal primitives to ensure we do not include downstream changes.

## Changes and Descriptions
Solely changes the node block type. The types are identical, but this protects us in the future, to have it in our own hands whether we wanna change the node block type.

# How Has This Been Tested?
This is not tested, besides the normal integration tests. But the typing of both node primitives and our block are identical see

* node primitives reference
   * Block and header: https://github.com/paritytech/substrate/blob/cc4d5cc8654d280f03a13421669ba03632e14aa7/bin/node/primitives/src/lib.rs#L62-L64
   * Opaque extrinsic: https://github.com/paritytech/substrate/blob/cc4d5cc8654d280f03a13421669ba03632e14aa7/primitives/runtime/src/lib.rs#L855
* our reference: https://github.com/centrifuge/centrifuge-chain/blob/09a5e3006d6f5aa5eed5002b8a2549ea450f68e9/libs/primitives/src/lib.rs#L95-L98

# Checklist:


NunoAlexandre (approved):
The changes look good to me! Is there any scenario we can think of where not adopting the same block type as the node-primitives could cause issues or incompatibilities? I can't think of one now but it's a good exercise to do before merging this

mustermeiszer:
> The changes look good to me! Is there any scenario we can think of where not adopting the same block type as the node-primitives could cause issues or incompatibilities? I can't think of one now but it's a good exercise to do before merging this

Not sure, but all teams use their own types there. As far as I remember the relay-chain also uses an opaque block type and decodes firstly on the cumulus side, where there is awareness of the type we choose.

NunoAlexandre (approved):

Issue #931, opened 2022-08-31
Custom Rounding for `Rate`


mustermeiszer:
Implement a few methods for our custom `FixedPointNumber` type `Rate` in `runtime_common`, that allow us to choose a rounding preference from time to time. One important aspect here is, that we use `Rate` always behind a trait bound. Meaning we can not use non-trait methods. Hence, we must define a trait for this.

### Implementation Draft
```rust
// Not sure if we actually need FixedPointOperand as a super trait
pub trait FixedPointOperandExtension: FixedPointOperand {
       fn checked_mul_floor(&self, int: &Self::Inner) -> Option<Self::Inner>;

       fn checked_mul_ceil(&self, int: &Self::Inner) -> Option<Self::Inner>;

       fn saturating_mul_floor(&self, int: &Self::Inner) -> Self::Inner;

       fn saturating_mul_ceil(&self, int: &Self::Inner) -> Self::Inner;
}
```

thea-leake:
PR for issue now in review: https://github.com/centrifuge/centrifuge-chain/pull/1006

PR #995, opened 2022-09-23
Pallet-Investments: Integration into pallet-pools

### LINKED ISSUE #916: Integrate `pallet-investments` and `pallet-pools`
once #909 is close, we must integrate the `pallet-investments` into the `pallet-pools` logic. 

This includes among others: 
* Remove the investment fields on the `struct Tranche`
* Remove all now invalid methods on `struct Tranche` and `struct Tranches` that are concerned with investments
* Add a field to the pools config that is bound by `trait InvestmentManager`
* Implement `trait AssetAccountant` and `trait AssetProperties` by the pools pallet
* Adapt logic to require orders via `T::InvestmentManager::orders()` for every tranche upon epoch closing and call `T::InvestmentManager::fulfilments` once we close the epoch and know the fulfillments.

mustermeiszer:
# Description
This PR integrates the new pallet-investment logic into the pallet-pools.

Closes #916

## Changes and Descriptions
The most notable changes are listed below. A special focus should lie on:
- `pallets/pools/src/lib.rs`
   - Removed two storage items completely - `Orders` and `Epoch`
   - Preparing of tranches - i.e. burning and minting of tranche tokens - is now done by the `trait Accountant` implementation
-  `pallets/pools/src/tranche.rs` 
   - Removes all information about investments and redemptions from both tranche structures
   - Introduces a new generic parameter - `TrancheCurrency` - that is a currency of a tranche. This currency must implement the new `trait TrancheCurrency`, which allows to generate it from `PoolId` and `TrancheId` and also allows to retrieve a `PoolId` and `TrancheId` from the currency struct itself again. 
- `pallets/pools/src/migrations/mods.rs`
   - Removes the above mentioned storage items
   - Migrates `EpochExecution` and `Pools` as we have a new tranche currency now
## Type of change

# How Has This Been Tested?
- Unit tests
- Try-runtime logic for migration as a test

# Checklist:

# Tasks
- [x] Integrate investment logic into pool
- [x] Adapt unit-tests to be green
- [x] Write migration for `Pool`
- [x] Write migration for `EpochExecution`
- [ ] ~~Write migrations for `Order`~~ 
We will rather remove all orders and collects from the pool on Altair
- [x] Include migration in Altair
- [x] Test migration locally
- [x] Adapt Pallet-Investment logic with explicit floor rounding and adapt tests (needs #931)


mustermeiszer:
@branan and @gruberb you two can already take a look. I still need to adjust the tests and the runtimes, write migrations.

But the logic in the pallet itself is adapted.

libs/primitives/src/lib.rs
```diff
@@ -39,9 +39,12 @@ pub mod types {
 	/// PoolId type we use.
 	pub type PoolId = u64;
 
-	/// OrderId type we to identify order per epoch.
+	/// OrderId type we use to identify order per epoch.
 	pub type OrderId = u64;
 
+	/// EpochId type we use to identify epochs in our revolving pools
+	pub type EpochId = u32;
```
offerijns, line 46: The rewards that @lemunozm and @mikiquantum are working on will also have epochs. Should we separate the types, and call this `PoolEpochId`? Or use the same type for both?

pallets/pools/src/lib.rs
```diff
@@ -166,104 +257,71 @@ where
...
-			epoch: One::one(),
-		}
-	}
-}
-
-/// The result of epoch execution of a given tranch within a pool
-#[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, Default, TypeInfo)]
-pub struct EpochDetails<BalanceRatio> {
-	pub invest_fulfillment: Perquintill,
-	pub redeem_fulfillment: Perquintill,
```
offerijns, line 195: I think we spoke about this before, but if we don't store this information anymore, how can we ensure (just using the `pools` pallet) that a scheduled update is only executed when all redemptions are fulfilled? Right now this is checking if `epochs[epochs.len - 1].redeem_fulfillment == 100%`.

pallets/pools/src/lib.rs
```diff
@@ -290,18 +348,12 @@ pub mod pallet {
...
-			+ TypeInfo
-			+ FixedPointNumber<Inner = Self::Balance>;
-
-		/// A fixed-point number which represents an
-		/// interest rate.
-		type InterestRate: Member
+		/// A fixed-point number which represents a Self::Balance
+		/// in terms of this fixed-point representation.
+		///
+		/// NOTE: One must be cautious with using One::one() in
```
offerijns, line 354: Why must one be cautious with using One::one() in this context?

pallets/pools/src/lib.rs
```diff
@@ -983,216 +991,6 @@ pub mod pallet {
 			})
 		}
 
-		/// Update an order to invest tokens in a given tranche.
```
offerijns, line 986: Nice to see all this code being cleaned up :)

pallets/pools/src/tranche.rs
```diff
@@ -1344,57 +1346,6 @@ where
...
-	pub fn acc_redemptions(&self) -> Result<Balance, DispatchError> {
-		self.tranches
-			.iter()
-			.fold(Some(Balance::zero()), |sum, tranche| {
-				sum.and_then(|acc| acc.checked_add(&tranche.redeem))
-			})
-			.ok_or(ArithmeticError::Overflow.into())
-	}
-
 	pub fn calculate_weights(&self) -> Vec<(Weight, Weight)> {
```
offerijns, line 1349: This is unrelated to this PR, but I noticed `calculate_weights` is duplicated in this file. Can/should we clean that up? If so, should we create a ticket for that?

pallets/pools/src/lib.rs
offerijns: Just saw the code in the runtime lib. That works, but it does not guarantee the update invariant anymore solely using the `pools` pallet, as it now relies on the `investments` pallet. I think that's fine for now, but let's discuss later if this is fine or there might be an alternative approach.

runtime/altair/src/lib.rs
```diff
@@ -1201,6 +1214,71 @@ impl pallet_interest_accrual::Config for Runtime {
...
+		};
+
+		if is_whitelisted {
+			Ok(())
+		} else {
+			// TODO: We should adapt the permissions pallets interface to return an error instead of a boolen. This makes the redundant has not role error
+			//       that downstream pallets always need to generate not needed anymore.
+			Err(DispatchError::Other(
+				"Investor not whitelisted for pool in tranche.",
+			))
```
offerijns, line 1277:
```suggestion
		let is_tranche_investor = match order {
			OrderType::Investment {
				who,
				investment_id: tranche,
				..
			} => P::has(
				PermissionScope::Pool(tranche.of_pool()),
				who,
				Role::PoolRole(PoolRole::TrancheInvestor(
					tranche.of_tranche(),
					T::now().as_secs(),
				)),
			),
			OrderType::Redemption {
				who,
				investment_id: tranche,
				..
			} => P::has(
				PermissionScope::Pool(tranche.of_pool()),
				who,
				Role::PoolRole(PoolRole::TrancheInvestor(
					tranche.of_tranche(),
					T::now().as_secs(),
				)),
			),
		};

		if is_tranche_investor {
			Ok(())
		} else {
			// TODO: We should adapt the permissions pallets interface to return an error instead of a boolen. This makes the redundant has not role error
			//       that downstream pallets always need to generate not needed anymore.
			Err(DispatchError::Other(
				"Account does not have the TrancheInvestor permission.",
			))
```

runtime/development/src/lib.rs
```diff
@@ -443,9 +446,9 @@ impl InstanceFilter<Call> for ProxyType {
 			ProxyType::Price => matches!(c, Call::Loans(pallet_loans::Call::price { .. })),
 			ProxyType::Invest => matches!(
 				c,
-				Call::Pools(pallet_pools::Call::update_invest_order{..}) |
-				Call::Pools(pallet_pools::Call::update_redeem_order{..}) |
-				Call::Pools(pallet_pools::Call::collect{..}) |
+				Call::Investments(pallet_investments::Call::update_invest_order{..}) |
```
offerijns, line 449: Good catch :)

runtime/development/src/lib.rs
```diff
@@ -1358,6 +1371,71 @@ impl pallet_keystore::pallet::Config for Runtime {
...
+			),
+		};
+
+		if is_whitelisted {
+			Ok(())
+		} else {
+			// TODO: We should adapt the permissions pallets interface to return an error instead of a boolen. This makes the redundant has not role error
+			//       that downstream pallets always need to generate not needed anymore.
+			Err(DispatchError::Other(
+				"Investor not whitelisted for pool in tranche.",
```
offerijns, line 1433:
```suggestion
		let is_tranche_investor = match order {
			OrderType::Investment {
				who,
				investment_id: tranche,
				..
			} => P::has(
				PermissionScope::Pool(tranche.of_pool()),
				who,
				Role::PoolRole(PoolRole::TrancheInvestor(
					tranche.of_tranche(),
					T::now().as_secs(),
				)),
			),
			OrderType::Redemption {
				who,
				investment_id: tranche,
				..
			} => P::has(
				PermissionScope::Pool(tranche.of_pool()),
				who,
				Role::PoolRole(PoolRole::TrancheInvestor(
					tranche.of_tranche(),
					T::now().as_secs(),
				)),
			),
		};

		if is_tranche_investor {
			Ok(())
		} else {
			// TODO: We should adapt the permissions pallets interface to return an error instead of a boolen. This makes the redundant has not role error
			//       that downstream pallets always need to generate not needed anymore.
			Err(DispatchError::Other(
				"Account does not have the TrancheInvestor permission.",
```

offerijns (commented):
Just took a first pass at this and left a few comments. This is looking really good, and it was clearly a huge amount of work 🔥  I really like how it cleans up the `pools` pallet and strips it down to the core logic that it really needs!

libs/primitives/src/lib.rs
mustermeiszer: Are the reward epochs referring to the pools epochs? If not, I would separate the naming.

pallets/pools/src/lib.rs
mustermeiszer: As `One::one().into_inner()` is of type `Balance` but has a random decimal (from the perspective of the pallet. Maybe the doc comment is more misleading than clarifying.

pallets/pools/src/lib.rs
mustermeiszer: That is indeed true. I wanted to bring this in with the `UpdateGuard` changes. Seperate from registry and this integration. 

This should definitely be stored in the pools. Just wanted to give this a little more thought, how we actually store infos in the pools that ensure the pool itself can ensure all invariants itself.

pallets/pools/src/tranche.rs
mustermeiszer: Removed.

libs/primitives/src/lib.rs
offerijns: No they are separate epochs. So let's rename this to `PoolEpochId` then? And @lemunozm maybe you want to use the `RewardEpochId` naming in your PR?

pallets/pools/src/lib.rs
offerijns: > Maybe the doc comment is more misleading than clarifying.
Yeah I think this is the case, I'd either remove it or add the context above.

pallets/investments/src/lib.rs
```diff
@@ -499,10 +501,36 @@ pub mod pallet {
...
 
-		/// Collect the results of another users orders for the given investment.
-		///
-		/// The type of collection will always be `CollectType::Closing`.
-		#[pallet::weight(0)]
+		/// Collect the results of a users invest orders for the given investment.
+		/// If any amounts are not fulfilled they are directly appended to the next active
+		/// order for this investment.
+		#[pallet::weight(80_000_000)]
+		pub fn collect_investments(
```
offerijns, line 508: Why do we need these? It seems like this is just adding complexity, while just `collect()` would suffice?

pallets/investments/src/lib.rs
```diff
@@ -975,9 +997,9 @@ where
 			.checked_add(
 				&fulfillment
 					.price
-					.reciprocal()
+					.reciprocal_floor()
```
offerijns, line 1000: This is great! Maybe you can add a comment though, explaining why you are rounding down here?

runtime/altair/src/lib.rs
```diff
@@ -1178,11 +1182,32 @@ impl PoolUpdateGuard for UpdateGuard {
...
+		let pool_id = pool.tranches.of_pool();
+
+		// We do not allow releasing updates during epoch
+		// closing.
+		//
+		// This is needed as:
+		// - investment side starts new order round with zero orders at epoch_closing
+		// - the pool might only fulfill x < 100% of redemptions
+		//         -> not all redemptions would be fulfilled after epoch_execution
+		if Pools::epoch_targets(pool_id).is_some() {
```
offerijns, line 1194: Good catch!

runtime/altair/src/lib.rs
```diff
@@ -1178,11 +1182,32 @@ impl PoolUpdateGuard for UpdateGuard {
...
-			.acc_outstanding_redemptions()
-			.unwrap_or(Balance::MAX);
+			.ids_non_residual_top()
+			.iter()
+			.map(|tranche_id| {
+				let investment_id = TrancheCurrency::generate(pool_id, *tranche_id);
+				Investments::redeem_orders(investment_id).amount
+			})
+			.fold(Balance::zero(), |acc, redemption| {
+				acc.saturating_add(redemption)
```
offerijns, line 1208: I wonder if this shouldn't be a `checked_add` rather than a `saturating_add` and error out if it overflows?

libs/primitives/src/lib.rs
mustermeiszer: Ping @lemunozm

mustermeiszer: I will rename

pallets/investments/src/lib.rs
mustermeiszer: Yes, thats definetly needed. Will add that.

pallets/investments/src/lib.rs
mustermeiszer: we are solely adding a public api endpoint. The logic is there anyways. The frontend should be able to see, if only investments should be collected and then call this endpoint. If there are both redemptions and investments to be collected it should call the other one. From a user perspective always collecting both is forcing them to pay for something they do not need. 

In cases, where this would duplicate logic or introduce complexity I wouldn't add this, but this is really just the endpoint.

offerijns: Yeah that makes sense. But, shouldn't we just remove `collect()` then, and users can just batch the other two methods if needed? Given that the weight of `collect` is simply `weight(collect_investments) + weight(collect_redemptions)`?

mustermeiszer: Yes, that is a good idea. For the collect_for I think forcing collecting of both is also not necessarily needed. I thought about any advantage one could have by only collecting one part for others. But there isn't actually one, when I think about it now. Wdyt?

offerijns: I can't think of any either, I think just rewriting that to `collect_investments/redemptions_for` would be fine.

branan (dismissed):
🙈 

In service of getting this in I refocused on tests and benchmarks to ensure those seemed sane. They do, and nothing jumps out in the rest of the code.

Some style nits, but as long as Clippy and rustfmt are happy I'm not gonna make a big deal of them.

mustermeiszer:
> In service of getting this in I refocused on tests and benchmarks to ensure those seemed sane. They do, and nothing jumps out in the rest of the code.
Some style nits, but as long as Clippy and rustfmt are happy I'm not gonna make a big deal of them.

I think it would be important to look especially at the changes to pallet-pools. @offerijns How time bound is this? Can we wait till tomorrow? Then @branan could take a look at the actual integration. As tests and benchmarks remain the same, I solely adapted them to cope with the investments pallet, but logically they are untouched.

branan (approved):
Still not seeing anything that looks wrong in the bookkeeping between pools&investments, or in the migration code.

Some more nits, but as before I don't think they're worth debating in this PR or blocking it on. 

I will probably toss up a PR with some proposed changes based on the things I'd have chosen to do slightly differently, and we can make decisions on that stuff there, instead of here.

branan:
And of course someone merged something so I can't land this now with my approval... But I think once you rebase you can get somebody in the EU timezone to add their approval and land this without needing to wait for my morning

mustermeiszer:
> Still not seeing anything that looks wrong in the bookkeeping between pools&investments, or in the migration code.

> Some more nits, but as before I don't think they're worth debating in this PR or blocking it on.

> I will probably toss up a PR with some proposed changes based on the things I'd have chosen to do slightly differently, and we can make decisions on that stuff there, instead of here.

Thanks for being so fast. I think that is a good idea.

NunoAlexandre (approved):
Relaying Branan's approval 🏅

PR #997, opened 2022-09-23
Drop pallet_anchors and pallet_fees migration

NunoAlexandre:
# Description

Drop all migration-related code for the the `pallet_anchors` and `pallet_fees`.

Fixes # (issue)

Closes https://github.com/centrifuge/centrifuge-chain/issues/996

## Changes and Descriptions

Cover in description

## Type of change

- Drop stale / no longer needed code

# How Has This Been Tested?

The migrations have already been executed so no more specific tests needed at this time.

# Checklist:


mikiquantum (approved):
LGTM, maybe is a good time to remove the pools/loans migrations of altair too. @branan WDYT? we can do it in a different PR either way this is good 👍

pallets/anchors/src/migration.rs
```diff
@@ -1,81 +0,0 @@
...
-		ensure!(
-			LatestEvictedDate::<T>::get() == None,
-			"State already initialized"
-		);
-		Ok(())
-	}
-
-	pub fn migrate<T: Config>() -> Weight {
-		if LatestEvictedDate::<T>::get().is_none() {
-			LatestEvictedDate::<T>::put(HARDCODED_EVICTED_DATE);
```
lemunozm, line 24: I can see this value in Altair, so the migration seems to be correctly applied. It can be removed 👍🏻

pallets/fees/src/migration.rs
```diff
@@ -1,102 +0,0 @@
...
-	<T as frame_system::Config>::Hash,
-	Fee<<T as frame_system::Config>::Hash, BalanceOf<T>>,
->;
-
-pub mod fee_balances {
-	use super::*;
-
-	#[cfg(feature = "try-runtime")]
-	pub fn pre_migrate<T: Config>() -> Result<(), &'static str> {
-		let count = Fees::<T>::iter_values().count();
```
lemunozm, line 44: I'm not sure how I can tests if these elements are removed because the storage is no longer referenced from `polkadot.js`. I'm 99.9% pretty sure it has already been removed correctly but do you know how to check it?

mikiquantum: @lemunozm You can either:
- Generate a valid storage key and query the raw storage on polkadotjs/apps
- Run this migration code with `try-runtime` code in your local machine to ensure that the pre_migrate fails :)

lemunozm: Ok thanks!

PR #1008, opened 2022-09-30
Utility: add ensure_op trait methods to simplify overflow errors.

lemunozm:
# Description

This PR adds `ensure_add` and `ensure_sub` trait methods to avoid being extremely verbose in math operations overflow checks, making the code more readable.

Before:
```rust
0u32.checked_sub(1).ok_or(ArithmeticError::Underflow)?;
```
After:
```rust
0u32.ensure_sub(1)?;
```

## Changes and Descriptions

Addition of new ops traits into `libs/traits`

## Type of change

# How Has This Been Tested?
```rust
cargo test -p cfg-traits // Which will run the associated doc examples
```

### DIFF
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -696,7 +696,7 @@ pub mod fees {
 	}
 }
 
-mod ops {
+pub mod ops {
 	pub use sp_runtime::{
 		traits::{CheckedAdd, CheckedDiv, CheckedMul, CheckedSub},
 		ArithmeticError,
@@ -709,7 +709,15 @@ mod ops {
 		/// `ArithmeticError::Overflow` is returned.
 		///
 		/// ```
-		/// u32::max().ensure_add(1)?
+		/// use cfg_traits::ops::EnsureAdd;
+		/// use sp_runtime::{DispatchResult, ArithmeticError, DispatchError};
+		///
+		/// fn extrinsic() -> DispatchResult {
+		///     u32::MAX.ensure_add(&1)?;
+		///     Ok(())
+		/// }
+		///
+		/// assert_eq!(extrinsic(), Err(DispatchError::Arithmetic(ArithmeticError::Overflow)));
 		/// ```
 		fn ensure_add(&self, v: &Self) -> Result<Self, ArithmeticError> {
 			self.checked_add(v).ok_or(ArithmeticError::Overflow)
@@ -723,7 +731,15 @@ mod ops {
 		/// `ArithmeticError::Underflow` is returned.
 		///
 		/// ```
-		/// 0u32.ensure_sub(1)?
+		/// use cfg_traits::ops::EnsureSub;
+		/// use sp_runtime::{DispatchResult, ArithmeticError, DispatchError};
+		///
+		/// fn extrinsic() -> DispatchResult {
+		///     0u32.ensure_sub(&1)?;
+		///     Ok(())
+		/// }
+		///
+		/// assert_eq!(extrinsic(), Err(DispatchError::Arithmetic(ArithmeticError::Underflow)));
 		/// ```
 		fn ensure_sub(&self, v: &Self) -> Result<Self, ArithmeticError> {
 			self.checked_sub(v).ok_or(ArithmeticError::Underflow)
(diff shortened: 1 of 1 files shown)


mustermeiszer (dismissed):
So clean and nice and tidy!!! ❤️

mustermeiszer:
Just one additional comment. Can we add the same for `checked_mul`, `checked_div`, `checked_pow`?

lemunozm:
Sure! even I was thinking of adding `checked_mul_int` from `FixedPointNumber` too

lemunozm:
Added `ensure_mul` and `ensure_div`. Nevertheless `ensure_pow` can not be (easily) done because there is no `CheckedPow` trait in `num_traits`.

Regarding `ensure_mul`, the `checked_mul` can fail for both overflow and underflow, but I'm not sure how I can distinguish them from within the default trait implementation. I think I would need to implement the trait for all cases by myself or making some other helping trait (that seems to no exists) as `Signum`:
```rust
pub trait EnsureMul: CheckedMul + Signum {
    fn ensure_mul(&self, v: &Self) -> Result<Self, ArithmeticError> {
         self.checked_mul(v).ok_or_else(|| match self.signum() * v.signum() {
             1 =>  ArithmeticError::Overflow,
             -1 =>  ArithmeticError::Underflow,
         })
    }
}
```

mustermeiszer:
Ah yes, I did not look correctly. It's coming from Substrate directly.

mustermeiszer (dismissed):
LGTM!

libs/traits/src/lib.rs
```diff
@@ -695,3 +695,71 @@ pub mod fees {
...
+	/// wrapping around on overflow.
+	pub trait EnsureMul: CheckedMul {
+		/// Multiplies two numbers, checking for overflow. If overflow happens,
+		/// `ArithmeticError::Overflow` is returned.
+		///
+		/// ```
+		/// u32::max().ensure_mul(2)?
+		/// ```
+		fn ensure_mul(&self, v: &Self) -> Result<Self, ArithmeticError> {
+			self.checked_mul(v).ok_or(ArithmeticError::Overflow)
```
lemunozm, line 743: I leave a tech debt here to fix the case where the overflow is, in fact, underflow.

libs/traits/src/lib.rs
```diff
@@ -695,3 +695,71 @@ pub mod fees {
...
+	/// wrapping around on overflow.
+	pub trait EnsureDiv: CheckedDiv {
+		/// Divides two numbers, checking for overflow. If overflow happens,
+		/// `ArithmeticError::DivisionByZero` is returned.
+		///
+		/// ```
+		/// u32::max().ensure_div(0)?
+		/// ```
+		fn ensure_div(&self, v: &Self) -> Result<Self, ArithmeticError> {
+			self.checked_div(v).ok_or(ArithmeticError::DivisionByZero)
```
lemunozm, line 757: I leave a tech debt here to fix the case where the division error is due to underflow or overflow. To be honest, I do not understand when it can happen in a division, but the [`checked_div`](https://docs.rs/num-traits/latest/num_traits/ops/checked/trait.CheckedDiv.html#tymethod.checked_div) says that it may happen (maybe with float numbers only?).

libs/traits/src/lib.rs
mustermeiszer: We can write a small macro and only implement it for unsigned and signed? WDYT?

libs/traits/src/lib.rs
mustermeiszer: We can write a small macro and only implement it for unsigned and signed? WDYT?

Who uses floats anyways ^^

mustermeiszer: Where did you get that info from?

lemunozm: From the documentation of `checked_div`: *Divides two numbers, checking for underflow, overflow, and division by zero. If any of that happens, None is returned.*

PS. floats are for losers 😎

libs/traits/src/lib.rs
lemunozm: Maybe with two different macros:
- If it's unsigned, ok, it will always be Overflow. Current behavior.
- If it's signed, it can be Overflow or Underflow, depending on the sign of multiplying both numbers.

mustermeiszer: Maybe lets make solely impls for unsigned. We have no signed logic currently

I was thinking of something like this.
```rust
ensure_ops_signed_impl!(u8);
ensure_ops_signed_impl!(u16);
ensure_ops_signed_impl!(32);
ensure_ops_signed_impl!(u64);
ensure_ops_signed_impl!(u128);
```

mustermeiszer (approved):
What do you think about restricting this to `u*` types?

PR #1010, opened 2022-10-03
pools: Add PalletIndex type

NunoAlexandre:
# Description

Currently, all of our native assets are described in `MultiLocation` as `X2(Parachain(_), GeneralKey(y))`. This works well for all assets with a static general key since we can _match_ on those static, const values. However, with Tranche tokens, as their general key is only obtained at runtime, we don't have a way of matching a MultiLocation representing a tranche apart from an exclusion-based approach.

Since we don't have tranche assets being used in production yet, hereby I suggest that we change the MultiLocation we use for tranche assets to be `X3(Parachain(x), PalletInstance(<pools_pallet_index>), GeneralKey(y))`, making it super easy to match on tranche tokens.

This comes particularly helpful in order to easily block transfers made through PolkadotXcm or XTokens that involve tranche tokens, which are disallowed to avoid transfers to non-whitelisted destination addresses. We now block all the extrinsics of those two pallets that allow for an arbitrary number of assets to be transferred, which we need to have enabled for USDC transfers where GLRM also needs to be sent alongside to cover for fee costs.

## Changes and Descriptions

1. We change the multilocation to which tranche tokens are represented as
2. We pass a `PalletIndex` type param to the pools pallet

## Type of change

^ I mean, in theory it's a breaking change but in practise I don't believe it's breaking anything.

# How Has This Been Tested?

We have adapted the affected tests and mocks in the pools pallet to cover this change.

# Checklist:

### DIFF
--- a/pallets/loans/src/mock.rs
+++ b/pallets/loans/src/mock.rs
@@ -26,7 +26,7 @@ use cfg_types::{
 };
 use frame_support::{
 	parameter_types,
-	traits::{AsEnsureOriginWithArg, Everything, GenesisBuild, SortedMembers},
+	traits::{AsEnsureOriginWithArg, Everything, GenesisBuild, PalletInfoAccess, SortedMembers},
 	PalletId,
 };
 use frame_system::{EnsureSigned, EnsureSignedBy};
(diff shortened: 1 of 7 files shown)


NunoAlexandre:
@mustermeiszer @offerijns can we get your reviews on this one please 🙏

pallets/loans/src/mock.rs
```diff
@@ -152,6 +152,9 @@ impl orml_tokens::Config for MockRuntime {
 parameter_types! {
 	pub const PoolPalletId: frame_support::PalletId = cfg_types::ids::POOLS_PALLET_ID;
 
+	/// The index with which this pallet is instantiated in this runtime.
+	pub PoolPalletIndex: u8 = <Pools as PalletInfoAccess>::index() as u8;
```
mustermeiszer, line 156: What size is `Index` that we need to cast it? Will this be safe? Can we have test in case this is overflowing? ^^

mustermeiszer (approved):
If you think this will be a good decision I trust you and it looks reasonable to me! 

The `GeneralKey` should then be unique across our Domain, right? At least in the context of the given `PalletInstance`, right?

mustermeiszer:
No, I like the `X3`! In general XCM seems to be really bad ad addressing tokens currently. At least without using `GeneralIndex` and `GeneralKey`...

PR #1017, opened 2022-10-05
runtime: Set sane BaseCallFilter

NunoAlexandre:
Some PolkadotXcm extrinsics allow any user to prepare and send raw XCM messages which is undesired so we block those calls, leaving those only open for Root to execute.

### DIFF
--- a/runtime/altair/src/lib.rs
+++ b/runtime/altair/src/lib.rs
@@ -192,7 +192,6 @@ impl frame_system::Config for Runtime {
 }
 
 /// Base Call Filter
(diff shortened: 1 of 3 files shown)


NunoAlexandre:
@mikiquantum I was able to verify the following in my local environment:

- Random users can NOT call polkadotXcm::send, it fails with the expected call filtered error
- Through sudo I could call polkadotXcm::send, i.e, Root can call that extrinsic
- Through sudo I could send execute batch calling polkadot::send

mikiquantum (approved):
👍

PR #1031, opened 2022-10-13
Add EnsureOpsAssign traits

lemunozm:
# Description

This PR continues the work on https://github.com/centrifuge/centrifuge-chain/pull/1008, extending the ops with *assign* traits.

Before:
```rust
let mut my_long_variable_name: u32 = 0;
my_long_variable_name = my_long_variable_name.checked_sub(1).ok_or(ArithmeticError::Underflow)?;
```
After:
```rust
let mut my_long_variable_name: u32 = 0;
my_long_variable_name.ensure_sub_assign(1)?;
```

## Changes and Descriptions

Addition of new ops traits into `libs/traits`:
- `EnsureAddAssign`
- `EnsureSubAssign`
- `EnsureMulAssign`
- `EnsureDivAssign`
- Correct multiplication underflow behavior
- Correct addition underflow behavior
- Correct subtraction overflow behavior

## Type of change

# How Has This Been Tested?

```sh
cargo test -p cfg-traits // Which will run the associated doc examples
```

# Checklist:

### DIFF
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -709,10 +709,47 @@ pub mod fees {
 }
 
 pub mod ops {
-	pub use sp_runtime::{
+	use sp_runtime::{
 		traits::{CheckedAdd, CheckedDiv, CheckedMul, CheckedSub},
 		ArithmeticError,
 	};
+
+	pub trait Signum {
+		fn signum(&self) -> i8;
+	}
+
+	macro_rules! signum_variant_impl {
+		($t:ty) => {
+			impl Signum for $t {
+				fn signum(&self) -> i8 {
+					(*self as $t).signum() as i8
+				}
+			}
+		};
+	}
+
+	macro_rules! signum_pos_impl {
+		($t:ty) => {
+			impl Signum for $t {
+				fn signum(&self) -> i8 {
+					1
+				}
+			}
+		};
+	}
+
+	signum_pos_impl!(u8);
+	signum_pos_impl!(u16);
+	signum_pos_impl!(u32);
+	signum_pos_impl!(u64);
+	signum_pos_impl!(u128);
+	signum_variant_impl!(i8);
+	signum_variant_impl!(i16);
+	signum_variant_impl!(i32);
+	signum_variant_impl!(i64);
+	signum_variant_impl!(i128);
+	signum_variant_impl!(f32);
+	signum_variant_impl!(f64);
 
 	/// Performs addition that returns `ArithmeticError::Overflow` instead of
 	/// wrapping around on overflow.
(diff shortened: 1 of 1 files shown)


NunoAlexandre:
Will have a look today 🚀 👀

libs/traits/src/lib.rs
```diff
@@ -776,6 +776,8 @@ pub mod ops {
 		/// assert_eq!(extrinsic(), Err(DispatchError::Arithmetic(ArithmeticError::Overflow)));
 		/// ```
 		fn ensure_mul(&self, v: &Self) -> Result<Self, ArithmeticError> {
+			// TODO: Improve the error checking the case where the multiplication shoul be
```
NunoAlexandre, line 779:
```suggestion
			// TODO: Improve the error checking the case where the multiplication should be
```

NunoAlexandre: Alternatively, you can add an error variant `ArithmeticError::OutOfBounds` which would be an union of Overflow and Underflow, which I guess would be ok in terms of error handling?

libs/traits/src/lib.rs
```diff
@@ -806,4 +808,105 @@ pub mod ops {
...
+		fn ensure_div_assign(&mut self, v: &Self) -> Result<(), ArithmeticError> {
+			*self = self.ensure_div(v)?;
+			Ok(())
+		}
+	}
+
+	impl<T: EnsureAdd> EnsureAddAssign for T {}
+	impl<T: EnsureSub> EnsureSubAssign for T {}
+	impl<T: EnsureMul> EnsureMulAssign for T {}
+	impl<T: EnsureDiv> EnsureDivAssign for T {}
```
NunoAlexandre, line 911: Sweet 👌

NunoAlexandre (approved):
Looks good to me 🎸

libs/traits/src/lib.rs
lemunozm: Thanks for the review Nuno 😃 

The problem is that `ArithmeticError` is something defined by Substrate 😢 . 
I have an idea in mind. I'm going to add a new commit to it soon.

lemunozm: Updated. I think I have covered all the cases.

libs/traits/src/lib.rs
```diff
@@ -804,6 +857,120 @@ pub mod ops {
...
+		///
+		/// fn extrinsic() -> DispatchResult {
+		///     let mut max = u32::MAX;
+		///     max.ensure_add_assign(&1)?;
+		///     Ok(())
+		/// }
+		///
+		/// assert_eq!(extrinsic(), Err(DispatchError::Arithmetic(ArithmeticError::Overflow)));
+		/// ```
+		fn ensure_add_assign(&mut self, v: &Self) -> Result<(), ArithmeticError> {
```
NunoAlexandre, line 881: Given that this is generic and could be used with a signed number type, couldn't this operation also lead to an Underflow error case? For example if the left hand side is the MIN of said type and right hand side is a some non-zero negative number?

NunoAlexandre: Same question applies to the `ensure_sub_assign`

NunoAlexandre (commented):
One question 👀

libs/traits/src/lib.rs
lemunozm: Totally right. I'm going to update it.

NunoAlexandre (approved):
Looks good to me :)

PR #1038, opened 2022-10-15
Ensure Ops extended

lemunozm:
# Description

Improve the `cfg_traits::ops::ensure` module with more functionality.

Follows this work #1034

## Changes and Descriptions

More support for the `ops::ensure` module:

- Added `EnsureFixedPointNumber` trait that adds `ensure_from_rational()`, `ensure_mul_int()` and `ensure_div_int()` for any `FixedPointNumber`.
- Added `EnsureInto` and `EnsureFrom` traits to check the overflow when numbers are converted.
- More accurate division error.
- Improve documentation.

## Type of change

- New feature (non-breaking change which adds *internal* functionality)

# How Has This Been Tested?

```sh
cargo test -p cfg-traits (run test in the examples)
cargo doc -p cfg-traits (check documentation links)
```

# Checklist:

### DIFF
--- a/libs/traits/src/ops.rs
+++ b/libs/traits/src/ops.rs
@@ -86,7 +86,7 @@ pub mod ensure {
 		/// Subtracts two numbers, checking for overflow.
 		/// If overflow happens, `ArithmeticError` is returned.
 		///
-		/// Similar to [`CheckedAdd::checked_sub()`] but returning an `ArithmeticError` error
+		/// Similar to [`CheckedSub::checked_sub()`] but returning an `ArithmeticError` error
 		///
 		/// ```
 		/// use cfg_traits::ops::ensure::EnsureSub;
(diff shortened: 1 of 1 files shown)


libs/traits/src/ops.rs
```diff
@@ -25,273 +22,464 @@ pub trait Signum: PartialOrd + Zero {
...
+		}
+
+		pub fn division<N: Signum, D: Signum>(n: N, d: D) -> ArithmeticError {
+			if d.is_zero() {
+				ArithmeticError::DivisionByZero
+			} else {
+				match n.signum() != d.signum() {
+					true => ArithmeticError::Underflow,
+					false => ArithmeticError::Overflow,
+				}
```
NunoAlexandre, line 481: You could choose to call `multiplication` here to display that `division` for non-zero `d` is the same as calling multiplication. Up to you :)

lemunozm:
> As a reviewer, I am lacking some high-level context of how all these changes are coming together tho.

Maybe I would have needed a feature to track it. My fault! (Maybe is not late 🙄 )
The changes come when I realize that some function that returns `Option<()>` or `Option<Self>` and works with arithmetics could return the `Result` version with `ArithmeticError` instead. The primary use case is to have an alternative that is less verbose and more error accurate that returns directly `Result<(), ArithmeticError>` and eases the arithmetics inside the pallets.

The returned value and parameters come from replicating the `std::ops` version (`Add` => `EnsureAdd`, `AddAsign` => `EnsureAddAsign`) or checked version functions. The idea in mind is to be as ergonomic as possible.

NunoAlexandre (approved):
🪴 good work!

PR #1043, opened 2022-10-19
Add GetFixedPointNumber trait to ops

lemunozm:
# Description

Add a new trait to `cfg_traits::ops` to be able to get the `FixedPointNumber` type of a `FixedPointOperand` operand.

It is useful when you need to know what `FixedPointNumber` you should use for some integer value. For example, it prevents you from passing this value as a `Config` parameter.

## Changes and Descriptions

Added `GetFixedPoint` to `cfg_traits::ops` trait.

## Type of change

# How Has This Been Tested?

```sh
cargo test -p cfg-traits
```

# Checklist:

### DIFF
--- a/libs/traits/src/ops.rs
+++ b/libs/traits/src/ops.rs
@@ -1,4 +1,8 @@
-use sp_runtime::traits::Zero;
+use sp_arithmetic::{
+	fixed_point::{FixedI128, FixedI64, FixedU128, FixedU64},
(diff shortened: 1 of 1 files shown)


lemunozm:
> It is useful when you need to know what FixedPointNumber you should use for some integer value.

Taking another view, I think this precondition is not true. By now I close this.

PR #907, opened 2022-08-12
Tests for tranches.rs

mustermeiszer:
This PR adds necessary test to the tranches module.

Issue #678, opened 2022-03-07
Pallet-pool: Tests for tranche.rs


mustermeiszer:
The new tranche.rs after the PR of #620 gets merged, contains no tests currently. 

Each method and the new functions should be unittested. 
Focus also on testing the reverse lib that we are using (needs to panic upon out-of-index).

thea-leake:
Note about odd issue encountered:

With the `validate_next_tranche`->`Residual`->`Residual` => `false` change several other tests in the Pools pallet are failing with `"InvalidTrancheStructure"` on my end.

It looks like `validate_next_tranche` is sometimes getting `Residual`, `Residual`, when it should be getting `Residual`, `NonRedisual`.

Not sure if actual issue or just on my end at the moment.

Example:
in `pallets/pools/src/tests.rs:948` (execute_info_removed_after_epoch_execute):
```
		assert_ok!(Pools::create(
			pool_owner_origin.clone(),
			pool_owner.clone(),
			0,
			vec![
				TrancheInput {
					tranche_type: TrancheType::Residual,
					seniority: None,
					metadata: TrancheMetadata {
						token_name: BoundedVec::default(),
						token_symbol: BoundedVec::default(),
					}
				},
				TrancheInput {
					tranche_type: TrancheType::NonResidual {
						interest_rate_per_sec: senior_interest_rate,
						min_risk_buffer: Perquintill::from_percent(10),
					},
					seniority: None,
					metadata: TrancheMetadata {
						token_name: BoundedVec::default(),
						token_symbol: BoundedVec::default(),
					}
				},
			],
			CurrencyId::AUSD,
			10_000 * CURRENCY,
			None
		));
```
with the following added to valid_next_tranche
```
		let res = match (self, next) {
			(TrancheType::Residual, TrancheType::Residual) => "Res--Res::False",
			(TrancheType::Residual, TrancheType::NonResidual { .. }) => "Res--NonRes::True",
			(TrancheType::NonResidual { .. }, TrancheType::Residual) => "NonRes--Res::False",
			(TrancheType::NonResidual { .. }, TrancheType::NonResidual { .. }) => "RateComp",
		};
		println!("Comp result: {}", res);
```
we get a match on Res, Res:
```
---- tests::execute_info_removed_after_epoch_execute stdout ----
Comp result: Res--Res::False
thread 'tests::execute_info_removed_after_epoch_execute' panicked at 'Expected Ok(_). Got Err(
    Module(
        ModuleError {
[...]

PR #1046, opened 2022-10-20
Tests/continue tranches tests

thea-leake:
# Description

 Adds tests for tranches in the Pools pallet, continues on work from https://github.com/centrifuge/centrifuge-chain/pull/907

Fixes # https://github.com/centrifuge/centrifuge-chain/issues/678

## Changes and Descriptions

 Adds tests for tranches in the Pools pallet

## Type of change

- Backfill tests and fixes/improvements for discovered issues.

# How Has This Been Tested?

Run unit tests, ticket is for addition of tranches tests.

# Checklist:

### DIFF
--- a/pallets/pool-system/src/tranches.rs
+++ b/pallets/pool-system/src/tranches.rs
@@ -112,6 +112,11 @@ where
 	///
 	pub fn valid_next_tranche(&self, next: &TrancheType<Rate>) -> bool {
 		match (self, next) {
+			// NOTE: If we change this: I.e. allowing multiple residual
+			//       tranches. One MUST be cautious to check ALL places
+			//       were we assume a single residual tranche.
+			//
+			//       E.g. Tranches.non_residual_tranches()
 			(TrancheType::Residual, TrancheType::Residual) => false,
 			(TrancheType::Residual, TrancheType::NonResidual { .. }) => true,
 			(TrancheType::NonResidual { .. }, TrancheType::Residual) => false,
@@ -609,7 +614,7 @@ where
 		F: FnMut(&Tranche<Balance, Rate, Weight, TrancheCurrency>) -> Result<R, DispatchError>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		for tranche in self.tranches.iter().rev() {
+		for tranche in self.non_residual_top_slice() {
 			let r = f(tranche)?;
 			res.push(r)
 		}
@@ -621,7 +626,7 @@ where
 		F: FnMut(&mut Tranche<Balance, Rate, Weight, TrancheCurrency>) -> Result<R, DispatchError>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		for tranche in self.tranches.iter_mut().rev() {
+		for tranche in self.non_residual_top_slice_mut() {
 			let r = f(tranche)?;
 			res.push(r)
 		}
@@ -638,7 +643,7 @@ where
 		I: IntoIterator<Item = W>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		let iter = self.tranches.iter().rev().zip(with.into_iter());
+		let iter = self.non_residual_top_slice().iter().zip(with.into_iter());
 
 		for (tranche, w) in iter {
 			let r = f(tranche, w)?;
@@ -661,7 +666,10 @@ where
 		I: IntoIterator<Item = W>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		let iter = self.tranches.iter_mut().rev().zip(with.into_iter());
+		let iter = self
+			.non_residual_top_slice_mut()
+			.iter_mut()
+			.zip(with.into_iter());
 
 		for (tranche, w) in iter {
 			let r = f(tranche, w)?;
@@ -680,7 +688,7 @@ where
 		F: FnMut(&Tranche<Balance, Rate, Weight, TrancheCurrency>) -> Result<R, DispatchError>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		for tranche in self.tranches.iter() {
+		for tranche in self.residual_top_slice() {
 			let r = f(tranche)?;
 			res.push(r)
 		}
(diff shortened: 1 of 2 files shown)


pallets/pool-system/src/tranches.rs
```diff
@@ -1360,12 +1477,1610 @@ where
...
+						TrancheWeight::from(u128::MAX),
+						TrancheWeight::from(u128::MAX)
+					),
+					(
+						TrancheWeight::from(u128::MAX),
+						TrancheWeight::from(u128::MAX)
+					)
+				]
+			)
+			// Verification of too many tranches unsurprisingly taking too long to run for tests
```
gruberb, line 2943: Is there a benefit of commenting them out vs just removing them? Maybe opening an issue?

gruberb (approved):

pallets/pool-system/src/tranches.rs
thea-leake: Ah, yeah would be better to just remove, I'll remove it.  I'm not sure if it should have an issue, it's testing a pretty extreme edge case though.

pallets/pool-system/src/tranches.rs
```diff
@@ -1133,14 +1190,26 @@ where
...
 		let mut res = Vec::with_capacity(self.tranches.len());
-		let iter = self.tranches.iter().rev().zip(with.into_iter());
-
-		for (tranche, w) in iter {
-			let r = f(tranche, w)?;
-			res.push(r);
+		let mut tranche_slice = self.non_residual_top_slice().iter();
+		let mut with_iter = with.into_iter();
+
+		for _ in 0..self.tranches.len() {
```
wischli, line 1196: Question: Do we need to distinguish between errors based on different lengths of `tranche_slice ` and `with_iter`? If not, we can revert back to zipping which automatically halts when the first iterator stopps. cc @offerijns 

If we care, I would like this custom `zipping` to be extracted to a function to reduce duplicate code as it is called thrice.

wischli (commented):
I have a couple of questions below. Additionally, do we need this PR to be merged before the end of the Polkadot Acadamy @offerijns ? If so, I would love to complete the missing unit tests as this would be great for getting more into the codebase.

pallets/pool-system/src/tranches.rs
```diff
@@ -523,8 +525,75 @@ where
...
+					.valid_next_tranche(&tranche.tranche_type),
+				DispatchError::Other(
+					"Invalid next tranche type. This should be catched somewhere else."
+				)
+			);
+			ensure!(
+				if i_at < self.tranches.len() - 1 {
+					tranche.tranche_type.valid_next_tranche(
+                &self.tranches
+						        .get(i_at + 1)
```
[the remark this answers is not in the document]
mustermeiszer: You are definitely right. We must always check: 

Assuming tranches are stored as: `| Tranche | Tranche | Tranche | ...`

*We should check:*

| Insert At | Checks |
|----------|---------|
| Most left | right &  residual |
| Middle | right & left |
|Most right | left |

*Currently we check:* 
| Insert At | Checks |
|----------|---------|
| Most left |  residual |
| Middle | right & left  |
| Most right | left |

pallets/pool-system/src/tranches.rs
```diff
@@ -523,8 +525,75 @@ where
...
+			at_idx,
+			id.clone(),
+			tranche.tranche_type,
+			tranche.seniority,
+			now,
+		)?;
+
+		self.validate_insert(at, &new_tranche)?;
 		self.remove(at)?;
-		self.add::<MaxTokenNameLength, MaxTokenSymbolLength>(at, tranche, now)
```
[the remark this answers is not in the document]
mustermeiszer: I wouldn't assume that. Although, when I started implementing this, my reasoning was paritally "the pool will check correct inputs", I came to the conclusion that we should implement `struct Tranches` in a self sufficient way. 

Meaning: The struct always keeps and checks core invariants and only allows the pool to have additional checks on top. 

I think duplicating the code here is fine for this reason. If you can refactor it without assuming transactionality, I am happy with that, too.

pallets/pool-system/src/tranches.rs
mustermeiszer: > If we care, I would like this custom zipping to be extracted to a function to reduce duplicate code as it is called thrice.

Yes we care. Would be nice to extract.

mustermeiszer: Comparing length would also be fine.

pallets/pool-system/src/tranches.rs
```diff
@@ -1360,12 +1477,1589 @@ where
...
+	use cfg_primitives::{Balance, PoolId, TrancheId, TrancheWeight};
+	use cfg_types::{fixed_point::Rate, tokens::TrancheCurrency};
 
-	#[test]
-	fn reverse_works_for_both_tranches() {}
+	use super::*;
+
+	// NOTE: We currently expose types in runtime-common. As we do not want
+	//       this dependecy in our pallets, we generate the types manually here.
+	//       Not sure, if we should rather allow dev-dependency to runtime-common.
```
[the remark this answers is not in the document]
mustermeiszer: I think I refactor this already. After I started with the tests. Not 100% sure though...

mustermeiszer (commented):
Just a first look and responding to @wischli.

If @thea-leake is fine with it I think it would be great if you take this to the finish line @wischli.

mustermeiszer:
> I have a couple of questions below. Additionally, do we need this PR to be merged before the end of the Polkadot Acadamy @offerijns ? If so, I would love to complete the missing unit tests as this would be great for getting more into the codebase.

We should. I want to bring in the rebalance changes before that and would build on top of that here in order to not make refactoring this here hell.

pallets/pool-system/src/tranches.rs
wischli: Seems like @thea-leake's approach is optimal here actually!

> Comparing length would also be fine.
Even though comparing input lengths is quicker and more readable, it requires adding the `Clone` trait to the iterator `I` because counting the elements in the iterator consumes it: 

```rustlang
let with_count = with.clone().into_iter().count();

match self.tranches.len().cmp(&with_count) {) {
    Ordering::Less => Err(DispatchError::Other(
        "Iterable contains more elements than EpochExecutionTranches tranche count",
    )),
    Ordering::Greater => Err(DispatchError::Other(
        "EpochExecutionTranches contains more tranches than iterables elements",
    )),
    Ordering::Equal => {
        self.non_residual_top_slice().iter().zip(with.into_iter()).try_fold(Vec::with_capacity(self.tranches.len()), |mut res, (tranche, w)| {
            res.push(f(tranche, w)?);
            Ok(res)
        })
    }
}
```

Extracting the _duplicate_ code also is not really possible due to the different assumptions of the tranche slize (either `IterMut` or `Rev<IterMut<...>>`) and `f` (either `mut` or not).

wischli:
Closed by #1173

PR #1053, opened 2022-10-24
Pallet-Pools: Rename to `PoolSystem`

### LINKED ISSUE #1052: Pallet-pools: Rename to Pools-System
* [x] Refactoring

### Description
We want to make the current pallet-pools a core pallet that is able to verify core invariants of our system, while having as little changes as possible over time. Hence, we want to move out a lot of logic into "surrounding" palelts. 
The name of the pallet should therefore be moved from `pallet-pools` to `pools-system`. The renaming must happen pre launch as the events and storage items will use this prefix

Naming changes. Especially, runtime integrations are critical.

* Migration on Altair  - Need to migrate EVERY storage item IF we rename the `Pools` in the runtime to `PoolsSystem`

mustermeiszer:
# Description
Renames the current `Pallet-Pools` to `Pools-System`

Fixes #1052 

## Changes and Descriptions
* Renaming
* Special look must be taken at the manual generation of the storage prefixes in the migration

## Type of change
The changing in the naming, might be relvant for breaking public apis, depending on how they are build. 

# Checklist:

# Tasks
- [x] Migrate storage on Altair - Sync migration with pallet-investments migration


mustermeiszer:
And did we aggree on `Pools-System` or `Pallet-Pools-System`?

offerijns:
> @branan @offerijns How do you feel about also changing how include pools into the construct_runtime macro?

How complicated would this migration be? I think changing the code on the apps side isn't a huge impact. It seems to me that if we are doing this renaming now, we should do it properly and change it everywhere.

> And did we agree on Pools-System or Pallet-Pools-System?

I don't have a strong opinion on this, will leave it to you and @branan

mustermeiszer:
> How complicated would this migration be? I think changing the code on the apps side isn't a huge impact. It seems to me that if we are doing this renaming now, we should do it properly and change it everywhere.

Not really complicated, but I guess something in the range of 2-3 days of work at least. The best would be to bring this in after #995.

branan:
I think the question of whether to change is really about whether we expect to have a higher-level pallet ever wanting the name "Pools" from an API perspective. If we think we might, we should change it. If we think that's unlikely and we'll be naming things based off of more specific chunks of functionality, then we can just leave it as `Pools` in the runtime and avoid the migration.

If we do change, it should be `Pools-System` or `PoolsSystem` or something like that. We omit the "Pallet" portion from the runtime name pretty much everywhere else (as do Parity and other parachain teams)

branan:
Oh, I see you've already done the rename to `PoolsSystem`. Those comments definitely predated our sync on Tuesday, sorry.

So at this point, is this just waiting on us making a call about closing the current Altair pool vs writing a migration? Is renaming back to "Pools" still on the table if we decide we need to keep what's currently on Altair?

mustermeiszer:
> So at this point, is this just waiting on us making a call about closing the current Altair pool vs writing a migration? Is renaming back to "Pools" still on the table if we decide we need to keep what's currently on Altair?

I already started the migration. But I am open to keeping it. But A clean cut would be nice. I won't work tomorrow, so we would bring this in after or during the offsite. 

The migration is not "too" complicated, but cumbersome. I can push my current WIP state if you wanna take over @branan ?

cc: @offerijns

branan:
I think landing it just after the offsite is fine (or if we want to take some hack time during).

mustermeiszer:
@branan This is ready for review. BUT for some reason the `iter()` in storage map values does not work. At least in the test-externalities. It always returns one... So the tests are not really useful in `try-runtime`. 

I think the logic is anyways correct. And I am not too much into debugging the actual reason, why the iterator does not count the correct amount of items in storage. 

NOTE: I also checked with the normal pool storage items and not just solely with the onces where I create the prefix manually.

mustermeiszer:
A special lock MUST be taken at the manually generated storage prefixes!

runtime/altair/src/lib.rs
```diff
@@ -1409,20 +1409,18 @@ mod upgrade {
 	impl OnRuntimeUpgrade for Upgrade {
```
mikiquantum, line 1409: Are you ready to get this included into the runtime executive? Or is this coming in a different PR?

mikiquantum (commented):
LGTM! 
One note; From a simple definition of "System" -> "a set of things working together as parts of a mechanism or an interconnecting network". Base on this, I think `PoolSystem` is more accurate since `Pools` indicate already plurality and `System` implies that already.

pallets/pools-system/src/migrations/mod.rs
```diff
@@ -89,9 +97,18 @@ pub mod altair {
 		pub reserve: ReserveDetails<Balance>,
 	}
 
-	#[frame_support::storage_alias]
-	type Pool<T: Config> = StorageMap<
-		Pallet<T>,
+	/// The old prefix we used when using pallet-pools as
+	/// `Pools` in the construct_runtime!-macro
+	pub struct PoolPrefix;
```
NunoAlexandre, line 102: I know this is temporary code, but we can name this code more clearly to not rely as much on the comments

```suggestion
	pub struct OldPoolsPrefix;
```

pallets/pools-system/src/migrations/mod.rs
```diff
@@ -89,9 +97,18 @@ pub mod altair {
...
+	impl StorageInstance for PoolPrefix {
+		const STORAGE_PREFIX: &'static str = "Pool";
+
+		fn pallet_prefix() -> &'static str {
+			"Pools"
+		}
+	}
+	type Pool<T> = StorageMap<
```
NunoAlexandre, line 110:
```suggestion

  /// The old Pool storage map
	type OldPoolStorage<T> = StorageMap<
```

pallets/pools-system/src/migrations/mod.rs
```diff
@@ -133,9 +150,18 @@ pub mod altair {
...
+	/// `Pools` in the construct_runtime!-macro
+	pub struct EpochExecutionPrefix;
+	impl StorageInstance for EpochExecutionPrefix {
+		const STORAGE_PREFIX: &'static str = "EpochExecution";
+
+		fn pallet_prefix() -> &'static str {
+			"Pools"
+		}
+	}
+	type EpochExecution<T> = StorageMap<
```
NunoAlexandre, line 163:
```suggestion

	type EpochExecution<T> = StorageMap<
```

pallets/pools-system/src/migrations/mod.rs
```diff
@@ -154,19 +180,33 @@ pub mod altair {
...
-		Pallet<T>,
+	pub struct EpochPrefix;
+	impl StorageInstance for EpochPrefix {
+		const STORAGE_PREFIX: &'static str = "Epoch";
+
+		fn pallet_prefix() -> &'static str {
+			"Pools"
+		}
+	}
+	type Epoch<T> = StorageDoubleMap<
```
NunoAlexandre, line 191:
```suggestion

	type Epoch<T> = StorageDoubleMap<
```

pallets/pools-system/src/migrations/mod.rs
```diff
@@ -154,19 +180,33 @@ pub mod altair {
...
 
-	#[frame_support::storage_alias]
-	pub type Order<T: Config> = StorageDoubleMap<
-		Pallet<T>,
+	pub struct OrderPrefix;
+	impl StorageInstance for OrderPrefix {
+		const STORAGE_PREFIX: &'static str = "Order";
+
+		fn pallet_prefix() -> &'static str {
+			"Pools"
```
NunoAlexandre, line 205: Maybe make `"Pools"` a constant that's returned by all of these `pallet_prefix` functions.

pallets/pools-system/src/migrations/mod.rs
```diff
@@ -196,7 +236,94 @@ pub mod altair {
...
+		}
+	}
+	pub type PoolDeposit<T> =
+		StorageMap<PoolDepositPrefix, Blake2_128Concat, <T as Config>::PoolId, PoolDepositOf<T>>;
+
+	fn migrate_pool_deposit<T: Config>() -> Weight {
+		let mut weight = 0u64;
```
NunoAlexandre, line 282: This weight doesn't need to be `mut`, it can just be instantiated or even used directly. for me there's no need to go and change these everywhere now since this is to be run once.

pallets/pools-system/src/migrations/mod.rs
```diff
@@ -258,27 +373,30 @@ pub mod altair {
 				})
 				.collect::<Vec<_>>();
 
-			Some(PoolDetails {
-				currency,
-				tranches: Tranches {
-					tranches: new_tranches,
-					ids,
-					salt,
+			crate::Pool::<T>::insert(
```
NunoAlexandre, line 376:
```suggestion
      /// Insert the data on the new PoolSystem::Pool storage
			crate::Pool::<T>::insert(
```

NunoAlexandre (dismissed):
Looks good overall, the storage prefixes seem correct. Left some minor suggestions but good to go imo.

I do agree with @mikiquantum tho on naming it `PoolSystem` instead of `PoolsSystem`

mustermeiszer:
> LGTM!
One note; From a simple definition of "System" -> "a set of things working together as parts of a mechanism or an interconnecting network". Base on this, I think PoolSystem is more accurate since Pools indicate already plurality and System implies that already.

cc: @offerijns

offerijns:
> One note; From a simple definition of "System" -> "a set of things working together as parts of a mechanism or an interconnecting network". Base on this, I think PoolSystem is more accurate since Pools indicate already plurality and System implies that already.

I think this makes sense, let's go with `PoolSystem` then

NunoAlexandre:
@mikiquantum @offerijns @mustermeiszer renamed to `PoolSystem` ✅

mikiquantum (approved):
Thanks @NunoAlexandre, the only two pieces I can see missing are:
- Add the migration struct to the Executive in the Altair runtime
- Increment the runtime & node version

Keep in mind that for the development environment we will need to have a hard reset and not a runtime upgrade, which is totally fine, just so we know what type of upgrade is needed next time we want to update the env.

mustermeiszer:
@mikiquantum Why the node?

mikiquantum (approved):
LGTM! 
@mustermeiszer until we agree on having different versioning for the node wrapper and the highest runtime version, this is how we are doing it. 
This is obviously not the best, so let's chat about it today and propose a new versioning policy :)

PR #1055, opened 2022-10-26
Add `taplo` to lint Cargo.toml files

NunoAlexandre:
Just found https://github.com/tamasfe/taplo, a tool for formatting toml files that a few other projects are using to format their Cargo files. Since I've seen a lot of inconsistent formatting in our Cargo files, let's give it a run.

We add a `taplo.fmt` file that, amogst other things, makes taplo order dependencies alphabetically which is much appreciated.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,6 +30,12 @@ jobs:
         with:
           toolchain: ${{ env.RUST_TOOLCHAIN }}
           default: true
+      - name: Install taplo
+        uses: actions-rs/install@v0.1
+        with:
+          crate: taplo-cli
+          version: 0.7.2
+          use-tool-cache: true
       - uses: Swatinem/rust-cache@cb2cf0cc7c5198d3364b9630e2c3d457f160790c
       - name: Run tests
         run: ./ci/script.sh
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -41,6 +41,7 @@ case $TARGET in
 
   lint)
     cargo fmt -- --check
+    taplo fmt --check
     ;;
 
   clippy)
--- a/taplo.toml
+++ b/taplo.toml
@@ -1,0 +1,37 @@
+## https://taplo.tamasfe.dev/configuration/#configuration-file
+
+include = ["**/Cargo.toml"]
+exclude = ["target/**/Cargo.toml"]
+
+[formatting]
+# Align consecutive entries vertically.
+align_entries = false
+# Append trailing commas for multi-line arrays.
+array_trailing_comma = true
+# Expand arrays to multiple lines that exceed the maximum column width.
+array_auto_expand = true
+# Collapse arrays that don't exceed the maximum column width and don't contain comments.
+array_auto_collapse = false
+# Omit white space padding from single-line arrays
+compact_arrays = true
+# Omit white space padding from the start and end of inline tables.
+compact_inline_tables = false
+# Maximum column width in characters, affects array expansion and collapse, this doesn't take whitespace into account.
+# Note that this is not set in stone, and works on a best-effort basis.
+column_width = 160
+# Indent based on tables and arrays of tables and their subtables, subtables out of order are not indented.
+indent_tables = false
+# The substring that is used for indentation, should be tabs or spaces (but technically can be anything).
+indent_string = '  '
+# Alphabetically reorder keys that are not separated by empty lines.
+reorder_keys = false
+# Add trailing newline at the end of the file if not present.
+trailing_newline = true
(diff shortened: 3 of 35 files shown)


NunoAlexandre:
We can choose to add `taplo` to our CI step and dev workflow. Wdyt?

mustermeiszer:
I actually like that idea a lot. @mikiquantum not sure if we must be cautious with respect to how many resources we spend here. But I think keeping out tomls clean will reduce overhead for finding missing feature flags and so a lot! Definitely pro having this!

NunoAlexandre:
> I actually like that idea a lot. @mikiquantum not sure if we must be cautious with respect to how many resources we spend here. But I think keeping out tomls clean will reduce overhead for finding missing feature flags and so a lot! Definitely pro having this!

Thanks for the feedback. I am working on adding this to the CI pipeline as I've given this some thought and it would be really helpful to have these Cargo files well formatted and sorted properly. The resources it takes are quite low I'd say

NunoAlexandre:
It seems that `taplo` doesn't support sorting features' lists, like `std`, `runtime_benchmarks`, etc, which is quite a shame. I was playing with an alternative, https://github.com/DevinR528/cargo-sort, but it doesn't support that either 😢

ci/script.sh
```diff
@@ -41,6 +41,7 @@ case $TARGET in
 
   lint)
     cargo fmt -- --check
+    taplo fmt --check
```
NunoAlexandre, line 44: Adding this as part of the `lint` step since that's what it is. I am planning on opening a follow up PR splitting these lint-related jobs from the `Tests` job into a separate `Lint` job.

.github/workflows/tests.yml
```diff
@@ -30,6 +30,12 @@ jobs:
         with:
           toolchain: ${{ env.RUST_TOOLCHAIN }}
           default: true
+      - name: Install taplo
+        uses: actions-rs/install@v0.1
+        with:
+          crate: taplo-cli
+          version: latest
```
mikiquantum, line 37: Should we lock to a specific version?

mikiquantum (approved):
@mustermeiszer @NunoAlexandre This sounds good, the lint step is already pretty quick and will not add too much overhead to CI. Same for the install step, it should not add too much to the already long >1h runs.

mustermeiszer:
> It seems that taplo doesn't support sorting features' lists, like std, runtime_benchmarks, etc, which is quite a shame. I was playing with an alternative, https://github.com/DevinR528/cargo-sort, but it doesn't support that either 😢

We could in this PR manually sort all features alphabetically, then we would have a clean start, that we could watch on new PRs. WDYT?

.github/workflows/tests.yml
NunoAlexandre: Just pushed that 👍

NunoAlexandre:
> > It seems that taplo doesn't support sorting features' lists, like std, runtime_benchmarks, etc, which is quite a shame. I was playing with an alternative, https://github.com/DevinR528/cargo-sort, but it doesn't support that either 😢
> 
> We could in this PR manually sort all features alphabetically, then we would have a clean start, that we could watch on new PRs. WDYT?
> 
> Clion has a nice sort function for marked stuff, that works well and is fast.

Good idea. We can also do that in a seperate PR and have this one focused on just adding `taplo` to the workflow with the default `taplo fmt` output.

mikiquantum (approved):
👍

mustermeiszer:
Please wait till #995 is merged. I will take care of updating this one here afterwards.

taplo.toml
```diff
@@ -0,0 +1,37 @@
...
+compact_inline_tables = false
+# Maximum column width in characters, affects array expansion and collapse, this doesn't take whitespace into account.
+# Note that this is not set in stone, and works on a best-effort basis.
+column_width = 160
+# Indent based on tables and arrays of tables and their subtables, subtables out of order are not indented.
+indent_tables = false
+# The substring that is used for indentation, should be tabs or spaces (but technically can be anything).
+indent_string = '  '
+# Alphabetically reorder keys that are not separated by empty lines.
+reorder_keys = false
```
mustermeiszer, line 27: Don't we want that? This will be really helpful I guess.

NunoAlexandre: not really. it messes with the whole cargo file, including the `[package]` fields etc. I had a look at other substrate projects using `taplo` and they have it disabled as well. btw this tool is planning on supporting sorting `features` arrays like `std` soon 🥳

mustermeiszer (approved):
LGTM! Thanks for bringing this in!

NunoAlexandre:
> LGTM! Thanks for bringing this in!

thanks for the feedback and reviews 🚀

PR #1057, opened 2022-10-26
Add domain concept to rewards

lemunozm:
# Description

Support different rewards for the same currency, adding a domain concept that splits the reward system.

## Changes and Descriptions

- Currency-related methods of the `rewards` trait now add a `domain_id` parameter.
- `pallet-rewards` and `pallet-liquidity-rewards` are now configured with a `DomainId` type.
 
## Type of change

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

# Checklist:

- [ ] I have added Rust doc comments to structs, enums, traits and functions
- [ ] I have made corresponding changes to the documentation
- [ ] I have performed a self-review of my code
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I rebased on the latest `main` branch

PR #1058, opened 2022-10-26
Support for reward domains as currency prefixes

lemunozm:
# Description

Support different rewards for the same currency, adding a domain concept that splits the reward 
system.

This PR is a simpler implementation of https://github.com/centrifuge/centrifuge-chain/pull/1057 where de `rewards` traits are no modified.

## Changes and Descriptions

- `pallet-rewards` and `pallet-liquidity-rewards` are now configured with a `DomainId` type.

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/pallets/rewards/src/lib.rs
+++ b/pallets/rewards/src/lib.rs
@@ -86,6 +86,14 @@ pub mod pallet {
 		#[pallet::constant]
 		type PalletId: Get<PalletId>;
 
+		/// Type used to identify domains.
+		type DomainId: TypeInfo
+			+ MaxEncodedLen
+			+ codec::FullCodec
+			+ Copy
+			+ PartialEq
+			+ sp_std::fmt::Debug;
+
 		/// Type used to identify currencies.
 		type CurrencyId: AssetId + MaxEncodedLen;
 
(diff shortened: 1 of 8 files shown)


pallets/rewards/src/lib.rs
```diff
@@ -154,7 +162,7 @@ pub mod pallet {
 		Blake2_128Concat,
 		T::AccountId,
 		Blake2_128Concat,
-		T::CurrencyId,
+		(T::DomainId, T::CurrencyId),
```
mikiquantum, line 165: Only "downside" is that when querying storage & iterating over the currencies, clients will need to know that this value is a composed one, since is most likely an enum, it should be easy.

lemunozm: Yes, I do not know how Substrate encodes/decodes a tuple of values, but I think should be easy to identify both things.

lemunozm: We could also use a `StorageNMap` here and avoid that tuple, but not sure if it's worth it.

mikiquantum: No, I think is fine with the tuple.

lemunozm:
I think now we can remove `AllowedCurrencies` from `pallet-liquidity-rewards` because now the pallet enforces the currencies used by it.

mikiquantum (commented):
LGTM! I agree we do not need AllowedCurrencies anymore since the prefix is filled by the correspondent implementer pallet.

PR #1059, opened 2022-10-26
ci: Split `tests` job into `tests` and `lints`

NunoAlexandre:
Built on top of #1055 

We organise the CI jobs a little better to be grouped by what they target. Now the `Tests` job handles both testing and linting so we split those as two different jobs: `Tests` and `Lints`.

The `Lints` jobs now has 3 targets, `fmt` (the former `lint`) which runs `cargo fmt`, `clippy` which runs `cargo clippy`, and (new) `taplo` which runs `taplo fmt`. Having different targets helps with easily identifying which step failed when things go wrong on the CI.

**NOTE**: Wait until #1055 before merging this one.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,44 @@
+on:
```
NunoAlexandre, line 1: @mikiquantum this file is basically a copy of `tests.yml` with only linting related stuff. Is there a way of drying the common pieces between this, `tests.yml`, `build.yml`, etc?

NunoAlexandre:
@mikiquantum @mustermeiszer could you have a look at this one today pls? 🙏

mustermeiszer (commented):
Looks good and sane to me. But I leave it to Miguel to approve. Don't wanna mess with the CI.

And don't we wanna have main as target branch?

.github/workflows/lint.yml
mikiquantum: I think is fine, we could start using reusable workflows but that would most likely mean that we need to refactor a bit the current ones.

mikiquantum (approved):
LGTM!

NunoAlexandre:
@mikiquantum I see on Github that the job "lint" is still waiting while the Lints jobs have completed with success so I wonder if it's because the filename doesn't match the job name. So I pushed a commit changing the file name to `lints.yml` to see if that does the trick.

NunoAlexandre:
@mikiquantum I still see this here on Github. do you have any clue maybe? 🤔 
Screenshot 2022-10-27 at 17 45 35

mikiquantum:
@NunoAlexandre I think is a cache thing on gh side. I just cleared that lints cache and see if that helps. Otherwise I am not sure. Will need to retrigger the job

NunoAlexandre:
thanks @mikiquantum, just pushed a commit.

mikiquantum (approved):

PR #1060, opened 2022-10-27
Minor rewards fixes/improvements

lemunozm:
# Description

Some minor reward fixes/improvements

## Changes and Descriptions

- Generalized error using `TokenError` from Substrate.
- Some code reorganizations to avoid a wrong state if the method fails.

## Type of change

# How Has This Been Tested?

```sh
cargo test -p pallet-rewards
```

# Checklist:

### DIFF
--- a/pallets/liquidity-rewards/src/lib.rs
+++ b/pallets/liquidity-rewards/src/lib.rs
@@ -62,7 +62,7 @@ use sp_runtime::{
 use sp_std::mem;
 use weights::WeightInfo;
 
-/// Type that contains the finish timestamp of an epoch.
+/// Type that contains the timestamp of an epoch.
 #[derive(Encode, Decode, TypeInfo, MaxEncodedLen, RuntimeDebug)]
 #[cfg_attr(test, derive(PartialEq))]
 pub struct EpochTimestamp<BlockNumber>(BlockNumber);
@@ -199,7 +199,8 @@ pub mod pallet {
 	// Although this value could be stored inside `EpochData`,
 	// we maintain it separately to avoid deserializing the whole EpochData struct each `on_initialize()` call.
 	// EpochData could be relatively big if there many groups.
-	// We dont have to deserialize the whole struct 99% of the time assuming a duration of 100 blocks.
+	// We dont have to deserialize the whole struct 99% of the time assuming a duration of 100 blocks,
+	// we only need to perform that action when the epoch finalized, 1% of the time.
 	#[pallet::storage]
 	pub(super) type EndOfEpoch<T: Config> = StorageValue<
 		_,
--- a/pallets/liquidity-rewards/src/lib.rs
+++ b/pallets/liquidity-rewards/src/lib.rs
@@ -62,7 +62,7 @@ use sp_runtime::{
 use sp_std::mem;
 use weights::WeightInfo;
 
-/// Type that contains the finish timestamp of an epoch
+/// Type that contains the finish timestamp of an epoch.
 #[derive(Encode, Decode, TypeInfo, MaxEncodedLen, RuntimeDebug)]
 #[cfg_attr(test, derive(PartialEq))]
 pub struct EpochTimestamp<BlockNumber>(BlockNumber);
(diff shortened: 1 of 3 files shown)


pallets/liquidity-rewards/src/lib.rs
```diff
@@ -62,7 +62,7 @@ use sp_runtime::{
 use sp_std::mem;
 use weights::WeightInfo;
 
-/// Type that contains the finish timestamp of an epoch
+/// Type that contains the finish timestamp of an epoch.
```
NunoAlexandre, line 65: This doc comment doesn't seem be quite what the code shows; the code tells me that `EpochTimestamp` is basically a wrapper around `BlockNumber`, not "the finish timestamp of an epoch" 🤔

pallets/liquidity-rewards/src/lib.rs
```diff
@@ -196,6 +196,10 @@ pub mod pallet {
 	#[pallet::generate_store(pub(super) trait Store)]
 	pub struct Pallet<T>(_);
 
+	// Although this value could be stored inside `EpochData`,
+	// we maintain it separately to avoid deserializing the whole EpochData struct each `on_initialize()` call.
+	// EpochData could be relatively big if there many groups.
+	// We dont have to deserialize the whole struct 99% of the time assuming a duration of 100 blocks.
```
NunoAlexandre, line 202: Very informative comment! I just don't quite get the last sentence. Can you explain?

pallets/rewards/src/lib.rs
```diff
@@ -28,8 +28,10 @@
 //!
 //! ### Terminology
 //!
-//! - **Currency ID**: The identification of a token used to make stake/unstake.
-//!   This ID is associated to a group used to reward the stake amount.
+//! - **Currency ID**: Identification of a token used to make stake/unstake.
```
NunoAlexandre, line 31:
```suggestion
//! - **Currency ID**: Identification of a token used to stake/unstake.
```

pallets/rewards/src/lib.rs
```diff
@@ -268,14 +267,18 @@ pub mod pallet {
 
 				Groups::<T>::try_mutate(group_id, |group| {
 					StakeAccounts::<T>::try_mutate(account_id, currency_id, |account| {
-						T::Currency::hold(currency_id.1, account_id, amount)?;
+						if !T::Currency::can_hold(currency_id.1, account_id, amount) {
```
NunoAlexandre, line 270: `thought`

I'm not a fan of creating type aliases for tuples and then reference the fields like we do here with `currency_id.1`. I'd rather have the current type alias become a struct called `DomainCurrency` that has a `DomainId` and a `CurrencyId`. Then here we could either deconstruct that struct and just reference it as `currency_id` or directly as `domain_currency.currency_id`.

pallets/rewards/src/lib.rs
```diff
@@ -268,14 +267,18 @@ pub mod pallet {
...
+							Err(TokenError::NoFunds)?;
+						}
 
 						account.try_apply_rpt_tallies(currency.rpt_tallies())?;
 						account.add_amount(amount, group.reward_per_token())?;
 
 						group.add_amount(amount)?;
 						currency.add_amount(amount)?;
 
+						T::Currency::hold(currency_id.1, account_id, amount)?;
```
NunoAlexandre, line 280: Maybe leave a comment explaining what `hold` is/does and its application in this context? I guess it's used here to stake `amount`?

NunoAlexandre (commented):
Some thoughts and comments, a step further anyways :) 👍

pallets/liquidity-rewards/src/lib.rs
lemunozm: Michael's Suggestion 😄 

Yes, for each `on_initialize()` call, you need to check the final timestamp of an epoch to check if the epoch has finalized, but only when the epoch finalizes do you need to read its content. If it happens every 100 blocks (for example), we only deserialize it in that case, in the last block of those 100 (1% of the cases).

NunoAlexandre: I would add this explanation to this comment block tbh :)

pallets/rewards/src/lib.rs
lemunozm: > I guess it's used here to stake amount?

Yes! It's the core of this pallet, `hold()` is the same as `Currency::reserve()` but for tokens. This change only moves down that line to the bottom.

lemunozm: Not sure about the comment here, but maybe it is good to explain this on top of the file as general description of how it works

NunoAlexandre: yeah sounds good!

NunoAlexandre (approved):
Nice 🚀

PR #1064, opened 2022-11-04
Ensure streamlined hashing

mustermeiszer:
# Description
Syntactical change that ensures primitives is the location where we define our chain hashing and hash tyoe.

Fixes # (issue)

## Changes and Descriptions
* Simple creation of new types
* Usage of new types in runtimes

# Checklist:

### DIFF
--- a/libs/primitives/src/lib.rs
+++ b/libs/primitives/src/lib.rs
@@ -30,9 +30,8 @@ pub mod types {
 	#[cfg(feature = "std")]
 	use serde::{Deserialize, Serialize};
 	use sp_core::{H160, U256};
-	use sp_runtime::traits;
 	use sp_runtime::{
-		traits::{BlakeTwo256 IdentifyAccount, Verify},
+		traits::{self, BlakeTwo256, IdentifyAccount, Verify},
 		OpaqueExtrinsic,
 	};
 	use sp_std::vec::Vec;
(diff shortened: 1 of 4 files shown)


libs/primitives/src/lib.rs
```diff
@@ -90,15 +90,20 @@ pub mod types {
 	pub type Index = u32;
 
 	/// A hash of some data used by the chain.
-	pub type Hash = sp_core::H256;
+	pub type Hash = <BlakeTwo256 as HashT>::Output;
```
NunoAlexandre, line 93: Do you mind just referencing `sp_runtime::traits::Hash` instead of importing it with an alias?

lemunozm: Following the Nuno's comment, maybe importing `sp_runtime::traits` and then using `traits::Hash`, is still readable without the annoyance of writing the entire path in the type.

lemunozm: matter of taste anyway!

NunoAlexandre: yeah I like that, @lemunozm. I am not fond of import alias, specially when there is no name conflict to resolve.

NunoAlexandre (approved):
🌟

lemunozm (approved):
LGTM!

PR #1069, opened 2022-11-07
Pallet pool registry

gruberb:
# Description

Adding `pallet-pool-registry` to our codebase. The intention is to make the `pallet-pool-system` surface smaller, and abstract away extrinsics etc. into `pallet-pool-registry`. This is the first step towards this goal. 

Fixes https://github.com/centrifuge/centrifuge-chain/issues/956

## Changes and Descriptions

Moving the following extrinsics out:

- `create`
- `update`
- `set_metadata`

## Type of change

# How Has This Been Tested?

Adding additional pallet tests, adjusting runtime tests for `altair` and `development`, as well as `pallet-pool-system`

# Checklist:


libs/traits/src/lib.rs
```diff
@@ -125,6 +128,47 @@ pub trait PoolInspect<AccountId, CurrencyId> {
...
+		tranche_inputs: Vec<Self::TrancheInput>,
+		currency: CurrencyId,
+		max_reserve: Balance,
+		metadata: Option<Vec<u8>>,
+	) -> DispatchResult;
+
+	fn update(
+		pool_id: PoolId,
+		changes: Self::PoolChanges,
+	) -> Result<(UpdateState, PostDispatchInfo), DispatchErrorWithPostInfo>;
```
branan, line 167: This return type scares me, but as long as it's not too awkward to work with it's fine....

pallets/pool-registry/src/benchmarking.rs
```diff
@@ -0,0 +1,48 @@
...
+use frame_benchmarking::{account, benchmarks, impl_benchmark_test_suite};
+use frame_system::RawOrigin;
+use sp_std::vec;
+
+use super::*;
+
+const POOL: u64 = 0;
+
+benchmarks! {
+		where_clause {
```
branan, line 24: Indentation here is inconsistent. Autoindenters don't like the benchmarks macro, I've noticed.

pallets/pool-registry/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+		///
+		/// If no delay is required for updates and redemptions
+		/// don't have to be fulfilled, then this is executed
+		/// immediately. Otherwise, the update is scheduled
+		/// to be executed later.
+		///
+		/// The caller must have the `PoolAdmin` role in order to
+		/// invoke this extrinsic.
+		#[pallet::weight(T::WeightInfo::update_no_execution(T::MaxTranches::get())
+        .max(T::WeightInfo::update_and_execute(T::MaxTranches::get())))]
```
branan, line 284: Indentation here as well

pallets/pool-system/src/benchmarking.rs
```diff
@@ -53,129 +53,117 @@ benchmarks! {
...
-		let n in 0..T::MaxSizeMetadata::get();
-		let caller: T::AccountId = create_admin::<T>(0);
-		let metadata = vec![0u8; n as usize];
-		prepare_asset_registry::<T>();
-		create_pool::<T>(1, caller.clone())?;
-	}: set_metadata(RawOrigin::Signed(caller), POOL, metadata.clone())
-	verify {
-		assert_eq!(get_pool::<T>().metadata, Some(metadata.try_into().unwrap()));
-	}
+	// create {
```
branan, line 56: Remove instead of leaving commented?

runtime/development/src/lib.rs
```diff
@@ -1601,19 +1619,20 @@ construct_runtime!(
...
-		NftSales: pallet_nft_sales::{Pallet, Call, Storage, Event<T>} = 100,
-		Nfts: pallet_nft::{Pallet, Call, Event<T>} = 103,
-		Bridge: pallet_bridge::{Pallet, Call, Storage, Config<T>, Event<T>} = 101,
-		InterestAccrual: pallet_interest_accrual::{Pallet, Storage, Event<T>, Config<T>} = 102,
-		Keystore: pallet_keystore::{Pallet, Call, Storage, Event<T>} = 104,
-		Investments: pallet_investments::{Pallet, Call, Storage, Event<T>} = 105,
-		Rewards: pallet_rewards::{Pallet, Storage, Event<T>} = 106,
-		LiquidityRewards: pallet_liquidity_rewards::{Pallet, Call, Storage, Event<T>} = 107,
-		Connectors: pallet_connectors::{Pallet, Call, Storage, Event<T>} = 108,
+		PoolRegistry: pallet_pool_registry::{Pallet, Call, Storage, Event<T>} = 96,
```
branan, line 1622: Renumbering is technically an API break (although the metadata update should handle it?)

Unless we are having issues with ordering of the pallets, I'd prefer to add the new one at the bottom

runtime/integration-tests/src/pools/pool_starts.rs
```diff
@@ -66,7 +66,7 @@ async fn create_init_and_price() {
 		Event,
 		EventRange::All,
 		Event::System(frame_system::Event::ExtrinsicFailed{..}) if [count 0],
-		Event::PoolSystem(pallet_pool_system::Event::Created { pool_id, .. }) if [pool_id == 0],
+		// Event::PoolSystem(pallet_pool_registry::Event::Created { pool_id, .. }) if [pool_id == 0],
```
branan, line 69: Remove commented out code?

pallets/pool-registry/src/benchmarking.rs
```diff
@@ -0,0 +1,48 @@
...
+//! Module provides benchmarking for the Pools Pallet
+use frame_benchmarking::{account, benchmarks, impl_benchmark_test_suite};
+use frame_system::RawOrigin;
+use sp_std::vec;
+
+use super::*;
+
+const POOL: u64 = 0;
+
+benchmarks! {
```
branan, line 23: This looks like it's still missing most of the benchmarks?

branan (commented):
This mostly looks pretty good. A few blocks of commented out code that should be nuked before we merge. Plus some other questions and comments.

The pallet index thing is my only strong request for change.

I'm curious what necessitated the removal of "Origin::Signed" from test inputs?

mustermeiszer (changes requested):
This looks already good! I made a few comments in the code where I would like to see some changes. 

Besides those I have one bigger request, that I would like to see: 

## Emit event on Pool-System side
At the time, when we execute an update I would like to trigger an event on the system side that signals the essence of changes. Something like: 

```rust
PoolUpdated {
   id: T::PoolId,
   old: PoolEssence,
   new: PoolEssence,
}
```

where `PoolEssence` is a struct that looks something like: 

```rust
struct PoolEssence {
   currency: T::CurrencyId,
   max_reserve: T::Balance, 
   max_nav_age: Moment,
   min_epoch_time: Moment,
   tranches: Vec<TrancheEssence>
}

struct TrancheEssence {
   currency: T::TrancheCurrency,
   type: TrancheType,
   meta: TrancheMetadata
}

// For convenience we should impl
impl From<PoolDetails> for PoolEssence {
   // TODO
}
```

## Emitting a more detailed event WHAT is planned
This is NOT needed right now, as I hope we can add additional events without needing extra code on the subgraph side. cc: @offerijns for clarification. 

But I am referring the idea here for completeness and discussion.
### Option A
We decide we want to split up the updating of a pool into more fine grained independent updates. This is for me a pro as it is more flexible and has smaller, easier to measure and check extriniscs (in the future, now we want to just have the events for this in place). The downside is, that it gets harder to execute, as the pool-system needs to know, that a "common" time-span passed. 

For this purpose we would split the current `PoolChanges` into an enum of fine grained actions: 

```rust
enum PoolChanges {
   MinEpochTime (Moment),
   MaxNavAge(Moment),
   // NOTE: Actually this should be split up into changing 
   //              possibly only meta, or only tranche, or both but not senority. etc. 
   TrancheUpdate {
      id: TrancheId,
       tranche: TrancheUpdate,
      meta: TrancheMetadata,
   }, 
   // NOTE: For the future solely
   TrancheRemoval {...}.
   TrancheInsert {...}
}
```

If we do it this way in the future, we must already now scan the given `PoolChanges` and trigger a distinct event for each change, so that we can adapt to the above enum without needing to change the events again. 

### Option B 
We won't split up changes into fine grained changes, then we must at least still trigger an event of that signals the `PoolChanges`.

libs/traits/src/lib.rs
```diff
@@ -125,6 +128,47 @@ pub trait PoolInspect<AccountId, CurrencyId> {
 	) -> Option<PriceValue<CurrencyId, Self::Rate, Self::Moment>>;
 }
 
+/// Variants for valid Pool updates to send out as events
+#[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+pub enum UpdateState {
+	NoExecution,
```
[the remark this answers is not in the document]
offerijns: I think the purpose of this was to cancel the outstanding scheduled update. So we should then add other functionality for achieving this.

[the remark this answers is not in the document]
mustermeiszer: Yes, canceling should definitely be a dedicated extrinsics instead of implicit submitting all `NoChanges`.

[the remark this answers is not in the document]
mustermeiszer: cc @gruberb

pallets/pool-registry/src/benchmarking.rs
mustermeiszer: I think we can merge this with "fix" weights at the first version. WDYT?

runtime/development/src/lib.rs
mustermeiszer: It will also break decoding past events, errors and calls, as far as I can recall.

pallets/pool-registry/src/benchmarking.rs
gruberb: Good point! I have to read up on benchmarking in Substrate more. Because now we the actual "weight" of the call is in `pool-system`, and `pool-registry` is just passing through the extrinsic call. So where do we weigh such a function? In both pallets?

libs/traits/src/lib.rs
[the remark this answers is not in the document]
gruberb: I didn't change the logic of `pool-system` here: https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/pool-system/src/lib.rs#L842-L846

This is why I added `NoExecution`, because currently we check inside `pool-system` if we just remove the scheduled updated and return `Ok(Some(T::WeightInfo::update_no_execution(0)).into());`.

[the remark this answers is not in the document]
gruberb: So you want me to adjust the `pool-system` logic and error out instead of returning `update_no_execution`?

libs/traits/src/lib.rs
```diff
@@ -125,6 +128,47 @@ pub trait PoolInspect<AccountId, CurrencyId> {
...
+	Executed,
+	Stored,
+}
+
+/// A trait that supports modifications of pools
+pub trait PoolMutate<
+	AccountId,
```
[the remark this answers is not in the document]
gruberb: `UpdateState` can't be an associated type afaik. It's an enum, and if it's associated, I can't have logic like `Self::UpdateState::Executed`. I tried it and we would have to create a different logic for this to work.

libs/types/Cargo.toml
```diff
@@ -24,6 +24,7 @@ sp-runtime = { git = "https://github.com/paritytech/substrate", default-features
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29" }
 
 orml-asset-registry = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.29" }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.29" }
```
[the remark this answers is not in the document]
gruberb: Left-over from some testing!

libs/traits/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: No, I would say lets do this in a follow up pr. Was not aware about the removing logic. This needs more changes then.

libs/traits/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: You can simply enforce the types on the trait. Generics is basically, name all, associated types is, name what you need. 

```rust
type Pool = PoolMutate <Self::AccountId, UpdateState = UpdateState>
```

You can of course define more types. Depedning on what you need here.

gruberb:
@mustermeiszer Based on your design (thanks for that!) I implemented the following:
https://github.com/centrifuge/centrifuge-chain/pull/1069/commits/cfa145d07fad48f5c5beb9f05f6c2530ab8f789b

mustermeiszer:
Commit looks good. Just not picking if the explicit `let clone = pool.clone()` is needed. 

Will take a more detailed look next week! 🙂

gruberb:
@mustermeiszer I hope I am wrong, but this is the only way to get the `TrancheMetadata` needed for our updated `UpdateExecuted` event: https://github.com/centrifuge/centrifuge-chain/pull/1069/commits/1ac60b2ed0d45eacd3e6dd2ace8d9d8242043837

pallets/pool-registry/src/benchmarking.rs
mustermeiszer: Usually, you weight them via the pool-registry. Othwise, you get the weight twice.

mustermeiszer:
> @mustermeiszer I hope I am wrong, but this is the only way to get the TrancheMetadata needed for our updated UpdateExecuted event: https://github.com/centrifuge/centrifuge-chain/commit/1ac60b2ed0d45eacd3e6dd2ace8d9d8242043837

To have a cleaner abstraction can we do: 

```rust
impl PoolDetails<...> {
   pub fn essence<T: Config>(&self) -> Result<PoolEssence, DispatchError> {
         let tranches = self.residual_top_slice(|tranche| {
              let essence = TranchEssence {
                    currency = tranche.currency;
                    tranche_type = tranche.tranche_type;
                    metadata = T::AssetRegistry::metadata(tranche.currency.into())
                        .ok_or(Error::<T>::FailedToUpdateTrancheMetadata)?;
              };

              Ok(essence)
         })?;

        Ok(
           PoolEssence {
               ...
           }
        )
   }
}
```

pallets/pool-system/src/lib.rs
```diff
@@ -255,6 +259,51 @@ impl<CurrencyId, TrancheCurrency, EpochId, Balance, Rate, MetaSize, Weight, Tran
 		self.epoch.last_executed += One::one();
 		Ok(())
 	}
+
+	pub fn essence<T: Config>(&self) -> Result<PoolEssenceOf<T>, DispatchError>
```
mustermeiszer, line 263: For simplicity do: 

`<T: Config<CurrencyId = CurrencyId, Balance = Balance, TrancheCurrency = TrancheCurrency, Rate = Rate>`

pallets/pool-registry/src/benchmarking.rs
```diff
@@ -0,0 +1,160 @@
...
+	where_clause {
+	where
+		T: Config<PoolId = u64>,
+	}
+
+	// create {
```
gruberb, line 29: It's dirty to leave commented code in there, but should we keep it as a reference for the follow up PR to implement benchmarking?

mustermeiszer (commented):
I still need to go over the changes of pool-system, but the rest looks good! 

Think we can bring this in soon as the first draft for the registry. Nice, I am really happy with this!

pallets/pool-system/src/impls.rs
```diff
@@ -70,6 +73,245 @@ impl<T: Config> PoolInspect<T::AccountId, T::CurrencyId> for Pallet<T> {
 	}
 }
 
+impl<T: Config> PoolMutate<T::PoolId> for Pallet<T> {
+	type AccountId = T::AccountId;
```
mustermeiszer, line 77: One minor thing, sorry for the annoyance. But Substrate usually has the `AccountId` as an generic, as it makes no sense to have different `AccountId`s over the runtime. At least my guess, why it is a generic. 

For us, the same applies for the `PoolId`. 

So could you change the trait to take two generics `PoolMutate<AccountId, PoolId>` rather than just one?

pallets/pool-system/src/lib.rs
```diff
@@ -1460,6 +1258,10 @@ pub mod pallet {
 			Pool::<T>::try_mutate(pool_id, |pool| -> DispatchResult {
 				let pool = pool.as_mut().ok_or(Error::<T>::NoSuchPool)?;
 
+				// Prepare PoolEssence struct for sending out UpdateExecuted event
+				let copy_of_pool = pool.clone();
+				let old_pool = copy_of_pool.essence::<T>()?;
```
mustermeiszer, line 1263: copy not needed anymore. `essence` takes `&self`

pallets/pool-system/src/lib.rs
```diff
@@ -1509,9 +1311,18 @@ pub mod pallet {
...
+				// Prepare PoolEssence struct for sending out UpdateExecuted event
+				let copy_of_pool = pool.clone();
+				let new_pool = copy_of_pool.essence::<T>()?;
+
+				Self::deposit_event(Event::PoolUpdated {
+					id: *pool_id,
+					old: old_pool,
+					new: new_pool,
+				});
+
```
mustermeiszer, line 1323: simply use `new: pool.essence::<T>()?`

mustermeiszer (commented):
Three minor things. Otherwise, it is good to go. If those are in, we can merge this from my perspective.

PR #1071, opened 2022-11-08
Drop all stale migrations

NunoAlexandre:
# Description

Drop migration code for a few pallets that should have already been executed.

## Type of change

- Deleting stale code


mikiquantum (commented):
LGTM, one question, there is a migration in the Altair runtime, that could be removed as well right? `pallet_pool_system`. Since this is going live soon.

NunoAlexandre:
@branan can you please have a look as well? We also delete the `pool_system` migrations so it's good to get your approval here

pallets/interest-accrual/src/lib.rs
```diff
@@ -417,16 +415,6 @@ pub mod pallet {
 			})
 		}
 
-		pub fn upgrade_to_v1() -> Weight {
```
mustermeiszer, line 420: I think `remove_unused_rates` was also part of the migration and can be dropped.

mustermeiszer (commented):
LGTM! Thanks for handling. There is one function in interest accrual that I think can be removed, too. Cc: @branan 

And am not approving as I was not involved in anchors.

pallets/interest-accrual/src/lib.rs
NunoAlexandre: It seems unused, dropped it 👍

NunoAlexandre: @branan appreciate your confirmation here

branan: Yup, that was part of the migration (a nasty back-and-forth of data between interest-accrual and loans). Kill it with fire

branan: There is also a `reference_active_rates` in pallet-loans that can go bye-bye now, which was part of the same dance

NunoAlexandre: thanks @branan, done 👍

mustermeiszer (approved):
Approving after everybody went through!

NunoAlexandre:
> Approving after everybody went through!

Thanks @mustermeiszer, a true gentleman 🤘🏻

PR #1072, opened 2022-11-10
Ignore nix build output

NunoAlexandre:
# Description

When running `nix --extra-experimental-features "nix-command flakes" build` locally with success, a `./result/bin/**` directory is created with the output of the nix command.

That requires a manual `rm -r result` clean up to avoid noise from being committed.

## Changes

We `.gitignore` any `bin` directory which both covers us in this case and covers the case of any `bin` build output which is desirable behaviour.


.gitignore
```diff
@@ -1,6 +1,7 @@
 # Generated by Cargo
 # will have compiled files and executables
 **/target/
+**/bin/
```
lemunozm, line 4: Why does adding `bin` to `.gitignore` cover us from not committing the `result` folder?

NunoAlexandre: because `/bin` is the only directory and contents of `result`, so by ignoring `**/bin` we both ignore `result` and any other `bin` that may be outputted somewhere else / by other tool.

lemunozm (approved):
LGTM! 🎉

NunoAlexandre:
Thanks for the review and question, @lemunozm 🏄‍♂️

PR #1074, opened 2022-11-10
Connectors POC v0.0.1 - Implement leftover `TODO`s

NunoAlexandre:
# Description

Implementing the two leftover `TODO`s from the Connectors POC by making use of the `TrancheToken` trait to convert `pool_id` and `tranche_id` into `CurrencyId` and thus be able to plug it with `Tokens::transfer` and `AssetRegistry::metadata`.

## Changes and Descriptions

We do the following:

1. Drop the top-level `CurrencyId` param as just point to the `xcm_transactor::CurrencyId` type to ensure we are not using different types across the board. This is essential so that we can use `TrancheToken` and plug it to other traits successfully.

2. Add `TrancheToken` and `AssetRegistry` type params to the connectors pallet Config.

## Type of change

# How Has This Been Tested?

- [ ] Need to test this on Moonbase
- [x] Write a test verifying the tranche transfer works as expected

# Checklist:

### TODO - To Discuss

Regarding the comment below https://github.com/centrifuge/centrifuge-chain/pull/1074#issuecomment-1311704523, note how the current implementation tries to convert the metadata's name and symbol fields into `[u8; 32]` in the connectors lib.rs. This is **not ok** since it will fail whenever the exact name and symbol lengths, respectively, are not 32-bytes long.

Ideally, the max size of both the name and the symbol fields share the same upper bound on the Connectors pallet as well as on the pool_system pallet. With that, we could then basically convert said fields from the asset_registry by filling it it up to 32 bytes and pad with 0's when their size is less than 32. I haven't found a very idiomatic and comphreensive way of doing this in Rust tho 🤔


NunoAlexandre:
@mikiquantum @offerijns @mustermeiszer this is still in draft because we should extend the tests to test these new features but the core code itself can already be reviewed if you choose to give it a look already 👍

offerijns (commented):
Implementation looks good to me!

NunoAlexandre:
@offerijns I just realised that the `MaxTokenNameLength` and `*Symbol` that we are for the tranche metadata is higher than the limits we set on the connectors' messages:

```
token_name: [u8; 32],
token_symbol: [u8; 32],
```

How do you envision this? Do we just truncate the metadata to fit in the connectors' 32 byte cap?

pallets/connectors/Cargo.toml
```diff
@@ -16,6 +16,7 @@ ethabi = { version = "16.0", default-features = false }
 frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29" }
 frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29" }
 hex = { version = "0.4.3", default_features = false }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.29" }
```
mustermeiszer, line 19: No `std` feature on traits?

pallets/connectors/src/lib.rs
```diff
@@ -157,9 +149,17 @@ pub mod pallet {
 		type Tokens: Mutate<Self::AccountId>
 			+ Inspect<
 				Self::AccountId,
-				AssetId = <Self as pallet::Config>::CurrencyId,
+				AssetId = CurrencyIdOf<Self>,
 				Balance = <Self as pallet::Config>::Balance,
 			> + Transfer<Self::AccountId>;
+
+		type TrancheToken: TrancheToken<PoolIdOf<Self>, TrancheIdOf<Self>, CurrencyIdOf<Self>>;
```
mustermeiszer, line 156: This trait is deprecated. Need to remove it. Sorry for that. Please use `TrancheCurrency`

`type TrancheCurrency: TrancheCurrency<PoolIdOf<Self>, TrancheIOf<Self>> + Into<CurrencyId<Of>;`

pallets/connectors/src/lib.rs
```diff
@@ -253,18 +259,27 @@ pub mod pallet {
 				Error::<T>::TrancheNotFound
 			);
 
-			// Send the message through the router
-			//
-			// TODO: retrieve token name and symbol from asset-registry
-			// Depends on https://github.com/centrifuge/centrifuge-chain/issues/842
-			//
+			// Look up the metadata of the tranche token
+			let currency_id = T::TrancheToken::tranche_token(pool_id.clone(), tranche_id.clone());
```
mustermeiszer, line 263: `let tranche_currency = T::TrancheCurrency::generate(pool_id.clone(), tranche_id.clone())`

pallets/connectors/src/lib.rs
```diff
@@ -253,18 +259,27 @@ pub mod pallet {
...
+			let tranche_metadata = T::AssetRegistry::metadata(&currency_id)
```
mustermeiszer, line 264: `let tranche_metadata = T::AssetRegistry::metadata(&currency_id.into())`

pallets/connectors/src/lib.rs
```diff
@@ -363,17 +378,17 @@ pub mod pallet {
...
-			// 	&DomainLocator {
-			// 		domain: address.domain,
-			// 	}
-			// 	.into_account_truncating(),
-			// 	amount,
-			// 	false,
-			// )?;
+			// Transfer to the domain account for bookkeeping
+			T::Tokens::transfer(
+				T::TrancheToken::tranche_token(pool_id, tranche_id),
```
mustermeiszer, line 383: `T::TrancheCurrency::generate(pool_id, tranche_id).into()`

runtime/development/src/lib.rs
```diff
@@ -1354,14 +1354,15 @@ impl pallet_interest_accrual::Config for Runtime {
...
+	type AssetRegistry = OrmlAssetRegistry;
 	type Balance = Balance;
-	type CurrencyId = CurrencyId;
 	type Event = Event;
 	type Permission = Permissions;
 	type PoolInspect = PoolSystem;
 	type Rate = Rate;
 	type Time = Timestamp;
 	type Tokens = Tokens;
+	type TrancheToken = TrancheToken;
```
mustermeiszer, line 1365: `type TrancheCurrency = TrancheCurrency`

There exists a `TrancheCurrency` struct in the cfg-types that implements all you need already.

mustermeiszer (commented):
Sorry for misleading you. You need to use `TrancheCurrency` trait. I made notes where you should change. But logic seems fine and will not really change with the other trait.

NunoAlexandre:
> Sorry for misleading you. You need to use `TrancheCurrency` trait. I made notes where you should change. But logic seems fine and will not really change with the other trait.

You didn't mislead me :) I tried using TrancheCurrency but its API didn't seem like the best fit while TrancheToken was super straight forward and with a smaller footprint.

We could consider keeping TrancheToken for a case like this one since its interface fits like a glove but I will try and follow your advise on how to integrate with TrancheCurrency and see how it goes 🙏

pallets/connectors/Cargo.toml
NunoAlexandre: It was already in `std` since it was already a `[dev.dependencies]` :)

mustermeiszer (commented):
Everything I see looks fine. Which capabilities has the connectors now?

And one thing regarding the tests:
- add_tranche: does this check some logic on the moonbeam side ? Does it check some state on the moonbeam side?

NunoAlexandre:
> Everything I see looks fine. Which capabilities has the connectors now?

We implement the metadata lookup when sending add_tranche and implement the actual tranche token transfer when sending transfer.

> 
> And one thing regarding the tests:
> 
> * add_tranche: does this check some logic on the moonbeam side ? Does it check some state on the moonbeam side?

mustermeiszer (dismissed):
Nice clean up of the tests!

mustermeiszer:
> Ideally, the max size of both the name and the symbol fields share the same upper bound on the Connectors pallet as well as on the pool_system pallet. With that, we could then basically convert said fields from the asset_registry by filling it it up to 32 bytes and pad with 0's when their size is less than 32. I haven't found a very idiomatic and comphreensive way of doing this in Rust tho 🤔

I think we should agree, (I think you already did agree on 128 bytes, right?) for both connectors and the registry (now still pool-system side). 
For the decoding, I would propose to use an additional byte for the messages to ethereum and always insert a "\end" sign so, that we know how do decode on the ethereum side. Maybe using the `03` byte from the standard https://www.fileformat.info/info/charset/UTF-8/list.htm. We can of course also use the NULL byte. 

TBH I have no idea, how to do this idiomatically, and you definitely code much more idiomatic and clean rust than I do. 
But couldnt' we wrap it with something like: 

```rust
fn populate(data: &[u8]) -> Result<[u8;129], DispatchError> {
     let mut assimilated = [0u8;129];
     match data.len().cmp(128) {
             Ordering::Equal => assimilated[128] = END_OF_TEXT_BYTE;
             Ordering::Less => (),
             Ordering::Greater => return Error;
     }

     let data = data.iter();
     for byte in assimilated.iter_mut() {
           if let Some(data_byte) = data.next() {
                *byte = *data_byte;
           } else {
                *byte = END_OF_TEXT_BYTE;
           }
     }
                      
    Ok(assimilated)
}


```

mustermeiszer:
We actually do not need and `END_OF_TEXT_BYTE` if we fully use the capacity. But it might make the decoding code easier.

NunoAlexandre:
@mustermeiszer let's discuss the metadata sizes subject on this new issue 🏄‍♂️  https://github.com/centrifuge/centrifuge-chain/issues/1080

mustermeiszer (approved):
Approving as I am fine with having the metadata issue solved afterwards.

NunoAlexandre:
@mustermeiszer sounds good, thanks for the reviews 🙌

PR #1077, opened 2022-11-11
cfg-types: Small touch on TokenCurrency + Drop TrancheToken

NunoAlexandre:
# Description

We do two things here:

1. Implement `From` instead of `Into` for `TrancheCurrency` -> `CurrencyId
By implementing `From`, we get `into()` for free, whereas the opposite is not true. Note that this clippy also complains about this.

2. Drop `TrancheToken` that has been deprecated in favour of `TrancheCurrency`

## Type of change

- [x] Small improvement to existing code with no breaking changes
- [x] Drop unused code

### DIFF
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -298,19 +298,6 @@ impl<T> PreConditions<T> for Never {
 	}
 }
 
-/// Trait for converting a pool+tranche ID pair to a CurrencyId
-///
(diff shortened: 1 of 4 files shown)


mustermeiszer (approved):
Thanks man!

NunoAlexandre:
My pleasure, mate :) @mustermeiszer

PR #1084, opened 2022-11-17
Rewards: Avoid first epoch event with zero duration

lemunozm:
# Description

Initially, a not configured epoch has a duration of 0 blocks. Currently, it is generating a NewEpoch event which it should not. This PR allows only to dispatch that event if the epoch has a duration.

## Type of change

# How Has This Been Tested?
```sh
```

# Checklist:

### DIFF
--- a/runtime/development/src/lib.rs
+++ b/runtime/development/src/lib.rs
@@ -1556,7 +1556,7 @@ frame_support::parameter_types! {
 	#[derive(scale_info::TypeInfo, Debug, PartialEq, Clone)]
 	pub const MaxChangesPerEpoch: u32 = 50;
 
-	pub const InitialEpochDuration: BlockNumber = prod_or_fast!(5 * MINUTES, 1 * MINUTES);
+	pub const InitialEpochDuration: BlockNumber = 1 * MINUTES;
 
 	pub const LiquidityDomain: RewardDomain = RewardDomain::Liquidity;
 }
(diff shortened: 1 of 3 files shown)


lemunozm (commented):
I've removed the changes on `on_initialize` because now, you can avoid having zero duration.

Also, as a note aside about the performance. In the previous case, you would fetch the same value from the database for each block, so the system would cache those in memory, and the access should be really fast. I think there are no performance issues on that, but is ok to have this constant definition. My only concern is if someone asks themself why they need to specify the epoch duration for the pallet if later they can change it by an extrinsic (because anyway they need to use an extrinsic later to configure groups and weights).

runtime/development/src/lib.rs
```diff
@@ -1556,6 +1556,8 @@ frame_support::parameter_types! {
 	#[derive(scale_info::TypeInfo, Debug, PartialEq, Clone)]
 	pub const MaxChangesPerEpoch: u32 = 50;
 
+	pub const InitialEpochDuration: BlockNumber = prod_or_fast!(5 * MINUTES, 1 * MINUTES);
```
mikiquantum, line 1559: Since this is the `development-runtime`, I would have a default of 1 * MINUTE only, without the `prod_or_fast` annotation. Since in most cases that is quick enough already to verify anything with a test. In most of the cases the regularly built wasm (no fast-runtime feature) is quick enough on most of the session, epochs etc cycles, so it doesnt need to use the `prod_or_fast`.
The `prod_or_fast` is more useful when testing on production runtimes like altair & centrifuge.

mikiquantum (approved):
LGTM!

mikiquantum (commented):
Sorry I meant to "soft-approve" due to the comment I added.

runtime/development/src/lib.rs
lemunozm: Makes sense!

mikiquantum (approved):
🙏

PR #1085, opened 2022-11-17
feat: Add cfg-utils crate

NunoAlexandre:
We add this crate to place general utils that can be used across other pallets, runtimes, tests, etc.

To bootstrap it, we added and test a `vec_to_fixed_array` function that is used in https://github.com/centrifuge/centrifuge-chain/pull/1083

### DIFF
--- a/libs/utils/Cargo.toml
+++ b/libs/utils/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+authors = ["Centrifuge <admin@centrifuge.io>"]
+description = 'Utility functions for Centrifuge'
+edition = '2021'
+license = "LGPL-3.0"
+name = 'cfg-utils'
+repository = "https://github.com/centrifuge/centrifuge-chain"
+version = '2.0.0'
+
+[package.metadata.docs.rs]
+targets = ['x86_64-unknown-linux-gnu']
+
+[dependencies]
+codec = { package = 'parity-scale-codec', version = '3.0.0', features = ['derive'], default-features = false }
+frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29" }
+scale-info = { version = "2.0", default-features = false, features = ["derive"] }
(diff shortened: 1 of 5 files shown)


mustermeiszer (dismissed):
LGTM! I Wonder if we should also have a math lib for all the calculations we are doing in pools and loans? 

Cc: @branan @thea-leake @gruberb

libs/utils/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+authors = ["Centrifuge <admin@centrifuge.io>"]
+description = 'Utility functions for Centrifuge'
+edition = '2021'
+license = "LGPL-3.0"
+name = 'cfg-utils'
+repository = "https://github.com/centrifuge/centrifuge-chain"
```
lemunozm, line 7: In some crates of our workspace, the link continues with something like `/libs/utils`, but not sure what should be the correct one

lemunozm (approved):
Looks perfect to me! 🚀

gruberb:
> LGTM! I Wonder if we should also have a math lib for all the calculations we are doing in pools and loans? 
> 
> Cc: @branan @thea-leake @gruberb 

Yeah, this also came to my mind during my WIP PR about the pool types. Depends if this counts as invariant though!

libs/utils/Cargo.toml
NunoAlexandre: Yeah I noticed that when I copy-pasted this cargo file. Those links are all 404 and I don't see the need to make them super specific to the sub-crate, just pointing to the repo itself is fine. I fixed that here: https://github.com/centrifuge/centrifuge-chain/pull/1086

mustermeiszer:
> > LGTM! I Wonder if we should also have a math lib for all the calculations we are doing in pools and loans? 
> > 
> > Cc: @branan @thea-leake @gruberb 
> 
> Yeah, this also came to my mind during my WIP PR about the pool types. Depends if this counts as invariant though!

I would say, basic computations shouldn't be. We also have fixed point arithmetic separate. But rebalancing definitely is System specific. But let's say interest calculation and so on maybe not.

PR #1086, opened 2022-11-17
cargo.toml: Fix and standardize [package] entries

NunoAlexandre:
We fix the [package] data for the pool-system pallet and standardize all the 'repository' entries to point to
github.com/centrifuge/centrifuge-chain instead of suffixing the base url with the name of the pallet which currently results in a 404 and adds excessive and unnecessary noise to that entry.


mustermeiszer (approved):
Thanks man! We are slowly tidying everting up.

NunoAlexandre:
> Thanks man! We are slowly tidying everting up.

One step at a time 😊  🧹 thanks!

PR #1087, opened 2022-11-18
Rewards: deferred mechanism with movement support

### LINKED ISSUE #1078: Rewards: currency movement for the deferred mechanism

### Description

Add currency movement support for the deferred mechanism.
This implies:
  - Design a method that allows moving currencies without breaking the deferred mechanism invariants.
  - Implement a mechanism with this behavior in `pallet-rewards`
  - Testing

### Research/based on
- See [these slides](https://docs.google.com/presentation/d/1pfl5cXYY2QnV01S90XQnrKp1JlOzw6aDTBT2nyeAU7Q/edit#slide=id.g19117097588_0_0) to implement the deferred currency movement.

lemunozm:
# Description

This PR implements the last mechanism required by `pallet-rewards` on https://github.com/centrifuge/roadmap/issues/163

Fixes #1078

## Changes and Descriptions

- Add a new mechanism for supporting currency movement in the deferred mechanism.
- Refactor all mechanisms to avoid one depending on others. Because this is very mathematical, trying to reuse code ends up writing more code and obfuscating the maths, which should be the most readable thing to consider for a reader of this part.
- Add new tests and refactor the old ones.
- Remove the UTs of each mechanism. Those UTs only were testing the machine behavior using their own behavior without a use case. This implies that if I write a bug, I would probably test against that bug, expected my own bug, and therefore not testing anything.

## Type of change

# How Has This Been Tested?

```sh
cargo test -p pallet-rewards
```

# Checklist:


mikiquantum (commented):
What is the disadvantage of combining `base` & `base_with_currency_movement` into one mechanism? Is it only the extra value stored in the Account and the additional move extrinsic? 
As a user, that will not want to move currencies, they can still not move any and use it without it, right? I think having less mechanisms, until we have more use cases makes sense. WDYT?

lemunozm:
> What is the disadvantage of combining base & base_with_currency_movement into one mechanism?

Currently, I see two advantages to having them as different configurable options that we would lose if we combine them (or simply if we remove the `base` one):
- Less storage per account, as you say, but also per currency.
- If someone doesn't want to use currency movement, the `base` would be easier to audit without the movement code noise. This is even more significant in the pair `deferred` & `deferred_with_currency_movement` because the movement there is more complex.

What are the advantages you see in combining them? I see this separation as something that comes free with no effort.

PR #1088, opened 2022-11-18
Cleanup pool types

gruberb:
# Description

Moving out types from `pool-registry` and `pool-system` into `lib/types/src/...`. We then can use the types throughout the codebase through `cfg_types::...`.

`tree libs/types`:

```
├── Cargo.toml
└── src
    ├── adjustments.rs
    ├── epoch.rs
    ├── fee_keys.rs
    ├── fixed_point.rs
    ├── fulfillments.rs
    ├── ids.rs
    ├── investments.rs
    ├── lib.rs
    ├── orders.rs
    ├── permissions.rs
    ├── pools.rs
    ├── reserves.rs
    ├── time.rs
    ├── tokens.rs
    ├── tranches.rs
    └── xcm.rs
```

For example, instead of:
`use cfg_types::CurrencyId`

we now do:
`use cfg_types::tokens::CurrencyId`.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

All current tests run through.

# Checklist:


mustermeiszer:
> However, I can't seem to build it with that feature flag. I removed it for now to validate the tests, and have to figure out now how we can, in a test scenario, pass the `test` flag to `cfg-types` and build it with the two additionals `impl` blocks, which are just used for testing.

You must create a feature flag like `std` and pass it to the dependency. Like `with_test_impls` and then we can hide them that way.

gruberb:
> You must create a feature flag like `std` and pass it to the dependency. Like `with_test_impls` and then we can hide them that way.

Wouldn't this alter our `test` command
`cargo test --workspace --release --features test-benchmarks,try-runtime,runtime-benchmarks`? 

I would need one of `test` or the feature flags being passed down and add this flag on top of the `impl` block.

libs/types/src/consts.rs
```diff
@@ -7,7 +7,7 @@
 // (at your option) any later version (see http://www.gnu.org/licenses).
 // Centrifuge is distributed in the hope that it will be useful,
 // but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See t
```
NunoAlexandre, line 10: Needs reverting 🦀

gruberb:
@mustermeiszer I solved it like that: 
https://github.com/centrifuge/centrifuge-chain/pull/1088/files#diff-e2c484318086774d9f6babc810eb35062e86fc411e6648eb3abb4e87ff2e616aR56

libs/test-utils/src/mocks/mod.rs
```diff
@@ -14,6 +14,7 @@
 //! Also does contain implementations for Substrate based traits for testing
 
 pub mod accountant;
+pub mod authority_origin;
```
mustermeiszer, line 17: 🥳

libs/types/src/fulfillments.rs
```diff
@@ -0,0 +1,23 @@
...
+use frame_support::RuntimeDebug;
+use scale_info::TypeInfo;
+use sp_runtime::Perquintill;
+use sp_std::cmp::PartialEq;
+
+#[derive(Copy, Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+pub struct FulfillmentWithPrice<BalanceRatio> {
+	pub of_amount: Perquintill,
+	pub price: BalanceRatio,
+}
```
mustermeiszer, line 23: All of this can go into orders. But no hard requirement.

libs/types/src/pools.rs
```diff
@@ -0,0 +1,227 @@
...
+		self.reserve.available = Zero::zero();
+
+		Ok(())
+	}
+
+	pub fn execute_previous_epoch(&mut self) -> DispatchResult {
+		self.reserve.available = self.reserve.total;
+		self.epoch.last_executed += One::one();
+		Ok(())
+	}
```
mustermeiszer, line 170: Can you create a ticket for us to keep track that we need to check, how we can ensure, that pools maintains all invariants?

libs/types/src/reserves.rs
```diff
@@ -0,0 +1,66 @@
...
+		self.total = self
+			.total
+			.checked_add(&acc_investments)
+			.ok_or(ArithmeticError::Overflow)?
+			.checked_sub(&acc_redemptions)
+			.ok_or(ArithmeticError::Underflow)?;
+
+		Ok(())
+	}
+}
```
mustermeiszer, line 66: Also applies to invariant maybe. I will make a comment on all sections that the issue must track.

libs/types/src/tranches.rs
```diff
@@ -10,75 +10,58 @@
...
+/// relevant. Most notably, in the `struct Tranche`.
+#[derive(
+	Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen,
+)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct TrancheCurrency {
+	pub(crate) pool_id: PoolId,
+	pub(crate) tranche_id: TrancheId,
+}
+
```
mustermeiszer, line 64: This should go into tokens IMO. But leave this up for discussion ^^

libs/types/src/tranches.rs
```diff
@@ -1354,7 +1360,7 @@ where
 	}
 }
 
-pub(crate) fn calculate_risk_buffers<Balance, BalanceRatio>(
+pub fn calculate_risk_buffers<Balance, BalanceRatio>(
```
mustermeiszer, line 1363: This is NOT the location, but I can not make comment on unchanged sections. 

Sections to keep track for invariant: 
* rebalance_tranche
* calculate_price

pallets/pool-system/src/solution.rs
```diff
@@ -386,13 +400,13 @@ where
...
-	pub invest_fulfillment: Perquintill,
-	pub redeem_fulfillment: Perquintill,
-}
+// // The solution struct for a specific tranche
+// #[derive(Encode, Decode, Copy, Clone, Eq, PartialEq, Default, RuntimeDebug, TypeInfo)]
+// #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+// pub struct TrancheSolution {
+// 	pub invest_fulfillment: Perquintill,
+// 	pub redeem_fulfillment: Perquintill,
+// }
```
mustermeiszer, line 409: RM

mustermeiszer (commented):
I like the changes. There is one section where code is commented and needs to be removed. 

Otherwise, the only thing currently of main importance to me is that we create a tracking issue to restore the pool-system to maintain all invariants. 

From my side I will approve, once the issue is out and the commented code is removed.

lemunozm:
Sorry, I came late to `test` the feature conversation. The `test` feature is not exported when you compile a crate, so you can not make something as utilities for testing in a crate that will be used by another crate. What I do (not sure if there are better ways, but seems that Substrate uses the same approach) is to implement it under the `std` feature that is always available when `test` is used. Also, because `std` is not used in production, you have the guarantee that this code can not be used wrong for real cases.

gruberb:
> Sorry, I came late to `test` the feature conversation. The `test` feature is not exported when you compile a crate, so you can not make something as utilities for testing in a crate that will be used by another crate. What I do (not sure if there are better ways, but seems that Substrate uses the same approach) is to implement it under the `std` feature that is always available when `test` is used. Also, because `std` is not used in production, you have the guarantee that this code can not be used wrong for real cases.

Nice, thank you for the insights! I tried it like that (inside the `std` feature flag) and it worked:
https://github.com/centrifuge/centrifuge-chain/pull/1088/commits/4400dcdb8c0417a524b83f4b1bec4f64ef135fce#diff-e2c484318086774d9f6babc810eb35062e86fc411e6648eb3abb4e87ff2e616aL54-L56

libs/types/src/tranches.rs
gruberb: https://github.com/centrifuge/centrifuge-chain/issues/1091

libs/types/src/reserves.rs
gruberb: https://github.com/centrifuge/centrifuge-chain/issues/1091

libs/types/src/pools.rs
gruberb: https://github.com/centrifuge/centrifuge-chain/issues/1091

lemunozm (commented):
I like it! LGTM, but I let the approval to Frederik that knows more about this part.

mustermeiszer:
> Also, because std is not used in production, you have the guarantee that this code can not be used wrong for real cases.

If I am not mistaken, this is not entirely correct. Collators can decide if they wanna use the std environment for building blocks, also std is used for producing the genesis block. 

I think this is not a problem for us, but wanted to mention it.

mustermeiszer (approved):
LGTM!

Issue #1056, opened 2022-10-26
CFG rewards improvements


lemunozm:
Issue to track possible improvements (not mandatories) on [CFG rewards](https://centrifuge.hackmd.io/9T9zAZaTRD6ZcHN-Gd_lzw?view#pallet-deferred-rewards-spec):

- [ ] **Configurable minting method**: by @mustermeiszer in https://github.com/centrifuge/centrifuge-chain/pull/944#discussion_r999456369 and [slack conversation](https://centrifugehq.slack.com/archives/C04221Z1NDA/p1666166081074319)
- [x] **Support for differents `pallet-rewards` in same runtime**. Could be cases where we want different rewards behavior for the same currencies.
    - ~~**Solution 1.** Several instances of `pallet-rewards`.~~
    - ~~**Solution 2.** Domain concept: https://github.com/centrifuge/centrifuge-chain/pull/1057~~
    - **Solution 3.** Currency domain prefix: https://github.com/centrifuge/centrifuge-chain/pull/1058
- [ ] **CurrencyId used for rewards specified in the Group**. Currently, we expect to only reward with CFG, but could be good to have the possibility to specify multiple rewards depending on the groups.
- [ ] **Struct for the tuple (DomainId/CurrencyId)**: _Originally posted by @NunoAlexandre in https://github.com/centrifuge/centrifuge-chain/pull/1060#discussion_r1006506405_
- [x] ~~**Variable used for counting currency moves can be parameterized by the user**. i.e, if `MaxCurrencyMovement` is configured to `100`, there is no need to store an `u32` per account to track the movements. It will be enough with an `u8`. Implementation here: https://github.com/centrifuge/centrifuge-chain/pull/1090~~
- [ ] **Different accounts for minting depending on the domain**: _Originally posted by @mikiquantum in https://github.com/centrifuge/centrifuge-chain/pull/1079#discussion_r1024667888_
- [x] **No duplication code in mechanisms**.

PR #1090, opened 2022-11-22
Rewards: Generic currency movement counter with support for no counters

lemunozm:
# Description

The type used to track the currency movements per account is fixed by the implementation (`u16` wight now). Ideally, we would want to allow the user to choose the type they want:

- If the user doesn't want to move currency, they should be able to use a zero overhead type as `()` to save storage.
- If the user wants to perform more than `u16` movements, they should be able to use `u32` as a counter type, for example.

This PR generalizes the counter type to allow this.

Related to https://github.com/centrifuge/centrifuge-chain/issues/1056

## Changes and Descriptions

- `last_currency_movement` is now generic.
- The type `MaxCurrencyMovements` now implements a new type that allows defining both the max currency movement value and the type used to hold that value: `MaxCurrencyMovement<u8, 50>`

## Type of change

- Refactorization

# How Has This Been Tested?

```sh
cargo test -p pallet-rewards
cargo test -p development-runtime
```

# Checklist:

### DIFF
--- a/pallets/rewards/src/mechanism.rs
+++ b/pallets/rewards/src/mechanism.rs
@@ -5,40 +5,13 @@ use sp_runtime::{traits::Get, ArithmeticError};
 pub mod base;
 pub mod deferred;
 
-pub trait DistributionId: Sized {
-	fn next_id(&mut self) -> Result<Self, ArithmeticError>;
-}
-
-impl DistributionId for () {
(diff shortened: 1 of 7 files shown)


lemunozm:
Closed because the complexity it adds could not deserve the storage it saves

PR #1102, opened 2022-11-29
Pallet-Restricted-XTokens

mustermeiszer:
# Description
This PR adds a pallet that enables running a "check" before forwarding to the logic of orml-xtokens. This pallet is needed in order to ensure the restriction of transfers in the future, not just for chain internal transfers as we currently already do with the pallet-restricted-transfers but also for xcm based transfers. 

## Changes and Descriptions
* Adds a new pallet
* Pallet mirrors API of orml-xtokens 1-to-1
* External integration will not recognize a difference between this pallet and orml-xtokens

## Type of change

# Tasks
- [ ] Adapt tests to have a failing version - i.e. implement a check that will return `DispatchError` instead of `Ok(())`
- [ ] Add Pallet as `XTokens` to the runtimes 
- [ ] Rename orml-pallet from `XTokens` to `OrmlXTokens` in the runtimes
- [ ] Remove exposure of the `orml-xtokens::Call` in the runtimes
- [ ] Add benchmarks


mustermeiszer:
A note regarding the benchmarks: 

Orml tokens currently does not use a normal benchmarking but rather has special calculation methods implement on the pallet. We could ask them make those public and re-use them. 

Problem: We do not account for the weight of the checking logic, which will be non negliable in the real runtime setup.

We could also try to write normal benchmarks. TBH, I am not sure, why the orml-xtokens does not use them, but they must have a reason...

gruberb:
I am happy to take this if no-one else wants it :)

mustermeiszer:
@NunoAlexandre Can you show we the filter? I did not find the proper location I guess. 

### Regarding what this will enable
We will implement a `TransferAllowlist` pallet where users can set a restriction for their sending account. Something like, account `x` is allowed to send `currency_a` to `MultiLocation_b` and account `y`. We check these lists when users send tokens via `pallet-restricted-tokens` and via `pallet-restricted-xtokens`. 

The permissions check is an additional check that only is of importance for `TrancheCurrency` tokens. 

### Regarding the transfer logic in connectors
I am not sure if we need the user-exposed transfer. I would instead force users to always go through the "official" transfer pallets. 
* Makes integration with third-party wallets easier
* One less thing to worry about restrictions
Connectors themselves definitely need the transfer to be possible internally via a connector's message. But I would rather have the message handling being it's own extrinsic (`Connectors::handle_incoming_message()`) on the connectors pallet that is only allowed to be called by our parachain origin. 

Let's talk about this in our following connectors sync.

mustermeiszer:
cc: @offerijns

NunoAlexandre:
> @NunoAlexandre Can you show we the filter? 

Actually, initially we had a call filter but we then switched to just rely on the fact that the `CurrencyIdConverter` of the xcm config does not support Tranche currencies. See https://github.com/centrifuge/centrifuge-chain/blob/main/runtime/centrifuge/src/xcm.rs#L209 and https://github.com/centrifuge/centrifuge-chain/blob/main/runtime/centrifuge/src/xcm.rs#L243-L253 👍

mustermeiszer:
done in different PR

PR #1106, opened 2022-12-01
test bench

mikiquantum:
# Description

Testing on comment gh action for benchmarks

# Checklist:

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -5,7 +5,7 @@ on:
 jobs:
   benchmark:
     name: benchmark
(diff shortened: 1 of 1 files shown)


mustermeiszer:
@mikiquantum What does this enable? Can we automatically generate benchmarks with this command? And who will be able to call it?

mikiquantum:
@mustermeiszer this is just a fix over the workflow so I can test if it works fine. The idea is that anyone with Member role of the org can run this by adding a comment as shown above, then the logic will run this in a large runner and push the weights as an artifact associated to the run, so the result can be pulled and pushed again. In the next iteration, most likely I would like to make this automatic by having a gh bot that pushes it directly to the PR, but this requires write permissions and I would like to think this a bit more.

NunoAlexandre (dismissed):

mustermeiszer:
Where are the results pushed to? I mean the artifact?

PR #1114, opened 2022-12-02
Testing Benchmark Job - Add status comments

mikiquantum:
## Description
Since there is no visible status for the `issue_comment` events, just added a comment at the beginning and at the end of the run for user feedback.

Here is an example of a successful run https://github.com/centrifuge/centrifuge-chain/actions/runs/3597875895 
You can find the zipped artifacts a the bottom of that page. There is no support to link the final artifact anywhere since the artifact_id is not available at the gh action level.


NunoAlexandre (approved):

gpmayorga:
@mikiquantum  The artifact_id is not known, but the workflow ID is and thus you can get your artifact URL using gh CLI or curl and the following method:
https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#list-workflow-run-artifacts

mikiquantum:
@gpmayorga thanks, I will look at that now.

PR #1115, opened 2022-12-02
Update polkadot v0.9.32

thea-leake:
# Description

Upgrades Polkado/Substrate/Cumulus to v0.9.32

## Type of change

- [x] Substrate Upgrade

# Checklist:

# Remaining Tasks:
- [x] Fix Fudge instantiation on Pools Integration Tests + Re-Enable tests
- [x] Include upstream migrations


mikiquantum:
@NunoAlexandre @thea-leake FYI, the benchmark action is wrongfully configured to pull the correct branch. Since by default this action only executes on main, then I pushed in this PR the correspondent fix to pull the PR branch instead. This means that we cannot call successfully the bench command until this is merged.

mikiquantum (approved):

branan (dismissed):
This looks good to me, assuming my eyes didn't glaze over too much on the repetitive changes :)

I'd like Frederik to look over the integration test env stuff, but nothing jumped out as evil to me (well, no more evil than that file normally is :see_no_evil:)

I've got a couple questions on runtime changes:
* Is the proof size of u64::MAX correct for maximum weight? I'd have thought this would have to come from the relay chain somehow? Do we just set it big and rely on our blocks being rejected if we're too big for the relay chain? Is this still unused at this point in the weights migration?
* How did we come up with the allowed transfer types for unvested funds?

And one thing to keep in mind is that we need to make sure that when we re-benchmark, we fix all the configs to point to the in-runtime weights again. It's quite a few of them, so I'm not going to flag them with comments.

mustermeiszer:
> I'd like Frederik to look over the integration test env stuff, but nothing jumped out as evil to me (well, no more evil than that file normally is 🙈)

Feel free to refactor that...

mustermeiszer:
> How did we come up with the allowed transfer types for unvested funds?

What do you mean with that?

runtime/altair/src/lib.rs
```diff
@@ -222,12 +224,6 @@ impl Contains<Call> for BaseCallFilter {
 				| pallet_xcm::Call::force_subscribe_version_notify { .. }
 				| pallet_xcm::Call::force_unsubscribe_version_notify { .. } => true,
 			},
-			Call::Multisig(method) => match method {
-				pallet_multisig::Call::as_multi { call, .. } => {
-					call.encoded_len() < MAX_MULTISIG_CALL_SIZE
-				}
-				_ => true,
-			},
```
mustermeiszer, line 230: Why don't we check that anymore? Was this just a weight reason for the PoV?

mustermeiszer (approved):
LGTM!

thea-leake:
@branan I looked into the proof size again and I had missed some context when I had set the set_proof_size param, I'm looking into that now.

ci/script.sh
```diff
@@ -32,7 +32,7 @@ case $TARGET in
     ;;
 
   tests)
-    RUST_MIN_STACK=8388608 cargo test --workspace --release --features test-benchmarks,try-runtime
+    RUST_MIN_STACK=8388608 cargo test --workspace --release --features test-benchmarks,try-runtime --exclude runtime-integration-tests
```
NunoAlexandre, line 35: Do we want this? If we still have issues with the pools tests, let's just disable those and not the whole integration-tests suite 🙏

src/service.rs
```diff
@@ -403,9 +402,6 @@ where
 			relay_chain_interface,
 			relay_chain_slot_duration,
 			import_queue,
-			collator_options: CollatorOptions {
```
NunoAlexandre, line 406: the new API doesn't take these collator options pointing to the relay chain and I am testing spinning off a local environment and I am unable to onboard the parachain (for now). Will update if this persists

ci/script.sh
mikiquantum: Since we are running the integration tests with its own command `./scripts.sh integration` and those all pass, it seems to me that the `./scripts.sh tests` should only focus on the unit tests then and we can disable the integration test from that run, since they are covered in the other one. 
If we still want to run the integration tests from the unit tests run we need to fix the context setup of the `./scripts.sh tests`, we can do that in a different PR. We are not losing coverage here.

mikiquantum:
@branan 
> How did we come up with the allowed transfer types for unvested funds?

They are the same standard types from Polkadot Runtime, they seem to make sense for now.

> And one thing to keep in mind is that we need to make sure that when we re-benchmark, we fix all the configs to point to the in-runtime weights again. It's quite a few of them, so I'm not going to flag them with comments.

Yep, this will be addressed in subsequent PRs

mikiquantum (approved):

mikiquantum (approved):

PR #1119, opened 2022-12-05
Rewards: added gap mechanism

lemunozm:
# Description

The deferred mechanism was an issue distributing wrong the lost reward in the same epoch (although it still works well if the lost reward is distributed to the next epoch). 

The `gap-mechanism` allows to distribute the correct lost reward in the same epoch. This new mechanism works slightly differently. Instead of deferring de reward one epoch later, it postpones the stake/unstake participant actions one epoch later, allowing to cancel them easily if the system state change. The participant needs to leave the stake there during the gap between two distributions to get the reward.

## Current reward state:

image

## Changes and Descriptions

- [x] Refactor `deferred-mechanism` to use its own storage.
- [x] Added `gap-mechanism`
- [x] Move some storage from `pallet-rewards` to the `gap-mechanism` & `deferred-mechanism`
- [x] Removed `DistributionId` trait.

## Type of change

- This change requires a documentation update

# How Has This Been Tested?

```sh
cargo test -p pallet-rewards
cargo test -p development-runtime
```

**NOTE:** *2 tests still failing*

# Checklist:


pallets/rewards/src/mechanism/gap.rs
```diff
@@ -0,0 +1,377 @@
...
+		}
+
+		fn move_currency(
+			currency: &mut Self::Currency,
+			prev_group: &mut Self::Group,
+			next_group: &mut Self::Group,
+		) -> Result<(), MoveCurrencyError> {
+			if prev_group.pending_total_stake > T::Balance::zero() {
+				Err(DispatchError::from(
+					Error::<T>::TryMovementAfterPendingState,
```
mikiquantum, line 341: This means that the on_initialize will always have to call move_currency after rewarding the group right? Which makes sense.

pallets/rewards/src/mechanism/gap.rs
```diff
@@ -0,0 +1,377 @@
...
+			currency: &mut Self::Currency,
+			group: &mut Self::Group,
+			amount: Self::Balance,
+		) -> Result<(), DispatchError> {
+			account.update(group, currency)?;
+
+			account.pending_stake.ensure_add_assign(amount)?;
+			group.pending_total_stake.ensure_add_assign(amount)?;
+
+			currency.total_stake.ensure_add_assign(amount)?;
```
mikiquantum, line 271: Why do we need to keep track of the total_stake at currency level? Is it so later we know how much to move out of the total pot?

pallets/rewards/src/mechanism/gap.rs
```diff
@@ -0,0 +1,377 @@
...
+			account.update(group, currency)?;
+
+			let pending_amount = amount.min(account.pending_stake);
+
+			account.pending_stake.ensure_sub_assign(pending_amount)?;
+			group
+				.pending_total_stake
+				.ensure_sub_assign(pending_amount)?;
+
+			let computed_amount = amount.ensure_sub(pending_amount)?;
```
mikiquantum, line 291: Maybe remaining_amount? Since it is what is left after subtracting from pending.
```suggestion
			let remaining_amount = amount.ensure_sub(pending_amount)?;
```

pallets/rewards/src/mechanism/gap.rs
```diff
@@ -0,0 +1,377 @@
...
+			account
+				.reward_tally
+				.ensure_add_assign(reward.ensure_into()?)?;
+
+			Ok(reward)
+		}
+
+		fn move_currency(
+			currency: &mut Self::Currency,
+			prev_group: &mut Self::Group,
```
mikiquantum, line 336:
```suggestion
			from_group: &mut Self::Group,
```

pallets/rewards/src/mechanism/gap.rs
```diff
@@ -0,0 +1,377 @@
...
+			next_group: &mut Self::Group,
```
mikiquantum, line 337:
```suggestion
			to_group: &mut Self::Group,
```

pallets/rewards/src/mechanism/gap.rs
```diff
@@ -0,0 +1,377 @@
...
+				Err(DispatchError::from(
+					Error::<T>::TryMovementAfterPendingState,
+				))?;
+			}
+
+			let rpt_change = next_group.rpt.ensure_sub(prev_group.rpt)?;
+
+			currency
+				.rpt_changes
+				.try_push(rpt_change)
```
mikiquantum, line 349: It is possible that the `rpt_change` is negative. Is that Arithmetically handled fine by the consumer of `currency.rpt_changes`?

pallets/rewards/src/mechanism/gap.rs
```diff
@@ -0,0 +1,377 @@
...
+				group.rpt.ensure_add_assign(rate)?;
+				reward = amount;
+			}
+
+			group
+				.total_stake
+				.ensure_add_assign(group.pending_total_stake)?;
+			group.pending_total_stake = T::Balance::zero();
+
+			RptHistory::<T>::insert(group.distribution_id, group.rpt);
```
mikiquantum, line 249: Shouldn't `group.rpt` added to the RptHistory refer to the previous rpt and not the one we just computed?

pallets/rewards/src/mechanism/gap.rs
```diff
@@ -0,0 +1,377 @@
...
+			last_currency_movement: 0,
+		}
+	}
+}
+
+impl<T: Config> Account<T> {
+	fn was_distribution(&self, group: &Group<T>, currency: &Currency<T>) -> bool {
+		if self.last_currency_movement as usize == currency.rpt_changes.len() {
+			self.distribution_id != group.distribution_id
+		} else {
```
mikiquantum, line 61: Is the purpose of this else branch to handle when a currency has been moved? Maybe adding some comments around why are we doing what would be useful.

pallets/rewards/src/mechanism/gap.rs
```diff
@@ -0,0 +1,377 @@
...
+		self.distribution_id = group.distribution_id;
+
+		Ok(())
+	}
+
+	fn get_tally_from_rpt_changes(
+		&self,
+		rpt_changes: &[T::Rate],
+	) -> Result<T::IBalance, ArithmeticError> {
+		let rpt_to_apply = &rpt_changes[self.last_currency_movement as usize..]
```
mikiquantum, line 121: In the worst case the call will iterate the value stored in MaxMovements, right? Do you know what value we have set this to?

mikiquantum (commented):
Overall looks good! Just a few comments and clarifications. 
I didnt review the deferred file since I think we are going to drop it.

pallets/rewards/src/mechanism/gap.rs
lemunozm: Correct 👍🏻

pallets/rewards/src/mechanism/gap.rs
lemunozm: Exactly, we need to know how much stake we need to subtract from the total stake of the current group and add it to the new group when we move.

Because it's mandatory not to have pending stake before moving, we know that when we will move, that pending stake added will no longer be pending.

pallets/rewards/src/mechanism/gap.rs
lemunozm: I call it computed, because it is the part of the amount that will be subtracted from the already computed stake for reward of the account. Not sure if "remaining" indicates that, maybe too generic?

If computed is not descriptive enough, maybe something like `rewarded_amount` fits better?

pallets/rewards/src/mechanism/gap.rs
lemunozm: I like it more 👍🏻

pallets/rewards/src/mechanism/gap.rs
lemunozm: Good question! Yes, it is possible if you move to a group with less rpt. It implies that the `Fixed` type used for `Rate` should be a signed type to handle this. The guard `<T::Rate as FixedPointNumber>::Inner: Signed` in the `RewardMechanism` impl protects this from a wrong configuration, with a `FixedU64` for example.

pallets/rewards/src/mechanism/gap.rs
lemunozm: Yes, and since we are postponing all previous stake/unstakes actions, the current rpt is actually computed with the previous state, so it's "the previous" rpt.

pallets/rewards/src/mechanism/gap.rs
lemunozm: With the restriction of only moving after distributing, I think I can refactor this to leave the condition simpler.

pallets/rewards/src/mechanism/gap.rs
lemunozm: Correct. It's set at configuration. Currently, in the development environment, It's set to 50.

pallets/rewards/src/mechanism/gap.rs
lemunozm: Done

mikiquantum (dismissed):

mikiquantum (dismissed):

mikiquantum (approved):

PR #1120, opened 2022-12-06
Use cargo patch to avoid duplicated crates

NunoAlexandre:
Fixes https://github.com/centrifuge/centrifuge-chain/discussions/1118

We introduce `cargo patch` rules to:

- avoid duplicated crates for paritytech/{substrate|polkadot|cumulus}, which will be helpful in runtime upgrades and in - together with nix - promoting reproducible builds.

- allow us to depend on Moonbeam's official repo directly instead of our own fork of their repo by patching their forked `{substrate|polkadot|cumulus}`-based crates to the official `paritytech/{substrate|polkadot|cumulus}` counterpart.


mikiquantum (approved):
Awesome, looks good to me. 
One question, when we have to upgrade from 0.9.29 to a later version of substrate/polkadot/cumulus, in addition to a sed replace on the polkadot branch version we will have to replace the rev version on each patch section right?

PR #1122, opened 2022-12-06
Removes `without_storage_info` macro from pallets

### LINKED ISSUE #1121: Remove `without_storage_info`
* [x] Refactoring

### Description
In order to be able to migrate to frame weights v2 we need to get rid of the `#[pallet::without_storage_info]` macro that currently is used at all our pallets. 

### Research/based on
* See: https://github.com/paritytech/substrate/issues/8629

This will enable us to upgrade to frame weights v2 which we need to be able to generate correct PoV blocks and submit them to polkadot. 

* Adding `MaxEncodedLen` to ALL storage items we use
* Should NOT need a migration

mustermeiszer:
# Description
In order to migrate to frame weights v2 we need to get rid of the `without_storage_info` macro on our pallets. In order to do this, we must implement `MaxEncodedLen` on ALL storage items. This PR accomplishes that goal.

As a result, once weights v2 are supported, the PoV size can be estimated by using the max encoded lengths (:= MEL) of changed storage items. Luckily, **this conversion does not require a migration** as long as `BoundedVec` max length is below the current length of its counterpart vector. This is the case for us!

Closes #1121 

## Changes and Descriptions
* Remove macro on all pallets
* Adapt storage types to implement `MaxEncodedLen` - mostly deriving it automatically
  * Two manual implementations were necessary due to missing MEL impls for three imported enums: [pool_system::pool_types::PoolChanges](https://github.com/centrifuge/centrifuge-chain/pull/1122/files#diff-2b408e35ea65dd77693af57564549eaf775b8362b71545746aef98b3dbf1cba4R169) and [connectors::routers::XcmDomain](https://github.com/centrifuge/centrifuge-chain/pull/1122/files#diff-ac76c92d600955f4c91edc72245817c0fc935f3fbbb44d647d89181619914724R36)
  * For `PoolChanges`, I [opened an ORML upstream PR](https://github.com/open-web3-stack/open-runtime-module-library/pull/867). In the future, we can remove that custom impl then.
* Convert all `Vec`s in any storage to their `BoundedVec` variants.
  * Turns out that bounding their length by `MaxTranches` seems sufficient for all new `BoundedVec` 

## Type of change

# Checklist:

~~- [ ] I have made corresponding changes to the documentation~~
- [x] My changes generate no new warnings (AFAICT)
~~- [ ] I have added tests that prove my fix is effective or that my feature works~~
~~ - [ ] Any dependent changes have been merged and published in downstream modules~~


mustermeiszer:
Just started this, but adding max encode len derivation is more cumbersome and time consuming than I thought. Leaving this here for the future. Me or others ^^

wischli:
> Just started this, but adding max encode len derivation is more cumbersome and time consuming than I thought. Leaving this here for the future. Me or others ^^

I was thinking about picking this up as it affects all pallets which helps me to get an overview. I also did this last year elsewhere. Definitely cumbersome 🫠

wischli:
@mustermeiszer Not sure if you want to take a quick look. Can't request a review as its your PR :) Apart from two custom impls (see PR description), mainly migrating Vecs.

mustermeiszer (commented):
Looks all straight forward. Thanks for picking this up.

I still need to take a look tomorrow at the custom impl in the connectors router. 

Looks good, but am on the phone and can not look at the deps.

mustermeiszer:
@wischli i will only have time for this tomorrow. So if you wanna merge this up today, rather ping someone else additionally.

wischli:
> @wischli i will only have time for this tomorrow. So if you wanna merge this up today, rather ping someone else additionally. 

No worries, it can wait until EOW for sure. I imagine you have bigger fish to fry.

pallets/connectors/src/routers.rs
```diff
@@ -22,9 +22,32 @@ pub struct XcmDomain<CurrencyId> {
...
+	fn max_encoded_len() -> usize {
+		// custom MEL bound for `VersionedMultiLocation`
+		xcm::v1::MultiLocation::max_encoded_len()
+			// VersionedMultiLocation is enum with two variants
+			.saturating_add(2)
+			.saturating_add(BoundedVec::<
+				u8,
+				ConstU32<{ xcm_primitives::MAX_ETHEREUM_XCM_INPUT_SIZE }>,
+			>::max_encoded_len())
+			// custom MEL bound for CurrencyId
```
NunoAlexandre, line 49: I'd rather not abbreviate this as MEL in the code, we won't know what this means in a few weeks time

```suggestion
			// The fee currency
```

pallets/connectors/src/routers.rs
```diff
@@ -22,9 +22,32 @@ pub struct XcmDomain<CurrencyId> {
...
+			.saturating_add(cfg_types::tokens::CurrencyId::max_encoded_len())
+			.saturating_add(H160::max_encoded_len())
```
NunoAlexandre, line 51: It would be great to simplify and extend these comments to inform which field of `XcmDomain` they are _add_ing to the `max_encoded_len`:

```suggestion
      // Contract address
			.saturating_add(H160::max_encoded_len())
			// The fee currency 
			.saturating_add(cfg_types::tokens::CurrencyId::max_encoded_len())

```

Moreover, these two types are in the inverse order now, the suggestion left here sorts them. It sums to the same but it's helpful to keep it in order

pallets/connectors/src/routers.rs
```diff
@@ -22,9 +22,32 @@ pub struct XcmDomain<CurrencyId> {
...
+
+// NOTE: Remove this custom implementation once the following underlying data implements MaxEncodedLen:
+/// * Polkadot Repo: xcm::VersionedMultiLocation
+/// * PureStake Repo: pallet_xcm_transactor::Config<Self = T>::CurrencyId
+impl<CurrencyId> MaxEncodedLen for XcmDomain<CurrencyId>
+where
+	XcmDomain<CurrencyId>: Encode,
+{
+	fn max_encoded_len() -> usize {
+		// custom MEL bound for `VersionedMultiLocation`
```
NunoAlexandre, line 41:
```suggestion
		// The domain's `VersionedMultiLocation`
```

NunoAlexandre (commented):
Looks good overall but two mains points:

1. Let's not abbreviate `MaxEncodedLength` in the code comments as `MEL` since it's not a thing and thee's no need to introduce that abbreviation

2. I don't think the custom impl for `XcmDomain` is correct because it just sums `2` as the max_encoded_length of the `VersionedMultiLocation` because of it having two variants (`V1` and `V2`) when this value can hold an arbitrarily long value.

pallets/connectors/src/routers.rs
```diff
@@ -22,9 +22,34 @@ pub struct XcmDomain<CurrencyId> {
...
+			.saturating_add(2)
+			// The ethereum xcm call index (default bound)
+			.saturating_add(BoundedVec::<
+				u8,
+				ConstU32<{ xcm_primitives::MAX_ETHEREUM_XCM_INPUT_SIZE }>,
+			>::max_encoded_len())
+			// The contract address (default bound)
+			.saturating_add(H160::max_encoded_len())
+			// The fee currency (custom bound)
+			.saturating_add(cfg_types::tokens::CurrencyId::max_encoded_len())
```
mustermeiszer, line 53: TBH, I have no clue how the `MaxEncodedLen` is derived. For an enum I would guess, assuming an enum has less then 256 variants, it would be 1 bytes + size of largest enum variant? 

As the differences between `v0` and `v1` are basically 
* `v1::MultiLocation` being a struct holding a `u8` and essentially `v0::MultiLocation`
* `v1::Junction` removing the named variant (`Parent`) and all other items being equal

I would say the implementation is correct. WDYT @NunoAlexandre?

The only thing I don't understand is why we need to add 2, wouldn't it be 1 as in 1 byte for indicating the variant number in the codec?

mustermeiszer (commented):
Re: Regarding custom impl of `MaxEncodedLen`. Need feedback from @wischli and @NunoAlexandre here.

pallets/connectors/src/routers.rs
NunoAlexandre: Maybe I don't understand `MaxEncodedLength` but I thought its point was to capture the max length that a value of a specific type could take; for example for the  If that is correct, than a `VersionedMultiLocation` will not be at most 2 bytes, it can in fact be arbitrarily long and that's probably why `MaxEncodedLength` is not implemented for that type.

[Help me understand](https://www.youtube.com/watch?v=tGu0_rFwfBc) 😆 🙏

wischli: > The only thing I don't understand is why we need to add 2, wouldn't it be 1 as in 1 byte for indicating the variant number in the codec?

Yes, you are right. I mixed that up with something else. Validated this with a small example:

```
#[derive(codec::Encode, codec::MaxEncodedLen)]
pub enum TestEnum<T: MaxEncodedLen> {
	A(T),
	B(T),
        C(T),
        D(T),
        E(T),
}

#[test]
assert_eq!(TestEnum::<u32>::max_encoded_len(), u32::max_encoded_len() + 1);
```

> Maybe I don't understand `MaxEncodedLength` but I thought its point was to capture the max length that a value of a specific type could take; If that is correct, then a `VersionedMultiLocation` will not be at most 2 bytes, it can in fact be arbitrarily long and that's probably why `MaxEncodedLength` is not implemented for that type.


`MaxEncodingLen` aims to provide an estimate to the maximum encoded size of an arbitrary storage. So even though a `u128` could have length 1, its [max encoded length would be 17](https://github.com/paritytech/parity-scale-codec/blob/master/src/max_encoded_len.rs#L75). Some more context can be found [in the initial PR which introduced it](https://github.com/paritytech/substrate/issues/8719). 

Unfortunately, [`VersionedMultiLocation`](https://github.com/paritytech/polkadot/blob/53253de77f1649ecbb650df2ec3e29d0d1110a2d/xcm/src/lib.rs#L73) does not impl `MaxEncodedLen`.

```
pub enum VersionedMultiLocation {
	V0(v0::MultiLocation),
	V1(v1::MultiLocation),
}
```

Luckily, `v1::MultiLocation` does! Since [v1 ](https://github.com/paritytech/polkadot/blob/53253de77f1649ecbb650df2ec3e29d0d1110a2d/xcm/src/v1/multilocation.rs#L51) is [v0](https://github.com/paritytech/polkadot/blob/53253de77f1649ecbb650df2ec3e29d0d1110a2d/xcm/src/v0/multi_location.rs#L46) with an extra field as pointed out above, v1 should be fine for the `MaxEncodedLen` plus the encoding size of an enum which is one.

 Therefore, it should be correct to use `v1::Multilocation::max_encoding_len().saturating_add(1)` for `VersionedMultiLocation::max_encoded_len()`.

NunoAlexandre: Oh I see, now I get it 🙏 thanks! Looks good then :)

NunoAlexandre (approved):
Sweet 🍯 solid work and thanks for the clarifications

Issue #1063, opened 2022-10-28
Rewards: Runtime API for list of currencies and claimable reward


lemunozm:
### Which part of the code is the issue addressing?

* Feature

### Description

- Allow the possibility to get the list of currencies an account has staked by Runtime API. 
- Allow the possibility to get the claimable reward amount for a list of currencies by Runtime API.

### Research/based on

Discussion thread here: by @mikiquantum in https://github.com/centrifuge/centrifuge-chain/pull/944#discussion_r998800249

mikiquantum:
Fixed by #1124

PR #1124, opened 2022-12-08
1063 add rewards runtime api

cdamian:
# Description

Fixes # (issue)

https://github.com/centrifuge/centrifuge-chain/issues/1063

## Changes and Descriptions

Added runtime API for:

- listing reward currencies of account
- computing reward for an account

## Type of change

# How Has This Been Tested?

- Adjusted the current tests that we have in the rewards pallet to confirm that listing the currencies behaves correctly.
- Tested API functionality using the newly added integration tests for runtime APIs.

# Checklist:


scripts/init.sh
```diff
@@ -64,7 +64,7 @@ start-parachain)
     --ws-external \
     --rpc-methods=Unsafe \
     --state-cache-size 0 \
-    --log="main,debug" \
+    --log="main,info" \
```
mikiquantum, line 67: We do not need debug here as default, too noisy.

libs/traits/src/rewards.rs
```diff
@@ -121,14 +121,14 @@ pub trait AccountRewards<AccountId> {
 	/// Type used as balance for all currencies and reward.
 	type Balance;
 
-	/// Type used to identify the currency
-	type CurrencyId;
```
NunoAlexandre, line 125: No intention to be nitpicking but given that this code is within the context of `rewards.rs`, I don't see the need to prefix this with type with `Reward`

runtime/development/src/lib.rs
```diff
@@ -1914,6 +1917,24 @@ impl_runtime_apis! {
...
+			Some(currencies)
+		}
+
+		fn compute_reward(reward_currency_id: (RewardDomain, CurrencyId), account_id: AccountId) -> Option<Balance> {
+			let reward = <pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance1> as AccountRewards<AccountId>>::compute_reward(reward_currency_id, &account_id);
+
+			match reward {
+				Ok(balance) => Some(balance),
+				Err(_) => None,
+			}
```
NunoAlexandre, line 1934:
```suggestion
		<pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance1> as AccountRewards<AccountId>>::compute_reward(reward_currency_id, &account_id).ok()
```

You can call `.ok()` on a `Result` to convert it to `Option` 👍

runtime/development/src/lib.rs
```diff
@@ -1914,6 +1917,24 @@ impl_runtime_apis! {
 		}
 	}
 
+	// RewardsApi
+	impl runtime_common::apis::RewardsApi<Block, AccountId, Balance, RewardDomain, CurrencyId> for Runtime {
+		fn list_currencies(account_id: AccountId) -> Option<Vec<(RewardDomain, CurrencyId)>> {
+			let currencies = pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance1>::list_currencies(account_id);
+
+			Some(currencies)
```
NunoAlexandre, line 1925: It seems like we have no error handling happening here so we always return `Some(x)`. Why not just make the return type `Vec(RewardDomain, CurrencyId)`?

NunoAlexandre (commented):
Looks good overall 👌  a few comments only

runtime/development/src/lib.rs
cdamian: Will change, thank you!

runtime/development/src/lib.rs
cdamian: I did that to keep it consistent with the rest of the APIs where we always return options.

libs/traits/src/rewards.rs
cdamian: But, to be fair, I also noticed that we do the same in other places, with `Balance` for example. So I can revert this naming change and then probably add a `CurrencyOf<T>` instead of `(T::DomainId, T::CurrencyId)`?

NunoAlexandre: Gotcha! That makes sense, let's keep it then 👍 thanks

runtime/development/src/lib.rs
NunoAlexandre: Yeah ok that's fine 👍

NunoAlexandre (dismissed):
LGTM 💯

libs/traits/src/rewards.rs
lemunozm: I come here a little bit late to clarify the `CurrencyId` naming. From the rewards traits perspective, what you put in the `currency_id` is a `CurrencyId`. Nevertheless, the `pallet-rewards` who implements this, add some domain as a prefix or namespace to that `CurrencyId`. This is why in the pallet the `currency_id` is `(T::DomainId, T::CurrencyId)`.

cdamian: It makes more sense if viewed from the perspective of the rewards trait. I reverted that naming change.

pallets/rewards/src/lib.rs
```diff
@@ -427,4 +427,17 @@ pub mod pallet {
...
+	impl<T: Config<I>, I: 'static> Pallet<T, I>
+	where
+		RewardAccountOf<T, I>: FullCodec + Default,
+	{
+		pub fn list_currencies(
+			account_id: T::AccountId,
+		) -> sp_std::vec::Vec<(T::DomainId, T::CurrencyId)> {
+			StakeAccounts::<T, I>::iter_prefix(account_id)
+				.map(|(currency_id, _)| currency_id)
+				.collect::<sp_std::vec::Vec<(T::DomainId, T::CurrencyId)>>()
```
lemunozm, line 440: As a hint, I think this could be simple:
```suggestion
				.collect()
```

pallets/rewards/src/tests/currency_movement.rs
```diff
@@ -47,6 +56,15 @@ macro_rules! currency_movement_tests {
 				}
 			}
 
+			fn has_currencies(user: u64, currency_ids: &[(DomainId, CurrencyId)]) {
```
lemunozm, line 59: I think rust allows to check the equality of two slices directly. Something like:
```rust
assert_eq!(&$pallet::list_currencies(user), currency_ids);
```

pallets/rewards/src/lib.rs
```diff
@@ -427,4 +427,17 @@ pub mod pallet {
...
 		}
 	}
+
+	impl<T: Config<I>, I: 'static> Pallet<T, I>
+	where
+		RewardAccountOf<T, I>: FullCodec + Default,
+	{
+		pub fn list_currencies(
+			account_id: T::AccountId,
+		) -> sp_std::vec::Vec<(T::DomainId, T::CurrencyId)> {
```
lemunozm, line 437: Maybe we can avoid the `Vec` allocation returning directly the iterator. Something like: `impl Iterator<Item = (T::DomainId, T::CurrencyId)>` in the result type.

lemunozm (changes requested):
LGTM! 😃  Only few comments

pallets/rewards/src/lib.rs
cdamian: Done, thank you!

pallets/rewards/src/tests/currency_movement.rs
cdamian: Dropped that ugly func and used that assert instead.

pallets/rewards/src/lib.rs
cdamian: Not sure if this will complicate things a bit when implementing it in the runtime APIs and if this optimization is worth it.

lemunozm: Its ok for me leave the `Vec` if the things get complicated dealing the Iterator trait 👍🏻

pallets/rewards/src/tests/currency_movement.rs
```diff
@@ -83,7 +74,10 @@ macro_rules! currency_movement_tests {
 					assert_ok!($pallet::attach_currency(DOM_1_CURRENCY_M, GROUP_B)); // MOVEMENT HERE!!
 					assert_ok!($pallet::deposit_stake(DOM_1_CURRENCY_M, &USER_A, STAKE_M));
 
-					has_currencies(USER_A, &[(DomainId::D1, CurrencyId::M)]);
+					assert_eq!(
```
lemunozm, line 77: The currency movement does not affect the number of currencies in an account, so maybe is better to move this check to its own test in `rewards/src/tests/common.rs` under `currency_common_tests` macro. To leave these tests free of this.

That test should associate different currencies to an amount and check the correct list of currencies in that amount.

lemunozm (changes requested):
Sorry for this late comment! 🙏🏻

pallets/rewards/src/tests/currency_movement.rs
lemunozm: Thank You!

libs/traits/src/rewards.rs
```diff
@@ -162,10 +162,10 @@ pub trait AccountRewards<AccountId> {
 
 /// Support for change currencies among groups.
 pub trait CurrencyGroupChange {
-	/// Type used to identify the group
+	/// Type used to identify the group.
 	type GroupId;
 
-	/// Type used to identify the currency
+	/// Type used to identify the reward currency.
```
lemunozm, line 168:
```suggestion
	/// Type used to identify the currency.
```

libs/traits/src/rewards.rs
```diff
@@ -121,7 +121,7 @@ pub trait AccountRewards<AccountId> {
 	/// Type used as balance for all currencies and reward.
 	type Balance;
 
-	/// Type used to identify the currency
+	/// Type used to identify the reward currency.
```
lemunozm, line 124:
```suggestion
	/// Type used to identify the currency.
```

pallets/rewards/src/tests/currency_movement.rs
```diff
@@ -468,6 +468,49 @@ macro_rules! currency_movement_tests {
...
+					];
+
+					assert_ok!($pallet::attach_currency(DOM_1_CURRENCY_A, GROUP_A));
+					assert_ok!($pallet::attach_currency(DOM_1_CURRENCY_B, GROUP_A));
+					assert_ok!($pallet::attach_currency(DOM_1_CURRENCY_C, GROUP_A));
+					assert_ok!($pallet::attach_currency(DOM_1_CURRENCY_M, GROUP_A));
+
+					assert_ok!($pallet::deposit_stake(DOM_1_CURRENCY_A, &USER_A, STAKE_A));
+					assert!(expected_currency_ids[0..1]
+						.to_vec()
```
lemunozm, line 489: I think there is no need to transform expected_currency_ids into a `Vec`. The slice is already iterable.

lemunozm (dismissed):
LGTM! I leave two minor comments, but for me it is approved. Very clean PR @Cosmin!

pallets/rewards/src/tests/currency_movement.rs
lemunozm: Anyway, I like how the test looks 😃

cdamian: Changed, thank you!

runtime/development/src/lib.rs
cdamian: OK, I went ahead and dropped the option. Thank you both for the input on this!

lemunozm (dismissed):

mikiquantum (dismissed):
👍

cdamian:
Marked as draft to avoid an accidental merge before the update to `polkadot-0.9.32` is done.

I will rebase this branch as well.

mikiquantum (approved):
Awesome on the integration tests for runtime APIs!

PR #1125, opened 2022-12-09
Add registry + weights to altair

gruberb:
# Description

So far we crated `pallet_pool_registry`, but haven't added it yet to the `altair` runtime. This PR fixes that.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

`cargo test --workspace --release --features test-benchmarks,try-runtime,runtime-benchmarks`

# Checklist:

- [ ] I have added Rust doc comments to structs, enums, traits and functions
- [ ] I have made corresponding changes to the documentation
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I rebased on the latest `main` branch

PR #1126, opened 2022-12-09
Adding registry to Altair

mustermeiszer:


runtime/altair/src/lib.rs
```diff
@@ -1140,6 +1198,24 @@ impl pallet_pool_system::Config for Runtime {
...
+	type CurrencyId = CurrencyId;
+	type Event = Event;
+	type InterestRate = Rate;
+	type MaxSizeMetadata = MaxSizeMetadata;
+	type MaxTokenNameLength = MaxTrancheNameLengthBytes;
+	type MaxTokenSymbolLength = MaxTrancheSymbolLengthBytes;
+	type MaxTranches = MaxTranches;
+	type ModifyPool = pallet_pool_system::Pallet<Self>;
+	type Permission = Permissions;
+	type PoolCreateOrigin = EnsureSigned<AccountId>;
```
gruberb, line 1212: We could use the type alias `PoolCreateOrigin`, we also use this in `PoolSystem`.

gruberb (changes requested):
Do we need to add `pallet_pool_registry` to the benchmarks as well?
https://github.com/centrifuge/centrifuge-chain/pull/1125/files#diff-dca6eccb11225cda4be1ac1e2a66128f7e7d140f567115a0ab702d8b0b9473d3R1642

mustermeiszer:
Good point will do that. Cc: @mikiquantum

mustermeiszer:
I checked again, and the registry does not expose all needed benchmarking yet. So this is not possible without manually adapting the weights afterwards. So won't add the benchmarks. @gruberb could you remove the request for changes.

runtime/altair/src/lib.rs
mustermeiszer: That is a good catch! We should change this to democracy!

mustermeiszer: The `PoolCreateOrigin` though, should be removed from `PoolSystem` ^^

gruberb: Can you explain why?

mustermeiszer:
Yes, but the registry does not implement all needed methods. I.e. it does not benchmark the extrinsic currently. See: https://github.com/centrifuge/centrifuge-chain/blob/4a10ae7a7eaff0062e7aca3bcba4b4129e909bc8/pallets/pool-registry/src/benchmarking.rs#L23-L150

This means, if we add it to the benchmarking clause in the runtimes, the next time we run the benchmarks for that runtime, we will get errors and need to manually add the older ones from the pool-systems. So I would like to finish the benchmarks in the registry first, and life we the imperfect benches till then.

runtime/altair/src/lib.rs
mustermeiszer: We don't use it anywhere, as we create the pool from the registry.

runtime/altair/src/lib.rs
```diff
@@ -413,7 +413,65 @@ impl InstanceFilter<Call> for ProxyType {
 	fn filter(&self, c: &Call) -> bool {
 		match self {
 			ProxyType::Any => true,
-			ProxyType::NonTransfer => !matches!(c, Call::Balances(..)),
+			ProxyType::NonTransfer => {
```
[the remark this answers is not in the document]
mustermeiszer: The reason for this is, that `Forbidlisting` is way harder to get right, than `Allowlisting`. Assume, we add another pallet that internally does some token transfers (e.g. investing into some pool). Then we will have an unnoticed breaking change. Where as, when we are explicit about what we are allowing, we will always recognize once we "need" to add something. 

Polkadot has the same setup.

mikiquantum (dismissed):
LGTM!

mikiquantum (approved):

PR #1127, opened 2022-12-09
Add portfolio_valuation and max_borrow_amount RPC calls to altair and dev

gruberb:
# Description

Adding `pool_valuation` RPC endpoint to `altair` and `development`

Closes: https://github.com/centrifuge/centrifuge-chain/issues/750

## Type of change

# How Has This Been Tested?

`cargo test --workspace --release --features test-benchmarks,try-runtime,runtime-benchmarks`

# Checklist:


runtime/development/src/lib.rs
```diff
@@ -1914,6 +1914,15 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl runtime_common::apis::LoansApi<Block, PoolId, Balance> for Runtime {
+		fn pool_valuation(pool_id: PoolId) -> Option<Balance>{
+			match pallet_loans::Pallet::<Runtime>::update_nav_of_pool(pool_id) {
```
gruberb, line 1919: The `impl PoolNAV` is using `update_nav`, the same name es the extrinsic for `pallet_loans`. If I try to use `update_nav` here, it tries to use the extrinsic instead of the one from the `PoolNAV` trait. Maybe some  trait bound magic we could do otherwise? 

In the end, `update_nav` is calling `update_nav_of_pool` internally, so I am using this instead.

runtime/altair/src/lib.rs
```diff
@@ -1573,6 +1574,15 @@ impl_runtime_apis! {
...
 
+	impl runtime_common::apis::LoansApi<Block, PoolId, Balance> for Runtime {
+		fn pool_valuation(pool_id: PoolId) -> Option<Balance>{
+			match pallet_loans::Pallet::<Runtime>::update_nav_of_pool(pool_id) {
+				Ok((_, value)) => Some(value),
+				Err(_) => None,
+			}
+		}
+	}
+
```
mustermeiszer, line 1585: The api should do something like: 

```rust
	impl runtime_common::apis::LoansApi<Block, PoolId, InstanceId, Balance> for Runtime {
		fn nav(id: PoolId) -> Option<Balance> {
			// TODO: Error out if pool does not exists in the loans-pallet
			if !pallet_pools::Pool::<Runtime>::get(id).is_some() {
				return None;
			}
			pallet_loans::Pallet::<Runtime>::update_nav_of_pool(id)
				.ok()
				.map(|(latest, _)| latest.into())
		}

		fn max_borrow_amount(id: PoolId, loan_id: InstanceId) -> Option<Balance> {
			let now = <pallet_timestamp::Pallet::<Runtime> as UnixTime>::now().as_secs();
			pallet_loans::Loan::<Runtime>::get(id, loan_id)
				.map(|loan_details| loan_details.max_borrow_amount(now).into())
		}
	}
```

mustermeiszer: Naming of course not right here. I just took the code from the old branch I implemented this on.

mustermeiszer (commented):
Looks already good. Just want to extend the API a bit more.

runtime/altair/src/lib.rs
mustermeiszer: Best would be to do all the checks that this section does https://github.com/centrifuge/centrifuge-chain/blob/4a10ae7a7eaff0062e7aca3bcba4b4129e909bc8/pallets/loans/src/functions.rs#L357-L393 also in the API code.

pallets/loans/src/functions.rs
```diff
@@ -766,4 +767,35 @@ impl<T: Config> Pallet<T> {
...
 		)
 	}
+
+	pub fn get_max_borrow_amount(pool_id: PoolIdOf<T>) -> Result<T::Balance, DispatchError> {
+		let now = Self::now();
+
+		let sum: T::Balance = Zero::zero();
+
+		let active_loans = ActiveLoans::<T>::get(pool_id);
+		for active_loan in active_loans.iter() {
```
gruberb, line 777: Using `.map()` didn't work out in a nice way, so I figure a `foreach` iss not the worst solution.

thea-leake (dismissed):
LGTM!

runtime/altair/src/lib.rs
```diff
@@ -1591,8 +1591,32 @@ impl_runtime_apis! {
...
+				return None;
+			}
+
+			match pallet_loans::Pallet::<Runtime>::update_nav_of_pool(pool_id) {
+				Ok((_, value)) => Some(value),
+				Err(_) => None,
+			}
+		}
+
+		fn max_borrow_amount(pool_id: PoolId) -> Option<Balance> {
```
offerijns, line 1606: I think there's a confusion about this endpoint. This should return the max borrow amount for an individual loan, not for the entire pool (I don't think there's a use case for that in the frontend).

pallets/loans/src/functions.rs
```diff
@@ -766,4 +767,35 @@ impl<T: Config> Pallet<T> {
...
+			};
+			ensure!(valid, Error::<T>::LoanMaturityDatePassed);
+
+			// check for max borrow amount
+			let current_debt = T::InterestAccrual::current_debt(
+				active_loan.interest_rate_per_sec,
+				active_loan.normalized_debt,
+			)?;
+
+			sum.saturating_add(active_loan.max_borrow_amount(current_debt));
```
mustermeiszer, line 796: What jeroen said!

mustermeiszer (changes requested):
Change from calculating maximal borrow amount of the whole pool, to the max borrow amount of a single loan, via submitting the loan id also. 

Furthermore, I would like to have a test for the runtime API. Cosmin has a PR open for the reward API that contains a test setup. In the integration tests folder.

gruberb:
> Furthermore, I would like to have a test for the runtime API. Cosmin has a PR open for the reward API that contains a test setup. In the integration tests folder.

Instead of copying it, I can wait for https://github.com/centrifuge/centrifuge-chain/pull/1124 to be merged.

@cdamian, do you have an idea when your PR might be merged so I can re-use the test-setup for RPC tests?

cdamian:
> > Furthermore, I would like to have a test for the runtime API. Cosmin has a PR open for the reward API that contains a test setup. In the integration tests folder.
> 
> Instead of copying it, I can wait for #1124 to be merged.
> 
> @cdamian, do you have an idea when your PR might be merged so I can re-use the test-setup for RPC tests?

@gruberb I think we want to merge that after the update to `polkadot-0.9.32` is done.

pallets/loans/src/functions.rs
branan: `.fold()` is probably the right iterator method here, but I think the for loop is fine

runtime/integration-tests/src/runtime_apis/pools.rs
```diff
@@ -0,0 +1,187 @@
...
+				<sr25519::Pair as sp_core::Pair>::from_string("//Bob", None)
+					.unwrap()
+					.public()
+					.into_account(),
+			);
+
+			let admin_account = account::<AccountId>("//Alice", 0, 0);
+			let borrower_account = account::<AccountId>("//Bob", 0, 0);
+
+			<development_runtime::Balances>::set_balance(
```
mustermeiszer, line 66: please instead of using the development runtime explicitly user the import `crate::chain::centrifuge::PALLET_NAME_HERE`. This will automatically import the right runtime.

runtime/integration-tests/src/runtime_apis/pools.rs
```diff
@@ -0,0 +1,187 @@
...
+			<development_runtime::Loans>::create(
+				RuntimeOrigin::signed(borrower),
+				pool_id,
+				collateral,
+			)
+			.expect("Creation of loan should not fail");
+		})
+		.with_api(|api, latest| {
+			let valuation = api.portfolio_valuation(&latest, 3).unwrap();
+			assert_eq!(valuation, Some(0));
```
mustermeiszer, line 185: Would be nice if we could warp the time a little and use a fixed value instead.

runtime/integration-tests/src/runtime_apis/pools.rs
gruberb: I used the `rewards` runtime api test as a sample and thought there was a reason we are using explicitly `development_runtime` there, since it was approved just a few weeks ago! Will change!

runtime/altair/src/lib.rs
```diff
@@ -1666,8 +1666,32 @@ impl_runtime_apis! {
...
+				return None;
+			}
+
+			match pallet_loans::Pallet::<Runtime>::update_nav_of_pool(pool_id) {
+				Ok((_, value)) => Some(value),
+				Err(_) => None,
+			}
+		}
+
+		fn max_borrow_amount(pool_id: PoolId, loan_id: ItemId) -> Option<Balance> {
```
lemunozm, line 1681: I'm going to make use of my pedantic new role 😝:

`Result` has an `ok()` to return a `None`, when there is an `Err`. Maybe it can be applied here. The whole function could be something like: 
```rust
fn max_borrow_amount(pool_id: PoolId, loan_id: ItemId) -> Option<Balance> {
    pallet_pool_system::Pool::<Runtime>::get(pool_id)?;
    pallet_loans::Pallet::<Runtime>::get_max_borrow_amount(pool_id, loan_id).ok()
}
```

runtime/altair/src/lib.rs
```diff
@@ -1666,8 +1666,32 @@ impl_runtime_apis! {
 			let pool = pallet_pool_system::Pool::<Runtime>::get(pool_id)?;
 			pool.tranches.tranche_currency(tranche_loc).map(Into::into)
 		}
+
+		fn portfolio_valuation(pool_id: PoolId) -> Option<Balance> {
```
lemunozm, line 1670: Similar here

src/rpc/pools.rs
```diff
@@ -201,4 +214,35 @@ where
 			.map_err(|e| runtime_error("Unable to query tranche currency.", e))?
 			.ok_or_else(|| invalid_params_error("Pool or tranche not found."))
 	}
+
+	fn portfolio_valuation(&self, pool_id: PoolId, at: Option<Block::Hash>) -> RpcResult<Balance> {
+		let api = self.client.runtime_api();
+		let at = if let Some(hash) = at {
```
lemunozm, line 220: What about:

```rust
let at = BlockId::hash(at.unwrap_or(self.client.info().best_hash))
```

to avoid the `if/else` ?

src/rpc/pools.rs
```diff
@@ -201,4 +214,35 @@ where
...
+	}
+
+	fn max_borrow_amount(
+		&self,
+		pool_id: PoolId,
+		loan_id: LoanId,
+		at: Option<Block::Hash>,
+	) -> RpcResult<Balance> {
+		let api = self.client.runtime_api();
+		let at = if let Some(hash) = at {
```
lemunozm, line 238: Same here

lemunozm (commented):
Added some minor Rust comments that you can avoid completely if you want 🤓

runtime/altair/src/lib.rs
```diff
@@ -1666,8 +1666,32 @@ impl_runtime_apis! {
...
+			if !pallet_pool_system::Pool::<Runtime>::get(pool_id).is_some() {
```
lemunozm, line 1671: No completely sure about, but I think `contains_key` avoid the deserialization of the data checking only if it exists

wischli: > No completely sure about, but I think `contains_key` avoid the deserialization of the data checking only if it exists

+1 from my side. If we don't act on the returned value, we should always check with `contains_key` instead of `get`. This API could be called quite frequently, so any simple optimization is very valuable.

gruberb: Good call, thanks!

runtime/integration-tests/src/runtime_apis/pools.rs
gruberb: I was curious about that too. Can you point me to the fudge feature where we can warp time?

gruberb:
@mustermeiszer Can you re-review this?

runtime/integration-tests/src/runtime_apis/pools.rs
```diff
@@ -0,0 +1,91 @@
...
+				90 * 60 * 60 * 24,
+				&mut nft_manager,
+			);
+
+			let alice = sp_runtime::AccountId32::from(
+				<sr25519::Pair as sp_core::Pair>::from_string("//Alice", None)
+					.unwrap()
+					.public()
+					.into_account(),
+			);
```
mustermeiszer, line 56: Remove

runtime/integration-tests/src/runtime_apis/pools.rs
```diff
@@ -0,0 +1,91 @@
...
+			for call in set_default_pool {
+				let res = UnfilteredDispatchable::dispatch_bypass_filter(
+					call,
+					RuntimeOrigin::signed(Keyring::Admin.into()),
+				);
+				// assert_ok!(res);
```
mustermeiszer, line 62: Unused?

runtime/integration-tests/src/runtime_apis/pools.rs
```diff
@@ -0,0 +1,91 @@
...
+			}
+			for call in set_loans_for_pools {
+				let res = UnfilteredDispatchable::dispatch_bypass_filter(
+					call,
+					RuntimeOrigin::signed(Keyring::Admin.into()),
+				);
+				// assert_ok!(res);
```
mustermeiszer, line 69: Unused?

runtime/integration-tests/src/runtime_apis/pools.rs
```diff
@@ -0,0 +1,91 @@
...
+			}
+			for call in issue_default_loans {
+				let res = UnfilteredDispatchable::dispatch_bypass_filter(
+					call,
+					RuntimeOrigin::signed(Keyring::Admin.into()),
+				);
+				// assert_ok!(res);
```
mustermeiszer, line 76: Unused?

runtime/integration-tests/src/runtime_apis/pools.rs
```diff
@@ -0,0 +1,91 @@
...
+			let currency = api.currency(&latest, 0).unwrap();
+			assert_eq!(currency, Some(CurrencyId::AUSD));
+
+			let valuation = api.portfolio_valuation(&latest, 0).unwrap();
+			assert_eq!(valuation, Some(0));
+
+			let max_borrow_amount = api
+				.max_borrow_amount(&latest, 0, LoanId::from(0_u16))
+				.unwrap();
+			assert_eq!(max_borrow_amount, None);
```
mustermeiszer, line 89: This means the loan is not calculating the max_borrow_amount correctly, right? 

So this test is currently not really helpful.

runtime/integration-tests/src/runtime_apis/pools.rs
mustermeiszer: you can use 

```
// Not sure if milliseconds or seconds here...
pallet_timestamp::Now::<Runtime>::set(TIME);
```

mustermeiszer: But in the `startup` closure

mustermeiszer (changes requested):
The test is currently not working IMO. 
* The dispatched calls are not checked if they succeeded
* The valuation is zero due to no time passing
* The max_borrow_amount fails -> Actually this would be interessting to check, why it is returning None here...

gruberb:
> The test is currently not working IMO.
> 
> * The dispatched calls are not checked if they succeeded
> * The valuation is zero due to no time passing
> * The max_borrow_amount fails -> Actually this would be interessting to check, why it is returning None here...

Uh, the latest commit wasn't there. The cleanups are in there. 

Yes, my question is how to warp time so we can get a loan amount. As I understand, we add the pool and loan to the latest block, and don't wait a few blocks to see how it progresses.

mustermeiszer:
This makes sense. I already wondered about the commented once ^^

Does it work with the proposal of mine? Setting the storage of `Now`?

gruberb:
> Does it work with the proposal of mine? Setting the storage of `Now`?

Playing around with different versions. I am getting:

```
thread 'runtime_apis::pools::test' panicked at 'Timestamp slot must match `CurrentSlot`', /Users/gruberbastian/.cargo/git/checkouts/substrate-f4423eebfa0046a3/5ea6d95/frame/aura/src/lib.rs:299:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

mustermeiszer:
Did you try setting the storage of Now in the pallet timeataml directly?

gruberb:
All updated again @mustermeiszer

pallets/loans/src/functions.rs
```diff
@@ -766,4 +766,40 @@ impl<T: Config> Pallet<T> {
...
+		ensure!(
+			active_loan.write_off_status == WriteOffStatus::None,
+			Error::<T>::WrittenOffByAdmin
+		);
+
+		// make sure maturity date has not passed if the loan has a maturity date
+		let valid = match active_loan.loan_type.maturity_date() {
+			Some(md) => md > now,
+			None => true,
+		};
```
NunoAlexandre, line 794: I think we could simplify this to the following but I'm happy keeping it as is

```suggestion
		let valid = active_loan.loan_type.maturity_date().map(|md| md > now).unwrap_or(true);
```

runtime/integration-tests/src/pools/utils/dispatch.rs
```diff
@@ -0,0 +1,30 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+//! Utilities around the dispatching calls
+use frame_support::{assert_ok, dispatch::UnfilteredDispatchable};
+
+use crate::chain::centrifuge::RuntimeOrigin;
+
+macro_rules! dispatch {
```
NunoAlexandre, line 18: Can we document this?

runtime/integration-tests/src/pools/utils/investments.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::{
+	chain,
+	chain::{
+		centrifuge,
+		centrifuge::{RuntimeCall, RuntimeOrigin},
+	},
+	pools::utils::{accounts::Keyring, loans::update_nav_call, pools::close_epoch_call},
+};
+
+/// Funds a Pool with a given PoolId. The pool must already exist and initalised.
```
NunoAlexandre, line 29: Usually docs are imperative (regarding the Fund~s~)
```suggestion
/// Fund a Pool with a given PoolId. The pool must already exist and be initialised.
```

runtime/integration-tests/src/pools/utils/time.rs
```diff
@@ -14,10 +14,20 @@
...
+/// NOTE: In milliseconds!
 pub const START_DATE: u64 = 1640995200u64;
 
+pub mod millisecs {
+	pub const MILLI_SECOND: u64 = 1000u64;
+	pub const MILLI_SECONDS_PER_MINUTE: u64 = 60 * MILLI_SECOND;
+	pub const MILLI_SECONDS_PER_HOUR: u64 = 60 * MILLI_SECONDS_PER_MINUTE;
+	pub const MILLI_SECONDS_PER_DAY: u64 = 24 * MILLI_SECONDS_PER_HOUR;
+	pub const MILLI_SECONDS_PER_YEAR: u64 = 365 * MILLI_SECONDS_PER_DAY;
+}
```
NunoAlexandre, line 27: This could actually be in libs/primitives but if it's only used here now this is also fine

runtime/integration-tests/src/pools/utils/time.rs
```diff
@@ -14,10 +14,20 @@
...
+///
+/// NOTE: In milliseconds!
 pub const START_DATE: u64 = 1640995200u64;
 
+pub mod millisecs {
+	pub const MILLI_SECOND: u64 = 1000u64;
+	pub const MILLI_SECONDS_PER_MINUTE: u64 = 60 * MILLI_SECOND;
+	pub const MILLI_SECONDS_PER_HOUR: u64 = 60 * MILLI_SECONDS_PER_MINUTE;
+	pub const MILLI_SECONDS_PER_DAY: u64 = 24 * MILLI_SECONDS_PER_HOUR;
+	pub const MILLI_SECONDS_PER_YEAR: u64 = 365 * MILLI_SECONDS_PER_DAY;
```
NunoAlexandre, line 26: Do we care about leap years?

NunoAlexandre (commented):
LGTM overall, just a few nits 🚀

runtime/integration-tests/src/pools/utils/time.rs
wischli: Polkadot does by counting 365.25 days per year. So we should probably as well?

gruberb: Since this is used for testing purposes only so far, we might not care about leap years? If we encounter a problem, this will come up in the tests first.

runtime/integration-tests/src/pools/utils/time.rs
gruberb: See comment below. Since it's not clear about leap year etc. we can leave it in the test module. If needed at some point, we can move it to the general `cfg-primitives` crate.

NunoAlexandre: On the same page 👍

runtime/integration-tests/src/pools/utils/dispatch.rs
gruberb: Good point, done!

NunoAlexandre (approved):
Awesome 👌

mustermeiszer (changes requested):
Still need to fix the `max_borrow_amount` either here or in a seperate PR.

mustermeiszer:
> Still need to fix the `max_borrow_amount` either here or in a seperate PR.

@gruberb the `NftManager` has a bug, that it is NOT increasing the ids. Just for you to know when creating a second loan.

runtime/integration-tests/src/runtime_apis/pools.rs
gruberb: I think I don't quite understand what should happen when. I think there is no `Loan` with the ID `0` as you mentioned, we start with `1`. So I would expect an `Error`, however in the APIs, we don't return errors so far. 

If that's not the case, what do you mean with "first borrow"?

PR #1132, opened 2022-12-14
Fix weight proof sizes

thea-leake:
# Description

Fixes weight proof size settings I had set incorrectly in upgrade.

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [ x] I have performed a self-review of my code
- [x ] My changes generate no new warnings
- [ x] New and existing unit tests pass locally with my changes
- [ x] I rebased on the latest `main` branch

### DIFF
--- a/runtime/altair/src/lib.rs
+++ b/runtime/altair/src/lib.rs
@@ -132,6 +132,8 @@ pub fn native_version() -> NativeVersion {
 	}
 }
 
+const MAX_BLOCK_WEIGHT: Weight = MAXIMUM_BLOCK_WEIGHT.set_proof_size(MAX_POV_SIZE as u64);
+
 parameter_types! {
 	pub const MaximumBlockWeight: Weight = MAXIMUM_BLOCK_WEIGHT;
 	pub const Version: RuntimeVersion = VERSION;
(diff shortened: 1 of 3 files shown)


mustermeiszer (approved):
That slip through all of our reviews! Really good that you looked through again! Thanks!

runtime/altair/src/lib.rs
```diff
@@ -132,6 +132,8 @@ pub fn native_version() -> NativeVersion {
 	}
 }
 
+const MAX_BLOCK_WEIGHT: Weight = MAXIMUM_BLOCK_WEIGHT.set_proof_size(MAX_POV_SIZE as u64);
+
 parameter_types! {
 	pub const MaximumBlockWeight: Weight = MAXIMUM_BLOCK_WEIGHT;
```
mikiquantum, line 138: This is not needed anymore right?

thea-leake: Good catch, I had spaced that.  It looks like it's actually being used once later for the scheduler config.

I've removed that in the scheduler config in the work to have the max proof size pulled from the relay chain here https://github.com/centrifuge/centrifuge-chain/pull/1134.

PR #1135, opened 2022-12-15
relay-cli: Handle None case for shared base path

cdamian:
# Description

When running the centrifuge-chain locally, if no `base-path` is provided to the relay chain args then a random temp directory is used which can cause inconsistent behavior such as the chain not producing blocks due to invalid/stale data in the db.

## Changes and Descriptions

The changes in this PR aim at handling the case where `shared_params().base_path()` is a `Result(None)` which results in the behavior described above.

## Type of change

# How Has This Been Tested?

- Tested by running the chain locally and confirming that the `base_path` set in the init script is used for the relay chain.

# Checklist:

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -536,9 +536,12 @@ impl CliConfiguration<Self> for RelayChainCli {
 	}
 
 	fn base_path(&self) -> Result<Option<BasePath>> {
-		self.shared_params()
+		Ok(self
+			.shared_params()
 			.base_path()
-			.or_else(|_| Ok(self.base_path.clone().map(Into::into)))
+			.ok()
+			.flatten()
+			.or_else(|| self.base_path.clone().map(Into::into)))
 	}
 
 	fn role(&self, is_dev: bool) -> Result<sc_service::Role> {


src/command.rs
```diff
@@ -536,9 +536,11 @@ impl CliConfiguration<Self> for RelayChainCli {
 	}
 
 	fn base_path(&self) -> Result<Option<BasePath>> {
-		self.shared_params()
+		Ok(self
+			.shared_params()
 			.base_path()
-			.or_else(|_| Ok(self.base_path.clone().map(Into::into)))
+			.unwrap_or_else(|_| self.base_path.clone().map(Into::into))
+			.or_else(|| self.base_path.clone().map(Into::into)))
```
mustermeiszer, line 543: This means in hierarchical order:

1. Take the path directly provided as relay chain argument
2. If not present use the one `RelayChainCli` was given as a field
3. ???

Why the last `or_else` clause()? If this intends to call the logic of the substrate core cli, shouldnt' this be `self.base_path()`

cdamian: This was my starting point for this:

```
match self.shared_params().base_path() {
			Ok(base_path_opt) => match base_path_opt {
				Some(base_path) => Ok(base_path.into()),
				None => Ok(self.base_path.clone().map(Into::into)),
			},
			Err(_) => Ok(self.base_path.clone().map(Into::into)),
		}
```

cdamian: Updated to a cleaner version, as per our Slack conversation. Thank you! :heart:

mustermeiszer (approved):
Thanks for the quick fix!

cc: @mikiquantum the reason being is that we had a silent logic change on the Substrate side, where the cli implementation now provides temp paths automatically when running in dev mode and no path is provided to the relay-chain cli.

mustermeiszer (approved):
LGTM!

PR #1136, opened 2022-12-15
Adding Codeowners

mustermeiszer:
# Description
This PR adds a CODEOWNER file to our repository. This alone does not enable anything but when the repository enforces code owners to do a review, only with a review of them PRs can be merged. 

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -29,6 +29,14 @@ pallets/keystore/* @cdamian
 # Changes to libs
 libs/* @branan @mikiquantum @mustermeiszer
 
+## Changes to specific libraries
+### cfg-traits crate
+#### rewards module
+libs/traits/src/rewards.rs
+
+#### ensure module
+libs/traits/src/ops.rs
+
 # Changes to runtime
 runtime/* @branan @mikiquantum @mustermeiszer
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,13 +4,13 @@
 
 # These owners will be the default owners for everything in
 # the repo. Unless a later match takes precedence,
-*       @offerijns @mikiquantum @mustermeiszer
+*       @branan @mikiquantum @mustermeiszer @offerijns
 
 # Changes to the service of our chain.
 src/*    @mikiquantum
 
 # Changes to pallets
-pallets/*  @mustermeiszer
+pallets/*  @branan @mikiquantum @mustermeiszer
 
 ## Changes to specific pallets
 ### Connectors related logic
@@ -20,12 +20,17 @@ pallets/connectors/* @NunoAlexandre
 pallets/rewards/* @lemunozm
 pallets/liquidity-rewards/* @lemunozm
 
+### Nft-sales related logic
+pallets/nft-sales/* @NunoAlexandre
+
+### Pallet Keystore
+pallets/keystore/* @cdamian
 
 # Changes to libs
-libs/* @mikiquantum @mustermeiszer @branan
+libs/* @branan @mikiquantum @mustermeiszer
 
 # Changes to runtime
-runtime/* @mikiquantum @mustermeiszer
+runtime/* @branan @mikiquantum @mustermeiszer
 
 ## Changes to folders inside runtime
 ### Changes to runtime/integration-tests
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,35 @@
+# Centrifudge Code Owners
+#
+# Order is important; the last matching pattern takes the most
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+*       @offerijns @mikiquantum @mustermeiszer
+
+# Changes to the service of our chain.
+src/*    @mikiquantum
+
+# Changes to pallets
+pallets/*  @mustermeiszer
+
+## Changes to specific pallets
+### Connectors related logic
+pallets/connectors/* @NunoAlexandre
+
+### Rewards related logic
+pallets/rewards/* @lemunozm
+pallets/liquidity-rewards/* @lemunozm
+
+
+# Changes to libs
+libs/* @mikiquantum @mustermeiszer @branan
+
+# Changes to runtime
+runtime/* @mikiquantum @mustermeiszer
+
+## Changes to folders inside runtime
+### Changes to runtime/integration-tests
+runtime/integration-tests/* @mustermeiszer
+
+### Changes to runtime/integration-tests/xcm/
+runtime/integration-tests/xcm/* @NunoAlexandre
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -32,10 +32,10 @@ libs/* @branan @mikiquantum @mustermeiszer
 ## Changes to specific libraries
 ### cfg-traits crate
 #### rewards module
-libs/traits/src/rewards.rs
+libs/traits/src/rewards.rs @lemunozm
 
 #### ensure module
-libs/traits/src/ops.rs
+libs/traits/src/ops.rs @lemunozm
 
 # Changes to runtime
 runtime/* @branan @mikiquantum @mustermeiszer
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,10 +1,10 @@
 # Centrifudge Code Owners
 
 # Changes to .github
-.github/* @mikiquantum @mustermeiszer @branan @NunoAlexandre @offerjins
+.github/* @mikiquantum @mustermeiszer @branan @NunoAlexandre @offerijns
 
 # Changes to ci
-ci/* @mikiquantum @mustermeiszer @branan @NunoAlexandre @offerjins
+ci/* @mikiquantum @mustermeiszer @branan @NunoAlexandre @offerijns
 
 # Changes to the service of our chain.
 src/* @mikiquantum @mustermeiszer @branan @NunoAlexandre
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,40 @@
...
+pallets/liquidity-rewards/* @lemunozm
+
+### Nft-sales related logic
+pallets/nft-sales/* @NunoAlexandre
+
+### Pallet Keystore
+pallets/keystore/* @cdamian
+
+# Changes to libs
+libs/* @branan @mikiquantum @mustermeiszer
```
lemunozm, line 30: Here could have sections where any other than these three can review it (in my case, fees, ensure, rewards modules, i.e). Is it needed to express that level too?

mustermeiszer: Yes, that would be needed. Tell me what you want and I add it

lemunozm: `fees`, `ensure`, `rewards` modules for me in `cfg_traits` 👍🏻

.github/CODEOWNERS
```diff
@@ -0,0 +1,48 @@
...
+
+#### ensure module
+libs/traits/src/ops.rs @lemunozm
+
+# Changes to runtime
+runtime/* @branan @mikiquantum @mustermeiszer
+
+## Changes to folders inside runtime
+### Changes to runtime/integration-tests
+runtime/integration-tests/* @mustermeiszer
```
NunoAlexandre, line 49: I would prefer to be included to this rule as well. Currently the integration tests are divided by pools and xcm tests, so imv it makes sense to have us both on the root level. wdyt?

.github/CODEOWNERS
```diff
@@ -0,0 +1,48 @@
+# Centrifudge Code Owners
+#
+# Order is important; the last matching pattern takes the most
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+*       @branan @mikiquantum @mustermeiszer @offerijns
```
NunoAlexandre, line 7:
```suggestion
* @branan @mikiquantum @mustermeiszer @offerijns
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,48 @@
...
+
+# Changes to the service of our chain.
+src/*    @mikiquantum
```
NunoAlexandre, line 10:
```suggestion
src/* @mikiquantum
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,48 @@
...
+
+# Changes to pallets
+pallets/*  @branan @mikiquantum @mustermeiszer
```
NunoAlexandre, line 13:
```suggestion
pallets/* @branan @mikiquantum @mustermeiszer
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,48 @@
...
+
+# Changes to runtime
+runtime/* @branan @mikiquantum @mustermeiszer
+
+## Changes to folders inside runtime
+### Changes to runtime/integration-tests
+runtime/integration-tests/* @mustermeiszer
+
+### Changes to runtime/integration-tests/xcm/
+runtime/integration-tests/xcm/* @NunoAlexandre
```
NunoAlexandre, line 52:
```suggestion
runtime/integration-tests/xcm/* @NunoAlexandre

```

.github/CODEOWNERS
NunoAlexandre, line 7: Btw, @offerijns usually doesn't have the bandwidth to respond to PRs in a short timeline. The other engineer in the Europe timezone is yourself, also not at full capacity. I would suggest adding myself here for that reason.

offerijns: Just sharing my two cents on this: I think it's important that we work on prioritising reviews above other work, and this is the right fix for that problem (and from what I've seen, this is going increasingly better in the protocol circle). But, as an example, if there is a PR that is touching the `pool-system` pallet, I think at least 1 of Branan/Frederik/me should review that (and the same logic applies to other pallets for other people, although `pool-system` is probably most critical). I don't think this should be reviewed solely by other people, it should imo always include someone from that list, even if this means waiting another day until they are available.

I'm wondering if this means we should not even remove this asterisk rule, and just have specific rules covering every individual folder. I checked the code owners docs, if a file is not covered by any of the rules (e.g. a change to just the README), then no code owner review would be required and any contributor with write access (i.e. the entire protocol circle) can review.

mustermeiszer: My main reasoning is, that the asterix prevents us from missing a rule. I.e. We need to adapt the file everytime we add a new pallet and define a owner. 

But this is just another argument, I am not against removing it. Just to take into account.

.github/CODEOWNERS
mustermeiszer: I will still keep you in that line, additionally, as this will notify you for review. Wdyt?

NunoAlexandre: Sounds good, thanks 🙏

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
...
+# Changes to .github
+.github/* @mikiquantum @mustermeiszer @branan @NunoAlexandre @offerjins
+
+# Changes to ci
+ci/* @mikiquantum @mustermeiszer @branan @NunoAlexandre @offerjins
+
+# Changes to the service of our chain.
+src/* @mikiquantum @mustermeiszer @branan @NunoAlexandre
+
+# Changes to chain-specs
```
mustermeiszer, line 12: Do we actually need that line?

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
+# Centrifudge Code Owners
+
+# Changes to .github
+.github/* @mikiquantum @mustermeiszer @branan @NunoAlexandre @offerijns
+
+# Changes to ci
```
mustermeiszer, line 6: Do we actually need that?

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
+# Centrifudge Code Owners
+
+# Changes to .github
+.github/* @mikiquantum @mustermeiszer @branan @NunoAlexandre @offerijns
```
mustermeiszer, line 4: Do we actually need that?

mustermeiszer (commented):
Some thoughts that need comments from the others.

.github/CODEOWNERS
mustermeiszer: This is referring to ci folder changes

.github/CODEOWNERS
offerijns, line 4: I don't know much about how the workflows are setup.
```suggestion
.github/* @mikiquantum @mustermeiszer @branan @NunoAlexandre
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
...
+
+# Changes to ci
+ci/* @mikiquantum @mustermeiszer @branan @NunoAlexandre @offerijns
```
offerijns, line 7: Similarly
```suggestion
ci/* @mikiquantum @mustermeiszer @branan @NunoAlexandre
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
...
+
+# Changes to the service of our chain.
+src/* @mikiquantum @mustermeiszer @branan @NunoAlexandre
+
+# Changes to chain-specs
+res/* @mikiquantum @mustermeiszer @branan @NunoAlexandre
+
+# Changes to specific pallets
+## Connectors related logic
+pallets/connectors/* @NunoAlexandre
```
offerijns, line 17: I think I can be a backup for this.
```suggestion
pallets/connectors/* @NunoAlexandre @offerijns
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
...
+
+## Rewards related logic
+pallets/rewards/* @lemunozm
+pallets/liquidity-rewards/* @lemunozm
```
offerijns, line 21:
```suggestion
pallets/rewards/* @lemunozmn @mikiquantum
pallets/liquidity-rewards/* @lemunozm @mikiquantum
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
...
+
+## Nft-sales related logic
+pallets/nft-sales/* @NunoAlexandre
+
+## Pallet Keystore
+pallets/keystore/* @cdamian
```
offerijns, line 27:
```suggestion
pallets/keystore/* @cdamian  @mikiquantum
```

.github/CODEOWNERS
mustermeiszer: I think I see your points now @NunoAlexandre and @offerijns. I adapted the owner file.

NunoAlexandre (approved):
LGTM! Thanks @mustermeiszer

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
...
+pallets/permissions/* @mustermeiszer
+
+## Pallet Restricted-Tokens
+pallets/restricted-tokens/* @mustermeiszer
+
+# Changes to libs
+## Changes to specific libraries
+### cfg-traits crate
+#### rewards module
+libs/traits/src/rewards.rs @lemunozm
```
lemunozm, line 54: Not sure If we need an entry for the rest of the `libs/*`?

mustermeiszer: I thought this would be a place that everybody can already own. WDYT?

lemunozm: I agree 👍🏻

lemunozm (approved):

PR #1137, opened 2022-12-15
Updated weight max proof size in primitives with const, will be updat…

thea-leake:
Maximum block weight proof size update.

# Description
Proof size set on maximum proof weight using culumuls/polkadot max pov size const.

This will be updated to pull from the relay chain in near future, however we're going to go with a lower risk change for the upgrade release -- https://github.com/centrifuge/centrifuge-chain/pull/1134

Fixes # 

Ensures proof size set correctly

## Type of change

# How Has This Been Tested?

# Checklist:


mikiquantum (approved):
LGTM! 🙏

PR #1139, opened 2022-12-17
[WIP] Benchmark fixes

branan:
# Description

Ceaning up and fixing benchmarking and some weight handling for our release with polkadot 0.9.32

## Changes and Descriptions

[x] Cleaned up the weight handling between pool-system and pool-registry for update weights
[x] Restore ability to generate per-pallet weight info with trait defined
[x] Enable pool-registry benchmarks on altair and development runtimes
[ ] fix pool-registry benchmarks
[ ] Enable other disable benches on altair and evelopment runtimes
[ ] fix those benchmarks if needed
[ ] Investigate status of benchmarks on centrifuge runtime
[ ] Switch altair and development runtimes to use their own benched results for all pallets
[ ] Run the benchmarks on proper bench hardware
 

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
Building and running the benchmarks. Over. and over. and over. and over.

# Checklist:


scripts/frame-weight-template.hbs
```diff
@@ -0,0 +1,86 @@
+//! Autogenerated weights for {{pallet}}
```
mikiquantum, line 1: Why do we need this? I removed it a few commits ago in favor of the default substrate one, and it works fine for the Centrifuge runtime.

pallets/pool-registry/src/benchmarking.rs
```diff
@@ -11,37 +11,74 @@
...
-	// }: create(origin, caller, POOL, tranches.clone(), CurrencyId::AUSD, MAX_RESERVE, None)
-	// verify {
-	// 	let pool = get_pool::<T>();
-	// 	assert_input_tranches_match::<T>(pool.tranches.residual_top_slice(), &tranches);
-	// 	assert_eq!(pool.reserve.available, Zero::zero());
-	// 	assert_eq!(pool.reserve.total, Zero::zero());
-	// 	assert_eq!(pool.parameters.min_epoch_time, T::DefaultMinEpochTime::get());
-	// 	assert_eq!(pool.parameters.max_nav_age, T::DefaultMaxNAVAge::get());
-	// 	assert_eq!(pool.metadata, None);
-	// }
```
mustermeiszer, line 44: We still need to write a seperate case for 
* `update`
   * I think here it is enough to simply use the worst case of update and execute
* `execute_update`

pallets/pool-registry/src/mock.rs
```diff
@@ -133,18 +202,40 @@ impl<T: Config + pallet_pool_registry::Config> PoolMutate<T::AccountId, T::PoolI
...
+	type Accountant = PoolSystem;
+	type Amount = Balance;
+	type BalanceRatio = Rate;
+	type InvestmentId = TrancheCurrency;
+	type MaxOutstandingCollects = MaxOutstandingCollects;
+	type PreConditions = Always;
+	type RuntimeEvent = RuntimeEvent;
+	type Tokens = Tokens;
+	type WeightInfo = ();
+}
```
mustermeiszer, line 227: We have a mock for this pallet.

pallets/pool-registry/src/mock.rs
```diff
@@ -163,6 +254,58 @@ impl Config for Test {
...
+	type PreFungibleTransfer = cfg_traits::Always;
+	type PreFungiblesInspect = pallet_restricted_tokens::FungiblesInspectPassthrough;
+	type PreFungiblesInspectHold = cfg_traits::Always;
+	type PreFungiblesMutate = cfg_traits::Always;
+	type PreFungiblesMutateHold = cfg_traits::Always;
+	type PreFungiblesTransfer = cfg_traits::Always;
+	type PreReservableCurrency = cfg_traits::Always;
+	type RuntimeEvent = RuntimeEvent;
+	type WeightInfo = ();
+}
```
mustermeiszer, line 307: Not needed, orml-tokens are suficcient.

pallets/pool-registry/src/mock.rs
```diff
@@ -171,12 +314,150 @@ frame_support::construct_runtime!(
...
+				) && P::has(
+					PermissionScope::Pool(pool_id),
+					recv,
+					Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, UNION)),
+				)
+			}
+			_ => true,
+		}
+	}
+}
```
mustermeiszer, line 440: Also not needed. We should have a mock for permissions here.

scripts/frame-weight-template.hbs
branan: We need a version of the template that will generate the actual trait. The default one only generates an impl

branan:
@gruberb From where you left this today, I was able to get all the benches running cleanly when run as benchmarks, but some still fail when run as unit tests (due to differences in the mocks, I suppose). If you can manage to resolve that, I think we're ready to get review from somebody else and get this landed.

It'd also be great if you could double-check my work in the benchmarks to make sure I didn't miss anything that's different in pool-registry compared to the monolithic pools pallet, but I *think* it captures all the necessary complexity.

gruberb:
> From where you left this today, I was able to get all the benches running cleanly when run as benchmarks, but some still fail when run as unit tests (due to differences in the mocks, I suppose).

Interesting. For me 4 benchmarks are still failing. One interesting case is here (you might have to manually "load diff" or it won't show):
https://github.com/centrifuge/centrifuge-chain/pull/1139/files#diff-4b43095be32ccb64ba5d121f5b69e418873a5d2d16770aa532b6ac01c7c23eebR134-R139

If I add `MinUpdateDelay` to `1`, the `update_no_execution` benchmarks runs through. But I wonder why, since it should not? I comment this part out. Maybe there is a logical error in the code. But from what I understand based on the `update` impl in `pool-system` here (https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/pool-system/src/impls.rs#L286-L289), the update should be executed right away.

pallets/pool-registry/src/benchmarking.rs
```diff
@@ -11,140 +11,201 @@
...
+				<T as pallet_pool_system::Config>::MaxTranches>,
+		>,
+	}
+	register {
+		let n in 1..<T as pallet_pool_system::Config>::MaxTranches::get();
+		let caller: <T as frame_system::Config>::AccountId = create_admin::<T>(0);
+		let tranches = build_bench_input_tranches::<T>(n);
+		let origin = RawOrigin::Signed(caller.clone());
+		prepare_asset_registry::<T>();
+		create_pool::<T>(n, caller.clone())?;
```
branan, line 93: I had removed this line because with it here, when running against the altair runtime, `register` below fails with a `PoolInUse` error. I'd expect this failure, since `create_pool` and `register` both end up trying to create the same pool in the underlying `pool-system` storage.

I'm still seeing that error, so I suspect this is a difference between running benches against the mocks and a real runtime?

gruberb: The `register` here actually doesn't create a pool when using the `mock` runtime. I will move `create_pool` into the `PoolMutate` mock, which should solve the problem.

pallets/pool-registry/src/benchmarking.rs
```diff
@@ -11,140 +11,201 @@
...
-	// verify {
-	// 	let pool = get_pool::<T>();
-	// 	assert_update_tranches_match::<T>(pool.tranches.residual_top_slice(), &tranches);
-	// 	assert_eq!(pool.parameters.min_epoch_time, SECS_PER_DAY);
-	// 	assert_eq!(pool.parameters.max_nav_age, SECS_PER_HOUR);
-	// }
+		// Withdraw redeem order so the update can be executed after that
+		pallet_investments::Pallet::<T>::update_redeem_order(RawOrigin::Signed(investor.clone()).into(), TrancheCurrency::generate(POOL, locator), 0)?;
+
+		update_pool::<T>(changes.clone())?;
```
branan, line 191: This needs to be before the `update_redeem_order` above, otherwise on the altair runtime it is executed immediately and the `execute_update` invocation below fails

branan:
Those are the only two issues I found in invoking `./scripts/run_benchmark.sh altair-local pallet-pool-registry` with your latest changes, and the latter probably doesn't even need a mock change. Nearly there!

gruberb:
All good from my side @branan! Thanks for the support and finding these little issues I didn't catch!

gruberb (approved):
LGTM!

pallets/pool-registry/src/benchmarking.rs
```diff
@@ -11,140 +11,200 @@
...
+		T: pallet_pool_system::Config<PoolId = u64,
+			  TrancheId = [u8; 16],
+			  Balance = u128,
+			  CurrencyId = CurrencyId,
+			  EpochId = PoolEpochId,
+			  Rate = <T as Config>::InterestRate,
+			  MaxTokenNameLength = <T as Config>::MaxTokenNameLength,
+			  MaxTokenSymbolLength = <T as Config>::MaxTokenSymbolLength,
+			  MaxTranches = <T as Config>::MaxTranches>,
+		<T as pallet_investments::Config>::Tokens: Inspect<T::AccountId, AssetId = CurrencyId, Balance = u128>,
```
mustermeiszer, line 67: Don't wanna block this, but this binding basically forces us to use the investments pallet always in the registry.

Instead bindings should be based on traits of functionality rather than traits of pallets wherever possible.

pallets/pool-system/src/impls.rs
```diff
@@ -321,7 +309,7 @@ impl<T: Config> PoolMutate<T::AccountId, T::PoolId> for Pallet<T> {
 		Self::do_update_pool(&pool_id, &update.changes)?;
 
 		let num_tranches = pool.tranches.num_tranches().try_into().unwrap();
-		Ok(Some(T::WeightInfo::execute_update(num_tranches)).into())
+		Ok(num_tranches)
```
mustermeiszer, line 312: We must have the same check as in the `update` logic.
I.e. check whether we are in submission period. Otherwise, an update would be dangerous for the rebalance and epoch closing.

mustermeiszer (changes requested):
There is a missing check in the execute update logic. That needs to be included.

Besides this, looks good

pallets/pool-registry/src/benchmarking.rs
gruberb: Hm, this part is just taken from the `pallet-pool-system` benchmarking without changing it (https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/pool-system/src/benchmarking.rs#L47). 

I assume we would need to create a trait for it first to abstract away the investments pallet?

pallets/pool-system/src/impls.rs
gruberb: I thought here we get a scheduled updated, and check weather it's in the timeframe via:

```
		ensure!(
			Self::now() >= update.scheduled_time,
			Error::<T>::ScheduledTimeHasNotPassed
		);
```

But I will add the `InSubmissionPeriod` check here as well then!

pallets/interest-accrual/src/lib.rs
```diff
@@ -66,6 +66,7 @@ use sp_runtime::{
 };
 
 pub mod weights;
+pub use weights::WeightInfo;
```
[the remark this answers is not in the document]
mikiquantum: I can take care of regenerating the weights for each runtime in the next pr.

scripts/frame-weight-template.hbs
mikiquantum: Ok, will test it out when generating the runtime ones.

mikiquantum (approved):
LGTM, there are some errors when generating the runtime benchmarks, but I can take care of those in a following pr

gruberb:
@mustermeiszer could you re-review? :) I think everything should be there now!

PR #1141, opened 2022-12-19
Rewards: Implement distribution to groups with no stake

### LINKED ISSUE #1140: Rewards: Allow distribution to groups with no stake

* Addressing a bug

### Description

At distribution, some reward mechanisms need to be called to the `reward_group()` to change their internal stake, although the group does not contain any rewarded stake. To support that, an `is_ready()` needs to be added to the reward traits, which tells the system if the group is ready to receive the reward or not. In case not, the system will still call to `reward_group()` but with 0 rewards allowing the mechanism to change its state without rewarding anything.

Still, two issues where the reward could not be given:
- The reward destinated to a group that fails is not redistributed to the other groups.
- If all groups of the distribution list are not ready, the reward is not given.
But both can be determined by checking if the distribution events sum the expected rewards.

lemunozm:
# Description

Fixes #1140

## Changes and Descriptions

- Add `is_ready()` method.
- Modify mock to support that.

## Type of change

# How Has This Been Tested?
```sh
cargo test -p cfg-traits
cargo test -p pallet-rewards
```

# Checklist:

### DIFF
--- a/libs/traits/src/rewards.rs
+++ b/libs/traits/src/rewards.rs
@@ -28,6 +28,10 @@ pub trait GroupRewards {
 	/// Type used to identify the group
 	type GroupId;
 
+	/// Check if the group is ready to be rewarded.
+	/// Most of the cases it means that the group has stake that should be rewarded.
+	fn is_ready(group_id: Self::GroupId) -> bool;
+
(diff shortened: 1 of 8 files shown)


mikiquantum (approved):
LGTM!

Issue #918, opened 2022-08-23
Benchmarks for `pallet-investments`


mustermeiszer:
Once #909 is merged, we need to implement benchmarks for the `pallet-investments`

PR #1151, opened 2022-12-23
`pool-system` review/questions/notes

lemunozm:
PR used as a channel for questions/notes/hints

### DIFF
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -488,6 +488,7 @@ pub trait InvestmentAccountant<AccountId> {
 	) -> Result<(), Self::Error>;
 }
 
+// This trait seems to be used only by the own Investment pallet in production.
 /// A trait that allows to retrieve information
 /// about an investment class.
 pub trait InvestmentProperties<AccountId> {
--- a/pallets/investments/src/lib.rs
+++ b/pallets/investments/src/lib.rs
@@ -459,7 +459,7 @@ pub mod pallet {
 		/// amount is less than the current order, the balance
 		/// will be transferred from the pool to the calling
 		/// account.
-		#[pallet::weight(80_000_000)]
+		#[pallet::weight(80_000_000)] // Should we use a computed weights here?
 		pub fn update_invest_order(
 			origin: OriginFor<T>,
 			investment_id: T::InvestmentId,
@@ -1071,6 +1071,7 @@ where
 	}
 }
 
+// This impl seems to only be used for testing purposes.
 impl<T: Config> Investment<T::AccountId> for Pallet<T>
 where
 	<T::Accountant as InvestmentAccountant<T::AccountId>>::InvestmentInfo:
--- a/pallets/pool-system/src/lib.rs
+++ b/pallets/pool-system/src/lib.rs
@@ -820,6 +820,7 @@ pub mod pallet {
 						.expect("Solution exists. qed.")
 						.solution();
 
+					// Requires transactional because of this.
 					Self::do_execute_epoch(pool_id, pool, epoch, solution)?;
 					Self::deposit_event(Event::EpochExecuted {
 						pool_id,
@@ -835,7 +836,7 @@ pub mod pallet {
 					.expect("MaxTranches is u32. qed.");
 
 				// This kills the epoch info in storage.
-				// See: https://github.com/paritytech/substrate/blob/bea8f32e7807233ab53045fe8214427e0f136230/frame/support/src/storage/generator/map.rs#L269-L284
+				// See: https://github.com/paritytech/substrate/blob/bea8f32e7807233ab53045fe8214427e0f136230/frame/support/src/storage/generator/map.rs#L269-L284 // Broken link
 				*epoch_info = None;
 				Ok(Some(T::WeightInfo::execute_epoch(num_tranches)).into())
 			})
@@ -943,15 +944,9 @@ pub mod pallet {
 				.checked_add(&epoch.reserve)
 				.ok_or(Error::<T>::InvalidSolution)?;
 
-			// Mostly a sanity check. This is catched above.
-			ensure!(
-				currency_available.checked_sub(&acc_redeem).is_some(),
-				Error::<T>::InsufficientCurrency
-			);
-
 			let new_reserve = currency_available
 				.checked_sub(&acc_redeem)
-				.expect("Ensures ensures there is enough liquidity in the reserve. qed.");
+				.ok_or(Error::<T>::InsufficientCurrency)?;
 
 			Self::validate_pool_constraints(
 				PoolState::Healthy,
@@ -1114,6 +1109,7 @@ pub mod pallet {
 			Ok(())
 		}
 
+		// This method have to be always called from a transaction
 		fn do_execute_epoch(
 			pool_id: T::PoolId,
 			pool: &mut PoolDetailsOf<T>,


pallets/investments/src/lib.rs
```diff
@@ -1071,6 +1071,7 @@ where
 	}
 }
 
+// This impl seems to only be used for testing purposes.
 impl<T: Config> Investment<T::AccountId> for Pallet<T>
```
lemunozm, line 1075: Maybe the `Investment` trait is not needed because the test can call directly to the pallet (?).

libs/traits/src/lib.rs
```diff
@@ -488,6 +488,7 @@ pub trait InvestmentAccountant<AccountId> {
 	) -> Result<(), Self::Error>;
 }
 
+// This trait seems to be used only by the own Investment pallet in production.
 /// A trait that allows to retrieve information
 /// about an investment class.
 pub trait InvestmentProperties<AccountId> {
```
lemunozm, line 494: Why not remove the trait and use directly the only type that implements this (`InvestmentInfo`)?

pallets/pool-system/src/lib.rs
```diff
@@ -943,15 +944,9 @@ pub mod pallet {
...
-			// Mostly a sanity check. This is catched above.
-			ensure!(
-				currency_available.checked_sub(&acc_redeem).is_some(),
-				Error::<T>::InsufficientCurrency
-			);
-
 			let new_reserve = currency_available
 				.checked_sub(&acc_redeem)
-				.expect("Ensures ensures there is enough liquidity in the reserve. qed.");
+				.ok_or(Error::<T>::InsufficientCurrency)?;
```
lemunozm, line 949: Minor simplification 🤓

pallets/pool-system/src/lib.rs
```diff
@@ -1114,6 +1109,7 @@ pub mod pallet {
 			Ok(())
 		}
 
+		// This method have to be always called from a transaction
```
lemunozm, line 1112: This method makes calls to other pallets that modify their own storage, and it could fail in the middle of the process. If it happens, some data could remain changed on those pallets after the failure.

Those who use this method, should use transaction, i.e, the `execute_epoch` call.

pallets/pool-system/src/tranches.rs
```diff
@@ -696,6 +698,7 @@ where
 	where
 		F: FnMut(&Tranche<Balance, Rate, Weight, TrancheCurrency>) -> Result<R, DispatchError>,
 	{
+		// Why not self.tranches.iter().map(f)?
```
lemunozm, line 701: I think all functions in `tranches.rs` that return a `Vec`, could return instead `impl Iterator` reducing a lot of code and making it handier/rusty. I can do a PR with this suggestion if you agree.

pallets/pool-system/src/tranches.rs
```diff
@@ -1323,7 +1326,7 @@ where
 						.checked_pow(
 							n_tranches
 								.checked_sub(tranche.seniority)
-								.unwrap_or(u32::MAX),
+								.unwrap_or(u32::MAX), //u32::MAX sure?
```
lemunozm, line 1329: Seems like if the subtraction fails, it should be 0 here. Maybe it should be a `saturating_sub`?

pallets/pool-system/src/impls.rs
```diff
@@ -48,6 +48,7 @@ impl<T: Config> PoolInspect<T::AccountId, T::CurrencyId> for Pallet<T> {
 		// Get cached nav as calculating current nav would be too computationally expensive
 		let (nav, nav_last_updated) = T::NAV::nav(pool_id)?;
 		let total_assets = pool.reserve.total.saturating_add(nav);
+		// why not checked/ensure?
```
lemunozm, line 51: I'm afraid that the `saturating_ops` could lead to breaking some invariants. In this case, if, for some reason, the `nav` value is very big (i.e. an issue on the `NAV` side), `pool-system` will not be able to realize that and will successfully compute this when it shouldn't.

pallets/pool-system/src/lib.rs
offerijns: I think since Substrate 0.9.26 or similar, all extrinsics are transactional by default.

mustermeiszer: Didn't know that, but if it is the case then we are good. Otherwise, I would add a `#[transactional]` to the method.. 

We should have a `assert_noop` test in this case in the integration tests.

pallets/pool-system/src/tranches.rs
mustermeiszer: I am actually not sure. The create logic ensures we have always `tranche.seniority <= tranche.len()` and `tranche.seniority` greater/equal to seniority of previous tranche.

pallets/investments/src/lib.rs
```diff
@@ -459,7 +459,7 @@ pub mod pallet {
 		/// amount is less than the current order, the balance
 		/// will be transferred from the pool to the calling
 		/// account.
-		#[pallet::weight(80_000_000)]
```
branan, line 462: We definitely need to benchmark this. I looked into whether we needed to do it as part of the current release cycle and determined that the weights here are high enough that the necessarily lookup to fail them is below the hardcoded weight - and they'll always fail, since nobody has permission to invest in pools :sunglasses:

pallets/pool-system/src/lib.rs
```diff
@@ -835,7 +836,7 @@ pub mod pallet {
 					.expect("MaxTranches is u32. qed.");
 
 				// This kills the epoch info in storage.
-				// See: https://github.com/paritytech/substrate/blob/bea8f32e7807233ab53045fe8214427e0f136230/frame/support/src/storage/generator/map.rs#L269-L284
+				// See: https://github.com/paritytech/substrate/blob/bea8f32e7807233ab53045fe8214427e0f136230/frame/support/src/storage/generator/map.rs#L269-L284 // Broken link
```
branan, line 839: We should probably try to find a proper docs link for how `try_mutate` works on values with optional storage and  link that in, if we still feel we need this commented. I'd consider this "basic substrate behavior", but it's perhaps still surprising enough to a new dev that it's worth a comment, especially since we don't use this pattern often?

pallets/pool-system/src/impls.rs
branan: Agreed, this should be a checked rather than saturating add

pallets/pool-system/src/tranches.rs
lemunozm: If I have time, I make a draft to see how it looks and then choose if we want to merge it.

pallets/pool-system/src/tranches.rs
lemunozm: I agree that if this is not possible, we should return an error or ensure the expect will never fail (even if in the future some precondition in other places changes)

mustermeiszer: Agreed, if we have a specific test for this.

pallets/pool-system/src/impls.rs
mustermeiszer: Good catch. @lemunozm should I create clean up issues for this or do you wanna do this after we closed this review pr?

pallets/pool-system/src/lib.rs
mustermeiszer: I think we should remove this one. Was more me being surprised by the api ^^

pallets/investments/src/lib.rs
mustermeiszer: https://github.com/centrifuge/centrifuge-chain/issues/918 ^^

I think I used the weight for closing an epoch. Would be surprised if this is above that. So I think this is definetly needed but lower in priority. Wdyt?

pallets/pool-system/src/impls.rs
lemunozm: Like you want 😃 . If in some time there are no created issues from this. I can go through this review PR and create the required issues, to ensure the important emerging points are caught.

mustermeiszer: I Hope i will have time tomorrow for creating them.
Will leave a comment here if it is done.

mustermeiszer: Done.

mustermeiszer: https://github.com/centrifuge/centrifuge-chain/issues/1165

lemunozm:
I've seen in `pool_types.rs` that we are using a `PoolLocator` struct that is built using a `pool_id` in order to get the account of the pool. Nevertheless, the automatic implementation of `into_account_truncating()` is not using that `pool_id` internally, so no matter which `pool_id` is used, the account will be the same. Is this what we want, or is the intention to have one account per pool? 

I'm asking this because this struct forces a dependency `pallet_loan` -> `pallet_pool_system`, which maybe can be avoided if it's the same account for all pool ids.

@mustermeiszer @branan WDYT?

PR #1155, opened 2023-01-02
Loan refactor

offerijns:
# Description

Starts work on refactoring the loan storage and methods in the loans pallet, to make it more flexible.

### pallet-loans
- Separate `LoanType` into `ValuationMethod`, `RepaymentSchedule`, and `LoanRestrictions`
- Pass `RepaymentSchedule` on `create`
- Add `extend` extrinsic to update maturity date (instead of re-pricing)
- Simplify DCF valuation method calculation code

#### Before:
image

#### After:
image

[diagram sources](https://centrifuge.hackmd.io/CBuqTX2ETa-gK8K-V5ExmQ?both)

## Type of change

# How Has This Been Tested?

Tests are still TODOs.

# Checklist:

### DIFF
--- a/pallets/loans/src/functions.rs
+++ b/pallets/loans/src/functions.rs
@@ -95,7 +95,7 @@ impl<T: Config> Pallet<T> {
 		collateral_owner: T::AccountId,
 		collateral: AssetOf<T>,
 		schedule: RepaymentSchedule,
-	) -> Result<T::LoanId, sp_runtime::DispatchError> {
+	) -> Result<T::LoanId, DispatchError> {
 		// check if the nft belongs to owner
 		let Asset(collateral_class_id, instance_id) = collateral;
 		let owner = T::NonFungible::owner(&collateral_class_id.into(), &instance_id.into())
@@ -406,10 +406,7 @@ impl<T: Config> Pallet<T> {
 					.present_value(old_debt, &[], active_loan.last_updated)
 					.ok_or(Error::<T>::LoanPresentValueFailed)?;
 
-				let new_total_borrowed = active_loan
-					.total_borrowed
-					.checked_add(&amount)
-					.ok_or(ArithmeticError::Overflow)?;
+				let new_total_borrowed = active_loan.total_borrowed.ensure_add(amount)?;
 
 				// calculate new normalized debt with adjustment amount
 				let normalized_debt = T::InterestAccrual::adjust_normalized_debt(
@@ -489,7 +486,7 @@ impl<T: Config> Pallet<T> {
 			pool_id,
 			loan_id,
 			|active_loan| -> Result<T::Balance, DispatchError> {
-				let now: Moment = Self::now();
+				let now = Self::now();
 
 				// ensure current time is more than origination time
 				// this is mainly to deal with how we calculate debt while trying to repay
@@ -527,10 +524,7 @@ impl<T: Config> Pallet<T> {
 				// ensure amount is not more than current debt
 				let repay_amount = amount.min(current_debt);
 
-				let new_total_repaid = active_loan
-					.total_repaid
-					.checked_add(&repay_amount)
-					.ok_or(ArithmeticError::Overflow)?;
+				let new_total_repaid = active_loan.total_repaid.ensure_add(repay_amount)?;
 
 				// calculate new normalized debt with repaid amount
 				let normalized_debt = T::InterestAccrual::adjust_normalized_debt(
@@ -601,10 +595,7 @@ impl<T: Config> Pallet<T> {
 						&write_off_groups,
 					)?;
 
-					sum.checked_add(&present_value)
-						.ok_or(sp_runtime::DispatchError::Arithmetic(
-							ArithmeticError::Overflow,
-						))
+					Ok(sum.ensure_add(present_value)?)
 				},
 			)?;
 
@@ -722,7 +713,7 @@ impl<T: Config> Pallet<T> {
 					active_loan.last_updated,
 				)?;
 
-				let now: Moment = Self::now();
+				let now = Self::now();
 
 				// get old present value accounting for any write offs
 				let old_pv = active_loan
@@ -741,8 +732,7 @@ impl<T: Config> Pallet<T> {
 				let interest_rate_with_penalty = active_loan
 					.pricing
 					.interest_rate_per_sec
-					.checked_add(&write_off_penalty_rate)
-					.ok_or(ArithmeticError::Overflow)?;
+					.ensure_add(write_off_penalty_rate)?;
 				T::InterestAccrual::reference_rate(interest_rate_with_penalty)?;
 				active_loan.normalized_debt = T::InterestAccrual::renormalize_debt(
 					previous_interest_rate,
--- a/pallets/loans/src/functions.rs
+++ b/pallets/loans/src/functions.rs
@@ -15,16 +15,14 @@
 use cfg_traits::ops::ensure::EnsureAdd;
 use cfg_types::adjustments::Adjustment;
 use pallet_pool_system::pool_types::PoolLocator;
-use sp_runtime::{traits::BadOrigin, ArithmeticError};
+use sp_runtime::{
+	traits::{BadOrigin, BlockNumberProvider},
+	ArithmeticError,
+};
 
 use super::*;
 
 impl<T: Config> Pallet<T> {
-	/// returns the account_id of the loan pallet
-	pub fn account_id() -> T::AccountId {
-		T::LoansPalletId::get().into_account_truncating()
-	}
-
 	pub fn ensure_role(
 		pool_id: PoolIdOf<T>,
 		sender: T::AccountId,
@@ -49,19 +47,6 @@ impl<T: Config> Pallet<T> {
 		ensure!(actual_owner == expected_owner, Error::<T>::NotAssetOwner);
 
 		Ok(Asset(loan_class_id, loan_id))
-	}
-
-	// TODO: Move this to test-utils
-	#[cfg(any(test, feature = "runtime-benchmarks"))]
-	pub(crate) fn get_active_loan(
-		pool_id: PoolIdOf<T>,
-		loan_id: T::LoanId,
-	) -> Option<PricedLoanDetailsOf<T>> {
-		let active_loans = ActiveLoans::<T>::get(pool_id);
-
-		active_loans
-			.into_iter()
-			.find(|active_loan| active_loan.loan_id == loan_id)
 	}
 
 	pub(crate) fn try_mutate_active_loan<R, F>(
@@ -112,7 +97,7 @@ impl<T: Config> Pallet<T> {
 		schedule: RepaymentSchedule,
 	) -> Result<T::LoanId, sp_runtime::DispatchError> {
 		// check if the nft belongs to owner
-		let (collateral_class_id, instance_id) = collateral.destruct();
+		let Asset(collateral_class_id, instance_id) = collateral;
 		let owner = T::NonFungible::owner(&collateral_class_id.into(), &instance_id.into())
 			.ok_or(Error::<T>::NFTOwnerNotFound)?;
 		ensure!(owner == collateral_owner, Error::<T>::NotAssetOwner);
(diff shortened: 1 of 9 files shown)


pallets/loans/src/types.rs
```diff
@@ -140,18 +142,133 @@ pub enum NAVUpdateType {
...
+#[derive(Encode, Decode, Copy, Clone, TypeInfo)]
+pub struct LoanPricingInput<Rate, Balance> {
+	pub(crate) collateral_value: Balance,
+	pub(crate) interest_rate_per_year: Rate,
+	pub(crate) valuation_method: ValuationMethod<Rate, Balance>,
+	pub(crate) restrictions: LoanRestrictions<Rate>,
+}
+
+#[derive(Encode, Decode, Copy, Clone, TypeInfo)]
+pub struct LoanPricing<Rate, Balance> {
```
lemunozm, line 233: Is this equals to `LoanPricingInput` because it's expected to differ in the future?

lemunozm: I overlooked that! 👍🏻

lemunozm:
Another possible refactor/clean: 

The storage `LoanNftClassToPool` is only read once at `create_loan()` method to check if the asset class is registered. But a few lines below, we get that class id. Could we simply check the equality between `collateral_class_id` and `loan_class_id` and completely avoid the `LoanNftClassToPool` storage?

pallets/loans/src/types.rs
offerijns: I will add a comment for it, it's clearly easy to miss ;)

lemunozm:
Some of the new enums added to `types.rs` are with a `None` variant. Do we have any specs about what could be possible values for them? Is it something not specified that we need to discuss?

pallets/loans/src/valuation_method.rs
```diff
@@ -0,0 +1,187 @@
...
+	#[allow(dead_code)]
+	pub fn new() -> Self {
+		Self {}
+	}
+
+	/// calculates the present value of the credit line loan
+	/// https://centrifuge.hackmd.io/uJ3AXBUoQCijSIH9He-NxA#Present-value1
+	/// The debt = current outstanding debt * (1 - written off percentage)
+	pub fn present_value(&self, debt: Balance) -> Option<Balance> {
+		Some(debt)
```
lemunozm, line 141: For `OutstadingDebt`, is this something that should be implemented or is it the current implementation?

lemunozm:
There are some places in `pallet-loans` where we use `Moment` as something generic and other places where we use the defined `Moment` in `cfg_primitives`. I understand I could use everywhere the one from `cfg_primitives`, right?

pallets/loans/src/valuation_method.rs
offerijns: Yeah the `present_value` for the `OutstandingDebt` valuation method is just the debt, so this implementation is correct. `The debt = current outstanding debt * (1 - written off percentage)` should be happening before this method already and the `debt` param here should already be the output of this.

pallets/loans/src/types.rs
lemunozm: Sorry for opening this again, I was wondering if it could fit in to extract the `interest_per_second_sec` from the `LoanPricing/LoanPricingInput` and store it in the `PricedLoanDetails`. I say that because is not easy to know over which `LoanPricing` you should check things. i.e should I check if it is valid from `LoanPricing` or from `LoanPricingInput`? extracting the interest per second from it would allow only using one struct for both, as an input parameter and as a storage. 

So my real question, is. It has sense to move `interest_per_second_sec` to the `PricedLoanDetails` or semantically is it something tied to the `LoanPricing`?

offerijns: It's just that the `price` method always accepts these 4 inputs to update. So if we separate out `interest_rate_per_sec`, it just means that the `price` and possibly `reprice` method would always have the params `loan_id`, `pricing`, `interest_rate_per_sec`

pallets/loans/src/lib.rs
```diff
@@ -526,64 +528,63 @@ pub mod pallet {
...
-				interest_rate_per_sec,
-				loan_type,
+				pricing,
 			});
 
 			Ok(Some(T::WeightInfo::price(active_count)).into())
 		}
 
+		#[pallet::weight(<T as Config>::WeightInfo::price(T::MaxActiveLoansPerPool::get()))]
+		pub fn extend(
```
lemunozm, line 572: Can it be called over an unpriced loan? A just created loan

pallets/loans/src/functions.rs
```diff
@@ -245,12 +248,39 @@ impl<T: Config> Pallet<T> {
...
+				active_loan
+					.loan
+					.schedule
+					.update_maturity_date(new_maturity_date);
 
-		Ok((
-			ActiveLoans::<T>::get(pool_id).len().try_into().unwrap(),
-			interest_rate_per_sec,
-		))
+				// TODO: should update PV of loan
```
lemunozm, line 279: Should this be the same implementation as found in `reprice_loan()` method?

pallets/loans/src/functions.rs
```diff
@@ -460,7 +471,7 @@ impl<T: Config> Pallet<T> {
 				false => Ok(old_pv
 					.checked_sub(&new_pv)
 					.and_then(|negative_diff| nav.latest.checked_sub(&negative_diff))
-					.unwrap_or_else(Zero::zero)),
+					.unwrap_or_else(Zero::zero)), // Error instead?
```
lemunozm, line 474: I mean, repaying more than the total nav value should not be possible, I guess

pallets/loans/src/lib.rs
```diff
@@ -187,12 +186,12 @@ pub mod pallet {
 	#[pallet::storage]
 	#[pallet::getter(fn get_next_loan_id)]
 	pub(crate) type NextLoanId<T: Config> =
-		StorageMap<_, Blake2_128Concat, PoolIdOf<T>, u128, ValueQuery, OnNextLoanIdEmpty>;
+		StorageMap<_, Blake2_128Concat, PoolIdOf<T>, u128, ValueQuery, OnNextLoanIdEmpty>; // Could it be ValueStorage?
```
lemunozm, line 189: I was thinking if `LoanId` is unique, maybe we don't need to store the "next" loan Id per pool, is enough with only one `NextLoanId` used for all pools.

pallets/loans/src/types.rs
```diff
@@ -110,23 +113,8 @@ pub enum WriteOffAction<Rate> {
 	},
 }
 
-/// The data structure for storing loan status
-#[derive(Encode, Decode, Copy, Clone, PartialEq, TypeInfo)]
-#[cfg_attr(any(feature = "std", feature = "runtime-benchmarks"), derive(Debug))]
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub enum LoanStatus<BlockNumber> {
```
lemunozm, line 117: I've removed `LoanStatus` because it can be inferred from the storage.

pallets/loans/src/types.rs
```diff
@@ -110,23 +113,8 @@ pub enum WriteOffAction<Rate> {
...
-	// this when asset is locked and loan nft is created.
-	Created,
-	// this is when loan is in active state. Either underwriters or oracles can move loan to this state
-	// by providing information like discount rates etc.. to loan
-	Active,
-	// loan is closed and collateral nft is transferred back to borrower and loan nft is burned
-	Closed { closed_at: BlockNumber },
```
lemunozm, line 124: Because there is no longer `LoanStatus`, I would not know exactly where to store that `closed_at`. Do we really need it?

pallets/loans/src/types.rs
```diff
@@ -196,29 +320,34 @@ where
...
-			LoanType::CreditLine(cl) => cl.max_borrow_amount(debt),
-			LoanType::CreditLineWithMaturity(clm) => clm.max_borrow_amount(debt),
+	pub fn max_borrow_amount(&self, debt: Balance) -> Result<Balance, ArithmeticError> {
+		Ok(match self.pricing.restrictions.max_borrow_amount {
+			MaxBorrowAmount::UpToTotalBorrowed { advance_rate } => advance_rate
+				.ensure_mul_int(self.pricing.collateral_value)?
+				.checked_sub(&self.total_borrowed),
+			MaxBorrowAmount::UpToOutstandingDebt { advance_rate } => advance_rate
+				.ensure_mul_int(self.pricing.collateral_value)?
+				.checked_sub(&debt),
```
lemunozm, line 343: We are currently saturating to 0 if this fails, but should we? I mean, if the `debt` is > `max_borrow_amount` should be an error instead?

pallets/loans/src/functions.rs
```diff
@@ -143,32 +145,36 @@ impl<T: Config> Pallet<T> {
...
+	pub(crate) fn price_loan(
 		pool_id: PoolIdOf<T>,
 		loan_id: T::LoanId,
-		interest_rate_per_year: T::Rate,
-		loan_type: LoanType<T::Rate, T::Balance>,
-	) -> Result<(u32, T::Rate), DispatchError> {
+		pricing: LoanPricingInput<T::Rate, T::Balance>,
+	) -> Result<(), DispatchError> {
 		let now = Self::now();
-		ensure!(loan_type.is_valid(now), Error::<T>::LoanValueInvalid);
```
lemunozm, line 164: Moving the `maturity_date` to the creation properties of the loan, means that now somebody can price something with a passed maturity date. Should we check this? or the caller of this method should be aware of it?

lemunozm (commented):
I have the crate compiling now isolated with the refactoring. I left some comments about some things are not clear for me.

pallets/loans/src/functions.rs
offerijns: Total nav value is separate from this. The NAV represents the valuation of the loan, this can be different from the outstanding debt. E.g. a loan can be written down reducing the valuation (which is basically saying, there's an expectation that this loan won't be fully repaid), but the debt will then reduce the same.

So it should not be possible to repay more than the outstanding debt, but the present value does not matter for this.

pallets/loans/src/functions.rs
offerijns: Yes.

pallets/loans/src/types.rs
offerijns: The idea here was the following: right now, we are storing all pricing info for closed loans. This isn't really needed on chain, but it is needed for the UI. One future improvement we thought of was storing the `closed_at` and then using an archive node to get the pricing info at the block of closing, so we don't need to store it for closed loans after closing anymore.

pallets/loans/src/types.rs
offerijns: No that should be possible. If a max borrowed amount is 100 and someone borrows 100, then immediately 1 block later `debt > max_borrow_amount`, but this is expected. They just won't be able to borrow more, but this is not an error case.

pallets/loans/src/functions.rs
offerijns: Good point, we should check this in `create` now. Maybe we should have an `is_valid` method on the `RepaymentSchedule` trait and check this there?

pallets/loans/src/types.rs
wischli: Sorry for side tracking but I wanted to share my thoughts regarding the "future archive node idea": Unfortunately then everyone relies on this/these single source(s) of truth without a guarantee that the provided pricing info can definitely be trusted and cannot be manipulated. On the minimum, some sort of fingerprinting technique should be applied. Pinning to IPFS with some sort of time frame to challenge the data integrity (e.g. compare data and fingerprint on full node with replication on decentralized db). could be another solution.

pallets/loans/src/functions.rs
lemunozm: Ok, understand... So in case it should not be possible, we can return an error instead there, and it could be fine

pallets/loans/src/functions.rs
lemunozm: Ok, and what would happen in the next scenario:
1. create() -> with a correct maturity_date.
1. Times passes until now > maturity_date
1. Somebody tries to price() over the loan. Should we also check the maturity also here?

In fact, I understand that any create/price/reprice/extend should check this too.

pallets/loans/src/types.rs
```diff
@@ -138,19 +116,156 @@ pub enum NAVUpdateType {
...
+#[derive(Encode, Decode, Copy, Clone, PartialEq, RuntimeDebug, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum PriceRestrictions {
+	None,
+}
+
+#[derive(Encode, Decode, Copy, Clone, PartialEq, RuntimeDebug, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum MaxBorrowAmount<Rate> {
+	/// Collateral value - advance rate * total borrowed
```
lemunozm, line 211: From ceiling computation [here](https://centrifuge.hackmd.io/pYxf6rE-S2yYApZZpkGC1w), and what is done currently, I understand that the comment should be:
```suggestion
	/// Collateral value * advance rate - total borrowed
```

lemunozm: Can I have a double-check of this?

offerijns: Yeah you're right!

Issue #1023, opened 2022-10-10
Update benchmark of NAV calculation for #1021


branan:
After #1021, the NAV calculation will be `O(N*M)` where `N` is the number of loans and `M` is the number of rates. For each loan, there will be a linear scan over all the rates to find the appropriate one (that is the `O(M)` component), and then a bunch of other work which is `O(1)`.

The benchmarks will need to be updated to account for this change in the weight.

* Currently, one rate is generated for every loan. These values should be made independent
* Each loan will need to have a rate assigned such that the worst-case linear scan time is accounted for (that is, they should all have the "last" rate)
* The extrinsic `weight` attribute will need to be updated to account for a default value for M
* The final weight of the NAV will need to be calculated at the end based on the real value of M

Issue #1022, opened 2022-10-10
Update benchmarks of loan operations for #1021


branan:
#1021 will create a new `O(n)` component for all loan operations which read or write interest rates - the accrual pallet will now have to do a linear scan of the `BoundedVec<RateInfo>` for each of these queries.

The following changes need to happen:

* Add a new integer parameter to the benchmarks to represent the number of rates which will be in storage
* Populate rates with this number (at present this is likely tied to the number of active loans)
* Ensure the operation we're benchmarking uses a loan with the last rate added (for worst-case linear scan time)
* In the pallet code itself, update the extrinsic rate to set this new benchmark parameter to `pallet_interest_accrual::Config::MaxRateCount`
* <Optional> Have some way to get the actual number of rates that were iterated over and adjust the final weight of the extrinsic

PR #1157, opened 2023-01-04
Convert interest-accrual storage to a vector

### LINKED ISSUE #1021: Convert interest-accrual pallet to using a single storage item
Currently, `pallet-interest-accrual` stores rate information in a Substrate `StorageMap`. This is ideal for accessing single rates at a time (typically the case when pricing, writing-off, or closing a loan). However, when accessing many rates at a time (as used for actually accruing rates at block initialization, and as used for NAV calculation), it requires a _lot_ of storage reads/writes, along with the associated hashing of trie keys.

Instead, we should move to a single storage item which holds a bounded data structure. `BoundedVec` and `BoundedBTreeMap`are the obvious contenders. `BoundedVec` is better for accrual, since that is just a linear scan & update of every rates. `BoundedBTreeMap` _may_ be better for NAV calculations, once there are enough rates that a linear search over the vec of rates becomes expensive.

For now, we should use `BoundedVec`. We know it works better for rate accrual which has very tight timing needs by virtue of running on every block, and at least initially N will be small enough that NAVs won't see much, if any, benefit from using a BTree.

If the number of rates gets big enough that the time spent scanning a vec of rates is overriding the time save in iteration efficiency in accrual, conversion is fairly straightforward (Just iterate over everything and build the BTree)

branan:
# Description

This updates `pallet-interest-accrual` to store all rates in a vec, rather than in an array. This makes rate accrual and nav updates much faster, at the cost of slightly slower rate lookup for operations on individual loans (borrow, repay, write-off).

Fixes #1021 #1022 #1023

### pallet-interest-accrual

* Change storage for rates to a boundedvec
* Add migration for the above
* Expose vec of rates as a `RateCollection` for NAV updates
* Expose `MaxRateCount` for loan extrinsic weights

### pallet-loans

* Use new `RateCollection` API in NAV updates
* Update benchmarks to account for the new linear lookups in all extrinsics which touch interest-accrual

## Type of change

# How Has This Been Tested?

- [x] Migration has been manually verified in local development chain 
- [x] All existing tests for loans and interest accrual pass
- [x] All benchmark tests for loans and interest accrual pass

# Checklist:


branan:
@mustermeiszer The only thing left is getting the final weights for each extrinsic to use the _real_ count of rates, rather than the max. There's two ways to do this:

1. In the various interest-accrual methods, we return the count of how many items we needed to iterate over. This is most precise, but makes every invocation into interest-accrual quite a bit more expensive - and we need to then plumb this out as part of our results through all the loans impls as well.
2. We just have a way to ask for how many rates there are in total, and assume the worst-case that we had to iterate over them all.
3. We don't bother, and just always use the max possible weight as is currently implemented.

My inclination is to go for 2 for now. It's a compromise between the complexity of plumbing through more information in all our results and having the user pay for way more weight than is necessarily needed. Curious what your thoughts are, though.

pallets/interest-accrual/src/lib.rs
```diff
@@ -240,7 +259,9 @@ pub mod pallet {
 			interest_rate_per_sec: T::InterestRate,
 			normalized_debt: T::Balance,
 		) -> Result<T::Balance, DispatchError> {
-			Rate::<T>::get(interest_rate_per_sec)
+			Rates::<T>::get()
+				.into_iter()
+				.find(|rate| rate.interest_rate_per_sec == interest_rate_per_sec)
```
offerijns, line 264: Seems like this can be refactored into a `get_rate(interest_rate_per_sec: Rate) -> Option<RateDetailsOf<T>>` method, as the code is re-used multiple times.

runtime/altair/src/weights/pallet_loans.rs
```diff
@@ -1,116 +1,260 @@
...
-			.saturating_add(T::DbWeight::get().reads(3 as u64))
-			.saturating_add(T::DbWeight::get().reads((1 as u64).saturating_mul(n as u64)))
+	// Storage: Loans PoolWriteOffGroups (r:1 w:0)
+	// Storage: InterestAccrual Rates (r:1 w:0)
+	// Storage: Loans ActiveLoans (r:1 w:1)
+	// Storage: Timestamp Now (r:1 w:0)
+	// Storage: Loans PoolNAV (r:0 w:1)
+	/// The range of component `n` is `[1, 50]`.
+	/// The range of component `m` is `[1, 50]`.
+	fn update_nav(n: u32, m: u32, ) -> Weight {
```
mustermeiszer, line 250: This looks a lot nicer!! No more n in the reads. Nice!

pallets/interest-accrual/src/lib.rs
```diff
@@ -474,4 +517,28 @@ impl<T: Config> InterestAccrual<T::InterestRate, T::Balance, Adjustment<T::Balan
...
 	}
+
+	fn rates() -> Self::Rates {
+		RateVec(Rates::<T>::get())
+	}
+}
+
+pub struct RateVec<T: Config>(BoundedVec<RateDetailsOf<T>, T::MaxRateCount>);
+
+impl<T: Config> RateCollection<T::InterestRate, T::Balance, T::Balance> for RateVec<T> {
```
mustermeiszer, line 528: This forces every pallet that uses interest accrual to be tight-coupled with the interest accrual pallet. I am not sure, we want that. 

Couldn't we return another error that the pallet loans converts?

mustermeiszer (commented):
The benchmarks look really nice! I haven't calculated it, but not having number of loans reading the storage is good. 

There are a few minor bits regarding style, mostly having future changes in mind and encapsulating logic. 

One thing I really would like to change is binding the `impl<T: Config> RateCollection` with the `Config` of intrest accrual. This forces every pallet to tight-couple with interest accrual. 

Wdyt about changing that @branan?

libs/traits/src/lib.rs
```diff
@@ -205,7 +205,9 @@ pub trait CurrencyPrice<CurrencyId> {
 
 /// A trait that can be used to calculate interest accrual for debt
 pub trait InterestAccrual<InterestRate, Balance, Adjustment> {
+	type MaxRateCount: Get<u32>;
```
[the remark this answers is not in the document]
branan: It's used for weight calculations for things that consume the InterestAccrual impl.

The other option is to duplicate it as a config option into all pallets that consume InterestAccrual, which feels worse to me

pallets/interest-accrual/src/lib.rs
branan: I'm not sure what you mean? `RateVec` and its `RateCollection` impl are only ever exposed to other pallets through the `Rates` type in the impl of the `InterestAccrual` trait. It only relies on the existing parameters of that trait, and only returns `DispatchError` as its error type. It should be totally decoupled?

pallets/interest-accrual/src/lib.rs
branan: Yeah. I think in my mind this was a "one-liner", but its enough stuff, and common enough, to be worth giving a name :+1:

pallets/interest-accrual/src/migrations.rs
```diff
@@ -0,0 +1,70 @@
...
+
+			Rates::<T>::set(
+				rates
+					.try_into()
+					.expect("Input to this vector was already bounded"),
+			);
+			v1::RateCount::<T>::kill();
+			StorageVersion::<T>::set(Release::V2);
+
+			T::DbWeight::get().reads_writes(1 + count, 3)
```
wischli, line 64: Another question: Are you sure the drain can be counted as a single write? IMO, it should be `count` writes as well since the iterator [calls `frame_support::storage::unhashed::kill` on each item](https://github.com/paritytech/substrate/blob/master/frame/support/src/storage/generator/map.rs#L105).

pallets/interest-accrual/src/lib.rs
mustermeiszer: Yes, of course you are right. Not sure where my head was.

pallets/interest-accrual/src/migrations.rs
```diff
@@ -0,0 +1,70 @@
...
+	#[storage_alias]
+	pub type RateCount<T: Config> = StorageValue<Pallet<T>, u32, ValueQuery>;
+}
+
+pub mod v2 {
+	use super::*;
+
+	pub struct Migration<T>(sp_std::marker::PhantomData<T>);
+
+	impl<T: Config> OnRuntimeUpgrade for Migration<T> {
```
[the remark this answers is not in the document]
mustermeiszer: Yeah, I also would like to have a test for the migration as sanity check!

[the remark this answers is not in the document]
branan: Added

pallets/interest-accrual/src/migrations.rs
branan: Good catch, fixed

mustermeiszer:
@branan WDYT about adding type checking for yearly and secondly interest rates? I feel like it is way to easy to misuse the interface of accrual. 

Basically do the following: 

```rust
/// Will be also used in the public interfaces, where user
/// provide the APY. 
//
// NOTE: Not sure, if front-end even needs to adapt. Scale codec 
//             probably just serializes to Rate.
pub struct YearlyRate<T: FixedPointNumberExtension>(T);

impl<T: FixedPointNumberExtension> YearlyRate<T> {
       fn new(T::InterestRate) ->  Self;
}

pub struct SecondlyRate<T: FixedPointNumberExtension>(T);

impl<T: FixedPointNumberExtension> SecondlyRate<T> {
       fn new(T::InterestRate) ->  Self;
}

impl<T: Config> InterestAccrual<T::InterestRate, T::Balance, Adjustment<T::Balance>> for Pallet<T> {
        /*
        // TODO: Remove and use in `RateCollection`. Would this be possible and still work for benchmarks?
	type MaxRateCount = T::MaxRateCount;
	*/
	type NormalizedDebt = T::Balance;
	type Rates = RateVec<T>;

	fn current_debt(
		interest_rate_per_sec: SecondlyRate<T::InterestRate>,
		normalized_debt: Self::NormalizedDebt,
	) -> Result<T::Balance, DispatchError> {
		Pallet::<T>::get_current_debt(interest_rate_per_sec, normalized_debt)
	}

	fn previous_debt(
		interest_rate_per_sec: SecondlyRate<T::InterestRate>,
		normalized_debt: Self::NormalizedDebt,
		when: Moment,
	) -> Result<T::Balance, DispatchError> {
		Pallet::<T>::get_previous_debt(interest_rate_per_sec, normalized_debt, when)
	}

	fn adjust_normalized_debt(
		interest_rate_per_sec: SecondlyRate<T::InterestRate>
		normalized_debt: Self::NormalizedDebt,
		adjustment: Adjustment<T::Balance>,
	) -> Result<Self::NormalizedDebt, DispatchError> {
		Pallet::<T>::do_adjust_normalized_debt(interest_rate_per_sec, normalized_debt, adjustment)
	}

	fn renormalize_debt(
		old_interest_rate: SecondlyRate<T::InterestRate>,
		new_interest_rate: SecondlyRate<T::InterestRate>,
		normalized_debt: Self::NormalizedDebt,
	) -> Result<Self::NormalizedDebt, DispatchError> {
		Pallet::<T>::do_renormalize_debt(old_interest_rate, new_interest_rate, normalized_debt)
	}

	fn reference_yearly_rate(
		interest_rate_per_year: YearlyRate<T::InterestRate>
	) -> Result<T::InterestRate, DispatchError> {
		Pallet::<T>::reference_yearly_interest_rate(interest_rate_per_year)
	}

        /* 
        // TODO: I would remove this one here. Everything should go through the yearly interface. Unreferencing through the secondly.
	fn reference_rate(interest_rate_per_sec: SecondlyRate<T::InterestRate>) -> Result<(), DispatchError> {
		Pallet::<T>::reference_interest_rate(interest_rate_per_sec)
	}
        */

	fn unreference_rate(interest_rate_per_sec: SecondlyRate<T::InterestRate>) -> Result<(), DispatchError> {
		Pallet::<T>::unreference_interest_rate(interest_rate_per_sec)
	}

        /*
        // TODO: See changes proposed in the WriteOff-Group issue/pr
	fn convert_additive_rate_to_per_sec(
		interest_rate_per_year: T::InterestRate,
	) -> Result<T::InterestRate, DispatchError> {
		Pallet::<T>::validate_rate(interest_rate_per_year)?;
		let interest_rate_per_sec = interest_rate_per_year
			.checked_div(&T::InterestRate::saturating_from_integer(SECONDS_PER_YEAR))
			.ok_or(ArithmeticError::Underflow)?;
		Ok(interest_rate_per_sec)
	}
        */

	fn rates() -> Self::Rates {
		RateVec(Rates::<T>::get())
	}
}

pub struct RateVec<T: Config>(BoundedVec<RateDetailsOf<T>, T::MaxRateCount>);

impl<T: Config> RateCollection<T::InterestRate, T::Balance, T::Balance> for RateVec<T> {
        type MaxRateCount = T::MaxRateCount;       

	fn current_debt(
		&self,
		interest_rate_per_sec: SecondlyRate<T::InterestRate>,
		normalized_debt: T::Balance,
	) -> Result<T::Balance, DispatchError> {
		self.0
			.iter()
			.find(|rate| rate.interest_rate_per_sec == interest_rate_per_sec)
			.ok_or(Error::<T>::NoSuchRate)
			.and_then(|rate| {
				Pallet::<T>::calculate_debt(normalized_debt, rate.accumulated_rate)
					.ok_or(Error::<T>::DebtCalculationFailed)
			})
			.map_err(Into::into)
	}
}
```

branan:
I think adding type safety here is good. I'd probably use associated types on the `InterestAccrual` like has been done for `NormalizedDebt` - it avoids some boilerplate. I also think we have three rate types - a yearly rate, a secondly rate, and an "additive secondly rate" for penalties.

I'd like to throw together a separate ticket for this just to keep track of alternate design ideas, and because it feels out-of-scope for this PR

mustermeiszer:
Sounds good!

pallets/interest-accrual/src/lib.rs
```diff
@@ -204,35 +205,56 @@ pub mod pallet {
...
+
+			// reads: timestamp, last updated, rates vec
+			// writes: last updated, rates vec
+			let mut weight = T::DbWeight::get().reads_writes(3, 2);
+
+			let rates = Rates::<T>::get();
+			let rates: Vec<_> = rates
+				.into_iter()
+				.filter_map(|rate| {
+					weight = weight.saturating_add(T::Weights::calculate_accumulated_rate(bits));
```
wischli, line 229: Could use `saturating_accrue`
```suggestion
					weight.saturating_accrue(T::Weights::calculate_accumulated_rate(bits));
```

pallets/interest-accrual/src/lib.rs
```diff
@@ -204,35 +205,56 @@ pub mod pallet {
...
-
-			let db_weight = T::DbWeight::get().reads_writes(2, 1);
-
-			let db_weight_accumulated = T::DbWeight::get().reads_writes(1, 1)
-				+ T::Weights::calculate_accumulated_rate(bits);
+				})
+				.collect();
 
-			db_weight + db_weight_accumulated.saturating_mul(count)
+			Rates::<T>::set(
```
wischli, line 252: I agree that in this case using `get` and `set` provides better readability than wrapping the `filter_map` inside `try_mutate`.

pallets/interest-accrual/src/lib.rs
```diff
@@ -241,47 +263,40 @@ pub mod pallet {
...
-				.ok_or(Error::<T>::NoSuchRate)
-				.and_then(|rate| {
-					Self::calculate_debt(normalized_debt, rate.accumulated_rate)
-						.ok_or(Error::<T>::DebtCalculationFailed)
-				})
-				.map_err(Into::into)
+			let rate = Self::get_rate(interest_rate_per_sec)?;
+			let debt = Self::calculate_debt(normalized_debt, rate.accumulated_rate)
+				.ok_or(Error::<T>::DebtCalculationFailed)?;
+			Ok(debt)
```
wischli, line 269: Could: End a bit shorter
```suggestion
			Self::calculate_debt(normalized_debt, rate.accumulated_rate)
				.ok_or(Error::<T>::DebtCalculationFailed.into())
```

pallets/interest-accrual/src/lib.rs
```diff
@@ -241,47 +263,40 @@ pub mod pallet {
...
+			if when > now {
+				return Err(Error::<T>::NotInPast.into());
 			}
+			let delta = now - when;
+			let rate_adjustment = checked_pow(interest_rate_per_sec, delta as usize)
+				.ok_or(ArithmeticError::Overflow)?;
+			let past_rate = rate
+				.accumulated_rate
+				.checked_div(&rate_adjustment)
+				.ok_or(ArithmeticError::Underflow)?;
```
wischli, line 288: Should: Use `ensure_div`

```suggestion
				.ensure_div(&rate_adjustment)?;
```

pallets/interest-accrual/src/lib.rs
```diff
@@ -370,40 +387,52 @@ pub mod pallet {
...
-
-					*rate = Some(RateDetailsOf::<T> {
-						accumulated_rate: One::one(),
-						reference_count: 1,
-					});
-					RateCount::<T>::set(rate_count.checked_add(1).ok_or(Error::<T>::TooManyRates)?);
+			Rates::<T>::try_mutate(|rates| {
+				for rate in rates.iter_mut() {
+					if rate.interest_rate_per_sec == interest_rate_per_sec {
+						rate.reference_count += 1;
```
wischli, line 393: Should: Use `ensure_add_assign`

```suggestion
						rate.reference_count.ensure_add_assign(1)?;
```

pallets/interest-accrual/src/lib.rs
```diff
@@ -370,40 +387,52 @@ pub mod pallet {
...
 		}
 
+		pub fn get_rate(
+			interest_rate_per_sec: T::InterestRate,
+		) -> Result<RateDetailsOf<T>, DispatchError> {
+			let rate = Rates::<T>::get()
+				.into_iter()
+				.find(|rate| rate.interest_rate_per_sec == interest_rate_per_sec)
+				.ok_or(Error::<T>::NoSuchRate)?;
+			Ok(rate)
```
wischli, line 433: Could: Reduce by one line 😇

```suggestion
			 Rates::<T>::get()
				.into_iter()
				.find(|rate| rate.interest_rate_per_sec == interest_rate_per_sec)
				.ok_or(Error::<T>::NoSuchRate.into())
```

pallets/loans/src/benchmarking.rs
```diff
@@ -310,6 +302,37 @@ fn pool_account<T: pallet_pool_system::Config>(pool_id: T::PoolId) -> T::Account
...
+	<T as LoanConfig>::Rate: From<Rate>,
+	<T as LoanConfig>::LoanId: From<u32>,
+	<T as InterestAccrualConfig>::InterestRate: From<Rate>,
+	<T as pallet_balances::Config>::Balance: From<u128>,
+	<T as pallet_uniques::Config>::CollectionId: From<u64>,
+	<T as pallet_pool_system::Config>::PoolId: Into<u64> + IsType<PoolIdOf<T>>,
+{
+	let mut collateral = None;
+	for idx in 1..rates {
+		let rate = Rate::saturating_from_rational(idx, 5000).into();
```
wischli, line 321: Since `5000` seems to be the default rate quotient for the benchmarks as it is used all three times, I think we should declare this a constant.

runtime/altair/src/lib.rs
```diff
@@ -1501,6 +1501,7 @@ pub type Executive = frame_executive::Executive<
 		pallet_preimage::migration::v1::Migration<Runtime>,
 		pallet_democracy::migrations::v1::Migration<Runtime>,
 		pallet_scheduler::migration::v3::MigrateToV4<Runtime>,
+		pallet_interest_accrual::migrations::v2::Migration<Runtime>,
```
wischli, line 1504: Q: Is there a process of removing deprecated migrations like all the ones above?

wischli (approved):
LGTM! I have a few minor optional nits.

Since there are currently no interest rates on Altair, I have no worries about the migration.

libs/traits/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I think since this type does not appear in the trait interface, it should be very well documented to teach the user how to use it and why it's needed.

pallets/loans/src/lib.rs
```diff
@@ -526,7 +526,7 @@ pub mod pallet {
 		/// Loan is accrued before transferring the amount to reserve.
 		/// If the repaying amount is more than current debt, only current debt is transferred.
 		/// Amount of token will be transferred from owner to Pool reserve.
-		#[pallet::weight(<T as Config>::WeightInfo::repay(T::MaxActiveLoansPerPool::get()))]
+		#[pallet::weight(<T as Config>::WeightInfo::repay(T::MaxActiveLoansPerPool::get(), MaxRateCountOf::<T>::get()))]
```
lemunozm, line 529: It feels like it's a lot of knowledge for the loans pallet to handle the `MaxRateCount` coming from a trait in a method that does not make use explicitly anything related to this count. The only way to know you must put this weight here is by looking deeper into how `InterestAccrual` implementation you've configured to see how it works.

In fact, the guilty method that forces you to add this is `interest_accrual::Pallet::get_rate()`, which is used by the implementation of `InterestAccrual::current_debt()` used in `price()`. Seems like a very long path to realizing about this.

I understand that it's needed for the benchmarks, but not sure if the complexity it adds here deserves to be measured it nowdays. Probably the Substrate framework should offer some kind of utility for this kind of things 🥲

pallets/interest-accrual/src/lib.rs
```diff
@@ -370,40 +387,52 @@ pub mod pallet {
...
+			Rates::<T>::try_mutate(|rates| {
+				let idx = rates
+					.iter()
+					.enumerate()
+					.find(|(_, rate)| rate.interest_rate_per_sec == interest_rate_per_sec)
+					.ok_or(Error::<T>::NoSuchRate)?
+					.0;
+				rates[idx].reference_count = rates[idx].reference_count.saturating_sub(1);
+				if rates[idx].reference_count == 0 {
+					rates.remove(idx);
```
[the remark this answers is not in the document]
lemunozm: Additionally, there is a `swap_remove()` that removes in `O(1)` that I think can be used here.

pallets/interest-accrual/src/lib.rs
```diff
@@ -241,47 +263,40 @@ pub mod pallet {
...
-					.checked_div(&rate_adjustment)
-					.ok_or(ArithmeticError::Underflow)?;
-				let debt = Self::calculate_debt(normalized_debt, past_rate)
-					.ok_or(Error::<T>::DebtCalculationFailed)?;
-				Ok(debt)
-			} else {
-				Err(Error::<T>::NoSuchRate.into())
+			let rate = Self::get_rate(interest_rate_per_sec)?;
+			let now = LastUpdated::<T>::get();
+			if when > now {
```
lemunozm, line 279: I think `when` could be the same as `now` if the "previous debt" was computed in the same block (same `now` value)

pallets/interest-accrual/src/lib.rs
```diff
@@ -241,47 +263,40 @@ pub mod pallet {
...
+				return Err(Error::<T>::NotInPast.into());
 			}
+			let delta = now - when;
+			let rate_adjustment = checked_pow(interest_rate_per_sec, delta as usize)
```
lemunozm, line 283: `Moment` maybe could be `u128` in the future.
```suggestion
			let rate_adjustment = checked_pow(interest_rate_per_sec, delta.ensure_into()?)
```

pallets/interest-accrual/src/lib.rs
lemunozm: I agree too, but here I recognize that by using `mutate()` we can avoid calling `collect()` and then avoid allocating a new big vector in memory. 🤔

lemunozm:
One more thought, if rates go from 0% to 100%, what about using `PerThing` instead of a more generic fixed point?

offerijns:
> One more thought, if rates go from 0% to 100%, what about using PerThing instead of a more generic fixed point?

It's not completely unthinkable that at some point we want to support rates larger than 100% or even negative rates (negative yielding bonds are definitely a thing, although becoming extinct quickly).

pallets/interest-accrual/src/lib.rs
branan: I went with the extra line so I could let the question mark operator do my error conversion, avoiding a manual `.into()`.

pallets/interest-accrual/src/lib.rs
branan: Yup, `when` can be ` now`, but the math still works out.

pallets/interest-accrual/src/lib.rs
branan: I don't think we can avoid the call to `collect` - part of what's happening with the `filter_map` is that any rates which overflow get removed (which will later cause extrinsics using those rates to fail). This isn't the _best_ error handling situation. Since there is no way to return an error from `on_initialize`, it's better to just nuke any bad data rather than allowing it to be used later on and further corrupt results

runtime/altair/src/lib.rs
branan: I think it is part of our release process now to double-check migrations? /cc @mikiquantum

pallets/interest-accrual/src/lib.rs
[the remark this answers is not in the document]
branan: `swap_remove` is a good call, since this is unsorted :+1:

branan:
`PerThing` is also kind of a pain to convert to a fixed-point for math later on (it needs to be in `1.xxxx` form for the `pow()` etc)

pallets/loans/src/lib.rs
branan: I agree it's a lot of knowledge. If we come up with a better way to abstract this I would love to implement it

pallets/interest-accrual/src/lib.rs
mustermeiszer: Agreeing with Branan here!"

pallets/interest-accrual/src/lib.rs
mustermeiszer: Does the `checked_pow` has an early return path if `delta == 0`?

mustermeiszer (approved):
I would say let's merge this and think about future improvements of the NAV in another PR. 

Thanks @branan, increases our bottle neck quite significantly!

pallets/interest-accrual/src/lib.rs
lemunozm: You're right! I didn't pay attention to the filter 👍🏻, (although maybe `Vec::drain()` could do the trick?)

pallets/interest-accrual/src/lib.rs
lemunozm: > but the math still works out

Although the result should not change, right? 
I mean, as a loan user, do I want to get an error here if in the same block I call `borrow()` and then `repay()/reprice()/modify()`? Because my previous `last_update` will be used later in this function, but the value will have not changed. I do not know if I explain it well.

> Does the checked_pow has an early return path if delta == 0?
Yes (checked)

PR #1160, opened 2023-01-05
Connectors v0.1.0

NunoAlexandre:
## Connectors v1

The latest revision of this PR has been tested on the Moonbase Alpha testnet and has passed all the tests where all the connectors messages (`AddPool`, `AddTranche`, `UpdateMember`, `UpdateTokenPrice`, and `Transfer`) were successfully sent from the Centrifuge development runtime through XCM to Moonbase, where they were forwarded through `ethereum_xcm` to the Moonbase EVM down to our `ConnectorXcmRouter` smart contract.

The changes which close the v1 are:

0. Several fixes in the encoding of messages, where numbers need to be encoded as big-endian to be correctly decoded on the Solidity side
1.  Redesign of the `Domain` and `DomainAddress` types to best map the mental model we have for connectors
2. Add `price` field to the `AddTranche` message with the latest tranche price
3. More++ unit and integration tests (the latter for covering all the setup requirements needed to send the respective messages)

---

#### To Do

- [x] Add `initial_price` to `Message::AddTranche` (https://github.com/centrifuge/connectors/pull/16#discussion_r1084201476)
- [x] Fix transfer message encoding test
- [x] Set `gas_limit` to max or have user pass it down when submitting a call
- [x] Fix/Test `update_member` derivation of DomainAddress
- [x] Test all messages' requirements

---


pallets/connectors/src/message.rs
```diff
@@ -207,6 +207,24 @@ mod tests {
 			assert_eq!(msg.encode(), vec![0]);
 		}
 
+		#[test]
+		fn encoding_domain_evm() {
+			let evm_domain = Domain::EVM(u64::MAX);
```
offerijns, line 212: Maybe you can add a few more sample tests for common chain IDs, e.g. Ethereum Mainnet (1), Avalanche C-cHain (43114), Moonbeam (1284). That way the connectors solidity tests can easily implement the same tests and ensure it all matches.

NunoAlexandre: @offerijns how do we want to handle the different environments within Polkadot? Meaning, `Domain::Parachain(ParachianId::Moonbeam)`, following to your examples, should have different values if we are running this pallet on Moobase Alpha, Rococo, Kusama, or Polkadot because "Moonbeam" has a different parachain id on each of those chains.

My idea was to just index the parachains by a fixed enumerated index (0, 1, 2, etc) so that it's not chain dependant. That's why, as the test below now shows, `Domain::Parachain(ParachainId::Moonbeam)` encodes to `0x0100`.

What do you think?

NunoAlexandre: Actually just realized that we need to make the domain encode to a fixed-size output. Now the `Evm` and the `Parachain` variants encode to different sizes.

offerijns: > Actually just realized that we need to make the domain encode to a fixed-size output. Now the Evm and the Parachain variants encode to different lengths.

I don't think it really needs to be fixed length. In the Solidity code, we can check if the first byte represents an EVM or Polkadot domain. Then we either parse 8 more bytes in the EVM case or skip to the next parameter in the Polkadot case. This will lead to more efficient messages than zero padding it.

> My idea was to just index the parachains by a fixed enumerated index (0, 1, 2, etc) so that it's not chain dependant. That's why, as the test below now shows, Domain::Parachain(ParachainId::Moonbeam) encodes to 0x0100.

I think that makes sense!

NunoAlexandre: Perfect, thanks @offerijns :)

NunoAlexandre (commented):
@offerijns on this block in `update_member`, we give the `TrancheInvestor` permission to the `DomainAddress` that we are adding as a member.

``` rust
T::Permission::add(
	PermissionScope::Pool(pool_id),
	address.into_account_truncating(),
	Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, valid_until)),
)?;
```

The issue with this current code is that `address` is of type `DomainAddress` it seems that the account derivation `address.into_account_truncating(),` doesn't actually take the inner address in consideration when deriving the account.

I noticed this by having first called `upate_member` for a given Moonbase evm address, which failed on the EVM side due to gas limit issues, and when I then wanted to retry that same call it failed because the permissions were already given; and changing the evm address I was specifying didn't fix it. I had to create a new pool all together to test this call.

Just want to confirm that the desired behaviour is to add that permission to an account that's derived from the domain + address pair?

We can also consider only adding the permission if it's not there yet; it adds one read but I don't think this should fail if the permission was already given. What do you think?

offerijns:
> Just want to confirm that the desired behaviour is to add that permission to an account that's derived from the domain + address pair?

Yeah that was the intention, bookkeeping which addresses were added as investors on other chains.

> We can also consider only adding the permission if it's not there yet; it adds one read but I don't think this should fail if the permission was already given. What do you think?

I think the idea here was that `update_member` would only need to be called once to add the permission. I think it's fine to add one read, but what's the advantage of doing this?

NunoAlexandre:
> Yeah that was the intention, bookkeeping which addresses were added as investors on other chains.

Thanks for confirming!

> I think the idea here was that `update_member` would only need to be called once to add the permission. I think it's fine to add one read, but what's the advantage of doing this?

The advantage would be to not have the `update_member` fail shall we need to call it again with the same params, which may be necessary for example if the call fails on the EVM side due to gas limits etc

offerijns:
> The advantage would be to not have the update_member fail shall we need to call it again with the same params, which may be necessary for example if the call fails on the EVM side due to gas limits etc

Ah yeah of course, makes sense then!

pallets/connectors/src/lib.rs
```diff
@@ -486,3 +547,28 @@ pub mod pallet {
...
+mod tests {
+	use cfg_primitives::AccountId;
+	use sp_runtime::traits::AccountIdConversion;
+
+	use super::DomainAddress;
+
+	#[test]
+	fn domain_address_account_derivation() {
+		assert_eq!(
+			account_from(DomainAddress::EVM(1284, [9; 20])),
```
NunoAlexandre, line 561: I just added these tests to verify that the account derivation was taking the inner address into account. We probably can drop these but leave it up to the reviewers.

pallets/connectors/src/lib.rs
```diff
@@ -45,23 +45,52 @@ pub enum ParachainId {
...
+
+/// Custom encode implementation for the `Domain` type
+/// Domain is encoded as a 9-byte long bytearray, where:
+/// Byte 0 - Flags the Domain variant, i.e, either Domain::Centrifuge (0) or Domain::EVM (1)
+/// Byte 1-9 - Encodes the domain id if applicable;
+///		- For Domain::Centrifuge there's no id, so we append 8 zeros
+///		- For Domain::EVM, encode the respective chain id (8 bytes) as little endian
+///
+/// We need to impl this as a custom encode to not overlap with the `#[derive(Encode, Decode)]`
+/// that are necessary for `Domain` to be used in the storage as a key.
```
NunoAlexandre, line 75: NOTE: Explained here why I've introduced this trait. I'd be supper happy to receive suggestions on a better way to do this 🙏

NunoAlexandre (commented):
Some notes for the reviewers

pallets/connectors/src/message.rs
```diff
@@ -48,19 +49,14 @@ where
 	Transfer {
 		pool_id: PoolId,
 		tranche_id: TrancheId,
-		domain: Domain,
-		destination: Address,
+		address: Address,
 		amount: Balance,
+		domain: Domain,
```
offerijns, line 54: This is a minor detail, but semantically I don't think it makes sense that `domain` is last. This should at least be before `address`, wdyt?

I would possibly also prefix these to make it clearer, so
```rust
Transfer {
		pool_id: PoolId,
		tranche_id: TrancheId,
		destination_domain: Domain,
		destination_address: Address,
		amount: Balance,
	}
```

NunoAlexandre: I think that makes sense, I thought about this only after we set this order in Solidity 🙈  Let's get this right now 👍

pallets/connectors/src/lib.rs
lemunozm: Thinking about this, if this new encoding is totally representative of the `Domain`, and it does not lose any information, why not use this method as a custom `Encode` and use it also for computing the storage's key?, The "downside" of this is that we need to ensure that now the `Decode` is the inverse of this encoding.

NunoAlexandre: > The "downside" of this is that we need to ensure that now the Decode is the inverse of this encoding.

Yeah, this. Plus our custom Encoding implementation is probably not as efficient as the default one which _can_ become an expensive matter when it comes to handling storage keys. Wdyt?

pallets/connectors/src/lib.rs
```diff
@@ -329,22 +387,30 @@ pub mod pallet {
...
+				domain_address.into_account_truncating(),
 				Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, valid_until)),
-			)?;
+			) {
+				T::Permission::add(
+					PermissionScope::Pool(pool_id),
+					domain_address.into_account_truncating(),
+					Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, valid_until)),
+				)?;
+			}
```
lemunozm, line 403: An idea for this. Since both parameters used are the same on `has()` and `add()`, and seems it can be a recurrent pattern, why not add a new method `add_idempotent()` (or similar) to the trait `Permissions` that does exactly this? The method could have a default implementation using only `has()` and `add()` so there is no implementation on the permission pallet.

NunoAlexandre: I actually wanted to suggest that we don't fail with `RoleAlreadyGiven` in the Permissions pallet with a permission was already set since it's not _really_ an error in my view. Wdyt?

pallets/connectors/src/lib.rs
```diff
@@ -434,7 +500,8 @@ pub mod pallet {
 				OriginKind::SovereignAccount,
 				TransactWeights {
 					// Specify a conservative max weight
-					transact_required_weight_at_most: 8_000_000_000,
+					transact_required_weight_at_most: xcm_domain.max_gas_limit * 25_000
+						+ 100_000_000,
```
lemunozm, line 504: Out of curiosity, where `25_000` and `100_000_000` come from?

pallets/connectors/src/message.rs
```diff
@@ -118,6 +110,10 @@ impl<
 				message.append(&mut token_name.encode());
 				message.append(&mut token_symbol.encode());
 
+				let mut encoded_price = price.encode();
+				encoded_price.reverse();
+				message.append(&mut encoded_price);
```
lemunozm, line 115: What about implementing a custom encoding for a `Price` type and here simply using `.encode()`?

lemunozm: Or something similar for types that needs a particular encoding, using a wrapper type. Maybe not now, but for the feature, if it grows, it could be easy to make mistakes

lemunozm (approved):
LGTM! I left some minor comments

pallets/connectors/src/lib.rs
lemunozm: I like too, but not sure if there are other places that expect not to add a `Role` if it's already been added. But it makes sense to have this idempotent 👍🏻

pallets/connectors/src/lib.rs
lemunozm: mmm... for this type of size, sure it's not a problem. But if we use this pattern for bigger types it could be (?). 
Another approach is to use a Wrapper type with custom encoding. Something like: `DomainSolidity`, which knows how to encode for solidity, and is used only when it requires to be sent to solidity.

pallets/connectors/src/message.rs
NunoAlexandre: > What about implementing a custom encoding for a Price type and here simply using .encode()?

May be a good idea. But I guess that would mess up with decoding the current data on storage? What I tried doing, actually, was having a bound on the type, `Into<u128>`, to which I could convert to and then call `.to_be_bytes()`. The problem is that this approach needs to cascade down to the `pallet_connectors` Config types (`Balance`, `Rate`) as well as the `PoolInspect::PoolId` and `PoolInspect::TrancheId`; that's why I didn't go after this approach.

> Or something similar for types that needs a particular encoding, using a wrapper type. Maybe not now, but for the feature, if it grows, it could be easy to make mistakes

I'm curious! can you expand?

lemunozm: Yeah, I was thinking in something like this:
```rust
#[derive(Decode, Clone, PartialEq, TypeInfo, Encode, Decode)]
pub enum Message<Domain, PoolId, TrancheId, Balance, Rate> {
    Invalid,
    AddPool {
        pool_id: XcmPoolId,
    },
    AddTranche {
        pool_id: XcmPoolId,
        tranche_id: XcmTrancheId,
        token_name: XcmName<TOKEN_NAME_SIZE>,
        token_symbol: XcmName<TOKEN_SYMBOL_SIZE>,
    },
    //...
}
// NOTE: No longer need an custom `encode()` method for `Message`.
```
And then, each XcmPoolId has a custom encoding for its special encoding:
```rust
struct XcmPoolId(PoolId);
impl Encode for XcmPoolId {
    fn encode(&self) -> Vec<u8> {
        let mut encoded_pool_id = self.0.encode();
        encoded_pool_id.reverse();
        encoded_pool_id
    }
}
// Same for price
```

NunoAlexandre: Thanks for expanding on this! Note that this are not "xcm" types, this is encoding requirements specific to Connectors. And if that's ok for you I'd rather leave this as is and consider revisit it when we work on Connectors v2. How does that sound?

pallets/connectors/src/lib.rs
NunoAlexandre: This is a formula given to us by Moonbeam. Would you like us to add a comment with that info?

pallets/connectors/src/message.rs
NunoAlexandre: Done 👍

pallets/connectors/src/message.rs
NunoAlexandre: @lemunozm I just created a simple `to_be` function that does this. This seems to be the cleanest, least invasive, and more straightforward way to improve this a bit to wrap up this PR.

pallets/connectors/src/lib.rs
lemunozm: I think it doesn't does damage 😃, or even `fn moombeam_formula(xcm_domain) -> Weight`. Up to you!

pallets/connectors/src/message.rs
lemunozm: I like much more the `to_be()` factorization 🙌🏻 . 
Following your approach, I was wondering if something like this could work, following your approach:
```rust
trait EncodeBigEndian: Encode {
    fn encode_be(&self) -> Vec<u8> {
        let encoded = self.encode();
        encoded.reverse();
        encoded
    }
}
impl<T: Encode> EncodeBigEndian for T {}
```
Usage:
```rust
price.encode_be()
```
Anyways, I'm ok as it is. 👍🏻

pallets/connectors/src/lib.rs
NunoAlexandre: What advantage do you see in having a `DomainSolidity` wrapper type VS the current `ConnectorEncode` trait? Also, note that the custom encoding is not necessarily "Solidity" dependant, it's the standard format defined for Connectors, thus the `ConnectorsEncode` trait name.

pallets/connectors/src/lib.rs
NunoAlexandre: Created a separate issue for that: https://github.com/centrifuge/centrifuge-chain/issues/1204

pallets/connectors/src/lib.rs
lemunozm: > it's the standard format defined for Connectors

Gotcha!

> What advantage do you see in having a DomainSolidity wrapper type VS the current ConnectorEncode trait?

Instead of having two kinds of encodings for a type, you would have two types with the same kind of encoding. I feel this is like "more" type-safety. Depending on where you pass the type, you protect/force the implementation by using the correct encoding call. No mistakes regarding which encoding I should use.

pallets/connectors/src/message.rs
```diff
@@ -273,27 +282,71 @@ mod tests {
...
+					"1231231231231231231231231231231231231231231231231231231231231231",
+				)
+				.expect(""),
+				valid_until: 1706260138,
+			};
+			let encoded = msg.encode();
+
+			let input = "040000000000000002811acd5b3f17c06841c7e41e9e04cb1b12312312312312312312312312312312312312312312312312312312312312310000000065b376aa";
+			let expected = <[u8; 65]>::from_hex(input).expect("Decoding failed");
+			assert_eq!(hex::encode(encoded), hex::encode(expected));
```
wischli, line 305: Why do we the expected hex into bytes and back into hex? Below we just compare `hex::encode(encoded)` with `input`.

runtime/integration-tests/src/xcm/development/tests/connectors.rs
```diff
@@ -141,15 +141,137 @@ fn add_tranche() {
...
+				Domain::EVM(1284),
+			),
+			pallet_connectors::Error::<development_runtime::Runtime>::MissingTranchePrice
+		);
+
+		Loans::update_nav(RuntimeOrigin::signed(ALICE.into()), pool_id.clone())
+			.expect("Should update nav");
+
+		// Verify it now works
+		assert_ok!(Connectors::update_token_price(
```
wischli, line 269: Question: Shouldn't we compare the token price before and after the invocation of `update_token_price` similar to the above permissions checks?

wischli (commented):
LGTM! I have a few minor comments

runtime/integration-tests/src/xcm/development/tests/connectors.rs
NunoAlexandre: This won't actually test the effect of having the connector message reach the smart contract, we are only testing here all the logic and requirements that need to be in place on the cent-chain to make these calls successfully.

pallets/connectors/src/message.rs
NunoAlexandre: Thanks for the catch. This was legacy code, so most tests were inconsistent in that regard. All standardized now 👍

pallets/connectors/src/lib.rs
```diff
@@ -485,3 +552,28 @@ pub mod pallet {
...
+	use sp_runtime::traits::AccountIdConversion;
+
+	use super::DomainAddress;
+
+	#[test]
+	fn domain_address_account_derivation() {
+		assert_eq!(
+			account_from(DomainAddress::EVM(1284, [9; 20])),
+			account_from(DomainAddress::EVM(1284, [9; 20])),
+		);
```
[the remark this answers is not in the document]
NunoAlexandre: Yeah good one, added :) thanks

pallets/connectors/src/lib.rs
```diff
@@ -71,14 +102,34 @@ impl<Domain> TypeId for DomainLocator<Domain> {
...
 
-#[derive(Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
+#[derive(Encode, Decode, Clone, PartialEq, TypeInfo)]
 #[cfg_attr(feature = "std", derive(Debug))]
-pub struct DomainAddress<Domain> {
-	pub domain: Domain,
-	pub address: [u8; 32],
+pub enum DomainAddress {
+	Centrifuge([u8; 32]),
+	EVM(EVMChainId, [u8; 20]),
```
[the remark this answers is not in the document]
NunoAlexandre: Left a doc comment but not sure what more to add given that Ethereum addresses are 20-bytes long, not sure what else to say. What do you suggest? 👀

pallets/connectors/src/lib.rs
NunoAlexandre: Thanks for all the input, @lemunozm. I decided to just implement `Encode` and `Decode` for `Domain` and drop `ConnectorEncode` all together. The custom impls are now covered by tests so we are good :)

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
wischli: Completely suffices!

wischli (approved):
LGTM!

PR #1173, opened 2023-01-16
tests: continue tranches tests pt. 2

wischli:
# Description

 * Adds more tests for tranches in the Pools pallet, continues on work from https://github.com/centrifuge/centrifuge-chain/pull/907 and https://github.com/centrifuge/centrifuge-chain/pull/1046
   * We could merge #1046 before reviewing this PR to reduce review reduncancy as #1046 has been soft reviewed by @gruberb and me at least 
* Fixes # https://github.com/centrifuge/centrifuge-chain/issues/678

## Changes and Descriptions

 Adds tests for tranches in the Pools pallet

## Type of change

- Backfill tests and fixes/improvements for discovered issues.

# How Has This Been Tested?

Run unit tests, ticket is for addition of tranches tests.

# Checklist:

### DIFF
--- a/pallets/pool-system/src/tranches.rs
+++ b/pallets/pool-system/src/tranches.rs
@@ -1525,18 +1525,15 @@ pub mod test {
 		buffer_in_perc: Option<u64>,
 		seniority: Seniority,
 	) -> TTranche {
-		let interest_rate_per_sec = if let Some(rate) = interest_rate_in_perc {
-			Rate::saturating_from_rational(rate, 100) / Rate::saturating_from_integer(SECS_PER_YEAR)
-				+ One::one()
-		} else {
-			Rate::one() / Rate::saturating_from_integer(SECS_PER_YEAR) + One::one()
-		};
+		let interest_rate_per_sec = interest_rate_in_perc
+			.map(|rate| Rate::saturating_from_rational(rate, 100))
+			.unwrap_or(Rate::one())
+			/ Rate::saturating_from_integer(SECS_PER_YEAR)
+			+ One::one();
 
-		let min_risk_buffer = if let Some(buffer) = buffer_in_perc {
-			Perquintill::from_rational(buffer, 100)
-		} else {
-			Perquintill::zero()
-		};
+		let min_risk_buffer = buffer_in_perc
+			.map(|buffer| Perquintill::from_rational(buffer, 100))
+			.unwrap_or(Perquintill::zero());
 
 		TTranche {
 			tranche_type: TrancheType::NonResidual {
--- a/pallets/pool-system/src/tranches.rs
+++ b/pallets/pool-system/src/tranches.rs
@@ -112,6 +112,11 @@ where
 	///
 	pub fn valid_next_tranche(&self, next: &TrancheType<Rate>) -> bool {
 		match (self, next) {
+			// NOTE: If we change this: I.e. allowing multiple residual
+			//       tranches. One MUST be cautious to check ALL places
+			//       were we assume a single residual tranche.
+			//
+			//       E.g. Tranches.non_residual_tranches()
 			(TrancheType::Residual, TrancheType::Residual) => false,
 			(TrancheType::Residual, TrancheType::NonResidual { .. }) => true,
 			(TrancheType::NonResidual { .. }, TrancheType::Residual) => false,
@@ -609,7 +614,7 @@ where
 		F: FnMut(&Tranche<Balance, Rate, Weight, TrancheCurrency>) -> Result<R, DispatchError>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		for tranche in self.tranches.iter().rev() {
+		for tranche in self.non_residual_top_slice() {
 			let r = f(tranche)?;
 			res.push(r)
 		}
@@ -621,7 +626,7 @@ where
 		F: FnMut(&mut Tranche<Balance, Rate, Weight, TrancheCurrency>) -> Result<R, DispatchError>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		for tranche in self.tranches.iter_mut().rev() {
+		for tranche in self.non_residual_top_slice_mut() {
 			let r = f(tranche)?;
 			res.push(r)
 		}
@@ -638,7 +643,7 @@ where
 		I: IntoIterator<Item = W>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		let iter = self.tranches.iter().rev().zip(with.into_iter());
+		let iter = self.non_residual_top_slice().iter().zip(with.into_iter());
 
 		for (tranche, w) in iter {
 			let r = f(tranche, w)?;
@@ -661,7 +666,10 @@ where
 		I: IntoIterator<Item = W>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		let iter = self.tranches.iter_mut().rev().zip(with.into_iter());
+		let iter = self
+			.non_residual_top_slice_mut()
+			.iter_mut()
+			.zip(with.into_iter());
 
 		for (tranche, w) in iter {
 			let r = f(tranche, w)?;
@@ -680,7 +688,7 @@ where
 		F: FnMut(&Tranche<Balance, Rate, Weight, TrancheCurrency>) -> Result<R, DispatchError>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		for tranche in self.tranches.iter() {
+		for tranche in self.residual_top_slice() {
 			let r = f(tranche)?;
 			res.push(r)
 		}
@@ -692,7 +700,7 @@ where
 		F: FnMut(&mut Tranche<Balance, Rate, Weight, TrancheCurrency>) -> Result<R, DispatchError>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		for tranche in self.tranches.iter_mut() {
+		for tranche in self.residual_top_slice_mut() {
 			let r = f(tranche)?;
 			res.push(r)
 		}
@@ -709,7 +717,7 @@ where
 		I: IntoIterator<Item = W>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		let iter = self.tranches.iter().zip(with.into_iter());
+		let iter = self.residual_top_slice().iter().zip(with.into_iter());
 
 		for (tranche, w) in iter {
 			let r = f(tranche, w)?;
@@ -732,8 +740,10 @@ where
 		I: IntoIterator<Item = W>,
 	{
 		let mut res = Vec::with_capacity(self.tranches.len());
-		// TODO: Would be nice to error out when with is larger than tranches...
-		let iter = self.tranches.iter_mut().zip(with.into_iter());
+		let iter = self
+			.residual_top_slice_mut()
+			.iter_mut()
+			.zip(with.into_iter());
 
 		for (tranche, w) in iter {
 			let r = f(tranche, w)?;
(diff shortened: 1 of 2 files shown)


pallets/pool-system/src/tranches.rs
```diff
@@ -728,14 +817,26 @@ where
...
+			};
 		}
 
-		Ok(res)
+		match (tranche_slice.next(), with_iter.next()) {
+			(None, None) => Ok(res),
+			_ => Err(DispatchError::Other(
+				"Iterable contains more elements than Tranches tranche count",
+			)),
+		}
```
mustermeiszer, line 839: Maybe encapsulate this? It is used in more than 3 places.

pallets/pool-system/src/tranches.rs
```diff
@@ -1051,41 +1174,41 @@ where
...
-			Some(tail)
-		} else {
-			None
-		}
+		self.tranches
+			.deref()
+			.as_slice()
+			.split_first()
+			.map(|(_head, tail)| tail)
+			.filter(|tail| !tail.len().is_zero())
```
mustermeiszer, line 1182: We are a little inconsistent here, if I am not mistaken. 
For the `struct Tranches` we have a better check than `split_first`. Maybe adapt the `EpochExecutionTranches` to behave the same.

pallets/pool-system/src/tranches.rs
```diff
@@ -1051,41 +1174,41 @@ where
...
 	) -> Option<&mut EpochExecutionTranche<Balance, BalanceRatio, Weight, TrancheCurrency>> {
-		if let Some((head, _tail)) = self.tranches.as_mut().split_first_mut() {
-			Some(head)
-		} else {
-			None
-		}
+		self.tranches
+			.get_mut(..)
+			.and_then(|slice| slice.split_first_mut())
+			.map(|(head, _tail)| head)
```
mustermeiszer, line 1211: See above

pallets/pool-system/src/tranches.rs
```diff
@@ -1394,12 +1577,2336 @@ where
...
-	#[test]
-	fn reverse_works_for_both_tranches() {}
+	use super::*;
+	use crate::mock::MaxTranches;
 
-	#[test]
-	fn accrue_overflows_safely() {}
+	// NOTE: We currently expose types in runtime-common. As we do not want
+	//       this dependecy in our pallets, we generate the types manually here.
+	//       Not sure, if we should rather allow dev-dependency to runtime-common.
```
mustermeiszer, line 1588: This can be removed and we can use all types from the primivites directly.

pallets/pool-system/src/tranches.rs
```diff
@@ -1394,12 +1577,2336 @@ where
...
+		#[test]
+		fn create_tranche_works() {
+			let mut tranches = default_tranches_with_seniority();
+			let tranche_id: TrancheId = [
+				103u8, 57, 22, 242, 127, 45, 18, 102, 173, 154, 105, 163, 156, 150, 75, 194,
+			];
+			let int_per_sec = Rate::saturating_from_integer(SECS_PER_YEAR);
+			let min_risk_buffer = Perquintill::from_rational(4u64, 5);
+
+			// Create tranch with explicit seniority
```
mustermeiszer, line 2062:
```suggestion
			// Create tranche with explicit seniority
```

pallets/pool-system/src/tranches.rs
```diff
@@ -1394,12 +1577,2336 @@ where
...
+		}
+
+		struct TokenSymLen;
+		impl Get<u32> for TokenSymLen {
+			fn get() -> u32 {
+				8u32
+			}
+		}
+
+		/// Replace should work if interest is lower than tranche w/ lower index ("next").
```
mustermeiszer, line 2157:
```suggestion
		// Replace should work if interest is lower than tranche w/ lower index ("next").
```

pallets/pool-system/src/tranches.rs
```diff
@@ -1394,12 +1577,2336 @@ where
...
+				Err(DispatchError::Other(
+					"Tranches struct contains more tranches than combining iterable"
+				))
+			);
+		}
+
+		mod calculate_prices {
+			use super::*;
+
+			// FIXME: To reviewer, can this somehow be derived? In practice, we only need the `total_issuance` but `calculate_price` requires the trait.
```
mustermeiszer, line 2583: I don't think so. We will in the near future add an issuance field to the tranches. As we need to ensure that ONLY the pool is able to increase/decrease those values, to protect us against minting bugs at the bridge levels.

pallets/pool-system/src/tranches.rs
```diff
@@ -1394,12 +1577,2336 @@ where
...
+						default_tranches_with_issuance()
+							.calculate_prices::<Rate, TTokens, TrancheCurrency>(
+								1_100_000_000,
+								i * SECS_PER_YEAR
+							),
+						"APR != APY after {} years",
+						i
+					);
+				}
+			}
```
mustermeiszer, line 2850: No changes with higher numbers? 

Can we try with explicit rounding through `FixedPointNumberExtension`

mustermeiszer (changes requested):
Looks really good already. 

What we need to change is the removal of tranches. Did not clearly spec this, my fault! Otherwise, just minor stuff.

pallets/pool-system/src/tranches.rs
```diff
@@ -1240,14 +1389,26 @@ where
...
-		for (tranche, w) in iter {
-			let r = f(tranche, w)?;
-			res.push(r);
+		let mut tranche_slice = self.residual_top_slice().iter();
+		let mut with_iter = with.into_iter();
+
+		for _ in 0..self.tranches.len() {
+			match (tranche_slice.next(), with_iter.next()) {
+				(Some(tranche), Some(w)) => res.push(f(tranche, w)?),
+				_ => {
```
lemunozm, line 1398: Cool pattern! 🚀

pallets/pool-system/src/tranches.rs
wischli: The comment was off, even without rounding the error % is around `10e-6`% or lower, see a7fdfb4d0ebb981f359b57268c032fc20c48d302 where I added rounding with a max precision of `10e-6`%.

pallets/pool-system/src/tranches.rs
wischli: Definitely! I ended up adding two generic builders for all `combine_*` functions in `Tranches` and `EpocExecutionTranches`, see 25e9a8fd1f5ddc332c09f18fecc66a849735f9a6

pallets/pool-system/src/tranches.rs
wischli: Props go to @thea-leake :)

pallets/pool-system/src/tranches.rs
```diff
@@ -1394,12 +1577,2336 @@ where
...
+
+			#[test]
+			/// If amount of assets is zwero zero, all price rates should be one.
+			fn calculate_prices_no_assets_works() {
+				assert_eq!(
+					default_tranches()
+						.calculate_prices::<_, TTokens, TrancheCurrency>(0, SECS_PER_YEAR),
+					Ok(vec![Rate::one(), Rate::one(), Rate::one(),])
+				);
+			}
```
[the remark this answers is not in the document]
wischli: a73defdb28030bf93bf216fc12be3301a77b7c81

pallets/pool-system/src/tranches.rs
```diff
@@ -1394,12 +1577,2336 @@ where
...
+				tranches_pre_removal
+					.get_tranche(TrancheLoc::Index(2))
+					.unwrap(),
+				tranches.get_tranche(TrancheLoc::Index(0)).unwrap()
+			);
+
+			// remove last remaining tranche
+			assert_ok!(tranches.remove(0));
+			assert!(tranches.num_tranches().is_zero());
+			assert!(tranches.remove(0).is_err());
```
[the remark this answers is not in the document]
wischli: Good catch! Fixed in 31d91c62f5c2c6bd25795e4608df59053055d714

pallets/pool-system/src/tranches.rs
mustermeiszer: Ah you are right. 

Mmh. Want to get rid of the epoch tranches anyways. But let's keep it for now.

wischli: Looking forward to getting rid of them! The code duplication strikes me.

pallets/pool-system/src/tranches.rs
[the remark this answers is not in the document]
wischli: > While looking through this I realized it should NOT be possible to remove a tranche if
> 
> * balance is != 0
> * tranche is residual tranche

Note: Needs to be applied to replacing as well.

pallets/pool-system/src/tranches.rs
mustermeiszer: Nice!

pallets/pool-system/src/tranches.rs
mustermeiszer: Yes, this is fine .

pallets/pool-system/src/tranches.rs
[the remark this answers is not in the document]
wischli: > > While looking through this I realized it should NOT be possible to remove a tranche if
> > 
> > * balance is != 0
> > * tranche is residual tranche
> 
> Note: Needs to be applied to replacing as well.

Answering myself: This is actually also covered by https://github.com/centrifuge/centrifuge-chain/commit/31d91c62f5c2c6bd25795e4608df59053055d714 since `replace(at)` calls `remove(at)` ensuring the invariance.

pallets/pool-system/src/tranches.rs
```diff
@@ -1392,14 +1456,2731 @@ where
...
+				))
+			);
+			// verify unchanged from default val
+			assert_eq!(
+				tranches
+					.get_tranche(TrancheLoc::Index(2))
+					.unwrap()
+					.seniority,
+				0
+			);
```
mustermeiszer, line 2179: I would rather test for equality of the whole tranche here, than the equality of the `seniority`

pallets/pool-system/src/tranches.rs
```diff
@@ -1392,14 +1456,2731 @@ where
...
+
+			// If amount of assets is zero, all price rates should be one.
+			#[test]
+			fn no_assets_works() {
+				assert_eq!(
+					default_tranches()
+						.calculate_prices::<_, TTokens, TrancheCurrency>(0, SECS_PER_YEAR),
+					Ok(vec![Rate::one(), Rate::one(), Rate::one(),])
+				);
+			}
```
mustermeiszer, line 2668: this is a bug! 

The `calculate_prices` does not work correctly. It should
* Return `Rate::zero()`, if `total_assets == 0 && total_issuance != 0`
* Return `Rate::one()`, if `total_assets == 0 && total_issuance == 0`

mustermeiszer (changes requested):
Thanks a lot @thea-leake and @wischli! The coverage looks great! 

I am just flagging this as a change request as I just noticed that we have a serious bug in the `caluclate_prices`. The wanted behaviour is noted in the comments. 

If this is fixed, we can merge this up!

pallets/pool-system/src/tranches.rs
wischli: > * Return `Rate::one()`, if `total_assets == 0 && total_issuance == 0`

Shouldn't it be `Rate:one()` anytime a token's total issuance is zero? Else we will always try to divide by zero independent of the tranche type due to returning.

```rust
Ok(BalanceRatio::ensure_from_rational(tranche_value, total_issuance)?)
```

wischli: See 11c963823b2704453e53378339e88cba916a6935

pallets/pool-system/src/tranches.rs
wischli: Apart from the `seniority` and the `tranche_type` it does not make sense to compare more since [the rest is boilerplate via `create_tranche`](https://github.com/centrifuge/centrifuge-chain/blob/tests%2Fcontinue-tranches-tests/pallets/pool-system/src/tranches.rs#L558):

Fixed in 9c5baf6e30868bfc6a011e68b09a866c233bacad

pallets/pool-system/src/tranches.rs
mustermeiszer: Yes. You are right. Sorry. If the issuance is 0 then the price should be 1.

mustermeiszer (dismissed):
Thanks a lot to both of you!

mustermeiszer (approved):
Re-approving after fixing the tests for good hopefully!

PR #1181, opened 2023-01-18
Update relay chain version for tests

mikiquantum:
# Description

Local relay chain should always match local parachain polkadot version.


mustermeiszer (approved):
This slipped through, right? In the 32 upgrade?

mikiquantum:
@mustermeiszer yep

PR #1182, opened 2023-01-19
Add Demo Spec

### LINKED ISSUE #1176: Create new Chain Spec for Demo environment

* [x] Feature

### Description

Create new set of keys for sudo + operational accounts

### Research/based on

[Which codebases, pallets or designs did you base your design on]

[Does it improve readability, performance? Will it introduce new complexity?]

[Are migrations needed, structural changes around testing etc?]

mikiquantum:
Closes #1176 

# Description

- Adds Demo Chain Spec with a newly generated set of keys. 
- Simplifies the Development Spec to only use 1 collator

# Impact
With these new keys, we will need to make sure that the provision environment script injects the session keys accordingly. @gpmayorga

### DIFF
--- a/src/chain_spec.rs
+++ b/src/chain_spec.rs
@@ -625,6 +625,40 @@ pub fn charcoal_local(para_id: ParaId) -> AltairChainSpec {
 	)
 }
 
+pub fn demo(para_id: ParaId) -> DevelopmentChainSpec {
+	let mut properties = Properties::new();
+	properties.insert("tokenSymbol".into(), "DEMO".into());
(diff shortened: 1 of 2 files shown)


wischli (approved):
LGTM! I suppose we want to add the chainspec to the `/res` directory as well?

mustermeiszer (approved):
LGTM! 

One question: Where do the endowed accounts come from?

mikiquantum:
@wischli I would do that for infinite living environments, but I am not certain that demo is such yet. 
@mustermeiszer I generated those accounts manually, they are available internally.

PR #1192, opened 2023-02-01
Rewards: fix initial block value

### LINKED ISSUE #1191: Liquidity rewards: Fix first `BlockNumber` value

* Addressing a bug

### Description

The `liquidity-pallet` does not work well in the initial stage when there is no value set on `EndOfEpoch`. The default `BlockNumber` returned by this storage is current, so the condition to update the epoch data is never dispatched.

This was skipped from UTs, because we are manually setting the `BlockNumber`.

### Research/based on

- Issue found in the development chain.

lemunozm:
# Description

Fix the initial block issue.

Fixes #1191

## Changes and Descriptions

- Remove the `System::block_number()` usage as a default storage value, because Substrate does not insert the value in the storage when the first value is fetched. This caused that each `get()` call returned the same current block number, and `on_initialize()` logic was never dispatched.
- Simplified generic usage.
- Minor docs additions.
- Adapt tests to catch the issue.

## Type of change

# How Has This Been Tested?

```sh
```

# Checklist:

### DIFF
--- a/pallets/liquidity-rewards/src/lib.rs
+++ b/pallets/liquidity-rewards/src/lib.rs
@@ -47,6 +47,7 @@ pub use cfg_traits::{
 use frame_support::{
 	pallet_prelude::*,
 	traits::tokens::{AssetId, Balance},
+	DefaultNoBound,
 };
 pub use frame_support::{
 	storage::{bounded_btree_map::BoundedBTreeMap, transactional},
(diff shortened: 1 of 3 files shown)


pallets/liquidity-rewards/src/lib.rs
```diff
@@ -55,92 +56,39 @@ pub use frame_support::{
...
-	FixedPointOperand,
-};
+use sp_runtime::{traits::Zero, FixedPointOperand};
 use sp_std::mem;
 use weights::WeightInfo;
 
-/// Type that contains the timestamp of an epoch.
-#[derive(Encode, Decode, TypeInfo, MaxEncodedLen, RuntimeDebug)]
-#[cfg_attr(test, derive(PartialEq))]
-pub struct EpochTimestamp<BlockNumber>(BlockNumber);
```
lemunozm, line 68: We no longer need a custom type because now the default value doesn't need to do any computation.

pallets/liquidity-rewards/src/lib.rs
```diff
@@ -55,92 +56,39 @@ pub use frame_support::{
...
-pub struct EpochData<BlockNumber, Balance, GroupId, Weight, MaxGroups>
-where
-	MaxGroups: Get<u32>,
-	GroupId: Ord,
-{
-	duration: BlockNumber,
-	reward: Balance,
-	weights: BoundedBTreeMap<GroupId, Weight, MaxGroups>,
+#[derive(Encode, Decode, TypeInfo, MaxEncodedLen, RuntimeDebugNoBound)]
+#[scale_info(skip_type_params(T))]
```
lemunozm, line 65: A few days ago, I learned about this. It allows to use the `T: Config` generic in any type without notification to the type info that it exists, reducing a lot of generic boilerplate. For types like this, it works pretty fine.

pallets/liquidity-rewards/src/lib.rs
```diff
@@ -299,9 +226,9 @@ pub mod pallet {
 						epoch_data.reward = changes.reward.unwrap_or(epoch_data.reward);
 						epoch_data.duration = changes.duration.unwrap_or(epoch_data.duration);
 
-						let ends_on = ends_on.ensure_add(epoch_data.duration)?;
+						let ends_on = ends_on.max(current_block).ensure_add(epoch_data.duration)?;
```
lemunozm, line 229: Because now, the initial `ends_on` not uses the current block number, we need to ensure our last `ends_on` is not too old. Suppose the following scenario:
- Block 1000 => Upgrade chain with `pallet-liquidiy-rewards`, initial epoch duration: 20
- Block 1001 => `on_initialize()` is called, next ends_on is 20 + 20 = 40 (**but we wanted 1021**)
- Block 1002 => `on_initialize()` is called, next ends_on is 40 + 20 = 60
- And so on until we reach the current chain block number

lemunozm (commented):
Explanation notes below

pallets/liquidity-rewards/src/lib.rs
```diff
@@ -206,54 +154,33 @@ pub mod pallet {
 	#[pallet::generate_store(pub(super) trait Store)]
 	pub struct Pallet<T>(_);
 
+	/// Contains the timestamp in blocks when the current epochs finalized.
```
mikiquantum, line 157:
```suggestion
	/// Contains the timestamp in blocks when the current epoch is finalized.
```

pallets/liquidity-rewards/src/lib.rs
```diff
@@ -267,7 +193,7 @@ pub mod pallet {
 	#[pallet::hooks]
 	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
 		fn on_initialize(current_block: T::BlockNumber) -> Weight {
-			let ends_on = EndOfEpoch::<T>::get().0;
+			let ends_on = EndOfEpoch::<T>::get();
```
mikiquantum, line 196: I guess in the initial call it will return the BlockNumber default of its primitive type, which should be 0, so the following comparison should be happy

mikiquantum (approved):
Thanks for the quick fix. I added a few informational comments. Keep in mind, that if we want to deploy this again, we will need to reset the development chain so it starts from empty storage so this logic can be triggered. Since the development runtime version was not incremented

pallets/liquidity-rewards/src/lib.rs
```diff
@@ -206,54 +154,33 @@ pub mod pallet {
...
-			T::BlockNumber,
-			T::Balance,
-			T::GroupId,
-			T::CurrencyId,
-			T::Weight,
-			T::MaxChangesPerEpoch,
-		>,
-		ValueQuery,
-	>;
+	pub(super) type NextEpochChanges<T: Config> = StorageValue<_, EpochChanges<T>, ValueQuery>;
```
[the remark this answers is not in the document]
lemunozm: Not because `#[scale_info(skip_type_params(T))]` skip that 😃

pallets/liquidity-rewards/src/lib.rs
lemunozm: Yes, default `BlockNumber` is  `zero()`

pallets/liquidity-rewards/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Also, `Config` (or any generic bound) contains only information at compile time. It never increments the size of the objects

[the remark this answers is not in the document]
wischli: Im a bit late to the party but wanted to share that even though the current design is leaner, Substrate actually recommends to use the former, e.g., to specifically declare all required generic parameters instead of pulling from the Config. I suppose one argument is that its a bit more opaque. Couldn't quickly find the PR with the comment about that. The skip macro resulted just from the inconvenience that the compiler else throws when pulling from the Config. 

Definitely no need to change this now but good to keep it in mind for the future and other parts of the code we touch. Or do you disagree @lemunozm ?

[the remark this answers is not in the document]
lemunozm: If you finally find that thread I would be happy to read it! Thanks for sharing this.

From my understanding of this, I think that if the type is (or should) be agnostic of the pallet, it shouldn't make use of `<T: Config>`, because it comes with the whole pallet dependency. But sometimes, the type is closely related to the pallet, and it contains a lot of types specified in it. Not using `<T: Config>` in that case adds a lot of boilerplate that seems like it's not really needed and difficult refactors and readability. Anyways, I do not have a strong thought about where to put the frontier between one decision or the other.

P.S: I took as a reference the Substrate's `nomination-pools` pallet where high-level types use `<T: Config>` and low-level types define the generics individually.

PR #1198, opened 2023-02-10
feat: impl block rewards

### LINKED ISSUE #1000: Rewards: pallet-block-rewards implementation

### Description

Implements the `pallet-block-rewards`. This implies:
- Implementation
- Testing
- Benchmarks

### Research/based on
- See [specs](https://centrifuge.hackmd.io/9T9zAZaTRD6ZcHN-Gd_lzw?view)

wischli:
# Description

Implements [`pallet-block-rewards`](https://centrifuge.hackmd.io/9T9zAZaTRD6ZcHN-Gd_lzw#pallet-block-rewards-spec).

Fixes #1000

## Changes and Descriptions

- [ ] Create `pallet_block_rewards` based on the [rewards spec](https://centrifuge.hackmd.io/9T9zAZaTRD6ZcHN-Gd_lzw?view)
- [ ] Add to Centrifuge runtime: Rewards should be minted
- [ ] Add to Develop runtime: Rewards should be minted
- [ ] Add to Altair runtime: Rewards should be pulled from Treasury 
  * Probably requires adding an interface for handling the token origin in `<<pallet_rewards::Pallet::<T, I>> as GroupRewards>::reward_group` which defaults to `T::Currency::mint_into`: Add interface Inflation<Currency> exposing fn inflate(amount) -> DispatchResult
  * Something similar for remainder handling
> Configurable minting method, that is, configure the pallet with an `Option<AccountId>` or similar to give the account used for rewards or None to mint it
- [ ] Benchmark `pallet_block_rewards`
- [ ] Benchmark `pallet_collator_selection`
- [ ] Benchmark `pallet_session`
- [ ] Add init migration
- [ ] TBD: Add runtime API call to amount of claimable rewards

## Type of change

- This change requires a documentation update

# How Has This Been Tested?

```sh
cargo test -p pallet-block-rewards --features runtime-benchmarks
PARA_CHAIN_SPEC=development ./scripts/init.sh benchmark pallet-block-rewards
```

# Checklist:


pallets/rewards/src/lib.rs
```diff
@@ -230,11 +235,22 @@ pub mod pallet {
...
+						reward,
+						true,
+					)?;
+				} else {
+					T::Currency::mint_into(
+						T::RewardCurrency::get(),
+						&T::PalletId::get().into_account_truncating(),
+						reward,
+					)?;
+				}
```
lemunozm, line 253: I like it as it is 🚀.

Because this is something independent from block rewards, maybe we could create an isolated PR with this change (with some test) and merge it before. Wdyt?

pallets/rewards/src/lib.rs
```diff
@@ -119,6 +119,11 @@ pub mod pallet {
 		/// Specify the internal reward mechanism used by this pallet.
 		/// Check available mechanisms at [`mechanism`] module.
 		type RewardMechanism: RewardMechanism;
+
+		/// Type used to identify the income stream for rewards.
+		/// By setting this `None`, the rewards will be minted.
+		#[pallet::constant]
+		type RewardSource: Get<Option<frame_support::PalletId>>;
```
lemunozm, line 126: I also have a proposal for this part. I think we should consider that different domains could use different mint methods. So maybe I will add something like:
`Get<BoundedBTreeMap<Domain, frame_support::PalletId>>;`
If the entry does not exist, then it mints; if an account exists for that domain then use that account.

lemunozm: I'm curious why you've chosen a `PalletId` instead of a more generic `AccountId`

wischli: I like your proposal! 

> I'm curious why you've chosen a `PalletId` instead of a more generic `AccountId`

To be honest, for now just because it compiled quicker it was clear to me, that this needed a conversation. I suppose we are favouring `AccountId` here?

lemunozm: I think an `AccountId` is more generic, and it wraps all possible `PalletId` values. Although right now, I'm not able to see a use case where `AccountId` works but `PalletId` does not. Maybe a CFG reserve that does not belong to a pallet itself?

runtime/development/src/lib.rs
```diff
@@ -1678,6 +1691,44 @@ impl pallet_liquidity_rewards::Config for Runtime {
...
 }
 
+frame_support::parameter_types! {
+	pub const BlockRewardsDomain: RewardDomain = RewardDomain::Block;
+	pub const EpocDurationBlockRewards: BlockNumber = 2 * Period::get();
+
+	// TODO: Convert to common consts
+	#[derive(scale_info::TypeInfo, Debug, PartialEq, Clone)]
+	// TODO: Ensure uniqueness, currently abbreviates "blockrewards/collator"
+	pub const CollatorCurrencyId: u32 = u32::from_be_bytes(*b"br/c");
```
lemunozm, line 1701: Not sure if we want this, but there is a `CurrencyId` enum at `libs/types/src/tokens.rs` with all currencies used in `centrifuge-chain`. Maybe we can add this currency there in order to track all our currencies. 
Maybe @mustermeiszer can clarify this.

runtime/development/src/lib.rs
```diff
@@ -1678,6 +1691,44 @@ impl pallet_liquidity_rewards::Config for Runtime {
...
+
+	pub const RemainingRewardCollector: Option<PalletId> = Some(TreasuryPalletId::get());
+}
+
+impl pallet_block_rewards::Config for Runtime {
+	type AdminOrigin = EnsureRootOr<HalfOfCouncil>;
+	type AuthorityId = AuraId;
+	type Balance = Balance;
+	type CollatorCurrencyId = CollatorCurrencyId;
+	type CollatorGroupId = CollatorGroupId;
```
lemunozm, line 1715: I think there is no need to specify this from the outside. It can be a constant inside the pallet (?).

runtime/development/src/lib.rs
```diff
@@ -1678,6 +1691,44 @@ impl pallet_liquidity_rewards::Config for Runtime {
...
+	type Currency = Tokens;
+	type CurrencyId = CurrencyId;
+	type DefaultCollatorStake = DefaultCollatorStake;
+	type Domain = BlockRewardsDomain;
+	type GroupId = u32;
+	type MaxChangesPerEpoch = MaxChangesPerEpoch;
+	type MaxCollators = MaxCandidates;
+	type MaxGroups = MaxGroups;
```
lemunozm, line 1723: Same here, should always be one?

runtime/development/src/lib.rs
```diff
@@ -1678,6 +1691,44 @@ impl pallet_liquidity_rewards::Config for Runtime {
...
+	type AdminOrigin = EnsureRootOr<HalfOfCouncil>;
+	type AuthorityId = AuraId;
+	type Balance = Balance;
+	type CollatorCurrencyId = CollatorCurrencyId;
+	type CollatorGroupId = CollatorGroupId;
+	type Currency = Tokens;
+	type CurrencyId = CurrencyId;
+	type DefaultCollatorStake = DefaultCollatorStake;
+	type Domain = BlockRewardsDomain;
+	type GroupId = u32;
```
lemunozm, line 1720: Same here

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+
+			NextEpochChanges::<T>::mutate(|changes| changes.reward = Some(balance));
+
+			Ok(())
+		}
+
+		/// Admin method to set the group weights used for the next epochs.
+		/// Current epoch is not affected by this call.
+		#[pallet::weight(T::WeightInfo::set_group_weight())]
+		pub fn set_group_weight(
```
lemunozm, line 250: I thought we finally decided to have only one group, but I probably remember badly. What's the use case for several groups?

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+					Self::deposit_event(Event::NewEpoch {
+						reward: epoch_data.reward,
+						last_changes: mem::take(changes),
+					});
+
+					Ok(())
+				})
+			})
+		})
+		.ok();
```
lemunozm, line 367: This applies also to `pallet-liquidity-rewards`. Maybe we should add a `log::error()` or similar in case this doesn't work as expected to realize that. Wdyt? 
BTW, I do not how the log works on Substrate, is this something I could see from the `polkadot.js` or only in the terminal where I run the node?

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+						.into_iter()
+						.filter(|n| !current.iter().any(|curr| curr == n))
+						.collect::<Vec<_>>();
+					let out = current
+						.clone()
+						.into_iter()
+						.filter(|curr| !next.iter().any(|n| n == curr))
+						.collect::<Vec<_>>();
+					collators.inc = BoundedVec::<_, T::MaxCollators>::truncate_from(inc);
+					collators.out = BoundedVec::<_, T::MaxCollators>::truncate_from(out);
```
lemunozm, line 420: What would happen with the currency minted by the collators that don't fit into the vec? (Could this even happen?)
Should we call `do_exit_collator()` for those?

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+					let inc = next
+						.clone()
+						.into_iter()
+						.filter(|n| !current.iter().any(|curr| curr == n))
+						.collect::<Vec<_>>();
+					let out = current
+						.clone()
+						.into_iter()
+						.filter(|curr| !next.iter().any(|n| n == curr))
+						.collect::<Vec<_>>();
```
lemunozm, line 418: Could this method be called with a list of validators with duplicated accounts, or Substrate ensures it can not happen?

lemunozm: If two collators have the same account, I understand that I want to be rewarded by both collators. Filtering here I'm losing half of the reward (?)

lemunozm (commented):
I really like how it looks! Good job William! 🙌🏻

Some comments/suggestions/questions below

pallets/block-rewards/src/lib.rs
wischli: Yes, we decided on one group. I haven't abstracted that away yet. IMO, `set_group_weight` needs to be kept to enable changing the split-ratio between collator rewards and the treasury. 

For now, the rule is 3% annual inflation as block rewards of which Collators receive 1000 CFG per month.

pallets/block-rewards/src/lib.rs
wischli: > BTW, I do not how the log works on Substrate, is this something I could see from the `polkadot.js` or only in the terminal where I run the node?

AFAIK, you can only see this in the node logs unfortunately.

pallets/block-rewards/src/lib.rs
wischli: > What would happen with the currency minted by the collators that don't fit into the vec? (Could this even happen?)

The minting does not happen here. Here, we just note an upcoming collator set change in the next session/epoch `i+1`. Then, in the next epoch `i+1`, we reward these collators. 

Say we have more collators than allowed by our pallets bound, then the overflowing ones don't receive rewards. But since `MaxCollators` should be configured as `pallet_collator_selection::Config::MaxCandidates`, this is impossible to happen.
I will look into utilizing `integrity-tests` to ensure such assumptions hold true at compile time. Haven't used that hook before but recently stumbled upon it.

lemunozm: You're right regarding the collator, I saw the `do_advance_epoch()` on the top, but it uses the previous ones.

> I will look into utilizing integrity-tests to ensure such assumptions hold true at compile time. Haven't used that hook before but recently stumbled upon it.

I'll take a look. Sounds interesting, thanks!

pallets/block-rewards/src/lib.rs
lemunozm: Ok, I understand 🙌🏻 

Only as a note, modifying the weight of a group does not have any effect if there is only one group. The weight is a proportion among groups. `effective_rate = weight_i/weight_total`

pallets/block-rewards/src/lib.rs
wischli: > Could this method be called with a list of validators with duplicated accounts, or Substrate ensures it can not happen?

In our case, [this is ensured](https://github.com/paritytech/cumulus/blob/master/pallets/collator-selection/src/lib.rs#L376) by the `pallet-collator-selection` which handles on- and offboarding of collator candidates.

In practice, it also wouldn't make sense to have the same account twice here. In the end, the same entity can easily run multiple collators by using pseudonyms/proxies.

pallets/block-rewards/src/lib.rs
wischli: Thanks for pointing this out, I need to rethink how to handle the split rewarding then without instantiating a group representing the Treasury 🥲

pallets/block-rewards/src/lib.rs
lemunozm: Thanks for the info! 🙌🏻

pallets/rewards/src/lib.rs
mustermeiszer: > If the entry does not exist, then it mints; if an account exists for that domain then use that account.
Wdyt @mikiquantum ?

* How are domains added? 
* Could this be an exploit of adding domains and then minting into the reward?

In general, do you think it is feasible to have an 
```rust
pub enum RewardSource<AccountId> {
    Mint, 
    AccountId(AccountId)
}
```

as the storage of the get call?

runtime/development/src/lib.rs
mustermeiszer: Do we need to move this currency via some `fungibles::*` trait? If not, I am happy to not have this as part of the `enum CurrencyId` if yes lets add it like something like this

```rust
pub enum CurrencyId {
    ...
    RewardCurrency {
        id: [u8; 12] // Not sure about the format here
    }
}

```

pallets/rewards/src/lib.rs
wischli: > In general, do you think it is feasible to have an
> 
> ```rust
> pub enum RewardSource<AccountId> {
>     Mint, 
>     AccountId(AccountId)
> }
> ```
> 
> as the storage of the get call?

The reason why I did not bother with assigning an `AccountId` here for the time being was that the conversion of the Treasury `PalletId` to its truncating `AccountId` is not static which causes trouble in the runtime configuration. Sure there will be ways.

AFAICT, the recommended way to handle this in Substrate is to make use of [the `OnUnbalanced<NegativeImbalance>` trait](https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/tokens/imbalance/on_unbalanced.rs#L32) which is implemented by the Treasury. The approach here is a bit different: Independent of whether the `Beneficiary` is set (e.g. `type Beneficiary = Treasury` in runtime) or not (`type Beneficiary = ()`), we mint the reward. In case it is set to the `Treasury`, the `Treasury`[ will _claim_ the unbalanced](https://github.com/paritytech/substrate/blob/master/frame/treasury/src/lib.rs#L600) amount. Else, it will be dropped automatically.

See 257308de3c273f2dea267d6b674357011ca7bc8c. WDYT?

runtime/development/src/lib.rs
lemunozm: Thanks! Yes, we need to use fungibles for it

pallets/rewards/src/lib.rs
lemunozm: > How are domains added?
It's something used internally to differentiate multiple reward systems on the same `pallet-rewards`. Nobody can manipulate them, only the chain when it's configured.
> Could this be an exploit of adding domains and then minting into the reward?
No, because the previous reason.

wischli: > How the system knows to mint the correct `CurrencyId` (which should be `RewardCurrency`).

Good point! It was configured incorrectly. Should be fixed in bdb6d126cd659e74b593404eaa439609f1d670ff. It is derived from the `Balances` of the respective runtime which is the native currency in our case.

runtime/development/src/lib.rs
wischli: > To clarify, it's not the currency given for reward. It's an emulated currency used to stake internally. I think a simple variant without content is enough.

So `RewardCurrency` should be entirely fine, right?

mustermeiszer: My main Objektive is to but clutter our currency id enum with those staking/reward logic based currencies.

pallets/rewards/src/lib.rs
wischli: Sorry @lemunozm, the `OnUnbalanced` stuff should have been posted to the `pallet_block_rewards::Config::Beneficiary` and not here. I mixed up the comments. Will look into the above option!

mikiquantum (commented):
Sorry for coming a bit late to the convo. Looks good overall!

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Claims the reward the associated to a currency.
+		/// The reward will be transferred to the target account.
+		#[pallet::weight(T::WeightInfo::claim_reward())]
+		#[transactional]
+		pub fn claim_reward(
+			origin: OriginFor<T>,
+			currency_id: T::CurrencyId,
```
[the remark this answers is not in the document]
wischli: Totally agree! I actually removed this earlier in 6fa7c4ae515ddfc866c04c06df427d586e62ac8d

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+				changes.collator_reward = Some(collator_reward_per_session);
+			});
+
+			Ok(())
+		}
+
+		/// Admin method to set the inflation amount for the Beneficiary used for the next epochs.
+		/// Current epoch is not affected by this call.
+		#[pallet::weight(T::WeightInfo::set_distributed_reward())]
+		pub fn set_beneficiary_reward(
```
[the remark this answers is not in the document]
wischli: We could definitely do it your way as well! I kind of went this path before [1ee21af](https://github.com/centrifuge/centrifuge-chain/pull/1198/commits/1ee21af95af8a31861dfd2ca22c08f9812777c67), when it became aware that rewarding a group with amount `x` consumes the entire `x`. 

> Everytime we include a new collator, we will need to update this value, which possibly for a period of time we would be giving more than what we should.

The current design does not need to be reconfigured when new collators join as the configured reward is per collator, at least under the assumption that `n * collator_reward + beneficiary_amount` does not need to stay static.
However, I definitely see the upside in your approach. Just on a note: We would still need to store two reward amounts, otherwise we would have to hardcode the share for a collator.

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+				ActiveEpochData::<T>::try_mutate(|epoch_data| {
+					num_joining = changes.collators.inc.len();
+					num_leaving = changes.collators.out.len();
+
+					// Apply collator set changes before rewarding
+					for leaving in changes.collators.out.iter() {
+						Self::do_exit_collator(leaving)?;
+					}
+					for joining in changes.collators.inc.iter() {
+						Self::do_init_collator(joining)?;
```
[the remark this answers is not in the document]
wischli: In the current design, all collators receive the same share of the reward. This is due to the fact that everyone has the same artificial stake invoked by `do_init_collators`. After removing the ability to adjust a single collator's stake via the `AdminOrigin`, it's currently not possible for collators to receive different amounts of rewards per epoc. Please see [my open question in the spec](https://centrifuge.hackmd.io/9T9zAZaTRD6ZcHN-Gd_lzw#Open-question) for details.

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+	/// Disables receiving rewards onwards.
+	fn do_exit_collator(who: &T::AccountId) -> DispatchResult {
+		let amount =
+			T::Rewards::account_stake((T::Domain::get(), T::CollatorCurrencyId::get()), who);
+		T::Rewards::withdraw_stake(
+			(T::Domain::get(), T::CollatorCurrencyId::get()),
+			who,
+			amount,
+		)?;
+		T::Currency::burn_from(T::CollatorCurrencyId::get(), who, amount).map(|_| ())
```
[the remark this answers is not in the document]
wischli: We are burning because `withdraw_stake` frees the currency from its holding state such that it is transferrable by default. Additionally, there is no reason for someone holding the artificial staking currency without being a collator.

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
mikiquantum: yes, the `collator_reward` is per collator and that doesn't have to change when changing the collator size, but that new value now affects the `beneficiary_amount` that goes into the treasury right? That value will need to be adapted accordingly when a new size of collators is defined. 
Am I understanding correctly that the `beneficiary_reward` is the `3% inflation rate - the collator reward`? If so those two values are interconnected.
Yes, in my proposal we still have to store 2 values.

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
mikiquantum: I agree that every collator should get the same reward per epoch, so for me we do not need to change that behavior. What I am not sure I understand, is that a new collator gets the old (default) collator reward and not the new updated one from storage. Since the collator reward is per collator and not per epoch for all collators.

runtime/development/src/lib.rs
lemunozm: Ok, gotcha, as a simple namespace 👍🏻. Although at least for now, we only have one currency of this kind.

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
wischli: There is no _default_ or _old collator reward_, only the `EpochData.collator_reward` stored in the `ActiveEpochData ` map as of now which denotes the reward each collator should receive per epoch. Thus, in total `num_collators * EpochData.collator_reward` for the group. For one epoch, all current collators receive exactly that amount, independent of whether they are new or not.

If in `session(i)` a change to the collator reward is configured and thus stored in `NextEpochChanges` StorageMap, then this change is applied in `session(i+1)` but AFTER rewarding the collators, so it becomes effective in `session(i+2)`. The argument of @lemunozm was

> I think it makes sense to reward first and change groups after. The main idea behind this is that once the epoch starts, nothing should change for that epoch, so I, as an investor, can evaluate if it's ok to leave my money there during the next whole period. If we allow updating the group before, an admin could change it 3 days before the 1-month epoch ends. Knowing that, as an investor, 1 month ago, I could have decided to not invest in that.
So, yes, you need at least 1 completed epoch to change group properties.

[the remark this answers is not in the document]
mikiquantum: ah! the `do_init_collator` sets/mints only the virtual stake. I thought it was already rewarding. Makes sense now.

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
wischli: Yes that is correct. Let's summarize: I will change `beneficiary_reward` to denote the total reward per epoch instead. From that, we subtract the rewards for all collators and the rest
* goes to the Treasury for Centrifuge chain
* stays in the Treasury for Altair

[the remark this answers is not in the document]
mikiquantum: I think so, that means that the `beneficiary_reward`, if set, has to be at least the same as the collator_reward * number of collators at that point in time. 
And for Altair, we basically do not set it, and the logic should be able to only mint/move the collator_reward without interacting with the Zero value of the `beneficiary_reward`.

pallets/rewards/src/lib.rs
wischli: To whom it may concern: Moved to https://github.com/centrifuge/centrifuge-chain/pull/1209

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
wischli: Fixed in [8b8c40f](https://github.com/centrifuge/centrifuge-chain/pull/1198/commits/8b8c40f805b8c8b6b1c0eed9b0ffa00df17e6ff3). Not sure if I should rename `Beneficiary` to `Treasury`.

ci/script.sh
```diff
@@ -36,7 +36,7 @@ case $TARGET in
     ;;
 
   integration)
-    RUST_MIN_STACK=8388608 cargo test --release --package runtime-integration-tests
+    RUST_MIN_STACK=8388608 cargo test --release --package runtime-integration-tests --features fast-runtime
```
wischli, line 39: Required for newly added `pallet-block-rewards` integration tests which need to iterate at least four sessions (20 blocks/~2 min with `fast-runtime`)

libs/types/src/permissions.rs
```diff
@@ -694,6 +694,11 @@ mod tests {
 				KSM => vec![2],
 				AUSD => vec![3],
 				ForeignAsset(id) => vec![4, id as u64, 0, 0, 0],
+				Rewards { id } => {
+					let mut r = vec![5];
+					r.append(&mut id.map(|x| x as u64).to_vec());
```
wischli, line 699: Honestly, not sure if I followed the schema correctly.

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+	pub(crate) fn do_exit_collator(who: &T::AccountId) -> DispatchResult {
+		let amount = T::Rewards::account_stake((T::Domain::get(), STAKE_CURRENCY_ID), who);
+		T::Rewards::withdraw_stake((T::Domain::get(), STAKE_CURRENCY_ID), who, amount)?;
+		T::Currency::burn_from(STAKE_CURRENCY_ID, who, amount).map(|_| ())
+	}
+
+	/// Apply session changes and distribute rewards.
+	///
+	/// NOTE: Noop if any call fails.
+	fn do_advance_session() {
```
wischli, line 349: It does not make sense to benchmark this hook as the a session change [blocks maximum weight](https://github.com/paritytech/substrate/blob/master/frame/session/src/lib.rs#L569-L572) anyways.

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+			})
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	/// Mint default amount of stake for target address and deposit stake.
+	/// Enables receiving rewards onwards.
+	///
+	/// Weight (6 reads, 6 writes):
```
wischli, line 326: For now, knowing the correct weight would only be necessary for the migration. We have 8 collators each on Altair and Centrifuge. So we should be fine without an accurate benchmark IMO. WDYT?

pallets/block-rewards/src/mock.rs
```diff
@@ -0,0 +1,383 @@
...
+impl pallet_block_rewards::Config for Test {
+	type AdminOrigin = EnsureRoot<AccountId>;
+	type AuthorityId = UintAuthorityId;
+	type Balance = Balance;
+	type Beneficiary = RewardRemainderMock;
+	type Currency = Tokens;
+	type Domain = BlockRewardsDomain;
+	type MaxChangesPerSession = MaxChangesPerSession;
+	type MaxCollators = MaxCollators;
+	type Rewards = Rewards;
```
wischli, line 244: @lemunozm I did not go with the mock here since I felt more comfortable staying closer to the runtime configuration. Also felt like the mock in its current form would have bloated up the tests with a lot of boilerplate. 

Now that I can be sure the tests are correct and sufficient, I am happy to revert this if you think it is to be preferred.

runtime/development/src/lib.rs
```diff
@@ -2094,12 +2160,25 @@ impl_runtime_apis! {
...
+			list_benchmark!(list, extra, pallet_session, SessionBench::<Runtime>);
+			list_benchmark!(list, extra, pallet_block_rewards, BlockRewards);
 
 			let storage_info = AllPalletsWithSystem::storage_info();
 
 			return (list, storage_info)
 		}
 	}
+
+	#[cfg(feature = "try-runtime")]
```
wischli, line 2172: NOTE: `try-runtime` was missing for development-runtime

runtime/development/src/lib.rs
```diff
@@ -1809,24 +1861,23 @@ pub type UncheckedExtrinsic =
 /// Extrinsic type that has already been checked.
 pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, RuntimeCall, SignedExtra>;
 
+parameter_types! {
```
wischli, line 1864: These should be the correct numbers we eventually want to have on Centrifuge, taken from the [gov proposal](https://gov.centrifuge.io/t/rfc-add-a-block-reward-and-improve-collators-cycle/4766).

runtime/development/src/lib.rs
```diff
@@ -1809,24 +1861,23 @@ pub type UncheckedExtrinsic =
...
+	// = 16.65 CFG per epoch (12h)
+	pub const CollatorRewards: Balance = 8_325 * MILLI_CFG;
+	// = 20,096 CFG per epoch (12h)
+	pub const TotalRewards: Balance = 10_048 * CFG;
+}
```
wischli, line 1869: These should be the correct numbers we eventually want to have on Centrifuge, taken from the [gov proposal](https://gov.centrifuge.io/t/rfc-add-a-block-reward-and-improve-collators-cycle/4766).

wischli:
@lemunozm @mikiquantum @mustermeiszer If you want, you can give this PR a final review. Please note the dependency to #1223 which fixes issues for updating `nix` in this PR.

pallets/block-rewards/src/mock.rs
```diff
@@ -0,0 +1,383 @@
...
+	pub const NativeToken: CurrencyId = CurrencyId::Native;
+
+	#[derive(scale_info::TypeInfo)]
+	pub const MaxCurrencyMovements: u32 = 1;
+}
+
+impl pallet_rewards::Config<pallet_rewards::Instance1> for Test {
+	type Currency = Tokens;
+	type CurrencyId = CurrencyId;
+	type DomainId = RewardDomain;
```
lemunozm, line 207: A simply `type DomainId = ()` is enough here if you get rid from the `Liquidity` name that is out of this scope.

pallets/block-rewards/src/mock.rs
lemunozm: It's true that the correct mock system (based on `mockall`) is hard and bolierplate. If you anyways needs to use `pallet_restricted_tokens` as a dependency it also loses part of its sense.

My only concern is, what happen if a user uses a different reward mechanism? Are we testing and make assumptions as if it was always the `base` mechanism? Are we not covered if the user of this pallet changes the mechanism?

By now, I think let leave it as it is, and once the `mock-builder` is completed, we'll see if then it simplify things and if it's worth using here 👍🏻

pallets/block-rewards/src/mock.rs
```diff
@@ -0,0 +1,383 @@
...
+
+	pub(crate) fn set_total_reward(mut self, reward: Balance) -> Self {
+		self.total_reward = reward;
+		self
+	}
+
+	pub(crate) fn set_run_to_block(mut self, run_to_block: BlockNumber) -> Self {
+		self.run_to_block = run_to_block;
+		self
+	}
```
lemunozm, line 344: likes a lot! 🚀

lemunozm (commented):
Minor review of the mock part

pallets/block-rewards/src/mock.rs
wischli: Unfortunately we cannot default to `()` as `DomainId` and `pallet_block_rewards::Config::Domain` require `TypedGet`. I played around with it and any change would be similar to the current version, e.g. some type of enum Domain. Thus, I propose to keep as is. WDYT?

lemunozm: Ok, I see. It's ok then 👍🏻 

To be purist, maybe `enum { MyDomain, OtherDomain }` is more representative for a reader of what `Domain` means there, but I'm perfectly ok with whatever you choose

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+
+		transactional::with_storage_layer(|| -> DispatchResult {
+			NextSessionChanges::<T>::try_mutate(|changes| -> DispatchResult {
+				ActiveSessionData::<T>::try_mutate(|session_data| {
+					// Reward collator group of last session
+					let total_collator_reward = session_data
+						.collator_reward
+						.saturating_mul(session_data.collator_count.into())
+						.min(session_data.total_reward);
+					T::Rewards::reward_group(COLLATOR_GROUP_ID, total_collator_reward)?;
```
lemunozm, line 361: I'm wondering if `COLLATOR_GROUP_ID` could be used by another pallet. Maybe we need to use some "Domain" mechanism here too.

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+						collator_reward: session_data.collator_reward,
+						last_changes: mem::take(changes),
+					});
+
+					Ok(())
+				})
+			})
+		})
+		.map_err(|e| {
+			log::error!("Failed to advance block rewards session: {:?}", e);
```
lemunozm, line 405: If this is the case, I would change the `saturating_ops` above to `ensure_ops`, to avoid "weird invisible behaviors" if those operations "fails". WDYT?
Maybe we should generate some kind of error event, too (not sure if this is the intended behavior of events, but sounds reasonable at least in infallible functions).

pallets/rewards/src/lib.rs
```diff
@@ -442,7 +441,7 @@ pub mod pallet {
 		RewardAccountOf<T, I>: FullCodec + Default,
 	{
 		pub fn list_currencies(
-			account_id: T::AccountId,
+			account_id: &T::AccountId,
```
lemunozm, line 444: 👍🏻

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+pub struct SessionChanges<T: Config> {
+	pub collators: CollatorChanges<T>,
+	pub collator_count: Option<u32>,
+	collator_reward: Option<T::Balance>,
+	total_reward: Option<T::Balance>,
+}
+
+pub const COLLATOR_GROUP_ID: u32 = 1;
+pub const DEFAULT_COLLATOR_STAKE: u32 = 1000;
+pub const STAKE_CURRENCY_ID: CfgCurrencyId = CfgCurrencyId::Rewards { id: *b"blkrwrds" };
```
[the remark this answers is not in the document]
wischli: My understanding was that liquidity rewards would fall under the same umbrella, e.g. `CurrencyId::Rewards { id: *b"lqdrwrds" }`. However, seems like we would rather want to distinguish them as separate variants of `CurrencyId`. WDYT @mustermeiszer ?

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+			GenesisConfig {
+				collators: Default::default(),
+				collator_reward: Default::default(),
+				total_reward: Default::default(),
+			}
+		}
+	}
+
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
```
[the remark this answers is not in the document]
wischli: For `Altair` and `Centrifuge` that is correct. However, `Development` is not supposed to be persistent and expected to launched from scratch frequently. Moreover, this `GenesisBuild` is at least required for the integration tests and is the correct way to initialize the pallet in the crate's mock.

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+
+		/// Type used to handle balances.
+		type Balance: Balance
+			+ MaxEncodedLen
+			+ FixedPointOperand
+			+ Into<<<Self as Config>::Currency as CurrencyT<Self::AccountId>>::Balance>
+			+ MaybeSerializeDeserialize;
+
+		/// Domain identification used by this pallet
+		type Domain: TypedGet;
```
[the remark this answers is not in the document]
wischli: I thought it would be wise to stay in sync with the the current PoC of `pallet-liquidity-rewards` which requires `TypedGet`.

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
lemunozm: The currency used for liquidity reward is the `CurrencyId::Tranche`, so I think you're the only one who will use that currency

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Ok, I see the reason why I did the same in liquidity 😅

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+		type Weight: Parameter + MaxEncodedLen + EnsureAdd + Unsigned + FixedPointOperand + Default;
+
+		/// The reward system used.
+		type Rewards: GroupRewards<Balance = Self::Balance, GroupId = u32>
+			+ AccountRewards<
+				Self::AccountId,
+				Balance = Self::Balance,
+				CurrencyId = (DomainIdOf<Self>, CfgCurrencyId),
+			> + CurrencyGroupChange<GroupId = u32, CurrencyId = (DomainIdOf<Self>, CfgCurrencyId)>
+			+ DistributedRewards<Balance = Self::Balance, GroupId = u32>;
```
lemunozm, line 154: I think you do not need this bound if uses the `reward_group()` directly.

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+						.collect::<Vec<_>>();
+					let out = current
+						.clone()
+						.into_iter()
+						.filter(|curr| !next.iter().any(|n| n == curr))
+						.collect::<Vec<_>>();
+					collators.inc = BoundedVec::<_, T::MaxCollators>::truncate_from(inc);
+					collators.out = BoundedVec::<_, T::MaxCollators>::truncate_from(out);
+
+					*collator_count = Some(next.len().saturated_into::<u32>());
```
mustermeiszer, line 462: I know we can not panic here. But an error log would be great. What would be the outfall of having this overflow? I know it will never happen ^^

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+			);
+		}
+	}
+
+	fn on_before_session_ending() {
+		// we don't care.
+	}
+
+	fn on_disabled(_validator_index: u32) {
+		// we don't care.
```
mustermeiszer, line 473: What means disabling from a session perspective?

mustermeiszer (commented):
Looks really, really good! 
Just minor stuff. Nothing blocking. Two question though: 
* Do the benchmarks of pallet-session work now?
* Did you already check how much weight increases from pallet-session perspective?

runtime/integration-tests/src/pools/env.rs
```diff
@@ -31,14 +32,19 @@ use crate::{
 async fn env_works() {
 	let mut env = env::test_env_default(Handle::current());
 
+	// FIXME: https://github.com/centrifuge/centrifuge-chain/issues/1219
```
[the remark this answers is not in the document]
wischli: Yes, it is still broken. The issue is the missing collators in the genesis setup for `pallet-session` as discussed on Slack.

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+pub use pallet::*;
+pub use sp_runtime::Saturating;
+use sp_runtime::{traits::Zero, FixedPointOperand, SaturatedConversion};
+use sp_std::{mem, vec::Vec};
+use weights::WeightInfo;
+
+#[derive(
+	Encode, Decode, DefaultNoBound, Clone, TypeInfo, MaxEncodedLen, PartialEq, RuntimeDebugNoBound,
+)]
+#[scale_info(skip_type_params(T))]
```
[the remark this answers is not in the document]
wischli: No, it has existed ever since `MaxEncodedLen` was added. It fixes compiler issues that arise from the derive macro for `T`. If we did not derive all types from `T::Config`, then deriving would be entirely fine.

runtime/integration-tests/src/pools/env.rs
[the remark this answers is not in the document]
mustermeiszer: But I found them in my tests. Lets maybe go through that together.

pallets/block-rewards/src/lib.rs
wischli: In the context of AuRa/Babe, it prevents such collators to author a block , see https://github.com/paritytech/substrate/pull/9414. However, slots assigned to disabled authorities are just skipped in AuRa e.g. no block is built. Thus, there is no incentive to disable a collator IMO.

In the context of a session, we could unstake these collators. However, their session keys are still part of `QueuedKeys`. If we wanted to handle this scenario, more checks would have to be introduced to `on_new_session`

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
wischli: Let's go with the single variant in the top level enum then!

```diff
pub enum CurrencyId {
        // ...
-	Rewards {
-		id: [u8; 8],
-	},
+	BlockRewards,
}
```

wischli:
> * Do the benchmarks of pallet-session work now?

Yes, they work!

> * Did you already check how much weight increases from pallet-session perspective?

It actually has decreased when compiling for `dev` runtime. Have to dig into why, should actually have added a read and write when looking at the current benches for `pallet_session` in Polkadot repo.

```diff
impl WeightInfo for () {
	fn set_keys() -> Weight {
		Weight::from_ref_time(24_000_000)
-    	.saturating_add(RocksDbWeight::get().reads(6 as u64))
-			.saturating_add(RocksDbWeight::get().writes(5 as u64))
+			.saturating_add(RocksDbWeight::get().reads(2 as u64))
+			.saturating_add(RocksDbWeight::get().writes(2 as u64))
	}

	fn purge_keys() -> Weight {
		Weight::from_ref_time(19_000_000)
-    	.saturating_add(RocksDbWeight::get().reads(2 as u64))
-			.saturating_add(RocksDbWeight::get().writes(5 as u64))
+			.saturating_add(RocksDbWeight::get().reads(1 as u64))
+			.saturating_add(RocksDbWeight::get().writes(2 as u64))
	}
}
```

pallets/block-rewards/src/lib.rs
lemunozm: The admin of the liquidity pallet can create a group; if they choose the same ID as this, what would happen?

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+			}
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Claims the reward the associated to a currency.
+		/// The reward will be transferred to the target account.
+		#[pallet::weight(T::WeightInfo::claim_reward())]
+		#[transactional]
```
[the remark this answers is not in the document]
lemunozm: Just tested and works as a transaction in test too, even if you do not put transactional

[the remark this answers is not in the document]
wischli: Removed, wasn't intentional.

pallets/block-rewards/src/lib.rs
wischli: AFAIU, `pallet-liquidity-rewards` and `pallet-block-rewards` each require a separate instance of `pallet-rewards`. Thus, even if the same GroupID exists in `pallet-liquidity-rewards`, it cannot influence `pallet-block-rewards`.

wischli: I definitely agree this could potentially lead to a silent bug. However, what do we win if we define it in `Config`?

lemunozm: Visibility. it's no longer an internal library value. However, it's true that we should configure some intervals to use in the liquidity, too to avoid collisions. So we do not win anything 😅. We need some kind of group domain level or always use a different `pallet-reward` each time.

Maybe we can remove the `Domain` concept that adds a lot of noise and always forces us to use different instances. This would simplify things a lot 🤔. What do you think about this? Of course, this will be another PR.

lemunozm: The `Domain` concept was something that we implemented before allowing `pallet-rewards` to have several instances. So maybe this is now a concept that can be deprecated

pallets/block-rewards/src/lib.rs
wischli: It could "overflow" if the `<Runtime as pallet_aura::Config>::MaxAuthorities` represents a larger value than it's current type definition of `u32` which is impossible. In the context of this pallet, the outfall would be that some collators miss out on their rewards.

Do you still want to have an error log? For now, I added a comment. See fdcd632d2bbf5a8b1bbc729676fb69b38d4738c0

pallets/block-rewards/src/lib.rs
wischli: > The Domain concept was something that we implemented before allowing pallet-rewards to have several instances. So maybe this is now a concept that can be deprecated

Thanks for the insights. I was wondering whether it could be simplified but did not want to add noise at the current stage.

> Maybe we can remove the `Domain` concept that adds a lot of noise and always forces us to use different instances. This would simplify things a lot 🤔. What do you think about this? Of course, this will be another PR.

I agree, will open a ticket!

wischli: https://github.com/centrifuge/centrifuge-chain/issues/1246

lemunozm: > I was wondering whether it could be simplified, but I did not want to add noise at the current stage.

Do it, do it. It is needed to avoid currency collisions using the same `pallet-rewards`. But I can be blind to one solution because I'm too biased on this. New mind sets are welcome!

pallets/block-rewards/src/lib.rs
wischli: Thanks! 5f845386ea3cb664fd71aec49f927725afd031ca

libs/primitives/src/lib.rs
```diff
@@ -246,6 +246,13 @@ pub mod constants {
 	pub const fn deposit(items: u32, bytes: u32) -> Balance {
 		items as Balance * 15 * CENTI_CFG + (bytes as Balance) * 6 * CENTI_CFG
 	}
+
+	/// The identifier of the group eligible to receive block rewards.
+	pub const COLLATOR_GROUP_ID: u32 = 1;
+
+	/// The default amount of stake for CurrencyId::BlockRewards which is inherently
+	/// assigned to any member of the only group in block rewards.
+	pub const DEFAULT_COLLATOR_STAKE: u32 = 1000;
```
mustermeiszer, line 255: I wouldn't put this into primitives, but rather types and then ids or so. But no blocker. Just feels like it could change and primitives should not change ^^

libs/types/src/tokens.rs
```diff
@@ -41,6 +41,10 @@ pub enum CurrencyId {
 
 	/// A foreign asset
 	ForeignAsset(ForeignAssetId),
+
+	/// An emulated internal, non-transferrable currency
+	/// Its issuance and holding is handled inherently
+	BlockRewards,
```
mustermeiszer, line 47: Why the change? I liked the id based approach. My reasoning here is that we will most likely need another currency CfgStake or so, that user can get by submitting CFG and then stake on pools or loans or so with it. I think it makes sense to have at least the following

```rust
pub enum CurrencyId {
     ....
     Staking(StakingCurrency) 
}

pub enum StakingCurrency {
    BlockRewards,
}
```

But just a thought out of my head. I also would like to get feedback on that. We can also go with the current approach, I just have the feeling that we are slowly cluttering the enum here and it gets really dirty. Just like `AUSD` is not really a good varaint, as we are not even having it nor have we a plan on how we want to extend the enum here with all the connectors tokens.

mustermeiszer (dismissed):
I still have some worries about the `CurrencyId` enum but otherwise looks good!

libs/types/src/tokens.rs
wischli: I mistook your silence and @lemunozm's rocket emoji [here](https://github.com/centrifuge/centrifuge-chain/pull/1198#discussion_r1131109660) as approval to go with the proposal. I was under the assumption we are not planning to add more "staking" currency ids anytime soon. However, I totally agree with the cluttering aspect and prefer this option here.

lemunozm: My rocket has been a destruction missile 😆. Currently, I'm not seeing another currency related to this (I mean, a special internal currency for reward purposes). But I'm ok with whatever solution we choose.

cdamian (approved):

mustermeiszer (commented):
Looks pretty ready. What do you think @wischli? I still need to look at the migration in detail. Otherwise, nice! 🚀

libs/types/src/tokens.rs
```diff
@@ -41,6 +41,19 @@ pub enum CurrencyId {
...
+
+	/// A staking token
+	Staking(StakingCurrency),
+}
+
+#[derive(
+	Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen,
+)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum StakingCurrency {
```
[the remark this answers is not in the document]
wischli: Totally agree! Done in [dd50f66](https://github.com/centrifuge/centrifuge-chain/pull/1198/commits/dd50f669063ec5a4140428382cb2ae25b96a0245)

pallets/block-rewards/src/lib.rs
```diff
@@ -237,7 +239,10 @@ pub mod pallet {
 	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
 		fn build(&self) {
 			T::Rewards::attach_currency(
-				(T::Domain::get(), CfgCurrencyId::BlockRewards),
+				(
+					T::Domain::get(),
+					CfgCurrencyId::Staking(StakingCurrency::BlockRewards),
```
lemunozm, line 244: Not sure if this has already been said. Why use this directly instead of a `CurrencyId` type parameter? I feel like the `StakingCurrency::BlockRewards` value should be added by config.

pallets/block-rewards/src/lib.rs
```diff
@@ -268,8 +273,14 @@ pub mod pallet {
 		pub fn claim_reward(origin: OriginFor<T>, account_id: T::AccountId) -> DispatchResult {
 			ensure_signed(origin)?;
 
-			T::Rewards::claim_reward((T::Domain::get(), CfgCurrencyId::BlockRewards), &account_id)
-				.map(|_| ())
+			T::Rewards::claim_reward(
+				(
+					T::Domain::get(),
+					CfgCurrencyId::Staking(StakingCurrency::BlockRewards),
+				),
```
lemunozm, line 280: I think this could by a `type CurrencyIdOf<T> =...` to avoid boilerplate

pallets/block-rewards/src/lib.rs
```diff
@@ -39,13 +39,15 @@ pub mod weights;
...
 use cfg_traits::ops::{EnsureMul, EnsureSub};
 pub use cfg_traits::{
 	ops::{EnsureAdd, EnsureAddAssign},
 	rewards::{AccountRewards, CurrencyGroupChange, DistributedRewards, GroupRewards},
 };
-use cfg_types::tokens::CurrencyId as CfgCurrencyId;
+use cfg_types::{
+	ids::{COLLATOR_GROUP_ID, DEFAULT_COLLATOR_STAKE},
+	tokens::{CurrencyId as CfgCurrencyId, StakingCurrency},
+};
```
lemunozm, line 50: Regarding my first comment. This probably is a long discussion that can be done in another issue, but, I think that in the `cfg_types` crate, we are mixing runtime things with common generic pallet things. In this case, our `CurrencyId` is almost sure part of how our runtime works.

lemunozm (commented):
Beyond one concern (but not blocking) and one minor comment, I think this PR is a really great work! 🙌🏻

pallets/block-rewards/src/lib.rs
wischli: Yes we can do this. I interpreted [your comment](https://github.com/centrifuge/centrifuge-chain/pull/1198#discussion_r1105795201) to include this for the `CurrencyId` as well.

lemunozm: Thanks for the change @wischli! 🙌🏻

lemunozm (approved):

pallets/block-rewards/src/migrations.rs
```diff
@@ -0,0 +1,159 @@
...
+				>= CollatorReward::get().saturating_mul(collators.len().saturated_into())
+		);
+
+		log::info!("💰 BlockRewards: Pre migration checks successful");
+
+		Ok(collators.encode())
+	}
+
+	// Weight: 2 + collator_count reads and writes
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
```
mikiquantum, line 79: 👍

runtime/development/src/lib.rs
```diff
@@ -1674,6 +1708,27 @@ impl pallet_liquidity_rewards::Config for Runtime {
...
+frame_support::parameter_types! {
+	pub const BlockRewardsDomain: RewardDomain = RewardDomain::Block;
+	pub const BlockRewardCurrency: CurrencyId = CurrencyId::Staking(BlockRewardsCurrency);
+}
+
+impl pallet_block_rewards::Config for Runtime {
+	type AdminOrigin = EnsureRootOr<HalfOfCouncil>;
+	type AuthorityId = AuraId;
+	type Balance = Balance;
+	type Beneficiary = ();
```
mikiquantum, line 1720: Since Centrifuge runtime will have the treasury here as the beneficiary, would it make sense to already add it here, so we can verify that piece works fine?

runtime/development/src/lib.rs
```diff
@@ -114,6 +116,7 @@ include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 impl_opaque_keys! {
 	pub struct SessionKeys {
 		pub aura: Aura,
+		pub block_rewards: BlockRewards,
```
mikiquantum, line 119: Since we couldn't use SessionManager because (i dont think it supports tuples) and you need to know the aura key changes so can adjust the collator stake accordingly, I think it makes sense. 
It seems clean, although I am not able to understand how the properties inside the SessionKeys struct makes the compiler happy about the "whole" SessionKeys struct implementing successfully the SessionHandler trait.

mikiquantum (commented):
Looked at the spec as well, and every requirement seems to be fulfilled. 
Thanks and good job @wischli !

pallets/block-rewards/src/lib.rs
```diff
@@ -131,7 +131,8 @@ pub mod pallet {
 			+ MaxEncodedLen
 			+ FixedPointOperand
 			+ Into<<<Self as Config>::Currency as CurrencyT<Self::AccountId>>::Balance>
-			+ MaybeSerializeDeserialize;
+			+ MaybeSerializeDeserialize
+			+ From<cfg_primitives::Balance>;
```
lemunozm, line 135: If this line is added because of `DEFAULT_COLLATOR_STAKE` type, I'll add this value as a constant too.

I'm thinking about what would happen if another chain wanted to use `block-rewards` as a dependency. It could be really cool to offer this crate in a more standalone way (it is already almost in this way).

I only wanted to leave this idea here. I'm ok as it is now 🚀

lemunozm (approved):
It's ready for me! 🚀 

Good job @wischli !!

libs/types/src/ids.rs
```diff
@@ -42,4 +43,4 @@ pub const CHAIN_BRIDGE_NATIVE_TOKEN_ID: [u8; 4] = *b"xCFG";
 pub const COLLATOR_GROUP_ID: u32 = 1;
 /// The default amount of stake for CurrencyId::Staking(StakingCurrency::BlockRewards) which is inherently
 /// assigned to any member of the only group in block rewards.
-pub const DEFAULT_COLLATOR_STAKE: u32 = 1000;
+pub const DEFAULT_COLLATOR_STAKE: Balance = CFG;
```
mustermeiszer, line 46: Does this mean 1 CFG, right?

pallets/block-rewards/src/lib.rs
mustermeiszer: Think this is a good idea.

pallets/block-rewards/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+cfg-primitives = { path = "../../libs/primitives", default-features = false }
+cfg-traits = { path = "../../libs/traits", default-features = false }
+cfg-types = { path = "../../libs/types", default-features = false }
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
+frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.32" }
+frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.32" }
+log = { version = "0.4.17", default-features = false }
```
mustermeiszer, line 21: Why do we need log?

pallets/block-rewards/src/migrations.rs
```diff
@@ -0,0 +1,159 @@
...
+		let collators = get_collators::<T>();
+		assert!(!collators.is_empty());
+
+		assert!(!CollatorReward::get().is_zero());
+		assert!(
+			TotalReward::get()
+				>= CollatorReward::get().saturating_mul(collators.len().saturated_into())
+		);
+
+		log::info!("💰 BlockRewards: Pre migration checks successful");
```
mustermeiszer, line 73: Ahh. I see.

pallets/block-rewards/src/migrations.rs
```diff
@@ -0,0 +1,159 @@
...
+				// TODO: Benching preferred to be precise.
+				// However, not necessarily needed as num of collators <= 10.
+				Pallet::<T>::do_init_collator(collator)
+					.map_err(|e| {
+						log::error!("Failed to init genesis collators for rewards: {:?}", e);
+					})
+					.ok();
+				weight.saturating_accrue(T::DbWeight::get().reads_writes(6, 6));
+			}
+			Pallet::<T>::current_storage_version().put::<Pallet<T>>();
```
mustermeiszer, line 114: Is this substrate magic? Where is the version defined?

runtime/development/src/lib.rs
```diff
@@ -1996,11 +2052,18 @@ impl_runtime_apis! {
 	// RewardsApi
 	impl runtime_common::apis::RewardsApi<Block, AccountId, Balance, RewardDomain, CurrencyId> for Runtime {
 		fn list_currencies(account_id: AccountId) -> Vec<(RewardDomain, CurrencyId)> {
-			pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance1>::list_currencies(account_id)
+			pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance1>::list_currencies(&account_id)
+			.into_iter().chain(
+				pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance2>::list_currencies(&account_id).into_iter()
+			).collect()
```
mustermeiszer, line 2058: Is `list_currencies` a trait method? 

Then we can impl that for tuples.

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+
+		/// Admin method to set the reward amount for a collator used for the next sessions.
+		/// Current session is not affected by this call.
+		#[pallet::weight(T::WeightInfo::set_collator_reward())]
+		pub fn set_collator_reward(
+			origin: OriginFor<T>,
+			collator_reward_per_session: T::Balance,
+		) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
+
```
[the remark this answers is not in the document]
mustermeiszer: I think this is not needed, but a good sanity check maybe as we also do this check in `set_total_reward`. Wdyt @wischli or is the check in `set_total_reward` otherwise a safety mesaure?

* `do_advance`: Already does check `let total_reward = min(num_collators*reward_collator, total_reward)`

mustermeiszer (dismissed):
Niiiiiiice!! 

Migrations looks good, code looks good, tests look good, integrations test are sick! 

The only thing before merging this one from my side would be to check miguels comment and wait for #1233

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
wischli: IMO we should have the check in both locations to minimize the chance of having less total rewards than needed for the collators in `do_advance`.

libs/types/src/ids.rs
```diff
@@ -36,3 +36,10 @@ pub const STAKE_POT_PALLET_ID: PalletId = PalletId(*b"PotStake");
 // Other ids
 pub const CHAIN_BRIDGE_HASH_ID: [u8; 13] = *b"cent_nft_hash";
 pub const CHAIN_BRIDGE_NATIVE_TOKEN_ID: [u8; 4] = *b"xCFG";
+
+// Reward related
+/// The identifier of the group eligible to receive block rewards.
+pub const COLLATOR_GROUP_ID: u32 = 1;
+/// The default amount of stake for CurrencyId::Staking(StakingCurrency::BlockRewards) which is inherently
+/// assigned to any member of the only group in block rewards.
+pub const DEFAULT_COLLATOR_STAKE: u32 = 1000;
```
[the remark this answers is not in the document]
wischli: Definitely! Done in 3f44316ac611cba7fae74ea86fcc95229688d799

libs/types/src/ids.rs
wischli: Yes. Clippy does not like `1 * CFG`.

pallets/block-rewards/src/migrations.rs
wischli: The function itself [is part of the](https://paritytech.github.io/substrate/master/frame_support/dispatch/trait.GetStorageVersion.html#tymethod.current_storage_version) `GetStorageVersion` trait. The pallet's storage version `current_storage_version` [is in scope](https://github.com/centrifuge/centrifuge-chain/blob/feat/impl-block-rewards/pallets/block-rewards/src/lib.rs#L120) of the `frame_support::pallet` magic. 

So anytime we bump a pallet's storage version we still need to do a migration to have the `on_chain_storage_version` and `current_storage_version` aligned.

pallets/block-rewards/src/lib.rs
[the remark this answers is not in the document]
wischli: [27e4653](https://github.com/centrifuge/centrifuge-chain/pull/1198/commits/27e46538b255be5e3ed457e97b60797270739fce)

.github/CODEOWNERS
```diff
@@ -19,6 +19,7 @@ pallets/connectors/* @NunoAlexandre @offerijns
 ## Rewards related logic
 pallets/rewards/* @lemunozm @mikiquantum
 pallets/liquidity-rewards/* @lemunozm @mikiquantum
+pallets/block-rewards/* @wischli @lemunozm
```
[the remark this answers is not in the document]
wischli: [7223a31](https://github.com/centrifuge/centrifuge-chain/pull/1198/commits/7223a31133cb022a481304b746825738d0c06618)

pallets/block-rewards/src/lib.rs
wischli: Definitely agree. Refactored this together with the collator group id to make the pallet generic enough to be used by others: [02060eb](https://github.com/centrifuge/centrifuge-chain/pull/1198/commits/02060ebdecb65d1eadacb8b62bfd441d4f54a984)

runtime/development/src/lib.rs
wischli: Good catch! [5be160c](https://github.com/centrifuge/centrifuge-chain/pull/1198/commits/5be160c8690ecda671f2aefa4615c6bda15d9ddd)

pallets/block-rewards/src/lib.rs
```diff
@@ -0,0 +1,499 @@
...
+				CurrencyId = (DomainIdOf<Self>, CfgCurrencyId),
+			> + CurrencyGroupChange<GroupId = u32, CurrencyId = (DomainIdOf<Self>, CfgCurrencyId)>;
+
+		/// Type used to handle currency minting and burning for collators.
+		type Currency: Mutate<Self::AccountId, AssetId = CfgCurrencyId, Balance = Self::Balance>
+			+ CurrencyT<Self::AccountId>;
+
+		/// The identifier of the artificial block rewards currency which is minted and burned for collators.
+		#[pallet::constant]
+		type StakeCurrencyId: Get<CfgCurrencyId>;
```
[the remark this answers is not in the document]
lemunozm: I mean, the pallet can works without knowing that the `CurrencyId` is exactly a `CfgCurrencyId`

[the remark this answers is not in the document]
wischli: Yes, we have to add an `AssetId` associated type then. Will do this tomorrow! Good catch.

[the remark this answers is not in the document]
lemunozm: No hurry! Maybe a `CurrencyIdOf<T>` using the `CurrencyId` of `Rewards` traits or The `Currency::AssetId` saves you form another Config parameter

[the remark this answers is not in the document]
wischli: ecd49deb50e1149617f4220be2589a9095c7f32b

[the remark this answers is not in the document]
wischli: Since we need to specify the `CurrencyId` for the `Rewards` trait, I don't see how we get around adding a `CurrencyId` parameter to the `pallet_block_rewards::Config`.

runtime/development/src/lib.rs
```diff
@@ -1907,9 +1907,94 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	pallet_block_rewards::migrations::InitBlockRewards<Runtime, CollatorRewards, TotalRewards>,
+	UpgradeDev1020,
 >;
 
+type UpgradeDev1020 =
+	pallet_block_rewards::migrations::InitBlockRewards<Runtime, CollatorRewards, TotalRewards>;
```
lemunozm, line 1914: Why is this migration needed for development?

runtime/development/src/weights/pallet_restricted_tokens.rs
```diff
@@ -2,23 +2,23 @@
...
+//! DATE: 2023-01-12, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! HOSTNAME: `runner`, CPU: `Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz`
+//! EXECUTION: Some(Wasm), WASM-EXECUTION: Compiled, CHAIN: Some("altair-dev"), DB CACHE: 1024
 
 // Executed Command:
 // target/release/centrifuge-chain
 // benchmark
 // pallet
-// --chain=development-local
+// --chain=altair-dev
```
lemunozm: I think the runtime should be `development-local` here

lemunozm (approved):
LGTM! Let's get this merged soon!

A couple of minor questions.

runtime/development/src/lib.rs
wischli: For the `demo` chain and to enable testing the migration locally before we add it to `Altair` and `Centrifuge` later.

lemunozm: Ok, so as a norm, we should always create migrations for any breaking change in `development,` right?

lemunozm (dismissed):

runtime/development/src/lib.rs
wischli: I would say so, yes.

mustermeiszer (approved):
Re-approve

offerijns (approved):

PR #1200, opened 2023-02-12
Clippy take 2

branan:
# Description

* Set up a clippy command line that will actually fail the lint script
* Fix clippy errors that have appeared since our first time

## Type of change

- [ ] Code cleanups

# How Has This Been Tested?

All existing tests still build/run

# Checklist:

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -48,7 +48,7 @@ case $TARGET in
     ;;
 
   clippy)
-    cargo clippy
+    cargo clippy --workspace -- -D warnings -A clippy::unnecessary-cast
     ;;
 
   benchmark)
--- a/libs/traits/src/rewards.rs
+++ b/libs/traits/src/rewards.rs
@@ -107,7 +107,7 @@ where
 					Self::Balance::zero()
 				};
 
-				Self::reward_group(group_id.clone(), group_reward)
+				Self::reward_group(group_id, group_reward)
 			})
 			.collect())
 	}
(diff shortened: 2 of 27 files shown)


branan:
I think all the code owners got pinged on this one

ci/script.sh
```diff
@@ -48,7 +48,7 @@ case $TARGET in
     ;;
 
   clippy)
-    cargo clippy
+    cargo clippy --workspace -- -D warnings -A clippy::unnecessary-cast
```
branan, line 51: The allow here is because the benchmarking script spit out bad code when a lot of our weights were generated. This was fixed in https://github.com/paritytech/substrate/commit/c29095026871c2b7d42734226dd45db8051d258e, but I didn't want to regenerate all the benches as part of this PR.

I've filed #1201 to track this separately

libs/traits/src/rewards.rs
```diff
@@ -238,11 +238,15 @@ pub mod mock {
 				amount: <Self as AccountRewards<AccountId>>::Balance,
 			) -> DispatchResult;
 
+			// Return type is too messy for clippy, but not messy enough to really clean up
+			#[allow(clippy::type_complexity)]
```
branan, line 242: I believe we can set the threshold for clippy type  complexity, if we so desire. I feel like these `Result <Self as Trait>::Type, Error>` patterns crop up quite a bit in our codebase, and I don't think its worth making a `TypeOf<Trait>` alias for every single time it comes up.

NunoAlexandre (approved):
Looks good to me! Would like to see a Clippy CI job fail tho 😁

mustermeiszer (approved):
LGTM! But agreeing with Nuno that I would like to see the failure case. 

@branan wheb would this be?

wischli (approved):
LGTM!

libs/traits/src/rewards.rs
lemunozm: To add context here, the `mockall::mock!` macro forces you to write the extended trait path version (not sure why) instead of `Result<Self::Type, Error>`

branan:
I've opened #1202 with an intentionally broken commit so we can see it fail in github CI.

lemunozm (approved):

PR #1202, opened 2023-02-13
Break clippy

branan:
This is a demonstration PR for #1200. Will close once we see clippy fail


NunoAlexandre:
Thanks Branan 👍 closing now

PR #1209, opened 2023-02-17
feat: reward issuance interface

wischli:
# Description

In #1198 it became aware that we required a more flexible handling of issuing rewards. Since this can be separated nicely from the impl of `pallet-block-rewards`, here's the PR for it.

## Changes and Descriptions

* Before: Mint rewards inside `GroupRewards` impl for `pallet_rewards::Pallet`
* After: Handle rewarding via `RewardIssuance` trait [at the same place](https://github.com/centrifuge/centrifuge-chain/pull/1209/files#diff-55e041b69f2a93f848c9f74d5dad78f3e1a127b2b0764a28b9093fdb1fdb36bcR242)
* Provide two implementations for it which should handle all currently existing requirements for our two prod runtimes:
  * Altair: Use `TransferReward` which transfers rewards from the configured `AccountID`, e.g. `TreasuryAccount`
  * Centrifuge: Use `MintReward` which mints out of thin air
* Add simple unit tests for implementations
* Extend mock and tests for `pallet_rewards` to take both implementations into account 

### Example configuration for Altair

```rust
impl pallet_rewards::Config<pallet_rewards::Instance1> for Runtime {
        // ...
	type RewardIssuance = pallet_rewards::issuance::TransferReward<
		AccountId,
		Balance,
		CurrencyId,
		Tokens,
		TreasuryAccount,
	>;
}
```

### Example configuration for Centrifuge

```rust
impl pallet_rewards::Config<pallet_rewards::Instance1> for Runtime {
        // ...
	type RewardIssuance = pallet_rewards::issuance::MintReward<AccountId, Balance, CurrencyId, Tokens>;
}
```

## Type of change

# How Has This Been Tested?

```
cargo test -p pallet-rewards
```

# Checklist:


mustermeiszer (dismissed):
LGTM!

One thing to note. The issuing of rewards is not able to detect for which group it is issuing them, making all groups forced to use the same mechanism. 

Not sure if we on the future need to differentiate there

wischli:
Good point! AFAICT, we don't need to distinguish issuance for the same instance of `pallet_rewards` as the reward currency seems to be the native one in all cases (block and liquidity rewards). 

When we need to add it in the future, we most likely won't get around introducing storage for that.

mikiquantum (dismissed):
👍

mustermeiszer (approved):
Reapprove after fix

libs/traits/src/rewards.rs
```diff
@@ -182,6 +182,24 @@ pub trait CurrencyGroupChange {
...
+	type AccountId;
+
+	/// Type used to identify the currency
+	type CurrencyId;
+
+	/// Type used as balance for all currencies and reward.
+	type Balance;
+
+	/// Issue the provided reward amount to a beneficiary account address.
+	fn issue_reward(
```
lemunozm, line 196: I'm thinking that all implementers of these traits will call `issue_reward()` in all implementations of `reward_group()`. Could we extract this logic and make this call automatic?. For example, in line 110 of this same file, where `reward_group()` is called, we can use the returned balance to call `issue_reward()` automatically just in the next line. This way the implementater no longer needs to care to call this explicitely. WDYT?

pallets/rewards/src/lib.rs
```diff
@@ -230,7 +239,7 @@ pub mod pallet {
 			Groups::<T, I>::try_mutate(group_id, |group| {
 				let reward_to_mint = T::RewardMechanism::reward_group(group, reward)?;
 
-				T::Currency::mint_into(
+				T::RewardIssuance::issue_reward(
```
lemunozm, line 242: From my previous comment, we would no longer need this call here.

libs/traits/src/rewards.rs
```diff
@@ -182,6 +182,24 @@ pub trait CurrencyGroupChange {
...
 
+pub trait RewardIssuance {
+	/// Type used to identify the beneficiary.
+	type AccountId;
+
+	/// Type used to identify the currency
+	type CurrencyId;
+
+	/// Type used as balance for all currencies and reward.
+	type Balance;
```
lemunozm, line 193: Seeing the implementations of these traits that end up using generic for specifying the types, do we really need the associated types in the trait? Why not `RewardIssuance<AccountId, CurrencyId, Balance`>?

lemunozm (approved):
Very good PR! Clean and well organized, as always 🙌🏻 

Some comments below in form of a question/suggestions that can be ignored/postponed.

libs/traits/src/rewards.rs
mustermeiszer: I am not a big fan of using generics. It forces consumers to define them while associated types have the advantage of being "unknonw".

wischli: I suppose @lemunozm 's point is that the [two implementations](https://github.com/centrifuge/centrifuge-chain/pull/1209/files#diff-b8fa994533f9de709bf0781bb706a698f7afdadda84d688a82c48823227d7f0cR25-R26) of the trait use generics, so why not bubble that down? In the current version, I tried to continue the existing style.

mustermeiszer: I know, but it is no advantage to use generics.

libs/traits/src/rewards.rs
wischli: I am a fan of the idea, however it blows up the code quite significantly as we would have to introduce 
* `RewardCurrency`
* Truncated pallet account id
* `RewardIssuance`

to the `RewardMechanism` trait or `RewardMechanism::reward_group` call. Wanted to re-check with you before I start the endeavour. Or am I missing something here?

libs/traits/src/rewards.rs
lemunozm: My rule of thumb for this is:
- If the caller chose the types => generics
- If the implementer can chose the types => associated types

It's true that in Substrate most of the traits are implemented by the pallet that has the `Config` trait, and the associated types can be linked to any `T::Whatever`, so the implementer most of the time knows which types choose for the associated types.

In this case, because the implementer is not the pallet, it needs to expose them anyways as a generic.

libs/traits/src/rewards.rs
lemunozm: I'm not seeing why it requires to modify the reward mechanism, could you expand? 👀 

As I see this is adding the bound of `trait DistributedRewards: GroupRewards + RewardIssuance`, and then, in the default implementation of `distribute_reward_with_weights()`.

```rust
fn distribute_reward_with_weights() {
    // ...
    let (account, currency, balance) = Self::reward_group(group_id, group_reward);
    RewardIssuance::issue_reward(account, currency, balance);
    balance
    // ...
}
```

It can add more "code" in the sense that `reward_group()` should return the account and the currency too. But it should be the `reward_group()` from the pallet, not from the `RewardMechanism`.

lemunozm: Rethinking this, maybe instead of 
```rust 
trait DistributedRewards: GroupRewards + RewardIssuance
```
is better:
```rust
fn distribute_reward_with_weights<RewardIssuance>()
```

wischli: > It can add more "code" in the sense that reward_group() should return the account and the currency too. But it should be the reward_group() from the pallet, not from the RewardMechanism.

1. This was one of my points: How does `reward_group()` have knowledge of the recipient account and reward currency if we don't add this to the input arguments?

> I'm not seeing why it requires to modify the reward mechanism, could you expand? 👀

2.  In #1998, I haven't used `distribute_reward_with_weights` yet but `reward_group` instead, as we only have a single group. It appears to me this not desired, e.g. `distribute_reward_with_weights` should be called for the distribution as the name suggests?

wischli: > It can add more "code" in the sense that reward_group() should return the account and the currency too. But it should be the reward_group() from the pallet, not from the RewardMechanism.

1. This was one of my points: How does `reward_group()` have knowledge of the recipient account and reward currency if we don't add this to the input arguments?

> I'm not seeing why it requires to modify the reward mechanism, could you expand? 👀

2.  In #1998, I haven't used `distribute_reward_with_weights` yet but `reward_group` instead, as we only have a single group. It appears to me this not desired, e.g. `distribute_reward_with_weights` should be called for the distribution as the name suggests?

lemunozm: > I haven't used distribute_reward_with_weights yet but reward_group instead

Ok, I think this is the main point. If `distribute_reward_with_weights` (or `distribute_reward`) is not used, all that I said lose sense 😅 . I always called to `distribute_reward` directly and I forgot it wasn't the same case here. 🙏🏻

libs/traits/src/rewards.rs
mustermeiszer: Can't the implementer always choose the type?

The caller can always enforce the type via `TraitA<Associated1=...>`. My personal opinion is just that generics are less flexible while providing no real advantage. Unless you wanna allow multiple implementations of this trait for the same implementer.

mustermeiszer: Shouldn't have started this. ^^ I actually don't have strong such strong feelings about that, so do what works best for you.

wischli: I propose to keep as is.

libs/traits/src/rewards.rs
wischli: It appears to me that we want pallets which use `pallet-rewards` to actually call `distribute_reward` instead of `reward_group` right? Then I would go forward with refactoring to stay consistent.

lemunozm: > It appears to me that we want pallets which use pallet-rewards to actually call distribute_reward instead of reward_group right?

Yes, that was at least what I had in mind. But it's true that your use case can bypass the `distribute_reward()` call without damage. For me, it's ok as it is 👍🏻 . Maybe we can change this in the future if `distribute_reward() has more relevance.

PR #1213, opened 2023-02-24
permissions: Add PODReadAccess permission

cdamian:
# Description

Added `PODReadAccess` permission.

## Type of change

# How Has This Been Tested?

- [x] Permissions unit tests

# Checklist:

### DIFF
--- a/libs/types/src/permissions.rs
+++ b/libs/types/src/permissions.rs
@@ -39,6 +39,7 @@ pub enum PoolRole<TrancheId = [u8; 16], Moment = u64> {
 	MemberListAdmin,
 	LoanAdmin,
 	TrancheInvestor(TrancheId, Moment),
+	PODReadAccess,
 }
 
 #[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, TypeInfo, Debug)]
@@ -93,6 +94,7 @@ bitflags::bitflags! {
 		const LIQUIDITY_ADMIN = 0b00001000;
 		const MEMBER_LIST_ADMIN = 0b00010000;
 		const RISK_ADMIN = 0b00100000;
+		const POD_READ_ACCESS = 0b01000000;
 	}
 
 	/// The current admin roles we support
@@ -220,6 +222,9 @@ where
 				}
 				PoolRole::LoanAdmin => self.pool_admin.contains(PoolAdminRoles::RISK_ADMIN),
 				PoolRole::TrancheInvestor(id, _) => self.tranche_investor.contains(id),
+				PoolRole::PODReadAccess => {
+					self.pool_admin.contains(PoolAdminRoles::POD_READ_ACCESS)
+				}
 			},
 			Role::PermissionedCurrencyRole(permissioned_currency_role) => {
 				match permissioned_currency_role {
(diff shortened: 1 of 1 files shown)


libs/types/src/permissions.rs
```diff
@@ -93,6 +94,7 @@ bitflags::bitflags! {
 		const LIQUIDITY_ADMIN = 0b00001000;
 		const MEMBER_LIST_ADMIN = 0b00010000;
 		const RISK_ADMIN = 0b00100000;
+		const POD_ADMIN = 0b01000000;
```
offerijns, line 97: I think this is a bit confusing as it's not an admin role, would just do:
```suggestion
		const POD_READ_ACCESS = 0b01000000;
```

cdamian: Agree, changed it.

libs/types/src/permissions.rs
```diff
@@ -258,6 +263,7 @@ where
 				}
 				PoolRole::LoanAdmin => Ok(self.pool_admin.remove(PoolAdminRoles::RISK_ADMIN)),
 				PoolRole::TrancheInvestor(id, delta) => self.tranche_investor.remove(id, delta),
+				PoolRole::PODReadAccess => self.pool_admin.remove(PoolAdminRoles::POD_READ_ACCESS),
```
offerijns, line 266: Comparing to the rest of the implementations, it seems like this should be:
```suggestion
				PoolRole::PODReadAccess => Ok(self.pool_admin.remove(PoolAdminRoles::POD_READ_ACCESS)),
```

libs/types/src/permissions.rs
```diff
@@ -289,6 +295,7 @@ where
 				}
 				PoolRole::LoanAdmin => Ok(self.pool_admin.insert(PoolAdminRoles::RISK_ADMIN)),
 				PoolRole::TrancheInvestor(id, delta) => self.tranche_investor.insert(id, delta),
+				PoolRole::PODReadAccess => self.pool_admin.insert(PoolAdminRoles::POD_READ_ACCESS),
```
offerijns, line 298: Likewise
```suggestion
				PoolRole::PODReadAccess => Ok(self.pool_admin.insert(PoolAdminRoles::POD_READ_ACCESS)),
```

libs/types/src/permissions.rs
```diff
@@ -289,6 +295,7 @@ where
 				}
 				PoolRole::LoanAdmin => Ok(self.pool_admin.insert(PoolAdminRoles::RISK_ADMIN)),
```
offerijns, line 296: This is unrelated to this PR, but the naming seems a bit off here, not all pool roles are pool admin roles. I wonder if we should change `self.pool_admin` to `self.pool` and `PoolAdminRoles` to just `PoolRoles`. What do you think?

cc @mustermeiszer

libs/types/src/permissions.rs
cdamian: Oh, sorry for missing this.

libs/types/src/permissions.rs
mustermeiszer: Sounds sane to me. I would be more pro having an additional struct like `PoolAdminRoles` and `PoolRoles` but this would mean we need to write a migration. So lets keep it simple.

mustermeiszer (approved):
Looks good to me!

PR #1220, opened 2023-03-01
Add `mock-builder` to the repo with pools & permissions instances

lemunozm:
# Description

First functional implementation of [mocking proposal](https://centrifuge.hackmd.io/bJSKpKK3SPCcwZRK13SFLA?view). The second iteration will add a procedural macro to remove the boilerplate, but for now, it is usable.

As an overview, it allows implementing customizable pallet mocks for generic traits. Exposing a user callback that the tester can configure in each test case.

Check the tests from https://github.com/centrifuge/centrifuge-chain/pull/1180 as a live example.

## Changes and Descriptions

This PR adds:
- `mock-builder` crate to easily build generic mocks with user callbacks.
- `cfg-mocks` crate where our specific mocks using `mock-buider` are placed. *Once the second iteration is done, we could consider removing this crate and building the mocks in place using the macro in the projects that need it.*

## Type of change

- New *testing* feature

# How Has This Been Tested?

```sh
cargo test -p mock-builder
cargo check -p cfg-mocks
```

# Checklist:


libs/mocks/src/permissions.rs
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		pub fn mock_add(f: impl Fn(T::Scope, T::AccountId, Role) -> bool + 'static) {
+			register_call!(move |(a, b, c)| f(a, b, c));
+		}
+
+		pub fn mock_remove(f: impl Fn(T::Scope, T::AccountId, Role) -> bool + 'static) {
+			register_call!(move |(a, b, c)| f(a, b, c));
+		}
+	}
```
mustermeiszer, line 40: Could also imagine something like
```suggestion
       #[mock_builder::builder]
	impl<T: Config> Pallet<T> {
	        #[builder::register]
		pub fn mock_has(f: impl Fn(T::Scope, T::AccountId, Role) -> bool;

		pub fn mock_add(f: impl Fn(T::Scope, T::AccountId, Role) -> bool + 'static) {
			register_call!(move |(a, b, c)| f(a, b, c));
		}

		pub fn mock_remove(f: impl Fn(T::Scope, T::AccountId, Role) -> bool + 'static) {
			register_call!(move |(a, b, c)| f(a, b, c));
		}
	}
```

mustermeiszer (commented):
I like the first draft also I think 
* many things can be moved to an procedural macro
  * the impl of the trait
  * probably also register logic
* the mock pallet can also mostly be implemented by a procedural macro 
  * `CallIds` storage mostly
 
 In general I think the idea is great. I would also not be against merging this up now and refine later. 
 
 But I think the `mock-builder` makes to many assumptions about the namings which I hope we eliviate with the procedural macros. Maybe, as these kind of mocks basically just create infrastructure so that consumers of this mock can set their desired return values, we can think about adding this as part of an implementation of a pallet. Via a flag `as_mock`

libs/mocks/src/permissions.rs
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		fn add(a: Self::Scope, b: T::AccountId, c: Self::Role) -> DispatchResult {
+			execute_call!((a, b, c))
+		}
+
+		fn remove(a: Self::Scope, b: T::AccountId, c: Self::Role) -> DispatchResult {
+			execute_call!((a, b, c))
+		}
+	}
```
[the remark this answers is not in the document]
lemunozm: I love this schema a lot! Especially this idea: `#[mock::call(mock_remove)]`😍, it makes explicit the name of the mock method. 

I would also add the whole pallet creation part as another macro (where `CallIds` is generated). I see 3 levels: pallet, trait, and call. Something like:
```rust
// build the pallet boilerplate
#[mock_builder::pallet] 
mod my_pallet_mock {
    //What you wrote:
    #[mock::trait(Permissions<T::AccountId>)]
       type Error = DispatchError;
       type Ok = ();
       type Role = Role;
       type Scope = T::Scope;

       #[mock::call(mock_has)]
       fn has(a: Self::Scope, b: T::AccountId, c: Self::Role) -> bool;

       #[mock::call(mock_add)]
       fn add(a: Self::Scope, b: T::AccountId, c: Self::Role) -> DispatchResult;

       #[mock::call(mock_remove)]
       fn remove(a: Self::Scope, b: T::AccountId, c: Self::Role) -> DispatchResult;

    #[mock::trait(OtherTrait)] 
    // Another trait for the same pallet here
    //...
}

File: libs/mock-builder/src/lib.rs [hunk 3]
```diff
@@ -0,0 +1,213 @@
...
+/// Register a call into the call storage.
+/// This macro should be called from the method that wants to register `f`.
+/// This macro must be called from a pallet with the `CallIds` storage.
+/// Check the main documentation.
+#[macro_export]
+macro_rules! register_call {
+	($f:expr) => {{
+		use frame_support::StorageHasher;
+
+		CallIds::<T>::insert(
```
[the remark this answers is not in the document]
lemunozm: Yes, I totally agree. My intention was to leave this macro to the second iteration. Basically, my idea is:
1. First iteration, make this functional from tests
2. Second iteration, make it easy to generate with macro/s

libs/mock-builder/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+	pub type CallId = u64;
+
+	trait Callable {
+		fn as_any(&self) -> &dyn Any;
+	}
+
+	thread_local! {
+		static CALLS: RefCell<HashMap<CallId, Box<dyn Callable>>>
+			= RefCell::new(HashMap::default());
+	}
```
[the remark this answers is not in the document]
lemunozm: The `CallId` is always different in each addition to the storage. No matter who uses it. So I swear not.

libs/mocks/src/permissions.rs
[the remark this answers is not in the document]
lemunozm: Yes! That's the idea. The problem here is that each call could have different arguments. We need that the macro "identify" those arguments in order to place them correctly in the `register_call!()`.

libs/mocks/src/permissions.rs
```diff
@@ -0,0 +1,60 @@
+pub use pallet_mock_permissions::*;
+
+#[allow(dead_code)]
+#[frame_support::pallet]
+pub mod pallet_mock_permissions {
+	use cfg_traits::Permissions;
+	use cfg_types::permissions::Role;
```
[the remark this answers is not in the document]
lemunozm: Not sure here what you refer to, but if you mean what you wrote in the example below, then yes, I agree.

libs/mock-builder/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+/// Gives the absolute string identification of a function.
+#[macro_export]
+macro_rules! function_locator {
+	() => {{
+		fn f() {}
+		fn type_name_of<T>(_: T) -> &'static str {
+			std::any::type_name::<T>()
+		}
+		let name = type_name_of(f);
+		&name[..name.len() - 3]
```
wischli, line 102: Where does the `3` come from?

libs/mocks/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+std = [
+  "codec/std",
+  "scale-info/std",
+  "frame-support/std",
+  "frame-system/std",
+  "sp-std/std",
+  "cfg-primitives/std",
+  "cfg-traits/std",
+  "cfg-types/std",
+]
```
wischli, line 36: As a result of #1223, I can claim the `try-runtime` feature is missing here 🙃

```suggestion
]
try-runtime = [
  "frame-support/try-runtime",
  "frame-system/try-runtime",
  "cfg-primitives/try-runtime",
  "cfg-traits/try-runtime",
  "cfg-types/try-runtime",
]

```

wischli (commented):
For a PoC this looks good to me! Definitely agree with the points made by @mustermeiszer, though I understand this might rather be handled in a subsequent PR. 

I have to admit I am lacking macro knowledge so treat my review like a feather in the wind.

mustermeiszer:
Then let's move the first iteration on and iterate on that one. 

I think for the grant we need to move it to its own GitHub repo anyways.

mustermeiszer (approved):

libs/mocks/Cargo.toml
lemunozm: Good point!

Only a last confirmation, `mocks` should only be used when testing in mocks, under `#[cfg(test)]` feature. Does add `try-runtime` applies in this context?

lemunozm: I've finally added them. It does not add damage. Although `cfg-traits` and `cfg-primitives` does not have that feature until #1223 is merged.

wischli: Thanks! True that. Just hoped #1223 would get merged first since rebasing requires re-approvals from all code-owners 🥲

wischli: > Only a last confirmation @wischli , `mocks` should only be used when testing in mocks, under `#[cfg(test)]` feature, and added as a `[dev-dependencies]`. Does add `try-runtime` applies in this context? I mean, I think nobody should use this crate with `try-runtime`

My understanding is that any crate which imports `frame-support` as dependency requires adding its `try-runtime` feature and hence the same for any other dependencies which also support `try-runtime`.

lemunozm: Thanks for clarify this 👍🏻

mustermeiszer (approved):
Tests are lovely!

PR #1221, opened 2023-03-01
Interest accrual API only with yearly rates

### LINKED ISSUE #1189: Improve type-safety in interest-accrual rate handling

* [ ] Addressing a bug
* [x] Refactoring
* [ ] Feature

### Description

From https://github.com/centrifuge/centrifuge-chain/pull/1157#issuecomment-1404666267

>  WDYT about adding type checking for yearly and secondly interest rates? I feel like it is way to easy to misuse the interface of accrual.

Improves readability and usability of internal APIs.

This is a purely internal API change that isn't planned to affect storage or user-facing APIs.

lemunozm:
# Description

Modify `InterestAccrual` to expose only an `interest_rate_per_year` parameter. No more interest rates per sec to handle. UI free from converting rates to show the correct yearly rates.

Fixes #1189

## Changes and Descriptions

- The same storage is used. No migration is required for `interest-accrual`.
- The rate representation changes on the `pallet-loans` side, the type is the same but the value is stored diferently, so **it will require a migration if centrifuge release is launched before this**.
- This yearly simplification allows removing a couple of methods from `InterestAccrual` trait.

## Requirements
- [x] Blocked by https://github.com/centrifuge/centrifuge-chain/pull/1266

### DIFF
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -233,7 +233,7 @@ pub trait InterestAccrual<InterestRate, Balance, Adjustment> {
 
 	/// Calculate the current debt using normalized debt * cumulative rate
 	fn current_debt(
-		interest_rate_per_sec: InterestRate,
+		interest_rate_per_year: InterestRate,
 		normalized_debt: Self::NormalizedDebt,
 	) -> Result<Balance, DispatchError>;
 
@@ -244,14 +244,14 @@ pub trait InterestAccrual<InterestRate, Balance, Adjustment> {
 	/// (effectively "rewinding the clock" to before the value was
 	/// valid)
 	fn previous_debt(
-		interest_rate_per_sec: InterestRate,
+		interest_rate_per_year: InterestRate,
 		normalized_debt: Self::NormalizedDebt,
 		when: Moment,
 	) -> Result<Balance, DispatchError>;
 
 	/// Increase or decrease the normalized debt
 	fn adjust_normalized_debt(
-		interest_rate_per_sec: InterestRate,
+		interest_rate_per_year: InterestRate,
 		normalized_debt: Self::NormalizedDebt,
 		adjustment: Adjustment,
 	) -> Result<Self::NormalizedDebt, DispatchError>;
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -263,23 +263,11 @@ pub trait InterestAccrual<InterestRate, Balance, Adjustment> {
 		normalized_debt: Self::NormalizedDebt,
 	) -> Result<Self::NormalizedDebt, DispatchError>;
 
-	/// Indicate that a yearly rate is in use
-	///
-	/// Validates that the rate is allowed, and converts it to a per-second rate for future operations
-	fn reference_yearly_rate(
-		interest_rate_per_year: InterestRate,
-	) -> Result<InterestRate, DispatchError>;
-
-	/// Indicate that a rate is in use
-	fn reference_rate(interest_rate_per_sec: InterestRate) -> DispatchResult;
+	/// Validate and indicate that a yearly rate is in use
+	fn reference_rate(interest_rate_per_year: InterestRate) -> DispatchResult;
 
 	/// Indicate that a rate is no longer in use
-	fn unreference_rate(interest_rate_per_sec: InterestRate) -> DispatchResult;
-
-	/// Verifies a yearly additive rate and converts it to a per-second additive rate
-	fn convert_additive_rate_to_per_sec(
-		interset_rate_per_year: InterestRate,
-	) -> Result<InterestRate, DispatchError>;
+	fn unreference_rate(interest_rate_per_year: InterestRate) -> DispatchResult;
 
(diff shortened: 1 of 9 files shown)


libs/traits/src/lib.rs
```diff
@@ -249,23 +249,11 @@ pub trait InterestAccrual<InterestRate, Balance, Adjustment> {
...
-	///
-	/// Validates that the rate is allowed, and converts it to a per-second rate for future operations
-	fn reference_yearly_rate(
-		interest_rate_per_year: InterestRate,
-	) -> Result<InterestRate, DispatchError>;
-
-	/// Indicate that a rate is in use
-	fn reference_rate(interest_rate_per_sec: InterestRate) -> DispatchResult;
+	/// Validate and indicate that a yearly rate is in use
+	fn reference_rate(interest_rate_per_year: InterestRate) -> DispatchResult;
```
lemunozm, line 253: `reference_yearly_rate` and `reference_rate` mixed into one

pallets/interest-accrual/src/lib.rs
```diff
@@ -480,28 +468,12 @@ impl<T: Config> InterestAccrual<T::InterestRate, T::Balance, Adjustment<T::Balan
...
-	}
-
-	fn convert_additive_rate_to_per_sec(
-		interest_rate_per_year: T::InterestRate,
-	) -> Result<T::InterestRate, DispatchError> {
-		Pallet::<T>::validate_rate(interest_rate_per_year)?;
-		let interest_rate_per_sec = interest_rate_per_year
-			.checked_div(&T::InterestRate::saturating_from_integer(SECONDS_PER_YEAR))
-			.ok_or(ArithmeticError::Underflow)?;
-		Ok(interest_rate_per_sec)
```
lemunozm, line 504: This method is no longer needed because the addition is now at year level. The resulting rate will be referenced as before and there evaluated by `validate_rate`.

pallets/interest-accrual/src/lib.rs
```diff
@@ -433,8 +422,7 @@ pub mod pallet {
 		pub(crate) fn validate_rate(interest_rate_per_year: T::InterestRate) -> DispatchResult {
 			let four_decimals = T::InterestRate::saturating_from_integer(10000);
 			ensure!(
-				interest_rate_per_year < One::one()
-					&& interest_rate_per_year > Zero::zero()
+				interest_rate_per_year > Zero::zero()
```
lemunozm, line 425: I've needed to remove the upper limit because add rates overpass this limitation. In fact, I do not understand why this upper limit

lemunozm (commented):
First iteration review. Currently, only the accrual pallet is compiling. If you agree with this structure, I will change the rest of the crates that are dependent on it. 

Self-comments about the most tricky parts:

lemunozm:
@branan, are you ok with this schema? Can I continue with this?

branan:
Yeah, this looks good :+1:

pallets/interest-accrual/src/lib.rs
lemunozm: > Why this upper limitation?

Not sure exactly what should be a good upper limitation, but we need to set one to avoid the user entering here a high value that causes later an overflow in the `pow`.

pallets/interest-accrual/src/lib.rs
```diff
@@ -490,18 +473,21 @@ pub mod pallet {
...
 		}
 
-		pub(crate) fn validate_rate(interest_rate_per_year: T::InterestRate) -> DispatchResult {
+		pub(crate) fn validate_interest_rate(
+			interest_rate_per_year: T::InterestRate,
+		) -> DispatchResult {
 			let four_decimals = T::InterestRate::saturating_from_integer(10000);
 			ensure!(
-				interest_rate_per_year < One::one()
+				interest_rate_per_year <= T::InterestRate::saturating_from_integer(2)
```
lemunozm, line 490: This would require another issue to tackle correctly. This `2` comes because is the "maximum" reasonable value coming from `pallet-loans`, which is a possible interest rate `< 1` plus a penalty from `[0, 1]`. Which in the worst cases could be near to `2`.

Theoretically, there is no upper bound for this, nevertheless, technically, the correct value that should be here is the maximum rate value that does not cause overflow after the whole life of centrifuge change accruing that rate (1000 years for example?).

lemunozm:
**If we get this merged before the release, we avoid these migrations: f077df955182696681dd8c8505cbb3599c1f6ac5**

pallets/interest-accrual/src/lib.rs
offerijns: Maybe add a constant `MAX_INTEREST_RATE = 2`, and add
```rust
// Possible interest rates < 1 plus a penalty from [0, 1]
// Which in the worst cases could be near to 2.
```
as a comment to this constant?

offerijns (approved):
LGTM!

pallets/interest-accrual/src/lib.rs
lemunozm: Agree!

offerijns (approved):

PR #1222, opened 2023-03-02
ChangeGuard definition with loans implementation

### LINKED ISSUE #1207: Implementation of `ChangeGuard` trait

* Feature

### Description
```rust
/// Trait for checking changes before allowing them to be performed.
trait PoolChangeGuard {
   /// Associated pool where evaluate the change
   type PoolId;

   /// Reference a `Change` waiting to be released.
   type ChangeId;

   /// Information about the change the caller wants to do.
   type Change;

   /// Notify a `change` that affect somehow to `pool_id`.
   /// It will be "blocked" and referenced by the returned `ChangeId` until the conditions
   /// will be ok to allow the caller of this trait to perform the desired action.
   fn note(pool_id: PoolId, change: Self::Change) -> Result<Self::ChangeId, DispatchError>;

   /// Ask for a `change_id` if it's ready to proceed.
   /// An error will be returned if:
   /// - The change is not ready yet.
   /// - The change was already released.
   /// - The change has expired.
   /// If `Ok()`, the caller can proceed.
   fn released(change_id: Self::ChangeId) -> DispatchResult;
}
```

The implementation of this trait amalgamates changes from multiple locations of the codebase.

### Research/based on

- Currently, there is a similar trait called `PoolUpdateGuard`.
- Slack [Thread](https://kflabs.slack.com/archives/C04HJN404RK/p1674572242912299) of the conversation.
- Task that will make use of this: #1195

- The `Change` type chosen by the implementation should contain information about the time.
We need to decide what is the relevant information to specify a `Change`.

lemunozm:
Fixes #1207

## Changes

- [x] Added `ChangeGuard` trait definition
- [x] Added `Change` types (basic implementation without `amalgamator` or `subenum` utilities).
- [x] Implemented `propose_modification()` and `modify()` extrinsics.
- [x] Add `ChangeGuard` mock
- [ ] Add tests 
- [ ] Add benchmarks (tests)
- [ ] Resolve TODOs

*ChangeGuard **implementation** is not in this PR and I think it should be in a new one.*

## Preconditions
- https://github.com/centrifuge/centrifuge-chain/pull/1180 is merged

### DIFF
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -41,6 +41,9 @@ pub mod ops;
 /// Traits related to rewards.
 pub mod rewards;
 
+/// Traits related to checked changes.
+pub mod changes;
+
 /// A trait used for loosely coupling the claim pallet with a reward mechanism.
 ///
 /// ## Overview
--- a/libs/types/src/changes.rs
+++ b/libs/types/src/changes.rs
@@ -1,0 +1,38 @@
+use codec::{Decode, Encode, MaxEncodedLen};
+use frame_support::RuntimeDebug;
+use scale_info::TypeInfo;
+
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub enum CfgChange {
+	Loan(LoanChange),
+	Pool(PoolChange),
+}
+
+impl From<LoanChange> for CfgChange {
+	fn from(value: LoanChange) -> Self {
+		CfgChange::Loan(value)
+	}
+}
+
+impl From<PoolChange> for CfgChange {
+	fn from(value: PoolChange) -> Self {
+		CfgChange::Pool(value)
+	}
+}
+
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub enum LoanChange {
+	Maturity,
+	InterestRate,
+	InterestPayments,
+	PayDownSchedule,
+	ValuationMethod,
+	ProbabilityOfDefault,
+	LossGivenDefault,
+	DiscountRate,
+}
+
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub enum PoolChange {
+	// Unimplemented
+}
--- a/pallets/loans-ref/src/lib.rs
+++ b/pallets/loans-ref/src/lib.rs
@@ -17,14 +17,16 @@
 //!
 //! The following actions are performed over loans:
 //!
-//! | Extrinsics                    | Role      |
-//! |-------------------------------|-----------|
-//! | [`Pallet::create()`]          | Borrower  |
-//! | [`Pallet::borrow()`]          | Borrower  |
-//! | [`Pallet::repay()`]           | Borrower  |
-//! | [`Pallet::write_off()`]       | Any       |
-//! | [`Pallet::admin_write_off()`] | LoanAdmin |
-//! | [`Pallet::close()`]           | Borrower  |
+//! | Extrinsics                         | Role      |
+//! |------------------------------------|-----------|
+//! | [`Pallet::create()`]               | Borrower  |
+//! | [`Pallet::borrow()`]               | Borrower  |
+//! | [`Pallet::repay()`]                | Borrower  |
+//! | [`Pallet::write_off()`]            | Any       |
+//! | [`Pallet::admin_write_off()`]      | LoanAdmin |
+//! | [`Pallet::close()`]                | Borrower  |
+//! | [`Pallet::propose_modification()`] | LoanAdmin |
+//! | [`Pallet::modify()`]               | Borrower  |
 //!
 //! The following actions are performed over a pool of loans:
 //!
@@ -63,11 +65,13 @@ pub use weights::WeightInfo;
 pub mod pallet {
 	use cfg_primitives::Moment;
 	use cfg_traits::{
+		changes::ChangeGuard,
 		ops::{EnsureAdd, EnsureAddAssign, EnsureInto},
 		InterestAccrual, Permissions, PoolInspect, PoolNAV, PoolReserve,
 	};
 	use cfg_types::{
 		adjustments::Adjustment,
+		changes::CfgChange,
 		permissions::{PermissionScope, PoolRole, Role},
 	};
 	use frame_support::{
@@ -90,7 +94,7 @@ pub mod pallet {
 	use sp_std::vec::Vec;
 	use types::{
 		self, ActiveLoan, AssetOf, BorrowLoanError, CloseLoanError, CreateLoanError, LoanInfoOf,
-		PortfolioValuationUpdateType, WrittenOffError,
+		LoanMutation, ModificationError, Mutation, PortfolioValuationUpdateType, WrittenOffError,
 	};
 	use write_off::{WriteOffRule, WriteOffStatus};
 
@@ -114,6 +118,9 @@ pub mod pallet {
 
 		/// Identify a curreny.
 		type CurrencyId: Parameter + Copy + MaxEncodedLen;
+
+		/// Identify a loan change.
+		type ChangeId: Parameter + Copy + MaybeSerializeDeserialize + MaxEncodedLen + TypeInfo;
 
 		/// Identify a non fungible collection
 		type CollectionId: Parameter
@@ -179,6 +186,13 @@ pub mod pallet {
 			Self::Balance,
 			Adjustment<Self::Balance>,
 			NormalizedDebt = Self::Balance,
+		>;
+
+		/// Used to confirm active loan modification properties.
+		type ChangeGuard: ChangeGuard<
+			PoolId = PoolIdOf<Self>,
+			ChangeId = Self::ChangeId,
+			Change = CfgChange,
 		>;
 
 		/// Max number of active loans per pool.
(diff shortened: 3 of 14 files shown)


pallets/loans-ref/src/types.rs
```diff
@@ -622,6 +676,41 @@ impl<T: Config> ActiveLoan<T> {
 		Ok(())
 	}
 
+	fn ensure_can_modify(&self, mutation: &Mutation<T::Rate>) -> DispatchResult {
+		// What happens if it has overdue? or has been repaid or written off?
```
lemunozm, line 680: Suppose the following:

I want to modify the interest rate.
1. I propose the change X
2. While time passes, it is repaid/reborrowed/written-off,
3. The change is ready to be applied.

Depending on change X or any amount of actions performed during that time, the answer of "what to do when the change is applied" could be different.

libs/traits/src/lib.rs
```diff
@@ -324,6 +324,38 @@ pub trait PoolUpdateGuard {
...
+	type Change;
+
+	/// Notify a `change` related to a `pool_id`.
+	/// The caller to this method ask for feedback for the implementation of this trait in order
+	/// be able to semantically proceed successful with that change.
+	/// The change intention will be noted by this method and identified by the returned ChangeId.
+	fn note<C: Into<Self::Change>>(
+		pool_id: Self::PoolId,
+		change: C,
+	) -> Result<Self::ChangeId, DispatchError>;
```
mustermeiszer, line 347: why generic over `impl`

libs/types/src/changes.rs
```diff
@@ -0,0 +1,38 @@
...
+	ValuationMethod,
+	ProbabilityOfDefault,
+	LossGivenDefault,
+	DiscountRate,
+}
+
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub enum PoolChange {
+	// Unimplemented
+}
```
mustermeiszer, line 38: ❤️

pallets/loans-ref/src/lib.rs
```diff
@@ -523,6 +565,33 @@ mod pallet {
...
+			origin: OriginFor<T>,
+			pool_id: PoolIdOf<T>,
+			change_id: T::ChangeId,
+		) -> DispatchResult {
+			ensure_signed(origin)?;
+
+			T::ChangeGuard::released(pool_id, change_id)?;
+
+			let modification = LoanModifications::<T>::take(pool_id, change_id)
+				.ok_or(DispatchError::Other("Expected a stored modification"))?;
```
mustermeiszer, line 580: I  would use a proper error at this level

pallets/loans-ref/src/lib.rs
```diff
@@ -584,6 +653,41 @@ mod pallet {
...
+		#[transactional]
+		pub fn propose_modification(
+			origin: OriginFor<T>,
+			pool_id: PoolIdOf<T>,
+			loan_id: T::LoanId,
+			mutation: Mutation<T::Rate>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			// TODO: check the correct permission for each LoanChange.
```
mustermeiszer, line 668: Maybe we need another role? or it can be the borrower cc @offerijns

libs/types/src/changes.rs
```diff
@@ -0,0 +1,38 @@
...
+pub enum LoanChange {
+	Maturity,
+	InterestRate,
+	InterestPayments,
+	PayDownSchedule,
+	ValuationMethod,
+	ProbabilityOfDefault,
+	LossGivenDefault,
+	DiscountRate,
+}
```
mustermeiszer, line 33: Although, I am not sure, if the `ChangeGuard` needs more context. 

I am always not sure, if we wanna store the change on the guard side or on the executing side...

mustermeiszer: WDYT?

pallets/loans-ref/src/types.rs
mustermeiszer: Given the current changes, I am not sure this is a problem.

mustermeiszer: The only thing I am not sure is interest rate if the rate has been adated through write-off. We could check that right? If the write-off-status of a loan is set

pallets/loans-ref/src/types.rs
```diff
@@ -635,6 +689,41 @@ impl<T: Config> ActiveLoan<T> {
...
+					.probability_of_default = rate
+			}
+			Mutation::LossGivenDefault(rate) => {
+				self.info
+					.mut_validation_method_dcf::<T>()?
+					.loss_given_default = rate
+			}
+			Mutation::DiscountRate(rate) => {
+				self.info.mut_validation_method_dcf::<T>()?.discount_rate = rate
+			}
```
mustermeiszer, line 721: We should definietly update the valuation here

mustermeiszer (commented):
Just a first scan and feedback.

pallets/loans-ref/src/lib.rs
offerijns: Currently we had discussed that this would be the `LoanAdmin` role. I need to think about this once more though to double check it makes sense, but let's go with that for now

pallets/loans-ref/src/lib.rs
lemunozm: I did it in this way because, logically, this error can never happen and, therefore, can not be tested (not belong to the model). The `released` method will fail if the pair `pool_id`, `change_id` does not exist, so if you reach `LoanModifications::take()`, it always contains something. We could use an `unwrap()` here too but it's not "safe".

I tried to rationale this more deeply in our (maybe "my" at this moment) [rust guidelines](https://centrifuge.hackmd.io/sLMxV5WoTy2q57fJTfbNLQ?view#Proposal)

libs/traits/src/lib.rs
lemunozm: I tried `impl` before, but I had problems mocking these kinds of methods; I have to investigate it more deeply.

mustermeiszer: Then don't bother. Will compiler will defer the generic anyways. Just the signature looks nicer, but who am I telling this, your code looks gorgeous!

pallets/loans-ref/src/types.rs
lemunozm: It's already done because `modify()` is called within the `update_active_loan()`

pallets/loans-ref/src/lib.rs
lemunozm: From here: https://docs.google.com/spreadsheets/d/1RJ5RLobAdumXUK7k_ugxy2eDAwI5akvtuqUM2Tyn5ts/edit?pli=1#gid=0

It seems like some changes depends on `LoanAdmin` and others on `PoolAdmin`

libs/traits/src/lib.rs
lemunozm: I totally agree with the `impl`. Thanks for your comment!

libs/types/src/changes.rs
mustermeiszer: I am not 100-% sure either. Let's see in the next weeks. Not urgent

pallets/loans-ref/src/types.rs
lemunozm: Yes, in fact, I think currently I'm not doing it, and I should.

lemunozm: Done!

lemunozm:
To summarize pending open questions:
- Should `ChangeGuard` implementation knows that has been a simple a "loan change" or should it know which kind of change? Or even should it know the value of the variant that has changed? i.e:
  1. `ChangeGuard` implementation knows about `LoanChange`
  1. `ChangeGuard` implementation knows about `LoanChange::InterestRate`
  1. `ChangeGuard` implementation knows about `LoanChange::InterestRate(Rate)`

- Each change has different waiting times to get it done. That "waiting time" knowledge should be known by the `pallet-loans` and added as a parameter to `ChangeGuard` trait, or should the implementation of `ChangeGuard` who knows about this?
  - My first feeling is that it should be the first one, and we need to add a `Moment` parameter, but some parameters depend on epochs, and `pallet-loans` doesn't know about epochs, so it seems like `ChangeGuard` should know this info. Also, if `ChangeGuard` knows about this, it should distinguish which kind of loan change needs to be, to know if it should wait an epoch or not, so it should implement at least the second above option

lemunozm:
Regarding the last point, maybe we can add a parameter with an enum as follows:
```rust
enum Waiting {
   Fixed(Moment),
   UntilNextEpoch,
}

fn note(pool_id: PoolId, change: Change, waiting: Waiting)
```

With this, the `ChangeGuard` implementation only needs to know if it was a `LoanChange`, not exactly what change

libs/types/src/changes.rs
lemunozm: I left some thoughts about this in the issue thread: #1207

pallets/loans-ref/src/lib.rs
lemunozm: @mustermeiszer, which should check this permission, `pallet-loans` or `ChangeGuard` implementation?

libs/traits/src/lib.rs
lemunozm: `impl` is now supported, but we still need the `'static` lifetime, until #1322

pallets/loans-ref/src/lib.rs
lemunozm: > It seems like some changes depend on LoanAdmin and others on PoolAdmin

I think here you're right @offerijns and any property of a loan should be associated to a `LoanAdmin`.

> @mustermeiszer, which should check this permission, pallet-loans or ChangeGuard implementation?

This question no longer applies if we choose to have only one role

lemunozm:
Closed in favor of #1384

Issue #1226, opened 2023-03-03
Add pools proxy types to Centrifuge runtime


branan:
### Which part of the code is the issue addressing?

* [x] Feature

### Description

Add the following proxy types from the Development runtime to the Centrifuge runtime:

```
	Borrow,
	Price,
	Invest,
	ProxyManagement,
	KeystoreManagement,
	PodOperation,
	PodAuth,
	PermissionManagement,
```

Issue #1225, opened 2023-03-03
Configure restricted-tokens in Centrifuge runtime to handle tranche tokens


branan:
### Which part of the code is the issue addressing?

* [x] Feature

### Description

The `pallet_restricted_tokens` configuration for Centrifuge will need to be updated to handle tranche tokens. Its current configuration allows all transactions.

The configuration can be copied from the `RestrictedTokens` type in the development runtime

PR #1233, opened 2023-03-07
Add pools to centrifuge runtime

### LINKED ISSUE #1224: Add pools pallets to Centrifuge runtime
* [x] Feature

### Description
Add the following pallets to the Centrifuge runtime:

* interest-accrual
* investments
* loans
* permissions
* pool-registry
* pool-system

Configuration for all of these should be nearly-identical between Development and Altair, and those are the configuration settings we will want.

Things to keep in mind:
* `PoolCreateOrigin` should be `EnsureRoot`
* `PoolDeposit` needs a final value

mustermeiszer:
# Description
This PR adds all relvant pallets, struct logic to the the Centrifuge runtime.

Fixes #1224, #1225, #1226

## Changes and Descriptions
* Adds new weights for Centrifuge
* Adds new weights for Altair
* Fixes pallet-pool-registry benchmarks
* Adds new loans-pallet to Centrifuge

## Type of change

# How Has This Been Tested?
- [ ] Charcoal has been deployed and upgrade
 
# Tasks:
- [ ] ~~Write integration test for creating a pool~~ &rarr; not possible currently, as integration-tests, Issue see #1288
- [x] Pump runtime version number
- [ ] Charcoal has been upgraded successfully
- [x] Proxy setup has been reviewed
- [x] Loans is integrated
- [x] Deposit amount for Pools is right

# Release Process
Checklist: https://centrifuge.hackmd.io/@mq/H1tt-AT9s?type=view


mustermeiszer:
@offerijns can you take a look at the pool parameters?

runtime/centrifuge/src/lib.rs
```diff
@@ -557,6 +577,58 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
...
+					| RuntimeCall::Anchor(..)
+					| RuntimeCall::Utility(pallet_utility::Call::batch_all { .. })
+			),
+			// This type of proxy is used only for authenticating with the centrifuge POD,
+			// having it here also allows us to validate authentication with on-chain data.
+			ProxyType::PodAuth => false,
+			ProxyType::PermissionManagement => matches!(
+				c,
+				RuntimeCall::Permissions(pallet_permissions::Call::add { .. })
+					| RuntimeCall::Permissions(pallet_permissions::Call::remove { .. })
```
offerijns, line 630: Just realized, we should probably add `batch_all` and `batch` here.

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1072,151 @@ impl pallet_collator_selection::Config for Runtime {
...
+}
+
+// Pool config parameters
+parameter_types! {
+	pub const PoolPalletId: frame_support::PalletId = cfg_types::ids::POOLS_PALLET_ID;
+
+	/// The index with which this pallet is instantiated in this runtime.
+	pub PoolPalletIndex: u8 = <PoolSystem as PalletInfoAccess>::index() as u8;
+
+	pub const MinUpdateDelay: u64 = 0; // no delay
```
offerijns, line 1100: This should be 7 days

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1072,151 @@ impl pallet_collator_selection::Config for Runtime {
...
+	pub const ChallengeTime: BlockNumber = if cfg!(feature = "runtime-benchmarks") {
+		// Disable challenge time in benchmarks
+		0
+	} else {
+		2 * MINUTES
```
offerijns, line 1105: This should be 30 minutes

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1072,151 @@ impl pallet_collator_selection::Config for Runtime {
...
+	};
+
+	// Defaults for pool parameters
+	pub const DefaultMinEpochTime: u64 = 5 * SECONDS_PER_MINUTE; // 5 minutes
```
offerijns, line 1109: This should be 23 hours + 50 minutes

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1072,151 @@ impl pallet_collator_selection::Config for Runtime {
...
+	pub const DefaultMaxNAVAge: u64 = 1 * SECONDS_PER_MINUTE; // 1 minute
```
offerijns, line 1110: This should be 0 (forcing update nav + close epoch in same tx)

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1072,151 @@ impl pallet_collator_selection::Config for Runtime {
...
+
+	// Runtime-defined constraints for pool parameters
+	pub const MinEpochTimeLowerBound: u64 = 1; // at least 1 second (i.e. do not allow multiple epochs closed in 1 block)
```
offerijns, line 1113: This should be 1 hour

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1072,151 @@ impl pallet_collator_selection::Config for Runtime {
...
+parameter_types! {
+	#[cfg_attr(feature = "std", derive(frame_support::Serialize, frame_support::Deserialize))]
+	pub const MaxTranches: u32 = 5;
+
+	// How much time should lapse before a tranche investor can be removed
+	#[derive(Debug, Eq, PartialEq, scale_info::TypeInfo, Clone)]
+	pub const MinDelay: Moment = 7 * SECONDS_PER_DAY;
+
+	#[derive(Debug, Eq, PartialEq, scale_info::TypeInfo, Clone)]
+	pub const MaxRolesPerPool: u32 = 1_000;
```
offerijns, line 1174: Can we make this 10_000? 1000 investors per pool is actually a pretty low limit

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1072,151 @@ impl pallet_collator_selection::Config for Runtime {
...
+	// TODO: This is a stopgap value until we can calculate it correctly with updated benchmarks. See #1024
+	type MaxRateCount = MaxActiveLoansPerPool;
+	type RuntimeEvent = RuntimeEvent;
+	type Time = Timestamp;
+	type Weights = ();
+}
+
+parameter_types! {
+	pub const LoansPalletId: PalletId = cfg_types::ids::LOANS_PALLET_ID;
+	pub const MaxActiveLoansPerPool: u32 = 50;
```
offerijns, line 1216: This can be (much) higher right? I don't think we need to set it to 5000+ now, but maybe at least to 250-500?

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1072,151 @@ impl pallet_collator_selection::Config for Runtime {
...
+	pub const MaxWriteOffGroups: u32 = 10;
```
offerijns, line 1217: This should be higher as well, I'd say 100

runtime/centrifuge/src/lib.rs
branan: This will need to be reconfigured based on the benchmarks, but we should be able to safely set it to a couple of thousand.

runtime/centrifuge/src/lib.rs
branan: Permissions storage is all based on Substrate maps, so this is an arbitrary limit to keep storage down rather than anything required by a linear-scaling operation. We can set this to whatever makes sense.

runtime/centrifuge/src/lib.rs
branan: I think this will need to be configured based on whether we're building benchmarks

runtime/centrifuge/src/lib.rs
branan: This might also break benchmarks if set to a sane value? I am less-sure about this one

runtime/centrifuge/src/lib.rs
offerijns: This defines how long there needs to be between submitting a pool update and execution of the pool update. That's an economical/product decision.

runtime/centrifuge/src/lib.rs
offerijns: This is just to prevent issuers from setting the min epoch time to such a low number that investors are not given a fair chance to submit new orders. This should not relate to benchmarks.

branan: It won't affect the output of the benchmarks, but they need to be able to set the timestamp far enough forward so that `update` can actually succeed and be benchmarked

branan: I think the benchmarks rely on being able to set the epoch time of the pool so that they can measure closing an epoch. They'll need to account for the minimum (even if it's just increasing the value of a fake timestamp).It won't affect the actual benchmark result, of course.

mustermeiszer:
Ran the benchmarks locally: 

First failure here: 

```
Starting benchmark: pallet_pool_registry::register
Starting benchmark: pallet_pool_registry::update_no_execution
Error: Input("Benchmark pallet_pool_registry::update_no_execution failed: BalanceTooLow")
Error: Intermediate error occurred. Cleaning up artifacts in '/tmp/runtime/centrifuge'
Aborting!
```

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1080,410 @@ impl pallet_collator_selection::Config for Runtime {
...
+	// per byte deposit is 0.01 AIR
+	pub const DepositPerByte: Balance = CENTI_CFG;
+	// Base deposit to add attribute is 0.1 AIR
+	pub const AttributeDepositBase: Balance = 10 * CENTI_CFG;
+	// Base deposit to add metadata is 0.1 AIR
+	pub const MetadataDepositBase: Balance = 10 * CENTI_CFG;
+	// Deposit to create a class is 1 AIR
+	pub const CollectionDeposit: Balance = CFG;
+	// Deposit to create a class is 0.1 AIR
+	pub const ItemDeposit: Balance = 10 * CENTI_CFG;
```
branan, line 1460: @mustermeiszer @offerijns These are copied 1:1 from Altair, but these deposit values feel pretty low to me. My gut says we should set them 10x-100x, but I have absolutely nothing scientific to back that up with.

mustermeiszer: I agree. But not sure what we should base them on economically. Usually something I would ask Cassidy.

lemunozm:
I'm going to take the freedom to rebase this to the latest main in order to be able to create another PR with the loan addition.

lemunozm:
I rebase it and fix minor errors

mustermeiszer:
@lemunozm No, we have some really bad changes in this PR. Can we do a review round on the current state and then merge #1254 into this one her?

mustermeiszer:
@branan @offerijns can you please review the changes I needed to do to the 
* Runtime settings - update guard
  * I removed the check of the `MinUpdateDelay` as this is done on the system side already 
  * I added a lot of feature flag dependent configuration values. &rarr; maybe there is a cleaner way to do this??
* Pallet-pool-system 
  * I changed the update to track the time of submission, rather than the time of "estimated execution"
  * I changed how we create a pool to use the defaults rather than checking whether the defaults are valid and use the "most" valid value for nav and epoch parameters

mustermeiszer:
ping @offerijns and @branan

runtime/centrifuge/src/lib.rs
```diff
@@ -494,6 +548,14 @@ pub enum ProxyType {
 	/// Deprecated ProxyType, that we are keeping due to the migration
 	_Staking,
 	NonProxy,
+	Borrow,
+	Price,
```
offerijns, line 554: I think this can be removed as there is no `price` step for loans anymore.

offerijns:
> * Runtime settings - update guard
>   
>   * I removed the check of the `MinUpdateDelay` as this is done on the system side already
> * Pallet-pool-system
>   
>   * I changed the update to track the time of submission, rather than the time of "estimated execution"
>   * I changed how we create a pool to use the defaults rather than checking whether the defaults are valid and use the "most" valid value for nav and epoch parameters

I reviewed this, looks good to me 👍

runtime/centrifuge/src/lib.rs
mustermeiszer: Will remove Price then

runtime/centrifuge/src/lib.rs
```diff
@@ -1506,16 +1505,16 @@ impl pallet_keystore::pallet::Config for Runtime {
...
+	// per byte deposit is 0.1 CFG
+	pub const DepositPerByte: Balance = 10 * CENTI_CFG;
+	// Base deposit to add attribute is 1 CFG
+	pub const AttributeDepositBase: Balance = 1 * CFG;
+	// Base deposit to add metadata is 1 CFG
+	pub const MetadataDepositBase: Balance = 1 * CFG;
+	// Deposit to create a class is 10 CFG
+	pub const CollectionDeposit: Balance = 10 * CFG;
+	// Deposit to create a class is 1 CFG
+	pub const ItemDeposit: Balance = 1 * CFG;
```
offerijns, line 1517: This means that to create a pool with 1000 loans, the deposit is 1000 CFG (per item) + 3000 CFG (NFTs minted by the POD have 2 attributes and 1 metadata field) + metadata size based deposit. That's fine now but can get expensive later, and this disincentivizes on-chain transparency (cheaper to create 1 asset NFT with many bundled loans, than 1000 individual asset NFTs). I would rather keep all of these a bit lower, potentially just returning to what they were. I would probably increase the collection deposit in return, e.g. to 100 CFG.

runtime/centrifuge/src/lib.rs
mustermeiszer: After loans is merged.

pallets/collator-allowlist/src/lib.rs
```diff
@@ -171,6 +171,7 @@ impl<T: Config> ValidatorRegistration<T::ValidatorId> for Pallet<T> {
 	/// True iff
 	///   - the validator id is present in the allowlist and
 	///   - the validator id is registered in the underlying validator registration center
+	#[cfg(not(test))]
 	fn is_registered(id: &T::ValidatorId) -> bool {
 		let contains_key = if cfg!(feature = "runtime-benchmarks") {
```
lemunozm, line 176: I think you can also do:
```suggestion
		let contains_key = if cfg!(all(test, feature = "runtime-benchmarks")) {
```

mustermeiszer: Nah, this isn't what I want. We are building with `runtime-benchmakrs` during tests. But the tests actually need the logic to behave like it does when not using this flag.

mustermeiszer: Not saying that there might be a nicer, cleaner solution ^^

pallets/pool-system/src/pool_types.rs
```diff
@@ -95,7 +95,7 @@ where
 	MaxTranches: Get<u32>,
 {
 	pub changes: PoolChanges<Rate, MaxTokenNameLength, MaxTokenSymbolLength, MaxTranches>,
-	pub scheduled_time: Moment,
+	pub submitted_at: Moment,
```
lemunozm, line 98: Sure @onnovisser must know about this. Breaking change for the UI

runtime/centrifuge/src/lib.rs
```diff
@@ -288,7 +311,39 @@ impl pallet_restricted_tokens::Config for Runtime {
 	type PreFungiblesTransfer = cfg_traits::Always;
 	type PreReservableCurrency = cfg_traits::Always;
 	type RuntimeEvent = RuntimeEvent;
-	type WeightInfo = weights::pallet_restricted_tokens::WeightInfo<Self>;
+	type WeightInfo = ();
```
lemunozm, line 314: I'm pretty sure we don't want this. I think the calls are exposed.

Default implementations have 0 weights.

runtime/centrifuge/src/lib.rs
```diff
@@ -924,6 +1044,7 @@ impl chainbridge::Config for Runtime {
 	type ProposalLifetime = ProposalLifetime;
 	type RelayerVoteThreshold = RelayerVoteThreshold;
 	type RuntimeEvent = RuntimeEvent;
+	// TODO: Fix benchmarks
 	type WeightInfo = ();
```
lemunozm, line 1048: Highlight this

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1121,425 @@ impl pallet_collator_selection::Config for Runtime {
...
+
+	// Defaults for pool parameters
+	pub const DefaultMinEpochTime: u64 = if cfg!(feature = "runtime-benchmarks") {
+		0 // Allow short epoch time in benchmarks and multiple close in one block
+	} else {
+		23 * SECONDS_PER_HOUR + 50 * SECONDS_PER_MINUTE // 23h and 50 minutes
+	};
+
+	pub const DefaultMaxNAVAge: u64 = if cfg!(feature = "runtime-benchmarks") {
+		1 * SECONDS_PER_HOUR // 1 hour
```
lemunozm, line 1151: Why not `0` in benchmark too?

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1121,425 @@ impl pallet_collator_selection::Config for Runtime {
...
+
+	// Runtime-defined constraints for pool parameters
+	pub const MinEpochTimeLowerBound: u64 = if cfg!(feature = "runtime-benchmarks") {
+		0 // Allow short epoch time in benchmarks and multiple close in one block
+	} else {
+		1 * SECONDS_PER_HOUR // 1 hour
+	};
+	pub const MinEpochTimeUpperBound: u64 = 30 * SECONDS_PER_DAY; // 1 month
+	pub const MaxNAVAgeUpperBound: u64 = if cfg!(feature = "runtime-benchmarks") {
+		1 * SECONDS_PER_HOUR // Allow an aged NAV in benchmarks
```
lemunozm, line 1164: Why this?

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1121,425 @@ impl pallet_collator_selection::Config for Runtime {
...
+	type MaxNAVAgeUpperBound = MaxNAVAgeUpperBound;
+	type MaxSizeMetadata = MaxSizeMetadata;
+	type MaxTokenNameLength = MaxTrancheNameLengthBytes;
+	type MaxTokenSymbolLength = MaxTrancheSymbolLengthBytes;
+	type MaxTranches = MaxTranches;
+	type MinEpochTimeLowerBound = MinEpochTimeLowerBound;
+	type MinEpochTimeUpperBound = MinEpochTimeUpperBound;
+	type MinUpdateDelay = MinUpdateDelay;
+	type NAV = Loans;
+	// TODO: Add actual pallet-loans, after https://github.com/centrifuge/centrifuge-chain/pull/1180 is merged
```
lemunozm, line 1303: I think this comment is solved

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1121,425 @@ impl pallet_collator_selection::Config for Runtime {
...
+	type InterestRate = Rate;
+	// TODO: This is a stopgap value until we can calculate it correctly with updated benchmarks. See #1024
+	type MaxRateCount = MaxActiveLoansPerPool;
+	type RuntimeEvent = RuntimeEvent;
+	type Time = Timestamp;
+	type Weights = weights::pallet_interest_accrual::WeightInfo<Self>;
+}
+
+parameter_types! {
+	pub const LoansPalletId: PalletId = cfg_types::ids::LOANS_PALLET_ID;
```
lemunozm, line 1468: This parameter can be removed

runtime/centrifuge/src/lib.rs
```diff
@@ -1000,6 +1121,425 @@ impl pallet_collator_selection::Config for Runtime {
...
+	pub const MaxActiveLoansPerPool: u32 = 1000;
```
lemunozm, line 1469: Are we ok with 1000 or we need a higher value here?

lemunozm (commented):
Some comments, one of them maybe "important".

runtime/centrifuge/src/lib.rs
mustermeiszer: Done. Needed to wait for next benchmark

runtime/centrifuge/src/lib.rs
```diff
@@ -1044,7 +1052,10 @@ impl chainbridge::Config for Runtime {
 	type ProposalLifetime = ProposalLifetime;
 	type RelayerVoteThreshold = RelayerVoteThreshold;
 	type RuntimeEvent = RuntimeEvent;
-	// TODO: Fix benchmarks
+	// NOTE: No benchmarks available.
+	//       External pallet, we use a fork of.
+	//       If we extend using this bridge
+	//       we need to write appropriate benches.
```
lemunozm: Thanks for this explanations, very appreciated!

lemunozm (dismissed):
LGTM!

runtime/centrifuge/src/lib.rs
```diff
@@ -1065,6 +1614,19 @@ construct_runtime!(
...
+
+		// Synced pallets across all runtimes - Range: 180-240
+		// WHY: * integrations like fireblocks will need to know the index in the enum
+		//      * makes it easier, without parsing complete metadata
+		//      * makes it in-sync for XCM integrations -- same enum variant again
+		PoolRegistry: pallet_pool_registry::{Pallet, Call, Storage, Event<T>} = 180,
+		PoolSystem: pallet_pool_system::{Pallet, Call, Storage, Event<T>} = 181,
+		Permissions: pallet_permissions::{Pallet, Call, Storage, Event<T>} = 182,
+		Investments: pallet_investments::{Pallet, Call, Storage, Event<T>} = 183,
+		InterestAccrual: pallet_interest_accrual::{Pallet, Storage, Event<T>, Config<T>} = 184,
```
wischli, line 1626: `pallet_interest_accrual` sets the `StorageVersion` to `Release::v2` in its `GenesisBuild`. However, we don't run any migration for Centrifuge. As a result, the `StorageVersion` storage will default to `0`. This needs to be addressed. 

Additionally, `pallet_interest_accrual` still uses the old scheme of handling `StorageVersion`, but this is another story. With the "new" way, the pallet would actually be initialized with the latest version for Centrifuge runtime.

wischli (changes requested):
I have some remarks and found at least two blockers (happy to be corrected).

runtime/centrifuge/src/lib.rs
lemunozm: That was a very good catch!
By now, can it be fixed applying the migration as it is done in `altair`? 
cc @branan

wischli: All we need to do is set `pallet_interest_accrual::StorageVersion` to `ReleaseV2`. As there is no storage trie yet on the Centrifuge runtime, no other migration is required.

Theoretically, we could also handle this the next time we bump the version but let's not start such inconsistencies.

runtime/centrifuge/src/lib.rs
```diff
@@ -1665,6 +1665,7 @@ pub type Executive = frame_executive::Executive<
 		pallet_preimage::migration::v1::Migration<Runtime>,
 		pallet_democracy::migrations::v1::Migration<Runtime>,
```
mustermeiszer, line 1666: Do we still need those?

wischli: According to #1286, we don't. So let's remove them.

wischli (dismissed):
Once the branch is rebased to master, let's rerun benchmarks to include `pallet_elections_phragmen`. But even with the default weights, it should be fine. So no blocker.

runtime/centrifuge/src/lib.rs
lemunozm: Could they be included in `altair` but not in `centrifuge`? In that case, we need them here. In `altair` I checked a local migration with those, and I get an error because they were already migrated, doing the same thing for centrifuge would give the same error

lemunozm: I think this command should tell that:
```sh
RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace cargo run --release --features try-runtime try-runtime --execution Native --chain altair-local --no-spec-check-panic on-runtime-upgrade live --uri wss://fullnode.parachain.centrifuge.io:443
````

File: pallets/interest-accrual/src/migrations.rs [hunk 25]
```diff
@@ -110,6 +110,23 @@ pub mod v2 {
 	}
 }
 
+pub mod centrifuge {
+	use super::*;
+
+	pub struct SetStorageVersionToV2<T>(sp_std::marker::PhantomData<T>);
+
+	impl<T: Config> OnRuntimeUpgrade for SetStorageVersionToV2<T> {
```
branan, line 118: In theory we could change `Release::default()` to be V2 now, and the unpopulated storage key would be V2 on Centrifuge without us needing to run a migration

runtime/centrifuge/src/lib.rs
```diff
@@ -338,6 +394,9 @@ impl orml_asset_registry::Config for Runtime {
 	type Balance = Balance;
 	type CustomMetadata = CustomMetadata;
 	type RuntimeEvent = RuntimeEvent;
+	// NOTE: Need no weights as spamming is not possible as the
+	//       calls are only callable by `AuthorityOrigin`. In out
```
branan, line 398: type - should be "our case"

runtime/centrifuge/src/lib.rs
```diff
@@ -904,6 +1028,10 @@ impl pallet_bridge::Config for Runtime {
 	type NativeTokenId = NativeTokenId;
 	type NativeTokenTransferFeeKey = NativeTokenTransferFeeKey;
 	type RuntimeEvent = RuntimeEvent;
+	// NOTE: No benchmarks available.
+	//       Might need to do that if we
+	//       extend the bridge usage
+	//       For not fine.
```
branan, line 1034: typo - "for now"

runtime/centrifuge/src/lib.rs
branan: Now that benches have been run on this PR we can probably update this before merge?

runtime/centrifuge/src/lib.rs
```diff
@@ -1065,6 +1614,19 @@ construct_runtime!(
 		OrmlTokens: orml_tokens::{Pallet, Storage, Event<T>, Config<T>} = 151,
 		OrmlAssetRegistry: orml_asset_registry::{Pallet, Storage, Call, Event<T>, Config<T>} = 152,
 		OrmlXcm: orml_xcm::{Pallet, Storage, Call, Event<T>} = 153,
+
+		// Synced pallets across all runtimes - Range: 180-240
+		// WHY: * integrations like fireblocks will need to know the index in the enum
```
branan, line 1619: Not a blocker to merging this, but I wonder if it would be better for us to expose EVM precompiles for usecases that can't parse metadata. We'd then  have:

* Stable(-ish) Substrate APIs through native pallet calls, where we can do things like reorder enums, add error variants, or insert new calls
* 100% stable precompiles, where we'd introduce a new precompile version for any change that would require re-parsing metadata.

I'm very worried about the implications for our ability to develop our pallets if we need to maintain compatibility for users who can't re-parse metadata quickly after an update.

branan: XCM can invoke a precompile through the EVM pallet relatively easily. I haven't fully though through the tradeoffs. I'm just very familiar with the pain that maintaining ABI compatibility can be vs just maintaining API compatibility. Precompiles have a much larger namespace (they're just contract addresses), so we have an opportunity to let certain ABI levels "freeze" in a way we can't do with pallets.

mustermeiszer: SOunds like it is not a bad idea, but I think it will be really hard to have the correct `AccountId` mappings then. As we always derive them. Lets keep that in mind.

pallets/interest-accrual/src/migrations.rs
mustermeiszer: Removed and defaulted to V2.

runtime/centrifuge/src/lib.rs
mustermeiszer: Removed.

wischli (approved):
LGTM!

lemunozm (approved):
LGTM!

offerijns (dismissed):

Issue #1237, opened 2023-03-08
Cut `Loan -> Pools -> Investment` tight test & benchmarks dependencies


lemunozm:
### Which part of the code is the issue addressing?

* Refactoring

### Description

Although `pallet-loans` use traits to get information from `pallet-pool-system`, and `pallet-pool-system` uses traits to get information from `pallet-investments`, this is only true for the wasm code. The testing and benchmark parts import the libraries under those traits and use a bunch of methods directly from them. This adds a huge knowledge load for a library that only wants to test itself or make some minor benchmarks, because it forces you to create a whole safe initial state on these libraries.

- For testing, this can be fixed using mocks.
- For benchmarks, this can be fixed by allowing some methods with `runtime-benchmarks` from these libraries to create that "safe initial state" (that they know how it is) where any other library can perform its benchmarks. Avoiding carrying the knowledge load.

### How will this affect the code base

- Improves readability
- Any pallet that needs to do benchmarks does not need to know anything about the underlying libraries.
- A lot fewer dependencies in `Cargo.toml` for testing

### What are foreseen obstacles or hurdles to overcome?

- No migrations expected, only internal refactors, most of them under the `std` or `runtime-benchmarks` feature that comes without risk.

lemunozm:
a complete `PoolBechmarkHelper` trait utility:

_Originally posted by @mustermeiszer in https://github.com/centrifuge/centrifuge-chain/pull/1242#pullrequestreview-1334822517_

lemunozm:
Fo now, this is done for `Loans -> Pools` dependency

lemunozm:
Closed, Pools -> Investment dependencies should be done in another PR

PR #1242, opened 2023-03-09
Pallet loans refactor: essential benchmarks

lemunozm:
# Description

*Essential* benchmarks for the pallet refactor loans.
By now, the essential benchmark is the `update_portfolio_valuation` previously called `update_nav`. The rest of the extrinsics can work with a high weight by now.

Belongs to: #1249

## Changes and Descriptions

- [x] Add benchmark for `update_portfolio_valuation`.
- [x] Set weights with high values.
- [x] Add `runtime-benchmarks` trait helper for initializing pools in benchmarks. 
  - This avoids a benchmark dependencies against loans/investment pallets from loans pallets: #1237
- [x] Add `pallet-loans-ref` to `development` runtime.
- [x] Enable benchmarks from `development`.


pallets/loans-ref/src/benchmarking.rs
```diff
@@ -0,0 +1,152 @@
...
+			let item_id = (i as u16).into();
+			T::NonFungible::mint_into(&collection_id, &item_id, &borrower).unwrap();
+
+			let loan_id = create_loan::<T>(&borrower, pool_id, (collection_id, item_id));
+			borrow_loan::<T>(&borrower, pool_id, loan_id);
+		}
+	}: _(RawOrigin::Signed(borrower), pool_id)
+	verify {
+		assert!(Pallet::<T>::portfolio_valuation(pool_id).value() > Zero::zero());
+	}
```
lemunozm, line 149: I get the following results:
- `pallet_loans_ref`: `30_000_000 + 5_000_000 * n + 50_000 * m`
- `pallet_loans`: `40_000_000 + 19_000_000 * n + 100_000 * m`

This is better because now, we do not write into the `ActiveLoans` storage in this method. I would like another check. Could you please @Branan check if this code is equivalent to the previous version? 

I've avoided calling the `on_initialize` of `interest-accrual`, because it seems to have no effect on this benchmark. The rest, I think, is very, very similar, at least to not change the benchmarking results.

pallets/loans-ref/src/weights.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use frame_support::{
+	traits::Get,
+	weights::{constants::RocksDbWeight, Weight},
+};
+use sp_std::marker::PhantomData;
+
+/// Weight functions needed for pallet_loans_ref.
+pub trait WeightInfo {
+	fn update_portfolio_valuation(n: u32, m: u32) -> Weight;
```
lemunozm, line 33: Pretty sure we do not need the `m` (the MaxRateCount`) parameter because we use it later always with the max value.

pallets/loans-ref/src/lib.rs
```diff
@@ -611,7 +636,11 @@ pub mod pallet {
 				update_type: PortfolioValuationUpdateType::Exact,
 			});
 
-			Ok(())
+			Ok(Some(T::WeightInfo::update_portfolio_valuation(
+				count,
+				MaxRateCountOf::<T>::get(),
```
lemunozm, line 641: Here (from my previous comment)

lemunozm (commented):
Self comments below:

mustermeiszer:
I am not really into the benchmarks of loans. Would be great to get @branan eyes on this! 🙂

pallets/pool-system/src/impls.rs
```diff
@@ -394,3 +394,102 @@ impl<T: Config> InvestmentAccountant<T::AccountId> for Pallet<T> {
...
+			T::Permission::add(
+				PermissionScope::Pool(pool_id),
+				investor.clone(),
+				Role::PoolRole(PoolRole::TrancheInvestor(tranche, u64::MAX)),
+			)
+			.unwrap();
+
+			T::Tokens::mint_into(CurrencyId::AUSD, &investor, FUNDS.into()).unwrap();
+			T::Investments::update_investment(
+				&investor,
```
mustermeiszer, line 472: I would at least think about it a bit before merging this like this. I agree we need benchmarking utilities and those should be abstracted via traits. But this impl here makes so many assumptions. 

We should probably take a little time to think whether this is sufficient for all benchmarks that depends on pools or whether we just create more technical debt here.

mustermeiszer (dismissed):
I am approving this in order to not block @lemunozm. But I would also propose to think about the benchmarking trait a little more. Feels like technical debt rather than a good solution.

pallets/pool-system/src/impls.rs
lemunozm: If agree, I reference this conversation into #1237 to track this technical debt

mustermeiszer (approved):

pallets/loans-ref/src/benchmarking.rs
lemunozm: @branan if you find something weird I tackle it in another PR

branan: I think it looks good. As long as the loans are using the last-added interest rate (so that we get worst-case scanning of the vector) then we're all set, and it appears you've got that going in a much cleaner way than the mess that was in the old benchmarks :+1:

pallets/loans-ref/src/lib.rs
```diff
@@ -687,17 +716,26 @@ pub mod pallet {
...
+			let count = loans.len().ensure_into()?;
+			let value = loans.into_iter().try_fold(
 				T::Balance::zero(),
 				|sum, (loan, _)| -> Result<T::Balance, DispatchError> {
 					Ok(sum.ensure_add(loan.current_present_value(&rates)?)?)
 				},
-			)
+			)?;
+
+			Ok((value, count))
```
branan, line 732: We could (and maybe should?) return the count of rates here to satisfy the `m`, rather than always assuming the max. But 1000 weights is the equivalent of only about three DB reads, so it's not a _huge_ waste of block time to just assume the max

pallets/loans-ref/src/benchmarking.rs
lemunozm: Great! Thanks for your eyes on this Branan 🙌🏻

pallets/loans-ref/src/lib.rs
lemunozm: Ok, I see it like you. Knowing the values, I would prefer not to give the loans all that knowledge and assume always the maximum

pallets/loans-ref/src/weights.rs
lemunozm: Answered below by Branan

PR #1251, opened 2023-03-13
Pallet transfers allowlist

### LINKED ISSUE #1144: Pallet-Transfer-Allowlist
* [x] Feature

* Spec: https://centrifuge.hackmd.io/EL3WEpBuR7Gmzh904WYmaw

### Assumptions
* If a user has NO restriction set for a specific currency, then it MUST be assumed that every receiver is valid
* If a user has ONE restriction set for a specific currency, then ONLY allowlisted accounts are valid receivers
* The restrictions are checked on EVERY transfer via the `pallet_restricted_transfers::Pallet::transfer(..)` method
* Restrictions are set on a per currency base
   *E.g.: User A has a restriction set for CFG, which is only User B is a valid receiver for him*
* A receiver can be 
    * A local `AccountId32`
    * A XCM based `MultiLocation`
    * A connecotors based `DomainAddress`

### Description
A pallet that provides a means of allowing specific addresses to transfer to and only to transfer to those. The pallet implements a trait that others can consume.

The Config of the pallet should roughly look like the following: 
```rust
trait Config: frame_system::Config {      
    type Deposit: Get<BalanceOf<Self>>
}
```

and the pallet should define a type for a `Location` that is able to provide the following. This might just be the plain `Multilocation` from xcm, but maybe a custom struct is better suited.

```rust
impl From<AccountId> for Location {
   ...
}

impl From<Multilocation> for Location {
   ...
}
```

#### Traits
```rust
trait TransferAllowance<AccountId, Location> {
    type CurrencyId;
    
    fn allowance(send: AccountId, recv: Location, currency: SelfCurrencyId) -> DispatchResult;
}
```
Implemented by the pallet in two ways:
* Local AccountId based
    ```rust
    impl<T: Config> TransferAllowance<Self::AccountId, Self::AccountId> for Pallet<T> {
        type CurrencyId = CurrencyIdOf<T>;

        fn allowance(send: Self::AccountId, recv: Self::AccountId, currency: Self::CurrencyId) -> DispatchResult {
            if // Allowance set 
               // * If recv is part of list for currency
               //    * If yes -> has delay set
               //           * If yes -> has delay since adding passed
               //           * If not -> Fail
               //.     * If not -> Pass
               //  * If not -> pass 
            else 
                Ok(())
        }
    }
    ```

* MultiLocation based
    ```rust
    impl<T: Config> TransferAllowance<Self::AccountId, Multilocation> for Pallet<T> {
        type CurrencyId = CurrencyIdOf<T>;

        fn allowance(send: Self::AccountId, recv: Multilocation, currency: Self::CurrencyId) -> DispatchResult {
           if // Allowance set 
               // * If recv is part of list for currency
               //    * If yes -> has delay set
               //           * If yes -> has delay since adding passed
               //           * If not -> Fail
               //.     * If not -> Pass
               //  * If not -> pass 
            else 
                Ok(())
        }
    }
    ```

* DomainAddress based
    ```rust
    impl<T: Config> TransferAllowance<Self::AccountId, DomainAddress> for Pallet<T> {
        type CurrencyId = CurrencyIdOf<T>;

        fn allowance(send: Self::AccountId, recv: DomainAddress, currency: Self::CurrencyId) -> DispatchResult {
           if // Allowance set 
               // * If recv is part of list for currency
               //    * If yes -> has delay set
               //           * If yes -> has delay since adding passed
               //           * If not -> Fail
               //.     * If not -> Pass
               //  * If not -> pass 
            else 
                Ok(())
        }
    }
    ```

#### Extrinsic
* adding allowlist extrinsic, with parameters
    * `Origin::signed`
    * `who //Probably MultiLocation`
    * `CurrencyId`

    Logical requirements:
    * Can be called multiple times. As adding multiple accounts. The `Origin` is then only allowed to send to the specified addresses.
    * If there is a delay set, the changes will only be active after the delay passed. 
    * Need to lock some amount for storage reasons
    * Calling this after removing will reset `blocked_at` to `BlockNumber::MAX`

* removing allowlist extrinsics, with parameters
    * `Origin::signed`
    * `who //Probably MultiLocation`
    * `CurrencyId`
    
    Logical requirements:
    * Can be called on any existing combination of the parameters. 
    * Sets the `blocked_at` to `current_block + delay`.
    
* purge_storage extrinsics, with parameters
    * `Origin::signed`
    * `who`
    * `CurrencyId`

    Logical requirements:
    * Can be called in order to retain locked tokens.
    * Checks whether `blocked_at >= current_block && blocked_at != BlockNumber::MAX`
        * If passes: remove storage entry and free token
    
* adding a delay extrinsics, with parameters
    * `Origin::sined`
    * `delay: BlockNumber`

    Logical requirements:
    * Updates the delay to the given value
    
    Question:
    * How to handle changing a delay if there exist already one?
    
#### Storage and Types
* Allowances for a user
    ```rust
    #[pallet::storage]
    type Allowance = StorageNMap<
	    _,
	    (
		    NMapKey<Twox64Concat, T::AccountId>,
                     NMapKey<Twox64Concat, T::CurrencyId>,
		    NMapKey<Twox64Concat, Location>,
	    ),
	    AllowanceDetails,
	    OptionQuery,
    >;
    ```
* Type `AllowanceDetails`
    ```rust=
    pub struct AllowanceDetails<BlockNumber> {
        allowed_at: BlockNumber // Defaults to 0
        blocked_at: BlockNumber // Defaults to BlockNumber::MAX
    }
    ```
* Delay settings of a user
    ```rust
    #[pallet::storage]
    type Delay = StorageMap<_, T::AccountId, Twox64Concat, T::CurrencyId, Twox64Concat, T::BlockNumber>
    ```

### Research/based on
* User needs ensuring they can NOT send tokens accidentally to the wrong address
* User needs ensuring they can connect a hot wallet and a single person with access to millions of dollars

thea-leake:
# Description

This adds a pallet to handle allowlist for transfers.

It allows users to add an allowance for an account, and if allowances present, then checks against the allowance for the sending account/currrency will be limited to the recievers for which allowances are defined, and allow/blocked at block range.

Fixes #1144

### Adds Transfer Allowlist pallet

Allowlist pallet, and tests added.

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [x] Pallet unit tests with externalities provided mock env

# Checklist:

- [x ] I have added Rust doc comments to structs, enums, traits and functions
- [x ] I have made corresponding changes to the documentation
- [ x] I have performed a self-review of my code
- [ x] My changes generate no new warnings
- [ x] I have added tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes
- [ x] I rebased on the latest `main` branch


mustermeiszer (commented):
Looks already neat!

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -272,28 +272,30 @@ pub mod pallet {
...
 		) -> DispatchResult {
 			let account_id = ensure_signed(origin)?;
 
+			let allowance_details = match Self::sender_currency_delay(&account_id, currency_id) {
+				Some(delay) => AllowanceDetails {
+					blocked_at: <frame_system::Pallet<T>>::block_number().saturating_add(delay),
+					..AllowanceDetails::default()
+				},
+				_ => AllowanceDetails::default(),
+			};
```
lemunozm, line 284: Very functional pattern, I like it! ❤️

pallets/transfer-allowlist/Cargo.toml
```diff
@@ -0,0 +1,58 @@
...
+scale-info = { version = "2.3.0", default-features = false, features = ["derive"] }
+serde = { version = "1.0.119" }
+
+cfg-traits = { path = "../../libs/traits", default-features = false }
+pallet-connectors = { path = "../connectors" }
+
+frame-benchmarking = { git = "https://github.com/paritytech/substrate", default-features = false, optional = true, branch = "polkadot-v0.9.32" }
+frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.32" }
+frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.32" }
+pallet-balances = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.32" }
```
lemunozm, line 26: I think `pallet-balances` can be enabled only for benchmarking

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+use xcm::{v1::MultiLocation, VersionedMultiLocation};
+
+#[cfg(test)]
+mod mock;
+
+#[cfg(test)]
+mod tests;
+
+/// AccountId type for runtime used in pallet.
+pub type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
```
lemunozm, line 41: Instead of `AccountIdOf<T>`, can not use directly `T::AccountId`?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+/// If there are no allowances specified, then the account is assumed to be allowed
+/// to send to any location without restrictions.
+/// However once an allowance for a sender to a specific recieving location and currency is made,
+/// /then/ transfers from the sending account are restricted for that currency to:
+/// - the account(s) for which allowances have been made
+/// - the block range specified in the allowance
+use codec::{Decode, Encode, MaxEncodedLen};
+use frame_support::{dispatch::DispatchError, RuntimeDebugNoBound};
+pub use pallet::*;
+use pallet_connectors::DomainAddress;
```
lemunozm, line 25: @NunoAlexandre If this type is used here, could it be moved to `cfg-types` to avoid the connector dependency?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+#![cfg_attr(not(feature = "std"), no_std)]
+
+/// This module checks whether an account should be allowed to make a transfer to
+/// a receiving location with a specific currency.
+/// If there are no allowances specified, then the account is assumed to be allowed
+/// to send to any location without restrictions.
+/// However once an allowance for a sender to a specific recieving location and currency is made,
+/// /then/ transfers from the sending account are restricted for that currency to:
+/// - the account(s) for which allowances have been made
+/// - the block range specified in the allowance
```
lemunozm, line 21:
```suggestion
//! This module checks whether an account should be allowed to make a transfer to
//! a receiving location with a specific currency.
//! If there are no allowances specified, then the account is assumed to be allowed
//! to send to any location without restrictions.
//! However once an allowance for a sender to a specific recieving location and currency is made,
//! /then/ transfers from the sending account are restricted for that currency to:
//! - the account(s) for which allowances have been made
//! - the block range specified in the allowance

```

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+pub trait TransferAllowance<AccountId> {
+	type CurrencyId;
+	/// Determines whether the `send` account is allowed to make a transfer to the  `recieve` loocation with `currency` type currency.
+	/// Returns result wrapped bool for whether allowance is allowed.
+	fn allowance(
+		send: AccountId,
+		recieve: Location,
+		currency: Self::CurrencyId,
+	) -> Result<bool, DispatchError>;
+}
```
lemunozm, line 112: Should be moved to `cfg-traits`?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+		/// Deposit Balance to reserve/release
+		#[pallet::constant]
+		type Deposit: Get<DepositBalanceOf<Self>>;
+
+		/// Currency for Reserve/Unreserve with allowlist adding/removal,
+		/// given that the allowlist will be in storage
+		type ReserveCurrency: Currency<Self::AccountId> + ReservableCurrency<Self::AccountId>;
+	}
+
+	pub type BlockNumberOf<T> = <T as frame_system::Config>::BlockNumber;
```
lemunozm, line 156: I think `T::BlockNumber` works

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+	pub type CurrencyIdOf<T> = <T as Config>::CurrencyId;
```
lemunozm, line 157: Is this new type really needed?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+					<AccountCurrencyTransferRestriction<T>>::insert(
+						account_id,
+						currency_id,
+						new_allowance_count,
+					);
+					Ok(())
+				}
+				Some(_) => Err(DispatchError::from(
+					Error::<T>::AllowanceCountArithmeticError,
+				)),
```
lemunozm, line 497: Is this error possible in the model?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+			match <AccountCurrencyTransferRestriction<T>>::get(account_id, currency_id) {
+				Some(allowance_count) if allowance_count <= 1 => {
+					<AccountCurrencyTransferRestriction<T>>::remove(account_id, currency_id);
+					Ok(())
+				}
+				Some(allowance_count) => {
+					// check in this case should not ever be needed
+					let new_allowance_count = allowance_count
+						.checked_sub(1)
+						.ok_or(Error::<T>::AllowanceCountArithmeticError)?;
```
lemunozm, line 523: I think you can use `cfg_traits::ops::ensure_sub`. These methods will be part of Substrate from `0.9.37`, and allow you to eliminate `Overflow` errors from the pallet. 

Related info from "our" guidelines: https://centrifuge.hackmd.io/sLMxV5WoTy2q57fJTfbNLQ?view#Proposal

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+		},
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T>
+	where
+		<T as frame_system::Config>::AccountId: Into<Location>,
+	{
+		#[transactional]
+		#[pallet::call_index(0)]
```
lemunozm, line 303: Do we need this line?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+			currency_id: CurrencyIdOf<T>,
+		},
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T>
+	where
+		<T as frame_system::Config>::AccountId: Into<Location>,
+	{
+		#[transactional]
```
lemunozm, line 302: This is added automatically for every extrisic

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+
+	/// Struct to define when a transfer should be allowed from
+	/// the sender, receiver, and currency combination.
+	/// Transfer allowed time set by range of block numbers
+	/// Defaults to starting at 0, and ending at MAX block value
+	/// as per default.
+	#[derive(Clone, Debug, Encode, Decode, Eq, PartialEq, Default, MaxEncodedLen, TypeInfo)]
+	pub struct AllowanceDetails<BlockNumber> {
+		pub allowed_at: BlockNumber,
+		pub blocked_at: BlockNumber,
```
lemunozm, line 176: Could you add a minor doc for these public attributes?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+					let current_block = <frame_system::Pallet<T>>::block_number();
+					match <AccountCurrencyTransferAllowance<T>>::get((&send, &currency, receive)) {
+						Some(AllowanceDetails {
+							allowed_at,
+							blocked_at,
+						}) if current_block >= allowed_at && current_block < blocked_at => Ok(true),
+						_ => Ok(false),
+					}
+				}
+				_ => Ok(true),
```
lemunozm, line 558: is `count == 0` an error in the model, something that can not happen? If so, we are returning `Ok(true)` in that case. Should be an error instead?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+		pub fn purge_transfer_allowance(
+			origin: OriginFor<T>,
+			currency_id: CurrencyIdOf<T>,
+			receiver: Location,
+		) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+			match <AccountCurrencyTransferAllowance<T>>::get((&account_id, &currency_id, &receiver))
+			{
+				Some(_) => {
+					T::ReserveCurrency::unreserve(&account_id, T::Deposit::get());
```
lemunozm, line 405: @mustermeiszer @mikiquantum should we use `pallet-fees` here?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+		pub fn add_transfer_allowance(
+			origin: OriginFor<T>,
+			currency_id: CurrencyIdOf<T>,
+			receiver: Location,
+		) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+
+			let allowance_details = match Self::sender_currency_delay(&account_id, currency_id) {
+				Some(delay) => AllowanceDetails {
+					allowed_at: <frame_system::Pallet<T>>::block_number().saturating_add(delay),
```
lemunozm, line 322: I would increase with `check_add` or `ensure_add` here. If we have an error obtaining a huge `delay` it silently works when it should not

pallets/transfer-allowlist/src/tests.rs
```diff
@@ -0,0 +1,577 @@
+use frame_support::{assert_noop, assert_ok};
+use frame_system::ensure_signed;
+use hex::FromHex;
+use pallet_connectors::DomainAddress;
+use xcm::v1::MultiLocation;
+
+use super::*;
+use crate::mock::*;
+
+#[test]
```
lemunozm, line 10: Clean tests and high coverage! 🥳

lemunozm (commented):
Very neat @thea-leake and very good coverage!

I left a lot of comments, but most of them are minor suggestions & questions. Good job!

pallets/transfer-allowlist/src/lib.rs
NunoAlexandre: @lemunozm yeah I think that's a good idea. Probably worth doing that in this PR, wdyt?

lemunozm: I agree!

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 I was using the config restriction on these types as opposed to storage items, but have moved it to the storage & storage struct runtime generics, so it's now using T::<type>.

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 updated to moduledoc

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 updated to use T::Blocknumber -- same as AccountId comment :)

pallets/transfer-allowlist/src/lib.rs
thea-leake: Updated to use T::CurrencyId -- same as AccountId comment updated

pallets/transfer-allowlist/src/lib.rs
thea-leake: The error should not be possible, just wanted to make sure all matches were accounted for -- updated error to indicate this.

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 Updated to use ensure_sub

pallets/transfer-allowlist/src/lib.rs
thea-leake: Ah, yeah given that we've gotten the conversions taking place before this with from/into, this would not be necessary.  Will remove.

pallets/transfer-allowlist/src/lib.rs
thea-leake: Ah, yeah I'll add a docstring & comments with the expected behaviour for the implementation to provide context.  It's currently working as intended--if there are 0 allowances added for the sending account and the currency then there shouldn't be any restrictions.

pallets/transfer-allowlist/src/lib.rs
NunoAlexandre: Thanks Thea :)

pallets/transfer-allowlist/src/lib.rs
thea-leake: My thinking was that if the delay + the current bock is greater than the max number of blocks then it would make sense to just cap at the max # of blocks.  But returning an error would also make sense, I'm good with changing if we want that changed.

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 updated docs for impl

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 added

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 removed deprecated transactional

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 removed

pallets/transfer-allowlist/src/lib.rs
thea-leake: No problem :)

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 Moved to cfg-traits

pallets/transfer-allowlist/src/lib.rs
mikiquantum: Pallet fees do not support reserve/unreserve, afaik, only one way transfer/burn/etc

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+	/// Using an StorageNMap would not allow us to look up whether there was a restriction for the sending account and currency, given that:
+	/// - we're checking whether there's an allowance specified for the receiver location
+	///   - we would only find whether a restriction was set for the account in this caseif:
+	///     - an allowance was specified for the receiving location, which would render blocked restrictions useless
+	/// - we would otherwise need to store a vec of locations, which is problematic given that there isn't a set limit on receivers
+	/// If a transfer restriction is in place, then a second lookup is done on
+	/// AccountCurrencyAllowances to see if there is an allowance for the reciever
+	/// This allows us to keep storage map vals to known/bounded sizes.
+	#[pallet::storage]
+	#[pallet::getter(fn sender_currency_restriction_set)]
```
mikiquantum, line 130: Why are the getter names different from the Storage name?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+		T::AccountId,
+		Twox64Concat,
+		T::CurrencyId,
+		u64,
+		OptionQuery,
+	>;
+
+	/// Storage item for allowances specified for a sending account, currency type and recieving location
+	#[pallet::storage]
+	#[pallet::getter(fn sender_currency_reciever_allowance)]
```
mikiquantum, line 143:
```suggestion
	#[pallet::getter(fn sender_currency_receiver_allowance)]
```

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+		_,
+		Twox64Concat,
+		T::AccountId,
+		Twox64Concat,
+		T::CurrencyId,
+		u64,
+		OptionQuery,
+	>;
+
+	/// Storage item for allowances specified for a sending account, currency type and recieving location
```
mikiquantum, line 141:
```suggestion
	/// Storage item for allowances specified for a sending account, currency type and receiving location
```

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+						.checked_add(1)
+						.ok_or(Error::<T>::AllowanceCountOverflow)?;
+					<AccountCurrencyTransferRestriction<T>>::insert(
+						account_id,
+						currency_id,
+						new_allowance_count,
+					);
+					Ok(())
+				}
+				Some(_) => Err(DispatchError::Other("Impossible allowance count")),
```
mikiquantum, line 411: In which scenario would this branch be possible?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+		type Location = Location;
+
+		/// This checks to see if a transfer from an account and currency should be allowed to a given location.
+		/// If there are no allowances defined for the sending account and currency, then the transfer is allowed.
+		/// If there is an allowance for the sending account and currency,
+		/// but the destination does not have an allowance added then the transfer is not allowed.
+		/// If there is an allowance for the sending account and currency,
+		/// and there's an allowance present:
+		/// then we check whether the current block is between the `allowed_at` and `blocked_at` blocks in the allowance.
+		fn allowance(
```
mikiquantum, line 459: Do you have an example of how this will be used by a consumer pallet in a runtime?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+	/// - we're checking whether there's an allowance specified for the receiver location
+	///   - we would only find whether a restriction was set for the account in this caseif:
+	///     - an allowance was specified for the receiving location, which would render blocked restrictions useless
+	/// - we would otherwise need to store a vec of locations, which is problematic given that there isn't a set limit on receivers
+	/// If a transfer restriction is in place, then a second lookup is done on
+	/// AccountCurrencyAllowances to see if there is an allowance for the reciever
+	/// This allows us to keep storage map vals to known/bounded sizes.
+	#[pallet::storage]
+	#[pallet::getter(fn sender_currency_restriction_set)]
+	pub type AccountCurrencyTransferRestriction<T: Config> = StorageDoubleMap<
```
mikiquantum, line 131: Wouldnt be possible to merge this storage with `AccountCurrencyDelay` and have a tuple value (u64, BlockNumber). There should be an storage reduction based on the storage prefix+key.

mikiquantum (commented):
Just trying to be curious. 😄  From my limited understanding, it looks good.

pallets/transfer-allowlist/src/lib.rs
lemunozm: But should `T::Deposit::get()` come from a key? At least we do so for `Anchors::pre_commit()`.

Maybe `reserve/unreserve` actions need to be added to `pallet-fees`. 🤔

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+//! This module checks whether an account should be allowed to make a transfer to
+//! a receiving location with a specific currency.
+//! If there are no allowances specified, then the account is assumed to be allowed
+//! to send to any location without restrictions.
+//! However once an allowance for a sender to a specific recieving location and currency is made,
+//! /then/ transfers from the sending account are restricted for that currency to:
+//! - the account(s) for which allowances have been made
+//! - the block range specified in the allowance
+use cfg_traits::TransferAllowance;
+use cfg_types::locations::Location;
```
lemunozm, line 23: WDYT about passing this type as a `Config` associated type to the pallet?
Not mandatory at all, only want to know your thoughts

lemunozm (commented):
I had a very clean review feedback, thanks for all your detailed answer to my question/suggestions @thea-leake

pallets/transfer-allowlist/src/lib.rs
lemunozm: Ok, I see what you mean. Then it's ok for me as it is.

pallets/transfer-allowlist/src/lib.rs
lemunozm: I just had a comment in the same line. This case doesn't belong to the model and can never happen.

pallets/transfer-allowlist/src/lib.rs
lemunozm: There are some similar codes using `checked_add`,  maybe it makes sense to use `ensure_add` too there

libs/types/src/domain_address.rs
```diff
@@ -0,0 +1,81 @@
...
+	pub domain: Domain,
+}
+
+impl<Domain> TypeId for DomainLocator<Domain> {
+	const TYPE_ID: [u8; 4] = *b"domn";
+}
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub enum DomainAddress {
```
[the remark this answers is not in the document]
NunoAlexandre: @mikiquantum no. `Domain` is used by connectors as the destination of a message to be sent. `DomainAddress` is a necessary type to ensure type-safety when specifying an address within a given domain (i.e, to avoid being possible to pass 32-byte addresses for EVM-based domains.

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+	}
+
+	pub type DepositBalanceOf<T> = <<T as Config>::ReserveCurrency as Currency<
+		<T as frame_system::Config>::AccountId,
+	>>::Balance;
+
+	//
+	// Storage
+	//
+	pub type AllowanceDetailsOf<T> = AllowanceDetails<<T as frame_system::Config>::BlockNumber>;
```
NunoAlexandre, line 82: FYI We usually place these type aliases at the top, before `pub trait Config: frame_system::Config {`

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+		pub allowed_at: BlockNumber,
+		/// Specifies a block number after-which transfers will be blocked
+		/// for the sender & currency and destination location.
+		/// This is by default set to `BlockNumber::Max()`, except when an allowance has been removed but not purged.
+		/// In that case it is set to the current block + delay.
+		/// if the allowance is later updated with the add allowance extrinsic, it is set back to max.
+		pub blocked_at: BlockNumber,
+	}
+
+	impl<BlockNumber> AllowanceDetails<BlockNumber>
```
NunoAlexandre, line 106: We should make this actually implement the `Default` trait

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+	// Pallet Errors and Events
+	//
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Number of allowances for a sending Account/Currency set at max allowance count storage type val (currently u64::MAX)
+		AllowanceCountOverflow,
+		/// An operation expecting one or more allowances for a sending Account/Currency set, where none present
+		NoAllowancesSet,
+		/// Attempted to create allowance for existing Sending Account, Currency, and Receiver combination
+		ConflictingAllowanceSet,
```
NunoAlexandre, line 177: From the /// comment, I guess this means that this allowance already exists? If so, let's make it consistent with similar error names:

```suggestion
		DuplicateAllowance,
```

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+		/// CatchAll for Allowance Count arithmetic errors -- largely for coverage for errors that should be impossible
+		AllowanceCountArithmeticError,
+		/// No matching allowance for Location/Currency
+		NoMatchingAllowance,
+		/// No matching delay
```
NunoAlexandre, line 182: Can you expand on this error doc?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+			origin: OriginFor<T>,
+			currency_id: T::CurrencyId,
+			receiver: Location,
+		) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+
+			let blocked_at = match Self::sender_currency_delay(&account_id, currency_id) {
+				Some(delay) => <frame_system::Pallet<T>>::block_number().saturating_add(delay),
+				_ => <frame_system::Pallet<T>>::block_number(),
+			};
```
NunoAlexandre, line 288: I think this would be a bit easier to read with a `.map().unwrap_or()` pattern. Up to you 👍

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+					Ok(())
+				}
+				None => Err(DispatchError::from(Error::<T>::NoMatchingAllowance)),
+			}
+		}
+
+		#[pallet::call_index(2)]
+		#[pallet::weight(10_000 + T::DbWeight::get().reads_writes(1, 2).ref_time())]
+		/// Removes a transfer allowance for a sending account/currency and receiving location
+		/// Decrements or removes the sending account/currency count.
```
NunoAlexandre, line 316: `nit` but I prefer having the doc comments above. I think that's what we do everywhere else

libs/types/src/locations.rs
```diff
@@ -0,0 +1,80 @@
...
+	/// Test
+	TestLocal(u64),
+	/// XCM MultiLocation sending destinations.
+	/// Using hash value here as Multilocation is large -- v1 is 512 bytes, but next largest is only 40 bytes
+	/// other values aren't hashed as we have blake2 hashing on storage map keys, and we don't want the extra overhead
+	XCM(H256),
+	/// DomainAddress sending location from connectors
+	Address(DomainAddress),
+	/// Etherium address, for cases where we would have a standalone Eth address
+	Eth(H160),
```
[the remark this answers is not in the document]
thea-leake: We can, it just seemed prudent to have an entry for standalone addresses in the case they're ever used.  I can remove though, it's likely that it wont be used with both domain address and multilocations.

libs/types/src/locations.rs
```diff
@@ -0,0 +1,80 @@
...
+	fn from(ml: MultiLocation) -> Self {
+		// using hash here as mulitlocation is significantly larger than any other enum type here
+		// -- 592 bytes, vs 40 bytes for domain address (next largest)
+		Self::XCM(BlakeTwo256::hash(&ml.encode()))
+	}
+}
+
+impl From<VersionedMultiLocation> for Location {
+	fn from(vml: VersionedMultiLocation) -> Self {
+		// using hash here as mulitlocation is significantly larger than any other enum type here
```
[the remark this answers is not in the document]
thea-leake: 👍 fixed

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+	/// Storage item for Allowance delays for a sending account/currency
+	#[pallet::storage]
+	#[pallet::getter(fn sender_currency_delay)]
+	pub type AccountCurrencyDelay<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		T::AccountId,
+		Twox64Concat,
+		T::CurrencyId,
+		T::BlockNumber,
```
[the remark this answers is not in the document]
thea-leake: Ah, good to know, thanks for the heads up :)

pallets/transfer-allowlist/src/lib.rs
thea-leake: The name seemed a bit more intuative for what was being fetched, but I've adjusted the names of the storage objects themselves, and updated the getters to match.

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 fixed

pallets/transfer-allowlist/src/lib.rs
thea-leake: I've combined the storage items for delay and count as per @mikiquantum's suggestion, which has removed that case as a possibility with the types so that branch is no longer there. :)
(was not possible to reach before, however types would have accommodated that branch)

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 Merged storage items

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 fixed

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 fixed

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 expanded

pallets/transfer-allowlist/src/lib.rs
thea-leake: With the merging of the delay and count storage items, there's another option to check/destructure, so I've kept the match, but I agree with how it was before `.map().unwrap_or()` would be clearer. :)

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 fixed

pallets/transfer-allowlist/src/lib.rs
thea-leake: I think that's a good idea, looking into it rn.

thea-leake: 👍 updated to use assoc type in pallet `Config` for `Location`.

pallets/transfer-allowlist/src/lib.rs
thea-leake: Iirc we were planning on using a `PreConditions` impl that would call the allowance trait fn on the pallet to filter out whether the request should be allowed or not.

pallets/transfer-allowlist/src/lib.rs
mikiquantum: Ah yes @lemunozm I would definitely define a key in the pallet fees and read the amount from it, that even if for now we do not support the reserve/unreserve logic yet there. That way the UIs can know up front operational costs.

pallets/transfer-allowlist/src/lib.rs
lemunozm: Cool!

thea-leake: thanks :), have the doc added as well 👍

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+		/// Currency for Reserve/Unreserve with allowlist adding/removal,
+		/// given that the allowlist will be in storage
+		type ReserveCurrency: Currency<Self::AccountId> + ReservableCurrency<Self::AccountId>;
+
+		type Fees: Fees<
+			AccountId = <Self as frame_system::Config>::AccountId,
+			Balance = DepositBalanceOf<Self>,
+		>;
+
+		type AllowanceFeeKey: Get<<Self::Fees as Fees>::FeeKey>;
```
thea-leake, line 90: Currently switching to use fees & new FeeKey to handle reserve amount.

thea-leake: switch done

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍  Pallet is switched over to use fee key & getting fee val with fees pallet.  Also have mock set up to set key val for allowance creation with fees pallet to ensure fees pulled w/ correct key -- `FeeKey` used in pallet itself is still loosely coupled in allowance pallet.

pallets/transfer-allowlist/Cargo.toml
```diff
@@ -31,7 +31,10 @@ sp-std = { git = "https://github.com/paritytech/substrate", default-features = f
 xcm = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.32" }
 
 [dev-dependencies]
+pallet-authorship = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.32" }
 pallet-balances = { git = "https://github.com/paritytech/substrate", default-features = true, branch = "polkadot-v0.9.32" }
+pallet-fees = { path = "../fees" }
```
lemunozm, line 36: You do not need to add `pallet-fees`/`pallet-treasury`/`pallet-authorship` to test this. You can do it much more manageably with the mock in `libs/mocks/src/fees.rs`

pallets/transfer-allowlist/src/lib.rs
lemunozm: Cool! I think if the future, we should be able to offer some kind the reserve/unreserve method to avoid the boilerplate on this

pallets/transfer-allowlist/src/lib.rs
lemunozm: I left a comment regarding the mock part @thea-leake.

Regarding `reserve()`/`unreserve()`, @mikiquantum I've created an issue to add them to `pallet-fees`: #1282

pallets/transfer-allowlist/Cargo.toml
thea-leake: Ah ok, thanks! Yeah that would be a lot more convenient :)  I wanted to verify the reserve/unreserve amounts, and verify the pallet was using the FeeKey correctly and ensure that the reserve/unreserve logic was all being handled correctly in here.

libs/types/src/locations.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use crate::domain_address::DomainAddress;
+
+/// Location types for destinations that can receive restricted transfers
+#[derive(Clone, RuntimeDebugNoBound, Encode, Decode, Eq, PartialEq, MaxEncodedLen, TypeInfo)]
+pub enum Location {
+	/// Local chain account sending destination.
+	Local(AccountId32),
+	/// Test
+	TestLocal(u64),
```
mustermeiszer, line 31: Can we hide this behind a feature flag? and maybe have a fixed codec index of 255 in order to net mess up tests that rely on decoding?

libs/types/src/locations.rs
```diff
@@ -0,0 +1,80 @@
...
+	/// XCM MultiLocation sending destinations.
+	/// Using hash value here as Multilocation is large -- v1 is 512 bytes, but next largest is only 40 bytes
+	/// other values aren't hashed as we have blake2 hashing on storage map keys, and we don't want the extra overhead
+	XCM(H256),
+	/// DomainAddress sending location from connectors
+	Address(DomainAddress),
+	/// Ethereum address, for cases where we would have a standalone Eth address
+	Eth(H160),
```
mustermeiszer, line 39: I would also remove it as we won't have native H160 support. At least not that I can think of.

libs/types/src/locations.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(Clone, RuntimeDebugNoBound, Encode, Decode, Eq, PartialEq, MaxEncodedLen, TypeInfo)]
+pub enum Location {
+	/// Local chain account sending destination.
+	Local(AccountId32),
+	/// Test
+	TestLocal(u64),
+	/// XCM MultiLocation sending destinations.
+	/// Using hash value here as Multilocation is large -- v1 is 512 bytes, but next largest is only 40 bytes
+	/// other values aren't hashed as we have blake2 hashing on storage map keys, and we don't want the extra overhead
+	XCM(H256),
```
mustermeiszer, line 35: But this means we are breaking allowance everytime that XCM is upgraded, right? 

i.e. 
* `hash(v2::Multilocation) != hash(v3::Multilocation)`

Although, I like the optimization.

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+			+ PartialEq
+			+ TypeInfo
+			+ Encode
+			+ EncodeLike
+			+ Decode
+			+ MaxEncodedLen;
+	}
+
+	//
+	// Storage
```
mustermeiszer, line 105: Nice ❤️

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Number of allowances for a sending Account/Currency set at max allowance count storage type val (currently u64::MAX)
+		AllowanceCountOverflow,
+		/// An operation expecting one or more allowances for a sending Account/Currency set, where none present
+		NoAllowancesSet,
+		/// Attempted to create allowance for existing Sending Account, Currency, and Receiver combination
+		DuplicateAllowance,
+		/// CatchAll for Allowance Count arithmetic errors -- largely for coverage for errors that should be impossible
+		AllowanceCountArithmeticError,
```
mustermeiszer: Can we use `enum ArithmeticError` here instead. It impls `Into<DIspatchError>`.

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+		pub fn increment_or_create_allowance_count(
+			account_id: &T::AccountId,
+			currency_id: &T::CurrencyId,
+		) -> DispatchResult {
+			// not using try_mutate here as we're not sure if key exits, and we're already doing a some value check on result of exists query check
+			match Self::get_account_currency_restriction_count_delay(account_id, currency_id) {
+				Some((allowance_count, delay)) => {
+					let new_allowance_count = allowance_count
+						.checked_add(1)
+						.ok_or(Error::<T>::AllowanceCountOverflow)?;
```
mustermeiszer: I think we can use `ensure_add` here

mustermeiszer: as It is a `u64`

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 updated the checked_add to use ensure_add as well :)

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+			receiver: T::Location,
+		) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+
+			let blocked_at = match Self::get_account_currency_restriction_count_delay(
+				&account_id,
+				currency_id,
+			) {
+				Some((_, Some(delay))) => {
+					<frame_system::Pallet<T>>::block_number().saturating_add(delay)
```
mustermeiszer, line 306: I was thinking if it might makes to have two values here. 
* One for delay of adding allowance
* One for delay of removing allowance

cc @offerijns do you think this will be an issue for issuers?

mustermeiszer: This would introduce quite some rewrite I guess, so I only want it if we think it makes sense. Thinking from a security perspective here. But actually, if a key is compromised and the holder did not set a delay upfront, then the attacker can do anything directly anyways. If a delay was set, then fast removal would not introduce any advantage. WDYT?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+				}
+				_ => <frame_system::Pallet<T>>::block_number(),
+			};
+			match <AccountCurrencyTransferAllowance<T>>::get((&account_id, &currency_id, &receiver))
+			{
+				Some(existing_allowance) => {
+					let allowance_details = AllowanceDetails {
+						blocked_at,
+						..existing_allowance
+					};
```
mustermeiszer, line 316: this works? it is just taking the remaing fields from the old allowance? Nice!

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+					Self::decrement_or_remove_allowance_count(&account_id, &currency_id)?;
+					Self::deposit_event(Event::TransferAllowancePurged {
+						sender_account_id: account_id,
+						currency_id,
+						receiver,
+					});
+					Ok(())
+				}
+				None => Err(DispatchError::from(Error::<T>::NoMatchingAllowance)),
+			}
```
mustermeiszer, line 365: This should ONLY be possible of the `blocked_at` has already passed!

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+			delay: T::BlockNumber,
+		) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+			let (count, _) =
+				Self::get_account_currency_restriction_count_delay(&account_id, &currency_id)
+					.unwrap_or((0, Some(0u32.into())));
+			<AccountCurrencyTransferCountDelay<T>>::insert(
+				&account_id,
+				&currency_id,
+				(count, Some(delay)),
```
mustermeiszer, line 384:
```suggestion
			let (count, delay) =
				Self::get_account_currency_restriction_count_delay(&account_id, &currency_id)
					.unwrap_or((0, Some(delay)));
			<AccountCurrencyTransferCountDelay<T>>::insert(
				&account_id,
				&currency_id,
				(count, Some(delay)),
```

mustermeiszer: Well, not sure if actually advantage ^^

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+				}
+				None => Err(DispatchError::from(Error::<T>::NoMatchingAllowance)),
+			}
+		}
+
+		#[pallet::call_index(3)]
+		#[pallet::weight(10_000 + T::DbWeight::get().reads_writes(0, 1).ref_time())]
+		/// Adds an account/currency delay
+		/// Calling on an existing combination will update the existing delay value
+		pub fn add_or_update_allowance_delay(
```
mustermeiszer, line 372: The problem here is that

* attacker can set delay to zero
* add arbitrary off-board address
* send tokens

all in one block and batch.

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+				currency_id,
+				delay,
+			});
+			Ok(())
+		}
+
+		#[pallet::call_index(4)]
+		#[pallet::weight(10_000 + T::DbWeight::get().reads_writes(0, 1).ref_time())]
+		/// Removes an existing sending account/currency delay
+		pub fn remove_allowance_delay(
```
mustermeiszer, line 397: Same goes as above.

mustermeiszer (changes requested):
This looks really, really good, clean and nice!

But if I understood the code correctly we have 3 logical problems that must be addressed (if I got it right)

* `add_or_update_allowance_delay` MUST only be active after the previously set `delay` has passed
* `remove_allowance_delay` MUST only be active  after the previously set `delay` has passed
*  `purge_transfer_allowance` MUST only be possible of the blocked_at has already passed!

libs/types/src/locations.rs
thea-leake: 👍 removed

libs/types/src/locations.rs
thea-leake: 👍 I've added a test with v1 & v2 to make sure that multilocations of the same type and value have the same hashed value.  If there's a different type/value/enum it seems that unless we were extracting the inner value we would have a different value regardless?

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 updated to use `ensure_add`

pallets/transfer-allowlist/src/lib.rs
thea-leake: 👍 updated to ensure `blocked_at` has already passed.

pallets/transfer-allowlist/src/lib.rs
thea-leake: I'm not sure it would make sense strictly from a security perspective, however I could see it from a permissions and  convienience with fee reclaiming perspective.  I'm open to however we want to handle this :)

thea-leake: Also, would we want to add another reserve for storage usage for the delay?

mustermeiszer: Mmh. Good point. I think it makes sense. But I am not sure, if it is not a little bit hard for users to follow how they can free it it up if we have lots of little locations. 

But I am nto against having an additional fee.

thea-leake: Missed seeing this earlier unfortunately.  I would prefer keeping no fee for simplicity, but I'm wondering now if having users be able to add a delay without a fee could be a DOS risk, or would the transaction/weight fee be enough of a deterrent for that?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+					..
+				}) => Err(DispatchError::from(Error::<T>::NoMatchingDelay)),
+				Some(AllowanceMetadata {
+					modifiable_at: None,
+					..
+				}) => Err(DispatchError::from(Error::<T>::DelayUnmodifiable)),
+				Some(AllowanceMetadata {
+					modifiable_at: Some(modifiable_at),
+					..
+				}) if modifiable_at > current_block => Err(DispatchError::from(Error::<T>::DelayUnmodifiable)),
```
lemunozm, line 482: Love this pattern! 🚀

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+						currency_id,
+						delay,
+					});
+					Ok(())
+				}
+			}
+		}
+
+		#[pallet::call_index(5)]
+		#[pallet::weight(10_000 + T::DbWeight::get().reads_writes(0, 1).ref_time())]
```
lemunozm, line 506: Is it intended to add benchmarks to these methods?

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+				sender_account_id: account_id,
+				currency_id,
+				delay,
+			});
+			Ok(())
+		}
+
+		#[pallet::call_index(4)]
+		#[pallet::weight(10_000 + T::DbWeight::get().reads_writes(0, 1).ref_time())]
+		/// Updates an allowance delay, only callable is the delay has been set to allow future modifications and
```
mustermeiszer, line 460:
```suggestion
		/// Updates an allowance delay, only callable if the delay has been set to allow future modifications and
```

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+		pub fn purge_transfer_allowance(
+			origin: OriginFor<T>,
+			currency_id: T::CurrencyId,
+			receiver: T::Location,
+		) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+			let current_block = <frame_system::Pallet<T>>::block_number();
+			match <AccountCurrencyTransferAllowance<T>>::get((&account_id, &currency_id, &receiver))
+			{
+				Some(AllowanceDetails { blocked_at, .. }) if blocked_at < current_block => {
```
mustermeiszer, line 391: If clause must check the other way around. 

`if blocked_at >= current_block`

pallets/transfer-allowlist/src/lib.rs
mustermeiszer, line 482: Must be 

`if current_block < modifiable_at`

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+					);
+					Self::deposit_event(Event::TransferAllowanceDelayFutureModifiable {
+						sender_account_id: account_id,
+						currency_id,
+						modifiable_at,
+					});
+					Ok(())
+				}
+			}
+		}
```
mustermeiszer, line 547: Change as talked about (see slack for requirements)

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+		) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+
+			let current_block = <frame_system::Pallet<T>>::block_number();
+			match Self::get_account_currency_restriction_count_delay(&account_id, &currency_id) {
+				Some(AllowanceMetadata {
+					allowance_count: 0,
+					modifiable_at: Some(modifiable_at),
+					..
+				}) if modifiable_at <= current_block => {
```
mustermeiszer, line 564: Must be 

`if current_block < modifiable_at`

mustermeiszer (changes requested):
If clauses are checking the wrong way for times. Otherwise changes look good!

pallets/transfer-allowlist/src/lib.rs
mustermeiszer: This is wrong from my side. The `blocked_at` needs to have passed in order to purge it!

pallets/transfer-allowlist/src/mock.rs
```diff
@@ -0,0 +1,222 @@
...
+// actual reserve/unreserve handled by reserve currency type
+impl pallet_fees::Config for Runtime {
+	type Currency = Balances;
+	type DefaultFeeValue = DefaultFeeValue;
+	type FeeChangeOrigin = EitherOfDiverse<EnsureRoot<Self::AccountId>, EnsureSignedBy<Admin, u64>>;
+	type FeeKey = FeeKey;
+	type RuntimeEvent = RuntimeEvent;
+	type Treasury = Treasury;
+	type WeightInfo = ();
+}
```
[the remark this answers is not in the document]
thea-leake: Yeah having the mocks would definitely be cleaner, I think it would be best to switch to the mocks after benchmarking is done so that the weight/time from those dependencies are taken into account (though in this case it shouldn't really affect the benchmarks much if at all), then switch to use the mocks so the tests aren't coupled with the other pallet dependencies/changes.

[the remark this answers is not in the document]
lemunozm: >I think it would be best to switch to the mocks after benchmarking is done so that the weight/time from those dependencies are taken into account

Mocks will run only under `test`. The real benchmarks that generates the weights will run without `test` and will use the correct implementations coming from the runtime. So there is no issue with benchmark times using trait mocks.

Anyway, no hurry!

pallets/transfer-allowlist/src/lib.rs
thea-leake: Is this for the weight or the call index line?  The weight line is just an approximation rn, but we have an issue open for the restricted transfers for benchmarks.  The call_index is to make sure that the extrinsic call index never changes for the rpc calls (extra explanation just because I'm not sure which line was being referred to :) ).

lemunozm: Yeah, I was referring to the `pallet::weight` line. Ok, thanks for explaining!

mustermeiszer (dismissed):
That looks really nice! Thanks for closing so quickly. I would say lets merge it.

mustermeiszer (approved):
Thanks for all the patience and fast changes @thea-leake! Lets merge that after #1241 is in, so that @NunoAlexandre does not kill us ^^

PR #1252, opened 2023-03-14
Feature/evm

branan:
# Description

EVM-compatible API/RPC enablement

Currently the ethereum chain ID is hardcoded to "999999", if anyone _wants_ to try to poke it with web3.js

Todo for safe deployment publicly:
* Finalize account conversion
* Think about weights in more than a copy-and-paste way
* Runtime call filter to disallow EVM pallet calls we don't want
* EVM transaction filtering to disallow ETH calls we don't want
* Fixing hardcoded values in service setup

## Changes and Descriptions

**TODO**

## Type of change

# How Has This Been Tested?

Manual validation of RPC calls:

* `ethereum` pallet storage contains block info
* `evm` pallet extrinsics to deploy and run a smart contract
* `eth` Substrate RPCs can query block data
* actual ethereum RPCs through web3.js (was holding off here until I had the node-side bits wired together)
* 
# Checklist:


runtime/development/src/lib.rs
```diff
@@ -1696,6 +1704,81 @@ impl pallet_liquidity_rewards::Config for Runtime {
 	type WeightInfo = ();
 }
 
+const WEIGHT_PER_GAS: u64 = 20_000;
```
branan, line 1707: This is copied directly from the Frontier template project - TODO is to determine if this is a good value (large enough to be safe without wasting block time)

runtime/development/src/lib.rs
```diff
@@ -1696,6 +1704,81 @@ impl pallet_liquidity_rewards::Config for Runtime {
...
+		if let Some(author_index) = F::find_author(digests) {
+			let authority_id = Aura::authorities()[author_index as usize].clone();
+			return Some(H160::from_slice(&authority_id.to_raw_vec()[4..24]));
+		}
+		None
+	}
+}
+
+impl pallet_evm::Config for Runtime {
+	type AddressMapping = HashedAddressMapping<BlakeTwo256>;
```
branan, line 1729: TODO - are we happy with this address mapping? It is one-way and not transparent

runtime/development/src/lib.rs
```diff
@@ -1696,6 +1704,81 @@ impl pallet_liquidity_rewards::Config for Runtime {
...
+	type BlockGasLimit = BlockGasLimit;
+	type BlockHashMapping = pallet_evm::SubstrateBlockHashMapping<Self>;
```
branan, line 1731: TODO: pallet_ethereum::EthereumBlockHashMapping<Self>;

runtime/development/src/lib.rs
```diff
@@ -1696,6 +1704,81 @@ impl pallet_liquidity_rewards::Config for Runtime {
...
+	fn ideal() -> Permill {
+		Permill::from_parts(500_000)
+	}
+
+	fn upper() -> Permill {
+		Permill::from_parts(1_000_000)
+	}
+}
+
+impl pallet_base_fee::Config for Runtime {
```
branan, line 1770: All again copied from the template project

runtime/development/src/lib.rs
NunoAlexandre: It may be a good idea to how Moonbeam does this

src/service.rs
```diff
@@ -202,13 +237,24 @@ where
 		client.clone(),
 	);
 
+	// TODO: is it bad if we create these when we're not in a runtime that uses EVM?
```
branan, line 240: As far as I can tell this doesn't cause any issues when it's instantiated on a node without frontier set up in other ways.

It might create a frontier DB file? I need to confirm that before we merge

src/service.rs
```diff
@@ -226,15 +278,18 @@ where
 /// Start a node with the given parachain `Configuration` and relay chain `Configuration`.
 ///
 /// This is the actual implementation that is abstract over the executor and the runtime api.
+#[allow(clippy::too_many_arguments)]
```
branan, line 281: `start_node_impl` is a bit more complex now, mostly to handle the fact that the altair & centrifuge nodes aren't EVM nodes yet. A lot of this can be simplified back down to how it was once all the runtimes support EVM.

src/service.rs
mustermeiszer: Does this open/create a db under the same path as the parachain one? If so, I would maybe tend to go a route where this is another path with a clear prefix maybe?

src/service.rs
mustermeiszer: Can't we have a command for running normal chain and one for running chain with EVM?

src/service.rs
```diff
@@ -889,6 +1051,70 @@ pub async fn start_development_node(
...
+	);
+
+	// Spawn Frontier EthFilterApi maintenance task.
+	// Each filter is allowed to stay in the pool for 100 blocks.
+	const FILTER_RETAIN_THRESHOLD: u64 = 100;
+	task_manager.spawn_essential_handle().spawn(
+		"frontier-filter-pool",
+		None,
+		EthTask::filter_pool_task(client.clone(), filter_pool, FILTER_RETAIN_THRESHOLD),
+	);
```
mustermeiszer, line 1107: What are filters good for?

mustermeiszer (commented):
I would move the evm setup into its seperate file, just like XCM stuff. 
And are there some notes on what we add here and what is the reasoning? 
Like what is the usage of those two: 
* node-side frontier [tasks](https://github.com/PureStake/frontier/blob/moonbeam-polkadot-v0.9.32/template/node/src/service.rs#L724-L772)
* node-side [import queue](https://github.com/PureStake/frontier/blob/moonbeam-polkadot-v0.9.32/template/node/src/service.rs#L724-L772)
 

Regarding the service: 
* Do you think it is possible to create a seperate command, that will only then use an start node impl that includes all the EVM stuff? 

Overall, I think it makes sense to let this only life in a feature branch. It seems to be breaking too much stuff on the service that I would not want to add to the production env. WDYTY?

src/service.rs
mikiquantum: The service file is by far the most complex piece to maintain at "parity" (no pun intended) with Polkadot/Substrate upgrades. If separating in two different implementations is doable, I would go with that route.

runtime/development/src/lib.rs
```diff
@@ -1674,6 +1685,94 @@ impl pallet_liquidity_rewards::Config for Runtime {
...
+	type OnChargeTransaction = ();
+	type PrecompilesType = CentrifugePrecompiles<Self>;
+	type PrecompilesValue = PrecompilesValue;
+	type Runner = pallet_evm::runner::stack::Runner<Self>;
+	type RuntimeEvent = RuntimeEvent;
+	type WeightPerGas = WeightPerGas;
+	type WithdrawOrigin = EnsureAddressTruncated;
+}
+
+impl pallet_evm_chain_id::Config for Runtime {}
```
mikiquantum, line 1742: How is the chainId value set for an ongoing running chain? Manual raw storage set?

runtime/development/src/lib.rs
```diff
@@ -1803,9 +1908,10 @@ pub type SignedExtra = (
 );
 /// Unchecked extrinsic type as expected by this runtime.
 pub type UncheckedExtrinsic =
-	generic::UncheckedExtrinsic<Address, RuntimeCall, Signature, SignedExtra>;
+	fp_self_contained::UncheckedExtrinsic<Address, RuntimeCall, Signature, SignedExtra>;
 /// Extrinsic type that has already been checked.
-pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, RuntimeCall, SignedExtra>;
+pub type CheckedExtrinsic =
+	fp_self_contained::CheckedExtrinsic<AccountId, RuntimeCall, SignedExtra, H160>;
```
mikiquantum, line 1914: Would still be possible to validate calls the non-evm way? I am wondering on changes needed on clients like GSRPC

src/service.rs
branan: It creates a separate `frontier/db` next to the main `db` folder:

```
branan@branan-laptop:~/proj/centrifuge/centrifuge-chain$ tree -d /tmp/centrifuge-chain/parachain/2000/data/chains/devel_local/
/tmp/centrifuge-chain/parachain/2000/data/chains/devel_local/
├── db
│   └── full
├── frontier
│   └── db
├── keystore
└── network

6 directories
```

src/service.rs
branan: My main concern with splitting into two different impls was exactly the maintenance cost - we'll need to update both `start_node_impl` versions whenever we do a substrate upgrade. But I also acknowledge that the code as written here makes the node startup flow even more complex than it used to be (and it was already pretty complex)
I'm happy to just create a second `start_evm_node_impl` that we can just migrate other runtimes to as they add EVM suport.

runtime/development/src/precompiles.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl<R> PrecompileSet for CentrifugePrecompiles<R>
+where
+	R: pallet_evm::Config,
+	R::RuntimeCall: Dispatchable<PostInfo = PostDispatchInfo> + GetDispatchInfo + Decode,
+	<R::RuntimeCall as Dispatchable>::RuntimeOrigin: From<Option<R::AccountId>>,
+{
+	fn execute(&self, handle: &mut impl PrecompileHandle) -> Option<PrecompileResult> {
+		// 1025 is chosen for compatibility with Moonbeam
+		if handle.code_address() == addr(1025) {
```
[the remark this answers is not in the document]
branan: It should be a `const` anyway ;)

I'd rather not rely on `is_precompile` in `execute` since we'll need to rewrite it again to use a match on addresses when we add more precompiles in the future.

runtime/development/src/lib.rs
branan: It comes from genesis. For adding EVM to a running chain we'll need to implement a migration (we'll also need that to define the "EVM Genesis Block" which will be based on the first block which contains EVM support. Lots of fun :) )

runtime/development/src/lib.rs
branan: `fp_self_contained` is pretty much a transparent wrapper that just exposes the necessary trait impls for handling EVM calls in the runtime API. I looked into the code to see if it would break SCALE-encoded compatibility with the existing Extrinsic/Call structures, and AFAICT it's safe

runtime/development/src/precompiles.rs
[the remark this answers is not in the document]
mikiquantum: Oh, my point is mostly about having to touch in two places (might oversee it) when updating that constant. But I do not know enough of how is the `is_precompile` used or will be used.

runtime/development/src/lib.rs
mikiquantum: Got it 👍

runtime/development/src/lib.rs
mikiquantum: Cool, and I assume the extrinsic payload to be signed is the same? Or a client would need to form the extrinsic to be signed differently?

mikiquantum (commented):
Nice! thanks @branan !

branan:
I'm working on rebasing this over the 0.9.37 work at the same time as shuffling the code around - there are major conflicts in the changes to services code, and I don't wan to do two refactors.

src/service.rs
mustermeiszer: @branan is there already some progress on splitting the two?

mustermeiszer (commented):
@branan do you have an ETA when we can merge this into main with a split node implementation? 
What is left to do for that?

branan:
Node and runtime EVM code has been split into separate modules now, and I've done a few more small cleanups. I'll ticket the remaining things to "productionize" the code, but otherwise this should be good to land as soon as 0.9.37 is merged. Tomorrow, if all the release shenanigans go correctly?

For now this is targeted at the 0.9.37 branch so that the diff is readable/manageable. It will rebase cleanly onto `main` once the 0.9.37 work is in.

branan:
Nix build failure is the same as 0.9.37 branch

branan:
`rustdoc` and `fmt` failures also look like they should be from the base 0.9.37 PR? I haven't touched the files that failed here

mustermeiszer (dismissed):
The split looks good. Thanks @branan!

branan:
@offerijns @NunoAlexandre @lemunozm This is ready for final review for merge into the dev environment. There's still some work to do for making it production-ready for other runtimes, but it's in a state where things that API consumers will interact with are stable.

runtime/development/src/lib.rs
```diff
@@ -2014,6 +2112,160 @@ impl_runtime_apis! {
...
+				let mut config = <Runtime as pallet_evm::Config>::config().clone();
+				config.estimate = true;
+				Some(config)
+			} else {
+				None
+			};
+
+			let is_transactional = false;
+			let validate = true;
+			let evm_config = config.as_ref().unwrap_or_else(|| <Runtime as pallet_evm::Config>::config());
```
lemunozm, line 2205:
```suggestion
            let mut config = <Runtime as pallet_evm::Config>::config();
			config.estimate = estimate;
			
			let is_transactional = false;
			let validate = true;
```
I think the following is equivalent 🤔

src/chain_spec.rs
```diff
@@ -762,6 +762,10 @@ fn endowed_accounts() -> Vec<cfg_primitives::AccountId> {
 		get_account_id_from_seed::<sr25519::Public>("Dave//stash"),
 		get_account_id_from_seed::<sr25519::Public>("Eve//stash"),
 		get_account_id_from_seed::<sr25519::Public>("Ferdie//stash"),
+		// Ethereum account 0x7F429e2e38BDeFa7a2E797e3BEB374a3955746a4
+		// Private key 0x4529cc809780dcc4bf85d99e55a757bc8fb3262d81fae92a759ec9056aca32b7
+		// kAJigCQycQoANDeHpjoSDt8hiXXqj9AmEvxfyzduSLDohfeZP
```
lemunozm, line 767: Double-check, should this be here?

src/service/evm.rs
```diff
@@ -0,0 +1,542 @@
...
+		ParachainBlockImport<RuntimeApi>,
+		Option<&Registry>,
+		Option<TelemetryHandle>,
+		&TaskManager,
+		Arc<dyn RelayChainInterface>,
+		Arc<sc_transaction_pool::FullPool<Block, FullClient<RuntimeApi>>>,
+		Arc<NetworkService<Block, Hash>>,
+		SyncCryptoStorePtr,
+		bool,
+	) -> Result<Box<dyn ParachainConsensus<Block>>, sc_service::Error>,
```
lemunozm, line 312: I think this is my record of seeing bound lines in a function 🏅

lemunozm (commented):
A very fast review, but I feel like I need to stay a long time to make a useful review. I'll try do this tomorrow. 

Could you have any related spec document about this?

runtime/development/src/evm.rs
```diff
@@ -0,0 +1,113 @@
...
+use pallet_evm::{AddressMapping, EnsureAddressTruncated};
+use sp_core::{crypto::ByteArray, H160, U256};
+use sp_runtime::{traits::AccountIdConversion, Permill};
+use sp_std::marker::PhantomData;
+
+use crate::{AccountId, Aura};
+
+pub mod precompile;
+
+pub struct FindAuthorTruncated<F>(PhantomData<F>);
```
NunoAlexandre, line 25: Can we document this?

runtime/development/src/evm.rs
```diff
@@ -0,0 +1,113 @@
...
+		if let Some(author_index) = F::find_author(digests) {
+			let authority_id = Aura::authorities()[author_index as usize].clone();
+			return Some(H160::from_slice(&authority_id.to_raw_vec()[4..24]));
+		}
+		None
+	}
+}
+
+#[derive(Encode, Decode, Default)]
+struct EthereumAccount(H160);
```
NunoAlexandre, line 40: Since this module is called `evm` I'd suggest we rename this to just `Account` or `Address`:

```suggestion
struct Account(H160);
```

runtime/development/src/evm.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl sp_runtime::TypeId for EthereumAccount {
+	const TYPE_ID: [u8; 4] = *b"ETH\0";
+}
+
+pub struct ExtendedAddressMapping;
+
+impl AddressMapping<AccountId> for ExtendedAddressMapping {
```
NunoAlexandre, line 48: This converts accounts of two different networks, it would be cool to refer to `AccountId` with a full path to make that more obvious. Like `fn into_account_id(address: H160) -> cfg_types::AccountId {`

runtime/development/src/evm.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub struct BaseFeeThreshold;
+
+impl pallet_base_fee::BaseFeeThreshold for BaseFeeThreshold {
+	fn lower() -> Permill {
+		Permill::zero()
+	}
+
+	fn ideal() -> Permill {
+		Permill::from_parts(500_000)
```
NunoAlexandre, line 62: Where do we get these from?

runtime/development/src/evm.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub struct ExtendedAddressMapping;
+
+impl AddressMapping<AccountId> for ExtendedAddressMapping {
+	fn into_account_id(address: H160) -> AccountId {
+		EthereumAccount(address).into_account_truncating()
+	}
+}
+
+pub struct BaseFeeThreshold;
```
NunoAlexandre, line 54: It would be cool to document all of these actually

runtime/development/src/evm/precompile.rs
```diff
@@ -0,0 +1,46 @@
+use core::marker::PhantomData;
```
NunoAlexandre, line 1: Don't we miss the license header here?

runtime/development/src/evm/precompile.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl<R> CentrifugePrecompiles<R> {
+	#[allow(clippy::new_without_default)] // We'll never use Default and can't derive it.
+	pub fn new() -> Self {
+		Self(Default::default())
+	}
+}
+
+// 1025 is chosen for compatibility with Moonbeam
+const DISPATCH_ADDR: H160 = addr(1025);
```
NunoAlexandre, line 18: Can you explain why the value 1025 grants that compatibility?

runtime/development/src/evm/precompile.rs
```diff
@@ -0,0 +1,46 @@
...
+		address == DISPATCH_ADDR
+	}
+}
+
+const fn addr(a: u64) -> H160 {
+	let b = a.to_be_bytes();
+	H160([
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, b[0], b[1], b[2], b[3], b[4], b[5], b[6], b[7],
+	])
+}
```
NunoAlexandre, line 46: Can we document what address this is building and why it's build this way?

runtime/development/src/lib.rs
```diff
@@ -1764,6 +1773,12 @@ construct_runtime!(
 		Migration: pallet_migration_manager::{Pallet, Call, Storage, Event<T>} = 199,
 		// admin stuff
 		Sudo: pallet_sudo::{Pallet, Call, Config<T>, Storage, Event<T>} = 200,
+
+		// EVM pallets
+		EVM: pallet_evm::{Pallet, Config, Call, Storage, Event<T>} = 160,
```
NunoAlexandre, line 1778: It seems that we are indexing the evm pallets with a lower number than pallets above that; is that intentional? Shouldn't we move them up to be sorted ASC?

src/rpc/evm.rs
```diff
@@ -0,0 +1,232 @@
...
+use fc_db::Backend as FrontierBackend;
+pub use fc_rpc::{
+	EthBlockDataCacheTask, OverrideHandle, RuntimeApiStorageOverride, SchemaV1Override,
+	SchemaV2Override, SchemaV3Override, StorageOverride,
+};
+pub use fc_rpc_core::types::{FeeHistoryCache, FeeHistoryCacheLimit, FilterPool};
+use fp_rpc::{ConvertTransaction, ConvertTransactionRuntimeApi, EthereumRuntimeRPCApi};
+use fp_storage::EthereumStorageSchema;
+use jsonrpsee::RpcModule;
+// Substrate
```
NunoAlexandre, line 25:
```suggestion
```

NunoAlexandre (commented):
Left many comments but no blockers but no blockers as nothing pops up as wrong. I am, however, a bit nervous to merge this amount of code without any added test coverage, which I would love to see in place, at the very least to test account version between the Substrate and the Evm sides 👀

runtime/development/src/evm/precompile.rs
branan: Yes! this is actually a workaround because the H160 method to convert from a u64 isn't const. I need to file an issue upstream with Parity for that still (I might even just put in a PR to make the methods const)

src/service/evm.rs
lemunozm: but bounds are cool 😎

runtime/development/src/lib.rs
```diff
@@ -2014,6 +2112,160 @@ impl_runtime_apis! {
 		}
 	}
 
+	// Frontier APIs
+	impl fp_rpc::EthereumRuntimeRPCApi<Block> for Runtime {
```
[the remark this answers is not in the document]
branan: Yup, I missed this in the move commit

src/service/evm.rs
lemunozm: Maybe, as an idea for this case, we could use a trait similar to `Config` is used for pallets, declaring all associated types with the required bounds and passing only that high-level trait.

runtime/development/src/evm.rs
branan: Totally copy and pasted from the frontier template. These constants are on the list of "things to understand before it lands in production". But I don't think it needs to be locked down to land it in dev

runtime/development/src/evm.rs
branan: I honestly don't 100% understand the fee pallets, but I'll do my best

runtime/development/src/evm/precompile.rs
branan: Good catch, will add

runtime/development/src/evm/precompile.rs
branan: Because it's the value that Moonbeam chose. As far as I know it was arbitrary on their end. But it made sense to me to have common precompiles have the same address within the polkadot ecosystem.

src/rpc/evm.rs
```diff
@@ -0,0 +1,232 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version (see http://www.gnu.org/licenses).
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+use std::{collections::BTreeMap, sync::Arc};
+
+// Frontier
```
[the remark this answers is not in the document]
branan: Leftover from moving the code out of the shared RPC namespace. Will remove.

src/rpc/evm.rs
branan: as above

runtime/development/src/lib.rs
NunoAlexandre: I think that's fine but I would still move it up to match the order

src/service/evm.rs
branan: I would normally agree that moving it to a trait would be good. But since the service code is usually the hardest to fix during a Polkadot upgrade, I think it is worthwhile to have it match what is in the Substrate/Frontier examples projects. This way we can more easily look at the changes made to those examples and copy the appropriate changes to our code.

The more we diverge from those example codebases, the harder Polkadot upgrades will be

src/chain_spec.rs
```diff
@@ -1084,6 +1088,10 @@ fn development_genesis(
 		parachain_system: Default::default(),
 		treasury: Default::default(),
 		interest_accrual: Default::default(),
+		base_fee: Default::default(),
+		evm_chain_id: development_runtime::EVMChainIdConfig { chain_id: 999_999 },
```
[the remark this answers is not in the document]
branan: Generally test networks have different IDs in the Ethereum ecosystem. That would mean even Catalyst should have a unique ID from the production Centrifuge chain. It is a 64-bit ID, so there is lots of room in the namespace.

I think it's better that any code we write (even tests) be chain-id-agnostic, rather than breaking the convention and reusing production chain IDs for local networks

src/service/evm.rs
lemunozm: Sure! I simply left that thought 👍🏻

runtime/development/src/evm.rs
wischli: I suppose all of the minor open TODOs will be handled in a subsequent PR?

runtime/development/src/lib.rs
wischli: Yes, we should move it up in this case. However, we should always be careful with these things as the order of the pallet in the `construct_runtime` macro decide the order of the `on_initialize` (in-order) and `on_finalize` (reverse of order) executions. 

So for the crucial part, the indices are irrelevant. Moreover, indices should not be swapped and most ideally not be reused. The later part will be interesting to see in the future, given we only have `u8` supported. But Kusama/Polkadot will hit this limitation much earlier 😅

runtime/development/src/evm.rs
branan: @wischli Yeah. Most of the TODOs are "make things configurable" or "understand this enough to decide for ourselves instead of copying the template". That's all part of what I've been calling the "productionization" work.

runtime/development/src/lib.rs
branan: I've been trying a bit to think about how the separation of "API stability" and "ABI stability" that I'm familiar with from the C world map to Substrate - things like indices I think are a big part of this. Let's chat about it some this week if we have time :)

Issue #1263, opened 2023-03-16
Improve CI times


lemunozm:
### Which part of the code is the issue addressing?

This issue tracks things that can be done in CI script to reduce the job times.

- Compile in debug: #1238 
- Seems that `taplo` is installed and executed in two jobs `Tests` and `Lints`. The `taplo` part takes around 7 min. We can remove it from `Tests`, which is the longest job, and save 7 min per PR.
- `tests.yml` runs over 2 targets: `tests` and `integration` maybe we can merge them to reuse the compilation. 
  - *Could we compile integration tests with `runtime-benchmarks` and `try-runtime`?*
- `rust-cache` seems not to be working, we are wasting around 10 mins saving the cache, but the next job fails to recover it. 😵‍💫 
  - *Is it because we are cleaning the docker in "Prep build on Ubuntu"?*

PR #1264, opened 2023-03-17
Remove taplo task from tests job

lemunozm:
# Description

The Taplo task is already in the `Lints` job. We could alleviate the longer job around 5-7 min removing it from this.

Related issue: #1263


mustermeiszer (dismissed):
If it works I am fine! Thanks again for the insane care you take! 🚀

branan:
Looks like this just failed with the disk space issue, so should be fine after a rebase.

mustermeiszer (approved):
Re-approve

PR #1266, opened 2023-03-17
Altair loans sustitution

### LINKED ISSUE #1265: Pallet loans sustitution in altair

This PR removes pallet_loans from development and adds pallet_loans-ref

**WARNING, double-check if we need any migrations in altair:**
- A migration to remove the current `pallet-loans` storage? I'm pretty sure we need this to avoid data collisions with the new pallet (the new pallet will have the same name, and some storage will have the same name too).
- After add this pallet, next calls from `pallet-pool-system` to `nav()` will return 0. Is it ok for the pools?

lemunozm:
Fixes #1265 #1281

Tasks:

- [x] Substitute `pallet-loans` by `pallet-loans-ref` in `altair`.
- [x] Make a migration in `pallet-loans-ref` that nukes the previous `pallet-loans` storage only in case it is the old one.
- [x] Removed deprecated `pallet-loans` dependency and others associated with it in `pallet-connectors`

**`pallet-loans` is now safe to be removed from `centrifuge-chain` if this PR is merged (do not need to wait for the migration).**


mustermeiszer (commented):
We definietly need to check needed migrations. For this we need to
a) query the current state of Altair 
b) we decide to nuke the old loans storage

@offerijns what is your thinking here? 

For option b) we also need a migration logic that nukes the old storage.

offerijns:
> b) we decide to nuke the old loans storage

This is fine, on Altair. There's only 1 pool with 1 closed loan.

pallets/loans-ref/src/migrations.rs
```diff
@@ -0,0 +1,87 @@
...
+		ensure!(
+			v0::NextLoanId::<T>::iter_values().count() == 1,
+			"Pallet loans contains doesn't contain old data"
+		);
+
+		Ok(Vec::new())
+	}
+
+	fn on_runtime_upgrade() -> Weight {
+		if v0::NextLoanId::<T>::iter_values().count() == 1 {
```
lemunozm, line 45: Ideally, I would have wanted to check here if the `on_chain_storage_version()` was 0, and bump the version to 1, but I could not do it because if I initialize the pallet with `STORAGE_VERSION = 1`, then it never enters here, because it will be already 1.

I think the issue was not setting the version to 0 in the past.

pallets/loans-ref/src/lib.rs
```diff
@@ -93,8 +94,11 @@ pub mod pallet {
 		<T as Config>::CurrencyId,
 	>>::PoolId;
 
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
```
lemunozm, line 97: We initialize the version to 0, for futures migrations

pallets/loans-ref/src/migrations.rs
```diff
@@ -0,0 +1,87 @@
...
+			let result = storage::unhashed::clear_prefix(&loan_prefix(), None, None);
+
+			match result.maybe_cursor {
+				None => log::info!("Loans: storage cleared successful"),
+				Some(_) => log::error!("Loans: storage not totally cleared"),
+			}
+
+			return T::DbWeight::get().writes(result.unique.into())
+				+ T::DbWeight::get().reads(result.loops.into());
```
lemunozm, line 54: It is expected 8 read and 8 writes

lemunozm (commented):
Some explanations below:

pallets/loans-ref/src/migrations.rs
lemunozm: > I think the issue was not setting the version to 0 in the past.

This makes me think that maybe we should add this version for all pallets, in this way, if in the future we have a new migration, we can easily fetch the correct version number 0 for them.

wischli: Yes, I have been meaning to open an issue for that. [Here it is](https://github.com/centrifuge/centrifuge-chain/issues/1295) 🌚

pallets/loans-ref/src/migrations.rs
```diff
@@ -0,0 +1,90 @@
...
+				None => log::info!("Loans: storage cleared successful"),
+				Some(_) => log::error!("Loans: storage not totally cleared"),
+			}
+
+			T::DbWeight::get().writes(result.unique.into())
+				+ T::DbWeight::get().reads(result.loops.into())
+		} else {
+			log::warn!("Loans: storage was already clear. This migration can be removed.");
+
+			Weight::zero()
```
wischli, line 59: AFAIK we still perform a `read`

```suggestion
			T::DbWeight::get().reads(1)
```

wischli (approved):
LGTM! Me gusta the check via the deprecated Storage entry 🥳

pallets/loans-ref/src/migrations.rs
lemunozm: Great! ❤️

pallets/loans-ref/src/migrations.rs
lemunozm: You're right!

lemunozm: Following your hint, I've made a minor change for the sake of purism

wischli (approved):
Re-approve

pallets/loans-ref/src/migrations.rs
```diff
@@ -0,0 +1,91 @@
...
+
+mod v0 {
+	use super::*;
+
+	/// This storage comes from the previous pallet loans.
+	/// It is used as an indicator that the previous pallet loans still exists.
+	/// If this storage is not found, the nuking process is aborted.
+	#[storage_alias]
+	pub(crate) type NextLoanId<T: Config> =
+		StorageMap<Pallet<T>, Blake2_128Concat, PoolIdOf<T>, u128, ValueQuery>;
```
mustermeiszer, line 23: Do we get the right prefix as `pallet-loans-ref` and `pallet-laons` are both using the same prefix (i.e. `Loans`)?

mustermeiszer (approved):
Just want to ensure we are getting the right prefix. I might hardcode it, but I think the given approach is fine to.

NunoAlexandre (approved):

PR #1275, opened 2023-03-21
Pallet `weights.rs` files without generation

### LINKED ISSUE #1267: Remove pallet `weights.rs` generation

* Refactoring

### Description

`weights.rs` is something that should be ideally generated always using a runtime. There is no reason to generate them inside the pallets. 

We should remove the `weights.rs` generation for pallets. We only need to generate benchmarks under the runtimes.

- Reduce benchmark development flow.
- Simplify things.

lemunozm:
# Description

Fixes #1267

## Changes and Descriptions

- Remove `run_benchmark.sh` script because it's no longer used after this PR.
- Clear `weights.rs` files from all pallets.
  - If there are benchmarks associated, the implementation contains `Weight::zero()`
  - If there are no benchmarks, the implementation contains a default weight value high enough to avoid security issues.
- Ensure all pallets from `centrifuge`/`altair` use their runtime-generated weights.(**double-check this**)
- Removed `bridge-mapping` stuff that seems to be deprecated (**check this**).

## Type of change

- Refactor

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -89,4 +89,9 @@ onboard-parachain)
   cd scripts/js/onboard
   yarn && yarn execute "ws://0.0.0.0:9944" "//Alice" ${para_id} "${genesis}" $wasm_location
   ;;
+
+benchmark)
+  pallet=$2
+  cargo run --features runtime-benchmarks benchmark pallet --chain $parachain --pallet="$pallet" --extrinsic=*
+  ;;
 esac


runtime/centrifuge/src/weights/mod.rs
```diff
@@ -23,6 +23,7 @@ pub mod pallet_migration_manager;
 pub mod pallet_multisig;
 pub mod pallet_preimage;
 pub mod pallet_proxy;
+pub mod pallet_restricted_tokens;
```
lemunozm, line 26: We have been using the weights from the pallet not from the runtime for `centrifuge`. Not sure if this was the intended behavior. I've changed it in order to use the correct weights from now

NunoAlexandre (dismissed):
🧹 ✨

scripts/run_benchmark.sh
mustermeiszer, line 1: @lemunozm can we please keep this one. It definitely makes sense to run benchmarks on a single pallet for testing puposes. The other script runs it for all. I would adapt the command to just not have template and ouput. If this is possible.

mustermeiszer (changes requested):

scripts/run_benchmark.sh
lemunozm: Ok, it can be simplified a lot, probably even do not need its own file, and can be run directly from `init.sh` as `./scripts/init.sh benchmark <runtime> <pallet>`, WDYT?

lemunozm: The simplified command is 
```sh
cargo run --features runtime-benchmarks benchmark pallet --pallet <pallet> --chain <runtime> --extrinsic="*"
```

lemunozm: Done and tested

mustermeiszer (approved):

Issue #1269, opened 2023-03-17
pallet-loans: multiples loan actions


lemunozm:
### Which part of the code is the issue addressing?

* Feature

### Description

Add the capability of performing multiple actions per loan. We need this to reduce the PoV overhead of multiple loan actions.

Ideas to implement:
- Create new extrinsics for different loan actions with `_mutiple` suffix.
- Make current extrinsics to accept a `BoundedVec` of their parameters.
- New extrinsic: `loan_action_multiple(`LoanAction`)` which accept several actions. `LoanAction` could be an `enum` with all parameters founds in `borrow`, `repay`, `write-off`, ...

### What are forseen obstacles or hurdles to overcome?

- Is this really needed if Substrate reverts the PoV when the same storage is overwritten? [slack thread](https://kflabs.slack.com/archives/C04HJN404RK/p1679051358713519)

PR #1283, opened 2023-03-24
Added `MutatedActiveLoan` storage for PoV optimization

### LINKED ISSUE #1193: Loans: Add `ActiveLoanMutations` storage to avoid PoV limits

* Feature

### Description

Most functions as `borrow\repay\write-off` should not write the whole `ActiveLoans` vector each time each loan changes. Instead, it should append an `ActiveLoanMutation` with the differential data that has changed.

These `ActiveLoanMutations` will be added to the `ActiveLoans` in `update_nav`

### Research/based on

This slack [thread](https://kflabs.slack.com/archives/C04HJN404RK/p1675093607773169)

- This would avoid reach the PoV limits of Substrate Weight V2.

- Minimize the computation impact on `update_nav` method.

lemunozm:
# Description

Fixes #1193

### Tasks
- [x] Implementation
- [x] Test passing
- [ ] Update benchmarks

## Changes and Descriptions

- Added new storage `MutatedActiveLoans`.
- Remove not needed `loan_id` from `ActiveLoan`.

## Type of change

- Breaking change

### DIFF
--- a/libs/traits/src/ops.rs
+++ b/libs/traits/src/ops.rs
@@ -4,6 +4,24 @@ pub use ensure::{
 	EnsureSubAssign,
 };
 
+pub mod storage {
+	use sp_runtime::{traits::Get, BoundedVec};
+
+	pub trait BoundedVecExt<T> {
+		/// Same as [`BoundedVec::try_push()`] but returns a reference to the pushed element in the
+		/// vector
+		fn try_push_fetch(&mut self, element: T) -> Result<&mut T, T>;
+	}
+
+	impl<T, S: Get<u32>> BoundedVecExt<T> for BoundedVec<T, S> {
+		fn try_push_fetch(&mut self, element: T) -> Result<&mut T, T> {
+			let len = self.len();
+			self.try_push(element)?;
+			Ok(self.get_mut(len).expect("This can not fail. qed"))
+		}
+	}
+}
+
 mod ensure {
 	use sp_runtime::{
 		traits::{CheckedAdd, CheckedDiv, CheckedMul, CheckedSub, Zero},
--- a/pallets/loans-ref/src/lib.rs
+++ b/pallets/loans-ref/src/lib.rs
@@ -401,10 +401,10 @@ pub mod pallet {
 				Some(created_loan) => {
 					Self::ensure_loan_borrower(&who, created_loan.borrower())?;
 
-					let mut active_loan = created_loan.activate(loan_id)?;
+					let mut active_loan = created_loan.activate()?;
 					active_loan.borrow(amount)?;
 
-					Self::insert_active_loan(pool_id, active_loan)?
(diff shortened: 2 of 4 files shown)


lemunozm:
I think this change adds a lot of complexity to the code and makes the benchmarks difficult to measure correctly. Please double-check this. Is this PR really really needed if we get #1269 done?

libs/traits/src/ops.rs
```diff
@@ -4,6 +4,24 @@ pub use ensure::{
 	EnsureSubAssign,
 };
 
+pub mod storage {
+	use sp_runtime::{traits::Get, BoundedVec};
+
+	pub trait BoundedVecExt<T> {
+		/// Same as [`BoundedVec::try_push()`] but returns a reference to the pushed element in the
+		/// vector
+		fn try_push_fetch(&mut self, element: T) -> Result<&mut T, T>;
```
lemunozm, line 13: A better name for this method?

mustermeiszer: `push_fetch_(trying_to_at_least)`

mustermeiszer:
> I think this change adds a lot of complexity to the code and makes the benchmarks difficult to measure correctly. Please double-check this. Is this PR really really needed if we get https://github.com/centrifuge/centrifuge-chain/issues/1269 done?

I agree. We should definitely wait till PoV is benched too. 

And given a second thought, why would this even be a reduction in PoV size? It rather looks like an increase, but maybe I am getting the size impact not correctly

For a successful state transition from block `x` to block `x+1`  the relay chain needs 
* block `x+1` &rarr; extrinsics and final root
* `(key, value)`-pairs of touched keys at the state of `x`

if this assumption is correct, than we will anyways send the whole vector all the time, due to reading.

lemunozm:
I think (or I understand from previous conversations) each extrinsic is measured isolated for the PoV Size. 

So, for each active loan modification, we are writing now into `MutatedActiveLoans` storage which is smaller than `ActiveLoans`, so we are not increasing too much the PoV Size.

This increases the PoV Size for `portfolio_valuation`, but it's ok if other calls are smaller. Statistically, if we sum all extrinsic done in `pallet-loans` per block, the PoV Size will be smaller than writing always the whole `ActiveLoan` vector for each modification of a loan.

libs/traits/src/ops.rs
lemunozm: I do not get well what you mean here 🙂

mustermeiszer:
> I think (or I understand from previous conversations) each extrinsic is measured isolated for the PoV Size.

Ahh, yeah, that is correct and a benchmarking issue... mmh

mustermeiszer:
cc @lemunozm

lemunozm:
I think we can close it 👍🏻. In case we want to go with this solution we can reopen, although I think the database has changed a lot from this.

PR #1285, opened 2023-03-27
Remove duplicate metadata from pool-system

### LINKED ISSUE #1261: Inconsistent Pool Metadata Hash Between PoolRegistry.poolMetadata and PoolSystem.pool methods

* [X] Addressing a bug
* [ ] Refactoring
* [ ] Feature

### Description

`api.query.poolRegistry.poolMetadata()` and `api.query.poolSystem.pool()` deliver different IPFS hashes for the same pool. 
Results from `api.query.poolSystem.pool()` seem not to be updated after initialization.

### Research/based on

Dev Live
centrifuge-devel/1007

Centrifuge Parachain Collator
 v0.10.25
api v10.1.3
apps v0.124.2-161

### Steps to reproduce
image
image

branan:
# Description

Remove duplicate metadata hash stored in the 

Fixes #1261

## Changes and Descriptions

* Remove metadata hash from `pool-system` pallet
* Remove unused error variant from `pool-system`
* Update benchmarks and tests for removal
* Update all runtimes for removal

# Checklist:


branan:
@mustermeiszer @lemunozm This is targeting the branch to add pools to Centrifuge since it will conflict with that branch, and since we need to make sure to include it as part of the final inclusion of the loans refactor + release.

As long as this ships to altair in the same release as the loans refactor (or at least, ships after we delete the existing test pool but before we add a new one), we don't need to bother with a migration.

lemunozm (dismissed):
LGTM!

> we don't need to bother with a migration.

Do you mean now? We are modifying the pools storage, although we are not shipping this to `altair`, we need to do some kind of migration for `altair` before we launch there, isn't it?

mustermeiszer (commented):
Nice! But we need a migration for Altair, right?

branan:
I think I misunderstood how much of the pools/loans storage we were removing, I had assume we would nuke the whole testing pool from Altair. If we're only going to do loans, then we will indeed need a migration here.

I got it half done already before I had the realization about removing storages, so I'll get it wrapped up and pushed :+1:

branan:
Since this is no longer tied to the loans refactor/release schedule, I'm going to retarget it at `main`. We can try to land it as part of the initial release on Centrifuge to avoid the migration, but it's not critical.

pallets/pool-system/src/lib.rs
```diff
@@ -165,14 +167,29 @@ pub type PoolEssenceOf<T> = PoolEssence<
...
+pub enum Release {
+	V0,
+	V1,
+}
+
+impl Default for Release {
+	fn default() -> Self {
+		Self::V0
+	}
+}
```
mustermeiszer, line 181: Can we use the new versioning that is automatically supported for pallets. Cc @wischli for how this works ^^

mustermeiszer (commented):
Can we maybe even nuke the storage? Easier to migrate, but the migration itself already looks good at the first view.

pallets/pool-system/src/lib.rs
branan: My understanding of that is that its based around the crate's version number. We... don't really use/maintain those right now. I agree we should move to the new versioning, though.

My thought was that we'll need to figure out some sort of versioning / release-stream system anyway once pools are on cent-chain (because we won't want new features to land simultaneously on altair & cfg). We should roll pallet versioning into that, and then we can switch to the new pallet version storage. WDYT? (again /cc @wischli )

branan:
As far as nuking goes, that's more complicated (many additional storage keys) and involves two pallets (this one and `pool-registry`). I don't think it's worth it, necessarily?

pallets/pool-system/src/lib.rs
mustermeiszer: I think it is a basic u16 value.

mustermeiszer (approved):
I would merge this directly after #1233.

offerijns (dismissed):

offerijns (approved):

PR #1286, opened 2023-03-27
Altair migration to remove previous pallet-loans storage

### LINKED ISSUE #1281: Remove storage from pallet-loans in Altair
### Description

Currently, Altair has only 1 closed loan.
We can remove it, nuking the storage of the pallet in a migration.

- Unlock the new `pallet-loans` to live in `altair` (the refactor)
- Allow removing the old pallet loans from the database, which unlocks/facilitates:
  - #1189
  - #1231
  - #1203
  - #908

Wait for the next release

lemunozm:
# Description

Fixes #1281

## Changes and Descriptions

- Added migration to remove all storages of `pallet-loans`. 
- ✅ Tested against `altair` using:
  ```sh
  RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace cargo run --release --features try-runtime try-runtime --execution Native --chain altair-local --no-spec-check-panic on-runtime-upgrade live --uri wss://fullnode.altair.centrifuge.io:443
  ```

### DIFF
--- a/pallets/loans/src/migrations.rs
+++ b/pallets/loans/src/migrations.rs
@@ -1,0 +1,36 @@
+use frame_support::{storage, traits::OnRuntimeUpgrade, weights::Weight};
+use sp_std::vec::Vec;
+
+use crate::*;
+
+pub struct Migration<T>(sp_std::marker::PhantomData<T>);
+
+impl<T: Config> OnRuntimeUpgrade for Migration<T> {
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
+		ensure!(
+			ClosedLoans::<T>::iter_values().count() == 1,
+			"There is not a closed loan"
+		);
+
+		Ok(Vec::new())
+	}
+
+	fn on_runtime_upgrade() -> Weight {
+		let loans_module = b"4c82a580ac33cceba8ed9766387f22b7";
+		let _ = storage::unhashed::clear_prefix(loans_module, None, None);
+
+		// Should be one/few elements per storage map in `pallet-loans`. The next number should be
+		// enough.
+		Weight::from_ref_time(200_000_000)
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(_: Vec<u8>) -> Result<(), &'static str> {
+		ensure!(
+			ClosedLoans::<T>::iter_values().count() == 0,
+			"There is still a closed loan"
+		);
+		Ok(())
+	}
+}
--- a/pallets/loans/src/migrations.rs
+++ b/pallets/loans/src/migrations.rs
@@ -1,4 +1,5 @@
 use frame_support::{traits::OnRuntimeUpgrade, weights::Weight};
+#[cfg(feature = "try-runtime")]
 use sp_std::vec::Vec;
 
 use crate::*;
--- a/pallets/loans/src/migrations.rs
+++ b/pallets/loans/src/migrations.rs
@@ -1,36 +1,88 @@
-use frame_support::{storage, traits::OnRuntimeUpgrade, weights::Weight};
+use frame_support::{traits::OnRuntimeUpgrade, weights::Weight};
 use sp_std::vec::Vec;
 
 use crate::*;
 
+/// This migration nukes all storages from the pallet individually.
 pub struct Migration<T>(sp_std::marker::PhantomData<T>);
 
 impl<T: Config> OnRuntimeUpgrade for Migration<T> {
 	#[cfg(feature = "try-runtime")]
 	fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
+		// The current state from altair when this migration should be done has only one element
+		// in each store.
+		ensure!(
+			PoolToLoanNftClass::<T>::iter_values().count() == 1,
+			"Err PoolToLoanNftClass"
+		);
+		ensure!(
+			LoanNftClassToPool::<T>::iter_values().count() == 1,
+			"Err LoanNftClassToPool"
+		);
+		ensure!(
+			NextLoanId::<T>::iter_values().count() == 1,
+			"Err NextLoanId"
+		);
+		ensure!(Loan::<T>::iter_values().count() == 1, "Err Loan");
+		ensure!(
+			ActiveLoans::<T>::iter_values().count() == 1,
+			"Err ActiveLoans"
+		);
 		ensure!(
 			ClosedLoans::<T>::iter_values().count() == 1,
-			"There is not a closed loan"
+			"Err ClosedLoans"
+		);
+		ensure!(PoolNAV::<T>::iter_values().count() == 1, "Err PoolNAV");
+		ensure!(
+			PoolWriteOffGroups::<T>::iter_values().count() == 1,
+			"Err PoolWriteOffGroups "
 		);
 
 		Ok(Vec::new())
 	}
 
 	fn on_runtime_upgrade() -> Weight {
-		let loans_module = b"4c82a580ac33cceba8ed9766387f22b7";
-		let _ = storage::unhashed::clear_prefix(loans_module, None, None);
+		let _ = PoolToLoanNftClass::<T>::clear(1, None);
+		let _ = LoanNftClassToPool::<T>::clear(1, None);
+		let _ = NextLoanId::<T>::clear(1, None);
+		let _ = Loan::<T>::clear(1, None);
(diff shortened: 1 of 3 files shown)


runtime/altair/src/lib.rs
```diff
@@ -1497,11 +1497,8 @@ pub type Executive = frame_executive::Executive<
 	Runtime,
 	AllPalletsWithSystem,
 	(
-		pallet_multisig::migrations::v1::MigrateToV1<Runtime>,
-		pallet_preimage::migration::v1::Migration<Runtime>,
-		pallet_democracy::migrations::v1::Migration<Runtime>,
-		pallet_scheduler::migration::v3::MigrateToV4<Runtime>,
 		pallet_interest_accrual::migrations::v2::Migration<Runtime>,
+		pallet_loans::migrations::Migration<Runtime>,
```
lemunozm, line 1501: I've removed the other migrations because they were failing in the `pre_upgrade()` phase (checked one by one), which I understand is because they were already applied.

**PLEASE, DOUBLE-CHECK THIS**

mustermeiszer: pallet-interets-accrual should be removed to

runtime/altair/src/lib.rs
```diff
@@ -1497,11 +1497,8 @@ pub type Executive = frame_executive::Executive<
 	Runtime,
 	AllPalletsWithSystem,
 	(
-		pallet_multisig::migrations::v1::MigrateToV1<Runtime>,
-		pallet_preimage::migration::v1::Migration<Runtime>,
-		pallet_democracy::migrations::v1::Migration<Runtime>,
-		pallet_scheduler::migration::v3::MigrateToV4<Runtime>,
 		pallet_interest_accrual::migrations::v2::Migration<Runtime>,
```
mustermeiszer, line 1500: I would also remove this one. cc @branan

lemunozm: Seems not to be migrated yet, or at least it migrates succesfully.

runtime/altair/src/lib.rs
lemunozm: comment below

runtime/altair/src/lib.rs
mustermeiszer: mmh. Branan can clarify that I guess

runtime/altair/src/lib.rs
mustermeiszer: You are right, I guess. https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Faltair.api.onfinality.io%2Fpublic-ws#/chainstate

mustermeiszer: You can see the state of altair.

mustermeiszer: But this is weird. Why where all the others still included? Are they mentioned in the last release notes of Altair
?

lemunozm: I think those were included in the last release but not removed after it. If I leave them there, they complain that they were already migrated. I suppose we added the interest accrual to that list without cleaning the previous migrations.

lemunozm: What is weirdest is that the `interest-accrual` pallet seems empty with version 0. Is it ok? 🤔

pallets/loans/src/migrations.rs
```diff
@@ -0,0 +1,89 @@
...
+		let _ = LoanNftClassToPool::<T>::clear(1, None);
+		let _ = NextLoanId::<T>::clear(1, None);
+		let _ = Loan::<T>::clear(1, None);
+		let _ = ActiveLoans::<T>::clear(1, None);
+		let _ = ClosedLoans::<T>::clear(1, None);
+		let _ = PoolNAV::<T>::clear(1, None);
+		let _ = PoolWriteOffGroups::<T>::clear(1, None);
+
+		T::DbWeight::get().writes(8)
+	}
```
wischli, line 56: I was a bit baffled why the `try-runtime` logs did not show any migration for `Loans` so I added logs to the `pre` and `post` hooks as well as the migration itself which were executed as expected. 

So we could add some assurance log at least to `on_runtime_upgrade`, better to both hooks as well.

```rust
log::debug!("Loans storage clearing migration successful");
```

pallets/loans/src/migrations.rs
```diff
@@ -0,0 +1,89 @@
...
+		ensure!(
+			PoolWriteOffGroups::<T>::iter_values().count() == 1,
+			"Err PoolWriteOffGroups "
+		);
+
+		Ok(Vec::new())
+	}
+
+	fn on_runtime_upgrade() -> Weight {
+		let _ = PoolToLoanNftClass::<T>::clear(1, None);
```
wischli, line 46: Should: Execute only for specific storage version, see below:

```rust
if Pallet::<T>::on_chain_storage_version() == StorageVersion::new(0) {
       // snip
}
```

wischli (approved):
LGTM but I have some improvement recommendations.

runtime/altair/src/lib.rs
lemunozm: Regarding the versioning of this migration. This migration tries to remove the whole pallet, in order to remove it from the codebase, and use the new one. In that case, makes sense to version it? I mean, If I version this, I would store some version number on the pallet storage that I don't want, because I'll like to start from the scratch with the new pallet

pallets/loans/src/migrations.rs
lemunozm: Ok, I agree!

pallets/loans/src/migrations.rs
lemunozm: I understand this check, but not sure if it works for this "migration" that is not a migration 😆. (See comment above)

runtime/altair/src/lib.rs
lemunozm: (resolution about my concerns in the bottom)

branan: I think storage versioning is a separate concern from what we're doing here, which is purely a purge of old data.

runtime/altair/src/lib.rs
branan: It will always migrate successfully, I wrote it to not fail if it was left in. It should log a warning, though.

It can safely be removed.

branan:
I don't think there's any need to add a V1 key, since the storage is all under a separate prefix from the old pallet. The point of a version key is to ensure that we don't re-run a migration and corrupt data (and secondarily to make storage version information available to API consumers). Since the migration purely removes data under an old prefix, there's no risk of it ever being unsafe to run (it becomes a no-op if re-run). And API consumers won't be confused since it's a totally new pallet.

I agree that moving the migration to the new pallet is a better indication of what is happening :+1:

runtime/altair/src/lib.rs
branan: Apologies, I posted wrong info.

We've definitely shipped since we landed this code, but it must have been for a security issue and thus been a minimal release instead of from `main`. Altair is still on V1 not V2. We still need this migration.

branan:
I'll hold off on dropping my official github approval until the migration is moved to its new home and named clearly to indicate what it does. But I'm happy with the general shape of things

lemunozm:
Thanks for your throughs @branan, only a minor detail:

>  Since the migration purely removes data under an old prefix

I think the pallet prefix is the same because both pallets are registering the pallet with the `"Loans"` name in the runtime

runtime/altair/src/lib.rs
lemunozm: @branan looking at the [chain state](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Faltair.api.onfinality.io%2Fpublic-ws#/chainstate) of `InterestAccrual` I see that the storage is empty. Maybe we do not need the migration

pallets/loans/src/migrations.rs
wischli: You are right. I have to admit I was blanking out on the fact that `pallet-loans` co-exists with `pallet-loans-ref` and will be deprecated. So no need to add versioning!

wischli:
> I think the pallet prefix is the same because both pallets are registering the pallet with the `"Loans"` name in the runtime

Yes, this will lead to the same prefix.

pallets/loans/src/migrations.rs
lemunozm: Don't worry! Sometimes I'm also confused with this mess and I'm the person in change of this 🤣

lemunozm:
Closed in favor of #1266

PR #1290, opened 2023-03-28
Fix discount rate conversion

### LINKED ISSUE #1289: Loans: discount rate as secs
### Description

Currently, `discount_rate` is used as the user configures it (expecting a "per sec" rate). Because all other rates are passed "per year", it is confusing to the user to transform that rates by them depending on what parameter they are writting.

Convert the rate internally to "per sec" to avoid that to the user.

*Note: breaks API in the sense that it changes how a value is interpreted on the UI side*

lemunozm:
# Description

Fixes #1289

## Changes and Descriptions

- Removed unused test
- Discount rate is configured "per year" but handled "per sec" internally.

*Note: breaks API in the sense that it changes how a value is interpreted on the UI side*

### DIFF
--- a/runtime/integration-tests/src/utils/loans.rs
+++ b/runtime/integration-tests/src/utils/loans.rs
@@ -30,15 +30,6 @@ use crate::{chain::centrifuge::RuntimeCall, utils::tokens::rate_from_percent};
 
 type Asset = (CollectionId, ItemId);
 
-// TODO: Remove once #1189 is merged
-fn interest_rate_per_year_to_sec(rate_per_annum: Rate) -> Rate {
-	rate_per_annum
-		.ensure_div(Rate::saturating_from_integer(SECONDS_PER_YEAR))
-		.unwrap()
-		.ensure_add(Rate::one())
-		.unwrap()
-}
-
 /// Structure that manages collateral and loan nft ids
 pub struct NftManager {
 	collaterals: HashMap<PoolId, ItemId>,
@@ -123,7 +114,7 @@ pub fn issue_default_loan(
 	.valuation_method(ValuationMethod::DiscountedCashFlow(DiscountedCashFlow {
 		probability_of_default: rate_from_percent(5),
 		loss_given_default: rate_from_percent(50),
-		discount_rate: interest_rate_per_year_to_sec(rate_from_percent(4)),
+		discount_rate: rate_from_percent(4),
 	}));
 
 	issue_loan(owner, pool_id, loan_info, manager)


pallets/loans-ref/src/valuation.rs
```diff
@@ -103,7 +113,7 @@ where
 {
 	pub fn is_valid(&self) -> bool {
 		match self {
-			ValuationMethod::DiscountedCashFlow(dcf) => dcf.discount_rate >= One::one(),
+			ValuationMethod::DiscountedCashFlow(_dcf) => true,
 			ValuationMethod::OutstandingDebt => true,
 		}
 	}
```
lemunozm, line 119: No longer need this check here (always true). Not sure if removing this

pallets/loans-ref/src/valuation.rs
```diff
@@ -76,11 +81,16 @@ impl<Rate: FixedPointNumber> DiscountedCashFlow<Rate> {
...
+		// once #1189 is merged
 		let acc_rate = checked_pow(interest_rate_per_sec, exp).ok_or(ArithmeticError::Overflow)?;
 		let ecf = acc_rate.ensure_mul_int(debt)?;
 		let ra_ecf = tel_inv.ensure_mul_int(ecf)?;
 
 		// Discount the risk-adjusted expected cash flows
+		// TODO: use InterestAccrual for this once #1189 is merged
+		// This would immply that discount_rate should be register/unregister.
 		let rate = checked_pow(self.discount_rate, exp).ok_or(ArithmeticError::Overflow)?;
 		let d = Rate::one().ensure_div(rate)?;
```
lemunozm, line 95: I think we can get a performance boost here if using interest-accrual for this

pallets/loans-ref/src/valuation.rs
mustermeiszer: I would keep. EVen the check. Checks that interest-accrual is not doing stuff it shouldn't after we use it, right?

mustermeiszer (dismissed):

pallets/loans-ref/src/valuation.rs
lemunozm: The check `discount_rate >= 1` needs to be removed since now we expect discount rates less than 1.

Regarding the `interest-accrual`, by now, it's not used with this rate. But when used, the own `interest-accrual` should anyways check its own values 🤷🏻‍♂️

lemunozm: This comment no longer apply because now we have an upper limit

mustermeiszer (dismissed):

PR #1292, opened 2023-03-28
feat: Connectors v2 message PoC

wischli:
# Description

Applies ConnectorsV2 message format [based on the spec](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A#Message-format)

## TODO
- [x] Discuss requirement of  `CurrencyId` to be representable by `u128`
- [x] Fix unit tests
- [x] Add mapping from `CurrencyId` to `u128` via the new `GeneralCurrencyIndex` trait
- [ ] Verify encoding/decoding updates in Solidity

## Changes and Descriptions

* Updates connectors `Message` enum [based on the spec](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A#Message-format)
  * Updates existing messages
  * Adds new messages
* Updates existing Centrifuge chain extrinsics
* Adds conversion for `CurrencyId` enum to `u128` via two `GeneralCurrencyIndex` impls
  * [First one](https://github.com/centrifuge/centrifuge-chain/pull/1292/files#diff-25fb8bbfd5cba07f4d02671436bf07aaf97c30403683f0b4116545c108a61097R83-R104) for `CurrencyId`
  * [Second one](https://github.com/centrifuge/centrifuge-chain/pull/1292/files#diff-af2bb0a52ae16d2600b291d87372520930880c0daac72256f235d72cd27363e9R668-R686) for the connectors `Pallet` 
* Does not add flow/extrinsics for new messages, will be done in a subsequent PR

# Checklist:


pallets/connectors/src/message.rs
```diff
@@ -56,13 +60,72 @@ where
...
+	CollectInvest {
+		pool_id: PoolId,
+		tranche_id: TrancheId,
+		address: Address,
+	},
+	CollectForInvest {
+		pool_id: PoolId,
+		tranche_id: TrancheId,
+		call_address: Address,
+		collect_address: Address,
```
NunoAlexandre, line 127: FYI, on the Solidity I am naming these `caller` and `user`, respectively. Same for the other messages. Happy to change but it would be good to have both ends aligned

pallets/connectors/src/message.rs
```diff
@@ -81,12 +144,21 @@ impl<
 	/// in other domains and MUST follow the defined standard.
 	fn call_type(&self) -> u8 {
 		match self {
-			Self::Invalid => 0,
+			Self::Invalid { .. } => 0,
```
NunoAlexandre, line 147: Out of curiosity, why is this change needed? 🤔

NunoAlexandre:
> * [ ]  Discuss requirement of  `CurrencyId` to be representable by `u128`

I have been thinking about this. The spec sort of presumes that every currency we will be identifying as a `u128` has an inner `GeneralIndex` value within its `MultiLocation` representation. But that's not always the case, neither can/should we rely on that.

I'd suggest that we have a storage mapping from `CurrencyId -> u128` that _we_ set for every `CurrencyId` we want to support for these transfers. This grants us some benefits:

- we can support new currencies to be transferred through Connectors without a runtime upgrade 
- we don't have a setup where EVERY currency registered in the asset registry is automatically connector-transferrable
- other dAPPs or our own can query this state to know how this mapping is taking place
- The community can have a saying on what tokens they find favourable to be connector-transferrable or not

On the Solidity side we will also need a similar mapping but from `uint128 -> address`.

What do you think? cc @offerijns @mustermeiszer

pallets/connectors/src/message.rs
wischli: It was a result of me deleting the entire match and using autofill to apply the changes 😅 I will revert it as it is not needed.

pallets/connectors/src/message.rs
```diff
@@ -220,6 +490,20 @@ fn encoded_message(call_type: u8, fields: Vec<Vec<u8>>) -> Vec<u8> {
...
+	account: &AccountId,
+) -> Result<[u8; 32], sp_runtime::DispatchError>
+where
+	AccountId: Encode,
+{
+	let account_vec = account.encode();
+	frame_support::ensure!(account_vec.len() == 32, "AccountId must be 32 bytes.");
+	let mut bytes = [0u8; 32];
+	bytes.copy_from_slice(&account_vec);
+	Ok(bytes)
```
lemunozm, line 504: I think `account.encode().try_into()` do this

NunoAlexandre: Yeah if we convert our `AccountId` (32 bytes long) to `[u8; 32]` that should do ☝️ . You can also always do `vec_to_fixed_array(account.encode())`

libs/types/src/tokens.rs
```diff
@@ -59,6 +60,39 @@ impl From<u32> for CurrencyId {
...
+				let mut r = [0u8; 16];
+				r[15] = 1;
+				Ok(u128::from_be_bytes(r))
+			}
+			// [0, ..., 0, 2]
+			CurrencyId::AUSD => {
+				let mut r = [0u8; 16];
+				r[15] = 2;
+				Ok(u128::from_be_bytes(r))
+			}
```
NunoAlexandre, line 83: `KSM` and `AUSD` will be dropped so in this conversion we can already ignore those two. And I still think that having a storage for this would grant us more granular control over what tokens we allow to be Connector-transferrable and to make (reverse-)maping more obvious than an hardcoded runtime feature IMHO (I can be wrong in this last one and happy to change my mind)

NunoAlexandre: If we do keep this logic, we need testing covering this 👀

pallets/connectors/src/lib.rs
```diff
@@ -316,6 +327,8 @@ pub mod pallet {
 		pub fn add_pool(
 			origin: OriginFor<T>,
 			pool_id: PoolIdOf<T>,
+			currency: CurrencyIdOf<T>,
+			decimals: u8,
```
NunoAlexandre, line 331: We are dropping `decimals` from `AddPool`

pallets/connectors/src/lib.rs
```diff
@@ -387,10 +411,11 @@ pub mod pallet {
 
 			Self::do_send_message(
 				who,
-				Message::UpdateTokenPrice {
+				Message::UpdateTrancheTokenPrice {
 					pool_id,
 					tranche_id,
 					price,
+					decimals,
```
NunoAlexandre, line 418: Also dropping `decimals` here

pallets/connectors/src/message.rs
```diff
@@ -56,13 +60,72 @@ where
...
+		tranche_id: TrancheId,
+		source_address: Address,
+		destination_address: Address,
+		amount: Balance,
+	},
+	IncreaseInvestOrder {
+		pool_id: PoolId,
+		tranche_id: TrancheId,
+		address: Address,
+		token: u128,
```
NunoAlexandre, line 83: In `AddPool` we are calling it `currency` while here we call it `token`. I don't have a strong preference for either but let's agree on one of the two and make that consistent both here and on the Solidity side cc @offerijns @AStox

pallets/connectors/src/message.rs
```diff
@@ -233,6 +517,13 @@ mod tests {
 
 	pub type ConnectorMessage = Message<Domain, PoolId, TrancheId, Balance, Rate>;
 
+	const AMOUNT: Balance = 100000000000000000000000000;
+	const POOL_ID: PoolId = 12378532;
+	const TOKEN_ID: u128 = 246803579;
+	const TRANCHE_HEX: &str = "811acd5b3f17c06841c7e41e9e04cb1b";
```
NunoAlexandre, line 523: I'd rather have `TRANCHE_ID: TrancheId = tranche_id_from_hex("811acd5b3f17c06841c7e41e9e04cb1b");` so that we don't have to call `tranche_id_from_hex` every time

NunoAlexandre (commented):
First round 🚀 

Things will still change a bit until the messages v2 spec is finalised so hang in there 🙏

pallets/connectors/src/message.rs
AStox: `currency` is widely used in tinlake contracts to refer to stable coins, so I think we could try to keep that naming convention here as well.

pallets/connectors/src/message.rs
wischli: That was my initial idea. Unfortunately that does not work as  calls in constants are limited to constant functions, tuple structs and tuple variants. Since `from_hex` is not a constant function, I feel like our hands are tied.

Another option could be to have default `get` functions instead of constants. WDYT? E.g.

```
	fn default_address20(hex: &str) -> [u8; 20] {
		<[u8; 20]>::from_hex("1231231231231231231231231231231231231231").expect("Should be valid 20 bytes")
	}
```

NunoAlexandre: @wischli that's why I din't include the `const` bit in my suggestion above; but your suggestion is also good :)

libs/types/src/tokens.rs
```diff
@@ -79,6 +80,29 @@ impl From<StakingCurrency> for CurrencyId {
...
 }
 
+impl<Prefix> GeneralCurrencyIndex<Prefix> for CurrencyId
+where
+	Prefix: Get<[u8; 12]>,
+{
+	type CurrencyId = CurrencyId;
+	type GeneralIndex = u128;
+
+	fn get_general_index(
```
NunoAlexandre, line 90: I would implement this as a `Into` / `From` `GeneralCurrencyIndex` to/from `CurrencyId` instead.

libs/types/src/tokens.rs
```diff
@@ -150,4 +174,7 @@ pub struct CustomMetadata {
 
 	/// Whether an asset can be used as a currency to fund Centrifuge Pools.
 	pub pool_currency: bool,
+	// TODO: Enable in follow-up PR
+	// /// The corresponding 20-byte EVM address of the asset.
+	// pub evm_address: [u8; 20],
```
NunoAlexandre, line 179: This won't work because this would assume that every token would have to have an `evm_address`. We could either make it optional, which is still a weak abstraction in my opinion, or we can use the `MultiLocation` field of the metadata to represent the EVM address of that token, assuming we are registering the asset as it being an asset in a Connector-connected Domain, if that makes sense? By this last part I mean, lets's say that Domain X has a Token `X` with an EVM address `EA`; if we register `X` on the `Centrifuge Chain` just to register/list a token we handle on Domain `X`, that approach could work, because when sending/receiving transfers from/to that Domain, we can expect the respective entry in the asset registry to have a MultiLocation holding an EVM address; if it doesn't, we throw an error.

Maybe I am making this sound too complicated, but I am coming from a perspective where every token listed in the asset registry (apart from Tranche tokens) is a potential token that could be transferred through XCM; by using `MultiLocation` to hold an `EVM` address that wouldn't fit _that_ purpose, but since that's not the purpose we would be registering the token in the registry in the first place, I think that's fine.

WDYT?

pallets/connectors/src/lib.rs
```diff
@@ -427,13 +453,89 @@ pub mod pallet {
...
+						.map_err(|_| DispatchError::Other("Conversion to 32 bytes failed"))?,
+					receiver: domain_address.address(),
+				},
+				domain_address.domain(),
+			)?;
+
+			Ok(())
+		}
+
+		/// Add a CurrencyId to the set of known currencies, enabling that currency for Pool creation.
```
NunoAlexandre, line 518:
```suggestion
		/// Add a CurrencyId to the set of known currencies on a given Domain
```

The "enabling that currency for Pool creation" is not necessarily correct for two reasons:
1. One can't create a pool through a Connector Domain, so adding a currency to a Domain doesn't enable that currency for pool creation
2. It's through `AllowPoolCurrency` that we enable a currency to be used as a pool currency and for investments etc

pallets/connectors/src/message.rs
```diff
@@ -56,13 +65,72 @@ where
 		address: Address,
 		valid_until: Moment,
 	},
+	// Bidirectional: Domain must not accept every incoming token.
+	// Sender must ensure beforehand that the receiver will not reject
```
NunoAlexandre, line 69: > // Sender must ensure beforehand that the receiver will not reject

This is not technically possible; As in, `sender` can't check with the receiver that the receiver will not reject handling a given token. I think the point of that sentence is to say that Sender can't ensure that receiver will handle it, so sender should make sure to only send tokens it knows that the receiver can handle. Since transfers are only to happen between Centrifuge and an EVM domain for now, this means that Centrifuge should make sure to have called `AddCurrency(C, _)` before `Transfer(C, _)`; The Transfer from a Domain down to Centrifuge should always work because if Domain as able to send `C`, it was by definition added already on that domain.

runtime/integration-tests/src/xcm/development/tests/connectors.rs
```diff
@@ -256,7 +260,7 @@ fn update_token_price() {
 }
 
 #[test]
-fn transfer() {
```
NunoAlexandre, line 259: Can we add tests for all the new messages? 🙏

NunoAlexandre (commented):
First round 🏄‍♂️ 

Looking good overall! Left some things to think about and to adjust, will continue the review after lunch.

We need to extend the encoding/decoding tests to cover the new messages btw 👍

libs/types/src/tokens.rs
wischli: I definitely get your point! This is purely based on the [last ConnectorsV2 sync](https://centrifuge.hackmd.io/ODJ-dNHUSsWcYNh6dHJtcg?both#Outcome) we had were we decided to add `evm_address` to the metadata of currencies in the asset registry. I am not sure though how to deterministically derive 32 bytes from the flexible `MultiLocation` which could be quite scaffolded, e.g. `Here` vs. `X2(para, pallet_instance, general_index)` etc. How would you handle that?

pallets/connectors/src/message.rs
wischli: Thanks for the explanation, makes absolute sense! This requirement was copy pasted [from the spec](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#Transfer) btw.

runtime/integration-tests/src/xcm/development/tests/connectors.rs
wischli: Of course, sorry for having missed out on these.

pallets/connectors/src/message.rs
NunoAlexandre: Yeah I noticed the source of that line :) my pleasure!

libs/types/src/tokens.rs
NunoAlexandre: We could use a `X2(DomainId, AccountKey20 { network: NetworkId, key: [u8; 20] }}`; but that's probably trying to make a shoe out of a shoe box. I remember @offerijns suggesting in that sync to maybe use a type like `DomainAddress`; the issue with that is that we already have the `location` field on the `Metadata` type. 

As I am thinking and writing about this, it becomes clearer how there's an abstraction leakage here, in the sense of trying to use a design / abstraction (the asset registry) for something it's not quite designed for. The best would be if we could change the `Meadata.location` field to be something like `enum GlobalLocation { Polkadot(VersionedMultiLocation), EVM(chainId, address) }` but we can't do that unless if we change the orml pallet itself.

I have to think a bit more about this 🙃

pallets/connectors/src/lib.rs
```diff
@@ -470,7 +470,10 @@ pub mod pallet {
 					amount,
 					domain: domain_address.domain(),
 					destination_address: domain_address.address(),
-					source_address: account_to_bytes(&who)?,
+					source_address: who
+						.encode()
+						.try_into()
+						.map_err(|_| DispatchError::Other("Conversion to 32 bytes failed"))?,
```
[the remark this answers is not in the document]
wischli: Unfortunately not.

libs/types/src/tokens.rs
wischli: Could you give a short feedback if [c064dca](https://github.com/centrifuge/centrifuge-chain/pull/1292/commits/c064dca5a9c793210246fcc47e65b63636f22b63) matches your expectations? Then I will add the rest.

runtime/integration-tests/src/xcm/development/tests/connectors.rs
wischli: AFAICS, basically only `collect_redeem` was missing. Added a few more `zero` cases. [7f60bd4](https://github.com/centrifuge/centrifuge-chain/pull/1292/commits/7f60bd4cc51be66013ec756b264762fadad83119)

libs/types/src/tokens.rs
offerijns: Why not use XCM MultiLocations for this, weren't they explicitly designed to support multiple consensus systems, so also able to represent EVM addresses? We can/should look at how Snowbridge intents to do this.

NunoAlexandre (approved):
Round two 🏄‍♂️ 

Some changes in `TransferTrancheTokens` and few other small things but we are getting there!

libs/primitives/src/lib.rs
```diff
@@ -249,6 +249,9 @@ pub mod constants {
 	pub const fn deposit(items: u32, bytes: u32) -> Balance {
 		items as Balance * 15 * CENTI_CFG + (bytes as Balance) * 6 * CENTI_CFG
 	}
+
+	/// The prefix for tokens managed by Connectors.
+	pub const GENERAL_CURRENCY_INDEX_PREFIX: [u8; 12] = *b"CfgCnctTknId";
```
NunoAlexandre, line 254:
```suggestion
	pub const GENERAL_CURRENCY_INDEX_PREFIX: [u8; 12] = *b"CfgCnctCurId";
```

Nit but since we call it `_CURRENCY_` I'd rather have the prefix value itself also be "Cur"rrency

pallets/connectors/src/lib.rs
```diff
@@ -550,6 +642,26 @@ pub mod pallet {
...
+
+		/// Returns the `u128` general index of a currency as the concatenation of the
+		/// configured `GeneralCurrencyPrefix` and its local currency identifier.
+		///
+		/// Assumes the currency to be registered in the `AssetRegistry`.
+		pub fn try_get_general_index(
+			currency: <T as pallet::Config>::CurrencyId,
+		) -> Result<u128, DispatchError> {
+			ensure!(
+				<T as Config>::AssetRegistry::metadata(&currency).is_some(),
```
NunoAlexandre, line 654: I think there should be a way of calling `contains_key` instead of `metadata(_).is_some()` but this is ok 👍

pallets/connectors/src/message.rs
```diff
@@ -35,34 +35,93 @@ where
...
 	},
 	UpdateMember {
 		pool_id: PoolId,
 		tranche_id: TrancheId,
-		address: Address,
+		member: Address,
 		valid_until: Moment,
 	},
+	// Bidirectional: Domain must not accept every incoming token.
+	// Sender cannot ensure whether the receiver rejects.
```
NunoAlexandre, line 69: I feel like this comment belongs at the header of the `Message` type definition to introduce concepts that will be mentioned below; then for each message we can document each message regarding its directionality and other properties. wdyt?

NunoAlexandre (commented):
Nice 👌 

A few small comments but it's 99.75% ready 🚀

pallets/connectors/src/lib.rs
```diff
@@ -427,13 +449,83 @@ pub mod pallet {
...
-				Message::Transfer {
+				who.clone(),
+				Message::TransferTrancheTokens {
 					pool_id,
 					tranche_id,
 					amount,
 					domain: domain_address.domain(),
-					address: domain_address.address(),
+					receiver: domain_address.address(),
+					sender: who
```
offerijns, line 459: I realize the order in this struct here doesn't matter technically, but I think for easy reading it would be nice to be consistent and have `sender` before `receiver` (and generally, just make sure the order of parameters matches that of the actual message parameter ordering)

pallets/connectors/src/message.rs
```diff
@@ -35,34 +35,93 @@ where
...
+		currency: u128,
+		evm_address: [u8; 20],
+	},
 	AddPool {
 		pool_id: PoolId,
 	},
+	AllowPoolCurrency {
+		currency: u128,
+		pool_id: PoolId,
+	},
```
offerijns, line 48: In the solidity implementation as well as in the spec, the order of these parameters was reversed. See discussion here: https://github.com/centrifuge/connectors/pull/59#discussion_r1174540624

NunoAlexandre: The spec is updated ☝️

pallets/connectors/src/lib.rs
wischli: We restricted `AssetRegistry` to the `Inspect` trait which [only exposes](https://github.com/open-web3-stack/open-runtime-module-library/blob/16b6c1149a15674d21c87244b7988a667e2c14d9/traits/src/asset_registry.rs#L42) `fn metadata` for our needs.

pallets/connectors/src/message.rs
wischli: I hope my attempt in [4ecddf9](https://github.com/centrifuge/centrifuge-chain/pull/1292/commits/4ecddf9d7515949ea43e8ad65375bfd1174d4ec4) aligns with your vision.

pallets/connectors/src/message.rs
```diff
@@ -35,34 +38,161 @@ where
...
+	/// Reduce the invest order amount for the specified pair of pool and
+	/// tranche token.
+	///
+	/// On success, triggers a message sent back to the sending domain.
+	/// The message will take care of re-funding the investor with the given
+	/// amount the order was reduced with. The `investor` address is used as
+	/// the receiver of that tokens.
+	///
+	/// Outgoing: Directional from Centrifuge to EVM Domain.
```
offerijns, line 133:
```suggestion
	/// Incoming: Directional from EVM Domain to Centrifuge. 
```

pallets/connectors/src/lib.rs
```diff
@@ -196,6 +214,8 @@ pub mod pallet {
 
 	#[pallet::error]
 	pub enum Error<T> {
+		/// Failed to map the asset to its u128 representation
```
NunoAlexandre, line 217:
```suggestion
		/// Failed to map the asset to its Connector's General Index representation
```

I guess we would be better off being more specific?

runtime/development/src/lib.rs
```diff
@@ -1492,10 +1492,17 @@ impl pallet_interest_accrual::Config for Runtime {
 	type Weights = ();
 }
 
+parameter_types! {
+	#[derive(scale_info::TypeInfo)]
+	pub const GeneralCurrencyPrefix: [u8; 12] = GENERAL_CURRENCY_INDEX_PREFIX;
+}
+
```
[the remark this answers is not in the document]
wischli: Since this is a constant, I feel like it's better suited in the current place, e.g the `constants` module of `libs/primitives/lib.rs`. WDYT @mustermeiszer?

pallets/connectors/src/message.rs
```diff
@@ -35,34 +38,161 @@ where
 	Rate: Encode + Decode,
 {
 	Invalid,
+	/// Add a currency to a domain, i.e, register the mapping of a currency id to
+	/// the corresponding EVM Address.
+	///
+	/// Outgoing: Directional from Centrifuge to EVM Domain.
```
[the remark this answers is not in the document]
wischli: I was also preferring arrow-style but not sure.

NunoAlexandre (approved):
🚀

pallets/connectors/src/lib.rs
```diff
@@ -427,13 +450,83 @@ pub mod pallet {
...
+			)?;
+
+			Ok(())
+		}
+
+		/// Add a `CurrencyId` to the set of known currencies on a given Domain.
+		// TODO: Replace weight after benchmarking
+		#[pallet::weight(< T as Config >::WeightInfo::add_connector())]
+		#[pallet::call_index(8)]
+		pub fn add_currency(
```
mustermeiszer, line 517: Do we have an issue for tracking that. This would be really important!

mustermeiszer: I mean the FIXME

runtime/development/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Agreeing with William here. All constants should stay in the primitives or the respective const of libs

mustermeiszer (changes requested):
Looks good already.

We should note that we can remove the hard coupling with `pallet_xcm_transactor` once the pallet-connectors-gateway takes care of routing messages.

Thinks we can remove then
* hard coupling
* `fn handle`
* `fn add_connector`
* `fn set_domain_router`

Few changes I think are necessary:
* Move `impl TypeId for DomainAddress` into `id` module of crate types
* Move `trait Codec` to traits
* `fn update_member`
    *The adding of members should always go through the pallet-permissions interface. As we also coded the member list bot to use this API. So the extrinsic must only check if the  permissions is existing. If yes, then submit message. Furthermore, we must block if the domain is local -- i.e. Centrifuge*
* `fn transfer_tranche_tokens` & `fn transfer` 
   &rarr; *for both of these messages to the local domain should be forbidden. I think it is debatable if the gateway logic @cdamian will implement will ever allow to send messages to our own domain. I would argue no, but this is debatable. SO we should either agree on blocking this on the gateway side or already on the connectors side. WDYT?*
   

Missing: 
* `fn allow_pool_currency` 
   &rarr; *This is needed. The restriction is currently not 100% clear. I would say we restrict calling this only from democracy/AdminOrigin for now. Maybe we find a way to make it clear in the asset-registry or in the pallet-foreign-investments whether another currency can be used for investing.*

pallets/connectors/src/lib.rs
wischli: We do not have an existing Github issue but this is part of the asset metadata update which is owned by @NunoAlexandre

mustermeiszer (dismissed):
Approval under the constraint that we do a follow-up PR. This is not production code so we can go with this temporary inconsistency in main.

pallets/connectors/src/lib.rs
wischli: Improved message in [6b06a9d](https://github.com/centrifuge/centrifuge-chain/pull/1292/commits/6b06a9dcc189f61db171ca44d005c68bb8e04153)

pallets/connectors/src/lib.rs
```diff
@@ -427,13 +450,83 @@ pub mod pallet {
...
+
+		/// Transfer non-tranche tokens to a given address
+		#[pallet::weight(< T as Config >::WeightInfo::transfer())]
+		#[pallet::call_index(7)]
+		pub fn transfer(
+			origin: OriginFor<T>,
+			asset_id: CurrencyIdOf<T>,
+			domain_address: DomainAddress,
+			amount: <T as pallet::Config>::Balance,
+		) -> DispatchResult {
```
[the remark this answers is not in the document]
wischli: Great catch! [05dd755](https://github.com/centrifuge/centrifuge-chain/pull/1292/commits/05dd755d6ca39069b4afbbcdce57477a4225efd3)

libs/primitives/src/lib.rs
```diff
@@ -249,6 +249,9 @@ pub mod constants {
 	pub const fn deposit(items: u32, bytes: u32) -> Balance {
 		items as Balance * 15 * CENTI_CFG + (bytes as Balance) * 6 * CENTI_CFG
 	}
+
+	/// The prefix for tokens managed by Connectors.
+	pub const GENERAL_CURRENCY_INDEX_PREFIX: [u8; 12] = *b"CfgCnctCurId";
```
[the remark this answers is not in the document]
wischli: It might be overkill but I am always supporting safer approaches. Unfortunately we lose out on the `const` property due to using `blake2_128` and splitting. WDYT? [d774926](https://github.com/centrifuge/centrifuge-chain/pull/1292/commits/d774926f35f09d6d1653379f025b26f257199850)

mustermeiszer (approved):
LGTM! Thanks for adapting to my comments.

PR #1298, opened 2023-03-31
Pallet loans: fix version after nuking it

lemunozm:
# Description

The migration removes the whole state of `pallet-loans`. This means it also removes the version we previously set to `0`. This PR fix it.

### DIFF
--- a/pallets/loans-ref/src/migrations.rs
+++ b/pallets/loans-ref/src/migrations.rs
@@ -1,6 +1,7 @@
 #[cfg(feature = "try-runtime")]
 use frame_support::ensure;
 use frame_support::{
+	dispatch::GetStorageVersion,
 	pallet_prelude::ValueQuery,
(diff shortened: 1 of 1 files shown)


wischli (commented):
We don't need to re-initiate `StorageVersion::new(0)` as querying the `StorageVersion` will default to `0` which is fine in this case. We would have to merge this PR if the current version was not the default value of zero.

You can verify yourself by adding these two checks to the `post_upgrade` hook of `NukeMigration`:

```rust
ensure!(
	Pallet::<T>::current_storage_version() == StorageVersion::new(0),
	"Current storage version is off, not zero"
);
ensure!(
	Pallet::<T>::on_chain_storage_version() == StorageVersion::new(0),
	"On chain storage version is off, not zero"
);
```

I just ran the migration with these two checks and it completed successfully.

PR #1300, opened 2023-04-03
Add benchmark CI job for fast checks

### LINKED ISSUE #1268: Add CI job for benchmarks

* CI Feature

### Description

Run tests benchmarks and run runtimes benchmarks are very different. We need to be sure the runtime benchmarks are not broken when we make a change.

Add a CI job that runs all runtime benchmarks with one iteration.

- Reliability of `main` code

- This CI job can be very long. We need to keep as simple as possible.

lemunozm:
# Description

This job does not generate weights, it only checks fastly for all runtimes if benchmarks work/compile.

You can test the script locally as follows for fast benchmark checks (it takes around 10-15s to check all runtime benchmarks once it's compiled):

```sh
./scripts/check_benchmarks.sh <runtime>
```

Being `<runtime>` the following possible values: `development`, `altair`, `centrifuge`

Fixes #1268

## Changes and Descriptions

- New script
- New CI job

### DIFF
--- a/.github/workflows/benchmark-check.yml
+++ b/.github/workflows/benchmark-check.yml
@@ -1,0 +1,40 @@
+on:
+  push:
+    branches: [main, 'release-v**']
+  pull_request:
+name: Check Benchmarks
+jobs:
+  tests:
+    name: ${{ matrix.target }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        target: [benchmark-test]
+        runtime: [development, altair, centrifuge]
+    runs-on: ${{ matrix.os }}
+    env:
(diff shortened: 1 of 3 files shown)


mustermeiszer (approved):
Looks good to me! But given we change CI I would also wait for @gpmayorga approval

NunoAlexandre (approved):
LGTM 🌞

lemunozm:
This PR fails as expected due to the `pool-registry` benchmark issue solved in #1233

Once that PR is merged, this can be safety merged.

lemunozm:
I think we can run better the three runtimes in the same job to save 1 hour per CI job

mustermeiszer (approved):
LGTM!

Issue #1304, opened 2023-04-04
Pallet-Interest-Accrual: Fix "past" timestamp usage for rates calculation


mustermeiszer:
### Description
We have the problem that the `on_initialize` logic in the `pallet-interest-accrual` uses the timestamp of `pallet_timestamp::Now`, which unfortunately is the timestamp of the last block and not the current one during initilization.

For reference see:
Block execution: https://github.com/paritytech/substrate/blob/master/frame/executive/src/lib.rs#L473-L490
Pallet-timestamp hooks: https://github.com/paritytech/substrate/blob/master/frame/timestamp/src/lib.rs#L168-L178
Pallet-timestamp inherent (an extrinsic) that updates the timestamp: https://github.com/paritytech/substrate/blob/master/frame/timestamp/src/lib.rs#L202-L217

#### Solution
As a solution we choose to let `pallet-interest-accrual` generate its own inherent. This inherent MUST run AFTER the `pallet_timestamp::set` inherent, so that the timestamp is up-to-date when we update the rates.

* define `pallet-interest-accrual::update_rates` inherent 
  * same logic as in current `on_initialize()`
   * checks to ensure it is ran once per block (it can be ran multiple times, as there is no external input, but MUST be included once)
     &rarr; easiest would be to do this in `on_finalize`
   * check to ensure it is ran after pallet-timestamp &rarr; easiest would be to store `LastTimestamp` in `on_finalize` and compare the timestamp used in the `update_rates` inherent with this value (it must strictly increase)

### How will this affect the code base
* fix bug

### What are foreseen obstacles or hurdles to overcome?
* Ordering of the inherents must be ensures -- `pallet_timestamp::Set` before `pallet_interest_accrual::UpdateRates`

### Accompanied Tasks
- [ ] re-revert #1303 (i.e. this brings back #1299)

PR #1299, opened 2023-04-03
InterestAccrual: Unify current & previous debt methods

lemunozm:
# Description

Fixes #1203 

## Changes and Descriptions

- `current_debt` and `previous_debt` merged into `calculate_debt`.

PR #1306, opened 2023-04-05
Add again #1299 with fix

lemunozm:
# Description

Adds again #1299 with a fast fix to avoid this issue: #1304

The fix consists in removing the `NotInPast` error completely, given the user the most updated debt in case `when` is one block in the "future".

**NOTE: The correct fix should be #1304**

### DIFF
--- a/pallets/interest-accrual/src/lib.rs
+++ b/pallets/interest-accrual/src/lib.rs
@@ -135,7 +135,7 @@ use sp_runtime::{
 	traits::{AtLeast32BitUnsigned, CheckedAdd, CheckedSub, Saturating},
 	ArithmeticError, DispatchError, FixedPointNumber, FixedPointOperand,
 };
-use sp_std::vec::Vec;
+use sp_std::{cmp::Ordering, vec::Vec};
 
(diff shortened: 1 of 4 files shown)


pallets/interest-accrual/src/lib.rs
```diff
@@ -335,30 +333,38 @@ pub mod pallet {
...
+					let delta = now.ensure_sub(when)?;
+					let rate_adjustment =
+						checked_pow(rate.interest_rate_per_sec, delta.ensure_into()?)
+							.ok_or(ArithmeticError::Overflow)?;
+					rate.accumulated_rate.ensure_div(rate_adjustment)?
+				}
+				Ordering::Greater => {
+					// TODO: This is a fast fix, the correct solution should be #1304
+					rate.accumulated_rate
+				}
```
lemunozm, line 364: Here it is

lemunozm:
The fix over the original commit is a742bab5f1a5a4844a5258305db9a5e53bdcc67b

offerijns (approved):

PR #1314, opened 2023-04-11
Loans: Add multiple triggers for write-off

### LINKED ISSUE #1305: Loans: Write off policy with different triggers
### Description

We can have different reasons for writing off a loan, not only overdue days. We need to model that:

From this [Slack conversation](https://kflabs.slack.com/archives/C050E3ZN2JU/p1680271126848299?thread_ts=1680107878.775689&cid=C050E3ZN2JU):

```rust
pub enum WriteOffTrigger {
    /// Number in days after the maturity date has passed
    PrincipalOverdueDays(u32),
    
    /// Seconds since the oracle valuation was last updated
    OracleValuationOutdated(Moment),
    
    /// Number in interest payments which are overdue
    NumberOfInterestPaymentsOverdue(u32),
}

pub struct WriteOffState<Rate> {
    /// If any of the triggers is valid, the write-off state can be applied
    pub triggers: Vec<WriteOffTrigger>,
    
    /// Percentage of present value we are going to write off on a loan
    pub percentage: Rate,
    
    /// Additional interest that accrues on the written off loan as penalty
    pub penalty: Rate,
}
```

lemunozm:
# Description

Fixes #1305 

## Tasks
- [x] Changes
- [x] migration

## Changes
- Refactor `WriteOffState::overdue_days` to `WriteOffState::triggers` of type `WriteOffTrigger` enum.
- Refactor `WriteOffState::{penalty, percentage} removed and used instead of the `WriteOffStatus` that already contains that information.
- Renamed `WriteOffState` to `WriteOffRule`.
- Find the best rule now is based of `percentage` and `penalty` instead of `overdue_days`


lemunozm:
Initial commit from Jeroen at: https://github.com/centrifuge/centrifuge-chain/commit/bf77eb6883ca764b8726eff179048e15b8999a0f

pallets/loans-ref/src/lib.rs
```diff
@@ -532,10 +533,11 @@ pub mod pallet {
...
 
 			let _count = Self::update_active_loan(pool_id, loan_id, |loan| {
-				let state = Self::find_write_off_state(pool_id, loan.maturity_date());
-				let limit = state.map(|s| s.status()).unwrap_or_else(|_| status.clone());
+				let rule = Self::find_write_off_rule(pool_id, loan)?;
+				let limit = rule.map(|r| r.status).unwrap_or_else(|| status.clone());
 
-				loan.write_off(&limit, &status)
+				loan.write_off(&limit, &status)?;
+				Ok(limit)
```
lemunozm, line 540: This is a "bug" fix. The previous `admin_writen_off` event was not registering the correct status applied. No security issue here

pallets/loans-ref/src/lib.rs
```diff
@@ -679,16 +681,38 @@ pub mod pallet {
...
+						.map(|trigger| loan.check_write_off_trigger(&trigger))
+						.find(|e| match e {
+							Ok(value) => *value,
+							Err(_) => true,
+						})
+						.map(|result| result.map(|_| rule))
+				})
+				.collect::<Result<Vec<_>, _>>()? // This exits if error before getting the maximum
+				.into_iter()
+				.max_by(|r1, r2| r1.status.cmp(&r2.status)))
```
lemunozm, line 715: Maybe it's not trivial to understand fastly. This is mostly because `Iter::try_find()` is unstable, and the workaround is easy to see. With `try_find` this should be 3-4 lines 😢 

Basically, we want to accept a "trigger" that has been true or with an error, to be able to catch that error successfully later and not avoid it without noise.

pallets/loans-ref/src/types.rs
```diff
@@ -129,65 +131,63 @@ pub enum PortfolioValuationUpdateType {
...
-		WriteOffStatus {
-			percentage: self.percentage,
-			penalty: self.penalty,
-		}
-	}
+/// The data structure for storing a specific write off policy
+#[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeDebug, MaxEncodedLen)]
+pub struct WriteOffRule<Rate> {
+	/// If any of the triggers is valid, the write-off rule can be applied
+	pub triggers: BoundedBTreeSet<WriteOffTrigger, WriteOffTrigger>,
```
lemunozm, line 166: I've used a set here instead of a vec because I understand the triggers must be unique per kind. This `BTreeSet` ensures that at the type level.

pallets/loans-ref/src/types.rs
```diff
@@ -486,6 +486,21 @@ impl<T: Config> ActiveLoan<T> {
...
+	) -> Result<bool, DispatchError> {
+		let now = T::Time::now().as_secs();
+		match trigger {
+			WriteOffTrigger::PrincipalOverdueDays(days) => {
+				let overdue_secs = SECONDS_PER_DAY.ensure_mul(days.ensure_into()?)?;
+				Ok(now >= self.maturity_date().ensure_add(overdue_secs)?)
+			}
+			WriteOffTrigger::OracleValuationOutdated(_seconds) => Ok(false),
+		}
+	}
```
lemunozm, line 502: I've moved this to the `ActiveLoan` type because if it grows, it will need more info coming from the loan type (overdue days, oracle timestamp, interest payments, ...) so it makes sense to be a loan method.

lemunozm (commented):
Some explanations of core parts:

offerijns (commented):
The new policy implementation looks great to me!

lemunozm:
@wischli I think the migration is ok (tested in `altair` and `centrifuge`), but it would be awesome if you could give a fast review in case you see something weird. It's the `v1`. Thanks 🙌🏻

lemunozm:
This is ready for a final review

offerijns (dismissed):

pallets/loans-ref/src/lib.rs
```diff
@@ -473,7 +480,7 @@ pub mod pallet {
 		/// - Write off by admin with percentage 0.5 and penalty 0.2
 		/// - Time passes and the policy can be applied.
 		/// - Write of with a policy that says: percentage 0.3, penaly 0.4
-		/// - The loan is written off with the maximum between the policy and the current state:
+		/// - The loan is written off with the maximum between the policy and the current rule:
 		///   percentage 0.5, penaly 0.4
```
wischli, line 484: Nit
```suggestion
		///   percentage 0.5, penalty 0.4
```

pallets/loans-ref/src/lib.rs
```diff
@@ -679,16 +688,38 @@ pub mod pallet {
...
+		/// From all overdue write off rules, it returns the one has highest percentage
+		/// (or highest penalty, if same percentage) that can be applied.
+		///
+		/// Suppose a policy with the following rules:
+		/// - overdue_days: 5,   percentage 10%
+		/// - overdue_days: 10,  percentage 30%
+		/// - overdue_days: 15,  percentage 20%
+		///
+		/// If the loan is not overdue, it will not return any rule.
+		/// If the loan overdue by 4 days, it will not return any rule.
```
wischli, line 700: Nit
```suggestion
		/// If the loan is overdue by 4 days, it will not return any rule.
```

pallets/loans-ref/src/lib.rs
```diff
@@ -679,16 +688,38 @@ pub mod pallet {
 			})
 		}
 
-		fn find_write_off_state(
+		/// From all overdue write off rules, it returns the one has highest percentage
```
wischli, line 691: Nit
```suggestion
		/// From all overdue write off rules, it returns the one with the highest percentage
```

pallets/loans-ref/src/lib.rs
```diff
@@ -679,16 +688,38 @@ pub mod pallet {
...
-			)
-			.ok_or_else(|| Error::<T>::NoValidWriteOffState.into())
+			loan: &ActiveLoan<T>,
+		) -> Result<Option<WriteOffRule<T::Rate>>, DispatchError> {
+			Ok(WriteOffPolicy::<T>::get(pool_id)
+				.into_iter()
+				.filter_map(|rule| {
+					rule.triggers
+						.iter()
+						.map(|trigger| loan.check_write_off_trigger(&trigger.0))
```
wischli, line 713: Nit: Maybe it would be better to destructure into the 0th tuple element directly to make it more explicit?

```suggestion
						.map(|(trigger, _)| loan.check_write_off_trigger(&trigger))
```

pallets/loans-ref/src/migrations/v1.rs
```diff
@@ -0,0 +1,133 @@
...
+					.try_into()
+					.expect("Size of the new vec can not be longer than previous one, qed"),
+			)
+		});
+
+		Pallet::<T>::current_storage_version().put::<Pallet<T>>();
+
+		log::info!("Successful migration: v0 -> v1. Items: {count}");
+
+		T::DbWeight::get().reads_writes(count, count)
```
wischli, line 65: Nit: It's actually `count + 1` since we read and write the `StorageVersion`:

```suggestion
		T::DbWeight::get().reads_writes(count.saturating_add(1), count.saturating_add(1))
```

You could also initialize `count = 1` and log 

```
		log::info!("Successful migration: v0 -> v1. Items: {count.saturating_sub(1)}");

```

pallets/loans-ref/src/migrations/v1.rs
```diff
@@ -0,0 +1,133 @@
...
+				policy.all(|(old, new)| {
+					*new == WriteOffRule::new(
+						[WriteOffTrigger::PrincipalOverdueDays(old.overdue_days)],
+						old.percentage,
+						old.penalty,
+					)
+				})
+			})
+			.then_some(())
+			.ok_or("Error: policies differ")
```
wischli, line 96: Very elegantly written assertion 😍

wischli (commented):
I only have a few nitpicks. The migration read-write-count is off by one but the rest looks splendid to me, great work again!

pallets/loans-ref/src/migrations/v1.rs
lemunozm: I thought the same at first, but then I checked some substrate pallets, and their Substrate returned `Weight::zero()` after reading the `StorageVersion`, which maybe means the version is something stored in the memory, not in the database 🤔.

I've tried to check if I'm right, but it's done procedurally...

lemunozm: Anyways adding 1 does not hurt!

pallets/loans-ref/src/lib.rs
lemunozm: It has only one element, the `map` part should be something like `.map(|UniqueWriteOffTrigger(trigger)|)`, which maybe is more confusing (line breaks and adds a new indentation). It if was a tuple, I'd agree with you.

pallets/loans-ref/src/lib.rs
lemunozm: Realizing that only means you did a deep review, thanks!!

pallets/loans-ref/src/migrations/v1.rs
lemunozm: Thanks! ❤️

wischli (approved):
Thanks for baking in my comments, looks great!

pallets/loans-ref/src/write_off.rs
```diff
@@ -0,0 +1,180 @@
...
+	Decode,
+	Clone,
+	PartialEq,
+	Eq,
+	PartialOrd,
+	Ord,
+	TypeInfo,
+	RuntimeDebug,
+	MaxEncodedLen,
+	EnumCount,
```
mustermeiszer, line 26: Why do we need that?

pallets/loans-ref/src/lib.rs
```diff
@@ -36,9 +36,14 @@
 //! The whole pallet is optimized for the more expensive extrinsic that is
 //! [`Pallet::update_portfolio_valuation()`] that should go through all active loans.
 
-pub mod migrations;
+pub mod migrations {
+	pub mod nuke;
+	pub mod v1;
+}
+
```
mustermeiszer, line 43: We should not keep them around forever or hide them behind a feature flag to not bloat the wasm. But rather minor.

pallets/loans-ref/src/write_off.rs
```diff
@@ -0,0 +1,180 @@
...
+)]
+pub enum WriteOffTrigger {
+	/// Number in days after the maturity date has passed
+	PrincipalOverdueDays(u32),
+
+	/// Seconds since the oracle valuation was last updated
+	OracleValuationOutdated(Moment),
```
mustermeiszer, line 33: What do we think we need this for? in which scenario will this be a trigger?

pallets/loans-ref/src/write_off.rs
offerijns: It's used to bound the `triggers` vec (as the triggers need to be unique, thus the bound should be on the number of variants in this enum)

pallets/loans-ref/src/write_off.rs
lemunozm: When an oracle, has not updated a price for a long time, the chain logic somehow differs from the real value of that price, and affects to the portfolio valuation, so we need to write off that loan in a percentage if it exceeds the maximum updated period for its price.

Issue #1315, opened 2023-04-12
Mock-builder proposal tracking


lemunozm:
# Description

Utility for building mocks from traits that works fine with Substrate pallets.

## Motivation
Loose coupling is great for avoiding the knowledge load of how other pallets work when you develop your own pallet. Nevertheless, this benefit and all the effort put into making this abstraction is lost once you need to choose an implementation for tests/benchmarks. Testing and benchmarking force you to have the dependencies and trait implementations you wanted to avoid by the usage of traits, getting the annoying extra load of knowledge again. 

Allow an automatic way of creating generic mocks allow to get rid of using real implementations in tests/benchmarks, and force you to improve the quality of your tests.

### In your trait `mocks` folder
A place where you create all mocks you need.

```rust
// Build the pallet boilerplate
#[mock_builder::pallet] 
mod pallet_mock_permission {
    //What you wrote:
    #[mock::trait]
    trait Permissions<Account> {
        #[mock::call(mock_has)]
        fn has(a: Self::Scope, b: AccountId, c: Self::Role) -> bool;

        #[mock::call(mock_add)]
        fn add(a: Self::Scope, b: AccountId, c: Self::Role) -> DispatchResult;

        #[mock::call(mock_remove)]
        fn remove(a: Self::Scope, b: AccountId, c: Self::Role) -> DispatchResult;
    }

    // Another trait for the same pallet here
    #[mock::trait]
    trait OtherTrait {
        //...
    }
}

// Other pallet mocks
```

### In your pallet `mock.rs` file
```rust
frame_support::construct_runtime!(
    pub enum Runtime where {
        System: frame_system,
        //...
        MockPermissions: pallet_mock_permissions,
    }
);

//...

impl pallet_mock_permissions::Config for Runtime {
    type Scope = PermissionScope<PoolId, CurrencyId>;
    type AccountId = AccountId;
    type Role = Role;
}
```

### In your tests
```rust
#[test]
fn check_borrower() {
    let borrower = 123;

    new_test_ext().execute_with(|| {
        MockPermission::mock_has(move |_, who, _| {
            // You can place any assert here or return any value you want.
            who == borrower
        });
        assert_ok!(Loans::borrow(borrower, ...));
    });
}
```

# Roadmap
- **Milestone 1**: Usage from test. Initial `mock-builder` implementation for using it in tests: #1220 
[...]

PR #1321, opened 2023-04-13
Mock-builder with generic methods support

### LINKED ISSUE #1316: Mock-builder: support trait methods with generic bounds
### Description

Currently, this is not supported:

```rust
trait MyTrait {
    fn my_generic_method<T: Into<i32> + 'static>(t: T);
}
```

`mock-builder` links a method with the name `mock_my_generic_method` with `my_generic_method`. The first one store a closure in the function storage, and the second one recovers that function and call it. The closure stored and recovered must have the same type. Nevertheless, each call to `my_generic_method` changes its type depending on the generics used when called:

```rust
    my_generic_method(42i8); // T is i8, a valid Into<i32>
    my_generic_method(42i16); // T is i16, a valid Into<i32>
```

This means that we'll have multiple different closures types stored for the same method. We need a way to distinguish and connect different versions of `my_generic_method` with different closures stored by `mock_my_generic_method`.

lemunozm:
# Description

This PR gives to `mock-builder` the capability of mocking methods with generic bounds

Fixes #1316

## Changes and Descriptions
- Added function identification with generics
- Fully refactorized to reduce macro usage and allows better organization
- Better error descriptions of what is going wrong.
- Added new tests

### DIFF
--- a/libs/mock-builder/src/lib.rs
+++ b/libs/mock-builder/src/lib.rs
@@ -87,7 +87,7 @@ where
 		.expect(&format!("Mock was not found. Location: {location:?}"));
 
 	storage::execute_call(call_id, input).expect(&format!(
-		"Mock was found but the its input/output types differ than {location:?}"
+		"Mock was found but its input/output types differ. Location: {location:?}"
 	))
 }
 
--- a/libs/mock-builder/src/lib.rs
+++ b/libs/mock-builder/src/lib.rs
@@ -39,194 +39,72 @@
 //!
 //! Take a look to the [pallet tests](`tests/pallet.rs`) to have a user view of how to use this crate.
 
-/// Provide methods for register/execute calls
-pub mod storage {
-	use std::{any::Any, cell::RefCell, collections::HashMap};
+/// Provide functions for register/execute calls
+pub mod storage;
 
-	/// Identify a call in the call storage
-	pub type CallId = u64;
+/// Provide functions for handle fuction locations
+pub mod location;
 
-	trait Callable {
-		fn as_any(&self) -> &dyn Any;
-	}
-
-	thread_local! {
-		static CALLS: RefCell<HashMap<CallId, Box<dyn Callable>>>
-			= RefCell::new(HashMap::default());
-	}
-
-	struct CallWrapper<Input, Output>(Box<dyn Fn(Input) -> Output>);
-
-	impl<Input: 'static, Output: 'static> Callable for CallWrapper<Input, Output> {
-		fn as_any(&self) -> &dyn Any {
-			self
-		}
-	}
-
-	/// Register a call into the call storage.
-	/// The registered call can be uniquely identified by the returned `CallId`.
-	pub fn register_call<F: Fn(Args) -> R + 'static, Args: 'static, R: 'static>(f: F) -> CallId {
-		CALLS.with(|state| {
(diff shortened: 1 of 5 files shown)


mustermeiszer (dismissed):
I did no deep deep dive but the changes look sane. But I also think we reached a point with this where moving to procedural macros makes sense in order to have the parsing happen during compilation.

lemunozm:
Thanks for your review! 

There is still a pending [issue](https://github.com/centrifuge/centrifuge-chain/issues/1322) that I would like to tackle before starting the procedural part. The current state is preferred to debug. [Roadmap](https://github.com/centrifuge/centrifuge-chain/issues/1315)

mustermeiszer (approved):

libs/mock-builder/src/storage.rs
```diff
@@ -0,0 +1,83 @@
...
+	fn different_output_type() {
+		let func_1 = |n: u8| -> usize { 23 * n as usize };
+		let call_id_1 = register_call(func_1);
+		let result = execute_call::<_, char>(call_id_1, 2u8);
+
+		assert_eq!(result, None);
+	}
+
+	#[test]
+	#[should_panic]
```
wischli, line 77: Maybe good to check for a more specific panic?

libs/mock-builder/tests/pallet.rs
```diff
@@ -190,4 +210,48 @@ mod test {
...
+				16
+			});
+
+			assert_eq!(MockAB::generic_input(1i8, 2u8), 8);
+			assert_eq!(MockAB::generic_input(3i16, 4u16), 16);
+		});
+	}
+
+	#[test]
+	#[should_panic]
```
wischli, line 234: Same as above: I would prefer to narrow down the panic.

wischli (approved):
Some really minor nit-picks from my side. Definitely not blocking, looks good!

libs/mock-builder/tests/pallet.rs
lemunozm: This panic comes from a bad macro usage, and I can not check from the test that "bad usage". Only I can expect it to panic at some point in the macro.

But this panic is ok from a use-case perspective. Tells the user they have configured its mock badly.

libs/mock-builder/src/lib.rs
```diff
@@ -39,177 +39,72 @@
...
-			std::any::type_name::<T>()
-		}
-		let name = type_name_of(f);
-		&name[..name.len() - "::f".len()]
-	}};
+/// Register a mock function into the mock function storage.
+/// This function should be called with a locator used as a function identification.
+pub fn register<Map, L, F, I, O>(locator: L, f: F)
+where
+	Map: StorageMap<<Blake2_128 as StorageHasher>::Output, CallId>,
```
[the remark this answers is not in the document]
lemunozm: I don't see right now a use case where a caller would want to choose another `Hasher` in this case. 

But thanks for your thoughts! I keep that in mind in case it evolves in a way a different `Hasher` can be useful.

libs/mock-builder/src/storage.rs
lemunozm: Ok, I've refactored the code a little bit, to support that case, with better error handling 👍🏻

mustermeiszer (approved):

PR #1323, opened 2023-04-17
Add transfer allowlist benchmarks

### LINKED ISSUE #1146: Pallet-Transfer-Allowlist: Benchmarking

* [x] Feature

### Description
* Implement benchmarking for pallet-transfer-allowlist #1144

thea-leake:
# Description
Adds benchmarks for allowlist pallet.

Fixes #1146

## Changes and Descriptions

Adds benchmarks for allowlist pallet.

# Checklist:


pallets/transfer-allowlist/src/benchmarking.rs
```diff
@@ -0,0 +1,298 @@
...
+	let sender: T::AccountId = account::<T::AccountId>("Sender", 1, 0);
+	let receiver: T::AccountId = account::<T::AccountId>("Receiver", 2, 0);
+	T::ReserveCurrency::deposit_creating(
+		&sender,
+		T::Fees::fee_value(T::AllowanceFeeKey::get()) * 4u32.into(),
+	);
+	(sender, receiver)
+}
+
+fn set_up_second_reciever<T: Config>() -> T::AccountId {
```
lemunozm, line 294:
```suggestion
fn set_up_second_receiver<T: Config>() -> T::AccountId {
```

pallets/transfer-allowlist/src/benchmarking.rs
```diff
@@ -0,0 +1,298 @@
...
+							sender,
+							CurrencyId::Native,
+					),
+					None
+			)
+	}
+
+	  purge_transfer_allowance_remaining_metadata {
+			  let (sender, receiver) = set_up_users::<T>();
+		let receiver_1 = set_up_second_reciever::<T>();
```
lemunozm, line 254: All benchmarks are not alignment due to be in the `benchmarks!` macro. We need to check it "manual". It's annoying, I know...

pallets/transfer-allowlist/src/benchmarking.rs
```diff
@@ -0,0 +1,298 @@
...
+			  Pallet::<T>::add_transfer_allowance(RawOrigin::Signed(sender.clone()).into(), CurrencyId::Native, receiver.clone().into())?;
+			  Pallet::<T>::add_transfer_allowance(RawOrigin::Signed(sender.clone()).into(), CurrencyId::Native, receiver_1.clone().into())?;
+			  Pallet::<T>::remove_transfer_allowance(RawOrigin::Signed(sender.clone()).into(), CurrencyId::Native, receiver.clone().into())?;
+	  }:purge_transfer_allowance(RawOrigin::Signed(sender.clone()), CurrencyId::Native, receiver.clone().into())
+		  verify {
```
lemunozm, line 259: A comment about `verify` in macros. The intention of this part of the macro is not to check the behavior itself (this is done well in the tests); it's to check that you're testing the worst case for that extrinsic. From my understanding of the library, if the extrinsic went ok, it computes the more expensive path, so in your case, you don't need to fill `verify` sections. So you can remove this safety and have less code to maintain (This also applies to other benchmarks)

pallets/transfer-allowlist/src/benchmarking.rs
```diff
@@ -0,0 +1,298 @@
...
+			  allowance_count: 0,
+			  current_delay: Some(200u32.into()),
+			  once_modifiable_after: None
+		  }
+	  )
+  }
+	add_allowance_delay_existing_metadata {
+			let (sender, receiver) = set_up_users::<T>();
+			Pallet::<T>::add_transfer_allowance(RawOrigin::Signed(sender.clone()).into(), CurrencyId::Native, receiver.clone().into())?;
+	}:add_allowance_delay(RawOrigin::Signed(sender.clone()), CurrencyId::Native, 200u32.into())
```
lemunozm, line 93: I see you're benchmarking the same extrinsic with different setups, I think this is not needed. Choosing the worst case is enough. (This applies to other extrinsics)

pallets/transfer-allowlist/Cargo.toml
```diff
@@ -20,12 +20,14 @@ cfg-traits = { path = "../../libs/traits", default-features = false }
...
+
 frame-benchmarking = { git = "https://github.com/paritytech/substrate", default-features = false, optional = true, branch = "polkadot-v0.9.37" }
 frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.37" }
 frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.37" }
 pallet-balances = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.37" }
 sp-core = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.37" }
-sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false, optional = true, branch = "polkadot-v0.9.37" }
+sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.37" }
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.37" }
 xcm = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.37" }
```
lemunozm, line 32: Does this `xcm` dependency still need?

libs/test-utils/src/system/time_travel.rs
```diff
@@ -0,0 +1,27 @@
...
+pub fn advance_n_blocks<T: frame_system::Config>(n: u64) {
+	for _ in 0..n {
+		let h = frame_system::Pallet::<T>::finalize();
+		let b = h
+			.number()
+			.checked_add(&T::BlockNumber::one())
+			.expect("Mock block count increase failed");
+		frame_system::Pallet::<T>::initialize(&b, h.parent_hash(), h.digest());
+	}
+}
```
lemunozm, line 27: If you only need to advance blocks, you can use `frame_system::Pallet<T>::set_block_number()` utility under `std` and `runtime-benchmarks`

lemunozm (changes requested):
First review round, looks good! 👍🏻 

I left some comments regarding things I've learned after flighting a lot with benchmarking.

libs/test-utils/src/system/time_travel.rs
wischli: That is true. However, if any hook from `on_initialize` or `on_finalize` is required, the above approach should be taken instead.

pallets/transfer-allowlist/src/benchmarking.rs
wischli: IMO, there should be at least one check still in `verify`. It is true that is a bit redundant with unit tests. However, the latter only work against a mock whereas real benchmarks apply the actual runtime configuration.

lemunozm: The system automatically verifies that it returns `Ok()`, so there is already an implicit "verification". In most cases, you don't need to go into more detail here. Although, it all depends on the extrinsic complexity of course.

libs/test-utils/src/system/time_travel.rs
lemunozm: Agree

pallets/transfer-allowlist/src/benchmarking.rs
thea-leake: In this particular case I wanted to make sure that the worst case was covered in case creating the metadata becomes more expensive in the future -- right now it's faster as it's writing the same amount when creating, but just pulling from default values.  I can remove the no existing metadata benchmark though.

pallets/transfer-allowlist/src/tests.rs
```diff
@@ -10,13 +10,13 @@ fn add_transfer_allowance_works() {
 	new_test_ext().execute_with(|| {
 		assert_ok!(TransferAllowList::add_transfer_allowance(
 			RuntimeOrigin::signed(SENDER),
-			CurrencyId::A,
```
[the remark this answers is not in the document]
thea-leake: Switched back to custom currency id in mocks with default for compatibility in tests/benchmarks across benchmark runs in tests, and runtime benchmarking.  Also similarly updated location type as well in tests/mocks.

pallets/transfer-allowlist/src/benchmarking.rs
mustermeiszer: I disagree with @lemunozm. Benchmarks, if done well and right, make the chain more efficient. Albeit knowing, that reads and writes are probably at all times the heavier instructions, we should not neglect the fact that we should bench all codepaths and let the system decide what is the heaviest. 

If we do the weights correctly, we should return the overweight fees via `DispatchWithPostInfo`. But defaulting to the above is not a good idea IMO.

mustermeiszer (commented):
This looks really good! 

I would counter argue the changes @lemunozm requested and move back to getting the max in the weight calculation. I don't see a reasoable argument to not do it with the max.

pallets/transfer-allowlist/src/benchmarking.rs
lemunozm: What you say is the next step. I mean, the first step is to determine and make a benchmark for the worst case (the safety step). Later the second step is what you say, where you fix your extremely bad measured extrinsic with `DispatchWithPostInfo` for the no-worst cases (the performance step).

lemunozm: Anyways I think these performance improvements should be evaluated to see if it worth it to do it or not, and if we are at the point to do it. Probably in simple functions like these ones, we do no need to optimize for this. And doing it adds a lot of complexity for very little improvement. As the famous sentence says: “Premature optimization is the root of all evil”

lemunozm: @mustermeiszer Looking the implementation of both paths, they are almost the same. Not sure if we want to add a `max()` complexity for an improvement close to the benchmark noise.

libs/test-utils/src/system/time_travel.rs
thea-leake: For this case it could be done with setting the block number.  My preference would be to keep this as is as it is more generic, and how the chain would actually function, but can also change.

lemunozm: No strong opinion 👍🏻

Only highlight that in order to leave this generically usable, we would need to call `on_finalize() / on_initialize()` over all runtime pallets. i.e. `pallet_timestamp` also change things between blocks, that could be required for the user of this function.

pallets/transfer-allowlist/Cargo.toml
```diff
@@ -31,6 +31,7 @@ sp-runtime = { git = "https://github.com/paritytech/substrate", default-features
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.37" }
 
 [dev-dependencies]
+cfg-mocks = { path = "../../libs/mocks" }
 pallet-authorship = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.37" }
```
lemunozm, line 35: `pallet-authorship`, `pallet-fees`, and `pallet-treasury` can now be removed 🎉

lemunozm (commented):
Great @thea-leake! I add some minor comments, but the change is just what you did! 👍🏻

pallets/transfer-allowlist/src/benchmarking.rs
mustermeiszer: I am not strongly against removing it in this case. It is just not that big of "not-good-looking-style" or complexity to have a single max call at the head of an extrinsic definition. But it ensures we are not missing possible changes when the extrinsic gets heavier in the future. 

I would always lean for more security.

libs/test-utils/src/system/time_travel.rs
mustermeiszer: It is really hard to generate a correct block -- i.e. run all hooks of a runtimes pallet. This is the main reason I build Fudge. But feel free to try it. For now, I think we can keep it, although, I agree with Luis.

pallets/transfer-allowlist/src/benchmarking.rs
lemunozm: > to have a single max call at the head of an extrinsic definition

It's not only that, for me, the big issue is to maintain the double benchmarks

Of course, each case has its own complexity and should be evaluated if it needs extra benchmarks for its different internal paths. Maybe I've sounded very strong in my first message against this, there are cases that need what you say.

pallets/transfer-allowlist/src/benchmarking.rs
```diff
@@ -0,0 +1,113 @@
...
+
+	add_transfer_allowance_no_existing_metadata {
+		let (sender, receiver) = set_up_users::<T>();
+	}:add_transfer_allowance(RawOrigin::Signed(sender.clone()), T::CurrencyId::default(), receiver.clone().into())
+
+
+	add_transfer_allowance_existing_metadata {
+		let (sender, receiver) = set_up_users::<T>();
+		Pallet::<T>::add_allowance_delay(RawOrigin::Signed(sender.clone()).into(), T::CurrencyId::default(), 200u32.into())?;
+	}:add_transfer_allowance(RawOrigin::Signed(sender.clone()), T::CurrencyId::default(), receiver.clone().into())
```
lemunozm, line 53: Not sure which resolution to [this thread](https://github.com/centrifuge/centrifuge-chain/pull/1323#discussion_r1187061233) you've chosen @thea-leake, if finally you go without the `max()`, you can remove one of this benchmarks

pallets/transfer-allowlist/src/mock.rs
```diff
@@ -180,20 +114,61 @@ pub enum CurrencyId {
...
+	TestLocal(u64),
+}
+
+impl From<u64> for Location {
+	fn from(a: u64) -> Self {
+		Self::TestLocal(a)
+	}
+}
+// Used to handle reserve/unreserve for allowance creation.
+// Loosely coupled with transfer_allowlist
```
lemunozm, line 147: Outdated comment I think

lemunozm (approved):
I think everything is ok. Good work @thea-leake!

pallets/transfer-allowlist/src/benchmarking.rs
thea-leake: I'm thinking atm that it's simpler and safer to have the multiple benchmarks for the different scenarios, than to place the burden on any future work that might need to be done with the pallet. I feel that without having to have the verify in the benchmarks maintaining the benchmarks for different scenarios is comparatively simple.

lemunozm (approved):
LGTM! 🚀

PR #1335, opened 2023-05-03
Mock-builder: add support for custom lifetimes

### LINKED ISSUE #1322: Mock-builder: support trait methods with no static lifetimes
### Description

Although the following are supported in #1316:
```rust
trait MyTrait {
    fn my_generic_method<T: Into<i32> + 'static>(t: T);
}
````

This not:
```rust
trait MyTrait {
    fn my_generic_method<T: Into<i32>>(t: T);
}
```

The `'static` lifetime should not be mandatory in the trait definition for creating mocks, and we should be able to mock any method with any lifetime.

This is quite annoying because it forces the trait definition to add that `'static` in the best cases, and in the worst cases, we can not modify it because it's a foreign trait (and then, we can not mock it). We need support for those traits.

### Research

This change is far to be trivial since we need a new way to store functions, different than `thread_local` (which can only store `'static` lifetimes), and we can no longer use `Any::downcast()` to recover the type (because `Any` is only implemented for `'static`).

There is no issue in using a non-static lifetime as a function parameter since those values are not stored anywhere.

#### Ideas
- Use `unsafe` to transmute a wide pointer as `&dyn Fn` to `u128` (storing the pointer and the metadata point in a single `u128`).
  ```rust
  fn any_to_wide_addr<F: std::any::Any>(f: F) -> u128 {
    unsafe { std::mem::transmute(&f as &dyn std::any::Any) }
  }
  ```
  because `F` is `'static` we can use `Any` in this case.

- Use the `FunctionLocation` done in #1316 to check the signature of the method in order to ensure at runtime that `mock_method` and `method` use the same signature, and it will not crash in the transmutation.

lemunozm:
# Description

This is the last technical requirement for `mock-builder`. After this PR, I think we can test any possible trait. 🎉 Future work is to create procedural macros to reduce boilerplate. [Roadmap here](https://github.com/centrifuge/centrifuge-chain/issues/1315)

New features:
- References supported.
- Generics with random lifetimes (other than `'static`) supported.

Fixes #1322

## Changes and Descriptions

- New storage implementation to support input/output types that could not be `'static`


libs/mock-builder/src/storage.rs
```diff
@@ -1,64 +1,87 @@
...
 }
 
 /// Register a call into the call storage.
 /// The registered call can be uniquely identified by the returned `CallId`.
-pub fn register_call<F: Fn(Args) -> R + 'static, Args: 'static, R: 'static>(f: F) -> CallId {
+pub fn register_call<F: Fn(I) -> O + 'static, I, O>(f: F) -> CallId {
+	let f = Box::new(f) as Box<dyn Fn(I) -> O>;
+	let ptr = Box::into_raw(f);
+
+	// SAFETY: transforming a wide pointer to an u128 is always safe.
```
wischli, line 42: Generally, I don't think this is true. Can you show a short proof?

libs/mock-builder/src/storage.rs
```diff
@@ -1,64 +1,87 @@
...
-			.ok_or(Error::TypeNotMatch)
+
+		if expected_type_signature != call.type_signature {
+			return Err(Error::TypeNotMatch(
+				expected_type_signature,
+				call.type_signature.clone(),
+			));
+		}
+
+		// SAFETY: The existance of this boxed clousure in consequent calls is ensured
```
wischli, line 71: nit
```suggestion
		// SAFETY: The existence of this boxed closure in consequent calls is ensured
```

wischli: This is some real inception level stuff: The existence is ensured by the `forget` call 🤯

wischli (dismissed):
I have to admit I can only follow on a much much higher level and on that, it looks good to me! I'd be interested in an argument for your safe `unsafe` transmutation, see below.

libs/mock-builder/src/storage.rs
lemunozm: Realize this means that you have done a very good review. Thanks so much! 🙌🏻

libs/mock-builder/src/storage.rs
lemunozm: It's safe in the same way a "normal" reference is always transformable into a `* const u64`. In fact, the following line does not require unsafe rust:

```rust
let reference: &u64 = ...
let ptr = reference as *const u64 
```

This is safe because, it is just an address, a number. What it's not safe is what you do with that address, because all invariants and protections are broken after this.

This is why the other unsafe is the critical one, and you must ensure that the ptr still exists and corresponds to the same type as the original.

lemunozm: Regarding "why exactly `u128`?" It's because any `&dyn Type` is a wide pointer: one part corresponds to the pointer itself and the other to the metadata pointer that contains the trait information of the closure. We do not care about how both pointers are sorted in memory, because we are undoing the same we did in the other method for the same type.

libs/mock-builder/src/storage.rs
```diff
@@ -1,64 +1,87 @@
...
 	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-		write!(
-			f,
-			"{}",
-			match self {
-				Error::CallNotFound => "Trying to call a function that is not registered",
-				Error::TypeNotMatch => "The function is registered but the type mismatches",
-			}
-		)
+		match self {
```
thea-leake, line 25: Nice!

libs/mock-builder/src/storage.rs
```diff
@@ -1,64 +1,87 @@
...
+use std::{cell::RefCell, collections::HashMap, fmt};
+
+use super::util::TypeSignature;
 
 /// Identify a call in the call storage
 pub type CallId = u64;
 
-trait Callable {
-	fn as_any(&self) -> &dyn Any;
+struct CallInfo {
```
thea-leake, line 8: Very cool that we can have this as known fixed size, though have comment below in casting thread about the transmute.

thea-leake (approved):
LGTM! :)
Added a few comments, but nothing blocking.

libs/mock-builder/src/storage.rs
```diff
@@ -1,64 +1,87 @@
...
-			.ok_or(Error::TypeNotMatch)
+
+		if expected_type_signature != call.type_signature {
+			return Err(Error::TypeNotMatch(
+				expected_type_signature,
+				call.type_signature.clone(),
+			));
+		}
+
+		// SAFETY: The existence of this boxed closure in consequent calls is ensured
```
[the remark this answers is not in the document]
lemunozm: Ok! I'll do it. Thanks for the advice and for taking the time to deep into it, it's not an easy PR.

libs/mock-builder/src/storage.rs
lemunozm: Agree 👍🏻  I'll add a better explanation. Regarding the 2 `usize` fields, I think separating could confuse in the sense that seems that they should be used separately when not. We only need to allocate that, but not use the internals. Again, I think a good explanation into the code is needed.

lemunozm:
@thea-leake, I've documented every weird case I was able to see. In the process, I've simplified a bit the `forget()` call part (which is no longer needed).

thea-leake (approved):
LGTM!

PR #1340, opened 2023-05-10
refactor: Drop tranche multiLocation

### LINKED ISSUE #1336: Drop Tranche `location`
Currently, the `pool_system::tranches::create_metadata` function defines the `AssetMetadata.location` field as follows:

``` rust
location: Some(VersionedMultiLocation::V3(MultiLocation {
	parents: 1,
	interior: X3(
		Parachain(parachain_id.into()),
		PalletInstance(pallet_index),
		// todo(nuno): revisit this
		GeneralKey {
			length: 16u8,
			data: vec_to_fixed_array(tranche_id),
		},
	),
})),
```

I remember suggesting this structure many months back, but as I am [updating the codebase to Polkadot v0.9.38](#1332) and XCM v3 and bumped into this, I realised we don't need a location for `tranche` tokens at all. Tranche tokens are not to be handled through XCM due to their permissions nature and we are building support for permissioned transfers through Connectors.

### Suggestion

Basically set `location` to `None` and adjust all tests.

``` diff
+ location: None,
- location: Some(VersionedMultiLocation::V3(MultiLocation {
- 	parents: 1,
- 	interior: X3(
- 		Parachain(parachain_id.into()),
- 		PalletInstance(pallet_index),
- 		// todo(nuno): revisit this
- 		GeneralKey {
- 			length: 16u8,
- 			data: vec_to_fixed_array(tranche_id),
- 		},
- 	),
- })),
```

NunoAlexandre:
Closes #1336 

We make `AssetMetadata.location` of a `Tranche` simply `None` since having a `MultiLocation` for Tranche tokens implies they are XCM-transferrable when in fact we have specific Xcm configs in place to forbid tranche tokens transfers. Tranche tokens are only transferrable through Centrifuge Connectors, which guarantees that the permissioned-nature of tranche tokens are respected.

### Changes

- [x] 1. We set the `AssetMetadata.location` of a `Tranche` token to `None`
- [x] 2. We add a migration for all the runtimes that ensure 1. for all the already registered Tranche tokens

### To Do

- [x] Write a migration that, for every `CurrencyId::Tranche` in entry in the `AssetRegistry`, updates that asset by setting `location` to `None`. 
  - [x] Centrifuge
  - [x] Altair
  - [x] Development

### Discussion

Can we dry the migration so that we define it once for all all runtimes? We need to reference directly on the (external dependency) `orml_asset_registry` which I don't think we can abstract over the runtime it's in 🤔


runtime/altair/src/lib.rs
```diff
@@ -1489,16 +1489,14 @@ pub type UncheckedExtrinsic =
...
-
 /// Executive: handles dispatch to the various modules.
 pub type Executive = frame_executive::Executive<
 	Runtime,
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	UpgradeAltair1027,
+	migrations::Migrations,
```
mustermeiszer, line 1499: Can we stay with the naming `UpgradeAltair1028` instead of tupling migrations here?

runtime/development/src/lib.rs
```diff
@@ -1922,14 +1922,9 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	UpgradeDev1020,
+	migrations::Migrations,
```
mustermeiszer, line 1925: Same as above.

runtime/centrifuge/src/lib.rs
```diff
@@ -1666,16 +1666,14 @@ pub type UncheckedExtrinsic =
...
-
 /// Executive: handles dispatch to the various modules.
 pub type Executive = frame_executive::Executive<
 	Runtime,
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	UpgradeCentrifuge1019,
+	migrations::Migrations,
```
mustermeiszer, line 1676: Same as above.

runtime/altair/src/migrations.rs
```diff
@@ -0,0 +1,74 @@
...
+					// symbol
+					None,
+					// existential_deposit
+					None,
+					// location: we do set it to `None`
+					Some(None),
+					// additional
+					None,
+				)
+				.expect("TrancheLocationMigration: Failed to update tranche token");
```
mustermeiszer, line 47: When would this fail? ^^

mustermeiszer (commented):
I would really like us to stick with the naming for upgrades, so use `UpgradeAltair1028` instead of just the plain migration. Makes it so much easier in the future to remove what needs to be removed. 

Otherwise, looks good!

### For approval:
Need clarification whether the following will never fail:
* `let location: MultiLocation = old_location.clone().try_into().map_err(|()| Error::<T>::BadVersion)?;`

pallets/pool-system/src/tests/mod.rs
```diff
@@ -20,15 +20,10 @@ use cfg_types::{
 use frame_support::{assert_err, assert_noop, assert_ok};
 use orml_traits::asset_registry::{AssetMetadata, Inspect};
 use rand::Rng;
-use sp_core::{storage::StateVersion, Encode};
+use sp_core::storage::StateVersion;
 use sp_runtime::{
-	traits::{ConstU32, One, Zero},
-	FixedPointNumber, Perquintill, TokenError, WeakBoundedVec,
-};
-use xcm::{
```
thea-leake, line 28: It looks like `xcm` has been removed from both places it's used in the `pool-system` pallet.  I think we can remove it from the pallets `Cargo.toml`.

pallets/pool-system/src/lib.rs
```diff
@@ -40,7 +40,6 @@ use orml_traits::{
 	Change,
 };
 pub use pallet::*;
-use polkadot_parachain::primitives::Id as ParachainId;
```
thea-leake, line 43: It looks like all usages of `polkadot_parachain` have been removed for the `pool-systems` pallet.  I think we can remove from pallet `Cargo.toml`.

thea-leake (commented):
Looks good!  Left some comments about crates we can likely remove from the `pool-system` deps.

pallets/pool-system/src/lib.rs
NunoAlexandre: Ah yes, I thought about this last night during my sleep xD thanks for the catch!

runtime/altair/src/lib.rs
NunoAlexandre: I also like the specific info about the version the migration is referring to. The reason why I named it `Migrations` was so that we don't have to change the code for each RUT migration (i.e, have to update the version in the name) and instead have that info as a comment above the migration tuple. If you (and @wischli) still prefer to keep the version info in the name I am cool with that 👍  let me know

pallets/pool-system/src/lib.rs
NunoAlexandre: Done in https://github.com/centrifuge/centrifuge-chain/pull/1340/commits/0bf232376e5723ed78f3eed59297f20def740a8f ✅

pallets/pool-system/src/tests/mod.rs
NunoAlexandre: Done in https://github.com/centrifuge/centrifuge-chain/pull/1340/commits/0bf232376e5723ed78f3eed59297f20def740a8f ✅

runtime/altair/src/migrations.rs
NunoAlexandre: Yeah this really shouldn't happen if the migration ran 🤷  so here we either ignore the result (i.e., `let _ = `) or we `expect(..)`. What's your preference? Or maybe you have a better alternative? :)

NunoAlexandre:
> I would really like us to stick with the naming for upgrades, so use `UpgradeAltair1028` instead of just the plain migration. Makes it so much easier in the future to remove what needs to be removed.
> 
> Otherwise, looks good!
> 
> ### For approval:
> Need clarification whether the following will never fail:
> 
> * `let location: MultiLocation = old_location.clone().try_into().map_err(|()| Error::<T>::BadVersion)?;`

@mustermeiszer what's the context for this code snippet? 🤔

mustermeiszer:
Just the only error that could happen behind the expect call in the upgrade logic. Not really a blocker though. . Was just curious or wanted to be sure they expect never gets triggered

NunoAlexandre:
> Just the only error that could happen behind the expect call in the upgrade logic. Not really a blocker though. . Was just curious or wanted to be sure they expect never gets triggered

Yeah, that's a valid concern. Maybe we can just `log` in case it fails?

NunoAlexandre:
@wischli @mustermeiszer I believe I have addressed everything, thanks for the feedback and I look forward for the next round :)

mustermeiszer (approved):
LGTM!

runtime/altair/src/migrations.rs
```diff
@@ -0,0 +1,75 @@
...
+					None,
+					// existential_deposit
+					None,
+					// location: we do set it to `None`
+					Some(None),
+					// additional
+					None,
+				) {
+					Err(e) => log::error!("TrancheLocationMigration: Failed to update asset with underlying error: {:?}", e),
+					_ => continue,
```
wischli, line 47: Nit: Since we are matching for two possible outcomes and only working on the error variant, it would be cleaner to map the error and then drop the result instead.

```suggestion
				OrmlAssetRegistry::do_update_asset(
					asset_id,
					// decimals
					None,
					// name
					None,
					// symbol
					None,
					// existential_deposit
					None,
					// location: we do set it to `None`
					Some(None),
					// additional
					None,
				)
				.map_err(|e| {
					log::error!("TrancheLocationMigration: Failed to update asset with underlying error: {:?}", e);
				})
				.ok();
```

runtime/altair/src/migrations.rs
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		// todo(nuno): not sure how to build this properly,
+		// setting it to a conservative value for now.
+		Weight::from_ref_time(200_000_000)
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<sp_std::vec::Vec<u8>, &'static str> {
+		Ok(Default::default())
```
wischli, line 59: Could: Return list of current orml registered assets and assert in `post_upgrade` that none of these have been tampered with apart from the location, if its a tranche token.

wischli (approved):
LGTM! I left two nitpicks which definitely don't need to be addressed.

UPDATE: Seeing the conflicts, I would recommend to move your runtime migration code into its own module inside `runtime/$runtime/migrations.rs`. In #1342, I ensured all imports tied to a specific migration where part of the module's scope such that deprecated migrations can be removed seamlessly without breaking clippy.

runtime/altair/src/migrations.rs
NunoAlexandre: Applied :) I forgot that in Rust a map can be just a closure that doesn't return a value per se. Thanks!

mustermeiszer (approved):

NunoAlexandre:
@wischli @mustermeiszer could you approve this again? :)

runtime/altair/src/migrations.rs
NunoAlexandre: It's a fair recommendation but given the simplicity of this migration I will skip it at this time.

pallets/pool-system/Cargo.toml
```diff
@@ -29,17 +29,16 @@ sp-arithmetic = { git = "https://github.com/paritytech/substrate", default-featu
...
-xcm = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.37" }
 
 # Benchmarking dependencies - optional
 frame-benchmarking = { git = "https://github.com/paritytech/substrate", default-features = false, optional = true, branch = "polkadot-v0.9.37" }
+orml-asset-registry = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, optional = true, branch = "polkadot-v0.9.37" }
 pallet-investments = { path = "../../pallets/investments", default-features = false, optional = true }
 
 [dev-dependencies]
 cfg-test-utils = { path = "../../libs/test-utils", default-features = true }
+orml-asset-registry = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.37" }
```
lemunozm, line 41: I think this is duplicated

lemunozm (approved):
Re-approving based on previous approvals 😃

offerijns (approved):

PR #1341, opened 2023-05-11
gateway: Move domain routers to separate crate

cdamian:
Intermediary PR for addressing the routers used in the upcoming `pallet-connectors-gateway`.


runtime/development/src/connectors.rs
mustermeiszer, line 1: Why do not have this in the gatway crate? I would not put such specific logic into the runtimes. I would really like the routers to be part of the gatway. WDYT?

mustermeiszer (commented):
Like like like! Just would like to have the existing router being a `MoonbeamRouter` struct in the gateway crate. Or maybe its own crate in the gatway crate for seperation.

runtime/development/src/connectors.rs
cdamian: Ah OK, I thought we wanted this out of gateway, gonna move it there then. Cheers!

mustermeiszer: Partially. I think creating new creates for those is a good setup. Inside of the Gateway crate. Then we can change them without touching the gateway itself. WDYT? But having them in modules in the gateway is also fine for me.

cdamian: Yup, looking back at this I think a separate crate with its own deps would make a lot of sense.

cdamian: Done, if this looks OK to you I suggest we merge this PR and then I can look into adding an Axelar router using the changes done in https://github.com/centrifuge/axelar-cgp-substrate/pull/15/files#diff-3b478a17b784d7eda1e3f5580ed1084447dd9e6495367f1e246af6bbd9a24c32

mustermeiszer (approved):
Merge whenever you think it is ready to go into the feature branch

pallets/connectors-gateway/src/lib.rs
```diff
@@ -140,29 +152,42 @@ pub mod pallet {
...
 		#[pallet::weight(< T as Config >::WeightInfo::add_submitter())]
 		#[pallet::call_index(1)]
 		pub fn add_submitter(origin: OriginFor<T>, submitter: DomainAddress) -> DispatchResult {
 			T::AdminOrigin::ensure_origin(origin.clone())?;
 
+			if let Domain::Centrifuge = submitter.domain() {
+				return Err(Error::<T>::DomainNotSupported.into());
+			}
+
 			<DomainSubmitters<T>>::try_mutate(submitter.domain(), |submitters| {
```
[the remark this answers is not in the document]
NunoAlexandre: Also, do we have an example of where the origin of a message coming from a Domain won't be the Connector address on that domain? At least that's the case with the POC Connector on the Moonbeam's EVM. Just curious about edge cases to that; and if the origin is always dependant on the signer of the source transaction, like how will we keep up with allowing all the users?

pallets/connectors-gateway/src/lib.rs
```diff
@@ -274,91 +254,13 @@ pub mod pallet {
...
-					overall_weight: None,
-				},
-			)?;
-
-			Ok(())
-		}
-	}
+			if let Domain::Centrifuge = destination {
+				return Err("invalid outbound destination".into());
+			}
```
NunoAlexandre, line 259: `ensure!` here as well

pallets/connectors-gateway/src/lib.rs
```diff
@@ -274,91 +254,13 @@ pub mod pallet {
...
-						+ 100_000_000,
-					overall_weight: None,
-				},
-			)?;
-
-			Ok(())
-		}
-	}
+			if let Domain::Centrifuge = destination {
+				return Err("invalid outbound destination".into());
```
NunoAlexandre, line 258: We _could_ be more specific with this error message and say that we can't send a message to the domain itself

runtime/development/Cargo.toml
```diff
@@ -115,6 +115,7 @@ pallet-claims = { path = "../../pallets/claims", default-features = false }
 pallet-collator-allowlist = { path = "../../pallets/collator-allowlist", default-features = false }
 pallet-connectors = { path = "../../pallets/connectors", default-features = false }
 pallet-connectors-gateway = { path = "../../pallets/connectors-gateway", default-features = false }
+gateway-routers = { path = "../../pallets/connectors-gateway/gateway-routers", default-features = false }
```
NunoAlexandre, line 118: I feel a bit uncomfortable that these sub pallets of connectors are not named accordingly 🤔 

I would even suggest that we do:

```
- pallets/
  - connectors/                    <--- "pallet-connectors"
  - connectors-gateway/    <--- "pallet-connectors-gateway"
    - routers                          <--- "pallet-connectors-gateway-routers"
```

Or even have `connectors-gateway/` under `connectors/` and just name it `gateway` while still keeping the absolute name `pallet-connectors-gateway`. I don't feel strongly about this last suggestion but we should at least consider the first one.

pallets/connectors-gateway/src/lib.rs
```diff
@@ -46,26 +40,47 @@ pub mod pallet {
...
+			+ PartialEq;
+
 		/// The type that processes incoming messages.
 		type Connectors: InboundQueue<Sender = Domain, Message = Self::Message>;
 
+		type WeightInfo: WeightInfo;
+
 		/// Maximum size of an Ethereum message.
 		#[pallet::constant]
 		type MaxEthMsg: Get<u32>;
```
NunoAlexandre, line 86: It seems like we don't use this anywhere at the moment. I'd rather not have code around that we don't use, or we it is where we should. Also, can we [add the units to the naming](https://ruudvanasseldonk.com/2022/03/20/please-put-units-in-names) please :) ?

libs/traits/src/connectors.rs
```diff
@@ -21,6 +21,15 @@ pub trait Codec: Sized {
 	fn deserialize<I: Input>(input: &mut I) -> Result<Self, codec::Error>;
 }
 
+/// The trait required for forwarding messages.
+pub trait Router {
+	type Message;
+
+	type Sender;
+
+	fn forward(&self, sender: Self::Sender, message: Self::Message) -> DispatchResult;
```
NunoAlexandre, line 30: On the `Solidity` side we named this `send`, which I personally prefer over `forward`. Can we align here and name it `send` as well?

pallets/connectors-gateway/gateway-routers/src/moonbeam.rs
```diff
@@ -0,0 +1,214 @@
...
+					+ 100_000_000,
+				overall_weight: None,
+			},
+		)?;
+
+		Ok(())
+	}
+
+	/// COPIED FROM CONNECTORS
+	///
```
NunoAlexandre, line 79:
```suggestion
```

NunoAlexandre (commented):
Good work overall! Aside from the review comments, some more general ones:

- Can we rename the title of this PR to `connectors: Add pallet gateway`

- The fact that this PR is adding changes on top of `[add-pallet-connectors-gateway](https://github.com/centrifuge/centrifuge-chain/tree/add-pallet-connectors-gateway)` made it a bit harder to review because some pieces of showing are new but didn't show as such and had me quite confused as I didn't remember to add stuff like the submitters 😆

pallets/connectors-gateway/gateway-routers/src/moonbeam.rs
```diff
@@ -0,0 +1,214 @@
+// Copyright 2021 Centrifuge Foundation (centrifuge.io).
```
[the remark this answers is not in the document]
mustermeiszer: Sounds reasonable.

cdamian:
> Good work overall! Aside from the review comments, some more general ones:
> 
>     * Can we rename the title of this PR to `connectors: Add pallet gateway`
> 
>     * The fact that this PR is adding changes on top of `[add-pallet-connectors-gateway](https://github.com/centrifuge/centrifuge-chain/tree/add-pallet-connectors-gateway)` made it a bit harder to review because some pieces of showing are new but didn't show as such and had me quite confused as I didn't remember to add stuff like the submitters laughing

Changed the title of the PR to something that better reflects the changes done.

pallets/connectors-gateway/gateway-routers/src/moonbeam.rs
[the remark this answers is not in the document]
cdamian: Are there any other parachains that this would work with? Otherwise I would keep this as it is in order to make sure there's a clear distinction between the `ethereum_xcm` setups that can be available.

NunoAlexandre:
> Changed the title of the PR to something that better reflects the changes done.

Thanks :p it doesn't matter in the end since we are merging it to the feature branch, sorry for the noise

pallets/connectors-gateway/gateway-routers/src/moonbeam.rs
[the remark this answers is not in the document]
NunoAlexandre: No, we only integrate with Moonbeam at the moment. I am fine with keeping it as is 👍

pallets/connectors-gateway/src/lib.rs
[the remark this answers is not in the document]
cdamian: To me, `AllowedDomainOrigin(s)` would be confusing since I would be thinking about the substrate specific origin, which should be an admin in this case. Maybe we can find a better replacement for `submitter`/`origin`, like `caller` or some other word which escapes me at this point? :D

pallets/connectors-gateway/gateway-routers/src/moonbeam.rs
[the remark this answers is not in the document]
cdamian: Let's keep it like this for now, maybe we'll find a more fitting name after we add more routers.

pallets/connectors-gateway/src/lib.rs
cdamian: Renamed to `MaxEthMsgSize`, and it is used in `process_msg`.

pallets/connectors-gateway/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: Right, I see your point. Maybe I just need to get used to the naming. But my most pressing question here is to understand how this will work; are we only allowing a subset of specific addresses to be able to - essentially - use connectors on another domain? That's the idea I am getting from this but that's not aligned to what I'd expect this to work like :P

pallets/connectors-gateway/src/lib.rs
NunoAlexandre: > and it is used in `process_msg`.

Thanks, I missed that!

So, basically this is used to represent the max size of a Connectors message; More specifically, incoming Connectors messages, although I don't think the specialisation is important.

With that in mind, to avoid confusing the `Ethereum` transaction message with the Connectors message sent through an Ethereum transaction, I suggest that we name this `MaxMsgBytes` with an appropriate `///` comment.

Alternative, I wonder if we can just use the `MaxEncodedLength` type of the `Message` type? That would make the code more resilient (since a change to the Message enum type wouldn't require changing this type here) reduce the sources of truth about the longest message byte length.

runtime/development/src/connectors.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// TODO(cdamian): Implement this for the connectors pallet.
+pub struct DummyInboundQueue {}
+
+impl InboundQueue for DummyInboundQueue {
+	type Message = ConnectorsMessage;
+	type Sender = Domain;
+
+	fn submit(_sender: Self::Sender, _msg: Self::Message) -> DispatchResult {
```
NunoAlexandre, line 49: Can't leave a comment in the trait itself but this on the Solidity side is called `handle` which I would also prefer to have here over `submit`

pallets/connectors-gateway/src/lib.rs
cdamian: Hmm, I think there's some confusion here. This should be the maximum size of the eth message that eventually gets deserialized into a connectors message.

NunoAlexandre: I don't think so? Following to the code, it is the `Connectors message`:

``` rust
               pub fn process_msg(
			origin: OriginFor<T>,
			msg: BoundedVec<u8, T::MaxEthMsgSize>, // HERE
		) -> DispatchResult {
			let domain_address = T::LocalOrigin::ensure_origin(origin)?;

			match domain_address {
				DomainAddress::EVM(_, _) => {
					DomainSubmitters::<T>::get(domain_address.domain())
						.iter()
						.find(|s| s.eq(&&domain_address))
						.ok_or(Error::<T>::UnknownSubmitter)?;

                                         // AND HERE
					let incoming_msg = T::Message::deserialize(&mut msg.as_slice())
						.map_err(|_| Error::<T>::EthereumMessageDecode)?; 
```

cdamian: My bad, I'm still confused as to how this would be called then. Wouldn't it make more sense for the caller to just pass a connectors message so that we can avoid deserializing here?

pallets/connectors-gateway/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: That doesn't state what it is trying to accomplish, so it's hard to judge. Regardless, this would mean that every user on any EVM address would need to be added as an allowed submitter, i.e, we would need to submit a Centrifuge Chain proposal to add that valid submitter's address before they would be able to use Centrifuge through Connectors. I don't think that's what we want. 

What we had in place already was the `knownConnectors` storage where we stored the origin / address of the known connector's contracts deployed on whatever EVM Domain; with that storage, we can verify that a given message came from Connectors and can therefore be trusted AFAIK. 

Currently on our Moonbeam POC, all messages sent from our Connector deployed in their EVM have the very contract's address as the origin of that message itself; So we don't have to whitelist every individual user who wants to use that contract, as we know it came from _that_ Connector, which we know at the Centrifuge chain level to be trusted.

Note as well that with this _submitters_ setup, users would still be able to send messages from the EVM side but they would just fail on the Centrifuge chain.

cc @offerijns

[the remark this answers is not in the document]
offerijns: Isn't that just going to be the address of the `Router` contract of the connectors deployment on every target chain? As in, we need to check somewhere that it actually was the connectors contract that sent a message over some transport layer to Centrifuge Chain.

At least, I would assume the origin here to be the EVM address that triggered the bridge message (basically, the `msg.sender` that sends a transaction in to the Axelar gateway, in Axelar's case; this `msg.sender` would then be the `Router`)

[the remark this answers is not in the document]
cdamian: > Isn't that just going to be the address of the Router contract of the connectors deployment on every target chain?

This is also what I thought it would be. AFAIK we need to have a way of whitelisting who gets to submit these messages.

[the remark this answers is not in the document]
offerijns: Yes that was definitely intentional because without that, migrating Connectors would be really hard. As you'd need to make sure there are no messages in transit before changing the address on the Cent Chain side.

pallets/connectors-gateway/src/lib.rs
NunoAlexandre: Yap :)

libs/traits/src/connectors.rs
```diff
@@ -21,19 +21,31 @@ pub trait Codec: Sized {
...
-	/// The connector message enum.
+	/// The message type that is processed.
 	type Message;
 
 	/// The destination this message should go to.
 	type Destination;
 
 	/// Submit a message to the outbound queue.
-	fn submit(
+	fn handle(
```
mustermeiszer, line 48: Sorry, to chip in here, but I don't like the naming of `handle`. I would assume that handle in most cases means a direct processing, where a queue in most cases firstly receives something and then acts when it has capacity for it. But no blocker.

libs/traits/src/connectors.rs
```diff
@@ -49,5 +61,5 @@ pub trait InboundQueue {
 	type Message;
 
 	/// Submit a message to the inbound queue.
-	fn submit(sender: Self::Sender, msg: Self::Message) -> DispatchResult;
+	fn handle(sender: Self::Sender, msg: Self::Message) -> DispatchResult;
```
mustermeiszer, line 64: Same as above.

pallets/connectors-gateway/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Do we need any more clarification here or are we all on the same page? ^^

mustermeiszer (approved):
Looks nice and clean. Would maybe really rename the moonbeam router but not a must have for me.

pallets/connectors-gateway/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: Makes all sense :) in this case, my suggestion is that we merge `knownConnectors` and `domainSubmitters`. I find "domain submitters" too broad and ambiguous tbh, as it sparked this whole discussion. We can basically keep the storage definition of `domainSubmitters` but rename it to `connectorsAllowlist`; which leaves us with:

- Drop `knownConnectors`
- Rename `MaxSubmittersPerDomain` to `MaxConnectorsPerDomain`
- Rename `domainSubmitters` to `connectorsAllowlist`, keeping the same definition:

```
StorageMap<
	_,
	Blake2_128Concat,
	Domain,
	BoundedVec<DomainAddress, T::MaxConnectorsPerDomain>,
	ValueQuery,
>;
```

NunoAlexandre (commented):
Almost there 🧗 

Last points are:

- 1. https://github.com/centrifuge/centrifuge-chain/pull/1341#discussion_r1194727497
- 2. On naming:
  - Name the functions handling incoming messages as `handle`
    (i.e, the Connectors pallet is _handling_ an incoming message, not _submitting_ it; I get the idea behind the gateway _submitting_ it to the Connector but I don't think that naming to be particularly helpful. On Solidity those are also called `handle` and I'd love us to keep that naming here)
  - Name the functions _sending_ a message to a remote Connector as `send`
    (now these are also called `submit`


So basically:

``` rust
/// The trait required for processing outbound connectors messages.
pub trait OutboundQueue {
	/// The sender type of the outgoing message
	type Sender;
	/// The message type that is processed.
	type Message;
	/// The destination this message should go to
	type Destination;

	/// Send a message to the the destination
	fn send(
		destination: Self::Destination,
		sender: Self::Sender,
		msg: Self::Message,
	) -> DispatchResult;
}

/// A trait that handles incoming messages
pub trait InboundQueue {
	/// The sender type of the incoming message
	type Sender;
	/// The incoming message
	type Message;

	/// Handle an incoming message
	fn handle(sender: Self::Sender, msg: Self::Message) -> DispatchResult;
}

```

mustermeiszer:
But as the naming of the trait states it is a queue. Not a sender and not a handler. I am not to find to base our argumentation on the naming of the Solidity code as this is a different architecture, different types, etc. Types are a different topic if the have a corresponding representation in every domain.

wischli (commented):
Great work dude! I am a bit late to the party but wanted to raise three nits. Feel free to ignore them (for now) 😅

NunoAlexandre (approved):
Awesome work @cdamian 🙌  🚀 

Also on the patience 😜

wischli (approved):
Thanks for applying my comment!

Issue #1246, opened 2023-03-10
Remove concept of `Domain` in `pallet-rewards`


wischli:
### Which part of the code is the issue addressing?

* [ ] Addressing a bug
* [x] Refactoring
* [ ] Feature

### Description

* Remove concept of `Domain` from `pallet-rewards`
* Add documentation to force a unique instance of `pallet-rewards` per reward system, e.g. `pallet-block-rewards` and `pallet-liquidity-rewards`

### Research/based on

https://github.com/centrifuge/centrifuge-chain/pull/1198#discussion_r1132164762

### How will this affect the code base

* Reduces complexity
* Improves readability

### What are forseen obstacles or hurdles to overcome?

* Investigate potential currency collisions using the same `pallet-rewards`
* We need to enforce separate instances of `pallet-rewards` in the future if we add more consumers

lemunozm:
@mikiquantum WDYT about this? If you agree, I'll go with this first before adding rewards to `altair`/`centrifuge` runtime to avoid migrations later.

mikiquantum:
So is the general idea that the rewards pallet doesnt know anything about the domains is dealing with underneath and since it the access to its storage is merely internal, delegating this identification to the `block-rewards` & `liquidity-rewards` pallets, by having a Config trait implemented on each defining its `Domain` ?
Then we can use the same instance of the pallet_rewards for every parent pallet that uses the same mechanisms. 

I think this change makes sense to me.

lemunozm:
> Then we can use the same instance of the pallet_rewards for every parent pallet that uses the same mechanisms.

Yes, and this is what we want to avoid. It's true that domains remove the `CurrencyId` collision when two pallets use the same instance of `pallet-rewards`, but we still need to differentiate between groups. So we would need to add a domain to a group. 

What I propose is to use *different* instances of `pallet-rewards` *even* if each pallet uses the same mechanisms:

PRO:
- We do not need to deal with `Domain`s which is not trivial to understand well.
- We do not need to implement a `Domain` for groups that makes it complex

CONS:
- We need to create several instances of `pallet-rewards` in case we have 2 pallets that use the exact mechanism. But I do not think it's really a problem. The storage used should be almost the same.

mikiquantum:
[...]

PR #1342, opened 2023-05-11
feat: add block rewards to altair and centrifuge runtimes

wischli:
# Description

* Adds block rewards implementation introduced in #1198 to the Altair and Centrifuge runtimes
* ~~Reduces kick threshold for collators to 2 hours from 6 hours~~  UPDATE: In order for this to become effective, we would have to change the session duration as kicking is [only applied on session transitions](https://github.com/paritytech/cumulus/blob/8ec0778fe386a272adb2c862cc554f880a28c858/pallets/collator-selection/src/lib.rs#L511)
  * Automatically removes collators which fail to produce a block within that threshold from the set of block authors
* Fixes issue of of unfunded sovereign `pallet_rewards` accounts which receive rewards
  * Without the fix: 
    * Rewards are not be minted into that account if they are below the `ExistentialDeposit` 
    * Rewards can not be claimed by all parties as this would leave the account without balance
  * Adds funding of 1 `CFG` or `AIR` to the `pallet_rewards` sovereign account via
    * `GenesisConfig` for ephemeral networks 
    * `FundExistentialDeposit` migration for running networks
* Adds required migration for `SessionKeys`
* Removes migrations for dev runtime as a result of the agreement made at the offsite 

### Altair

Based on [this RFC](https://gov.centrifuge.io/t/adjustment-of-altair-collator-rewards/4844), Altair rewards are currently `12000 AIR` per month. Looking at the time calculation of the Centrifuge block rewards proposal, a year is counted as 365 days. Thus, on average, a month equals `30.4166666667` days.

* Collator rewards: `12000 AIR per month / 30.417 days per month / 4 sessions per day = 98.63 AIR = 98630 MILLI_AIR`
* Total rewards: In Altair, ~~we transfer the rewards from the treasury and do not want to mint extra rewards into the treasury as the latter is already sufficiently funded~~. 
  * UPDATE 2023-05-15: We actually want to mint the rewards here as well to be closer to Centrifuge. There will be a follow-up RFC to burn some of the treasury.
  * We could set this to infinity and it would have no effect because `pallet_block_rewards::Config::Beneficiary` is set to `()`. We only need to ensure that `TotalRewards` will always be at least `CollatorReward * num_collators`. Of course, `TotalRewards` can also be adjusted via `RootOrHalfOfCouncil`. However, I do not see any reason to set it to `CollatorRewards * MaxCandidates` for now as the multiplication with `MaxCandidates` just serves as a max bound. The remainder of `TotalRewards - CollatorRewards` will be **dropped** in Altair.

### Centrifuge

In Centrifuge we mint rewards both for the collators as well as the treasury. These values have been pre-approved by https://gov.centrifuge.io/t/rfc-add-a-block-reward-and-improve-collators-cycle/4766

* Collator rewards = `16.65 CFG per epoch / 2 sessions per epoch = 8.325 CFG = 8_325 * MILLI_CFG`
* Total rewards = `20,096 CFG per epoch / 2 sessions per epoch = 10048 CFG`

## About the migrations

Unfortunately, adding block rewards requires three migrations to Altair and Centrifuge. Here's a short summary:

### Rewards

While testing block rewards on a local chain environment, I stumbled into a minor issue which needs to be fixed: When claiming rewards, we [currently transfer](https://github.com/centrifuge/centrifuge-chain/blob/bd4a96bf6e9bb6606f6b285cfe0493b7415a8734/pallets/rewards/src/lib.rs#L370) from the pallet id of the  base `pallet_rewards` instance to the recipient with `keep_alive` enabled. This blocks any reward claims which would reduce the balance of the pallet's sovereign account below the existential deposit (ED). The `keep_alive` should stay enabled. Thus, we need to fund the `pallet_rewards` sovereign account (at least) with the ED which is done by the migration [added here](https://github.com/centrifuge/centrifuge-chain/pull/1342/files#diff-0946c08ffa10c8e8b0c1113bc058b68fda4c798a1724cad4c5b7e6241ce6cd6cR55).

Without this migration, not all rewards can be claimed from the reward recipients.

### BlockRewards

This migration was introduced in #1198. It sets up the required block rewards storage for our current collators. 

Without this migration, none of the collators would be eligible to claim rewards.

### SessionKeys

Without this migration, all of our collators would have to update their session keys manually. E.g., none of the collators would actually be recognized by the session transition inside `pallet_block_rewards` because the `block_rewards` entry of `SessionKeys` is unset. 

## TODO

- [x] Migration for `SessionKeys`: Add new `block_rewards` entry
- [x] In separate PR: https://github.com/centrifuge/centrifuge-chain/issues/1246 --> https://github.com/centrifuge/centrifuge-chain/pull/1344
- [x] Maybe: Implement `RewardsApi` for Tuple --> Can be done later when `LiquidityRewards` are pushed to prod

# Checklist:


pallets/rewards/src/lib.rs
```diff
@@ -138,6 +141,46 @@ pub mod pallet {
...
+		fn build(&self) {
+			if !self.amount.is_zero() {
+				T::Currency::mint_into(
+					self.currency_id,
+					&T::PalletId::get().into_account_truncating(),
+					self.amount,
+				)
+				.map_err(|_| log::error!("Failed to mint ED for sovereign pallet account",))
+				.ok();
+			}
```
lemunozm, line 180: Out of curiosity, what would happen if we mint zero? Is it an error?

pallets/rewards/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
...
+impl<T, I, Currency, ED> OnRuntimeUpgrade for FundExistentialDeposit<T, I, Currency, ED>
+where
+	T: frame_system::Config + Config<I>,
+	I: 'static,
+	Currency: Get<<T as Config<I>>::CurrencyId>,
+	ED: Get<BalanceOf<T, I>>,
+{
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
+		assert!(!pallet_is_funded::<T, I, Currency, ED>());
```
lemunozm, line 49: Should it be an `ensure!` instead in order to avoid panicking the whole try runtime execution?

pallets/rewards/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
...
+			log::info!(
+				"💶 Rewards: ED funding for sovereign pallet account not required anymore. This probably should be removed"
+			);
+			T::DbWeight::get().reads_writes(1, 0)
+		}
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(_pre_state: Vec<u8>) -> Result<(), &'static str> {
+		assert!(pallet_is_funded::<T, I, Currency, ED>());
```
lemunozm, line 77: Same here

pallets/rewards/src/mock.rs
```diff
@@ -94,6 +94,12 @@ pub enum CurrencyId {
 	M,
 }
 
+impl Default for CurrencyId {
+	fn default() -> Self {
+		CurrencyId::Reward
+	}
+}
```
lemunozm, line 101: There is also a [default for enums](https://users.rust-lang.org/t/derive-default-for-enum-non-only-struct/44046). Anyways, no strong opinion here 👍🏻

runtime/altair/src/lib.rs
```diff
@@ -1447,6 +1527,8 @@ construct_runtime!(
 		InterestAccrual: pallet_interest_accrual::{Pallet, Storage, Event<T>, Config<T>} = 101,
 		Investments: pallet_investments::{Pallet, Call, Storage, Event<T>} = 102,
 		PoolRegistry: pallet_pool_registry::{Pallet, Call, Storage, Event<T>} = 103,
+		BlockRewardsBase: pallet_rewards::<Instance1>::{Pallet, Storage, Event<T>, Config<T>} = 104,
```
lemunozm, line 1530: Not sure if we want to use `Instance2` here to be according to the development Instances, if not, later, it can be a mess. WDYT?

pallets/rewards/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
+// Copyright 2021 Centrifuge Foundation (centrifuge.io).
```
lemunozm, line 1: I've found quite useful to place migrations under a file with a custom name (as `v1.rs` or so), later if in a new PR we remove an old migration and create a new one, there is not an awful git diff, and it becomes easy to review. Not sure if I explain the situation well 😅

runtime/altair/src/weights/pallet_block_rewards.rs
```diff
@@ -0,0 +1,58 @@
...
+// pallet
+// --chain=altair_local
+// --steps=50
+// --repeat=20
+// --pallet=pallet_block_rewards
+// --extrinsic=*
+// --execution=wasm
+// --wasm-execution=compiled
+// --heap-pages=4096
+// --output=/tmp/runtime/development/src/weights/pallet_block_rewards.rs
```
lemunozm, line 21: It should be `altair` here, not sure if it was badly generated or it's till not generated.

runtime/development/src/lib.rs
```diff
@@ -1824,11 +1824,11 @@ construct_runtime!(
 		Nfts: pallet_nft::{Pallet, Call, Event<T>} = 103,
 		Keystore: pallet_keystore::{Pallet, Call, Storage, Event<T>} = 104,
 		Investments: pallet_investments::{Pallet, Call, Storage, Event<T>} = 105,
-		Rewards: pallet_rewards::<Instance1>::{Pallet, Storage, Event<T>} = 106,
+		Rewards: pallet_rewards::<Instance1>::{Pallet, Storage, Event<T>, Config<T>} = 106,
```
lemunozm, line 1827: Out of curiosity, what capabilities add this `Config<T>` here?, something related to the Genesis?

runtime/centrifuge/src/lib.rs
```diff
@@ -1143,7 +1146,75 @@ impl pallet_collator_selection::Config for Runtime {
...
+	#[derive(scale_info::TypeInfo, Debug, PartialEq, Eq, Clone)]
+	pub const MaxChangesPerEpoch: u32 = 50;
+	pub const RewardsPalletId: PalletId = PalletId(*b"d/reward");
+	pub const RewardCurrency: CurrencyId = CurrencyId::Native;
+}
+
+impl pallet_rewards::Config<pallet_rewards::Instance1> for Runtime {
+	type Currency = Tokens;
+	type CurrencyId = CurrencyId;
+	type DomainId = RewardDomain;
```
lemunozm, line 1177: Regarding the `Domain` change, what's the idea? Merge this, and later add a migration for removing `Domain`s?

runtime/centrifuge/src/lib.rs
```diff
@@ -1614,6 +1685,8 @@ construct_runtime!(
 		CrowdloanReward: pallet_crowdloan_reward::{Pallet, Call, Storage, Event<T>} = 97,
 		Tokens: pallet_restricted_tokens::{Pallet, Call, Event<T>} = 98,
 		CollatorAllowlist: pallet_collator_allowlist::{Pallet, Call, Storage, Config<T>, Event<T>} = 99,
+		BlockRewardsBase: pallet_rewards::<Instance1>::{Pallet, Storage, Event<T>, Config<T>} = 100,
```
lemunozm, line 1688: Same here to `Instance2`

runtime/centrifuge/src/lib.rs
```diff
@@ -1666,7 +1739,24 @@ pub type UncheckedExtrinsic =
 /// Extrinsic type that has already been checked.
 pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, RuntimeCall, SignedExtra>;
 
-type UpgradeCentrifuge1019 = pallet_loans_ref::migrations::v1::Migration<Runtime>;
+parameter_types! {
+	// = 16.65 CFG per epoch (12h)
+	// https://gov.centrifuge.io/t/rfc-add-a-block-reward-and-improve-collators-cycle/4766
+	pub const CollatorRewards: Balance = 8_325 * MILLI_CFG;
+	// = 20,096 CFG per epoch (12h)
+	pub const TotalRewards: Balance = 10_048 * CFG;
```
lemunozm, line 1747: Why here are we not multiplying this by `MaxCandidates` as in Altair?

src/chain_spec.rs
```diff
@@ -876,6 +882,19 @@ fn centrifuge_genesis(
...
+				.iter()
+				.cloned()
+				.map(|(acc, _)| acc)
+				.collect(),
+			collator_reward: 6_000 * AIR,
+			total_reward: 6_000 * AIR * 100,
+		},
+		block_rewards_base: centrifuge_runtime::BlockRewardsBaseConfig {
+			currency_id: CurrencyId::Native,
+			amount: 1 * AIR,
```
lemunozm, line 896: Should it be the `ED` constant instead? (It applies to all chains)

lemunozm (commented):
Very clean PR @wischli. I really like it! 🚀 
I left some comments, but most of them are questions

pallets/rewards/src/lib.rs
branan: If the account would be below the minimum balance after minting it will error

lemunozm: Then, should it be `if self.amount.is_zero() < ED {` or similar?

NunoAlexandre:
@wischli looking good already but I will give it a proper review once the ToDos are addressed 👍

runtime/centrifuge/src/lib.rs
wischli: In Altair, we transfer the rewards from the treasury and do not want to mint extra rewards into the treasury as the latter is already sufficiently funded. So the multiplication with `MaxCandidates` serves there as a max bound. The remainder of `TotalRewards - CollatorRewards` will be dropped in Altair.

However, in Centrifuge we mint rewards both for the collators as well as the treasury. These values have been pre-approved by https://gov.centrifuge.io/t/rfc-add-a-block-reward-and-improve-collators-cycle/4766

runtime/development/src/lib.rs
wischli: This is required for the `GenesisConfig` builder applied in `[src/chain_spec.rs](https://github.com/centrifuge/centrifuge-chain/pull/1342/files#diff-8eaf99be7c59a06fae89dfafbab2ec3873044be161513bb28863e3c99af34d69R1146)`. Without this setting, we would not expose the Config and thus could not do any genesis setting.

runtime/altair/src/weights/pallet_block_rewards.rs
wischli: These weights are temporary anyways. Before the release, we will rerun all benchmarks.

pallets/rewards/src/migrations.rs
wischli: I get what you are saying but I am not sure I support this. Wouldn't it suffice to push the code into a `mod` inside `migrations.rs` and still guarantee an easy-to-follow git diff in the future?

This particular migration though should actually stick around because we need to run it for every `rewards` instance we spawn.

lemunozm: > Wouldn't it suffice to push the code into a mod inside migrations.rs and still guarantee an easy-to-follow git diff in the future?

If you do not remove the previous migration when adding the new one, the diff will be ok, but if you remove it, git will think you're instead modifying the previous one.

pallets/rewards/src/migrations.rs
wischli: Will do. On a note, in Substrate there are migrations with `assert` as well. Luckily, in the future `try-runtime` does not bail on the first error anymore[ according to this PR](https://github.com/paritytech/substrate/pull/14083)  🚀

runtime/altair/src/lib.rs
lemunozm: Ok, I see your point.

> So I would rather propose to change it in the dev runtime instead.

Would it need a migration?

pallets/rewards/src/lib.rs
wischli: > If the account would be below the minimum balance after minting it will error

AFAICS, it would actually only error if we the account has reached the [max consumer threshold pre minting](https://github.com/paritytech/substrate/blob/6fa7fe1326ecaab9921c2c3888530ad679cfbb87/frame/assets/src/functions.rs#L79). Or if we [cannot have more accounts on the chain](https://github.com/paritytech/substrate/blob😅/6fa7fe1326ecaab9921c2c3888530ad679cfbb87/frame/assets/src/functions.rs#L71-L72)

runtime/centrifuge/src/lib.rs
wischli: I am more or less done with it. I wanted to open a separate PR with base set to this branch.

pallets/rewards/src/lib.rs
lemunozm: Almost never 🤣

lemunozm: > It would just return Ok(())

Then, is the `if !self.amount.is_zero() {` really needed?

pallets/rewards/src/migrations.rs
wischli: But does that also apply if we keep them in separate scopes via `mod v1 { ... } mod v2 { ... }` inside the same file?

runtime/altair/src/lib.rs
wischli: It would not! During the offsite, we decided there should not be migrations for `dev`.

pallets/rewards/src/lib.rs
wischli: Fair point 😅 It's not.

pallets/rewards/src/migrations.rs
lemunozm: git doesn't know about "scopes", and it will think that you've changed `v1` by `v2`, and will mix every line inside of those blocks.

lemunozm: Anyways, nothing mandatory here at all

runtime/altair/src/lib.rs
wischli: It doesn't make sense to write migrations for a runtime which is only supposed to be used in ephemeral networks 🙃 If Demo was still based on `dev`, we would still be in that dilemma. That was the main motivation to base Demo on `altair` runtime.

lemunozm: Ok, I understand 👍🏻. Agree totally with it

pallets/rewards/src/migrations.rs
wischli: I will follow your proposal! More readable git diffs are definitely worth the little extra work :)

lemunozm: To avoid redundant files for the mod.rs, I'm adding them like [this](https://github.com/centrifuge/centrifuge-chain/blob/bd4a96bf6e9bb6606f6b285cfe0493b7415a8734/pallets/loans-ref/src/lib.rs#L40) (in case it can be useful)

pallets/rewards/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
...
+	Currency: Get<<T as Config<I>>::CurrencyId>,
+	ED: Get<BalanceOf<T, I>>,
+{
+	let balance = <<T as Config<I>>::Currency as Inspect<T::AccountId>>::balance(
+		Currency::get(),
+		&<T as Config<I>>::PalletId::get().into_account_truncating(),
+	);
+	balance >= ED::get()
+}
+
```
mustermeiszer, line 39: Could it be possible that an account already has funding from rewards and we still would need to fund 1 times ED in case all collators claim?

runtime/integration-tests/src/rewards/block/env.rs
mustermeiszer, line 1: Nice!

src/chain_spec.rs
```diff
@@ -876,6 +882,19 @@ fn centrifuge_genesis(
 		},
 		treasury: Default::default(),
 		interest_accrual: Default::default(),
+		block_rewards: centrifuge_runtime::BlockRewardsConfig {
+			collators: initial_authorities
+				.iter()
+				.cloned()
+				.map(|(acc, _)| acc)
+				.collect(),
+			collator_reward: 6_000 * AIR,
```
thea-leake, line 891: Should `AIR` here be `CFG`?  It looks like AIR is set to `CFG` in altair consts, but wondering if it would be better/clearer to use `CFG` given that this is for cent runtime.   Not blocking at all, just curious :).

pallets/rewards/src/migrations.rs
wischli: Funding the pallet's sovereign account once seems to be sufficient based on the unit tests and my local e2e testing. When we update the rewards, they are associated to a group. This ensures that a group cannot consume more than it should. 

However, before the migration, the balance of the pallet's sovereign account only consisted of the block rewards. So it was impossible to claim all rewards.

src/chain_spec.rs
wischli: Great catch! Absolutely. Thanks!

thea-leake (approved):
LGTM! 🚀

pallets/rewards/src/migrations/new_instance.rs
```diff
@@ -0,0 +1,82 @@
...
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
+		assert!(!pallet_is_funded::<T, I, Currency, ED>());
+
+		log::info!("💶 Rewards: Pre funding ED checks successful");
+		Ok(vec![])
+	}
+
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		if !pallet_is_funded::<T, I, Currency, ED>() {
```
mustermeiszer, line 56: Isn't `LiquidityRewards` instance funded but not with the `ED` as we want to? As we know no intance has been funden, shouldn't we simply mint 1 ED into the pallets account without this check?

mustermeiszer (commented):
Only one question regarding the `ED` minting in the migration, where I am not sure, it is working as we intend it to. But might be due to my improper understanding. If cleared, will approve.

runtime/altair/src/lib.rs
```diff
@@ -1106,6 +1110,82 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
...
+	Block,
+}
+
+frame_support::parameter_types! {
+	// BlockRewards have exactly one group and currency
+	#[derive(scale_info::TypeInfo)]
+	pub const SingleCurrencyMovement: u32 = 1;
+	#[derive(scale_info::TypeInfo, Debug, PartialEq, Eq, Clone)]
+	pub const MaxChangesPerEpoch: u32 = 50;
+	pub const RewardsPalletId: PalletId = PalletId(*b"d/reward");
```
NunoAlexandre, line 1131: Out of curiosity, why `d/` in `d/reward`?

runtime/altair/src/lib.rs
```diff
@@ -1501,6 +1604,101 @@ pub type Executive = frame_executive::Executive<
...
+	impl_opaque_keys! {
+		pub struct OldSessionKeys {
+			pub aura: Aura,
+		}
+	}
+
+	#[storage_alias]
+	type QueuedKeys<T: pallet_session::Config> = StorageValue<
+		pallet_session::Pallet<T>,
+		// hex_literal::hex!("cec5070d609dd3497f72bde07fc96ba0"),
```
NunoAlexandre, line 1622: I am not sure I get this comment; is this what the storage value prefix encodes to or is this the only value we expect to exist when running the migration?

runtime/altair/src/lib.rs
```diff
@@ -1501,6 +1604,101 @@ pub type Executive = frame_executive::Executive<
...
+			Ok(queued.encode())
+		}
+
+		fn on_runtime_upgrade() -> frame_support::weights::Weight {
+			use sp_runtime::SaturatedConversion;
+
+			// We must not bump the storage version of pallet_session to stay in sync with
+			// Substrate. As this migration is tightly coupled to the initialization of
+			// BlockRewards, it should be sane to trigger the session migration based on the
+			// BlockRewards storage version.
```
NunoAlexandre, line 1646: Thanks for this comment 👍

runtime/altair/src/lib.rs
```diff
@@ -1501,6 +1604,101 @@ pub type Executive = frame_executive::Executive<
...
+					&queued[i].0
+				);
+				assert!(
+					&queued[i].1.aura == &pre_state[i].1.aura,
+					"Failed to match old aura key at index {}. old {:?} vs {:?} new",
+					i,
+					&pre_state[i].1.aura,
+					&queued[i].1.aura,
+				);
+			}
```
NunoAlexandre, line 1695: Can we do a `forEach` of sorts instead of `for i n 0.. x.len()`? I find that easier to read and safer as well. Since we want to compare the `queued` with the  `pre_state`, we can just `zip` the two, forEach on the zipped tuples, and go with that :)

src/chain_spec.rs
```diff
@@ -941,6 +960,19 @@ fn altair_genesis(
...
+				.iter()
+				.cloned()
+				.map(|(acc, _)| acc)
+				.collect(),
+			collator_reward: 98_630 * MILLI_AIR,
+			total_reward: 98_630 * MILLI_AIR * 100,
+		},
+		block_rewards_base: altair_runtime::BlockRewardsBaseConfig {
+			currency_id: CurrencyId::Native,
+			amount: 1 * MICRO_AIR,
```
NunoAlexandre, line 974: Is this correct that the reward for a block  is 1 _micro_ AIR?

runtime/altair/src/lib.rs
```diff
@@ -1106,6 +1110,82 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
...
+
+#[derive(Clone, Copy, PartialEq, Eq, Encode, Decode, TypeInfo, MaxEncodedLen, RuntimeDebug)]
+#[cfg_attr(
+	feature = "std",
+	derive(frame_support::Serialize, frame_support::Deserialize)
+)]
+pub enum RewardDomain {
+	Liquidity,
+	Block,
+}
```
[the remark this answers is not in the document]
wischli: Definitely. It is removed in #1344 though.

runtime/altair/src/lib.rs
wischli: Sorry, this hex literal comment should be removed. It was part of my struggle to get this migration running.

pallets/rewards/src/migrations/new_instance.rs
mustermeiszer: cc @offerijns

runtime/altair/src/lib.rs
mustermeiszer: The ID please into the `types` crate and to ids.

wischli: Shouldn't it rather be a primitives constant?

src/chain_spec.rs
wischli: That's not the block reward but the funding of the `pallet_rewards::Instance` sovereign account which holds all the rewards until they are claimed. As long as this account does not have the ED, not all rewards can be claimed at the same time as it would kill this account.

runtime/altair/src/lib.rs
mustermeiszer: We already have an ids mod in the types. I am not sure why I did not move it to the primitives. I guess my reasoning was that it is not a primitive to the logic of our chain. But no feelings where it lives, just want them grouped so that we notice when we change them.

pallets/rewards/src/migrations/new_instance.rs
wischli: Sorry, my mistake. Liquidity rewards should be in native [according to the spec's goal](https://centrifuge.hackmd.io/j_7liWuyTDu7w9Qa-jFPZw#-Goal).

So we must run this migration (which I would have advised for anyways), once we add them to prod runtimes.

offerijns: Yeah main goal is receiving CFG tokens as reward for staking tranche tokens.

wischli:
There was a last-minute change to the Altair configuration: Lucas and Cassidy [proposed to actually mint AIR](https://centrifugedao.slack.com/archives/C04HAJHFL4A/p1684157298039919?thread_ts=1683791457.639279&cid=C04HAJHFL4A) instead of transferring from the treasury. There will be a follow-up proposal to burn some/all of the Altair treasury. The main reason is that we want to be more close to Centrifuge.

lemunozm (dismissed):
LGTM!

pallets/block-rewards/src/lib.rs
```diff
@@ -258,12 +251,9 @@ pub mod pallet {
...
 	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
 		fn build(&self) {
-			T::Rewards::attach_currency(
-				(T::Domain::get(), T::StakeCurrencyId::get()),
-				T::StakeGroupId::get(),
-			)
-			.map_err(|e| log::error!("Failed to attach currency to collator group: {:?}", e))
-			.ok();
+			T::Rewards::attach_currency(T::StakeCurrencyId::get(), T::StakeGroupId::get())
+				.map_err(|e| log::error!("Failed to attach currency to collator group: {:?}", e))
```
mustermeiszer, line 255: Genesis builds usually `panic!`

pallets/rewards/src/lib.rs
```diff
@@ -138,6 +142,44 @@ pub mod pallet {
...
+	where
+		BalanceOf<T, I>: MaybeSerializeDeserialize,
+	{
+		fn build(&self) {
+			T::Currency::mint_into(
+				self.currency_id,
+				&T::PalletId::get().into_account_truncating(),
+				self.amount,
+			)
+			.map_err(|_| log::error!("💶 Rewards: Failed to mint ED for sovereign pallet account",))
```
mustermeiszer, line 178: Same here.

runtime/centrifuge/src/lib.rs
```diff
@@ -1143,7 +1147,62 @@ impl pallet_collator_selection::Config for Runtime {
...
+	pub const BlockRewardCurrency: CurrencyId = CurrencyId::Staking(StakingCurrency::BlockRewards);
+	pub const StakeAmount: Balance = cfg_types::consts::rewards::DEFAULT_COLLATOR_STAKE;
+	pub const CollatorGroupId: u32 = cfg_types::ids::COLLATOR_GROUP_ID;
+}
+
+impl pallet_block_rewards::Config for Runtime {
+	type AdminOrigin = EnsureRootOr<HalfOfCouncil>;
+	type AuthorityId = AuraId;
+	type Balance = Balance;
+	// Must not set change this as long as we want to mint rewards into the treasury
```
mustermeiszer, line 1193:
```suggestion
	// Must not change this as long as we want to mint rewards into the treasury
```

src/chain_spec.rs
mustermeiszer: Just out of interest, we could use the ED defined in the runtime instead of hardcoding, right?

mustermeiszer (approved):
Would prefer to panic instead of logging an error in the genesis build. This is at least what Substrate is doing, IIRC and easier to spot. But no blocker. 

Looks clean and nice!

pallets/block-rewards/src/lib.rs
wischli: [1fb344e](https://github.com/centrifuge/centrifuge-chain/pull/1342/commits/1fb344edd5f19d9c5540aa9b79012025bab50373)

NunoAlexandre (approved):
Solid 🤘

PR #1343, opened 2023-05-11
refactor: Clean up `CurrencyId`

### LINKED ISSUE #826: refactor: `CurrencyId`
## What

Our current `CurrencyId` type is defined as follows:

``` rust
pub enum CurrencyId {
	// The Native token, representing AIR in Altair and CFG in Centrifuge.
	Native,

	// A Tranche token
	Tranche(u64, [u8; 16]),

	/// Kusama KSM
	KSM,

	/// Karura Dollar
	KUSD,

	/// Acala Dollar
	/// Note: KUSD and AUSD will be merged into a single token, AUSD.
	AUSD,

	Permissioned(PermissionedCurrency),
}
```

This type is then used as the currency id type for other pallets, namely `restricted_tokens`, `orml_tokens`, and `xTokens`. I.e, each variant is encoded and used as a key identifying an asset on those pallets. With the exception of `xTokens`, that encoded key is stored on chain, for example when storing the balance of an account for a specific asset.

## Requirements

1. Allow for identifying assets stored in `asset-registry` (see #825)
2. Allow for specifying or categorise assets based on some properties
   - Whether a token is _permissioned_, such as the `Tranche` token
   - Whether a token should be `Mintable` - an upcoming use case
3. Support for specific native currencies for the different runtimes, i.e, `AIR`, `CFG`, `DEVEL`, instead of just a single `Native`
    - This will become specifically important when cross-network XCM transfers become possible
    - Note: this would require a migration

Requirement 1. is mandatory and urgent; 
Requirement 2. is an important one but not yet urgent;
Requirement 3. is a nice to have but can wait.

## Proposal

To prevent us to dive into a gigantic PR, I propose that we break this in phases and focus on deliverables for the timelines we aim to respect.

### Phase 1

In phase 1, we focus on delivering on Requirement 1 and lay the foundation for Requirement 2.

**Steps**

- Add a `ForeignAsset` variant to `CurrencyId`
- Define a `CustomMetadata` type used in `asset-registry` that could capture whether a token is:
  - Permissioned
  - Mintable
  - Optional XCM-useful fields (for instance, `fee_per_second`)

This allows us to:

- Register and thus support new tokens to be XCM-transferrable, namely USDC
- Register other tokens for testing purposes with the Permissioned & Mintable trait flags allowed

### Phase 2

- Drop the `CurrencyId::Permissioned` variant and register those tokens in `asset-registry` with the correct `Permissioned` & `Mintable` flags instead
- Possibly extend the `CustomMetdata` type with more useful flags to categorise tokens more gradually

### Phase 3

- Deprecate the `CurrencyId::Native` variant and introduce at least `Currency::AIR` and `Currency::CFG`
   - Make `CurrencyId::AIR` the native token in `Altair` and `CurrencyId::CFG` the native token in `Centrifuge`
   - Write migration code to migrate all data associated with `Native` to be associated to `AIR` and `CFG` in `Altair` and `Centrifuge`, respectively 

Additionally, we can also care to:
 - Drop the harcoded variants we have for `KSM, `KUSD`, and `AUSD` to have them as registered assets in the `asset-registry` and merge `KUSD` as `AUSD`
       - Write migration code to migrate all data associated with said variants to be associated with the respective `CurrencyId::RegisteredToken(<id>)` variant.

NunoAlexandre:
Closes #826 

### Description

`CurrencyId` currently two legacy variants: `CurrencyId::AUSD` and `CurrencyId::KSM` that we want to drop in favour of using the `CurrencyId::ForeignAsset` variant class that was introduced with the purpose of supporting new foreign assets without the need of code changes and therefore runtime upgrades.

To drop those legacy variants, we have registered those same assets as `CurrencyId::ForeignAssets` in the `AssetRegistry`, both for Centrifuge and Altair. This allows us to 1) continue XCM support for those assets and 2) have a `CurrencyId` value to migrate all the balance data to.

### Changes

- [x] Drop `CurrencyId::AUSD` and `CurrencyId::KSM`
- [x] Altair: Write a migration to move all balances under `CurrencyId::AUSD` to the respective `ForeignAsset` id

## Notes

As state below, the only migrations we need to handle are moving AUSD tokens on Altair. Unfortunately, we can't drop the `CurrencyId::AUSD` variant right away in this PR because we need to have it in order to call the `burn_from` function of `orml_tokens`; i.e., unlike calling `translate` on a storage item which allows for old types to be decoded. Once this migration has run, we can drop that variant.

Also note that I don't check for a specific runtime version to decided to run the migration or not since this migration is idempotent.

### DIFF
--- a/runtime/altair/src/migrations.rs
+++ b/runtime/altair/src/migrations.rs
@@ -56,7 +56,7 @@ mod asset_registry {
 
 	impl OnRuntimeUpgrade for CrossChainTransferabilityMigration {
 		fn on_runtime_upgrade() -> Weight {
-			if VERSION.spec_version != 1028 {
+			if VERSION.spec_version > 1028 {
 				return Weight::zero();
 			}
 


NunoAlexandre:
@mustermeiszer @wischli I am looking at this again and I am trying to come up with a safe and sensible way of migrating all the storage items of `orml_tokens` under `CurrencyId::AUSD` to the new corresponding `CurrencyId::ForeignAsset` id.

The `orml_tokens` pallet has the following 4 storage items:

``` rust
/// The total issuance of a token type.
#[pallet::storage]
#[pallet::getter(fn total_issuance)]
pub type TotalIssuance<T: Config> = StorageMap<_, Twox64Concat, T::CurrencyId, T::Balance, ValueQuery>;

/// Any liquidity locks of a token type under an account.
/// NOTE: Should only be accessed when setting, changing and freeing a lock.
#[pallet::storage]
#[pallet::getter(fn locks)]
pub type Locks<T: Config> = StorageDoubleMap<
		_,
		Blake2_128Concat,
		T::AccountId,
		Twox64Concat,
		T::CurrencyId,
		BoundedVec<BalanceLock<T::Balance>, T::MaxLocks>,
		ValueQuery,
	>;

/// The balance of a token type under an account.
///
/// NOTE: If the total is ever zero, decrease account ref account.
///
/// NOTE: This is only used in the case that this module is used to store
/// balances.
#[pallet::storage]
#[pallet::getter(fn accounts)]
pub type Accounts<T: Config> = StorageDoubleMap<
		_,
		Blake2_128Concat,
		T::AccountId,
		Twox64Concat,
		T::CurrencyId,
		AccountData<T::Balance>,
		ValueQuery,
	>;

/// Named reserves on some account balances.
#[pallet::storage]
#[pallet::getter(fn reserves)]
pub type Reserves<T: Config> = StorageDoubleMap<
		_,
		Blake2_128Concat,
		T::AccountId,
		Twox64Concat,
		T::CurrencyId,
		BoundedVec<ReserveData<T::ReserveIdentifier, T::Balance>, T::MaxReserves>,
		ValueQuery,
	>;
```


### Suggestion

There's no way, that I am aware of, to basically just swap the key under which a storage entry is stored. So we can't swap `CurrencyId::AUSD` to `CurrencyId::ForeignAsset` just like that. Neither to transfer all balances under an asset id to another (which makes sense obviously).

The way to accomplish this, it seems, is to basically delete any entry we find in all of the storage items listed above that are under `CurrencyId::AUSD` and insert new entries with the same values but under the `CurrencyId::ForeignAsset` key.

Are you aware of any other way? Or do you have any suggestion or blocker against what seems to me the only way to run this migration?

libs/types/src/permissions.rs
```diff
@@ -666,41 +665,4 @@ mod tests {
 		assert!(!roles.exists(Role::PoolRole(PoolRole::InvestorAdmin)));
 		assert!(!roles.exists(Role::PoolRole(PoolRole::PODReadAccess)));
 	}
-
-	/// Sanity check for every CurrencyId variant's encoding value.
-	/// This will stop us from accidentally moving or dropping variants
-	/// around which could have silent but serious negative consequences.
-	#[test]
-	fn currency_id_encode_sanity() {
```
NunoAlexandre, line 674: Move this to `tokens.rs`, not sure why this was here 🙃

NunoAlexandre (commented):
Notes

libs/types/src/tokens.rs
```diff
@@ -41,23 +51,24 @@ use crate::{xcm::XcmMetadata, EVMChainId};
...
 
-	/// Karura KSM
-	KSM,
-
-	/// Acala Dollar
-	/// In Altair, it represents AUSD in Kusama;
-	/// In Centrifuge, it represents AUSD in Polkadot;
+	#[codec(index = 2)]
+	/// DEPRECATED - Will be removed in the following up Runtime Upgrade once
+	/// the orml_tokens' balances are migrated to the new CurrencyId for AUSD.
```
mustermeiszer, line 63: `KSM` can be removed as we don't have balances for that, right?

libs/types/src/tokens.rs
```diff
@@ -322,17 +375,114 @@ mod tests {
...
+			);
+			assert_eq!(after::CurrencyId::Native.encode(), vec![0]);
+
+			// Tranche
+			assert_eq!(
+				before::CurrencyId::Tranche(33, default_tranche_id()).encode(),
+				after::CurrencyId::Tranche(33, default_tranche_id()).encode()
+			);
+			assert_eq!(
+				after::CurrencyId::Tranche(33, default_tranche_id()).encode(),
```
mustermeiszer, line 410: Do we plan on keeping before forever then?

pallets/pool-system/src/migrations/mod.rs
```diff
@@ -176,6 +176,11 @@ mod test {
 	use super::*;
 	use crate::mock::*;
 
+	// todo(nuno): figure out if this migration code is still needed and whether
+	// this needs to match the real foreign asset id in production and for which
+	// runtime if so.
+	const AUSD_CURRENCY_ID: CurrencyId = CurrencyId::ForeignAsset(1);
+
```
mustermeiszer, line 183: What are the implications of that? I do not really get it. ^^

runtime/altair/src/lib.rs
```diff
@@ -1371,12 +1371,12 @@ pub struct PoolCurrency;
...
-			| CurrencyId::KSM
-			| CurrencyId::ForeignAsset(_)
-			| CurrencyId::Staking(_) => false,
-			CurrencyId::AUSD => true,
+			// Deprecated, remove in the next runtime upgrade
+			CurrencyId::AUSD => false,
+			CurrencyId::Tranche(_, _) | CurrencyId::Native | CurrencyId::Staking(_) => false,
+			CurrencyId::ForeignAsset(_) => OrmlAssetRegistry::metadata(&id)
+				.map(|m| m.additional.pool_currency)
+				.unwrap_or(false),
```
mustermeiszer, line 1379: nice and flexible!

runtime/integration-tests/src/xcm/polkadot/tests/asset_registry.rs
```diff
@@ -108,7 +108,7 @@ fn register_foreign_asset_works() {
 fn register_tranche_asset_blocked() {
 	Centrifuge::execute_with(|| {
 		let meta: AssetMetadata<Balance, CustomMetadata> = AssetMetadata {
-			decimals: 12,
+			decimals: 18,
```
mustermeiszer, line 111: ?

mustermeiszer (dismissed):
Looks good! 
Just some minor questions. Nothing to be changed, just to get a overall compelte picture.

libs/types/src/tokens.rs
NunoAlexandre: No! We can even drop this now, I just wrote this test to ensure that the different variants would encode accordingly to the new structure. Wdyt?

libs/types/src/tokens.rs
NunoAlexandre: Yap, this is just in the temporarily `mod before` to have the exact same previous definition of `CurrencyId` and ensure the other variants encode the way we expect them to in the new representation.

pallets/pool-system/src/migrations/mod.rs
NunoAlexandre: The migrations I have been working lately were well against third-party pallets, such as `orml_tokens`, `orml_asset_registry`, etc. Those migrations are therefore written and copy-pasted for each specific runtime.

This migration is for a pallet of our own and is re-used for all runtimes. In this specific case, I first wonder if this migration code can already be removed or not; and if not, whether we have to find a way of matching the `ForeignAsset` for `centrifuge` and for `altair` respectively (given they have different ids across those two runtimes)

runtime/integration-tests/src/xcm/polkadot/tests/asset_registry.rs
NunoAlexandre: My bad, forgot to revert. I was under the impression due to a discussion on the solidity side that tranche tokens were always 18 decimals so I thought this was wrong. Will revert 👍

libs/types/src/tokens.rs
mustermeiszer: Sounds good! Thanks for the explainer

mustermeiszer: Maybe let's remove in the follow up clean up pr. Makes the pr clearer for future us. Wdyt?

pallets/pool-system/src/migrations/mod.rs
mustermeiszer: Ahhh. Maybe have a getter in the migration code? 

Is this migration already included in the runtimes?

libs/types/src/tokens.rs
NunoAlexandre: Yeah let's do that ✌🏻 thanks for the suggestion!

pallets/pool-system/src/migrations/mod.rs
NunoAlexandre: Just opened a PR removing this code since it's seems to have been last touched 2 months ago and applied already: https://github.com/centrifuge/centrifuge-chain/pull/1421

runtime/altair/src/migrations.rs
```diff
@@ -166,3 +169,145 @@ mod asset_registry {
...
+
+					Some(details)
+				},
+			);
+
+			let n = pallet_pool_system::Pool::<Runtime>::iter().count() as u64;
+			<Runtime as frame_system::Config>::DbWeight::get().reads_writes(n, n)
+		}
+	}
+}
```
NunoAlexandre, line 313: ℹ️  This is a super simple, idempotent migration. It will only run on Altair and apply to the single pool we have there.

mustermeiszer (approved):
LGTM! @wischli we should be a bit careful when testing that. Maybe create a dedicated amount of `AUSD` accounts in the test environments.

libs/types/src/tokens.rs
```diff
@@ -23,6 +23,16 @@ use sp_runtime::{traits::Get, DispatchError, TokenError};
 
 use crate::{xcm::XcmMetadata, EVMChainId};
 
+/// The type for all Currency ids that our chains handles.
```
wischli, line 26: Nit: Could we enforce a line break to have `#[codec(index =_)]` in the same line? It would also be great, if `NOTE`s always started a new paragraph, either with or without an empty line before.

Moreover, the notes appear to say the same?

```suggestion
/// The type for all Currency ids that our chains handles.
/// Foreign assets gather all the tokens that are native to other chains, such
/// as DOT, AUSD, UDST, etc. 
///
/// NOTE: We MUST never change the `#[codec(index = _)]`
/// below as doing so results in corrupted storage keys. If changing the
/// index value of a variant is mandatory, a storage migration must take place
/// to ensure that the values under an old codec-encoded key are moved to the
/// new key.
```

libs/types/src/tokens.rs
wischli: The `AUSD` variant should be assigned index `3` as `2` currently maps to `KSM`. We must leave a note that both indices are deprecated.

libs/types/src/tokens.rs
```diff
@@ -322,17 +375,116 @@ mod tests {
...
+	/// around which could have silent but serious negative consequences.
+	#[test]
+	fn currency_id_encode_sanity() {
+		// Verify that every variant encodes to what we would expect it to.
+		// If this breaks, we must have changed the order of a variant, added
+		// a new variant in between existing variants, or deleted one.
+		vec![
+			Native,
+			Tranche(42, [42; 16]),
+			AUSD,
```
wischli, line 461: AFAIK, we must add `KSM` here as well. Moreover, how do you envision the test to stay in its current form if we remove `AUSD` from the `CurrencyId` enum? It appears to me that this test is not foolproof.

libs/types/src/tokens.rs
```diff
@@ -322,17 +375,116 @@ mod tests {
...
+		fn expected_encoded_value(id: CurrencyId) -> Vec<u8> {
+			match id {
+				Native => vec![0],
+				Tranche(pool_id, tranche_id) => {
+					let mut r: Vec<u8> = vec![1];
+					r.append(&mut pool_id.encode());
+					r.append(&mut tranche_id.to_vec());
+					r
+				},
+				AUSD => vec![2],
```
wischli, line 480: Again, should be 3.

pallets/loans/src/benchmarking.rs
```diff
@@ -133,6 +133,9 @@ where
 		)
 		.unwrap();
 
+		// todo(nuno): I guess this is the right place to register AUSD in the asset
```
wischli: Can you add some context to this `todo` please? E.g. why do we not resolve it in this PR or can we remove it?

runtime/altair/src/lib.rs
```diff
@@ -1438,12 +1438,12 @@ pub struct PoolCurrency;
 impl Contains<CurrencyId> for PoolCurrency {
 	fn contains(id: &CurrencyId) -> bool {
 		match id {
-			CurrencyId::Tranche(_, _)
-			| CurrencyId::Native
-			| CurrencyId::KSM
-			| CurrencyId::ForeignAsset(_)
-			| CurrencyId::Staking(_) => false,
-			CurrencyId::AUSD => true,
+			// Deprecated, remove in the next runtime upgrade
```
wischli, line 1441: Maybe we want to specify a version number here? For me, `next` would be literal next runtime upgrade, e.g. the one which migrates `AUSD`. However, the comment wants to say `one after next` right? We can assume the migration will be part of `1028` such that it can be removed when bumping to `1029`.

runtime/altair/src/migrations.rs
```diff
@@ -166,3 +169,145 @@ mod asset_registry {
...
+			);
+
+			for (account, account_data) in old_state.entries {
+				ensure!(
+					OrmlTokens::accounts(&account, NEW_AUSD_CURRENCY_ID) == account_data.clone(),
+					"The account data under the new AUSD Currency does NOT match the old one"
+				);
+			}
+
+			return Ok(());
```
wischli, line 233: Nit: The dark side (Solidity) has shown its impact!

```suggestion
			Ok(())
```

runtime/altair/src/migrations.rs
```diff
@@ -166,3 +169,145 @@ mod asset_registry {
...
+					.map_err(|e| {
+						log::error!(
+							"Failed to call burn_from({:?}, {:?}, {balance}): {:?}",
+							DEPRECATED_AUSD_CURRENCY_ID,
+							account,
+							e
+						)
+					})
+					.ok();
+					// // Now mint the amount under the new CurrencyID
```
wischli, line 261: Nit

```suggestion
					// Now mint the amount under the new CurrencyID
```

runtime/altair/src/migrations.rs
```diff
@@ -166,3 +169,145 @@ mod asset_registry {
...
+
+					migrated_entries += 1;
+				});
+
+			// Approximate weight given for every entry migration there are two calls being
+			// made, so counting one read and one write per call. Given that we have very
+			// few entries to be migrated, the specific value isn't quite relevant and we
+			// could just set this to a high-enough value.
+			<Runtime as frame_system::Config>::DbWeight::get()
+				.reads_writes(migrated_entries * 2, migrated_entries * 2)
```
wischli, line 285: AFAICS, it should be one read and two writes per entry right?

runtime/centrifuge/Cargo.toml
```diff
@@ -129,6 +129,10 @@ runtime-common = { path = "../common", default-features = false }
 # bridge pallets
 chainbridge = { git = "https://github.com/centrifuge/chainbridge-substrate.git", default-features = false, branch = "polkadot-v0.9.38" }
 
+# tmp - NUNO DELETE THIS
```
wischli, line 132: Can this be removed in this PR?

runtime/altair/src/migrations.rs
wischli: Would it be possible to add the `try-runtime` pre and post upgrade hooks? I understand it seems unnecessary in this scenario but I think we should NEVER run a migration without pre and post hooks. Otherwise there is no way to test it apart from locally running the migration. Since executing `try-runtime` is part of the release process, it should be enforced.

runtime/centrifuge/src/lib.rs
```diff
@@ -1275,11 +1275,9 @@ pub struct PoolCurrency;
 impl Contains<CurrencyId> for PoolCurrency {
 	fn contains(id: &CurrencyId) -> bool {
 		match id {
-			CurrencyId::Tranche(_, _)
-			| CurrencyId::Native
-			| CurrencyId::KSM
-			| CurrencyId::Staking(_) => false,
-			CurrencyId::AUSD => true,
+			// Deprecated, remove in the next runtime upgrade
```
wischli, line 1278: See comment above about potentially specifying the spec version here as we can expect the migration to be part of `1020`.

wischli (commented):
I have a few nits, some question on TODOs and think the `CurrencyId` index deprecation is not entirely correct.

The migrations are well written! I was a bit hesitant on the fact that we don't theoretically handle locked and reserved `AUSD`s on Altair but checked that a user does not seem to be able to lock or reserve these anyway. If so, we would unlock/-reserve them in the worst case, which would still be fine.

pallets/loans/src/benchmarking.rs
NunoAlexandre: This has been removed

runtime/altair/src/migrations.rs
NunoAlexandre: Ai 😱  thanks!

runtime/altair/src/migrations.rs
NunoAlexandre: `burn_from` reads `TotalIssuance::<T>::get(` and `Self::accounts(who, currency_id).free`, so that's at least 2 reads per entry. As written in the comment - correct me if I am wrong - it doesn't really matter; this applies to just a few entries and the main purpose of weights and fees is to avoid spamming which is not a risk with these migrations.

libs/types/src/tokens.rs
NunoAlexandre: No, KSM can already and is already dropped since we don't have to migrate anything for KSM.

>  It appears to me that this test is not foolproof.

It's not, indeed. This test has value if we don't go in messing around with the CurrencyId variants, which is exactly what we did in this PR (removing AUSD and then adding it back to make the migration possible).

Once we remove AUSD in a follow up PR (after the migration is applied), we drop the `CurrencyId::AUSD` variant and drop `AUSD` from this list.

libs/types/src/tokens.rs
NunoAlexandre: Thanks! apparently I wrote the same NOTE twice, something went wrong there. Fixed

runtime/altair/src/migrations.rs
wischli: > main purpose of weights and fees is to avoid spamming which is not a risk with these migrations.

For extrinsics, this is true. For migrations however, we should always strive to provide a worst-case weight which always overestimates. Otherwise we could end up with tuple of migrations which exceed the maximum block weight such that the block in which the runtime upgrade is applied cannot be produced by para collators and relay validators with hardware close to our benchmarks. As long as this block is not produced, collators will continue trying to produce the same block and get it validated by relay validators. Depending on the amount of overweight, the chain could stall until relay governance comes to the rescue.

NunoAlexandre:
> I have a few nits, some question on TODOs and think the `CurrencyId` index deprecation is not entirely correct.

Thanks A LOT for the catch, it's a massive one 👌 

> 
> The migrations are well written! I was a bit hesitant on the fact that we don't theoretically handle locked and reserved `AUSD`s on Altair but checked that a user does not seem to be able to lock or reserve these anyway. If so, we would unlock/-reserve them in the worst case, which would still be fine.

Indeed, and it's not possible to move these assets cross chain atm anyways, so let's just get this migration executed asap and be done with this deprecated `CurrencyId::AUSD` for good ✨

runtime/centrifuge/Cargo.toml
NunoAlexandre: Done

runtime/altair/src/migrations.rs
NunoAlexandre: Thanks for the context! 💛  

> For migrations however, we should always strive to provide a worst-case weight which always overestimates

That's why I specified what seems to me the correct estimation of reads and rights.

Do you think there's more to do on this thread?

wischli: The current estimation does not overestimate! It should be 5 reads and 4 writes per item as burning and minting each read and write both `TotalIssuance` as well as `Account`. The last read comes from iterating the `Accounts` storage map.

NunoAlexandre: Done

runtime/altair/src/migrations.rs
NunoAlexandre: Done ✅

NunoAlexandre:
@wischli thank you for your extensive and careful review 👍  I believe everything is addressed and ready for a final review. Thanks!

wischli (changes requested):
Thanks for applying my requested changes. Code LGTM! However, `try-runtime` fails to decode some keys, so something must be off in the migration. I can look into this further tomorrow. Here's the error:

```
2023-06-28 22:33:54.271 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 14, 146, 103, 243, 186, 79, 175, 118, 176, 54, 118, 166, 23, 214, 239, 194, 192, 153, 124, 79, 43, 58, 131, 235, 7, 239, 116, 168, 103, 207, 103, 42, 37, 162, 163, 12, 198, 26, 188, 147, 109, 204, 153, 77, 247, 123, 168, 74, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
2023-06-28 22:33:54.272 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 94, 207, 253, 123, 108, 15, 120, 117, 27, 170, 157, 40, 30, 11, 250, 58, 109, 111, 100, 108, 112, 121, 47, 116, 114, 115, 114, 121, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
2023-06-28 22:33:54.273 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 14, 146, 103, 243, 186, 79, 175, 118, 176, 54, 118, 166, 23, 214, 239, 194, 192, 153, 124, 79, 43, 58, 131, 235, 7, 239, 116, 168, 103, 207, 103, 42, 37, 162, 163, 12, 198, 26, 188, 147, 109, 204, 153, 77, 247, 123, 168, 74, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
2023-06-28 22:33:54.274 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 94, 207, 253, 123, 108, 15, 120, 117, 27, 170, 157, 40, 30, 11, 250, 58, 109, 111, 100, 108, 112, 121, 47, 116, 114, 115, 114, 121, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
```

On a note: Independent from this PR, the `CrossChainTransferabilityMigration` also fails with 
```
ERROR main runtime: panicked at 'called `Result::unwrap()` on an `Err` value: "The metadata of a NON tranche token wasn't just updated by setting `transferability` to `Xcm`"'
```

NunoAlexandre:
@wischli I will look into this myself, thanks a lot for the report ✌🏻

NunoAlexandre:
I placed some logs on my local revision and the `CurrencyIdRefactorMigration` and disabled the other Altair migrations (on branch `refactor/currency-id-debug`) and the former seems to be working well as well. Not sure right off the bat what those decoding errors have to do with 🤔 

```
2023-06-29 09:53:25.806  INFO main altair_runtime::migrations::orml_tokens_migration: CurrencyIdRefactorMigration: pre_upgrade
2023-06-29 09:53:25.806 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 14, 146, 103, 243, 186, 79, 175, 118, 176, 54, 118, 166, 23, 214, 239, 194, 192, 153, 124, 79, 43, 58, 131, 235, 7, 239, 116, 168, 103, 207, 103, 42, 37, 162, 163, 12, 198, 26, 188, 147, 109, 204, 153, 77, 247, 123, 168, 74, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
2023-06-29 09:53:25.806 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 94, 207, 253, 123, 108, 15, 120, 117, 27, 170, 157, 40, 30, 11, 250, 58, 109, 111, 100, 108, 112, 121, 47, 116, 114, 115, 114, 121, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
2023-06-29 09:53:25.807  INFO main altair_runtime::migrations::orml_tokens_migration: CurrencyIdRefactorMigration: pre_upgrade old state OldState { total_issuance: 4997544866265088, entries: [(, <wasm:stripped>), (, <wasm:stripped>), (, <wasm:stripped>), (, <wasm:stripped>), (, <wasm:stripped>), (, <wasm:stripped>), (, <wasm:stripped>), (, <wasm:stripped>)] }
2023-06-29 09:53:25.807 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 14, 146, 103, 243, 186, 79, 175, 118, 176, 54, 118, 166, 23, 214, 239, 194, 192, 153, 124, 79, 43, 58, 131, 235, 7, 239, 116, 168, 103, 207, 103, 42, 37, 162, 163, 12, 198, 26, 188, 147, 109, 204, 153, 77, 247, 123, 168, 74, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
2023-06-29 09:53:25.809 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 94, 207, 253, 123, 108, 15, 120, 117, 27, 170, 157, 40, 30, 11, 250, 58, 109, 111, 100, 108, 112, 121, 47, 116, 114, 115, 114, 121, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
2023-06-29 09:53:25.811  INFO main altair_runtime::migrations::orml_tokens_migration: CurrencyIdRefactorMigration: post_upgrade
2023-06-29 09:53:25.811  INFO main altair_runtime::migrations::orml_tokens_migration: CurrencyIdRefactorMigration: post_upgrade new total issuance 4997544866265088
2023-06-29 09:53:25.811  INFO main altair_runtime::migrations::orml_tokens_migration: CurrencyIdRefactorMigration: great success
```

On it, but it's good to know that things are running as expected or would have under the normal circumstances.

NunoAlexandre:
it is clear that those `ERROR main frame_support::storage: (key, value) failed to decode at` errors only occur within the `CurrencyIdRefactorMigration`, but the migration itself executes correctly. I have tried a few things to make sure I wasn't decoding the wrong types but that seems fine as well.

The error indicates it's an issue within the underlying `frame_support` storage, which is not immeditately obvious as to how that's caused by reading the storage of the `orml_tokens` pallet, aside from shared types like `AccountId`.

@wischli have you experienced anything similar in the past?

mustermeiszer:
@NunoAlexandre I guess the trace is misleading. IIRC the error results from an decoding error. This means, the bytes in storage can not be decoded into the type that should be the result of the query. The code for that lives in the frame_support. Hence, the trace information.

NunoAlexandre:
> @NunoAlexandre I guess the trace is misleading. IIRC the error results from an decoding error. This means, the bytes in storage can not be decoded into the type that should be the result of the query. The code for that lives in the frame_support. Hence, the trace information.

Thanks for the input! That makes sense but I don't follow how that seems to be triggered from the `CurrencyIdRefactorMigration`, which itself executes correctly 🤔

wischli:
> Thanks for the input! That makes sense but I don't follow how that seems to be triggered from the `CurrencyIdRefactorMigration`, which itself executes correctly 🤔

It might seemingly execute correctly as all keys/values which fail to decode are just skipped: https://paritytech.github.io/substrate/master/src/frame_support/storage/types/map.rs.html#323

NunoAlexandre:
> https://paritytech.github.io/substrate/master/src/frame_support/storage/types/map.rs.html#323

I see. I'd assume it would not decode anything at all. In any case, that seems an unrelated issue to this PR given that the migration works and migrates the values its meant to migrate as expected.

FWIW I tried creating a storage alias for the `orml_tokens::Accounts` using `before::CurrencyId`, to make sure the issue wasn't' that it was trying to decode a `KSM` entry we didn't know about and that didn't make a difference.

How do you guys propose to follow further on this? I would like to close this PR that's been dragging forever and would thus suggest that we create an issue to debug what potential account data is corrupted on orml_tokens. Open for other suggestions 👍

NunoAlexandre:
I will debug this further, if I need help I will let you guys know 👍

NunoAlexandre:
In the `refactor/currency-id-decode-error` I've pushed a commit that turns the `CurrencyIdRefactorMigration` to just this:

```rust
impl OnRuntimeUpgrade for CurrencyIdRefactorMigration {
	fn on_runtime_upgrade() -> Weight {
		// Collect all the orml_tokens::Accounts entries
		let _: Vec<_> = orml_tokens::Accounts::<Runtime>::iter().collect::<_>();

		return Weight::zero();
	}
}
```

The error still shows up:

```
2023-06-29 12:26:01.341 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 14, 146, 103, 243, 186, 79, 175, 118, 176, 54, 118, 166, 23, 214, 239, 194, 192, 153, 124, 79, 43, 58, 131, 235, 7, 239, 116, 168, 103, 207, 103, 42, 37, 162, 163, 12, 198, 26, 188, 147, 109, 204, 153, 77, 247, 123, 168, 74, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
2023-06-29 12:26:01.341 ERROR main frame_support::storage: (key, value) failed to decode at [206, 171, 181, 23, 175, 131, 191, 94, 198, 102, 51, 20, 164, 91, 11, 163, 142, 231, 65, 138, 101, 49, 23, 61, 96, 209, 246, 168, 45, 143, 77, 81, 94, 207, 253, 123, 108, 15, 120, 117, 27, 170, 157, 40, 30, 11, 250, 58, 109, 111, 100, 108, 112, 121, 47, 116, 114, 115, 114, 121, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 50, 103, 157, 214, 62, 218, 185, 100, 4]: Error
```

This proves my suspicion that the issue is not with the migration of the AUSD entries (which, again, seems to work as expected) but rather with some other corrupted data in the storage. We see less error logs but that's because in this derived branch we don't pre & post iter over the storage but only on `on_runtime_upgrade`.

NunoAlexandre:
FYI I've created a vanilla PR based off `main` without the changes to the `CurrencyId` type and the decoding error still shows there: https://github.com/centrifuge/centrifuge-chain/pull/1427/files

So there's definitely some corrupted entries in Altair's `orml_tokens`. The only approach I can think of is to try and `translate` the entries by reading them as bytearrays and try and decode them as the right expected values to pin point which entries are corrupted and then we see what we can do about it.

Given that this was apparently not known to be an issue in the first place, I don't expect this to have priority over other more important tasks? @mustermeiszer wdyt?

wischli (approved):
Thanks for digging into the issue which became visible in the new migrations, despite being independent from these. Let's open an issue and merge this PR.

PR #1344, opened 2023-05-12
refactor: remove rewards domain concept

wischli:
# Description

* Removes concept of `Domain` for `pallet-rewards` and its consumers
* Fixes https://github.com/centrifuge/centrifuge-chain/issues/1246
* Based on #1342 

# Checklist:


pallets/rewards/src/lib.rs
```diff
@@ -51,6 +48,12 @@
 //!
 //! The Rewards pallet provides implementations for the Rewards trait.
 //!
+//! ### Assumptions
+//!
+//! Each consuming reward system must have its unique instance of this pallet
+//! independent of the underlying reward mechanism. E.g., one instance for Block
+//! Rewards and another for Liquidity Rewards.
+//!
```
lemunozm, line 56: Great comment! 🙌🏻

pallets/rewards/src/tests/command_based.rs
```diff
@@ -105,11 +105,11 @@ where
 /// | claim      | A, B         | 2      | 4      |
 /// | distribute |  -           | 2      | 2      |
 ///
-/// It uses 1 group, 1 domain and 1 currency.
+/// It uses 1 group and 1 currency.
 /// It uses the `base` mechanism.
 #[test]
 fn silly_sample_for_fuzzer() {
-	const DOM_CURR: (DomainId, CurrencyId) = (DomainId::D1, CurrencyId::A);
+	const DOM_CURR: CurrencyId = CurrencyId::A;
```
lemunozm, line 112: I would rename it to `CURR`

runtime/development/src/lib.rs
```diff
@@ -2187,15 +2155,15 @@ impl_runtime_apis! {
...
 			.into_iter().chain(
 				pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance2>::list_currencies(&account_id).into_iter()
 			).collect()
 		}
 
-		fn compute_reward(currency_id: (RewardDomain, CurrencyId), account_id: AccountId) -> Option<Balance> {
+		fn compute_reward(currency_id: CurrencyId, account_id: AccountId) -> Option<Balance> {
 			<pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance1> as AccountRewards<AccountId>>::compute_reward(currency_id, &account_id)
 			.or_else(|_|
 				<pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance2> as AccountRewards<AccountId>>::compute_reward(currency_id, &account_id)
```
lemunozm, line 2169: I think this can no longer be chained, because we do not know from which pallet is comming the Currency. 

I think here we need to have two RPC calls like `collator_list_currencies` and `liquidity_list_currencies` or even better returning something like HashMap<String, Vec<CurrencyId>>`, with some identificatory name. Or using the Domain concept just as the key here.

lemunozm (commented):
Only one detail, but the change itself is just perfect!

pallets/rewards/src/tests/currency_movement.rs
```diff
@@ -245,10 +245,10 @@ macro_rules! currency_movement_tests {
...
-						&(DomainId::D1, CurrencyId::B),
-						&(DomainId::D1, CurrencyId::C),
-						&(DomainId::D1, CurrencyId::M),
+						&CurrencyId::A,
+						&CurrencyId::B,
+						&CurrencyId::C,
+						&CurrencyId::M,
 					];
 
 					assert_ok!($pallet::attach_currency(DOM_1_CURRENCY_X, GROUP_1));
```
thea-leake, line 254: Would we want to remove `DOM_x_` from `DOM_[0-9]_CURRENCY_{[A-Z]}`?

thea-leake (approved):
Looks good to me!  Saw name that should maybe be changed, but not blocker for me :).

runtime/centrifuge/src/lib.rs
```diff
@@ -1934,13 +1921,19 @@ impl_runtime_apis! {
...
+			}
 		}
 
-		fn compute_reward(currency_id: (RewardDomain, CurrencyId), account_id: AccountId) -> Option<Balance> {
-			<pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance1> as cfg_traits::rewards::AccountRewards<AccountId> >::compute_reward(currency_id, &account_id).ok()
+		fn compute_reward(domain: runtime_common::apis::RewardDomain, currency_id: CurrencyId, account_id: AccountId) -> Option<Balance> {
+			match domain {
+				runtime_common::apis::RewardDomain::Block => <pallet_rewards::Pallet::<Runtime, pallet_rewards::Instance1> as cfg_traits::rewards::AccountRewards<AccountId>>::compute_reward(currency_id, &account_id).ok(),
+				_ => None,
+			}
```
mustermeiszer, line 1936: Out of interest, why don't we compute for `liquidity`?

mustermeiszer (commented):
Looks really clean and nice! Before approving though

* Did we already have storage in the rewards pallet? 
  * If YES: 
    * Do we need to migrate this, as the instance will change the prefix?

wischli:
> Did we already have storage in the rewards pallet?

We don't (apart from the `Dev` network) which is why we wanted to refactor this before go-live.

mustermeiszer (approved):

runtime/development/src/lib.rs
wischli: Great catch! I went with your third proposal of using the Domain concept as a key. The main reason for this is that using a HashMap introduced custom `encoding` and `decoding` complexity as it not available by default. 

[2a25440](https://github.com/centrifuge/centrifuge-chain/pull/1344/commits/2a254407c6165621f02a456b9818007192cd0007)

PR #1345, opened 2023-05-12
connectors: Sort messages test by call index

NunoAlexandre:
NIT, just moving some tests around to follow the messages call index which is helpful when cross testing with the Solidity counter part.


wischli (approved):
LGTM, thanks for cleaning up after me 🌚

NunoAlexandre:
> LGTM, thanks for cleaning up after me 🌚

ah, this is not even "cleaning up" :) the fact that this is all there's to do after cross testing the messages speaks wonders of your work!

PR #1346, opened 2023-05-16
Mocks: Use mock test utility without dependencies

lemunozm:
# Description

Each time you want to use the mock test utility macro, you need to add a bunch of dependencies to use it correctly. Some of them are not trivial at first view. This change allows using the macro without importing dependencies


mustermeiszer (dismissed):
Sane change!

mustermeiszer (approved):
Re-approve!

PR #1347, opened 2023-05-16
Register pools with write-off policy

offerijns:
# Description

This adds a parameter to the pool registration extrinsic to add a write-off policy, which will be inserted in the loans pallet. This way, the pool that CFG holders are voting on before launch already contains this information.

## Changes and Descriptions

- [x] Add write-off policy mutation trait
- [x] Integrate write-off policy parameter in registry pallet

# Checklist:


libs/traits/src/lib.rs
```diff
@@ -187,6 +187,18 @@ pub trait PoolReserve<AccountId, CurrencyId>: PoolInspect<AccountId, CurrencyId>
 	fn deposit(pool_id: Self::PoolId, from: AccountId, amount: Self::Balance) -> DispatchResult;
 }
 
+/// A trait that supports modifications of pool write-off policies
+pub trait PoolWriteOffPolicyMutate<PoolId> {
+	type Rate;
```
lemunozm, line 192: I think this `Rate` is not needed

libs/traits/src/lib.rs
```diff
@@ -187,6 +187,18 @@ pub trait PoolReserve<AccountId, CurrencyId>: PoolInspect<AccountId, CurrencyId>
...
+	type WriteOffRule: Encode + Decode + Clone + TypeInfo + Debug + PartialEq;
```
lemunozm, line 193: Although for these bounds I think it doesn't matter. Personally, I prefer to leave the associated types as generic as possible, adding here only the bounds required by the trait itself. Later it will ease the mocking/testing part and reason about what's really needed for the implementer.

pallets/pool-registry/src/mock.rs
```diff
@@ -237,6 +240,33 @@ impl<
...
+	type Rate = <T as pallet_pool_registry::Config>::Rate;
+	type WriteOffRule = WriteOffRule<Self::Rate>;
+
+	fn update(
+		pool_id: PoolId,
+		policy: BoundedVec<Self::WriteOffRule, Self::MaxWriteOffPolicySize>,
+	) -> DispatchResult {
+		Ok(())
+	}
+}
```
lemunozm, line 268: I think this trait is the perfect use case for using `mock-builder` and avoiding the need to import a `pallet-loans-ref` dependency in this pallet.

libs/traits/src/lib.rs
```diff
@@ -187,6 +187,18 @@ pub trait PoolReserve<AccountId, CurrencyId>: PoolInspect<AccountId, CurrencyId>
...
+	type MaxWriteOffPolicySize: Get<u32>;
+
+	fn update(
+		pool_id: PoolId,
+		policy: BoundedVec<Self::WriteOffRule, Self::MaxWriteOffPolicySize>,
```
lemunozm, line 198: I feel like having a `BoundedVec` with a `MaxPolicySize` seems to be too much implementation detail of how it is currently handled from `pallet-loans-ref`. What do you think about having:

```rust
policy: impl IntoIterator<Item = Self::WriteOffRule>
```

In this way, any kind of list can be used as a policy, and it is the implementation that needs to ensure it works for the way it is using it.

lemunozm (commented):
I like the idea! 🚀 

Some comments below with some thoughts

pallets/pool-registry/src/tests.rs
```diff
@@ -75,6 +75,7 @@ fn register_pool_and_set_metadata() {
 				currency,
 				max_reserve,
 				metadata.clone(),
+				BoundedVec::default()
```
mustermeiszer, line 78: I would add a test for checking they are added validly here.

mustermeiszer (commented):
Looks already nice. I think this is a good change. Good that we introduced the registry so that this does not need to land in the system.

libs/traits/src/lib.rs
branan: Since the data is coming in as a `BoundedVec` from an extrinsic call anyway, I think converting it to an iterator and then re-collecting on the other side isn't really worth the slightly nicer API. We can always clean it up later if we end up with another potential consumer of this trait.

lemunozm: No strong opinion if now we have the same structure in both places 👍🏻

pallets/pool-registry/src/lib.rs
```diff
@@ -241,6 +254,7 @@ pub mod pallet {
 			currency: T::CurrencyId,
 			max_reserve: T::Balance,
 			metadata: Option<Vec<u8>>,
+			write_off_policy: BoundedVec<WriteOffRuleOf<T>, MaxWriteOffPolicySizeOf<T>>,
```
lemunozm, line 257: Why, instead of adding here `write_off_policy` as a parameter we are not doing a batch with:
- PoolRegistry::register(..)
- Loans::update_write_off_policy(..)
?

offerijns: The batch transaction will go through CFG governance. Two issues with the above proposal:
1. The origin of the transaction will be `Root` rather than the pool admin, since CFG governance is accepting it. This means we'd need to modify `update_write_off_policy` to be callable also by `Root`, which I would prefer not to.
2. The call will be displayed in governance UIs like Polkadot.js, so keeping it not a batch makes it much more readable.

libs/traits/src/lib.rs
offerijns: I would keep it as is then :)

pallets/pool-registry/src/tests.rs
offerijns: Ticket created: https://github.com/centrifuge/centrifuge-chain/issues/1499

lemunozm:
If `pool-registry` never iterates the policy, maybe it could be simpler to pass the whole policy directly as a generic type, without understanding what is `WriteOffRule` and `MaxWriteOffPolicySizeOf`, WDYT?

pallets/pool-registry/src/tests.rs
```diff
@@ -78,6 +80,7 @@ fn register_pool_and_set_metadata() {
 				currency,
 				max_reserve,
 				metadata.clone(),
+				()
```
lemunozm, line 83: Updated trait and add pallet mock

From `pool-registry` perspective, we do not need to know more than the policy itself (whatever it is) reach the `update()` call.

runtime/altair/src/lib.rs
```diff
@@ -1511,6 +1511,7 @@ impl pallet_pool_registry::Config for Runtime {
 	type MaxTokenSymbolLength = MaxTrancheSymbolLengthBytes;
 	type MaxTranches = MaxTranches;
 	type ModifyPool = pallet_pool_system::Pallet<Self>;
+	type ModifyWriteOffPolicy = pallet_loans_ref::Pallet<Self>;
```
lemunozm, line 1514:
```suggestion
	type ModifyWriteOffPolicy = pallet_loans::Pallet<Self>;
```

lemunozm (approved):
LGTM! I think everything is fine!

mustermeiszer (approved):

mustermeiszer (approved):

libs/mocks/src/write_off_policy.rs
```diff
@@ -34,5 +34,10 @@ pub mod pallet {
 		fn update(a: T::PoolId, b: T::Policy) -> DispatchResult {
 			execute_call!((a, b))
 		}
+
+		#[cfg(feature = "runtime-benchmarks")]
+		fn worst_case_policy() -> Self::Policy {
```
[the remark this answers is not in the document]
lemunozm: To test this:
```sh
cargo test -p pallet-pool-registry -F runtime-benchmarks
```

lemunozm (approved):

mustermeiszer (dismissed):

mustermeiszer (approved):

lemunozm (approved):

PR #1349, opened 2023-05-17
Evm part 2

branan:
# Description

This brings the EVM support to Altair and Centrifuge runtimes, with additional restrictions compared to the Development runtime. I particular:

* The production runtimes filter all non-root calls to the EVM pallet
* The production runtimes disallow creating contracts via the EVM RPC calls

## Changes and Descriptions

* Add EVM support to Altair runtime
* Add EVM support to Centrifuge runtime
* Add EVM support to altair/centrifuge node services
* Various minor EVM cleanups

# Checklist:

- [X] I have added Rust doc comments to structs, enums, traits and functions
- [X] I have performed a self-review of my code


branan:
I definitely meant to make this a draft, it's still missing a couple of parts:

* Additional command line parameters for some hardcoded node values
* The updated account conversion as discussed last week
* Migrations to set the EVM chain ID for our live networks.
  * I've actually got this mostly done locally, but I need a way to handle a migration recognizing when it's on Centrifuge vs Catalyst so that it can set different chain IDs for each

branan:
There are no benchmarks in this version of Frontier :tada:. It'll be something to add in 0.9.40

runtime/altair/src/lib.rs
```diff
@@ -1563,6 +1579,88 @@ pub type Executive = frame_executive::Executive<
...
+				)))
+			}
+			_ => None,
+		}
+	}
+}
+
+pub struct TransactionConverter;
+
+impl fp_rpc::ConvertTransaction<UncheckedExtrinsic> for TransactionConverter {
```
thea-leake, line 1642: Just as a question to make sure I understand correctly what's happening, signing verification is being done via `SelfContainedCall` impl validate and apply methods?

thea-leake (commented):
Not sure if it's still supposed to be marked as draft as per PR comment, but looks good to me :)

runtime/altair/src/lib.rs
branan: Yup, It all eventually ends up down in https://github.com/paritytech/frontier/blob/polkadot-v0.9.37/primitives/self-contained/src/unchecked_extrinsic.rs#L94-L119 which does the actual signature validation

NunoAlexandre (commented):
One thing that popped up for me on this first round. Will give it a more detailed looked later 👍

runtime/altair/src/evm.rs
```diff
@@ -0,0 +1,104 @@
...
+			let authority_id = Aura::authorities()[author_index as usize].clone();
+			return Some(H160::from_slice(&authority_id.to_raw_vec()[4..24]));
+		}
+		None
+	}
+}
+
+// From Moonbeam:
+//
+// Current approximation of the gas/s consumption considering
```
NunoAlexandre, line 49: If that a typo or what does `gas/s` mean? 🐒 

```suggestion
// Current approximation of the gas consumption considering
```

runtime/altair/src/evm.rs
```diff
@@ -0,0 +1,104 @@
...
+};
+use sp_core::{crypto::ByteArray, H160, U256};
+use sp_runtime::Permill;
+use sp_std::marker::PhantomData;
+
+use crate::Aura;
+
+// To create valid Ethereum-compatible blocks, we need a 20-byte
+// "author" for the block. Since that author is purely informational,
+// we do a simple truncation of the 32-byte Substrate author
```
NunoAlexandre, line 32:
```suggestion
/// To create valid Ethereum-compatible blocks, we need a 20-byte
/// "author" for the block. Since that author is purely informational,
/// we do a simple truncation of the 32-byte Substrate author
```

runtime/altair/src/evm.rs
```diff
@@ -0,0 +1,104 @@
...
+	where
+		I: 'a + IntoIterator<Item = (ConsensusEngineId, &'a [u8])>,
+	{
+		if let Some(author_index) = F::find_author(digests) {
+			let authority_id = Aura::authorities()[author_index as usize].clone();
+			return Some(H160::from_slice(&authority_id.to_raw_vec()[4..24]));
+		}
+		None
+	}
+}
```
NunoAlexandre, line 45: This kind of runtime-agnostic trait impl could to to `runtime-commons` and just be referenced in the different runtimes instead of duplicated in all of them 🧹

NunoAlexandre (commented):
Nothing major :)

runtime/altair/src/evm.rs
branan: I think it needs to be made generic over the Runtime (that `Aura` in there is really `pallet_aura::Pallet::<Runtime>`), but indeed could be made properly runtime agnostic. Worth doing for sure

runtime/altair/src/evm.rs
branan: gas-per-second. I'll expand the abbreviation

runtime/altair/src/evm.rs
NunoAlexandre: Ah, my bad, I didn't spot the `Aura` bit 🙃 I don't know how to make stuff generic over a runtime when referencing a specific pallet from that runtime tho.

src/service/evm.rs
```diff
@@ -55,6 +55,34 @@ use substrate_prometheus_endpoint::Registry;
...
+	/// Maximum number of logs in a query.
+	#[clap(long, default_value = "10000")]
+	pub max_past_logs: u32,
+
+	/// Maximum fee history cache size.
+	#[clap(long, default_value = "2048")]
+	pub fee_history_limit: u64,
+
+	#[clap(long)]
+	pub enable_dev_signer: bool,
```
mustermeiszer, line 70: maybe default to `false` 

WDYT?

mustermeiszer (approved):
Nice and clean!! Thanks a lot @branan!

runtime/altair/src/evm.rs
mustermeiszer: We could add a `Get<Vec<Authorities>>` and impl that for Aura in the actual runtime. But feel like this makes it more opaque than actually easier to understand.

src/service/evm.rs
branan: `bool` in clap defaults to false implicitly

runtime/common/src/account_conversion.rs
```diff
@@ -0,0 +1,54 @@
...
+		let chain_id = pallet_evm_chain_id::Pallet::<R>::get();
+		Self::convert_evm_address(chain_id, address.0)
+	}
+}
+
+// Implement connectors account conversion using our shared conversion code
+impl<R> Convert<DomainAddress, AccountId> for AccountConverter<R> {
+	fn convert(domain_address: DomainAddress) -> AccountId {
+		match domain_address {
+			DomainAddress::Centrifuge(addr) => AccountId::new(addr),
```
branan, line 50: @NunoAlexandre this behavior is different from what `DomainAddress::into_address_truncating` did. This is a direct unwrapping, whereas `into_address_truncating` would have done something like `[<tag_bytes>,<truncated_address>]`. I didn't see anything in the comments or commit messages that made that behavior seem intentional, and there were no tests that covered it.

We can change this so that `DomainAddress::Centrifuge` encodes into something different if we so desire

branan:
Regarding account conversion:

* The MultiLocation based idea won't work until we're on XCM V3. We're still on polkadot v0.9.37 and v2, which is missing several pieces we need to represent addresses correctly. Since for now we only need EVM account conversion, I've gone ahead and implemented without going through MultiLocation.
* I've also skipped the Runtime API for now - the EVM account conversion is very simple to implement in JS, and it makes more sense to introduce the runtime APIs in terms of MultiLocations when we have the updated XCM version.

We do still now have a single, central location to expand account conversion as we need it, which is the important part.

NunoAlexandre:
> Regarding account conversion:
> 
> * The MultiLocation based idea won't work until we're on XCM V3. We're still on polkadot v0.9.37 and v2, which is missing several pieces we need to represent addresses correctly. Since for now we only need EVM account conversion, I've gone ahead and implemented without going through MultiLocation.

The update to 0.9.38 is pretty much finalised apart from a few failing tests so you can always choose to rebase this branch :) Out of curiosity, what's missing in 0.9.37/xcm v2 that would help us represent addresses correctly?

runtime/common/src/account_conversion.rs
NunoAlexandre: Ah, good point. For all I know, this is exactly what we want since the idea of `DomainAddress::Centrifuge(address)` is exactly to represent a direct account address on the Centrifuge chain.

branan:
I've flipped the auto-merge switch, this is ready for final review :tada:

mustermeiszer (approved):
Re-approve! Thanks for pushing this

PR #1351, opened 2023-05-17
CollectionDataFeed: Minor changes

lemunozm:
# Description

Minor changes over `collection-data-feed` to prepare it for the OracleValuation feature.

## Changes and Descriptions

- Prices returned as `Result` instead of Option.
- `MockDataCollection` with user callback.
- Add license headers
- Clear unused features

### DIFF
--- a/pallets/collection-data-feed/src/lib.rs
+++ b/pallets/collection-data-feed/src/lib.rs
@@ -110,7 +110,7 @@ pub mod pallet {
 		type Data = Result<DataValueOf<T>, DispatchError>;
 
 		fn get(data_id: &T::DataId) -> Self::Data {
-			T::DataProvider::get_no_op(data_id).ok_or(Error::<T>::DataIdWithoutData.into())
+			T::DataProvider::get_no_op(data_id).ok_or_else(|| Error::<T>::DataIdWithoutData.into())
 		}
 
 		fn collection(collection_id: &T::CollectionId) -> Self::Collection {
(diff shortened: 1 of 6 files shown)


pallets/collection-data-feed/src/lib.rs
```diff
@@ -91,6 +94,9 @@ pub mod pallet {
 		/// The used data ID is not in the collection.
 		DataIdNotInCollection,
 
+		/// Max collection number exceeded
```
offerijns, line 97: I think this comment is copied from the other error and needs to be updated

lemunozm: You're right! done

offerijns (approved):

pallets/collection-data-feed/src/lib.rs
```diff
@@ -152,9 +159,11 @@ pub mod pallet {
 			// for Data values.
 			for collection_id in Listening::<T>::get(data_id).keys() {
 				Collection::<T>::mutate(collection_id, |collection| {
-					collection
-						.get_mut(data_id)
-						.map(|value| *value = Self::get(data_id))
+					if let Some(value) = collection.get_mut(data_id) {
```
thea-leake, line 162: I'm wondering if It might read a little cleaner flattened with match, though feel free to disregard if not :)
```
match (collection.get_mut(data_id), Self::get(data_id)) {
    (Some(value), Ok(new_value)) => *value = new_value,
    _ => ()
}
```

thea-leake (approved):
LGTM!

Issue #1350, opened 2023-05-17
feat: Extend `CustomMetadata` with Connectors-specific metadata


NunoAlexandre:
### Requirements

- Possible Connectors related information is derivable from the metadata

### Spec

**Adaption of the Metadata**
For this part of the change we need to change the existing struct.

```rust
pub struct XcmMetadata {
    /// The fee charged for every second that an XCM message takes to execute.
    /// When `None`, the `default_per_second` will be used instead.
    pub fee_per_second: Option<Balance>,
}

pub struct CustomMetadata {
    /// XCM-related metadata.
    pub xcm: XcmMetadata,

    /// Whether an asset can be minted.
    /// When `true`, the right permissions will checked in the permissions
    /// pallet to authorize asset minting by an origin.
    pub mintable: bool,

    /// Whether an asset is _permissioned_, i.e., whether the asset can only
    /// be transferred from and to whitelisted accounts. When `true`, the
    /// right permissions will checked in the permissions pallet to authorize
    /// transfer between mutually allowed from and to accounts.
    pub permissioned: bool,

    /// Whether an asset can be used as a currency to fund Centrifuge Pools.
    pub pool_currency: bool,
}

pub struct AssetMetadata<Balance, CustomMetadata: Parameter + Member + TypeInfo> {
    pub decimals: u32,
    pub name: Vec<u8>,
    pub symbol: Vec<u8>,
    pub existential_deposit: Balance,
    pub location: Option<VersionedMultiLocation>,
    pub additional: CustomMetadata,
}
```

We extend the existing metadata with the following struct:

```rust
pub struct ConnectorsMetadata {
    /// The "location" of the token on the 
    /// connectors domain. This can be used to actually
    /// derive both domain and also address for "adding"
    /// this currency to another domain via the 
    /// `AddCurrency` message.
    address: DomainAddress,
}

impl ConnectorsMetadata {
    fn location(&self) -> (Domain, [u8; 32]) {
        (self.address.domain(), self.address.address())
    }
}
```

and the `struct CustomMetadata` will subsequently be changed too

```rust
pub struct CustomMetadata {
    /// Connectors related metadata
    ///
    /// Not every currency variant has nor needs that.
    pub connectors: Option<ConnectorsMetadata>
    
    /// XCM-related metadata.
    pub xcm: XcmMetadata,

    /// Whether an asset can be minted.
    /// When `true`, the right permissions will checked in the permissions
[...]

PR #1357, opened 2023-05-23
feat: Extend CustomMetadata with `transferability`

NunoAlexandre:
Part 1/2 of https://github.com/centrifuge/centrifuge-chain/issues/1350

### Changes

- [x] Extend the CustomMetadata type with `transferability`
- [x] Drop `CustomMetadata.xcm` and place that in `transferability` instead (thanks for the suggestion @mustermeiszer 🚀 )
- [x] Storage Migration
  - [x] Centrifuge
  - [x] Altair
- [x] Adapt XCM config
  - [x] Centrifuge
  - [x] Altair


runtime/centrifuge/src/xcm.rs
```diff
@@ -203,11 +204,13 @@ pub struct CurrencyIdConvert;
...
-			_ => OrmlAssetRegistry::multilocation(&id).ok()?,
-			// todo(nuno): verify this will work correctly
-		}
+		OrmlAssetRegistry::metadata(id)
+			.filter(|m| {
+				m.additional
+					.transferability
+					.includes(AssetTransferability::Xcm)
+			})
+			.and_then(|_| OrmlAssetRegistry::multilocation(&id).ok()?)
```
NunoAlexandre, line 213: We can just write the conversion from `m.location: Option<VersionedMultiLocation` to `MultiLocation` here instead of calling `OrmlAssetRegistry::multilocation` which looks up the `location` field we already have and does said conversion

libs/types/src/tokens.rs
```diff
@@ -204,6 +204,48 @@ pub struct CustomMetadata {
...
+	/// The asset is only transferable through Centrifuge Connectors
+	Connectors,
+
+	/// The asset is transferable through all available options.
+	/// NOTE: Once set to `All`, the asset is automatically transferrable
+	/// through any eventual new option being added at a later stage. A
+	/// migration might be required if that's undesirable.
+	All,
+
+	/// The asset is not transferable
```
NunoAlexandre, line 241:
```suggestion
	/// The asset is not cross-chain transferable
```

NunoAlexandre: With that, we could actually either:

- a. drop `None` and make it `Local` to say the asset is only transferrable locally (and not cross-chain) and make this the `#[default]` variant 
- b. Rename `AssetTransferability` to `AssetCrossChainTransferability`

I find that `b.` would make more sense because all fungible assets are locally transferable by nature so it seems like an unnecessary wide scope to cover; just don't like the longer name but maybe someone has a better suggestion? 😇

runtime/centrifuge/src/xcm.rs
NunoAlexandre: Done

libs/types/src/tokens.rs
```diff
@@ -204,6 +204,48 @@ pub struct CustomMetadata {
...
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum CrossChainTransferability {
+	/// The asset is only transferable through XCM
+	Xcm,
+
+	/// The asset is only transferable through Centrifuge Connectors
+	Connectors,
+
+	/// The asset is transferable through all available options.
+	/// NOTE: Once set to `All`, the asset is automatically transferrable
```
wischli, line 235: Maybe we should move/repeat part of this comment for the (missing) docs comment of the enum `CrossChainTransferability`.

libs/types/src/tokens.rs
```diff
@@ -204,6 +204,48 @@ pub struct CustomMetadata {
...
+	All,
+
+	/// The asset is not transferable cross-chain
+	#[default]
+	None,
+}
+
+impl CrossChainTransferability {
+	pub fn includes(self, other: Self) -> bool {
+		self == other || self == Self::All
```
wischli, line 247: We must add test(s) as the current check is also positive when checking `CrossChainTransferability::All.includes(CrossChainTransferability::None)`.

runtime/centrifuge/src/migrations.rs
```diff
@@ -203,4 +212,135 @@ mod asset_registry {
...
+		orml_asset_registry::Pallet<T>,
+		Twox64Concat,
+		CurrencyId,
+		AssetMetadata<Balance, v0::CustomMetadata>,
+		OptionQuery,
+	>;
+
+	impl OnRuntimeUpgrade for AssetTransferabilityMigration {
+		fn on_runtime_upgrade() -> Weight {
+			orml_asset_registry::Metadata::<Runtime>::translate(
```
wischli, line 236: It would be great if we made this migration conditional, e.g. on some pallet version. If for some reason this migration is not removed before the next RU, it panics as it fails to decode the values of the storage map.

However, we cannot use the `orml_asset_registry` version as this must not be overwritten by us. Otherwise, we will have harder times running default (from ORML itself) `orml_asset_registry` migrations in the future.

We can take a look at this later as I don't have a solution at the top of my head. We should think about whether we want to introduce versioning to our custom metadata.

runtime/centrifuge/src/migrations.rs
```diff
@@ -203,4 +212,135 @@ mod asset_registry {
...
+						)),
+						_ => Some(to_metadata_v1(
+							old_metadata,
+							v1::CrossChainTransferability::Xcm,
+						)),
+					}
+				},
+			);
+
+			Weight::from_ref_time(200_000_000)
```
wischli, line 252: Eventually, we should return the weight based on the number of read and written items, e.g. like

```suggestion
			let n = orml_asset_registry::Metadata::<Runtime>::iter().count();
			T::DbWeight::get().reads_writes(n, n)
```

runtime/centrifuge/src/migrations.rs
```diff
@@ -203,4 +212,135 @@ mod asset_registry {
...
+
+			for (asset_id, old_metadata) in old_state {
+				let new_metadata = OrmlAssetRegistry::metadata(asset_id)
+					.ok_or_else(|| "New state lost the metadata of an asset")?;
+
+				match asset_id {
+					CurrencyId::Tranche(_, _) => ensure!(new_metadata == to_metadata_v1(
+						old_metadata,
+						v1::CrossChainTransferability::Connectors,
+					), "The metadata of a tranche token wasn't just updated by setting `transferability` to `Xcm`"),
```
wischli, line 283:
```suggestion
					), "The metadata of a tranche token wasn't just updated by setting `transferability` to `Connectors `"),
```

libs/types/src/tokens.rs
```diff
@@ -204,6 +204,48 @@ pub struct CustomMetadata {
 
 	/// Whether an asset can be used as a currency to fund Centrifuge Pools.
 	pub pool_currency: bool,
+
+	/// The ways this token is cross-chain transferable
```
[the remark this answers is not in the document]
NunoAlexandre: But the comment about is referring to a `bool`, where `whether` makes sense. `transferability` is not _just_ "whether" it's cross-chain transferable but the ways that it might be so.

runtime/centrifuge/src/migrations.rs
NunoAlexandre: Thanks for the catch!

libs/types/src/tokens.rs
NunoAlexandre: What is potentially wrong with that expression being `true`? I can't think of any use case that could turn out dangerous or erroneous 🤔

libs/types/src/tokens.rs
NunoAlexandre: Done, let me me know how you find it now 👍

libs/types/src/tokens.rs
NunoAlexandre: Oh, I see. `All` should not imply `None`, so this indeed undesired from a logical perspective. Will tackle!

runtime/centrifuge/src/migrations.rs
NunoAlexandre: Thanks for the thoughts here. Can't we just use the runtime version the migration is running on to ensure 1) that it's being executed against the correct version and 2) that it won't be running again even if not deleted for a follow up RTU?

libs/types/src/tokens.rs
NunoAlexandre: @wischli I've made `transferability` option: https://github.com/centrifuge/centrifuge-chain/pull/1357/commits/2dc11a9b9585d776016ddcf1aeb4fc005ce4be26

runtime/centrifuge/src/migrations.rs
NunoAlexandre: Thanks William!

runtime/centrifuge/src/migrations.rs
wischli: Great idea. I think that totally suffices!

NunoAlexandre:
@wischli thank you very much for your careful first review. I have addressed everything and made this PR ready for a proper final review.

I will copy the migration to the Altair runtime once it's been approved for Centrifuge 👍

libs/types/src/tokens.rs
```diff
@@ -210,6 +199,61 @@ pub struct CustomMetadata {
...
+	Debug,
+	Encode,
+	Decode,
+	TypeInfo,
+	MaxEncodedLen,
+)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum CrossChainTransferability {
+	/// The asset is only transferable through XCM
+	#[default]
```
lemunozm, line 243: Is this default used only for tests/benchmarks, or does it make sense to have `Xcm` as the expected/default behavior in production?

I say this because it's easier to initialize this value without realizing it by default with a non-specified value in production code.

NunoAlexandre: Both, actually. By default, a token registered in the registry is XCM transferrable unless set otherwise. And for tests it's also very useful to have this default precisely because that's the default property.

runtime/centrifuge/src/xcm.rs
```diff
@@ -194,18 +195,22 @@ where
...
+		OrmlAssetRegistry::metadata(id)
+			.filter(|m| {
+				m.additional
+					.transferability
+					.map_or(false, |t| t.includes(CrossChainTransferability::Xcm))
+			})
+			.and_then(|m| m.location)
+			.and_then(|l| l.try_into().ok())
 	}
 }
```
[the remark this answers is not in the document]
NunoAlexandre: It does but I would like to do that in a separate PR alongside with a couple of other bits that we could dry

pallets/pool-system/src/tranches.rs
```diff
@@ -251,6 +254,7 @@ where
 				xcm: XcmMetadata {
 					fee_per_second: None,
 				},
+				transferability: Some(CrossChainTransferability::Connectors),
```
mustermeiszer, line 257: What does it again mean, if a token in the registry has no `location` set. I mean, in our system?

runtime/integration-tests/src/xcm/polkadot/tests/currency_id_convert.rs
```diff
@@ -131,37 +131,15 @@ fn convert_cfg_xcm_v2() {
...
-	});
+		register_no_xcm_token();
 
-	Centrifuge::execute_with(|| {
 		assert_eq!(
-			<CurrencyIdConvert as C2<_, _>>::convert(tranche_currency),
+			<CurrencyIdConvert as C2<_, _>>::convert(NO_XCM_ASSET_ID),
 			None
 		)
 	});
```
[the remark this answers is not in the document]
NunoAlexandre: Yes! Note how we registered the "no xcm token" and then verify that the conversion fails; the conversion is a key step in processing an XCM transfer since it tries to convert `CurrencyId` to `MultiLocation` for an outbound transfer and the reverse in an incoming one; we don't need to test the incoming case because we can't receive tokens that were not sent out in the first place.

I could write an explicit test showing that xtokens.transfer fails for a "no xcm token", would you like to see that? I didn't do that to avoid redundancy but let me know how you see it

runtime/centrifuge/src/migrations.rs
```diff
@@ -9,5 +9,164 @@
...
+
+			let old_state = sp_std::vec::Vec::<(
+				CurrencyId,
+				AssetMetadata<Balance, v0::CustomMetadata>,
+			)>::decode(&mut old_state_encoded.as_ref())
+			.map_err(|_| "Error decoding pre-upgrade state")?;
+
+			for (asset_id, old_metadata) in old_state {
+				let new_metadata = OrmlAssetRegistry::metadata(asset_id)
+					.ok_or_else(|| "New state lost the metadata of an asset")?;
```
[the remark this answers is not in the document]
NunoAlexandre: Eagle eyes! If won't return an error because we don't have (and assume that maybe too implicitly?) any `CurrencyId::Staking` registered in the registry, so that edge case should never occur.

pallets/pool-system/src/tranches.rs
NunoAlexandre: So, for a token to be transferrable through XCM it **must** have a `location`. If it doesn't have, and it supposed to be XCM-transferrable, it will not be actually transferrable; for Connectors-wrapped tokens, we discussed using the `location` as a type to encapsulate the corresponding EVM address of the token and it's evm chain id; so in that case having a `location` set, which would require having  `transferability: Some(CrossChainTransferability::Connectors)`, it would mean that the location served the purpose of holding that information that connectors would know how to unwrap.

Does that clarify or does it make it more confusing?

libs/types/src/tokens.rs
```diff
@@ -210,6 +199,61 @@ pub struct CustomMetadata {
...
+	fn default() -> Self {
+		Self {
+			xcm: Default::default(),
+			mintable: false,
+			permissioned: false,
+			pool_currency: false,
+			transferability: Some(CrossChainTransferability::Xcm),
+		}
+	}
+}
```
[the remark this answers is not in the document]
NunoAlexandre: I mean, by default tokens are transferable. The only exception to that, to my knowledge, is the `CurrencyId::Staking` tokens. And within that, XCM is the default because it's the general status quo in Polkadot, whereas Connectors is an ad-hoc, Centrifuge-specific solution.

In practise, this is most useful for testing indeed, because this default is never imposed anywhere else to my knowledge.

libs/types/src/tokens.rs
mustermeiszer: > I am open to that, and thanks for the suggestion, but do you have an example of where it's dangerous to have the default being Some(XCM)? Because having the default set to None is neither helpful for testing purposes neither useful in practise. Let me know 👍

Lets say we add registry default entries for all currencies and then test networks will be different. And it feels like too much allowance for a default to be transferable over xcm. But maybe just me not trusting xcm setups. 

Again, if it is only a default for testing, I would hide it behind std feature, so it does not hit the wasm.

libs/types/src/tokens.rs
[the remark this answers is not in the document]
mustermeiszer: See above :)

runtime/centrifuge/src/migrations.rs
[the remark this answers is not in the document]
mustermeiszer: Should be fine then!

libs/types/src/tokens.rs
mustermeiszer: But not blocking of course. Just the typical reviewing eye ^^

runtime/integration-tests/src/xcm/polkadot/tests/currency_id_convert.rs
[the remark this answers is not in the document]
mustermeiszer: > we don't need to test the incoming case because we can't receive tokens that were not sent out in the first place.

But we can receive foreign assets that never hit our chain before, right?

libs/types/src/tokens.rs
```diff
@@ -210,6 +199,61 @@ pub struct CustomMetadata {
 
 	/// Whether an asset can be used as a currency to fund Centrifuge Pools.
 	pub pool_currency: bool,
+
+	/// The ways, if any, this token is cross-chain transferable
+	pub transferability: Option<CrossChainTransferability>,
```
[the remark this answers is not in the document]
NunoAlexandre: That doesn't work since it wouldn't allow for that value / XCM metadata to be set for CrossChai Transferability::All, unless if we would also have XcmMetadata as a value for that variant, but that seems a bit odd.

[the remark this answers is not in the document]
NunoAlexandre: Wdyt?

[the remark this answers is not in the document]
wischli: It actually works, happy to push my commit which moves `XcmMetadata` into `CrossChainTransferability::XCM`, e.g. ... with fixes

```rust
pub enum CrossChainTransferability {
	/// The asset is only transferable through XCM
	Xcm(XcmMetadata),

	/// The asset is only transferable through Centrifuge Connectors
	Connectors,

	/// The asset is transferable through all available options
	All,
}

impl Default for CrossChainTransferability {
	fn default() -> Self {
		Self::Xcm(Default::default())
	}
}
```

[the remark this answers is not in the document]
NunoAlexandre: @wischli The issue that I am describing above with this approach is that if we set a token to `CrossChainTransferability::All`, which presumes that the asset at hand is also XCM transferable, we won't have a `XcmMetadata` to retrieve anymore; we will only have that if we _also_ have `XcmMetadata` as a field of `CrossChainTransferability::All`:

``` rust
pub enum CrossChainTransferability {
	/// The asset is only transferable through XCM
	Xcm(XcmMetadata),

	/// The asset is only transferable through Centrifuge Connectors
	Connectors,

	/// The asset is transferable through all available options
	All(XcmMetadata), // HERE
}
```

I said above that this alternative seems a bit awkward but coming back at this today I can't really find a good argument as to why it's awkward so if it sounds good to you both, it sounds good to me :)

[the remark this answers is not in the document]
wischli: Ah good point, I was blinding that out. I would still prefer this over having the xcm metadata separated. Going this path, we also ensure we need to run migrations if we add more variants to `CrossChainTransferability` with additional metadata that needs to be added to `All`.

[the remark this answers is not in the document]
wischli: Not really. It's true I am dependent on this PR but I won't finish before anyways.

wischli (commented):
I agree with @mustermeiszer that defaulting to XCM transferability could be dangerous ([thread](https://github.com/centrifuge/centrifuge-chain/pull/1357/files#r1219849466)). But definitely not a blocker!

Everything else looks good to me, especially the migration. Please note that this still needs to be copied over to Altair.

libs/types/src/tokens.rs
NunoAlexandre: No, this is a good point. I added the `None` variant back to `CrossChainTransferability` and have that being the default.

wischli (dismissed):
LGTM! Thanks bringing back the default `None` version!

libs/types/src/tokens.rs
```diff
@@ -212,6 +212,49 @@ pub struct CustomMetadata {
...
+	/// The asset is only transferable through Centrifuge Connectors
+	Connectors,
+
+	/// The asset is transferable through all available options
+	All(XcmMetadata),
+}
+
+impl CrossChainTransferability {
+	pub fn includes_xcm(self) -> bool {
+		self != Self::None && matches!(self, Self::Xcm(..) | Self::All(..))
```
mikiquantum, line 254: Seems that the `matches!(self, Self::Xcm(..) | Self::All(..))` already covers the !Self::None case right? We might be able to simplify to: 
```suggestion
		matches!(self, Self::Xcm(..) | Self::All(..))
```

pallets/pool-system/src/tranches.rs
```diff
@@ -248,9 +248,7 @@ where
 				mintable: false,
 				permissioned: true,
 				pool_currency: false,
-				xcm: XcmMetadata {
-					fee_per_second: None,
-				},
+				transferability: CrossChainTransferability::Connectors,
```
mikiquantum, line 251: Just for my own education, tranche tokens (for now) will not be xcm transferable?

runtime/altair/src/migrations.rs
```diff
@@ -9,5 +9,163 @@
...
+		AssetMetadata<Balance, v0::CustomMetadata>,
+		OptionQuery,
+	>;
+
+	impl OnRuntimeUpgrade for CrossChainTransferabilityMigration {
+		fn on_runtime_upgrade() -> Weight {
+			// todo(nuno): this migration will **most likely** be part of the set of
+			// migrations when upgrading to Centrifuge 1020. Refine when we know
+			// for sure.
+			if VERSION.spec_version != 1020 {
```
mikiquantum, line 55: I guess this would be updated accordingly to the Altair spec version once we know the final version

runtime/centrifuge/src/migrations.rs
```diff
@@ -9,5 +9,163 @@
...
+		AssetMetadata<Balance, v0::CustomMetadata>,
+		OptionQuery,
+	>;
+
+	impl OnRuntimeUpgrade for CrossChainTransferabilityMigration {
+		fn on_runtime_upgrade() -> Weight {
+			// todo(nuno): this migration will **most likely** be part of the set of
+			// migrations when upgrading to Centrifuge 1020. Refine when we know
+			// for sure.
+			if VERSION.spec_version != 1020 {
```
mikiquantum, line 55: 👍

runtime/centrifuge/src/migrations.rs
```diff
@@ -9,5 +9,163 @@
...
+						)),
+						_ => Some(to_metadata_v1(
+							old_metadata.clone(),
+							v1::CrossChainTransferability::Xcm(old_metadata.additional.xcm),
+						)),
+					}
+				},
+			);
+
+			let n = orml_asset_registry::Metadata::<Runtime>::iter().count() as u64;
```
mikiquantum, line 75: I assume that the number of items is low, or empty, in Centrifuge right?

mikiquantum (commented):
Code & migration looks good! I have only some educational questions and simplification of the `includes_xcm` check.

runtime/centrifuge/src/migrations.rs
NunoAlexandre: It's low for sure :) just a few entries now

runtime/altair/src/migrations.rs
NunoAlexandre: Ah, good catch! I copied this over from the Centrifuge runtime, we have to adjust this

pallets/pool-system/src/tranches.rs
NunoAlexandre: That's right, they won't be transferable barebones through XCM since that way we can't ensure the correct permissions are respected. So tranche tokens will only be transferrable through Connectors; connectors might use xcm as the transport layer, as our POC with Moonbeam does, but it's not a direct xcm transfer.

libs/types/src/tokens.rs
NunoAlexandre: Thank you 😇 the reason why I was checking that it wasn't None was because of legacy. We initially had the None variant and a function includes(other: CrosschainTransferability) but because now that we added None again but the XCM variant has a value, it wasn't ergonomic to have a function with that signature, so we chose for includes_xcm(self) and uses matches!; But with that I forgot that we no longer have to check for None. Thanks again ✌🏻

runtime/altair/src/migrations.rs
NunoAlexandre: Done, thanks again!

mustermeiszer (approved):
LGTM!

mikiquantum (approved):
🙏

PR #1311, opened 2023-04-11
Oracle valuation implementation

lemunozm:
# Description

🚧 Under construction 🚧 

Epic #1279

## Changes and Descriptions
- `pallet-loans`
  - [x] #1305 
  - [ ] Update loans to support oracle prices
  - [ ] Tests using oracle valuations
  - [ ] Update `pallet-loans` benchmarks
  - [ ] Write migration
- Runtime
  - [ ] Add `orml-oracle` to `development`
  - [ ] Add `pallet-pool-price-feed` to `development`
  - [ ] Adapt `pallet-loans` in `development`
- Integration test:
  - [ ] End to End tests using oracles to update prices.
 
### Prerequisites
- #1324

PR #1359, opened 2023-05-23
Fix default timestamp for portfolio valuation

lemunozm:
# Description

The `last_updated` field of portfolio valuation is only updating when the valuations are computed. This is ok, nevertheless, if a loan is added to a pool, and modified, it will affect the portfolio valuation without updating the `last_updated`, which by default will be created with `0` instead of `now()`

This means that if some entity call `nav()`, before computing once the portfolio valuation, it will obtain an incorrect time (?). Not sure if this is a possible case or if we always call the `update_portfolio_valuation()` at least once.

Until I see it, the `pool-system` is protected when closing the epoch giving an `NAVTooOld` error, but that means the epoch can not be closed (unless `update_porfolio_valuation()` is called).

If it is a security issue, I understand we need to deliver a fix soon. In case not, and we can live with it, this will be fixed properly in the [Oracle Valuation PR](1311)

### DIFF
--- a/pallets/loans-ref/src/types.rs
+++ b/pallets/loans-ref/src/types.rs
@@ -13,19 +13,16 @@
 
 use cfg_primitives::{Moment, SECONDS_PER_DAY};
 use cfg_traits::{
-	ops::{EnsureAdd, EnsureAddAssign, EnsureFixedPointNumber, EnsureInto, EnsureMul, EnsureSub},
+	ops::{
+		EnsureAdd, EnsureAddAssign, EnsureFixedPointNumber, EnsureInto, EnsureMul, EnsureSub,
+		EnsureSubAssign,
+	},
 	InterestAccrual, RateCollection,
 };
 use cfg_types::adjustments::Adjustment;
 use codec::{Decode, Encode, MaxEncodedLen};
 use frame_support::{
-	ensure,
-	pallet_prelude::DispatchResult,
-	traits::{
-		tokens::{self},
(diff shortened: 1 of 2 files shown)


pallets/loans-ref/src/types.rs
```diff
@@ -83,27 +80,35 @@ pub enum CloseLoanError {
...
-{
-	pub fn new(value: Balance, when: Moment) -> Self {
+impl<T: Config> Default for PortfolioValuation<T> {
+	fn default() -> Self {
+		Self {
+			value: T::Balance::zero(),
+			last_updated: T::Time::now().as_secs(),
+		}
+	}
+}
```
lemunozm, line 101: No longer created with 0 by default

lemunozm:
@mustermeiszer @offerijns It's not clear what should be the `last_updated` initial value. 
In the original `pallet-loans` implementation, it's initialized in the `initialise_pool()` extrinsic, but we do not have such a call. That value is defined as "last time when nav is performed for the entire pool", but what if it has never been called? Should we return `None` instead or the time of first borrowing in the pool?

mustermeiszer (approved):

mustermeiszer:
I agree it should be the created time.

lemunozm:
The intent of this commit is to add a fast fix, but if we can wait until the next release where Oracle Valuation fixes this well is better not to merge this PR

lemunozm:
Close it because the fix is already in #1311

PR #1362, opened 2023-05-24
mock-builder: Main documentation

lemunozm:
# Description

This PR adds the main documentation with examples for understanding and using `mock-builder`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -31,19 +31,19 @@ jobs:
         with:
           command: doc
           args: --all --no-deps
-      
+
       - name: Build Documentation failed
         if: always() && steps.build_docs.outcome == 'failure'
         run: echo ":::error::cargo doc --all --no-deps failed"
         # Job will stop here and the check will be red if Build documentation failed
 
       - name: Create Index file
-        if: github.ref == 'ref/head/main'
+        if: github.ref == 'refs/heads/main'
         run: ./ci/create_index_for_rust_docs.sh
 
       - name: Deploy Docs
         uses: peaceiris/actions-gh-pages@v3
-        if: github.ref == 'ref/head/main'
+        if: github.ref == 'refs/heads/main'
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_branch: gh-pages
--- a/libs/mock-builder/src/lib.rs
+++ b/libs/mock-builder/src/lib.rs
@@ -29,19 +29,19 @@
 //! the associated types of your pallet `Config`.
 //!
 //! Then, you are mostly forced to use other pallet configurations
-//! geting a [tight coupling](https://docs.substrate.io/build/pallet-coupling/)
+//! getting a [tight coupling](https://docs.substrate.io/build/pallet-coupling/)
 //! with them. It has some downsides:
 //! - You need to learn how to configure other pallets.
 //! - You need to know how those pallets work, because they affect directly the
 //!   behavior of the pallet you're testing.
 //! - The way they work can give you non-completed tests. It means that some
-//!   paths of your pallet can no be tested because some dependency works in a
+//!   paths of your pallet can not be tested because some dependency works in a
 //!   specific way.
 //! - You need a lot of effort maintaining your tests because each time one
 //!   dependency changes, it can easily break your tests.
 //!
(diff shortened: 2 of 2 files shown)


libs/mock-builder/src/lib.rs
```diff
@@ -11,10 +11,139 @@
...
+//! }
+//!
+//! trait TraitB {
+//!     type AssocB;
+//!
+//!     fn bar(a: u64, b: Self::AssocB) -> u32;
+//! }
+//! ```
+//!
+//! We have a real huge pallet that implements a specific behavior for those
```
mustermeiszer, line 74:
```suggestion
//! We have a really huge pallet that implements a specific behavior for those
```

libs/mock-builder/src/lib.rs
```diff
@@ -11,10 +11,139 @@
...
+//! traits, but we want to get rid of such dependency so we [generate a *mock
```
mustermeiszer, line 75:
```suggestion
//! traits, but we want to get rid of such dependency so we generate a [*mock
```

mustermeiszer (approved):
LGTM! Thanks for pushing this up in your backlog.

libs/mock-builder/src/lib.rs
lemunozm: I wanted to pick up the `generate` part because I'm redirecting to the creation section

.github/workflows/docs.yml
```diff
@@ -31,19 +31,19 @@ jobs:
...
 
       - name: Create Index file
-        if: github.ref == 'ref/head/main'
+        if: github.ref == 'refs/heads/main'
         run: ./ci/create_index_for_rust_docs.sh
 
       - name: Deploy Docs
         uses: peaceiris/actions-gh-pages@v3
-        if: github.ref == 'ref/head/main'
+        if: github.ref == 'refs/heads/main'
```
lemunozm, line 46: As talked with @gpmayorga, we had our doc deploy job broken. This change fixes it.

libs/mock-builder/src/lib.rs
```diff
@@ -93,7 +289,7 @@ where
 }
 
 /// Register a mock function into the mock function storage.
-/// Same as `register()` but with using the locator who calls this macro.
+/// Same as `register()` but it uses as locator who calls this macro.
```
wischli, line 292: Should this mean that it is use**d as** a locator or that it use**s a** locator?

wischli (commented):
Stellar documentation! I put on my language nitpicker head and made some suggestions. Feel free to ignore any of them.

wischli (commented):
Stellar documentation! I put on my language nitpicker hat and made some suggestions. Feel free to ignore any of them.

libs/mock-builder/src/lib.rs
lemunozm: 🤔
The macro generates a locator parameter that is passed to the `register()` function, That locator represents the position of the caller in the code. So I think "it uses as locator" is ok, but not sure if I get to transmit the idea correctly

wischli: Thanks for clariyfing. Then let's leave it as is.

wischli (approved):
Thanks for applying my suggestions! LGTM

NunoAlexandre (approved):
Nit comments, approving anwyays. This is awesome and it reminded me of your presentation on the offsite :) Nice work!

libs/mock-builder/src/lib.rs
```diff
@@ -11,10 +11,139 @@
...
+//! Nevertheless, those dependencies still exist in testing because when the
+//! `mock.rs` file is defined, you're forced to give some implementations for
+//! the associated types of your pallet `Config`.
+//!
+//! Then, you are mostly forced to use other pallet configurations
+//! getting a [tight coupling](https://docs.substrate.io/build/pallet-coupling/)
+//! with them. It has some downsides:
+//! - You need to learn how to configure other pallets.
+//! - You need to know how those pallets work, because they affect directly the
+//!   behavior of the pallet you're testing.
```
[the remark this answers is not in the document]
lemunozm: Yeah I think it's more formal. I write it down these comments for the next doc update. Thanks!

PR #1363, opened 2023-05-25
connectors: Add new `Executed*` messages

### LINKED ISSUE #1356: Connectors v2: Add `ExecutedCollect*` messages
### Description

During our multi chain sync, we realised that the this flow doesn't work because we have no way to link the `Transfer` message to a specific pool, which is necessary for the https://github.com/centrifuge/connectors/pull/66 work taking place on the Solidity side.

The current flow:

1. Message `CollectInvest` | `CollectRedeem` is sent from EVM -> Centrifuge Chain
2. The message is received and processed on the Centrifuge chain
3. On epoch execution, `Transfer` would be sent from Centrifuge Chain to the EVM

### Changes

A. Introduce two new message types which will essentially be a specialised version of `Transfer`:

1. `ExecutedCollectInvest`
2. `ExecutedCollectRedeem`

Both of these messages will contain:

- pool id
- tranche id
- currency id
- executed amount (how much was invested or redeemed, respectively)
- investor's address

B. Send the appropriate message back to the EVM on epoch execution instead of `Transfer`

NunoAlexandre:
Closes #1356

### DIFF
--- a/pallets/connectors/src/message.rs
+++ b/pallets/connectors/src/message.rs
@@ -201,6 +201,93 @@ where
 		tranche_id: TrancheId,
 		investor: Address,
 	},
+	/// The message sent back to the domain from which a `DecreaseInvestOrder`
+	/// message was received, ensuring the correct state update on said domain
+	/// and that the `investor`'s wallet is updated accordingly.
+	///
+	/// Directionality: Centrifuge -> EVM Domain.
+	ExecutedDecreaseInvestOrder {
+		/// The pool id
+		pool_id: PoolId,
+		/// The tranche id
+		tranche_id: TrancheId,
+		/// The investor's address
+		investor: Address,
+		/// The currency in which `DecreaseInvestOrder` was realised
+		currency: u128,
+		/// The amount of `currency` that was actually executed in the original
+		/// `DecreaseInvestOrder` message, i.e., the amount by which the
+		/// investment order was actually decreased by.
+		currency_payout: Balance,
+		/// The outstanding order, in `currency` units
+		remaining_invest_order: Balance,
+	},
+	/// The message sent back to the domain from which a `DecreaseRedeemOrder`
+	/// message was received, ensuring the correct state update on said domain
+	/// and that the `investor`'s wallet is updated accordingly.
+	///
+	/// Directionality: Centrifuge -> EVM Domain.
+	ExecutedDecreaseRedeemOrder {
+		/// The pool id
+		pool_id: PoolId,
+		/// The tranche id
+		tranche_id: TrancheId,
+		/// The investor's address
+		investor: Address,
+		/// The currency in which `DecreaseRedeemOrder` was realised
+		currency: u128,
+		/// The amount of tranche tokens that was actually executed in the
+		/// original `DecreaseRedeemOrder` message, i.e., the amount by which
+		/// the redeem order was actually decreased by.
+		tranche_tokens_payout: Balance,
+		/// The remaining amount of tranche tokens the investor still has locked
+		/// to redeem at a later epoch execution
+		remaining_redeem_order: Balance,
+	},
+	/// The message sent back to the domain from which a `CollectInvest` message
+	/// has been received, which will ensure the `investor` gets the payout
+	/// respective to their investment.
+	///
+	/// Directionality: Centrifuge -> EVM Domain.
+	ExecutedCollectInvest {
+		/// The pool id
+		pool_id: PoolId,
+		/// The tranche
+		tranche_id: TrancheId,
+		/// The investor's address
+		investor: Address,
+		/// The currency in which the investment was realised
+		currency: u128,
+		/// The amount that was actually collected
+		currency_payout: Balance,
+		/// The amount of tranche tokens received for the investment made
+		tranche_tokens_payout: Balance,
+		/// The remaining amount of `currency` the investor still has locked to
+		/// invest at a later epoch execution
+		remaining_invest_order: Balance,
+	},
+	/// The message sent back to the domain from which a `CollectRedeem` message
+	/// has been received, which will ensure the `investor` gets the payout
+	/// respective to their redemption.
+	///
+	/// Directionality: Centrifuge -> EVM Domain.
+	ExecutedCollectRedeem {
+		/// The pool id
+		pool_id: PoolId,
+		/// The tranche id
+		tranche_id: TrancheId,
+		/// The investor's address
+		investor: Address,
+		/// The stable coin currency in which the payout takes place
+		currency: u128,
+		/// The amount of `currency` being paid out to the investor
+		currency_payout: Balance,
+		/// How many tranche tokens were actually redeemed
+		tranche_tokens_payout: Balance,
+		/// The remaining amount of tranche tokens the investor still has locked
+		/// to redeem at a later epoch execution
+		remaining_redeem_order: Balance,
+	},
 }
 
 impl<
@@ -234,6 +321,10 @@ impl<
 			Self::DecreaseRedeemOrder { .. } => 12,
 			Self::CollectInvest { .. } => 13,
 			Self::CollectRedeem { .. } => 14,
+			Self::ExecutedDecreaseInvestOrder { .. } => 15,
+			Self::ExecutedDecreaseRedeemOrder { .. } => 16,
+			Self::ExecutedCollectInvest { .. } => 17,
+			Self::ExecutedCollectRedeem { .. } => 18,
 		}
 	}
 }
@@ -415,6 +506,82 @@ impl<
 				self.call_type(),
 				vec![encode_be(pool_id), tranche_id.encode(), address.to_vec()],
 			),
+			Message::ExecutedDecreaseInvestOrder {
+				pool_id,
+				tranche_id,
+				investor,
+				currency,
+				currency_payout,
+				remaining_invest_order,
+			} => encoded_message(
+				self.call_type(),
+				vec![
+					encode_be(pool_id),
+					tranche_id.encode(),
+					investor.to_vec(),
+					encode_be(currency),
+					encode_be(currency_payout),
+					encode_be(remaining_invest_order),
+				],
+			),
+			Message::ExecutedDecreaseRedeemOrder {
+				pool_id,
+				tranche_id,
+				investor,
+				currency,
+				tranche_tokens_payout,
+				remaining_redeem_order,
+			} => encoded_message(
+				self.call_type(),
+				vec![
+					encode_be(pool_id),
+					tranche_id.encode(),
+					investor.to_vec(),
+					encode_be(currency),
+					encode_be(tranche_tokens_payout),
+					encode_be(remaining_redeem_order),
+				],
+			),
+			Message::ExecutedCollectInvest {
+				pool_id,
+				tranche_id,
+				investor,
+				currency,
+				currency_payout,
+				tranche_tokens_payout,
+				remaining_invest_order,
+			} => encoded_message(
+				self.call_type(),
+				vec![
+					encode_be(pool_id),
+					tranche_id.encode(),
+					investor.to_vec(),
+					encode_be(currency),
+					encode_be(currency_payout),
+					encode_be(tranche_tokens_payout),
+					encode_be(remaining_invest_order),
+				],
+			),
+			Message::ExecutedCollectRedeem {
+				pool_id,
+				tranche_id,
+				investor,
+				currency,
+				currency_payout,
+				tranche_tokens_payout,
+				remaining_redeem_order,
+			} => encoded_message(
+				self.call_type(),
+				vec![
+					encode_be(pool_id),
+					tranche_id.encode(),
+					investor.to_vec(),
+					encode_be(currency),
+					encode_be(currency_payout),
+					encode_be(tranche_tokens_payout),
+					encode_be(remaining_redeem_order),
+				],
+			),
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


NunoAlexandre:
@mustermeiszer based on your follow up discussion with Alina and considering what @wischli said on the associated issue, would it be possible to only have one message to serve the purpose of paying an investor out and inform of the _remaining_, for all operations (invest/redeem etc)? Or do we benefit from or need to have separate `ExecuteCollect*` messages for each flow?

As I said in the issue, I rather not fall into the trap of early abstraction but considering that we were going to use the very generic `Transfer` message for all those flows, I guess that one "Executed*` message should suffice?

offerijns:
The difference is that a `ExecutedCollectInvest` is transferring (pool) currency but `ExecutedCollectRedeem` is transferring tranche tokens. So I would think it makes sense to separate these.

NunoAlexandre:
> The difference is that a `ExecutedCollectInvest` is transferring (pool) currency but `ExecutedCollectRedeem` is transferring tranche tokens. So I would think it makes sense to separate these.

Ah, thanks. I missed that bit, good to know! In that case the PR is ready 👍

thea-leake (approved):
LGTM, I really appreciate the explanations in the `Message` enum :).

Happy to re-approve if needed for rustdoc job.

NunoAlexandre:
> LGTM, I really appreciate the explanations in the `Message` enum :).
> 
> Happy to re-approve if needed for rustdoc job.

Thank you so much for taking a look, Thea! I am working on the rust doc issue on the 0.9.38 PR 👀  will ping you as needed!

pallets/connectors/src/message.rs
```diff
@@ -201,6 +201,39 @@ where
...
+		pool_id: PoolId,
+		tranche_id: TrancheId,
+		investor: Address,
+		/// The currency in which the payout takes place
+		currency: u128,
+		/// The amount being paid out to the investor
+		payout: Balance,
+		/// The remaining amount the investor can still collect
+		remaining: Balance,
+	},
```
mustermeiszer, line 236: The docs are not correct and if we implement it like said in the docs it is not the expected information that this message should signal. Especially, for `ERC-4626`. 

Please [see ](https://github.com/centrifuge/centrifuge-chain/issues/1356)

mustermeiszer: This is why I would be really specific with the names.

NunoAlexandre: > For sure. It is not the remaining amount one can still collect, but rather the amount one has still "locked", either stable coin for investing or tranche token for redeeming.

Right, so that was my understanding as well. When I wrote "... the investor can still collect" I had in mind that it was an amount that would be executed at a later epoch. But if you are absolutely right that it's not the right choice of words.

> This is why I would be really specific with the names

I want to find common ground here, so let's try to work this out. For _me_,  e.g. `remaining_investment_invest` is not any clearer, it's just repeating words that are already in context (`Invest` in `ExecutedCollectInvest`, same for `Redeem` in `ExecutedCollectRedeem`).

Would it help to name it `remaining_locked_investment` and `remaining_locked_redeem`? Don't mean to introduce new language into the table but usually the language used when explaining this stuff low key to each other is usually helpful. Let me know your thoughts 👍

NunoAlexandre: And thanks for the careful review dissecting the `///` docs, it's really important to get all this stuff right both from a code and from an understanding perspective!

mustermeiszer: Naming changes sound good!

offerijns: I think `currency` is also redundant for `ExecutedCollectInvest`, since the currency that's being paid out is the tranche token, which is already implicit through the `pool_id` and `tranche_id` parameters.

mustermeiszer (approved):
LGTM

pallets/connectors/src/message.rs
```diff
@@ -201,6 +201,39 @@ where
...
+	ExecutedCollectRedeem {
+		pool_id: PoolId,
+		tranche_id: TrancheId,
+		investor: Address,
+		/// The currency in which the payout takes place
+		currency: u128,
+		/// The amount being paid out to the investor
+		payout: Balance,
+		/// The remaining amount the investor still has locked
+		remaining_locked_redeem: Balance,
```
wischli, line 235: I agree with @offerijns that we should keep the naming consistent, e.g. one of the following

> * `remaining_locked_investment` / `remaining_locked_redemption`
> * `remaining_locked_invest` / `remaining_locked_redeem`

wischli (commented):
I would appreciate if we could make the `remaining_*` variants consistent in their naming. Definitely not a blocker.

pallets/connectors/src/message.rs
NunoAlexandre: > I think `currency` is also redundant for `ExecutedCollectInvest`, since the currency that's being paid out is the tranche token, which is already implicit through the `pool_id` and `tranche_id` parameters.

@mustermeiszer can you comment on this? 👀

wischli (dismissed):
Thank for renaming - LGTM!

NunoAlexandre:
Moving this back to draft since we have further points around these messages to discuss later this week

cdamian (approved):

pallets/connectors/src/message.rs
```diff
@@ -201,6 +201,90 @@ where
...
+		investor: Address,
+		/// The currency in which `DecreaseInvestOrder` was realised
+		currency: u128,
+		/// The amount of `currency` that was actually executed in the original
+		/// `DecreaseInvestOrder` message, i.e., the amount by which the
+		/// investment order was actually decreased by.
+		decreased_by: Balance,
+		/// The outstanding order, in `currency` units
+		outstanding_order: Balance,
+	},
```
NunoAlexandre, line 224: Based on

> InvestmentDecreased
>  - How much is the outstanding order
>  - By How much it was decreased
>  - Investor’s address

pallets/connectors/src/message.rs
```diff
@@ -201,6 +201,90 @@ where
...
+	/// The message sent back to the domain from which a `DecreaseRedeemOrder`
+	/// message was received, ensuring the correct state update on said domain
+	/// and that the `investor`'s wallet is updated accordingly.
+	///
+	/// Directionality: Centrifuge -> EVM Domain.
+	ExecutedDecreaseRedeemOrder {
```
NunoAlexandre, line 230: Based on:

> RedemptionDecreased
>  - How much is the outstanding order
>  - By How much it was decreased
>  - Investor’s address

pallets/connectors/src/message.rs
```diff
@@ -201,6 +201,90 @@ where
...
+		decreased_by: Balance,
+		/// The outstanding order, in `currency` units
+		outstanding_order: Balance,
+	},
+	/// The message sent back to the domain from which a `CollectInvest` message
+	/// has been received, which will ensure the `investor` gets the payout
+	/// respective to their investment.
+	///
+	/// Directionality: Centrifuge -> EVM Domain.
+	ExecutedCollectInvest {
```
NunoAlexandre, line 251: Based on

```
ExecutedCollectInvest
  -  Shares instead of stable coins
  -  How much Currency (stable coins) was actually invested
  -  optional currencyID
  -  How many shares were received for the invested currency
  -  How much currency left to invest (could not be included in the epoch execution)
```

pallets/connectors/src/message.rs
```diff
@@ -201,6 +201,90 @@ where
...
+		/// The remaining amount of `currency` the investor still has locked to
+		/// invest at a later epoch execution
+		remaining_locked_investment: Balance,
+	},
+	/// The message sent back to the domain from which a `CollectRedeem` message
+	/// has been received, which will ensure the `investor` gets the payout
+	/// respective to their redemption.
+	///
+	/// Directionality: Centrifuge -> EVM Domain.
+	ExecutedCollectRedeem {
```
NunoAlexandre, line 273: Based on

```
ExecutedCollectRedeem
 -  How many shares (TIN/DROP) were actually redeemed
 -  How much currency (stable coins) was received for the redeemed shares
 -  optional: currencyID
 -  How many shares left to be redeemed (could not be included in the epoch execution)
```

mustermeiszer (commented):
Looks good. Just want to get a better understanding of one field that I am not getting.

pallets/connectors/src/message.rs
```diff
@@ -201,6 +201,90 @@ where
...
+		/// The pool id
+		pool_id: PoolId,
+		/// The tranche
+		tranche_id: TrancheId,
+		/// The investor's address
+		investor: Address,
+		/// The currency in which the investment was realised
+		currency: u128,
+		/// The amount that was actually invested
+		amount_invested: Balance,
```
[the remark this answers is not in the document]
NunoAlexandre: ah, good catch. I _think_ that this was a typo in the original text-based spec pasted above and the goal was to say "how much currency (stable coins) was actually **collected**? Which would mean we would rename this field to `amount_collected`. cc @ilinzweilin 👀

pallets/connectors/src/message.rs
```diff
@@ -201,6 +201,90 @@ where
...
+		/// The amount of tranche tokens received for the investment made
+		tranches_received: Balance,
```
offerijns, line 263: The investor is receiving tranche tokens, not tranches.

```suggestion
		tranche_tokens_received: Balance,
```

pallets/connectors/src/message.rs
offerijns: I actually don't understand still why we need `ExecutedDecreaseInvestOrder` and `ExecutedDecreaseRedeemOrder` but not the `Increase` equivalents 🤔 @ilinzweilin @mustermeiszer can you explain ?

pallets/connectors/src/message.rs
offerijns: I think 
 -  How many shares (TIN/DROP) were actually redeemed
 
is missing in this message.

pallets/connectors/src/message.rs
```diff
@@ -201,6 +201,90 @@ where
...
+		pool_id: PoolId,
+		/// The tranche id
+		tranche_id: TrancheId,
+		/// The investor's address
+		investor: Address,
+		/// The stable coin currency in which the payout takes place
+		currency: u128,
+		/// The amount of `currency` being paid out to the investor
+		payout: Balance,
+		/// The remaining amount of tranches the investor still has locked to
```
offerijns, line 284:
```suggestion
		/// The remaining amount of tranche tokens the investor still has locked to
```

pallets/connectors/src/message.rs
wischli: My understanding was that `Decrease*` and `Collect*` calls require a message being sent back to the source chain in order to (re-)fund the investor whereas `Increase*` does not require callback handling.

pallets/connectors/src/message.rs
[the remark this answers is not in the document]
wischli: > Which would mean we would rename this field to `amount_collected`

That makes sense to me.

pallets/connectors/src/message.rs
NunoAlexandre: I sort of remember @ilinzweilin saying that that's something we could calculate on the EVM side by having the `payout` amount and the price of the tranche token. But I guess that just having that info in the message saves computation, fees, and margin of error on the EVM side.

pallets/connectors/src/message.rs
offerijns: Makes sense, thanks both for explaining!

offerijns:
I think it's good to review all messages together. The list now is (focusing on the difference in args):
```rust
	ExecutedDecreaseInvestOrder {
		...,
		decreased_by: Balance,
		outstanding_order: Balance,
	},

	ExecutedDecreaseRedeemOrder {
		...,
		decreased_by: Balance,
		outstanding_order: Balance,
	},

	ExecutedCollectInvest {
		...,
		amount_collected: Balance,
		tranches_received: Balance,
		remaining_locked_investment: Balance,
	}

	ExecutedCollectRedeem {
		...,
		payout: Balance,
		redeemed_shares: Balance,
		remaining_locked_redemption: Balance,
	}
```

I think:
- Remaining and outstanding are being mixed, even though they are the same.
- Shares and tranche tokens are being mixed. I don't think we should use the shares terminology here, that's 4626 specific and not something we use on Centrifuge Chain.
- `amount_collected` does not make clear whether it's denominated in currency or tranche tokens

Just aligning everything to be consistent, we could do something like:

```rust
	ExecutedDecreaseInvestOrder {
		...
		payout_currency: Balance,
		remaining_invest_order: Balance,
	},

	ExecutedDecreaseRedeemOrder {
		...
		payout_tranche_tokens: Balance,
		remaining_redeem_order: Balance,
	},

	ExecutedCollectInvest {
		payout_currency: Balance,
		payout_tranche_tokens: Balance,
		remaining_invest_order: Balance,
	}

	ExecutedCollectRedeem {
		payout_currency: Balance,
		payout_tranche_tokens: Balance,
		remaining_redeem_order: Balance,
	}
```

I don't know whether `payout_currency` / `payout_tranche_tokens` are ideal, maybe someone has a better proposal for this. But I think it's important we make clear what is denominated in currency and what in tranche tokens.

NunoAlexandre:
I will sound fishy but I left some naming inconsistencies on purpose to see what spontaneous discussions would spark from there; this motivated by the mixed language when discussing and proposing this messages in different channels. I like your suggestions!

> I don't know whether payout_currency / payout_tranche_tokens are ideal, 

I find them fine. Even better, imho, would be to make them type-safe, so `payout: TrancheTokensBalance` and `payout: CurrencyBalance`, which would make the code itself unambiguous. But going with just the naming specificity is also fine for me.

Will wait for more thoughts from the other peers 👀

NunoAlexandre:
> Definitely in favour of distinguishing between tranche and payout balance types! But that does not solve the naming question or can the type-safety also be applied to Solidity?

The naming question is only an issue because we want to make sure we don't confuse tranche tokens with a stable coin currency; type-safety solves that at a deeper level than naming can. On Solidity I think we could have a similar approach, but the Rust side shouldn't be deprived of better solutions whenever the Solidity side can't match them, imo. Otherwise we would be stuck coding Solidity in Rust.

But so, again, I am cool with just go ahead with the naming improvement now and go from there. Your suggestion to have `tranche_payout` and `currency_payout ` works for me, even tho we should name it `tranche_tokens_payout`.

offerijns:
> ```diff
> - payout_currency
> + currency_payout
> - payout_tranche_tokens
> + tranche_tokens_payout
> ```

I agree with this ☝️ 

I did just realize, it might be a bit confusing that this is implying there's 2 payouts, of both currency and tranche tokens. But I suppose there's no easy fix for that.

offerijns:
I'm wondering how this would work with different currencies.

Suppose you invest in a pool denominated in Statemint USDC, with 6 decimals. But the currency you use to invest from Ethereum is DAI, with 18 decimals.

- You trigger an `IncreaseInvestOrder(..)` of 100 DAI (so specified as `100 * 10**18`). On Centrifuge Chain, `foreign-investments` will submit a swap order from DAI to USDC, which will implicitly change the decimals (as the swap will basically remove the 12 additional decimals). So this will result in something like `99 * 10**6` USDC being invested (excl swap fee).
- This gets locked as an investment order in a pool and executed in the next epoch.
- Some bot triggers the collect and this should send a message back. What would the `ExecutedCollectInvest` now look like? I think for proper bookkeeping in the 4626 implementation, it should be
  - `currency` = index of Ethereum DAI
  - `currency_payout` = `99 * 10**18` as the amount of currency actually invested in the pool, but converted to the currency that the user invested with
  - `tranche_token_payout` = assuming 1.0 token price, `99 * 10**6` (as tranche tokens are always denominated in the decimals of the pool currency)
  - `remaining_invest_order` = 0
  
@ilinzweilin does this sound right to you? Especially the `currency_payout` value.

But @NunoAlexandre @wischli @mustermeiszer  getting to that `currency_payout` value on the cent chain side will probably be complex 🤔

NunoAlexandre:
@offerijns @wischli @mustermeiszer @ilinzweilin appreciate another round 👀 🙏

offerijns:
`currency_payout`/ `payout_currency ` and `payout_tranche_tokens` / `tranche_tokens_payout` are now both used. Is that intentional? I would use one or the other, so always `currency_payout` and `tranche_tokens_payout` as discussed above.

NunoAlexandre:
> payout_tranche_tokens

Sorry, that wasn't intentional. Let me have a look and fix that 👍

mustermeiszer:
> But @NunoAlexandre @wischli @mustermeiszer getting to that currency_payout value on the cent chain side will probably be complex 🤔

I think it won't be too hard. We must anyways know which fulfilled order needs to trigger such an `Execute*` message. When must hence, just additionally store the "incoming" currency. 

But given we have this know, `pallet-foreign-investments` will need to take care of this earlier than expected. I mean in V2 and not V3.

NunoAlexandre:
@offerijns thanks for the quick review and catch! I believe it's all fixed now 👍

wischli (approved):
LGTM!

offerijns (approved):
Let's get this merged!

pallets/connectors/src/message.rs
NunoAlexandre: ℹ️  This was renamed to `tranche_tokens_payout`

PR #1364, opened 2023-05-25
feat: add connectors v2 extrinsics & inbound msg handlers

wischli:
# Description

* Extends Connectors v2 messages added in #1292 based on the [technical spec](https://centrifuge.hackmd.io/SERpps-URlG4hkOyyS94-w?view)
   * Adds extrinsics for outbound msgs
   * Adds inbound msg handlers

Please note that the spec is not finalized. I marked the corresponding sections in the code with `TODO(@review)` and added a couple of other ones which were not crystal clear to me. Of course, this does not mean, that you reviewers should solve this issue here. It should merely raise awareness as this PR is quite large such that there's a lot of noise. In order to simplify the reviewing process, I have left a couple of notes.

As discussed, we want to solve issues such as handling `Executed*` messages in subsequent PRs. Thus, I have divided open tasks related to this PR into three sections: Now, either now or later and later.

## Changes and Descriptions

* Adds `InvestmentCollector` trait required for `{Collect | Redeem}For` inbound messages
  * Could also be part of `Investment`
* Adds `ForeignInvestment` trait required to verify a given currency...
  * Is the matching currency of the given pool
  * Has `pool_currency` metadata enabled
* Adds `CurrencyInspect` trait to enable checking whether a currency is a tranche token
* Adds thorough integration (rather unit) tests for outgoing and incoming connectors messages
* Fixes bug in `do_collect_redeem`, see commit [6158fca](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/6158fcacaacd49fb197b2fc0fd0e178807f61919) and this [DAO Slack message](https://centrifugedao.slack.com/archives/C04HCK98KLL/p1686745223438149) for details 
* Temporarily: Disables CI for draft PRs  

### TODOs in this PR

- [x] Write tests (preferably integration, otherwise using the `mock-builder`)
- [x] Rebase to `main` after #1332 is merged
- [x] Solve open questions marked by `TODO(@review)`

### TODOs in this PR or later

- [x] Add trait `PoolMetadata` that allows retrieving the necessary data and the registry implements it ([details](https://centrifuge.hackmd.io/SERpps-URlG4hkOyyS94-w?both#pallet-connectors)) --> https://github.com/centrifuge/centrifuge-chain/pull/1432
- [x] Fix outbound domain derivation from `CurrencyId` after #1357
- [x] Derive sending domain via Gateway `InboundQueue` for inbound tranche token transfer and redemptions (blocked by https://github.com/centrifuge/centrifuge-chain/pull/1376)

### TODO in other PRs

- [ ] Changing `PoolInspect` trait based [on this proposal](https://centrifuge.hackmd.io/SERpps-URlG4hkOyyS94-w?both#fn-update_tranche_token_price)
- [ ] `valid_until` [rework](https://centrifuge.hackmd.io/SERpps-URlG4hkOyyS94-w?view#fn-update_member)
- [ ] Handling and tests for Executed*: Write `pallet-foreign-investments` which should handle the storage and dispatch of `Executed*` connectors messages and `pallet-connectors` relays the storage relevant data to `pallet-foreign-investments` when it executes inbound messages `Decrease{Invest, Redeem}Order` and `Collect{Invest, Redeem}`
- [x] Remove hard-coupling with `pallet-xcm-transactor` and remove/rework `handle`
- [ ] E2E Tests

# Checklist:


pallets/connectors/src/lib.rs
```diff
@@ -486,11 +506,137 @@ pub mod pallet {
...
+			pool_id: PoolIdOf<T>,
+			currency_id: CurrencyIdOf<T>,
+		) -> DispatchResult {
+			// TODO(subsequent PR): In the future, should be permissioned by trait which
+			// does not exist yet See spec: https://centrifuge.hackmd.io/SERpps-URlG4hkOyyS94-w?view#fn-update_member
+
+			// TODO(@review): According to spec, this should be restricted to
+			// `AdminOrigin`. However, `do_send_message` requires a 32-byte address for the
+			// payment of the fee. We could set the treasury.
+			// https://centrifuge.hackmd.io/SERpps-URlG4hkOyyS94-w?view#fn-add_pool_currency
```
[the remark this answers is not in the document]
offerijns: I think for now, I would just make this callable by anyone and only allow the equivalent currency on the target domain of the local pool currency. So if the pool currency is Statemint USDC, we would allow USDC on all other domains by default.

In the future we can figure out what would be the right flow to support multiple currencies, but I don't think we need to solve that now.

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+		let currency_id = Self::try_get_currency_id(currency)?;
+		T::Tokens::mint_into(currency_id, &receiver, amount)?;
+
+		Ok(())
+	}
+
+	/// Executes a transfer from the `DomainLocator` account of the origination
+	/// domain to the receiver exclusively for tranche tokens.
+	///
+	/// Assumes that the amount of tranche tokens has been minted in the
```
offerijns, line 56: Tranche tokens are always minted by investments, only transferred into the `DomainLocator` account on `connectors.Transfer`
```suggestion
	/// Assumes that the amount of tranche tokens has been locked in the
```

pallets/connectors/src/message.rs
```diff
@@ -383,26 +382,26 @@ impl<
...
 					encode_be(pool_id),
 					tranche_id.encode(),
-					address.to_vec(),
+					member.to_vec(),
 					valid_until.to_be_bytes().to_vec(),
 				],
 			),
 			Message::Transfer {
-				currency: token,
+				currency,
```
NunoAlexandre, line 397: nice 👌

pallets/connectors/src/lib.rs
```diff
@@ -224,36 +226,54 @@ pub mod pallet {
 	#[pallet::error]
 	pub enum Error<T> {
 		/// Failed to map the asset to the corresponding Connector's General
-		/// Index representation
+		/// Index representation and thus cannot be used as an
+		/// ivnestment currency.
```
NunoAlexandre, line 230:
```suggestion
		/// investment currency.
```

pallets/connectors/src/lib.rs
```diff
@@ -224,36 +226,54 @@ pub mod pallet {
...
-		/// A pool could not be found
+		/// The metadata of the given asset does not declare it as a pool
+		/// currency and thus it cannot be used as an investment currency.
+		AssetMetadataNotPoolCurrency,
+		/// The given asset does not match the currency of the pool.
+		AssetNotPoolCurrency,
+		/// A pool could not be found.
 		PoolNotFound,
-		/// A tranche could not be found
+		/// A tranche could not be found.
```
NunoAlexandre, line 239: I have the idea the convention with rust docs is to have a `.` at the end of a one liner

pallets/connectors/src/lib.rs
```diff
@@ -224,36 +226,54 @@ pub mod pallet {
...
 		InvalidIncomingMessageOrigin,
-		/// Failed to decode an incoming message
+		/// Failed to decode an incoming message.
 		InvalidIncomingMessage,
-		/// A transfer attempt from the local to the local domain
-		InvalidTransferDomain,
+		/// The destination domain is invalid.
+		InvalidDomain,
+		/// The validity is in the past.
+		InvalidValidity,
```
NunoAlexandre, line 262: Maybe we should make this error name a bit more specific

pallets/connectors/src/lib.rs
```diff
@@ -378,48 +400,44 @@ pub mod pallet {
...
-			domain_address: DomainAddress,
 			pool_id: PoolIdOf<T>,
 			tranche_id: TrancheIdOf<T>,
+			domain: DomainAddress,
 			valid_until: Moment,
 		) -> DispatchResult {
 			let who = ensure_signed(origin.clone())?;
 
-			// Check that the origin is allowed to add other members
+			// Sanity checks
```
NunoAlexandre, line 410:
```suggestion
```

pallets/connectors/src/lib.rs
```diff
@@ -378,48 +400,44 @@ pub mod pallet {
 		#[pallet::call_index(5)]
 		pub fn update_member(
 			origin: OriginFor<T>,
-			domain_address: DomainAddress,
 			pool_id: PoolIdOf<T>,
 			tranche_id: TrancheIdOf<T>,
+			domain: DomainAddress,
```
NunoAlexandre, line 405: I don't really agree with this renaming because it's not a `Domain`, it's a `DomainAddress`. At worse, we could name it `address`, not `domain`.

pallets/connectors/src/lib.rs
```diff
@@ -486,11 +508,139 @@ pub mod pallet {
...
+			ensure!(
+				!CurrencyIdOf::<T>::is_tranche_token(currency_id),
+				Error::<T>::InvalidTransferCurrency
+			);
+			let currency = Self::try_get_general_index(currency_id)?;
+
+			// FIXME: Ensure receiving domain is issuer of token, e.g. we never transfer
+			// tokens from OtherDomain A to OtherDomain B
+			// NOTE: Blocked by asset metadata update
+			// https://github.com/centrifuge/centrifuge-chain/pull/1393
```
NunoAlexandre, line 537: This has been merged ✅

NunoAlexandre (approved):
A first and superficial round to scan it through. Looking very good already ✨

pallets/connectors/src/lib.rs
wischli: Do you think `InvalidtTrancheInvestorValidity` suffices?

NunoAlexandre: Yeah I like that :) Or `InvalidUpateMemberValidity` but I think we are renaming  `update_member` to `update_investor` eventually I guess?

libs/types/src/tokens.rs
```diff
@@ -264,7 +289,7 @@ impl CrossChainTransferability {
 	Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen,
 )]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub enum ConnectorsWrappedToken {
+pub enum ConnectorsWrappedCurrency {
```
NunoAlexandre, line 292: Why the rename? I get it that we use token/currency/asset interchangeably but this rename is out of scope and in reality we will be talking about these (connector-wrapped tokens) as tokens and not as currencies so I would prefer if we revert this

pallets/connectors/src/contract.rs
```diff
@@ -13,7 +13,7 @@
 use ethabi::{Bytes, Contract};
 use sp_std::{vec, vec::Vec};
 
-/// The ConnectorsXcmContract handle function name
+/// The ConnectorsXcmContract handle function name.
```
NunoAlexandre, line 16: `nnnit`

```suggestion
/// The ConnectorsXcmContract handle function name
```

in my previous comment I meant to say that the convention I was aware with rust one line docs is to **not** have a dot at the end. I looks kinda weird in the generated docs .

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+use crate::{
+	pallet::Error, Config, CurrencyIdOf, GeneralCurrencyIndexOf, Pallet, PoolIdOf, TrancheIdOf,
+};
+
+impl<T: Config> Pallet<T> {
+	/// Executes a transfer from another domain exclusively for
+	/// non-tranche-tokens.
+	///
+	/// Directly mints the currency into the receiver address.
+	pub fn do_transfer_from_other_domain(
```
NunoAlexandre, line 40:
```suggestion
	pub fn do_transfer(
```

or 

```suggestion
	pub fn do_transfer(
```


Given that this module is called `inbound`, the context of these functions being about messages coming from other domains is a given, so I would drop the redundant and verbose `_from_other_domain(`. Using `handle` instead of `do` also suggests that it's handling something that's coming in, but maybe that's just how I interpret that.

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+
+		Ok(())
+	}
+
+	/// Executes a transfer from the `DomainLocator` account of the origination
+	/// domain to the receiver exclusively for tranche tokens.
+	///
+	/// Assumes that the amount of tranche tokens has been locked in the
+	/// `DomainLocator` account of the origination domain beforehand.
+	pub fn do_transfer_tranche_tokens_from_other_domain(
```
NunoAlexandre, line 58: Considering the comment above, I'd rename this to:

```suggestion
	pub fn handle_tranche_tokens_transfer(
```

pallets/connectors/src/lib.rs
```diff
@@ -100,19 +72,27 @@ pub type MessageOf<T> =
...
-		tokens::{CustomMetadata, GeneralCurrencyIndex},
+		tokens::{ConnectorsWrappedCurrency, CustomMetadata},
 	};
-	use frame_support::{error::BadOrigin, pallet_prelude::*, traits::UnixTime};
+	use frame_support::{pallet_prelude::*, traits::UnixTime};
 	use frame_system::pallet_prelude::*;
 	use pallet_xcm_transactor::{Currency, CurrencyPayment, TransactWeights};
 	use sp_runtime::traits::{AccountIdConversion, Zero};
-	use xcm::latest::OriginKind;
+	use xcm::{latest::OriginKind, v3::MultiLocation};
```
NunoAlexandre, line 95: I think you can just import `MultiLocation` from `latest`?
```suggestion
	use xcm::{latest::{OriginKind, MultiLocation};
```

pallets/connectors/src/lib.rs
```diff
@@ -160,7 +139,27 @@ pub mod pallet {
 			> + Transfer<Self::AccountId>;
 
 		type TrancheCurrency: TrancheCurrency<PoolIdOf<Self>, TrancheIdOf<Self>>
-			+ Into<CurrencyIdOf<Self>>;
+			+ Into<CurrencyIdOf<Self>>
+			+ Clone;
+
+		type ForeignInvestment: Investment<
```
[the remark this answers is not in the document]
NunoAlexandre: Could you leave that as a comment in the code? :)

pallets/connectors/src/lib.rs
```diff
@@ -177,14 +177,22 @@ pub mod pallet {
...
+				<<<Self as Config>::ForeignInvestmentAccountant as InvestmentAccountant<
+					Self::AccountId,
+				>>::InvestmentInfo as InvestmentProperties<Self::AccountId>>::Currency,
+			> + CurrencyInspect<CurrencyId = <Self as pallet::Config>::CurrencyId>;
 
 		/// The converter from a DomainAddress to a Substrate AccountId
 		type AccountConverter: Convert<DomainAddress, Self::AccountId>;
 
+		/// The converter from a [ConnectorsWrappedCurrency] to `MultiLocation`.
+		type CurrencyConverter: Convert<ConnectorsWrappedCurrency, MultiLocation>
```
NunoAlexandre, line 192: Not sure about this name since this is not converting currencies. No suggestions now but something to think about

pallets/connectors/src/lib.rs
```diff
@@ -378,38 +412,36 @@ pub mod pallet {
...
+			ensure!(
+				T::Permission::has(
 					PermissionScope::Pool(pool_id),
 					T::AccountConverter::convert(domain_address.clone()),
-					Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, valid_until)),
-				)?;
-			}
+					Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, valid_until))
+				),
+				Error::<T>::DomainNotWhitelisted
```
NunoAlexandre, line 443: So now this is only testing that the destination domain is whitelisted, no? But we have to ensure that the specific address to which tranche tokens are being sent is an investor and not the destination domain as a whole.

NunoAlexandre: Unless if `T::AccountConverter::convert(domain_address.clone()),` considers both the domain and the address; if it doesn't this is wrong and the error name as well 👀

pallets/connectors/src/lib.rs
```diff
@@ -486,11 +506,137 @@ pub mod pallet {
...
+		}
+
+		/// Add a currency to the set of known currencies on the domain derived
+		/// from the given currency.
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1).ref_time())]
+		#[pallet::call_index(8)]
+		pub fn add_currency(origin: OriginFor<T>, currency_id: CurrencyIdOf<T>) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			let currency = Self::try_get_general_index(currency_id)?;
```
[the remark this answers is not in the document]
NunoAlexandre: I think we are covered with `try_get_wrapped_currency` now 👍

pallets/connectors/src/lib.rs
```diff
@@ -612,11 +865,108 @@ pub mod pallet {
...
-			let general_index: GeneralCurrencyIndex<u128, T::GeneralCurrencyPrefix> =
-				CurrencyIdOf::<T>::try_into(currency)?;
+			let general_index: GeneralCurrencyIndexOf<T> = CurrencyIdOf::<T>::try_into(currency)?;
 
 			Ok(general_index.index)
 		}
+
+		/// Returns the local currency identifier from from its general index.
+		///
+		/// Assumes the currency to be registered in the `AssetRegistry`.
```
NunoAlexandre, line 875: How do you mean, "assumes"? Because it handles the error case when it is not in the registry so I wouldn't say it assumes anything

pallets/connectors/src/lib.rs
```diff
@@ -612,11 +865,108 @@ pub mod pallet {
...
+			let metadata =
+				T::AssetRegistry::metadata(&currency).ok_or(Error::<T>::AssetNotFound)?;
+			ensure!(
+				metadata.additional.pool_currency,
+				Error::<T>::AssetMetadataNotPoolCurrency
+			);
+
+			T::PoolInspect::currency_for(pool_id)
+				.filter(|c| c == &currency)
+				.map(|_| ())
```
NunoAlexandre, line 967: Do we need this `.map(|_| ())`?

NunoAlexandre (commented):
The code looks really good and solid 🏄‍♂️  I will review the e2e tests in a follow up review

libs/traits/src/lib.rs
```diff
@@ -629,16 +669,24 @@ pub mod fees {
...
 		currency: Self::CurrencyId,
 	) -> DispatchResult;
 }
+
+/// Trait to retrieve information about currencies.
+pub trait CurrencyInspect {
+	type CurrencyId;
+
+	/// Checks whether the provided currency is a tranche token.
+	fn is_tranche_token(currency: Self::CurrencyId) -> bool;
```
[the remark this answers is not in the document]
wischli: I am not sure it would be cleaner to move this to `TrancheCurrency` as with the current implementation this is per definition a tranche token. However, we want to check whether an arbitrary currency is in fact a tranche token. Adding such a static check to the trait seems misplaced to me.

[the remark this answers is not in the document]
NunoAlexandre: That's not what I meant, sorry. I meant could just do `TrancheCurrency::generate(_, _) == currency`, which would ensure the input currency is a tranche token currency. But it is clear and clean as is so fine for me to keep it that way

pallets/connectors/src/contract.rs
wischli: I see, looking at the edit you said the exact opposite though. I have no strong opinion and will revert the changes then.

libs/types/src/tokens.rs
NunoAlexandre: hm, how is `ConnectorsWrappedToken` coupled with `CurrencyId`? 🤔

pallets/connectors/src/contract.rs
NunoAlexandre: > I see, looking at the edit you said the exact opposite though. I have no strong opinion and will revert the changes then.


Not sure if I misunderstand you but I edited the comment to correct it so that it say the right thing, since I noticed I had missed the **not** which said the opposite of what I meant. Mea culpa!

pallets/connectors/src/lib.rs
wischli: How about `Wrapped{Currency/Token}Converter`? We have named `AccountConverter` based on the desired output type (i.e. a Substrate Account). Here, we use the converter for retrieving a `ConnectorsWrapped{Currency/Token}` from a `MultiLocation`.

NunoAlexandre: I like that :)

pallets/connectors/src/lib.rs
NunoAlexandre: Got it 👍🏻

pallets/connectors/src/contract.rs
wischli: I was phrasing poorly - sorry. totally got what you were saying. Looking at Substrate Doc comments, there are actually periods for one liners. I also think it is much more consistent than leaving the existence of periods up to the number of lines. Can we just leave it as is?

libs/types/src/tokens.rs
wischli: Ok you got me: If at all, then implicitly: Right now we rely on the `OrmlAssetRegistry` to provide the `MultiLocation` which can be converted to and from `ConnectorsWrappedToken`. However, all key type of storage entries of this registry is `CurrencyId`.

pallets/connectors/src/contract.rs
NunoAlexandre: Sure! :) This is absolutely not important, sorry for all the noise

pallets/connectors/src/lib.rs
wischli: Good catch. We can either entirely remove this comment or rephrase it to `requires`.

pallets/connectors/src/lib.rs
wischli: Yeah I think it's cleaner than doing `filter_map` in this case. Of course, we could also do a simple `if else`.

pallets/connectors/src/lib.rs
wischli: > Unless if `T::AccountConverter::convert(domain_address.clone()),` considers both the domain and the address; if it doesn't this is wrong and the error name as well 👀

This [seems to be the case](https://github.com/centrifuge/centrifuge-chain/blob/0d60c08d7326499582b457733bc04f7ceb790868/runtime/common/src/account_conversion.rs#L58): It basically converts to `[...evm_address.bytes(), ...chain_id.bytes(), ...b"EVM]`.

Then it should be better to phrase the error `DomainLocatorNotWhitelisted` or do you have a better suggestion?

libs/traits/src/lib.rs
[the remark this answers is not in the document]
wischli: Looking over it again, I feel like we can entirely remove this trait. Since `CurrencyId` implements `TrancheCurrency` and we are generating a `TrancheCurrency` before checking `is_tranche_token` anyway, it feels completely redundant.

runtime/integration-tests/src/xcm/development/tests/connectors.rs
```diff
@@ -1088,8 +1089,81 @@ fn allow_pool_should_fail() {
...
+		// Should fail if currency does not have any MultiLocation in metadata
+		assert_ok!(OrmlAssetRegistry::update_asset(
+			RuntimeOrigin::root(),
+			pool_currency_id,
+			None,
+			None,
+			None,
+			None,
+			None,
+			Some(CustomMetadata {
```
NunoAlexandre, line 1126: Am I not seeing this well or is this `update_asset` exactly the same as the one above?

NunoAlexandre (commented):
We are very close :) Although, it would be nice to get more eyes on this before merging since it's a big PR with a lot to look through

pallets/connectors/src/lib.rs
NunoAlexandre: The _thing_ being whitelisted is an investor's address; it's just that such address is on another chain. So it's not the `DomainLocatorNotWhitelisted` that's not whitelisted but the cross-chain Investor's address. Maybe `InvestorEvmAddressNotWhitelisted`? Or just `InvestorNotWhitelisted`?

libs/types/src/tokens.rs
NunoAlexandre: I wouldn't extrapolate and rename it for that reason. The truth is that we have been and will refer to this type and to these tokens as connectors-wrapped tokens, so Id' rather have the code reflecting the language we have and will adopt to avoid unnecessary divergence.

runtime/integration-tests/src/xcm/development/tests/connectors.rs
wischli: The update above that has `CrossChainTransferability::None` whereas this one has `CrossChainTransferability::Connectors`. I improved awareness of the changes in [5396f5f](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/5396f5fbd3a1995a6748c17f6cae3603473c154f) as this very difficult to follow due to the amount of entries required for an update.

libs/types/src/tokens.rs
wischli: Thanks for taking the time to discuss this with me. Renamed it back to `ConnectorsWrappedToken` in [4589410](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/45894105b130e5547e3a219b1e7480a8c0ed0fee) :)

pallets/connectors/src/lib.rs
wischli: Went with `InvestorEvmAddressNotWhitelisted` as this is more explicit. Thanks for the explanation!

pallets/connectors/src/lib.rs
```diff
@@ -378,38 +442,36 @@ pub mod pallet {
...
+			// Ensure that the destination address has been whitelisted as a TrancheInvestor
+			// beforehand.
+			ensure!(
+				T::Permission::has(
 					PermissionScope::Pool(pool_id),
 					T::AccountConverter::convert(domain_address.clone()),
-					Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, valid_until)),
-				)?;
-			}
+					Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, valid_until))
```
mustermeiszer, line 471: Did we alredy adapt the behaviour of the permissions checking to take into account the `valid_until`? ^^

pallets/connectors/src/lib.rs
```diff
@@ -378,38 +442,36 @@ pub mod pallet {
...
+				),
+				Error::<T>::InvestorEvmAddressNotWhitelisted
```
mustermeiszer, line 473: I would opt in for a more generic error. Like `InvestorDomainAddressNotWhitelisted`

pallets/connectors/src/lib.rs
```diff
@@ -453,11 +514,16 @@ pub mod pallet {
 				Error::<T>::UnauthorizedTransfer
 			);
 
-			ensure!(!amount.is_zero(), Error::<T>::InvalidTransferAmount);
+			// Ensure pool and tranche exist and derive invest id
+			let invest_id = Self::derive_invest_id(pool_id, tranche_id)?;
+			ensure!(
+				CurrencyIdOf::<T>::is_tranche_token(invest_id.clone().into()),
+				Error::<T>::InvalidTransferCurrency
+			);
```
mustermeiszer, line 522: The `TrancheCurrency::generate` will always generate the right currency. So this check is not needed IMO.

pallets/connectors/src/lib.rs
```diff
@@ -123,10 +103,13 @@ pub mod pallet {
 
 	#[pallet::config]
 	pub trait Config: frame_system::Config + pallet_xcm_transactor::Config {
+		/// The overarching event type.
 		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
 
+		/// Weight information for extrinsics in this pallet.
 		type WeightInfo: WeightInfo;
 
+		/// The source of truth for the balance of accounts in native currency.
```
mustermeiszer, line 112: We do not need the balance type I think. We could just refer always to the `BalanceOf<T> = <<T as Config>::Tokens as Inspect>::Balance`

pallets/connectors/src/lib.rs
```diff
@@ -486,11 +552,139 @@ pub mod pallet {
...
+			ensure!(!amount.is_zero(), Error::<T>::InvalidTransferAmount);
+			// Check that the destination is not the local domain
+			ensure!(
+				receiver.domain() != Domain::Centrifuge,
+				Error::<T>::InvalidDomain
+			);
+			ensure!(
+				!CurrencyIdOf::<T>::is_tranche_token(currency_id),
+				Error::<T>::InvalidTransferCurrency
+			);
```
mustermeiszer, line 575: That is actually a good reason for the trait. ^^ 

I think it makes sense to keep it and implement it for our currency id.

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: See comment above.

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+	/// Executes a transfer from another domain exclusively for
+	/// non-tranche-tokens.
+	///
+	/// Directly mints the currency into the receiver address.
+	pub fn do_transfer(
+		currency: GeneralCurrencyIndexOf<T>,
+		receiver: T::AccountId,
+		amount: <T as Config>::Balance,
+	) -> DispatchResult {
+		ensure!(!amount.is_zero(), Error::<T>::InvalidTransferAmount);
```
mustermeiszer, line 45: @NunoAlexandre do we have the same check on the Solidity side?

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+
+		// Check that the source is not the local domain
+		ensure!(
+			sending_domain.domain() != Domain::Centrifuge,
+			Error::<T>::InvalidDomain
+		);
```
mustermeiszer, line 71: I think this is not needed. As mentioned earlier we should be consistent.

mustermeiszer: I include myself here definitely ^^

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+				Role::PoolRole(PoolRole::TrancheInvestor(tranche_id, Self::now())),
+			),
+			Error::<T>::UnauthorizedTransfer
+		);
+
+		let invest_id = Self::derive_invest_id(pool_id, tranche_id)?;
+		ensure!(
+			CurrencyIdOf::<T>::is_tranche_token(invest_id.clone().into()),
+			Error::<T>::InvalidTransferCurrency
+		);
```
mustermeiszer, line 85: redundant as we generate the tranche currency in `Self::derive_invest_id`

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+
+		T::Tokens::transfer(
+			invest_id.into(),
+			&T::AccountConverter::convert(sending_domain),
```
mustermeiszer, line 89: I think this is not right. This should be the `DomainLocator` instead, right?

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+		let invest_id: T::TrancheCurrency = Self::derive_invest_id(pool_id, tranche_id)?;
+
+		// Determine post adjustment amount
+		let pre_amount = T::ForeignInvestment::redemption(&investor, invest_id.clone())?;
+		let post_amount = pre_amount.ensure_add(amount)?;
+
+		// Transfer tranche tokens from `DomainLocator` account of origination domain
+		T::Tokens::transfer(
+			invest_id.clone().into(),
+			&T::AccountConverter::convert(sending_domain),
```
mustermeiszer, line 187: Again, I think this should be `DomainLocator`, right?

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+		let post_amount = pre_amount.ensure_sub(amount)?;
+
+		T::ForeignInvestment::update_redemption(&investor, invest_id.clone(), post_amount)?;
+
+		// Transfer tranche tokens to `DomainLocator` account of
+		// origination domain
+		T::Tokens::transfer(
+			invest_id.into(),
+			&investor,
+			&T::AccountConverter::convert(sending_domain),
```
mustermeiszer, line 232: also `DomainLocator`

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+			amount,
+			false,
+		)?;
```
mustermeiszer, line 235: I think this should also be handled by the `pallet-foreign-invesments`. WDYT?

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+	/// next active order for this investment.
+	pub fn do_collect_investment(
+		pool_id: PoolIdOf<T>,
+		tranche_id: TrancheIdOf<T>,
+		investor: T::AccountId,
+	) -> DispatchResult {
+		let invest_id: T::TrancheCurrency = Self::derive_invest_id(pool_id, tranche_id)?;
+
+		T::ForeignInvestment::collect_investment(investor, invest_id)
+	}
```
mustermeiszer, line 256: TODO: Needs to trigger an `ExecutedCollectInvest` message.

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,270 @@
...
+	/// next active order for this investment.
+	pub fn do_collect_redemption(
+		pool_id: PoolIdOf<T>,
+		tranche_id: TrancheIdOf<T>,
+		investor: T::AccountId,
+	) -> DispatchResult {
+		let invest_id: T::TrancheCurrency = Self::derive_invest_id(pool_id, tranche_id)?;
+
+		T::ForeignInvestment::collect_redemption(investor, invest_id)
+	}
```
mustermeiszer, line 269: TODO: Needs to also trigger an `ExecutedCollectRedeem` message.

mustermeiszer (changes requested):
That looks really good and nice already! Nice to see the progress here!

Some minor but important change requests:
* `fn add_tranche` must derive decimals of tranche from regitry meta not user input
* impl for `InboundQueue` is missing
* use `OutboundQueue` instead of `fn do_send_message`
&rarr; Basically, I think we should assume that the `OutboundQueue` is provided by the system already
   * remove tight coupling to `pallet-xcm-transactor`


Some minor proposals:
* adapt `trait Investments` (see comments)
* remove `ConnectorsWrappedCurrency` (see comments)

pallets/connectors/src/lib.rs
```diff
@@ -486,11 +552,139 @@ pub mod pallet {
...
+			amount: <T as pallet::Config>::Balance,
+		) -> DispatchResult {
+			let who = ensure_signed(origin.clone())?;
+
+			ensure!(!amount.is_zero(), Error::<T>::InvalidTransferAmount);
+			// Check that the destination is not the local domain
+			ensure!(
+				receiver.domain() != Domain::Centrifuge,
+				Error::<T>::InvalidDomain
+			);
```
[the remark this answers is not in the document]
wischli: I was also hesitant and support that it should rather be handled via the Gateway.

pallets/connectors/src/lib.rs
offerijns: Let's not use the term whitelisting in code, but rather something like `InvestorDomainAddressNotAMember`

pallets/connectors/src/lib.rs
mustermeiszer: Lazy me, did not read carefully enough. Sorry ^^

pallets/connectors/src/inbound.rs
mustermeiszer: Sorry, in did not wanted to create more confusion. Given connectors pallet is already huge, I would suggest we create an additional new one

* `pallet-connectors-investments`

This pallet impls `trait Investments` BUT only forwards to an internal type that also implements `trait Investments` BUT does some additional bookkeeping for triggering return messages and minting tokens etc. 

Then we would have 

* `pallet-investments`: Doing the actual investing logic
* `pallet-foreign-investments`: Creating swap orders and async using `pallet-investments` for actual investing
* `pallet-connectors-investments`: Taking care of return messages and conenctors related bookeeping using `pallet-foreign-investments` for actual investing

Does this make sense for you all?

mustermeiszer: IF we agree then:

* `pallet-pool-links` uses `pallet-foreign-investments`
* `pallet-connectors` uses `pallet-connectors-investments`

offerijns: And `pallet-connectors-investments` uses `pallet-foreign-investments`, right?

offerijns: Seems to me like "Taking care of return messages and conenctors related bookeeping" still fits within the intended goal of the `pallet-connectors`, and this only needs more code organization within `pallet-connectors` rather than a separate pallet. But ultimately I don't have a very strong opinion on this and would leave this to you @NunoAlexandre @wischli et al to decide :)

mustermeiszer: I honestly also am not opinionated here. Just be aware that connectors will need to expose some "automation" logic to get start working on investments if they can make progress later down the road.

NunoAlexandre: > Taking care of return messages and conenctors related bookeeping" still fits within the intended goal of the pallet-connectors


I agree ☝️

wischli: Do I understand correctly that the pallet hierarchy for an incoming connectors investment message would be (leaving out gateway) `connectors -> connectors-investments -> foreign-investments -> investments`? E.g.

1. `pallet-connectors` forwards to `pallet-connectors-investments` which basically implements `Investment` as the done in `inbound.rs` in this PR (see associated type `ForeignInvestment`)
2. The `Investment` implementation of `pallet-connectors-investments` forwards the actual investment logic to `pallet-foreign-investments`
3. The `Investment` implementation of `pallet-foreign-investments` creates swap orders and once settled, forwards the investment to the `Investment` implementation of `pallet-investment`

mustermeiszer: > I agree ☝️

I am not completely against it, but we should keep in mind that we will have a lot of async stuff to handle in the future if we allow to change orders on the fly. Just am not so confident that this will not blow the connectors codebase too much. 

But we can go with it of course.

pallets/connectors/src/lib.rs
```diff
@@ -486,11 +552,139 @@ pub mod pallet {
...
+			T::Tokens::transfer(
+				currency_id,
+				&who,
+				&DomainLocator::<Domain> {
+					domain: receiver.domain(),
+				}
+				.into_account_truncating(),
+				amount,
+				false,
+			)?;
```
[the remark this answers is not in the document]
wischli: Just to be sure: We do want `AllowDeath` right?

pallets/connectors/src/lib.rs
```diff
@@ -605,18 +888,124 @@ pub mod pallet {
...
+				metadata.additional.pool_currency,
+				Error::<T>::AssetMetadataNotPoolCurrency
+			);
+
+			T::PoolInspect::currency_for(pool_id)
+				.filter(|c| c == &currency)
+				.map(|_| ())
+				.ok_or(Error::<T>::AssetNotPoolCurrency.into())
+		}
 	}
```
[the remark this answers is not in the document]
wischli: We would still need to check in Connectors whether the `metadata.additional.pool_currency` flag is set to `true` for now, right? I cant imagine we want to add the `AssetRegistry` to the `Investment` trait.

libs/traits/src/lib.rs
```diff
@@ -419,6 +421,44 @@ pub trait Investment<AccountId> {
...
+	type Error: Debug;
+	type PoolId;
+
+	/// Checks whether the currency can be used for investing into the given
+	/// pool.
+	fn can_invest_currency_into_pool(
+		invest_id: Self::PoolId,
+		currency: CurrencyId,
+	) -> Result<(), Self::Error>;
+}
```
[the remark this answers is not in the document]
wischli: IIUC, `accepted_payment_currency ` and `accepted_payout_currency` can be implemented in the same way for the time being? They can diverge later, e.g. when `pallet-foreign-investments` exists for handling the currency swaps.

pallets/connectors/src/lib.rs
```diff
@@ -486,11 +552,139 @@ pub mod pallet {
...
+		#[pallet::call_index(90)]
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1).ref_time())]
+		pub fn allow_pool_currency(
+			origin: OriginFor<T>,
+			pool_id: PoolIdOf<T>,
+			currency_id: CurrencyIdOf<T>,
+		) -> DispatchResult {
+			// TODO(subsequent PR): In the future, should be permissioned by trait which
+			// does not exist yet.
+			// See spec: https://centrifuge.hackmd.io/SERpps-URlG4hkOyyS94-w?view#fn-add_pool_currency
```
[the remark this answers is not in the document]
mustermeiszer: The tranche is not really needed as `InvestmentId` will already carry that information.

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Yes, sounds reasonable!

pallets/connectors/src/lib.rs
```diff
@@ -486,11 +552,139 @@ pub mod pallet {
...
+
+			// Check that the registered asset location matches the destination
+			match Self::try_get_wrapped_currency(&currency_id)? {
+				ConnectorsWrappedToken::EVM { chain_id, .. } => {
+					ensure!(
+						Domain::EVM(chain_id) == receiver.domain(),
+						Error::<T>::InvalidDomain
+					);
+				}
+			}
```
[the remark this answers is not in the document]
NunoAlexandre: We are coming full circle :P We came to `ConnectorsWrappedToken` after discussing that `DomainAddress` is NOT a suitable type to represent tokens across different chains. It works for EVM tokens (`EVM(chain_id, [u8; 20])`) but `Centrifuge(AccountId32)` does not suit anymore; `DomainAddress` is design to represent an account across different chains, `ConnectorsWrappedToken` for representing connectors wrapped tokens across the different chains we deploy connectors on (currently only `EVM`).

libs/traits/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Yes, exactly!

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Haha. You are right! 

I would still opt for, having a local method that does the conversion and checks. Especially, as we can get the `Connectors` index in the runtime in the pallet itself, this allows us to not leak the type to the runtime. Which makes it less confusing.

pallets/connectors/src/lib.rs
wischli: That does not work as the `Inspect` trait requires setting the associated `Balance` type: https://github.com/paritytech/substrate/blob/ecf71223e6d34204418616d935faad1c35ea36d5/frame/support/src/traits/tokens/fungibles/regular.rs#L51

So we would end up in an evaluation overflow.

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: For me, as long as we are able to have a type that will be ready to handle connectors-wrapped tokens from other chains in the future, it's fine. I personally don't feel the confusion you guys report on the current setup with `ConnectorsWrappedToken` but you are two (handsome ones, for that matter) and I am one so majority takes the vote!

pallets/connectors/src/inbound.rs
wischli: Can we handle this in a subsequent PR as proposed in the PR description? Some parameters of the the `Executed*` messages are not trivial to derive and thus should be tested properly as well.

mustermeiszer: Of course. Just wanted to have todos in the codebase as there are for the other left ones

pallets/connectors/src/lib.rs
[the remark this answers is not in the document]
wischli: I moved the conversion from the specific `MultiLocation` to `ConnectorsWrappedToken` into `fn try_get_wrapped_token` and removed the other conversions as well as unit tests in [714cfd9](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/714cfd915c6092a925c6849eaf95adae8a75ac15) + [d4e382d](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/d4e382de2ad8241a56b4a352fe5c521f632b4328). I was hesitant about the unit tests but we can add the bidirectional conversion and tests back once we need them. Until then, they are just noise IMO.

pallets/connectors/src/lib.rs
```diff
@@ -319,8 +381,8 @@ pub mod pallet {
 			);
 
 			// Look up the metadata of the tranche token
-			let currency_id = T::TrancheCurrency::generate(pool_id, tranche_id).into();
-			let metadata = T::AssetRegistry::metadata(&currency_id)
+			let currency_id = Self::derive_invest_id(pool_id, tranche_id)?;
```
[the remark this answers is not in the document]
wischli: [44469c9](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/44469c9b937f13d6d8398505dc3a8afb989e8190)

libs/traits/src/lib.rs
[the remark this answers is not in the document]
wischli: [c04a1fe](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/c04a1fe58ca1fba73f282b4e5683c1f747f702b4)

pallets/connectors/src/lib.rs
```diff
@@ -486,11 +552,139 @@ pub mod pallet {
...
+			Ok(())
+		}
+
 		/// Handle an incoming message
 		/// TODO(nuno): we probably need a custom origin type for these messages
 		/// to ensure they have come in through XCM. For now, let's have a POC
 		/// here to test the pipeline Ethereum ---> Moonbeam --->
 		/// Centrifuge::connectors
+		/// TODO(subsequent PR): Should be removed as extrinsic before adding to
+		/// prod runtimes. Should be handled automatically via Gateway.
```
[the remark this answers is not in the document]
wischli: [a04aef7](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/a04aef75c42a9b99bde9a2d6ca11d694a9578619)

pallets/connectors/src/inbound.rs
wischli: [cbba52c](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/cbba52ca61042bbf52ca841a4f8338d0d4e0630f)

pallets/connectors/src/inbound.rs
wischli: [bd7ed80](https://github.com/centrifuge/centrifuge-chain/pull/1364/commits/bd7ed80f7c660b12db28f171943082f32af38657)

wischli:
@NunoAlexandre @mustermeiszer I have applied all requested changes and linked the corresponding commits. I would love to get a re-review soonish so we can finally get this PR merged.

mustermeiszer:
Will do tonight or tomorrow!

libs/traits/src/connectors.rs
```diff
@@ -0,0 +1,58 @@
...
+/// An encoding & decoding trait for the purpose of meeting the
+/// Connectors General Message Passing Format
+pub trait Codec: Sized {
+	fn serialize(&self) -> Vec<u8>;
+	fn deserialize<I: Input>(input: &mut I) -> Result<Self, codec::Error>;
+}
+
+/// The trait required for processing outbound connectors messages.
+pub trait OutboundQueue {
+	// pub trait OutboundQueue<Sender, Message, Destination> {
```
mustermeiszer, line 27:
```suggestion
```

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,275 @@
...
+		let currency = Self::try_get_payment_currency(invest_id.clone(), currency_index)?;
+
+		// Determine post adjustment amount
+		let pre_amount = T::ForeignInvestment::investment(&investor, invest_id.clone())?;
+		let post_amount = pre_amount.ensure_sub(amount)?;
+
+		T::ForeignInvestment::update_investment(&investor, invest_id, post_amount)?;
+
+		// Burn decreased amount
+		T::Tokens::burn_from(currency, &investor, amount)?;
```
mustermeiszer, line 141: This must be removed. We can not burn already. The `T::ForeignInvestments` will not instantaneously reduce the order. Burning must happen just before we send the `ExecutedDecreaseInvestOrder` message back.

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,275 @@
...
+		T::Tokens::transfer(
+			invest_id.into(),
+			&investor,
+			&DomainLocator::<Domain> {
+				domain: sending_domain.domain(),
+			}
+			.into_account_truncating(),
+			amount,
+			false,
+		)?;
```
mustermeiszer, line 230: This must be removed. Same applies as for burning tokens. The moment we trigger the return message is the moment we must do "bookkeeping" for tokens reaching and leaving the chain.

pallets/connectors/src/lib.rs
```diff
@@ -480,143 +533,359 @@ pub mod pallet {
...
+	{
+		type Message = MessageOf<T>;
+		type Sender = DomainAddress;
+
+		#[transactional]
+		fn process(sender: DomainAddress, msg: MessageOf<T>) -> DispatchResult {
+			ensure!(
+				KnownConnectors::<T>::contains_key::<&T::AccountId>(&sender.address().into()),
+				Error::<T>::InvalidIncomingMessageOrigin
+			);
```
mustermeiszer, line 799: Not needed. All validation of message submitters happen in the Gatway. Must not be addressed here. Can also be addressed by the gatway PR.

libs/traits/src/lib.rs
```diff
@@ -377,9 +380,10 @@ pub trait TrancheCurrency<PoolId, TrancheId> {
 /// A trait, when implemented allows to invest into
 /// investment classes
 pub trait Investment<AccountId> {
-	type Error: Debug;
-	type InvestmentId;
 	type Amount;
+	type CurrencyId;
+	type Error: Debug;
+	type InvestmentId: Into<Self::CurrencyId>;
```
mustermeiszer, line 386: Why is the `Into<Self::CurrencyId>` bound needed here?

mustermeiszer (changes requested):
Two changes we need to do IMO. Otherwise, good to go.

libs/traits/src/lib.rs
NunoAlexandre: Same question 👀

pallets/connectors/src/inbound.rs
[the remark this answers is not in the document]
NunoAlexandre: @mustermeiszer should we emit an event in such a case so that we can more easily monitor and debug those scenarios?

pallets/connectors/src/inbound.rs
NunoAlexandre: Yeah maybe let's leave a TODO to do that once we have the `foreign-investments` pallet implemented and plugged in

pallets/pool-system/src/tests/mod.rs
```diff
@@ -2614,7 +2613,7 @@ mod changes {
 			util::default_pool::create();
 
 			let change = PoolChangeProposal::new([Requirement::BlockedByLockedRedemptions]);
-			let change_id = PoolSystem::note(DEFAULT_POOL_ID, change).unwrap();
+			let _change_id = PoolSystem::note(DEFAULT_POOL_ID, change).unwrap();
```
NunoAlexandre, line 2616: Warnings we fix 🖖

runtime/common/src/lib.rs
```diff
@@ -471,152 +471,3 @@ pub mod changes {
 		}
 	}
 }
-
-pub mod connectors {
```
NunoAlexandre, line 475: 💔

NunoAlexandre (commented):
GH tells me I left 16 comments but I think this is pretty much ready apart from Frederik's last review which are important points.

Amazing work, @wischli 💯

libs/traits/src/connectors.rs
```diff
@@ -0,0 +1,58 @@
...
+/// The trait required for processing incoming connectors messages.
+pub trait InboundQueue {
+	/// The sender type of the incoming message.
+	type Sender;
+
+	/// The connector message enum.
+	type Message;
+
+	/// Process a message from the inbound queue.
+	fn process(sender: Self::Sender, msg: Self::Message) -> DispatchResult;
```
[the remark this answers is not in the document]
wischli: The Gateway PR called this method `submit` indeed. Since this replaced the previous `handle` function, I wanted to give it a more active name. However, since the Gateway will actually call this, `submit` makes more sense thinking about this retrospectively.

libs/traits/src/lib.rs
wischli: Good catch, this can just be removed!

pallets/investments/src/lib.rs
```diff
@@ -1119,15 +1127,21 @@ where
...
 
+	fn accepted_payout_currency(
+		investment_id: Self::InvestmentId,
+		currency: Self::CurrencyId,
+	) -> bool {
+		T::Accountant::info(investment_id)
+			.map(|info| info.payment_currency() == currency)
+			.unwrap_or(false)
+	}
+
```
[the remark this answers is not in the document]
wischli: I was wondering the same in https://github.com/centrifuge/centrifuge-chain/pull/1364#discussion_r1239514314

pallets/investments/src/lib.rs
```diff
@@ -1410,3 +1424,31 @@ where
...
+	type InvestmentId = T::InvestmentId;
+	type Result = ();
+
+	fn collect_investment(
+		who: T::AccountId,
+		investment_id: Self::InvestmentId,
+	) -> Result<Self::Result, Self::Error> {
+		Pallet::<T>::do_collect_invest(who, investment_id)
+			.map_err(|e| e.error)
+			.map(|_| ())
```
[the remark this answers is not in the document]
wischli: Right now, `do_collect_invest` would return the post dispatch weight. We might need to add the collected amount in a subsequent PR which handles the `Executed*` messages.

pallets/investments/src/lib.rs
```diff
@@ -1410,3 +1424,31 @@ where
...
+	}
+
+	fn collect_redemption(
+		who: T::AccountId,
+		investment_id: Self::InvestmentId,
+	) -> Result<Self::Result, Self::Error> {
+		Pallet::<T>::do_collect_redeem(who, investment_id)
+			.map_err(|e| e.error)
+			.map(|_| ())
```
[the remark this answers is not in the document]
mustermeiszer: In the future yes. But given we want to close that I thought, this will be part of the foreign investments PR.

libs/traits/src/connectors.rs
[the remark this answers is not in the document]
mustermeiszer: I would also go with `submit`.

mustermeiszer:
> @mustermeiszer should we emit an event in such a case so that we can more easily monitor and debug those scenarios?

GH not showing what this is related to. Could you list it here again @NunoAlexandre  ^^

wischli:
> GH not showing what this is related to. Could you list it here again @NunoAlexandre ^^

I also had to dig: https://github.com/centrifuge/centrifuge-chain/pull/1364#discussion_r1236639487

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,275 @@
...
+};
+
+use crate::{pallet::Error, Config, GeneralCurrencyIndexOf, Pallet, PoolIdOf, TrancheIdOf};
+
+impl<T: Config> Pallet<T> {
+	/// Executes a transfer from another domain exclusively for
+	/// non-tranche-tokens.
+	///
+	/// Directly mints the currency into the receiver address.
+	pub fn do_transfer(
```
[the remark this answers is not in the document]
wischli: Kept `handle` prefix for `transfer` and `transfer_tranche_tokens` as these calls also exist as extrinsics.

NunoAlexandre (approved):
Ready to 🚢 

Kudus for your work, perseverance, and patience, @wischli 🎖️

pallets/connectors/src/inbound.rs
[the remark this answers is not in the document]
mustermeiszer: Probably a good idea. Let's have this when we do pallet foreign investments

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,263 @@
...
+	/// Decreases an existing investment order of the investor.
+	///
+	/// Directly burns the decreased investment amount from the investor
+	/// account.
+	///
+	/// Initiates a return `ExecutedDecreaseInvestOrder` message to refund the
+	/// decreased amount on the source domain. The dispatch of this message is
+	/// delayed until the execution of the investment, e.g. at least until the
+	/// next epoch transition.
+	pub fn do_decrease_invest_order(
```
mustermeiszer, line 120: We got really inconsistent here now. Haha.

pallets/connectors/src/inbound.rs
```diff
@@ -0,0 +1,263 @@
...
+		let currency = Self::try_get_payment_currency(invest_id.clone(), currency_index)?;
+
+		// Determine post adjustment amount
+		let pre_amount = T::ForeignInvestment::investment(&investor, invest_id.clone())?;
+		let post_amount = pre_amount.ensure_sub(amount)?;
+
+		T::ForeignInvestment::update_investment(&investor, invest_id, post_amount)?;
+
+		// Burn decreased amount
+		T::Tokens::burn_from(currency, &investor, amount)?;
```
mustermeiszer, line 138: Here we keep the burn, below we removed the transfer.

mustermeiszer (dismissed):
Approve for now. Although, we got inconsistent now in naming and handling actions after hitting pallet investments... 

But we will handle this in the subsequent PR so I am fine. Just noteing here for visibility.

pallets/connectors/src/inbound.rs
wischli: Sorry, this slipped through. Will handle this.

pallets/connectors/src/inbound.rs
wischli: Oh boy, I fully relied on search and replace without double checking 🥶 Will fix. This should not be merged.

wischli: I normalized all inbound handlers with the `handle` prefix as some prefix was required for the transfer functions.

mustermeiszer (approved):
❤️

mustermeiszer (approved):
Re-approve

NunoAlexandre (approved):
🤟🏻

PR #1367, opened 2023-05-25
Loans: LoanId out from ActiveLoan

lemunozm:
# Description

This is a nitpicky change. But ideally, a `LoanId` should not belong to the `ActiveLoan` itself. A `LoanId` is an id used to identify such a loan, and should be handled at the pallet level, not at the loan level.

To rationale this, if instead of storing the `ActiveLoans` in a `Vec`, we would store it in a `StorageMap`, we should not place the `LoanId` in the `ActiveLoan` at the very beginning, as happen currently with `CreatedLoans` and `ClosedLoans`, so how we store the loans should not affect the data they hold.

### DIFF
--- a/pallets/loans-ref/src/lib.rs
+++ b/pallets/loans-ref/src/lib.rs
@@ -247,7 +247,7 @@ pub mod pallet {
 		_,
 		Blake2_128Concat,
 		PoolIdOf<T>,
-		BoundedVec<ActiveLoan<T>, T::MaxActiveLoansPerPool>,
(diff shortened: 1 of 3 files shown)


offerijns (approved):
👍

offerijns (approved):

wischli (approved):
LGTM! Thanks for being lightning fast ⚡️

wischli (approved):
LGTM! Thanks for being lightning fast ⚡️

PR #1368, opened 2023-05-26
Loans: Add unchecked amount support

### LINKED ISSUE #1352: Loans: add `extraordinary_amount` parameter to `repay()` method.
### Description

We want a way to repay an extra amount for a loan (without checking it) and without having an effect on the debt of the loan.

That amount should be deposited directly in the pool, without being handled by the loan.

### Research/based on

[Slack thread](https://kflabs.slack.com/archives/C050E3ZN2JU/p1684392717897869).

lemunozm:
# Description

- Added `unchecked_amount` in `repay()` extrinsic
- Added `total_repaid_unchecked` to `ActiveLoan` struct.

Fixes #1352

### DIFF
--- a/pallets/loans-ref/src/lib.rs
+++ b/pallets/loans-ref/src/lib.rs
@@ -307,6 +307,7 @@ pub mod pallet {
 			pool_id: PoolIdOf<T>,
 			loan_id: T::LoanId,
 			amount: T::Balance,
+			unchecked_amount: T::Balance,
 		},
 		/// A loan was written off
 		WrittenOff {
@@ -509,6 +510,7 @@ pub mod pallet {
 				pool_id,
 				loan_id,
 				amount,
+				unchecked_amount,
 			});
 
 			Ok(())
--- a/pallets/loans-ref/src/lib.rs
+++ b/pallets/loans-ref/src/lib.rs
@@ -494,12 +494,13 @@ pub mod pallet {
 			pool_id: PoolIdOf<T>,
 			loan_id: T::LoanId,
 			amount: T::Balance,
+			unchecked_amount: T::Balance,
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
 			let (amount, _count) = Self::update_active_loan(pool_id, loan_id, |loan| {
 				Self::ensure_loan_borrower(&who, loan.borrower())?;
-				loan.repay(amount)
+				loan.repay(amount, unchecked_amount)
 			})?;
 
 			T::Pool::deposit(pool_id, who, amount)?;
--- a/pallets/loans-ref/src/lib.rs
+++ b/pallets/loans-ref/src/lib.rs
@@ -482,12 +482,13 @@ pub mod pallet {
 		/// Transfers amount borrowed to the pool reserve.
 		///
 		/// The origin must be the borrower of the loan.
-		/// If the repaying amount is more than current debt, only current debt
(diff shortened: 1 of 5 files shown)


pallets/loans-ref/src/lib.rs
```diff
@@ -494,12 +494,13 @@ pub mod pallet {
...
 			amount: T::Balance,
+			unchecked_amount: T::Balance,
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
 			let (amount, _count) = Self::update_active_loan(pool_id, loan_id, |loan| {
 				Self::ensure_loan_borrower(&who, loan.borrower())?;
-				loan.repay(amount)
+				loan.repay(amount, unchecked_amount)
 			})?;
```
offerijns, line 504: We should add `unchecked_amount: T::Balance` as the last parameter to the `Repaid` event.

lemunozm: Makes sense!

offerijns: Actually, I see what's happening now, the `amount` logged in the event will include the unchecked amount. I fear an indexing layer needs to be able to know what the checked vs unchecked amount was.

offerijns: Maybe `Loan::repay` can return separate amount and unchecked amounts, rather than adding them together?

pallets/loans-ref/src/tests.rs
```diff
@@ -1088,12 +1113,34 @@ mod repay_loan {
...
+			config_mocks(COLLATERAL_VALUE);
+			assert_ok!(Loans::repay(
+				RuntimeOrigin::signed(BORROWER),
+				POOL_A,
+				loan_id,
+				0,
+				COLLATERAL_VALUE,
+			),);
+
+			// Nothing repaid with unchecked amount,
```
offerijns, line 1139: Good straightforward test to check this behaviour 👍

offerijns (approved):
LGTM!

offerijns (approved):
LGTM!

pallets/loans-ref/src/lib.rs
lemunozm: Yes, In fact, I think `Loan::repay` shouldn't do anything with `unchecked` except stores it.

wischli (commented):
I am not fully onboard with the context but the code and changes LGTM based on the PR description.

offerijns (commented):
I think it would be good to update the comment above the `repay` extrinsic. Something like from
```rust
/// The origin must be the borrower of the loan.
/// If the repaying amount is more than current debt, only current debt
/// is transferred. The borrow action should fulfill the borrow
/// restrictions configured at [`types::LoanRestrictions`]. The `amount`
/// will be transferred from borrower to pool reserve. The portfolio
/// valuation of the pool is updated to reflect the new present value of
/// the loan.
```
to
```rust
/// The origin must be the borrower of the loan.
/// The repay action should fulfill the repay restrictions 
/// configured at [`types::RepayRestrictions`].
/// If the repaying `amount` is more than current debt, only current debt
/// is transferred. This does not apply to `unchecked_amount`, which
/// can be used to repay more than the outstanding debt.
/// The portfolio  valuation of the pool is updated to reflect the new
/// present value of the loan.
```

offerijns (approved):

PR #1370, opened 2023-05-26
Fix: Failing integration tests

mustermeiszer:
# Description
The logic previously fetched events at block zero, which where simply not existing as not populated. Substrate changed to panicking now, when somebody does this. This change fetches events only starting at block 1.


NunoAlexandre (approved):
Sweet

PR #1371, opened 2023-05-26
Algol

branan:
Add the new "algol" test network, based off of an altair runtime with two changes (gated by a feature flag)
* Algol allows pools to be created by anyone, rather than requiring governance
* Algol allows smart contracts to be deployed by anyone, rather than requiring governance

Ideally these differences will eventually be eliminated, but for now we want to keep both of those actions restricted on Altair.

The Algol genesis is 20M ALGL, distributed equally between twenty accounts:
* 3 for the three genesis collators
* 7 for extra developer accounts
* 9 for the initial council (also expected to be developers)
* 1 for the faucet account (using the same address as the current Demo faucet


offerijns (commented):
I'm not the best to review the chain spec and related changes so will let someone else approve, but
- Algol allows pools to be created by anyone, rather than requiring governance
- Algol allows smart contracts to be deployed by anyone, rather than requiring governance

definitely makes sense to me 👍

lemunozm:
As Jeroen, I'm not the best to check that part. But all changes look sane 👍🏻

NunoAlexandre (approved):
Looks correct to me!

wischli (commented):
Overall LGTM! Thanks for working on this. I have two comments:

In order to be closer to the `development` runtime, we are missing a few proxy types, [see here](https://centrifuge.hackmd.io/0ZuveSTsROKqPkoVGQfM-A#Runtime-Obstacles) for details. 

Additionally, we need to alter the CI to also build `testnet-runtime` images. I suppose it suffices to do that on releases.

branan:
@wischli Now includes the missing proxy types, as well as the keystore pallet.

I've set it to just build algol runtimes on every push for now, but we can change that if it makes sense. We don't currently have any on-tag runtime builds :shrug:

branan:
I cannot for the life of me figure out why the keystore benchmarks aren't running in altair. Would love if somebody got put a second set of eyeballs on it

offerijns:
This is a pretty wild guess, was just comparing diffs: any chance keystore needs to be added in this file? https://github.com/centrifuge/centrifuge-chain/blob/241ff57760b674e78ec7f09b1a7ddae1c3eb9f53/runtime/altair/src/weights/mod.rs

branan:
> This is a pretty wild guess, was just comparing diffs: any chance keystore needs to be added in this file? https://github.com/centrifuge/centrifuge-chain/blob/241ff57760b674e78ec7f09b1a7ddae1c3eb9f53/runtime/altair/src/weights/mod.rs

I was desperate enough to have already tried copying over the weights from development just in case the zero weights were doing something weird. No luck

wischli:
Benchmarks fixed by removing a typo `l`, see [08fa74f](https://github.com/centrifuge/centrifuge-chain/pull/1371/commits/08fa74fc31e7595bb25453dae6aecff1208ad662) @branan 😅 Amazed how the benchmark macro magic did not bubble up such error to the compiler.

thea-leake:
Nice find @wischli!

thea-leake (approved):
LGTM!

runtime/altair/src/lib.rs
```diff
@@ -1532,7 +1532,7 @@ impl<
 
 parameter_types! {
 	pub const MaxKeys: u32 = 10;
-	pub const DefaultKeyDeposit: Balance = 100 * CFG;
+	pub const DefaultKeyDeposit: Balance = 100 * AIR;
```
lemunozm, line 1535: I don't remember if I've already talked with someone about it, but maybe it can be worth it to generalize our constants and use the same CFG for all runtimes. Then, each chain will call it as they want, but internally they can be the same constants.

wischli: Actually, they are already [the same constant internally](https://github.com/centrifuge/centrifuge-chain/blob/08fa74fc31e7595bb25453dae6aecff1208ad662/runtime/altair/src/constants.rs#L25-L26) 😅 So this was just a style/consistency change.

branan:
Thanks @wischli! I am also surprised the macro lets that compile.

wischli (dismissed):
LGTM! Thanks for your additions based on my last review. Happy we solved the benchmark "issue".

runtime/altair/src/lib.rs
branan: Purely copy and paste on my part :).

It's nice to keep them separate in case we do adjust our decimals, ex when AIR goes to the moon ;)

branan: Purely copy and paste on my part :).

It's nice to keep them separate in case we do adjust our decimals, ex when AIR goes to the moon ;)

lemunozm: But being Altair a "test" of Centrifuge net, does it make sense to have a different decimal approach? I mean, the idea of this is to ensure later things will work well in the real network. If we differ on it, maybe could appear some minor precision bug no tested later in centrifuge.

lemunozm (dismissed):
LGTM!

runtime/altair/src/lib.rs
NunoAlexandre: @lemunozm you are right in theory, but in practise often that happens. For example, `DOT` is a 10-decimal token while `KSM` is 12. The point here is that it can be happen that need arises and given that internally they match to the same constant, I'd leave this as is to keep one in mind that they are different assets in reality. Also, keep in mind that soon ™️  cross-network transfers will be possible, so we could have AIR being transferred to Centrifuge and CFG to Altair.

Issue #1279, opened 2023-03-22
Oracle valuation method


lemunozm:
### Description

We want to be able to offer a new `ValuationMethod` for `pallet-loans`. This valuation will get the value from the outside through an oracle. 

### Research/based on

- [Slack: wg-oracle-valuation](https://kflabs.slack.com/archives/C050E3ZN2JU)
- [Requirements](https://centrifuge.hackmd.io/qcTAtX_9Q3WDXHRt1o9oJg?view)
- [Specs](https://centrifuge.hackmd.io/ayzluh1RR-aNWTZh6NMZ8g?view)

### Tracking

- [x] Write off triggers: #1305
- [x] Data Collection traits #1324
- [x] Implement `pallet-collection-data-feed`: #1324
- [x] Adapt `pallet-loans`: #1311
- [x] ~~Migration~~ => Not needed
- [ ] Add oracles to development runtime: #1377
- [x] ~~Add oracles to altair & centrifuge runtimes~~ => It's own future task
- [x] Fix benchmarks: #1372 
- [x] ~~Integration tests~~ => It's own task #1375

PR #1372, opened 2023-05-29
Loans: Support for benchmarking prices

lemunozm:
# Description

- Added a `MaxCollectionSize` associated type that is only used under `runtime-benchmarks`.

By now, I think always using the maximum size of a price collection in benchmarks is ok. In comparison with the work done in `portfolio_valuation` over each `ActiveLoan`, deserializing once a small or a big vector of prices should not affect too much the final value.

If in the future, we need to struggle more with the benchmarks, we can add this as a variable parameter.

Associated to #1279

### DIFF
--- a/libs/mocks/src/data.rs
+++ b/libs/mocks/src/data.rs
@@ -10,6 +10,8 @@ pub mod pallet {
 		type CollectionId;
 		type Collection: DataCollection<Self::DataId>;
 		type Data;
+		#[cfg(feature = "runtime-benchmarks")]
+		type MaxCollectionSize: Get<u32>;
 	}
 
 	#[pallet::pallet]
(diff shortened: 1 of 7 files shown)


offerijns (commented):
I would have expected that we needed to enforce the `MaxCollectionSize` somewhere, e.g. when creating a new loan with a specific price ID. Is there already somewhere implicitly checked?

lemunozm:
The price collection is only increased when a new price is registered using `DataCollector::register_id()` once a loan becomes active. Inside the `register_id()` is the check I think you're looking for.

offerijns (approved):

lemunozm:
Rebased

offerijns (approved):

PR #1376, opened 2023-05-30
Add pallet connectors gateway

cdamian:
# Description

This PR adds the `Connectors Gateway` pallet which will be used to process inbound and outbound Connectors messages.

## Changes and Descriptions

- `Connectors Gateway` pallet - entrypoint for Connectors messages that's using the `InboundQueue` for processing incoming messages in the case of successful validation, and that implements the `OutboundQueue` for processing outgoing messages that are handled by the according XCM or EVM router.
- Connectors Gateway routers - separate crate that's part of the `Connectors Gateway` pallet which holds the routers that are used in the gateway.

# Checklist:


pallets/connectors-gateway/connectors-gateway-routers/src/axelar_evm.rs
```diff
@@ -0,0 +1,217 @@
...
+		//
+		// Sender T::AccountId -> Sender H160
+		//
+		// pallet_evm::Pallet::account_basic:
+		//
+		// Sender H160 -> Unknown T::AccountId
+		//
+		// This will most likely mean that the information for
+		// the unknown account will be empty i.e. 0 nonce and 0 balance.
+		pallet_evm::Pallet::<T>::call(
```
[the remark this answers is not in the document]
branan: tl;dr - we probably need to have a broader sync about requirements for calling EVM from substrate code and how we want to deal with gas, weights, and fees - there are lots of complications here.

Normally, the EVM wants to handle charging transaction fees for us. This is fine, until it's not. It's kinda not fine here.

* As noted, the AccountID doesn't round-trip cleanly.
* We need to account for the weight of the EVM work. Normally this is done by a `PostDispatchInfo` with `pays_fee: false`, since the EVM has already charged the fees. But in this case we have additional work in the top-level extrinsic that needs to be paid for.

We have a few ways of dealing with this:
* We can continue to use `pays_fee: false`, by letting the EVM do its thing. We'd also need to manually handle charging the fee for the non-EVM parts of any extrinsic which reaches this code path. This will require making accounts round-trip so that the EVM can charge the correct top-level AccountID for this call.
* We can pay fees normally with a standard `PostDispatchInfo`. This means making the EVM `OnChargeTransaction` handle the case where it should lock fees but not finalize them, letting the total weight be handled at the very end.
  * I think for this to work correctly we would need to adjust fee calculations so that EVM gas prices are tied to substrate weights by using `pallet_transaction_payment` instead of a separate EVM price computation. Otherwise the EVM could lock too few tokens for payment if gas prices are cheaper than substrate weight prices.
* We could reimplement the parts of the `Runner` that are needed to nicely call from within an extrinsic, bypassing all EVM fee logic entirely and just getting back a weight that we can add to the rest of our tranasaction.

All of this gets more complicated if we expect the call to come from an extrinsic which might have been triggered by an Ethereum call in the first place, because now we're dealing with gas->weight->gas conversions and possibly fees being charged in multiple places. I haven't dived into those codepaths yet at all.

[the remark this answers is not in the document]
branan: As for the best way to actually _invoke_ the call: I belive it'll be `pallet_ethereum::Pallet<T>::execute(origin, transaction)`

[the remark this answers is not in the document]
cdamian: Thank you for the input, I think it makes sense to refrain from making any changes until we figure out how to deal with the fee issue that you described. Initially I wanted to change to `pallet_ethereum::Pallet<T>::execute` but since that might also not suffice I'll hold it until we clarify that and focus on the moonbeam router tests for the time being.

cc @mustermeiszer Please check this and let us know if we should sync more on this topic.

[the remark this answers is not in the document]
branan: Another complication: since this is for axelar, we probably need a transaction + events to appear on the Ethereum block. So maybe this needs to look more like how Moonbeam does `ethereum_xcm`?

Regardless, I still think it'd be good for us to fully understand the requirements here so we can sort out the right solution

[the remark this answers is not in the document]
mustermeiszer: @branan I think we can get away with charging the fees twice. Once for the actual calley, the user who wants to use the bridge, and then secondly on the EVM side charging the pallet-account that we accept on the ethereum side. 

All we need to do is (re-)fund the pallet-account form the treasury. Either automatically or via democracy.

libs/mocks/src/connectors.rs
```diff
@@ -0,0 +1,73 @@
...
+	fn deserialize<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let call_type = input.read_byte()?;
+
+		match call_type {
+			0 => Ok(MessageMock::First),
+			1 => Ok(MessageMock::Second),
+			_ => Err("unsupported message".into()),
+		}
+	}
+}
```
lemunozm, line 33: Maybe I'm going too far, but maybe a simple number can be used as a message to avoid this type?

An later use something like:
```rust
domain_router.send(sender, 42).unwrap()
```

lemunozm (commented):
A couple of comments about mocking, but looks quite good!

libs/mocks/src/connectors.rs
lemunozm: Mmm... that's weird 🤔 . I'm pretty sure `Codec` and `Serialize` is implemented for basic types. Could it be because the compiler doesn't know exactly what that `42` is? Sometimes in these cases, you need to help the compiler with something like `42u32`.

cdamian: You are probably referring to a different `Codec` trait, I'm referring to the one that we have in connectors - https://github.com/centrifuge/centrifuge-chain/blob/add-pallet-connectors-gateway/libs/traits/src/connectors.rs#L19

lemunozm: Ouch, you're right! Never mind then

lemunozm: Although... would it make sense to add the basic types to the `Codec` trait along with the implementation (as it happens with the substrate `Coded` and `Serialize` traits)? @NunoAlexandre @wischli

pallets/ethereum-transaction/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+		fn get_transaction_signature() -> Option<TransactionSignature> {
+			TransactionSignature::new(
+				TRANSACTION_RECOVERY_ID,
+				H256::from_low_u64_be(1u64),
+				H256::from_low_u64_be(1u64),
+			)
+		}
+	}
+
+	impl<T: Config> EthereumTransactor for Pallet<T> {
```
cdamian, line 59: cc @branan

pallets/connectors-gateway/connectors-gateway-routers/src/axelar_evm.rs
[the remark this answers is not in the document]
mustermeiszer: @cdamian I would for now move forward with this! We should keep in mind that we need to look into fee managmanet later on. WDYT>

mustermeiszer:
Coming back to this: https://github.com/centrifuge/centrifuge-chain/pull/1403#discussion_r1236906068 

I think we will just benchmark the call as we usually do. IIUC this will result in the caller having to pay fees twice. For now I think it is fine, we must just create a todo to fix this! 

The only thing we must ensure is that these lines 

```rust
// Use the same conversion as the one used in `EnsureAddressTruncated`.
let sender_evm_address = H160::from_slice(&sender.as_ref()[0..20]);
```

create the backwards conversion to the mapping that `type AddressMapping: AddressMapping<Self::AccountId>;` is doing. So we need roundtripping for that. 

cc @branan for consolidation.

pallets/connectors-gateway/connectors-gateway-routers/src/ethereum_xcm.rs
```diff
@@ -0,0 +1,256 @@
...
+	pub xcm_domain: XcmDomain<CurrencyIdOf<T>>,
+	pub _marker: PhantomData<T>,
+}
+
+impl<T> EthereumXCMRouter<T>
+where
+	T: frame_system::Config + pallet_xcm_transactor::Config + pallet_connectors_gateway::Config,
+{
+	pub fn do_init(&self) -> DispatchResult {
+		pallet_xcm_transactor::Pallet::<T>::set_transact_info(
```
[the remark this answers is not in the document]
NunoAlexandre: The idea was to require a governance proposal in order to submit and thus be able to use a router, for the sake of transparency. I don't necessarily oppose this approach tho, but it could happen that by the time this is published we might need to change some of the values through a proposal anyways. I think it's fine to keep it, unless if someone has different opinions 👍

pallets/connectors-gateway/connectors-gateway-routers/src/axelar_evm.rs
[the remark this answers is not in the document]
cdamian: @mustermeiszer I agree.

pallets/connectors-gateway/connectors-gateway-routers/src/ethereum_xcm.rs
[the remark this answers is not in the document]
cdamian: Adding a router will also be done via a governance proposal, this logic will be triggered when the router is added. Just wanted to mention that.

runtime/development/src/lib.rs
```diff
@@ -707,7 +713,7 @@ parameter_types! {
 	pub const LaunchPeriod: BlockNumber = 7 * DAYS;
 	pub const VotingPeriod: BlockNumber = 7 * DAYS;
 	pub const FastTrackVotingPeriod: BlockNumber = 3 * HOURS;
-	pub const InstantAllowed: bool = false;
+	pub const InstantAllowed: bool = true;
```
cdamian, line 716: Are we OK with this for our development runtime? Otherwise, I'm not sure how to speed up the current integration test that goes through democracy.

pallets/connectors-gateway/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+	///
+	/// This can only be modified by an admin.
+	#[pallet::storage]
+	pub(crate) type ConnectorsAllowlist<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat,
+		Domain,
+		BoundedVec<DomainAddress, T::MaxConnectorsPerDomain>,
+		ValueQuery,
+	>;
```
NunoAlexandre, line 128: @mustermeiszer @cdamian what about making this a `StorageDoubleMap`?  By doing so, we don't have the `O(n)` expense of checking whether a connector address is whitelisted for a given `Domain` but rather we can make it `O(1)`.

If we still need the `MaxConnectorsPerDomain` max in that case, we can always check that the `ConnectorsAllowlist::get(domain).keys().len() <= MaxConnectorsPerDomain`.

Regarding details, we could make the `Value` of the `StorageDoubleMap` a simple unit (`()`):

``` rust
#[pallet::storage]
	pub(crate) type ConnectorsAllowlist<T: Config> = StorageDoubleMap<
		_,
		Blake2_128Concat,
		Domain,
    Blake2_128Concat,
		DomainAddress,
    ()
		ValueQuery,
	>;
```

pallets/connectors-gateway/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+				}
+
+				connectors
+					.try_push(connector.clone())
+					.map_err(|_| Error::<T>::MaxConnectorsReached)?;
+
+				Self::deposit_event(Event::ConnectorAdded { connector });
+
+				Ok(())
+			})
```
NunoAlexandre, line 206:
```suggestion
				ensure!(
				  !<ConnectorsAllowlist<T>>:contains(connector.domain(), connector), 
				  Error::<T>::ConnectorAlreadyAdded
				);
     
				<ConnectorsAllowlist<T>>::insert(connector.domain(), connector, ());
				Self::deposit_event(Event::ConnectorAdded { connector });

				Ok(())
			})
```

NunoAlexandre (commented):
One suggestion 👀

pallets/connectors-gateway/src/lib.rs
cdamian: I'm cool with this, awesome suggestion!

pallets/connectors-gateway/src/lib.rs
NunoAlexandre: It would be even better if `insert` fails if it already exists, but as I looked into the code I couldn't find that error case being handled there

runtime/development/src/lib.rs
cdamian: Ah, sure. Thanks for this!

pallets/connectors-gateway/connectors-gateway-routers/src/axelar_evm.rs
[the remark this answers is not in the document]
cdamian: Resolving this for now, let's start another discussion on this topic if needed.

pallets/connectors-gateway/src/lib.rs
cdamian: Done.

pallets/connectors-gateway/src/lib.rs
cdamian: Done, thanks again!

mustermeiszer (dismissed):
LGTM!

mustermeiszer:
Failing tests can be fixed by applying the latest changes of the precompile branch here. I just do not wnat the node changes in this PR. THe latest precompile stuff can already be included.

mustermeiszer (approved):
Not an perfect in debth review, but I have seen this PR before and it still looks good. Great work again @cdamian! 

But would be good to get another pair of eyes on this at least! ^^

cdamian:
This PR will be split into multiple smaller PRs that focus on one part.

cdamian:
Part 1 - adding `Connectors Gateway` pallet - https://github.com/centrifuge/centrifuge-chain/pull/1456

NunoAlexandre:
@cdamian do we want to close this?

NunoAlexandre:
> > @cdamian do we want to close this?
> 
> @NunoAlexandre Let's keep it open until we merge all remaining parts, it's also helpful for me to keep track of things ^^

Sure, I was just trying to follow these gateway-related PRs an given your comments [https://github.com/centrifuge/centrifuge-chain/pull/1376#issuecomment-1635746946] above I thought this PR was now obsolete given those two Parts have been merged already 👍

cdamian:
Closing this since the [last related PR](https://github.com/centrifuge/centrifuge-chain/pull/1478) was just merged. Other work related to the topic, such as the [precompiles](https://github.com/centrifuge/centrifuge-chain/pull/1484) will be done separately and documented accordingly.

PR #1377, opened 2023-05-30
Loans: Add oracles to development runtime

### LINKED ISSUE #1279: Oracle valuation method

lemunozm:
# Description

This PR adds price support to `pallet-loans` for `development` runtime.

Fixes #1279

## Changes and Descriptions

- Added `pallet-membership` pallet to handle oracle members.
- Added `orml-oracle` pallet to fetch prices
- Added `pallet-data-collector` to organize prices in collections.
- Added common utility structs into `runtime-common`. 

WIP:
 - [ ] Looking for a way to feed the oracle from the benchmark without attaching the oracle directly.


runtime/development/src/lib.rs
```diff
@@ -1295,6 +1295,46 @@ impl pallet_xcm_transactor::Config for Runtime {
 parameter_types! {
 	pub const MaxActiveLoansPerPool: u32 = 50;
 	pub const MaxWriteOffPolicySize: u32 = 10;
+	pub const MaxHasDispatchedSize: u32 = 1;
+	pub const MaxPools: u32 = 50;
+	pub const MaxPriceOracleMembers: u32 = 10;
+	pub RootOperatorOraclePrice: AccountId = PalletId(*b"orac_pri").into_account_truncating();
```
offerijns, line 1301: I guess the pallet ID here should be part of https://github.com/centrifuge/centrifuge-chain/blob/main/libs/types/src/ids.rs?

runtime/development/src/lib.rs
```diff
@@ -1295,6 +1295,46 @@ impl pallet_xcm_transactor::Config for Runtime {
 parameter_types! {
 	pub const MaxActiveLoansPerPool: u32 = 50;
 	pub const MaxWriteOffPolicySize: u32 = 10;
+	pub const MaxHasDispatchedSize: u32 = 1;
+	pub const MaxPools: u32 = 50;
```
offerijns, line 1299: This is not really `MaxPools`, but rather `MaxPoolsWithExternalPrices`, right?

runtime/development/src/lib.rs
```diff
@@ -1295,6 +1295,46 @@ impl pallet_xcm_transactor::Config for Runtime {
 parameter_types! {
 	pub const MaxActiveLoansPerPool: u32 = 50;
```
offerijns, line 1296: Unrelated but could we bump this? I think we soon will want to test pools with larger # of assets per pool.

lemunozm: What is an ok number here? Maybe 300 or so by now?

runtime/development/src/lib.rs
lemunozm: Added there, and I've removed the `LOANS_PALLET_ID` because it's no longer used.

This was another pending talk. To be honest, not sure if we want a pallet id as a root account. This account should be handled by us or by root/conceal. I think adding a simple public key account here we do not let us handle this later.

runtime/development/src/lib.rs
offerijns: What's the theoretical upper limit now based on how many loans can fit in say half a block based on the benchmarks?

Assuming the theoretical is now a multiple of 1000, I would probably just set this to something like 1000, at least in the development env, for testing purposes

lemunozm: Ok! I do not remember the number well, but once I had the benchmark working I can test it and adjust this number if needed. By now, I set it to 1000 👍🏻

mustermeiszer (dismissed):

lemunozm:
Thanks for the review! It's not ready yet because of this: https://github.com/open-web3-stack/open-runtime-module-library/issues/919

But I think I can get a workaround in the meantime.

runtime/development/src/lib.rs
```diff
@@ -1302,13 +1371,13 @@ impl pallet_loans_ref::Config for Runtime {
...
 	type LoanId = LoanId;
-	type MaxActiveLoansPerPool = MaxActiveLoansPerPool;
+	type MaxActiveLoansPerPool = MaxActiveLoansPerPoolBenchmark;
 	type MaxWriteOffPolicySize = MaxWriteOffPolicySize;
 	type NonFungible = Uniques;
 	type Permissions = Permissions;
 	type Pool = PoolSystem;
-	type PriceId = PriceId;
-	type PriceRegistry = pallet_loans_ref::util::NoPriceRegistry<Runtime>;
+	type PriceId = OracleKey;
```
lemunozm, line 1379: I think this is a breaking change if we do it in `altair` or `centrifuge`. Do we need a migration for it, or are we still in the same state as the previous release, where it's still not used?

pallets/loans-ref/src/benchmarking.rs
```diff
@@ -221,14 +224,21 @@ where
...
 
+		// TODO: ideally should this be computed by a parameter?
 		for i in 0..MaxCollectionSizeOf::<T>::get() {
-			let price_id = T::PriceId::from(i + 1);
+			let price_id = i.into();
+			// This account is different in each iteration because of how oracles works.
+			// This restriction no longer exists once
+			// https://github.com/open-web3-stack/open-runtime-module-library/pull/920 is merged
+			let feeder = account("feeder", i, 0);
+			T::PriceRegistry::feed_value(feeder, price_id, 0.into()).unwrap();
```
lemunozm, line 241: This is quite tricky. Because of how oracles are implemented, the same user can not feed twice in the same block. And the internal trait they use (`DataFeeder`) only allows us to feed one price id each time.

That's why we need to create a new account per the price we want price here. And consequently, we need to mock the member's pallet in the runtime. To allow to use all these users.

lemunozm (commented):
Some important comments below:

runtime/development/src/lib.rs
```diff
@@ -1291,8 +1293,75 @@ impl pallet_xcm_transactor::Config for Runtime {
...
 	pub const MaxWriteOffPolicySize: u32 = 10;
+	pub const MaxPriceOracleMembers: u32 = 10;
+	// The benchmark distintion can be removed once
+	// <https://github.com/open-web3-stack/open-runtime-module-library/issues/920> be merged.
+	pub const MaxHasDispatchedSize: u32 = production_or_benchmark!(
+		MaxPriceOracleMembers::get(),
+		MaxActiveLoansPerPool::get()
+	);
+	pub const MaxPoolsWithExternalPrices: u32 = 50;
+	pub RootOperatorOraclePrice: AccountId = PRICE_ORACLE_PALLET_ID.into_account_truncating();
```
lemunozm, line 1312: Not sure how this can be used from outside without the private key. But I need to put some account as a root.

runtime/development/src/lib.rs
offerijns: I would think it doesn't require a migration if there are no loans with external pricing, right?

lemunozm: I think so, it all depends on that

runtime/centrifuge/src/lib.rs
```diff
@@ -1573,7 +1573,7 @@ impl pallet_loans_ref::Config for Runtime {
 	type NonFungible = Uniques;
 	type Permissions = Permissions;
 	type Pool = PoolSystem;
-	type PriceId = PriceId;
+	type PriceId = u32;
```
mustermeiszer, line 1576: Why make it a raw type again?

mustermeiszer (approved):

runtime/centrifuge/src/lib.rs
lemunozm: Our real type is `OracleKey`. It doesn't matter what you put as type there in altair & centrifuge runtimes because the price feature is not allowed as long as `NoPriceRegistry` type is attached to the loan pallet configuration.

Basically, I removed the `PriceId` from `cfg-primitives` and I placed an `u32` as a placeholder.

offerijns (approved):

PR #1383, opened 2023-06-05
Loans: interest rate value stored without penalty

lemunozm:
# Description

In the previous pallet loan refactor, we modified the loan's interest rate to carry on the penalty with it. This works, but I think it has some downsides:

- From an understanding point of view, all `LoanInfo` attributes from the creation are immutable. Nevertheless, `interest_rate` is mutable under write-off actions, which can be ackward.
- With the future loans modification feature, having a modified `interest_rate` with the penalty makes things harder to handle and understand. 

This PR reverts that change. Originally it was done so because it was difficult to know where to use the `interest_rate` and where to use the `interest_rate_with_penalty`. Nevertheless, after oracle's refactor, we have all `interest_rate` computations under the same place, and now it is easier to manage.

**NOTE: This can imply a migration and a UI change to correctly show the `interest_rate` and the penalty. However, both things should be easy to do.**

### DIFF
--- a/pallets/loans-ref/src/loan.rs
+++ b/pallets/loans-ref/src/loan.rs
@@ -368,7 +368,7 @@ impl<T: Config> ActiveLoan<T> {
 
 	pub fn write_off(&mut self, new_status: &WriteOffStatus<T::Rate>) -> DispatchResult {
 		if let ActivePricing::Internal(inner) = &mut self.pricing {
-			inner.update_penalty(new_status.penalty)?;
+			inner.set_penalty(new_status.penalty)?;
 		}
 
 		self.write_off_percentage = new_status.percentage;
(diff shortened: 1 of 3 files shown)


lemunozm:
EDIT: it could be not really needed for loan modification feature (WIP)

mustermeiszer:
Looks good, but do you need a review right now or should we wait until modifitcation is done>

lemunozm:
This can be merged before (it was the intention). I need to add tests to the modification PR yet. But no hurry at all!

lemunozm:
After implementing it in the modification PR, I'm not sure if we really need it (so we do not need to merge this before the other). 

This PR depends more on a philosophical way of how to see the `interest_rate`. Do we want to hold the base and add the penalty each time? or do we want to carry on the modification in the `interest_rate`? What is easier for the UX?

lemunozm:
Finally not needed

PR #1384, opened 2023-06-05
Loans: `ChangeGuard` trait and loans modifications

### LINKED ISSUE #1207: Implementation of `ChangeGuard` trait

lemunozm:
# Description

This PR overwrites #1222 with some simplifications.

- The mutation storage is no longer in `pallet-loans`. Now the implementor of `ChangeGuard`, is who should allocate those mutations. This way, we extract from `pallet-loans` and any future `ChangeGuard`'s user the need to manage mutations with its own storage
- `pallet-loans` no longer has a dependency on `CfgChange`. The runtime is who transforms the `LoanMutation` type into a `CfgChange` that the implementor of `ChangeGuard` understand.

Fixes #1207

## Changes and Descriptions

- [x] Added `ChangeGuard` trait.
- [x] Added `ChangeGuard` trait mock.
- [x] Added 2 new extrinsic for loan modifications to `pallet-loans`.
- [x] Added a utility to configure `pallet-loans` without the modification feature.
- [x] Legacy tests passing
- [x] Add new tests
- [x] Runtimes working with the feature disabled


lemunozm (commented):
Important parts commented below:

offerijns (approved):
Clippy is failing but LGTM!

lemunozm:
I've renamed the main functions of this feature to `propose()` and `apply()`, I think these names are more scalable for future applications. Examples:

```rust
let change = Change::Loan(loan_id, LoanMutation::InterestRate(...));
let change_id = Loans::propose(origin, pool_id, change)?;
Loans::apply(origin, pool_id, change_id);
```

But if you do not like or suggest others, please tell me.

offerijns:
@lemunozm I do think these extrinsic names are a bit confusing 😅  Since the pallet name is loans, these will be `loans.propose()` and `loans.apply()`, which implies you are proposing and applying loans, rather than changes. Maybe just `propose_change/mutation` and `apply_change/mutation`?

lemunozm:
I agree, it makes total sense what you say! 👍🏻  

I'll add the suffix `_change` that I think fits better with the parameter name.

pallets/loans-ref/src/lib.rs
```diff
@@ -33,6 +33,13 @@
 //! | [`Pallet::update_write_off_policy()`]    | PoolAdmin |
 //! | [`Pallet::update_portfolio_valuation()`] | Any       |
 //!
+//! The following actions are performed based on changes:
+//!
+//! | Extrinsics                     | Role      |
+//! |--------------------------------|-----------|
+//! | [`Pallet::propose_change()`]   | LoanAdmin |
+//! | [`Pallet::apply_change()`]     | Borrower  |
```
offerijns, line 41: Why the differing roles for this btw? Not sure it's wrong, just curious about the choice. Maybe `apply_change` could even be permissionless..?

pallets/loans-ref/src/lib.rs
```diff
@@ -681,6 +715,50 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			pool_id: PoolIdOf<T>,
+			change_id: T::ChangeId,
+		) -> DispatchResult {
+			ensure_signed(origin)?;
+
+			let Change::Loan(loan_id, mutation) = T::ChangeGuard::released(pool_id, change_id)?;
+
+			let (_, _count) = Self::update_active_loan(pool_id, loan_id, |loan| {
+				loan.modify_with(mutation.clone())
```
offerijns, line 751: Realized we are not interchanging `change`, `modify/modification` and `mutation`. Maybe good to remove one of those synonyms?

lemunozm: I understand this is as follows:
- Change: something more general, can be whatever, can be a loan mutation, a policy,... and it's more at `ChangeGuard` level.
- modify (as a verb): Action that _modifies_ an entity based on some input data.
- Mutation: _Small information unit_ that can be applied to another entity to modify it.

Maybe `Mutation` can be `Modification` instead or vice versa, and avoid one synonym.

offerijns: That actually makes sense :) I think I'd personally still change `Mutation` to `Modification` then, but I can also see the reasoning for separating it.

lemunozm: And `modify_with()` to `mutate_with()` would also work for you? To leave the `Mutation` name that I think expresses better that it is _small information unit_ ready to be applied. I feel like `Modification` is too open. WDYT? Maybe I'm too nitpick 😅

offerijns: Yeah I think that's fine!

pallets/loans-ref/src/loan.rs
```diff
@@ -411,6 +411,22 @@ impl<T: Config> ActiveLoan<T> {
...
+			LoanMutation::InterestPayments(payments) => self.schedule.interest_payments = payments,
+			LoanMutation::PayDownSchedule(schedule) => self.schedule.pay_down_schedule = schedule,
+			LoanMutation::Internal(mutation) => match &mut self.pricing {
+				ActivePricing::Internal(inner) => inner.modify_with(mutation)?,
+				ActivePricing::External(_) => {
+					Err(Error::<T>::from(ModificationError::InternalPricingExpected))?
+				}
+			},
+		};
+
```
offerijns, line 426: I think this should actually trigger a recalculation of the PV? Since many (almost all) of these changes will affect the valuation

offerijns: Ah, that's a nice design!

pallets/loans-ref/src/lib.rs
```diff
@@ -17,14 +17,14 @@
...
-//! | [`Pallet::repay()`]                | Borrower  |
-//! | [`Pallet::write_off()`]            | Any       |
-//! | [`Pallet::admin_write_off()`]      | LoanAdmin |
-//! | [`Pallet::close()`]                | Borrower  |
+//! | Extrinsics                    | Role      |
+//! |-------------------------------|-----------|
+//! | [`Pallet::create()`]          | Borrower  |
+//! | [`Pallet::borrow()`]          | Borrower  |
+//! | [`Pallet::repay()`]           | Borrower  |
+//! | [`Pallet::write_off()`]       | Any       |
```
offerijns: Super minor nit, but: to be consistent with `Pallet::apply_change()`, I would remove `Any` here

lemunozm: Makes sense, done!

pallets/loans-ref/src/lib.rs
```diff
@@ -681,6 +710,51 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			pool_id: PoolIdOf<T>,
+			loan_id: T::LoanId,
+			mutation: Mutation<T::Rate>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			Self::ensure_role(pool_id, &who, PoolRole::LoanAdmin)?;
+
+			T::ChangeGuard::note(pool_id, (loan_id, mutation.clone()))?;
+
```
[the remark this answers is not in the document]
mustermeiszer: Makes sense.

mustermeiszer (dismissed):
LGTM! Gives us enough flexibility for future changes and we can channel down core changes to the pool-system and let the rest be handled by the `ChangeGuard` impl

lemunozm:
This is ready for a final review 🤓

pallets/loans-ref/src/lib.rs
```diff
@@ -314,6 +335,12 @@ pub mod pallet {
 			loan_id: T::LoanId,
 			status: WriteOffStatus<T::Rate>,
 		},
+		/// A loan was modified
```
offerijns, line 338:
```suggestion
		/// A loan was changed
```

lemunozm: Good eyes!

offerijns (approved):
LGTM!

PR #1388, opened 2023-06-07
Transfer Ethereum fees to Treasury & Author, rather than burning them

branan:
This implements a custom `OnChargeEVMTransaction` for our chain.

Unfortunately the EVM handling of base_fee+priority is slightly different from how pallet_transaction_payment handles fees+tips, so we have to reimplement quite a bit more logic than I would have ideally liked. Fortunately, it's all pretty straightforward fee handling bits.

## Changes and Descriptions

* EVM Fees are now distributed identically to Substrate fees

# Checklist:

### DIFF
--- a/runtime/common/src/evm/fees.rs
+++ b/runtime/common/src/evm/fees.rs
@@ -15,7 +15,6 @@
 
 use core::marker::PhantomData;
 
-use cfg_primitives::constants::TREASURY_FEE_RATIO;
 use frame_support::traits::{
 	Currency, ExistenceRequirement, Imbalance, OnUnbalanced, SignedImbalance, WithdrawReasons,
 };
@@ -25,10 +24,10 @@ use pallet_treasury::Pallet as Treasury;
 use sp_core::{H160, U256};
 use sp_runtime::{
 	traits::{UniqueSaturatedInto, Zero},
-	Perbill, Saturating,
+	Saturating,
 };
 
-use crate::fees::{NegativeImbalance, ToAuthor};
+use crate::fees::{DealWithFees, NegativeImbalance};
 
 type AccountIdOf<R> = <R as frame_system::Config>::AccountId;
 type Balance<R> = <Balances<R> as Currency<AccountIdOf<R>>>::Balance;
@@ -36,6 +35,7 @@ type PositiveImbalance<R> = <Balances<R> as Currency<AccountIdOf<R>>>::PositiveI
 
 type Error<R> = pallet_evm::Error<R>;
 
+/// Handler for EVM fees
 pub struct DealWithEVMFees<R>(PhantomData<R>);
 
 impl<R> OnChargeEVMTransaction<R> for DealWithEVMFees<R>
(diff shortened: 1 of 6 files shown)


runtime/common/src/evm/fees.rs
```diff
@@ -0,0 +1,130 @@
...
+		let account_id = R::AddressMapping::into_account_id(*who);
+		Balances::<R>::withdraw(
+			&account_id,
+			fee.unique_saturated_into(),
+			WithdrawReasons::FEE,
+			ExistenceRequirement::AllowDeath,
+		)
+		.map(Some)
+		.map_err(|_| Error::<R>::BalanceLow)
+	}
```
lemunozm, line 66: I wondered if we could implement this trait with calls to our `pallet-fees`. Does it make sense? i.e. in this case this would be a call to `Fees::withdraw_fee()`

runtime/common/src/evm/fees.rs
```diff
@@ -0,0 +1,130 @@
...
+
+	fn correct_and_deposit_fee(
```
lemunozm, line 68: And this, in turn, would be calls to `Fees::fee_to_author()` and `Fees::fee_to_treasury()`.

branan: `fee_to_author` and `fee_to_treasury` both do a `withdraw` internally, so they're not suitable here.

lemunozm: Ok, I understand 👍🏻

runtime/common/src/evm/fees.rs
```diff
@@ -0,0 +1,130 @@
...
+		let (base_fee, tip) = adjusted_paid.split(base_fee.unique_saturated_into());
+
+		// Handle base fee and tips
+		let (treasury_amount, mut author_amount) = base_fee.ration(
+			TREASURY_FEE_RATIO.deconstruct(),
+			(Perbill::one() - TREASURY_FEE_RATIO).deconstruct(),
+		);
+		tip.merge_into(&mut author_amount);
+		<Treasury<R> as OnUnbalanced<_>>::on_unbalanced(treasury_amount);
+		<ToAuthor<R> as OnUnbalanced<_>>::on_unbalanced(author_amount);
```
lemunozm, line 122: This section reminds me to `DealWithFees` type, located in `runtime/common/src/lib.rs`. I think the structure is the same. Maybe we can call it to avoid duplicate code:

```rust
DealWithFees::on_unbalanceds([base_fee, tip].iter());
```

Or similar.

branan: :+1: yeah, originally I had the tip handling done in `pay_priority_fee`, but since I rolled it all into one place it makes sense to call up into `DealWithFees` so we only have one place to update

runtime/common/src/evm/fees.rs
```diff
@@ -0,0 +1,130 @@
...
+		.map_err(|_| Error::<R>::BalanceLow)
+	}
+
+	fn correct_and_deposit_fee(
+		who: &H160,
+		corrected_fee: U256,
+		base_fee: U256,
+		already_withdrawn: Self::LiquidityInfo,
+	) -> Self::LiquidityInfo {
+		let Some(paid) = already_withdrawn else { return None };
```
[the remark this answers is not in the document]
branan: rustfmt also unfortunately does not handle `let else` properly yet, so we should probably only use it for small things like this. Otherwise we'll end up with a bunch of reformats when rustfmt finally gets updated

lemunozm (dismissed):
Would be great another couple of eyes, but LGTM!

runtime/common/src/evm/fees.rs
```diff
@@ -0,0 +1,124 @@
...
+		// that case we don't refund anything.
+		let refund_imbalance = Balances::<R>::deposit_into_existing(&account_id, refund_amount)
+			.unwrap_or_else(|_| PositiveImbalance::<R>::zero());
+
+		// Make sure this works with 0 ExistentialDeposit
+		// https://github.com/paritytech/substrate/issues/10117
+		// If we tried to refund something, the account still empty and the ED is set to
+		// 0, we call `make_free_balance_be` with the refunded amount.
+		let refund_imbalance = if Balances::<R>::minimum_balance().is_zero()
+			&& refund_amount > <Balance<R>>::zero()
```
wischli, line 93:
```suggestion
			&& !refund_amount.is_zero()
```

wischli (dismissed):
LGTM! I have a tiny nit that totally can be ignored. 

A dumb question from my side would be, why we do not simply apply the `impl_on_charge_evm_transaction` macro boilerplate [as done in Moonbeam](https://github.com/PureStake/moonbeam/blob/f00e9205beed9e41bff0d44a4a55ee65b70370d7/runtime/common/src/impl_on_charge_evm_transaction.rs#L18)?

mustermeiszer:
What is the state of this?

wischli:
@mustermeiszer I suppose we still want this at some point or can we close it?

mustermeiszer:
Yes, I would say so.

PR #1392, opened 2023-06-08
fix: get para_id from chain spec extension

wischli:
# Description

Today, I was debugging a block production issue on Altair with a collator (StakerSpace). This resulted in the discovery, that we do not make use of the `Extensions` struct which can be optionally added to the chain spec by setting an Extension to the  `sc_service::GenericChainSpec<runtime::GenesisConfig, Extension` which we currently do not such that it defaults to `NoExtension`:

```rust
pub type AltairChainSpec = sc_service::GenericChainSpec<altair_runtime::GenesisConfig>;

pub struct Extensions {
	/// The relay chain of the Parachain.
	pub relay_chain: String,
	/// The id of the Parachain.
	pub para_id: u32,
}
```

This definitely is not a critical issue but it leaves our start node command in a limbo in which we still depend on setting the  `parachain-id` which was [deprecated on Cumulus end of 2021](https://github.com/paritytech/cumulus/pull/739). By defaulting to an ambiguous parachain id `10001`, any node operator can run into issues debugging the block production failure as other chains do not support `parachain-id` CLI option anymore. AFAICT, most node operators copy paste their start commands. Moreover, we are lacking documentation.

**There are multiple options how to solve this, this PR just drafts one of the options with the goal to prevent having to update existing chain specs.** 

1. We just remove the default unwrapping to `10001` and soon™️ write collator documentation which makes aware of our dependency to the `parachain-id` CLI. This could create issues at some point, e.g. if new commands depend on information provided from the Chainspec Extension.
2. We go ahead with the approach of the PR. Please note that this is not the default Substrate "recommended code" as I made both parameters of `Extensions` optional. I am also aware of the fact that the code could be optimized a bit.
```diff
pub struct Extensions {
	/// The relay chain of the Parachain.
-	pub relay_chain: String,
+	pub relay_chain: Option<String>,
	/// The id of the Parachain.
-	pub para_id: u32,
+	pub para_id: Option<u32>,
}
```
3. We fully follow Substrate/Cumulus and update our existing chainspecs for persistent chains Catalyst, Altair, Centrifuge. I fear this can create issues for node operators as all of them have to point to new chain spec. In case they are using Docker, that won't be an issue.
4. Anything our crowd intelligence can come up with.

## About this approach

Since none of the specs for our persistent chains currently includes the default `Extensions` parameters, the `from_json_bytes` method panics if we don't make both of them optional. I did not find peace with the compiler when trying to make the `Extensions` optional, e.g.

```
pub type AltairChainSpec = sc_service::GenericChainSpec<altair_runtime::GenesisConfig, Option<Extensions>>;
```

# Checklist:

### DIFF
--- a/src/chain_spec.rs
+++ b/src/chain_spec.rs
@@ -46,9 +46,11 @@ use xcm::{
 const POLKADOT_TELEMETRY_URL: &str = "wss://telemetry.polkadot.io/submit/";
 
 /// Specialized `ChainSpec` instances for our runtimes.
-pub type AltairChainSpec = sc_service::GenericChainSpec<altair_runtime::GenesisConfig>;
-pub type CentrifugeChainSpec = sc_service::GenericChainSpec<centrifuge_runtime::GenesisConfig>;
-pub type DevelopmentChainSpec = sc_service::GenericChainSpec<development_runtime::GenesisConfig>;
+pub type AltairChainSpec = sc_service::GenericChainSpec<altair_runtime::GenesisConfig, Extensions>;
+pub type CentrifugeChainSpec =
+	sc_service::GenericChainSpec<centrifuge_runtime::GenesisConfig, Extensions>;
+pub type DevelopmentChainSpec =
+	sc_service::GenericChainSpec<development_runtime::GenesisConfig, Extensions>;
 
 /// Helper function to generate a crypto pair from seed
 pub fn get_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pair>::Public {
@@ -64,9 +66,9 @@ pub fn get_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pair>::Pu
 #[serde(deny_unknown_fields)]
 pub struct Extensions {
 	/// The relay chain of the Parachain.
-	pub relay_chain: String,
+	pub relay_chain: Option<String>,
 	/// The id of the Parachain.
-	pub para_id: u32,
+	pub para_id: Option<u32>,
 }
 
 impl Extensions {
@@ -112,10 +114,16 @@ where
 }
 
 pub fn centrifuge_config() -> CentrifugeChainSpec {
-	CentrifugeChainSpec::from_json_bytes(
+	let mut spec = CentrifugeChainSpec::from_json_bytes(
 		&include_bytes!("../res/genesis/centrifuge-genesis-spec-raw.json")[..],
 	)
-	.unwrap()
+	.unwrap();
+	let extension = spec.extensions_mut();
+	*extension = Extensions {
+		relay_chain: Some("polkadot".into()),
+		para_id: Some(2031),
+	};
+	spec
 }
 
 pub fn centrifuge_staging(para_id: ParaId) -> CentrifugeChainSpec {
@@ -169,7 +177,10 @@ pub fn centrifuge_staging(para_id: ParaId) -> CentrifugeChainSpec {
 		Some("centrifuge"),
 		None,
 		Some(properties),
-		Default::default(),
+		Extensions {
+			relay_chain: Some("rococo-local".into()),
+			para_id: Some(para_id.into()),
+		},
 	)
 }
 
@@ -210,7 +221,10 @@ pub fn centrifuge_dev(para_id: ParaId) -> CentrifugeChainSpec {
 		None,
 		None,
 		Some(properties),
-		Default::default(),
+		Extensions {
+			relay_chain: Some("rococo-local".into()),
+			para_id: Some(para_id.into()),
+		},
 	)
 }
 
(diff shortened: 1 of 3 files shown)


lemunozm:
Thanks a lot for your detailed investigation of this! 🎉 

> I did not find peace with the compiler when trying to make the Extensions optional, e.g.

I'm curious, what was the error here?

wischli:
> > I did not find peace with the compiler when trying to make the Extensions optional, e.g.
> 
> I'm curious, what was the error here?

I suppose it was too late for my head. Fixed in [1b47a97](https://github.com/centrifuge/centrifuge-chain/pull/1392/commits/1b47a9750251c01a425ef7434edfc8e4ab1fb00f) which is a much prettier approach IMO.

src/chain_spec.rs
```diff
@@ -112,10 +145,17 @@ where
...
-	CentrifugeChainSpec::from_json_bytes(
+	let mut spec = CentrifugeChainSpec::from_json_bytes(
 		&include_bytes!("../res/genesis/centrifuge-genesis-spec-raw.json")[..],
 	)
-	.unwrap()
+	.unwrap();
+	let extension = spec.extensions_mut();
+	*extension = Extensions {
+		relay_chain: "polkadot".into(),
+		para_id: 2031,
```
NunoAlexandre, line 155: We can use `parachains::polkadot::centrifuge::PARA_ID` here and same for similar const usages 👍

NunoAlexandre (commented):
Thank you for looking into this, and the great write-up. This is extremely helpful not only for us to understand what this PR is addressing but also as a history record of this change and for others in the community who might bump into something similar in their chains 👏 

Just left one comment :)

src/command.rs
```diff
@@ -443,8 +443,8 @@ pub fn run() -> Result<()> {
 						.chain(cli.relaychain_args.iter()),
 				);
 
-				let para_id = chain_spec::Extensions::try_get(&*config.chain_spec)
-					.map(|e| e.para_id).unwrap_or_else(|| cli.parachain_id.unwrap_or(10001));
+				let para_id = chain_spec::ExtensionsWrapper::try_get(&*config.chain_spec)
+					.map(|e| e.para_id).unwrap_or_else(|| cli.parachain_id.expect("Could not find parachain ID in chain spec or CLI."));
```
mikiquantum, line 447: I think this line is necessary. 👍

src/chain_spec.rs
```diff
@@ -112,10 +145,17 @@ where
...
 
 pub fn centrifuge_config() -> CentrifugeChainSpec {
-	CentrifugeChainSpec::from_json_bytes(
+	let mut spec = CentrifugeChainSpec::from_json_bytes(
 		&include_bytes!("../res/genesis/centrifuge-genesis-spec-raw.json")[..],
 	)
-	.unwrap()
+	.unwrap();
+	let extension = spec.extensions_mut();
+	*extension = Extensions {
```
mikiquantum, line 153: The only thing I am a bit unsure of this approach is that there are values that are not part of the genesis file that are added at runtime. I would love to have static files, that are verifiable, hashable and so on for the chain specs. 
If possible I would regenerate the specs (genesis hash should not change since this is node wrapper stuff) using the regular extension logic. 
We might have to tweak the `$env_staging` functions to regenerate the plain ones and then the raw ones. We can check locally that a node syncs up with the live networks to verify it works.

mikiquantum (commented):
Overall the approach is clean. On the other hand, I am on the fence on not having static and verifiable chain specs.

src/chain_spec.rs
wischli: I fully agree! Happy to hear that you think updating the existing chain specs should be feasible. How do we retrieve non-raw specs from our raw specs though? That's not supported by Substrate AFAICT.

src/chain_spec.rs
```diff
@@ -561,16 +625,33 @@ pub fn antares_local(para_id: ParaId) -> AltairChainSpec {
...
 		None,
 		Some(properties),
-		Default::default(),
+		development_extensions(para_id.into()),
 	)
 }
 
 pub fn algol_config() -> AltairChainSpec {
-	AltairChainSpec::from_json_bytes(&include_bytes!("../res/algol-spec.json")[..]).unwrap()
+	let mut spec =
```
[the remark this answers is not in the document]
wischli: I don't think it matters as we have to update Catalyst, Altair and Centrifuge anyways. I envision this all to happen as part of this PR. The question is whether it's fine for me to just update all specs or whether we want to make some ceremony out of it. Building a raw spec from a non-raw one should be deterministic I suppose?

src/chain_spec.rs
wischli: Updated the chain specs in [a9ba1a7](https://github.com/centrifuge/centrifuge-chain/pull/1392/commits/a9ba1a7f07669d90ed71e0f0c6787bf8ab769cd5) and reverted the `ExtensionsWrapper` in [4d2553f](https://github.com/centrifuge/centrifuge-chain/pull/1392/commits/4d2553fd12674d8ac97e12480fd4ed1f5a3b6946). Unfortunately, our non-raw specs (except for Algol which only exists in human readable format) are so outdated, that building the raw one from these fails with the current client code. They are missing quite a large number of genesis config entries that had been added in the meantime and also contain some properties which do not exist anymore such as `genssis.runtime.system.changesTrieConfig`. Luckily, the extensions properties [are not encoded in the raw specs](https://github.com/paritytech/cumulus/blob/1379520ac152a15dc6f28722538c4ba64449740d/parachains/chain-specs/asset-hub-kusama-genesis.json#L16), so I could just add them there.

mikiquantum (dismissed):
@wischli thanks! 
Can we add the 2 attributes `relay_chain` & `para_id` to the "plain" genesis JSONs of our live (Centrifuge & Altair) networks? (not only the raw ones). Although It is very unlikely that we will regenerate anything out of those, just for consistency sake. 

In addition, the "*_staging" network is a leftover from the steps of going live. I think we can get rid of those functions in the chain spec. (We do not have to do it now, but just to think about)

wischli:
> @wischli thanks! Can we add the 2 attributes `relay_chain` & `para_id` to the "plain" genesis JSONs of our live (Centrifuge & Altair) networks? (not only the raw ones). Although It is very unlikely that we will regenerate anything out of those, just for consistency sake.
> 
> In addition, the "*_staging" network is a leftover from the steps of going live. I think we can get rid of those functions in the chain spec. (We do not have to do it now, but just to think about)

Done! Thanks for reminding me about the non-raw ones.

mikiquantum (approved):
Awesome, thank you!

PR #1393, opened 2023-06-08
connectors: Wrapped token lookup

NunoAlexandre:
Part 2/2 of https://github.com/centrifuge/centrifuge-chain/issues/1350

> - Connector-wrapped tokens should have a `AssetMetadata.location` that encapsulates the EVM chain id and the specific EVM address of said token
>
>    - We, therefore, can implement a lookup for Connector-wrapped assets that knows how to lookup that data from the asset's `location`

### Changes

- Add `ConnectorsWrappedToken`, an enum type representing the Connectors Wrapped tokens we handle. At this moment, that's only EVM tokens. As described in the docs of that type, in the future it might be others such as Cosmos or Avalanche.

- We implement conversion between `ConnectorsWrappedToken` and `MultiLocation`

### Usage

- When registering a `ConnectorsWrappedToken` in the `OrmlAssetRegistry`, we define the `Metadata.location` as `ConnectorsWrappedTokenConvert<ConnectorsPalletIndex>>::convert(connectors_wrapped_token)`;

- In the connectors pallet we add a new type:
``` rust
ConnectorsWrappedTokenConvert: Convert<ConnectorsWrappedToken, MultiLocation> + Convert<MultiLocation, Result<ConnectorsWrappedToken, ()>
```

- In the `add_currency` extrinsic of the connectors pallet, we will _roughly_ do:

``` rust
fn add_currency(currency_id: T::CurrencyId, domain: Domain) {
	let meta = T::AssetRegistry::metadata(currency_id).ok_or(Error::CurrencyMetadataNotFound)?;

	ensure!(
		meta.additional.transferability.includes_connectors(),
		Error::NotConnectorsTransferable
	);
	let wrapped_token = ConnectorsWrappedTokenConvert::convert(meta.location)
		.map_err(Error::NotConnectorsWrappedToken)?;

	match wrapped_token {
		ConnectorsWrappedToken::EVM { chain_id, address } => ensure!(
			chain_id == domain.chain_id,
			Error::DifferentWrappedTokenAndDomain
		),
	}

	// send the message
}
````

### CODE REVIEW

File: libs/types/src/tokens.rs [hunk 1]
```diff
@@ -255,6 +259,64 @@ impl CrossChainTransferability {
...
+							key: address,
+						},
+					),
+			} => Ok(Self::EVM { chain_id, address }),
+			_ => Err(()),
+		}
+	}
+}
+
+impl Into<MultiLocation> for ConnectorsWrappedToken {
```
NunoAlexandre, line 303: This basically defines how we describe an EVM Connectors wrapped token as a `MultiLocation`.
The non-final structure proposed here is to have `parents: 0` to symbolise that it's relative to the local chain, and then in the `interior` we say that it's part of the  `GlobalConsensus(NetworkId::Ethereum { chain_id }),`, and finally the `AccountKey20` with the evm token address.

The way I read this is "a Centrifuge-wrapped token that is actually an EVM token". Now, what I am missing here is  being more specific that this is a Centrifuge **Connectors**-wrapped token.

We could address that by making the `interior` an `X3 ( PalletInstance( ConnectorsPalltetIndex), GlobalConsensus(NetworkId::Ethereum { chain_id }), AccountKey20  { .. }`. This would mean that conversion `ConnectorsWrappedToken` to `MultiLocation` becomes runtime-dependant, but that would be fine.

Thoughts?

libs/types/src/tokens.rs
```diff
@@ -255,6 +259,64 @@ impl CrossChainTransferability {
...
+///
+/// Currently, Connectors are only deployed on EVM-based chains and therefore
+/// the only supported type of connectors-wrapped tokens are Evm tokens. In the
+/// far future, we might support wrapped tokens from other chains, such as
+/// Cosmos, Avalanche, etc.
+#[derive(
+	Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen,
+)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum ConnectorsWrappedToken {
```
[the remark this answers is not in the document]
wischli: If we follow this idea, we have to rework the undirected mapping for `u128 <--> CurencyId`. Right now, we use the 4 bytes from the `ForeignAsset(u32)` identifier and add a 12-byte prefix. 

In order to guarantee collision resistance, we would have to use both the `u64` `EVMChainId` as well as the 20 bytes from the token address plus preferably a sufficiently long prefix. We would end up with 224 bytes only from those two fields already.

IMO, this is enough reason to not register such currencies as `CurrencyId::ConnectorsWrapped(ConnectorsWrappedToken)` but I might be missing something here.

libs/types/src/tokens.rs
```diff
@@ -255,6 +259,64 @@ impl CrossChainTransferability {
...
+
+impl TryFrom<MultiLocation> for ConnectorsWrappedToken {
+	type Error = ();
+
+	fn try_from(location: MultiLocation) -> Result<Self, Self::Error> {
+		match location {
+			MultiLocation {
+				parents: 0,
+				interior:
+					X2(
```
wischli, line 290: Do we not need to support derivation for Moonbeam?

wischli (commented):
Overall LGTM. Added my two cents on the asset registry question and I feel like we are missing at least a Moonbeam derivation?

libs/types/src/tokens.rs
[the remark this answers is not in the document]
NunoAlexandre: Yeah that's right, William. I am not attached to this idea, I was just throwing it out there. Maybe it's better to just keep things as they are 👍

libs/types/src/tokens.rs
NunoAlexandre: How do you mean, @wischli? Moonbeam, from this perspective, would be seen as an EVM chain so that's already encapsulated in the `GlobalConsensus(NetworkId::Ethereum { chain_id }` part. See https://github.com/centrifuge/centrifuge-chain/pull/1393#discussion_r1224080128 for more context and info 👍

wischli: Sorry, I missed out on that comment as it is not displayed in the `Files changed` section since the line it refers to is outdated.

libs/types/src/tokens.rs
wischli: > This would mean that conversion ConnectorsWrappedToken to MultiLocation becomes runtime-dependant, but that would be fine.

IMO, that is totally fine and I favour this variant over the `X2` one.

NunoAlexandre: Yes, me too. I started with X2 to kick off this discussion with something to look at already and because I recall @mustermeiszer not being into the idea of including the connectors pallet index here when I suggest this a while back 🤔

NunoAlexandre: @wischli @mustermeiszer I have implemented the alternative containing the X3 with the Connectors pallet index. I have done so following a similar pattern used with the `CurrencyIdConvert` type. This is necessary given that the conversion to make it generic on the specific runtime it will be used while being runtime-dependant, and since adding the pallet index to the `ConnectorsWrappedToken` type didn't make sense.

libs/types/src/tokens.rs
```diff
@@ -255,6 +255,25 @@ impl CrossChainTransferability {
 	}
 }
 
+/// Connectors-wrapped tokens
+///
+/// Currently, Connectors are only deployed on EVM-based chains and therefore
+/// we only support EVM tokens. In the far future, we might support wrapped
+/// tokens from other chains, such as Cosmos, Avalanche, etc.
```
mustermeiszer, line 262:
```suggestion
/// tokens from other chains, such as Cosmos based ones.
```

Avalanche is EVM based.

runtime/common/src/lib.rs
```diff
@@ -453,3 +453,152 @@ pub mod changes {
...
+	impl<Index: Get<PalletIndex>> Convert<MultiLocation, Result<ConnectorsWrappedToken, ()>>
+		for ConnectorsWrappedTokenConvert<Index>
+	{
+		fn convert(location: MultiLocation) -> Result<ConnectorsWrappedToken, ()> {
+			match location {
+				MultiLocation {
+					parents: 0,
+					interior:
+						X3(
+							PalletInstance(pallet_instance),
```
mustermeiszer, line 484: Why do we want the `PalletInstance` here again?

mustermeiszer (commented):
Looks good. Just wanted to highlight that we could imlement this purely on the connectors level without moving to the runtime level if we drop the pallet instance. But no blocker.

libs/types/src/tokens.rs
mustermeiszer: My two cents are, that we never leak this `ConnectorsWrapedToken` to the outside, right? It is only used as we need a way to verify the arguements of the users, so making this runtime dependent seems to be leaking a type that has no use to the outside. But maybe I am missing something.

NunoAlexandre: There's for sure validity in that argument. But I do find that being specific in the structure of a MultiLocation is beneficial because it makes it more future-proof; e.g. imagine we would have some new feature in the future involving wrapped tokens with a similar nature. Not being able to distinguish between them looking at the MultiLocation is a drawback if not a problem, it would require migrations, etc. So if we can compromise on the current solution I'd vote to keep it 👍

libs/types/src/tokens.rs
NunoAlexandre: Thanks 👍

PR #1395, opened 2023-06-09
fix: Address EVM-related import warnings

NunoAlexandre:
### Description
Currently building the project throws the following warnings:

Screenshot 2023-06-09 at 13 39 18

### Changes

Move the appropriate imports to the place of their usage, under the feature flag where they are used.


runtime/altair/src/lib.rs
```diff
@@ -1676,6 +1671,10 @@ impl fp_self_contained::SelfContainedCall for RuntimeCall {
 
 	#[cfg(not(feature = "testnet-runtime"))]
 	fn check_self_contained(&self) -> Option<Result<Self::SignedInfo, TransactionValidityError>> {
+		use pallet_ethereum::TransactionAction;
+		use runtime_common::evm::GetTransactionAction;
+		use sp_runtime::transaction_validity::InvalidTransaction;
+
```
NunoAlexandre, line 1677: It seems that `cargo` doesn't handle `#[cfg(not` rules sensibly when it comes to imports, because the warnings we get are not actually an issue 🤷‍♂️  but by moving the imports here everyone is happy

lemunozm (approved):
You fixed in the simplest way 👍🏻

wischli (approved):
Thanks for cleaning up!

PR #1396, opened 2023-06-12
ChangeGuard pool-system implementation

### LINKED ISSUE #1385: Loans: `ChangeGuard` trait implementation
### Description

Once https://github.com/centrifuge/centrifuge-chain/pull/1384 is merged, we must implement the behavior of `ChangeGuard`. 

Pending questions:
- Could be in its own pallet or into pool-system, see this [thread](https://kflabs.slack.com/archives/C050E3ZN2JU/p1685452550780189). 
- Should `CfgChange` also have a "severity level" indicator, to abstract pallet-dependent data?

lemunozm:
# Description

Fixes #1385 

Check the related specs here: https://centrifuge.hackmd.io/_lquEXx8Q_69mQE0HRGHSw?view

### Changes
- [x] `ChangeGuard` implemented into `pool-system`.
- [x] Linked `pallet-loans` changes with `pallet-pool-system` changes.
- [x] Generated Event with `ChangeId` in `pallet-pool-system`
- [x] Convert `LoanMutation` into `PoolChangeProposal` based on this [sheet](https://docs.google.com/spreadsheets/d/1RJ5RLobAdumXUK7k_ugxy2eDAwI5akvtuqUM2Tyn5ts/edit?pli=1#gid=0)
- [x] Tests
- [x] Process `PoolChangeProposal` for each requirement.
  - [x] non-submitted period (default requirement)
  - [x] DelayTime
  - [x] NextEpoch
  - [x] ~~BlockByLockedRedemptions.~~ Another issue: https://github.com/centrifuge/centrifuge-chain/issues/1407


pallets/pool-system/src/lib.rs
```diff
@@ -260,6 +261,8 @@ pub mod pallet {
 
 		type CurrencyId: Parameter + Copy + MaxEncodedLen;
 
+		type ExtChange: Parameter + Member + MaxEncodedLen + TypeInfo + Into<ExtPoolChange>;
```
offerijns, line 264: Why `ExtChange`? I think we should use this for both `loan` and `pool` changes, right, so then only some of them are external? 

We should also add a TODO (can be done later of course) to migrate https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/pool-system/src/impls.rs#L199 to use this `ChangeGuard` implementation

pallets/pool-system/src/pool_types.rs
```diff
@@ -332,3 +332,10 @@ impl<
 		})
 	}
 }
+
+#[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, Default, TypeInfo, MaxEncodedLen)]
+pub struct ExtPoolChange {
+	submitted_time: Moment,
+	delay_time_required: Moment,
+	block_by_locked_redemptions: bool,
+}
```
offerijns, line 341: Maybe would be more extensible if this was
```suggestion

pub enum ExtPoolChangeRequirements {
  DelayTime(Moment),
  BlockedByLockedRedemptions,
}

#[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, Default, TypeInfo, MaxEncodedLen)]
pub struct ExtPoolChange {
	submitted_time: Moment,
	requirements: BoundedBTreeSet<ExtPoolChangeRequirements, RequirementsSize>,
}
```

pallets/pool-system/src/lib.rs
lemunozm: Make sense!

pallets/pool-system/src/pool_types.rs
lemunozm: I still do not know well what the requirements will be, but I like this very much!

Do you know another future requirement as an example?

pallets/pool-system/src/lib.rs
lemunozm: Issue to track the migration: https://github.com/centrifuge/centrifuge-chain/issues/1397

pallets/pool-system/src/impls.rs
```diff
@@ -367,6 +372,60 @@ impl<T: Config> InvestmentAccountant<T::AccountId> for Pallet<T> {
...
+			Changes::<T>::get(pool_id, change_id).ok_or(Error::<T>::ChangeNotFound)?;
+
+		let pool_change: PoolChangeProposal = change.clone().into();
+
+		let mut allowed = true;
+		for requirement in pool_change.requirements() {
+			allowed &= match requirement {
+				Requirement::NextEpoch => {
+					match Pool::<T>::get(pool_id) {
+						Some(details) => submitted_time < details.epoch.last_closed,
```
lemunozm, line 408: Not sure if we should use `last_closed` or `last_executed` here

pallets/pool-system/src/impls.rs
```diff
@@ -367,6 +372,60 @@ impl<T: Config> InvestmentAccountant<T::AccountId> for Pallet<T> {
...
+						None => true, // If the pool doesn't exists, there are no restrictions.
+					}
+				}
+				Requirement::DelayTime(secs) => {
+					Self::now().saturating_sub(submitted_time) >= secs as u64
+				}
+				// TODO
+				Requirement::BlockedByLockedRedemptions => false,
```
lemunozm, line 416: I do not know how to know if this restriction is valid. Could it mean the we are blocked in the following interval?:
```
last_closed < submitted_time < last_executed
```

lemunozm (commented):
Some important questions before merging this PR 

cc @offerijns @mustermeiszer

pallets/pool-system/src/impls.rs
offerijns: See here how this restriction is implemented: https://github.com/centrifuge/centrifuge-chain/blob/main/runtime/centrifuge/src/lib.rs#L1323

offerijns: Actually we should also have that "do not allow releasing updates during epoch closing" check for all possible changes, both pool and loan.

lemunozm: > Actually we should also have that "do not allow releasing updates during epoch closing" check for all possible changes, both pool and loan.

What do you mean by this? to have the `BlockedByLockedRedemptions` as a default implicit requirement for all changes?

pallets/pool-system/src/impls.rs
mustermeiszer: I would opt for `last_executed` as we can have cash flows in between. Although, those should not change the the execution/solution of previous. But I think this will change when we change the rebalancing.

pallets/pool-system/src/impls.rs
```diff
@@ -367,6 +372,60 @@ impl<T: Config> InvestmentAccountant<T::AccountId> for Pallet<T> {
...
+				Requirement::NextEpoch => {
+					match Pool::<T>::get(pool_id) {
+						Some(details) => submitted_time < details.epoch.last_closed,
+						None => true, // If the pool doesn't exists, there are no restrictions.
+					}
+				}
+				Requirement::DelayTime(secs) => {
+					Self::now().saturating_sub(submitted_time) >= secs as u64
+				}
+				// TODO
```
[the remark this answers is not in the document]
mustermeiszer: Or better IMO. 

* pallet-investments stores some info about redemptions in the pallet-system. As opaque as the `RuntimeChange` for the pallet-pool-system. 
* But it allows to query whether there are redemtions outstainding

This would allow us to mitigate the need to call into pallet-investments and iterate over all tranches and possible outstanding redemptions. But just a first thought.

pallets/pool-system/src/impls.rs
```diff
@@ -367,6 +372,60 @@ impl<T: Config> InvestmentAccountant<T::AccountId> for Pallet<T> {
...
+
+		let pool_change: PoolChangeProposal = change.clone().into();
+
+		let mut allowed = true;
+		for requirement in pool_change.requirements() {
+			allowed &= match requirement {
+				Requirement::NextEpoch => {
+					match Pool::<T>::get(pool_id) {
+						Some(details) => submitted_time < details.epoch.last_closed,
+						None => true, // If the pool doesn't exists, there are no restrictions.
```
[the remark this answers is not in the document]
lemunozm: I thought at first to do it in that way. But then I realized that if we remove the pool, then we'll never remove that storage. Unless the removing pool action (maybe in the future) also removes its related changes.

If the case is the second one, then I think we can go with an error here

pallets/pool-system/src/impls.rs
```diff
@@ -367,6 +372,60 @@ impl<T: Config> InvestmentAccountant<T::AccountId> for Pallet<T> {
...
 }
 
+impl<T: Config> ChangeGuard for Pallet<T> {
+	type Change = T::RuntimeChange;
+	type ChangeId = T::Hash;
+	type PoolId = T::PoolId;
+
+	fn note(pool_id: Self::PoolId, change: Self::Change) -> Result<Self::ChangeId, DispatchError> {
+		let registered_change = (Self::now(), change.clone());
+		let change_id: Self::ChangeId = T::Hashing::hash(&registered_change.encode());
```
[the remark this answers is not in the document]
lemunozm: Yes (confirmed previously with @offerijns), because the hash also takes into account the change content. So, if in the same block, we have the exact change, it will count as only one change. Make sense for all use cases that came to my mind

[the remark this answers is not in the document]
offerijns: Would only be the case if there are multiple updates that have exactly the same content (same change it's applying), right? So I think thats' fine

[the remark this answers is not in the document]
offerijns: Since the contents of the change are included in the hash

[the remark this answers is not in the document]
mustermeiszer: Ahh yeah.

pallets/pool-system/src/impls.rs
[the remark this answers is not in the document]
mustermeiszer: But the loans does not check existance of a pool in every call, so I would say, it is something weird if loans go on and a pool is not there..

pallets/pool-system/src/impls.rs
lemunozm: Ok, so we have a kind of default requirement for all possible changes

pallets/pool-system/src/impls.rs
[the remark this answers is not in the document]
lemunozm: Mmm... 🤔 Ok. 

> pallet-investments stores some info about redemptions in the pallet-system

How is it stored in the `pallet-system`? Is there a way to store things there?

Not sure if this is what you mean, but maybe `pallet-investments` can create a change (a `RuntimeChange`) as a way to send information to the `pallet-pool-system` in a transparent way. Although maybe it could be even simple to offer a trait from `pallet-pool-system` that `pallet-investment` can call to update the redemption state in the pool side

pallets/pool-system/src/impls.rs
lemunozm: Ok, `last_executed` makes more sense!

lemunozm: Looking into it, I thought the `last_executed` was the `Moment`, but it's an `EpochId` so I can not compare it with the `submitted_time` I'm already handle. Solutions:
1. Keep it by now as it is, using the `last_closed`.
2. Store the `EpochId` when the event was noted, to check it later against `last_executed`.

lemunozm: I think we can go with 1., if this is ensured for all changes: https://github.com/centrifuge/centrifuge-chain/pull/1396#discussion_r1229242865

pallets/pool-system/src/impls.rs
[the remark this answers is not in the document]
lemunozm: Done [701770d](https://github.com/centrifuge/centrifuge-chain/pull/1396/commits/701770dcaee6a0ac9164913c95c5c3e647cde3df)

pallets/pool-system/src/impls.rs
offerijns: I think `last_closed` is fine actually. This requirement just needs to ensure that at least the epoch is inbetween noting the change and releasing it. This is true once `last_closed > submitted_time`. The time inbetween closing and execution does not matter if changes cannot be executed during the submission period anyways, which is the other default requirement we are adding.

pallets/pool-system/src/impls.rs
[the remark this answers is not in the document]
offerijns: I think it should check not if there are redemptions outstanding right now, but rather if all redemptions were fulfilled in the last epoch. Otherwise someone can block all changes indefinitely by continuously submitting a 1 wei redemption order in every block.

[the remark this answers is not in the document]
offerijns: Which was actually incorrectly implemented previously as well

pallets/pool-system/src/impls.rs
offerijns: Yes exactly. Should never allow releasing an update if the pool is in a submission period

pallets/pool-system/src/impls.rs
lemunozm: Yeah, composing this rule with the default rule we are safe 👍🏻

pallets/pool-system/src/impls.rs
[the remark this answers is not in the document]
lemunozm: Ok, I need to dive first into this part that is new for me, but I get the idea, thanks 👍🏻

pallets/pool-system/src/impls.rs
lemunozm: Done 👍🏻

offerijns: Looks good to me! Nice implementation :)

pallets/pool-system/src/impls.rs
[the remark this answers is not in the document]
lemunozm: > I think it should check not if there are redemptions outstanding right now, but rather if all redemptions were fulfilled in the last epoch.

But the change we are processing at that moment will affect the current redemptions, isn't it? So are the current redemptions which should be protected against the change.

I see the issue you commented on, but I'm not sure how it can fix it...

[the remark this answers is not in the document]
lemunozm: > the last executed epoch had a redemption fulfillment of 100% in all tranches

But this information isn't anywhere right now, right? I guess we need to track this somehow to be able to check it later in the `released()` method.

pallets/pool-system/src/pool_types.rs
```diff
@@ -332,3 +332,106 @@ impl<
...
+	}
+
+	impl PoolChangeProposal {
+		pub fn new(requirements: impl IntoIterator<Item = Requirement>) -> Self {
+			Self {
+                requirements: BTreeSet::from_iter(requirements.into_iter().map(UniqueRequirement))
+                    .try_into()
+                    .expect(
+                        "Cannot exist more unique requirements in a set than `MaxRequirements`, qed",
+                    ),
```
mustermeiszer, line 420: Given that we are passing a `Vec<Requierment>` to the `PoolChangeProposal::new()` in the runtime-common implementation, what would be the behaviour if multiple `Requirements` are passed just with different "inner" values. Does the BTreeSet deterministically choose the first one or the last one?

mustermeiszer (approved):
Just a minor question regarding the behaviour of BTreeSet inserts for equal values. Otherwise, nice!!

pallets/pool-system/src/pool_types.rs
lemunozm: That is a very good question! After checking it, the first one is the chosen.

PR #1400, opened 2023-06-13
Fix local environment

NunoAlexandre:
I was trying to spin up a local environment and run into two issues:

- Docker error [`"no matching manifest for linux/arm64/v8 in the manifest list entries"`](https://github.com/OpenLiberty/ci.docker/issues/332)
Fixed by specifying the platform in the yaml file. I wonder if this cause issues for devs in other platforms?

- The development genesis tries to register AUSD twice which caused the dev note to fail to load all together
Fixed by dropping the duplicate entry.


lemunozm (dismissed):
LGTM!

wischli (commented):
Thanks for shortening my future docker commands! Can you point me to the other location where we register `AUSD`?

src/chain_spec.rs
```diff
@@ -1264,31 +1264,5 @@ fn asset_registry_assets() -> Vec<(CurrencyId, Vec<u8>)> {
 			}
 			.encode(),
 		),
-		(
```
NunoAlexandre, line 1267: @wischli the block above is registering AUSD, this deleted block was a duplicate. I don't know how we merged this in 🤔

wischli (dismissed):
LGTM thanks for 🧹

src/chain_spec.rs
NunoAlexandre: ahah no worries! glad it's clear!

mikiquantum (commented):
I noticed that the `./res/rococo-local.json` is still on `0.9.37` but we are in `0.9.38`. The local setup works bc, even though we are pulling the `0.9.38` the spec is the 0.9.37 and with that runtime. 
Can you regenerate the `./res/rococo-local.json` using the `parity/polkadot:v0.9.38` so we test with the correspondent runtime?

NunoAlexandre:
> I noticed that the `./res/rococo-local.json` is still on `0.9.37` but we are in `0.9.38`. The local setup works bc, even though we are pulling the `0.9.38` the spec is the 0.9.37 and with that runtime. Can you regenerate the `./res/rococo-local.json` using the `parity/polkadot:v0.9.38` so we test with the correspondent runtime?

Thanks for the note and the context, @mikiquantum 👍 

Just pushed the new `rococo-local`, the output of the following command:

``` bash
λ export DOCKER_DEFAULT_PLATFORM=linux/amd64
λ docker run parity/polkadot:v0.9.38 build-spec --chain rococo-local --raw --disable-default-bootnode > res/rococo-local.json
```

I got these _interesting_ logs tho:

```
<jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
<jemalloc>: (This is the expected behaviour if you are running under QEMU)
2023-06-13 18:41:31 Building chain spec
2023-06-13 18:41:31 Took active validators from set with wrong size
2023-06-13 18:41:31 Took active validators from set with wrong size
2023-06-13 18:41:31 Took active validators from set with wrong size.
2023-06-13 18:41:31 Took active validators from set with wrong size
```

I can take a look again tomorrow morning _if_ something is actually wrong.

wischli:
> I can take a look again tomorrow morning if something is actually wrong

AFAIK these logs are normal for rococo local and can be ignored.

NunoAlexandre:
> > I can take a look again tomorrow morning if something is actually wrong
> 
> AFAIK these logs are normal for rococo local and can be ignored.

Yeah it appears to be the case 👍

mikiquantum:
> > I noticed that the `./res/rococo-local.json` is still on `0.9.37` but we are in `0.9.38`. The local setup works bc, even though we are pulling the `0.9.38` the spec is the 0.9.37 and with that runtime. Can you regenerate the `./res/rococo-local.json` using the `parity/polkadot:v0.9.38` so we test with the correspondent runtime?
> 
> Thanks for the note and the context, @mikiquantum 👍
> 
> Just pushed the new `rococo-local`, the output of the following command:
> 
> ```shell
> λ export DOCKER_DEFAULT_PLATFORM=linux/amd64
> λ docker run parity/polkadot:v0.9.38 build-spec --chain rococo-local --raw --disable-default-bootnode > res/rococo-local.json
> ```
> 
> I got these _interesting_ logs tho:
> 
> ```
> <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
> <jemalloc>: (This is the expected behaviour if you are running under QEMU)
> 2023-06-13 18:41:31 Building chain spec
> 2023-06-13 18:41:31 Took active validators from set with wrong size
> 2023-06-13 18:41:31 Took active validators from set with wrong size
> 2023-06-13 18:41:31 Took active validators from set with wrong size.
> 2023-06-13 18:41:31 Took active validators from set with wrong size
> ```
> 
> I can take a look again tomorrow morning _if_ something is actually wrong.

Yep those are normal 👍

mikiquantum (approved):
Thanks!

PR #1403, opened 2023-06-15
pallets: Add Ethereum Transaction pallet

cdamian:
Follow-up for the connectors gateway pallet discussion around EVM found [here](https://github.com/centrifuge/centrifuge-chain/pull/1376#discussion_r1210457853).

The approach is similar to the one found in [ethereum-xcm](https://github.com/PureStake/moonbeam/blob/master/pallets/ethereum-xcm/src/lib.rs#L89).

Note this a PR against - https://github.com/centrifuge/centrifuge-chain/pull/1376, meant to facilitate discussion around this specific topic.


pallets/ethereum-transaction/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+					},
+					error: e.error,
+				}
+			})?;
+
+			match info {
+				CallOrCreateInfo::Call(call_info) => {
+					Nonce::<T>::put(nonce.saturating_add(U256::one()));
+
+					//TODO(cdamian): calculate this without base weight?
```
cdamian, line 108: Am I calculating this the right way?

cdamian:
> > The approach is similar to the one found in [ethereum-xcm](https://github.com/PureStake/moonbeam/blob/master/pallets/ethereum-xcm/src/lib.rs#L89).
> 
> Where can we find information about the thought process around implementing a custom one instead of using that already solution from PureStake?

Here is the [spec](https://centrifuge.hackmd.io/yd0ak2uKQYy92I4bR8ViDA).

NunoAlexandre:
> 

Cool, thank you!

> In order for Ethereum calls from the Substrate side to be present in an ethereum block, a full fake transaction needs to be created

What is the goal behind this feature? Is it so that we have all the substrate calls / blocks accessible in the EVM? And won't it be an issue that all of the mirrored blocks on the EVM are all signed by that mock account?

pallets/ethereum-transaction/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+
+	impl<T: Config> EthereumTransactor for Pallet<T> {
+		fn call(
+			from: H160,
+			to: H160,
+			data: &[u8],
+			value: U256,
+			gas_price: U256,
+			gas_limit: U256,
+		) -> DispatchResultWithPostInfo {
```
cdamian, line 54: Do we want to propagate this and then charge for the fees in one of the callers of this OR do we want to charge all the necessary fees here, in this pallet?

cdamian: Upon further testing I noticed that some fees are actually charged by the [EVM runner](https://github.com/paritytech/frontier/blob/polkadot-v0.9.38/frame/evm/src/runner/stack.rs#L187).

cdamian: In this case, the above mentioned fees would be charged from the derived account, meaning that we'll have to ensure that the original caller will have sufficient funds to execute the EVM call.

mustermeiszer: Thanks for the research. What would happen if the evm fails? Would it roll-back the complete transaction including possible EVM changes?

If this is the case, I would just assume we fail in runtime then.

pallets/ethereum-transaction/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+			let nonce = Nonce::<T>::get();
+			let read_weight = T::DbWeight::get().reads(1);
+
+			//TODO(cdamian): Same signature as the one in ethereum-xcm.
+			let signature = TransactionSignature::new(
+				42,
```
mustermeiszer, line 55: ^^

pallets/ethereum-transaction/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+					Err(DispatchErrorWithPostInfo {
+						post_info: PostDispatchInfo {
+							actual_weight: Some(weight.saturating_add(read_weight)),
+							pays_fee: Pays::Yes,
+						},
+						error: DispatchError::Other("unexpected execute result"),
+					})
+				}
+			}
+		}
```
mustermeiszer, line 131: I think we should just remove this one mathing arm or panic and have a test case for that. WDYT?

mustermeiszer (approved):
The only think I would clear out is whether we can just fail in runtime/EVM runner if the calley has not enough funds. cc @branan

pallets/ethereum-transaction/src/lib.rs
cdamian: I wanted to avoid panics and be safe here, not sure if I can reproduce this kind of error when testing. Especially since the tests that I currently have are passing, without having a contract deployed in there.

mustermeiszer: Alrigh! Seems reasonable

pallets/ethereum-transaction/src/lib.rs
cdamian: I will have to check that.

cdamian: > If there is an error before [this point](https://github.com/paritytech/frontier/blob/5644279da588baab7caa9f3a4eb69363e1606461/frame/evm/src/runner/stack.rs#L187) we will get back a `RunnerError` that contains a weight that we should charge. Otherwise, the total fee for the EVM transaction will be charged by the runner and we would only have to charge for reading the nonce.

Given this, would it make sense to charge for reading the nonce in this pallet? This would the allow us to just use a `DispatchResult` here and keep other parts free of fee-related logic. WDYT @mustermeiszer

cdamian:
As agreed, gonna merge this for now and continue working on the original branch.

PR #1406, opened 2023-06-19
Move node crate to its own folder

### LINKED ISSUE #1278: `node` crate under its own folder
The `Cargo.toml` of the node crate is mixed with the `Cargo.toml` of the project. And sometimes it is difficult to see what belongs to what.

Following the updated [substrate repo template](https://github.com/substrate-developer-hub/substrate-node-template), the *convention* is to move the node to its own `node` folder.

lemunozm:
# Description

Fixes #1278

- Updated `connectors` rust edition to `2021` @NunoAlexandre


NunoAlexandre:
> The compilation issue is because of this: [mcginty/snow#146](https://github.com/mcginty/snow/issues/146) I've tried to solve it without success. Not sure if any of you have found this before @NunoAlexandre @wischli
> 
> Note that the `Cargo.lock` is not updated in this PR, so all dependencies should work as they are.

@lemunozm I would try two things:

1. Run the nix command to make sure we don't have any messed up duplicated dependencies.
  I read the note about cargo.lock not changing but still I would double check this

2. Try `cargo update`

NunoAlexandre:
> Thanks @NunoAlexandre, `cargo update` worked! But I'm still wondering why simply moving files adds a dependency error even though the `Cargo.lock` is not modified. Pretty weird!

Nice. yeah that is weird indeed. Sometimes cargo deals with dependencies differently when they are set in the root `cargo.toml` file, that could have something to do with it

NunoAlexandre (dismissed):
Thanks for doing this :) it was in the back of my mind for over a year.

mustermeiszer (commented):
Looks mostly good. 

Two things:

* Pipeline is failing. Space in runner I guess and somehow we are missing an impl, I guess this is a feature flag issue, now that the node was moved to a non top level crate
* Why the `cargo update`?

NunoAlexandre:
> I'll recheck the features! Thanks for the suggestion
> 
> Regarding why the `cargo update`... this is the big question 😆

You can try and run this `subalfred` at the root of the repo (credit to @wischli)

``` bash
#!/usr/bin/env bash

# Find all child directories containing Cargo.toml files
dirs=$(find . -name Cargo.toml -print0 | xargs -0 -n1 dirname | sort -u)

# Execute the command "subalfred check" on each directory
for dir in $dirs; do
    echo $dir;
    subalfred check features $dir
done
```

NunoAlexandre:
> `subalfred` says that everything is fine 😓

I can take a look later today / tomorrow morning, I have experienced s.o. m.a.n.y. cargo issues updating our chain to newer versions of polkadot in the last 1.5 years that I don't feel the pain anymore 😆

lemunozm:
Hahaha, I appreciate it a lot @NunoAlexandre. Hope your experienced eye finds something

wischli:
Couldn't find anything obvious as well comparing it with another parachain. So I agree it might be a feature flag issue. Please note that subalfred only checks `std`, `runtime-benchmarks` and `try-runtime`.

lemunozm:
Thanks @wischli for taking the time to look into it. I didn't know about `subalfred` only check those three features.

wischli:
> Thanks @wischli for taking the time to look into it. I didn't know about `subalfred` only check those three features.

I wasn't fully sure if this is correct as I remembered it from looking through the Subalfred repo but the documentation states the same unfortunately: https://subalfred.hack.ink/user/cli/check.html#episode-2

lemunozm:
Thanks a lot @NunoAlexandre for your research on this!! 🙌🏻  I'll give it another try with more time.

lemunozm:
I still don't know what the bug is, but at least I know why it's falling now and not before.

Any command with the form `cargo check` in the top of the project folder was running the node crate before, it was translated to `cargo check -p centrifuge-chain`. Now if we do `cargo check` on the top folder, we are doing what previously we did as `cargo check --workspace`. So, the jobs are working in a different way, and we are finding a hidden issue we already had.

By now I think we can get this working by modifying the jobs to run the node instead of all workspace.

lemunozm:
A little digging:
- `cargo check` ✅
- `cargo check -r` ✅ 
- `cargo check -F runtime-benchmarks` ❌ 
- `cargo check -r -F runtime-benchmarks` ❌ 
- `cargo check -p centrifuge-chain` ✅ 
- `cargo check -r -p centrifuge-chain` ✅ 
- `cargo check -p centrifuge-chain -F runtime-benchmarks` ✅ 
- `cargo check -r -p centrifuge-chain -F runtime-benchmarks` ✅ 

The issue comes from compiling the workspace (not the node) with `runtime-benchmarks` enabled. Release mode does not take effect.

lemunozm:
And surprisingly, the issue is fixed if we add try-runtime along with `runtime-benchmarks`:
- `cargo check` ✅
- `cargo check -F runtime-benchmarks` ❌
- `cargo check -F runtime-benchmarks,try-runtime` ✅

lemunozm:
I think all issues are solved. CI is now failing because we have reached the max size limit

lemunozm:
No idea here why jobs are failing because of storage limits. Changes I did should not add extra storage, or at least I'm not able to see why

lemunozm:
I do not now what have been changed in our CI config but it seems like this finally passes the CI jobs 😭

lemunozm:
The last CI errors were about space/memory. All "code" issues were already fixed before. 

Maybe the doc CI job that was fixed generated a lot of artifacts in the docker, and made other CI jobs fail because of the available space. 🤔

lemunozm:
I'm going to fix the conflict and maybe we can this finally merged 🎉

lemunozm:
Closed in favor of: https://github.com/centrifuge/centrifuge-chain/pull/1632

PR #1412, opened 2023-06-21
Loans: write off policy though ChangeGuard

### LINKED ISSUE #1386: Loans: write off policy should go through `ChangeGuard`
### Description

We should protect policy changes for the investor through `ChangeGuard`. This would imply adding a new variant to `pallet_loans::Change`: 

```rust
enum Change {
    Loan(LoanId, LoanMutation), // Check permissions for this
    WriteOffPolicy(Vec<WriteOffRule>), //Check another permissions for this
}
```

And we would need to remove the current extrinsic and add a new couple of them:
- `propose_write_off_policy()` 
- `apply_write_off_policy()`

lemunozm:
# Description

Fixes #1386 

## Changes and Descriptions

- [x] Updated/Added two new extrinsics:
  - `propose_write_off_policy()`
  - `apply_write_off_policy()`
- [x] `pallet loans` tests
- [x] `pallet-loans` benchmarks
- [ ] Runtime common changes to adapt the new loan's change to Requirements.
- [ ] Benchmarks running correctly from for real runtimes. 
  - We would need some `pallet-pool-system` benchmark utility to enable this through the `PoolBenchmark` trait.

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -15,30 +15,30 @@
 
 //! This pallet offers extrinsics to handle loans.
 //!
-//! The following actions are performed over loans:
-//!
-//! | Extrinsics                    | Role      |
-//! |-------------------------------|-----------|
-//! | [`Pallet::create()`]          | Borrower  |
-//! | [`Pallet::borrow()`]          | Borrower  |
-//! | [`Pallet::repay()`]           | Borrower  |
-//! | [`Pallet::write_off()`]       |           |
-//! | [`Pallet::admin_write_off()`] | LoanAdmin |
-//! | [`Pallet::close()`]           | Borrower  |
-//!
-//! The following actions are performed over a pool of loans:
-//!
-//! | Extrinsics                               | Role      |
-//! |------------------------------------------|-----------|
-//! | [`Pallet::update_write_off_policy()`]    | PoolAdmin |
-//! | [`Pallet::update_portfolio_valuation()`] | Any       |
-//!
-//! The following actions are performed based on changes:
+//! The following actions are performed over a loan:
 //!
 //! | Extrinsics                          | Role      |
 //! |-------------------------------------|-----------|
+//! | [`Pallet::create()`]                | Borrower  |
+//! | [`Pallet::borrow()`]                | Borrower  |
+//! | [`Pallet::repay()`]                 | Borrower  |
+//! | [`Pallet::write_off()`]             |           |
+//! | [`Pallet::admin_write_off()`]       | LoanAdmin |
+//! | [`Pallet::close()`]                 | Borrower  |
 //! | [`Pallet::propose_loan_mutation()`] | LoanAdmin |
-//! | [`Pallet::apply_change()`]          |           |
+//!
+//! The following actions are performed over an entire pool of loans:
+//!
+//! | Extrinsics                               | Role      |
+//! |------------------------------------------|-----------|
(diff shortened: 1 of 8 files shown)


pallets/loans/src/lib.rs
```diff
@@ -779,6 +780,29 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Apply a proposed change identified by a change id.
+		/// It will only perform the change if the requirements for it
+		/// are fulfilled.
+		#[pallet::weight(100_000_000)]
+		#[pallet::call_index(10)]
+		pub fn apply_write_off_policy(
```
lemunozm, line 789: I've added this as a new extrinsic. The signature is the same for both `apply_loan_mutation()` and `apply_write_off_policy()` but having them in separated extrinsic allow to calculate correctly the weights for each case. Otherwise we would need to run the extrinsics and fix the weight which is not desirable for a user to know how many fees will need.

offerijns: Makes sense!

pallets/loans/src/lib.rs
```diff
@@ -646,6 +645,63 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// Propose a change.
+		/// The change is not performed until you call
+		/// [`Pallet::apply_change()`].
```
offerijns, line 650:
```suggestion
		/// [`Pallet::apply_loan_mutation()`].
```

pallets/loans/src/lib.rs
```diff
@@ -849,6 +871,15 @@ pub mod pallet {
 			policy::find_rule(rules, |trigger| loan.check_write_off_trigger(trigger))
 		}
 
+		fn get_change(
```
offerijns, line 874: I was wondering initially whether we were properly checking whether the changes were released. This function name doesn't make that clear. I would suggest making this more explicit, something like
```suggestion
		fn get_change_if_released(
```

or `get_released_change` or something else that's better that makes this clear

lemunozm: Ok, it's reasonable! Good suggestion

lemunozm:
@offerijns if it passes all jobs, this should be ready for the final review. Thanks!!

pallets/loans/src/lib.rs
```diff
@@ -646,6 +645,63 @@ pub mod pallet {
...
+
+			Self::deposit_event(Event::<T>::Mutated {
+				pool_id,
+				loan_id,
+				mutation,
+			});
+
+			Ok(())
+		}
+
```
lemunozm, line 704: This has not changed, I've only moved upwards in the file where they make more sense.

Not sure if we can also modify `call_index` of these calls to leave with the same order, or if it's too late. I think we could because by now nobody has used it. 
cc @NunoAlexandre @wischli

NunoAlexandre: @lemunozm considerations when changing pallets or call indexes:

1. Is there any governance proposal that encodes a call that will be changed?
2. Is there any dApp integration (our frontend included) that encodes a call to a pallet of any of our runtimes based on the pallet and call index?
3. Is there any data on storage that references a target pallet or call index?

If none apply, should be fine to change :) hope I didn't forget any edge case

lemunozm: Thanks Nuno!

AFAIK, none of them apply here 🚀

offerijns (approved):
LGTM!

PR #1417, opened 2023-06-27
Loans: Avoid check price id when creating loans

lemunozm:
# Description

From [slack thread](https://kflabs.slack.com/archives/C050E3ZN2JU/p1687798229887369)


offerijns (dismissed):
That was quick, as always :)

offerijns (approved):

PR #1420, opened 2023-06-27
Mock builder: Support nested mock calls

lemunozm:
# Description

Now you can define mocks inside mock executions. It unlocks the set-get mock pattern, meaning you can return what you've previously set in another mock. In other words, you will always get the last you've set.

```rust
MyMock::mock_set(|value| MyMock::mock_get(move || value));
```

## Changes and Descriptions
- Implementation
- Docs
- Tests


NunoAlexandre (approved):
Clean like water 💧 👌

lemunozm:
Thanks Nuno!

PR #1421, opened 2023-06-27
pool-system: Drop applied migration code

NunoAlexandre:
Triggered by these code changes and discussion https://github.com/centrifuge/centrifuge-chain/pull/1343#discussion_r1239672234, this migration seems old and that we can drop it.


lemunozm (approved):
Cool!

lemunozm (approved):

PR #1422, opened 2023-06-28
Rewards: Liquidity epoch duration as time instead of blocks

### LINKED ISSUE #1415: Rewards: liquidity epoch measured in time instead of block numbers.
### Description

Using a block number is difficult (impossible?) to predict what will be the moment an investor can claim the reward. Blocks can fail and be retried. Instead, if we use a time interval, the investor knows precisely when they can claim the reward.

lemunozm:
# Description

Fixes #1415

## Changes and Descriptions

- Changed `BlockNumber` intervals to `Moment` intervals.
- Split `on_initialize()` logic to be able to extract the epoch change logic and apply it in benchmarks (required now).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7469,6 +7469,7 @@ dependencies = [
  "frame-support",
  "frame-system",
  "num-traits",
+ "pallet-timestamp",
  "parity-scale-codec 3.4.0",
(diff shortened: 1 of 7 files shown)


mikiquantum (approved):
LGTM

PR #1425, opened 2023-06-29
Rewards: liquidity using gap & remove pallet-timestamp dependency

### LINKED ISSUE #1416: Rewards: Liquidity in development should use `gap` mechanism instead of `base`
### Description

Considering that altair & centrifuge will use the `gap` mechanism, we should change the liquidity mechanism in development to use `gap` instead.

lemunozm:
# Description

Fixes #1416 

## Changes and Descriptions

- Replace `base` by `gap` mechanism (which requires instantiating a new pallet)
- Remove `pallet-timestamp` dependency from `pallet-rewards` using a mock.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7469,7 +7469,6 @@ dependencies = [
  "frame-support",
  "frame-system",
  "num-traits",
- "pallet-timestamp",
  "parity-scale-codec 3.4.0",
  "scale-info",
  "sp-core",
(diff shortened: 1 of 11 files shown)


mustermeiszer:
Why do we need this?

lemunozm:
We want to configure big epochs in the liquidity system, and we want to protect from people that just stake in the moment of reward and later unstake them. `gap` mechanism protect from that

mikiquantum (approved):
LGTM!

PR #1426, opened 2023-06-29
altair: CrossChainTransferabilityMigration version check

NunoAlexandre:
Backport from https://github.com/centrifuge/centrifuge-chain/pull/1343, were @wischli reported the migration was failing when trying it out with `try-runtime`.

The issue being that the version check only allows the migration to run under Altair v1028, the version in which we expect the migration to actually be executed along with the respective the runtime upgrade. The issue is that it won't run until we bump the spec_version, making it impossible to test it without manually and temporarily bump Altair's `spec_version` to satisfy the check.


NunoAlexandre:
Closing since the original fix in the triggering PR has been merged already 👍

PR #1428, opened 2023-06-29
Loans: unlimit external pricing borrows

lemunozm:
# Description

Enable different max borrow models for external pricing

[slack thread](https://kflabs.slack.com/archives/C050E3ZN2JU/p1688037812005439)

### DIFF
--- a/pallets/loans/src/entities/pricing/external.rs
+++ b/pallets/loans/src/entities/pricing/external.rs
@@ -5,7 +5,7 @@ use cfg_traits::{
 };
 use cfg_types::adjustments::Adjustment;
 use codec::{Decode, Encode, MaxEncodedLen};
-use frame_support::{self, ensure, RuntimeDebugNoBound};
+use frame_support::{self, ensure, RuntimeDebug, RuntimeDebugNoBound};
 use scale_info::TypeInfo;
 use sp_runtime::{
 	traits::{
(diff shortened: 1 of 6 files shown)


pallets/loans/src/entities/pricing/external.rs
```diff
@@ -16,15 +16,25 @@ use sp_runtime::{
 
 use crate::pallet::{Config, Error, PoolIdOf, PriceOf};
 
+/// Define the max borrow amount of a loan
+#[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeDebug, MaxEncodedLen)]
+pub enum MaxBorrowAmount<Balance> {
```
offerijns, line 21: Maybe we should rename this to `ExternalMaxBorrowAmount` and vice verse for the internal variant, since otherwise the metadata will have two enums with the same name?

lemunozm: I think `TypeInfo` should somehow add the path information of where this type is, otherwise there would be a lot of collisions with any other pallet type in Substrate.

Let me check about this

lemunozm: Checked it uses `core::any::TypeId` which is the same one I've used for differentiate calls in `mock-builder` and it adds the path of the type. So both types are differents.

offerijns: Alright 👍 Thanks for checking

PR #1429, opened 2023-06-29
Add orderbook 

thea-leake:
# Description

Adds the orderbook pallet.

## Changes and Descriptions

Adds pallet providing an orderbook to exchange assets.
Adds pallet to dev runtime, and dev runtime weights.

# Checklist:

- [x] I have performed a self-review of my code -- in progress


pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+			);
+			ensure!(
+				T::TradeableAsset::can_reserve(asset_in, &account_id, amount),
+				Error::<T>::InsufficientAssetFunds,
+			);
+
+			ensure!(
+				T::ReserveCurrency::can_reserve(
+					&account_id,
+					T::Fees::fee_value(T::OrderFeeKey::get())
```
mustermeiszer, line 241: Could be a conflict if we allowed `trading_curreny` = `reserve_curreny`

mustermeiszer (commented):
Looks good! Definitely going in the right direction.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+		/// Provided for frontend to grab all open orders
+		pub fn get_all_orders() -> Result<
+			sp_std::vec::Vec<(
+				T::Hash,
+				Order<T::Hash, T::AccountId, T::AssetCurrencyId, T::ForeignCurrencyBalance>,
+			)>,
+			Error<T>,
+		> {
+			Ok(<Orders<T>>::iter().collect())
+		}
```
Chralt98, line 367: You might want to use this for a RPC to let it be queried by the front end easier, right? But can't you just access the storage items `Orders` and `UserOrders` directly instead of using an RPC? I always think if you have complex computations, it's cool to provide a RPC for it, but just querying the storage might not be a big deal for the front end. In addition: If you access the functions inside the runtime, I think you know that `iter...().collect()` is a bad practice and usually leads to overweight of blocks. But I assume the functions will never be called inside the runtime.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+			// UserOrders using Resultquery, if signed account
+			// does not match user for order id, we will get an Err Result
+			Self::cancel_order(order_id)?;
+			Ok(())
+		}
+
+		#[pallet::call_index(2)]
+		// dummy weight for now
+		#[pallet::weight(10_000 + T::DbWeight::get().reads_writes(2, 2).ref_time())]
+		pub fn fill_order_full(origin: OriginFor<T>, order_id: T::Hash) -> DispatchResult {
```
Chralt98, line 291: Are there plans to partially fill orders? I mean assume one big ask order comes in at the best price (lowest ask). Now a small fish can't really buy at this price, because the order is too big for the small fish.

The other case is the trickier: Assume there are many small ask orders at the best price (lowest ask). Now a big whale places a market order and buys all small ask orders. First, this often creates slippage, as you might know already.. Second, there might be not enough open ask orders to fill the big bid of the whale. Third, the whale needs to transact with many small fish accounts to fulfil the small orders. This leads to a high computational demand.

So is your plan to just allow the orders to be filled 100% to mitigate this and sacrifice that a small fish can fill the possible big order? 

There are two

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+		OrderNotFound,
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T>
+	where
+		<T as frame_system::Config>::Hash: PartialEq<<T as frame_system::Config>::Hash>,
+	{
+		/// Create an order, with the minimum fulfillment amount set to the buy
+		/// amount, as the first iteration will not have partial fulfillment
```
Chralt98, line 258: Oh I saw this after [this](https://github.com/centrifuge/centrifuge-chain/pull/1429/files#r1269004599) message. 

But what are your plans to realise not partial fulfillment, but the other way around: If a big order needs to eat many small orders?

> Assume there are many small ask orders at the best price (lowest ask). Now a big whale places a market order and buys all small ask orders. First, this often creates slippage, as you might know already.. Second, there might be not enough open ask orders to fill the big bid of the whale. Third, the whale needs to transact with many small fish accounts to fulfil the small orders. This leads to a high computational demand.

pallets/order-book/Cargo.toml
```diff
@@ -49,6 +53,8 @@ runtime-benchmarks = [
   'sp-runtime/runtime-benchmarks',
   'cfg-types/runtime-benchmarks',
   'cfg-traits/runtime-benchmarks',
+  'orml-tokens',
+  'orml-traits'
```
lemunozm, line 57: This should be:
```toml
`orml-tokens?/runtime-benchmarks`,
`orml-traits?/runtime-benchmarks`,

**Comment by thea-leake** [comment]
Still reviewing on my end, but putting it in review now as it would be good to get it merged sooner if possible to have available for frontend.  I would like to extend tests and benchmarks more, but that might be better to do in second PR.

File: pallets/order-book/src/lib.rs (hunk 2 above)
Comment (thea-leake) [reply pallets/order-book/src/lib.rs:416]: That's a very good point, as it is only for frontend it won't be called in the runtime, and therefore won't affect blockweights, but definitely a good thing to check! :)

File: pallets/order-book/src/lib.rs (hunk 4 above)
Comment (thea-leake) [reply pallets/order-book/src/lib.rs:301]: Thats a good question.  The initial plan was to match orders of corresponding currency pairs and compatible prices and min fulfillment amounts, however there was some concern with the weight of matching those up, it's something I believe we plan on revisiting in the future.

File: pallets/order-book/src/lib.rs [hunk 6]
```diff
@@ -0,0 +1,639 @@
...
+		/// (asset out)
+		pub price: ForeignCurrencyBalance,
+		/// Minimum amount of an order that can be fulfilled
+		/// for partial fulfillment
+		pub min_fullfillment_amount: ForeignCurrencyBalance,
+		/// Maximum amount of outgoing currency that can be sold
+		pub max_sell_amount: ForeignCurrencyBalance,
+	}
+
+	#[derive(Clone, Copy, Debug, Encode, Decode, Eq, PartialEq, MaxEncodedLen, TypeInfo)]
```
lemunozm, line 172: You would probably need to add here the following: `#[scale_info(skip_type_params(T))]`

pallets/order-book/src/lib.rs
lemunozm: Agree with @Chralt98, I would remove these methods to avoid bad usage from the runtime. If they're needed for testing, the tester can call `iter_prefix()` or `iter()` directly.

Forcing to use `iter` or `iter_prefix` the user can see that this implies a lot of read accesses.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,639 @@
...
+		/// account, asset IDs, and pallet nonce.
+		pub fn gen_hash(
+			placer: &T::AccountId,
+			asset_out: T::AssetCurrencyId,
+			asset_in: T::AssetCurrencyId,
+			nonce: T::Nonce,
+		) -> T::Hash {
+			(&placer, asset_in, asset_out, nonce).using_encoded(T::Hashing::hash)
+		}
+	}
```
lemunozm, line 439: Nonce should not be enough to identify it?

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,639 @@
...
+				buy_amount != T::ForeignCurrencyBalance::zero(),
+				Error::<T>::InvalidBuyAmount
+			);
+			ensure!(
+				sell_price_limit != T::ForeignCurrencyBalance::zero(),
+				Error::<T>::InvalidMinPrice
+			);
+
+			<Orders<T>>::try_mutate_exists(order_id, |maybe_order| -> DispatchResult {
+				let mut order = maybe_order.as_mut().ok_or(Error::<T>::OrderNotFound)?;
```
lemunozm, line 572: Should this always exist because it's inside of a `try_mutate_exists`?

lemunozm (commented):
Fast first review with some nitpick comments and questions to understand this better. I would make a second deep review focusing in the main methods and implementation logic.

But everything seems clear and well structured!

pallets/order-book/src/lib.rs
thea-leake: I was thinking that it would be better to have IDs that would not be easily determinable, but the `Nonce` would be usable for order identification.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,639 @@
...
+
+	/// Stores Nonce for orders placed
+	/// Given that Nonce is to ensure that all orders have a unique ID, we can
+	/// use just one Nonce, which means that we only have one val in storage,
+	/// and we don't have to insert new map values upon a new account/currency
+	/// order creation.
+	#[pallet::storage]
+	pub type NonceStore<T: Config> = StorageValue<_, T::Nonce, ValueQuery>;
+	#[pallet::storage]
+	pub type AssetPairOrders<T: Config> = StorageDoubleMap<
```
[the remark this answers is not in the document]
thea-leake: Thanks for catching, added!

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,639 @@
...
+				T::AssetRegistry::metadata(&currency_out).is_some(),
+				Error::<T>::InvalidAssetId
+			);
+			ensure!(
+				T::ReserveCurrency::can_reserve(
+					&account,
+					T::Fees::fee_value(T::OrderFeeKey::get())
+				),
+				Error::<T>::InsufficientReserveFunds,
+			);
```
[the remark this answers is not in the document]
thea-leake: I was thinking it would be cheaper to try and check before reserve, but we can also remove if wanted as well.

libs/traits/src/lib.rs
```diff
@@ -685,3 +685,46 @@ pub trait CurrencyInspect {
...
+	/// Swap tokens buying a `buy_amount` of `currency_in` using the
+	/// `currency_out` tokens. The implementator of this method should know
+	/// the current market rate between those two currencies.
+	/// `sell_price_limit` defines the lowest price acceptable for
+	/// `currency_in` currency when buying with `currency_out`. This
+	/// protects order placer if market changes unfavourably for swap order.
+	/// Returns the order id created with by this buy order if it could not
+	/// be inmediately and completelly fullfilled. If there was already an
+	/// active order with the same account currencies, the order is
+	/// increased/decreased and the same order id is returned.
```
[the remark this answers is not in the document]
thea-leake: `update_order` allows accounts to update active orders they have, and can also be used when we add partial fulfillment.

pallets/order-book/src/mock.rs
```diff
@@ -0,0 +1,223 @@
...
+frame_support::construct_runtime!(
+	  pub enum Runtime where
+		Block = Block,
+		NodeBlock = Block,
+		UncheckedExtrinsic = UncheckedExtrinsic,
+	  {
+			Balances: pallet_balances,
+			Fees: pallet_mock_fees,
+			System: frame_system,
+		  OrmlTokens: orml_tokens,
```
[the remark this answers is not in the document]
thea-leake: Agreed, I think this would definitely be a good change for the next PR.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,639 @@
...
+		/// Type for pallet weights
+		type Weights: WeightInfo;
+	}
+	//
+	// Storage and storage types
+	//
+	/// Order Storage item.
+	/// Contains fields relevant to order information
+	#[derive(Clone, Copy, Debug, Encode, Decode, Eq, PartialEq, MaxEncodedLen, TypeInfo)]
+	pub struct Order<OrderId, AccountId, AssetId, ForeignCurrencyBalance> {
```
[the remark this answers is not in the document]
thea-leake: Agreed, I've added an OrderOf type to restrict to appropriate Config types.

pallets/order-book/src/lib.rs
thea-leake: Ah, that struct was superfluous, I had missed removing it before.  Removed now.

pallets/order-book/src/lib.rs
thea-leake: As long as a valid order is submitted it should always exist.

libs/traits/src/lib.rs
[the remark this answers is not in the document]
lemunozm: But given that comment, it seems like `place_order` does the same if the order already exits.

pallets/order-book/src/benchmarking.rs
```diff
@@ -0,0 +1,139 @@
...
+
+	Fees::mock_fee_value(|_| 0);
+	Fees::mock_fee_to_author(|_, _| Ok(()));
+}
+
+const CURRENCY_0: u128 = 1_000_000_000_000_000_000u128;
+const CURRENCY_1: u128 = 1_000_000_000_000_000u128;
+
+benchmarks! {
+		where_clause {
```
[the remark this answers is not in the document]
lemunozm: Thanks! 🙌🏻

pallets/order-book/src/lib.rs
lemunozm: > it would be better to have IDs that would not be easily determinable

Why this? Can you expand?

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
lemunozm: No strong opinion here 👍🏻. It's just I'm always checking lines that could be removed 😆.

To give my reasoning, I think in case of an error, the weight of this call will be the same, which is higher than the computation time the node wasted on it for an error. The weight will be added to the current block weight as if the call was ok. So, from my understanding, we would not be saving computation or making it cheaper.

pallets/order-book/src/lib.rs
lemunozm: But if it was not submitted, will the callback of `try_mutate_exists` not be called?

thea-leake: Ah, I meant if a valid order id is passed to the update fn.  If there isn't a matching order then it should return an error for no matching order.

pallets/order-book/src/lib.rs
thea-leake: It's updated to check the fee currency against the asset out currency, and consolidate the reserves when they match.

thea-leake: Currently updating other reserve/unreserve locations to make sure correct currency type is reserved/unreserved when the outgoing currency matches the fees so the mechanism is always consistent (i.e. so we don't have any reserves done through native balances trait impl that are undone via orml impl.)

thea-leake: All reserve/unreserve should be covered now, cleaning it up, adding additional coverage around this as well now.

pallets/order-book/src/lib.rs
```diff
@@ -120,6 +120,14 @@ pub mod pallet {
 		/// Fee Key used to find amount for allowance reserve/unreserve
 		type OrderFeeKey: Get<<Self::Fees as Fees>::FeeKey>;
 
+		/// Token Id for token used for fee reserving
+		/// as represented by `AssetCurrencyId`.
+		/// Used for reserve-able fund checking to ensure
+		/// amount traded and storage fees can be reserved
+		/// when trading for fee currency.
+		/// This should typically be native chain currency.
+		type FeeCurrencyId: Get<Self::AssetCurrencyId>;
```
lemunozm, line 133: When this will not be the native currency? Is it just a way to identify our currency in the `MultiReservableCurrency`?

thea-leake: Yeah exactly :) -- there aren't any forseeable situations where it wouldn't be native, but we still need to be able to see if it is native based off the asset id.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,662 @@
...
+		#[pallet::weight(T::Weights::fill_order_full())]
+		pub fn fill_order_full(origin: OriginFor<T>, order_id: T::Hash) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+			let order = <Orders<T>>::get(order_id)?;
+			// maybe move to ensure if we don't need these later
+			// might need decimals from currency, but should hopefully be able to use FP
+			// price/amounts from FP balance
+			T::AssetRegistry::metadata(&order.asset_out_id).ok_or(Error::<T>::InvalidAssetId)?;
+
+			T::AssetRegistry::metadata(&order.asset_in_id).ok_or(Error::<T>::InvalidAssetId)?;
```
lemunozm, line 369: I think this can never be possible because an order was created if both were in the metadata.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,662 @@
...
+				let total_reserve_amount = Self::get_combined_reserve(max_sell_amount)?;
+				T::ReserveCurrency::reserve(&account, total_reserve_amount)?;
+			} else {
+				T::ReserveCurrency::reserve(&account, fee_amount)?;
+
+				T::TradeableAsset::reserve(currency_out, &account, max_sell_amount)?;
+			}
+
+			let new_nonce = <NonceStore<T>>::get();
+			let order_id = Self::gen_hash(&account, currency_in, currency_out, new_nonce);
```
lemunozm, line 505: I do not understand why not just use `new_nonce` as an `order_id`. So instead of using a Hash type, use the nonce number directly as identification.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,662 @@
...
+			ensure!(
+				T::AssetRegistry::metadata(&currency_in).is_some(),
+				Error::<T>::InvalidAssetId
+			);
+			ensure!(
+				T::AssetRegistry::metadata(&currency_out).is_some(),
+				Error::<T>::InvalidAssetId
+			);
+			<NonceStore<T>>::try_mutate(|n| {
+				*n = n.ensure_add(T::Nonce::one())?;
```
lemunozm, line 489: Totally nitpick: there is a `ensure_add_assign` for this cases

lemunozm (commented):
Second review round, mostly questions to understand it better and some suggestions. I would appreciate another eyes on this 🙌🏻.

Thanks for your work on this! It's pretty complex and seems like you've thought in weird border cases.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,662 @@
...
+	#[pallet::call]
+	impl<T: Config> Pallet<T>
+	where
+		<T as frame_system::Config>::Hash: PartialEq<<T as frame_system::Config>::Hash>,
+	{
+		/// Create an order, with the minimum fulfillment amount set to the buy
+		/// amount, as the first iteration will not have partial fulfillment
+		#[pallet::call_index(0)]
+		#[pallet::weight(T::Weights::create_order_v1())]
+		pub fn create_order_v1(
```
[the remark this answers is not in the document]
thea-leake: Ah, the initial orders are pretty simple, it's currently just specifying a buy price with full fulfillment, but we're going to have partial fulfillment as well in the future, so I wanted to make sure that we could add that functionality in the future without having to modify an existing extrinsic.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,662 @@
...
+			Self::place_order(
+				account_id, asset_in, asset_out, buy_amount, price, buy_amount,
+			)?;
+			Ok(())
+		}
+
+		///  Cancel an existing order that had been created by calling account.
+		#[pallet::call_index(1)]
+		#[pallet::weight(T::Weights::user_cancel_order())]
+		pub fn user_cancel_order(origin: OriginFor<T>, order_id: T::Hash) -> DispatchResult {
```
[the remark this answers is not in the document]
thea-leake: Yeah, it's the name collision with the trait, so wanted to specify that this would be the one callable by users.

runtime/development/src/lib.rs
```diff
@@ -1845,6 +1845,24 @@ impl pallet_transfer_allowlist::Config for Runtime {
...
+	type AssetCurrencyId = CurrencyId;
+	type AssetRegistry = OrmlAssetRegistry;
+	type FeeCurrencyId = NativeToken;
+	type Fees = Fees;
+	type ForeignCurrencyBalance = Balance;
+	type Nonce = u64;
+	type OrderFeeKey = OrderBookCreationFeeKey;
+	type ReserveCurrency = Balances;
+	type RuntimeEvent = RuntimeEvent;
+	type TradeableAsset = OrmlTokens;
```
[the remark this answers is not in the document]
thea-leake: Ah, yeah I wanted to use the `MultiReserveCurrency` trait, but if we want to use the `restricted-transfers` pallet we can switch it over.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,662 @@
...
+
+			ensure!(
+				T::TradeableAsset::can_reserve(order.asset_out_id, &account_id, sell_amount),
+				Error::<T>::InsufficientAssetFunds,
+			);
+			T::TradeableAsset::unreserve(
+				order.asset_out_id,
+				&order.placing_account,
+				order.max_sell_amount,
+			);
```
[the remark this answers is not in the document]
thea-leake: I was using `reserve`/`unreserve` as those where provided by orml via `MultiReservableCurrency`, which seemed the safest bet for foreign asset trades.  But `hold`/`release` do seem more preferable here.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,662 @@
...
+	/// Map of Vec containing OrderIds of same asset in/out pairs.
+	/// Allows looking up orders available corresponding pairs.
+	#[pallet::storage]
+	pub type AssetPairOrders<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		T::AssetCurrencyId,
+		Twox64Concat,
+		T::AssetCurrencyId,
+		BoundedVec<T::Hash, ConstU32<1_000_000>>,
```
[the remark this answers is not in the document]
thea-leake: Good call on the bounded vec size, I'll get that update.

The AssetPairOrders is to provide a quick means of querying storage for orders that match a desired trade.

pallets/order-book/src/lib.rs
thea-leake: Ah, yeah it seemed better to have the check in place, but it would make sense to move to the extrinsic fn from the `TokenSwaps` impl.

pallets/order-book/src/lib.rs
thea-leake: Yeah, the nonce would work here, and simplify things a bit.  I might have been a bit paranoid with the order id determinability.

pallets/order-book/src/lib.rs
```diff
@@ -457,23 +469,25 @@ pub mod pallet {
...
+					T::Fees::fee_value(T::OrderFeeKey::get())
+						.try_into()
+						.map_err(|_| Error::<T>::BalanceConversionErr)?;
+				let total_reserve_amount = max_sell_amount.ensure_add(fee_reserve_balance)?;
+
+				T::TradeableAsset::reserve(currency_out, &account, total_reserve_amount)?;
+			} else {
+				T::ReserveCurrency::reserve(&account, T::Fees::fee_value(T::OrderFeeKey::get()))?;
+
+				T::TradeableAsset::reserve(currency_out, &account, max_sell_amount)?;
```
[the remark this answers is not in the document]
thea-leake: This is for when we have the storage fee reserve, and the trade currency reserve to ensure there are enough funds in the account reserve to complete the trade when the order is fulfilled.

pallets/order-book/src/lib.rs
thea-leake: There's only one way of identifying the order in the system right now, the nonce is just used to ensure each order gets a new hash, but I haven't thought of any ways that someone would be able to maliciously use a predicted order id, so I think it would be fine to change it to nonce.

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
thea-leake: Updated to use pallet const conf type.

pallets/order-book/src/lib.rs
thea-leake: It was mostly in case there were anything potentially with predictable order ids, however I'm not aware of any situations where that would be an issue here.  I've updated to use nonce 👍

pallets/order-book/src/lib.rs
thea-leake: Updated to use nonce for order id.

pallets/order-book/src/lib.rs
thea-leake: Ah, yeah in this case if the order was successfully updated there wouldn't be any invalid assets.

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Ok 👍🏻 . Anyways, from my understanding, I think it's not a problem if the existing extrinsic is modified later, communicating a break API change (or internal change)

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I think it is totally fine having both names at the same time. If the user wants to call the trait once, they can use `<Self as Trait>::cancel_order()`. I also say that because `user_` is a generic word. It changes depending on the context: every call has a "user" who calls it.

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Not sure exactly what the difference is, TBH. Is it the same thing, or one of them does some transfer under the hood?

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
lemunozm: >The AssetPairOrders is to provide a quick means of querying storage for orders that match a desired trade.

And where is it used? Is it intended to be used by the front-end?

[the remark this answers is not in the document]
thea-leake: Yes, this is primarily for front end right now, it will also be necessary if/when we add fulfilling orders by opposite asset ids with compatible prices.

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
thea-leake: All of the reserve checks except the reserve check before the transfers to ensure enough funds can actually be transferred were removed.  Also just Found/fixed an error with fund check & added tests when double checking again for extra superfluous checks.

libs/traits/src/lib.rs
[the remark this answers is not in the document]
thea-leake: Ah, the current `place_order` creates a new order instead of updating the order, the docstring wasn't updated to match the behaviour, but it might be better to have behaviour match docstring.
@wischli would it be better for `place_order` to update an existing order for an account/currency pair?
If that's the case it would make sense to switch order id back to hash, but without nonce--this should be a simple change though.

lemunozm (dismissed):
Approved! We can merge it to unlock Apps team. Thanks for your work on this Thea! 🙌🏻 

Anyways, I would love to have other eyes on it, because I'm not fully inmersed in this topic. Would be great to have a postmorten review maybe from @mustermeiszer @wischli

libs/traits/src/lib.rs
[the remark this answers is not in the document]
thea-leake: Verified current behaviour is good, and updated docstring to make sure it's accurate.

lemunozm (approved):

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,655 @@
...
+			+ EnsureAdd
+			+ MaybeSerializeDeserialize
+			+ TypeInfo
+			+ MaxEncodedLen;
+
+		/// Type for currency orders can be made for
+		type TradeableAsset: MultiReservableCurrency<
+			Self::AccountId,
+			Balance = <Self as pallet::Config>::ForeignCurrencyBalance,
+			CurrencyId = Self::AssetCurrencyId,
```
mustermeiszer, line 165: I think this should be a Substrate native trait instead of orml one.

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,655 @@
...
+			+ Copy
+			+ EnsureAdd
+			+ MaybeSerializeDeserialize
+			+ TypeInfo
+			+ MaxEncodedLen;
+
+		/// Type for currency orders can be made for
+		type TradeableAsset: MultiReservableCurrency<
+			Self::AccountId,
+			Balance = <Self as pallet::Config>::ForeignCurrencyBalance,
```
mustermeiszer, line 164: The `Balance` type should be equal for `Reservable` and `ForeignCurrency`

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,655 @@
...
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::Weights::fill_order_full())]
+		pub fn fill_order_full(origin: OriginFor<T>, order_id: T::OrderIdNonce) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+			let order = <Orders<T>>::get(order_id)?;
+			// maybe move to ensure if we don't need these later
+			// might need decimals from currency, but should hopefully be able to use FP
+			// price/amounts from FP balance
+
+			let sell_amount = order.buy_amount.ensure_mul(order.price)?;
```
mustermeiszer, line 376: IIUC we are missing decimal conversion here. WDYT?

mustermeiszer (commented):
Maybe I am missing something, but I think we are missing some essential decimal conversion. 

Furthermore, I am not understanding `sell_price_limit`. ^^ I need an example with numbers here. @thea-leake

mustermeiszer:
I also think we should make `sell_price_limit` a `FixedPointType` and name it `sell_rate_limit`. @lemunozm or do you think this is misusing `Rate` where we need a `Balance`?

If I get it right, this limit is saying. I want at least $x\frac{CurrencyOut}{CurrencyIn}$

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,655 @@
...
+					&order.placing_account,
+					order.max_sell_amount,
+				);
+				T::ReserveCurrency::unreserve(
+					&order.placing_account,
+					T::Fees::fee_value(T::OrderFeeKey::get()),
+				);
+			};
+			Ok(())
+		}
```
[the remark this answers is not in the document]
lemunozm: >I was wondering how much work it would be to make fees also usable for handling deposits at a centralised location in our system

Could you expand this? Not sure what you mean by a centralized location

pallets/order-book/src/lib.rs
lemunozm: Yes... I think you're right. If used `Balance` here have different decimal representation, we should unify them.

libs/traits/src/lib.rs
```diff
@@ -687,3 +687,44 @@ pub trait CurrencyInspect {
...
 }
+
+pub trait TokenSwaps<Account> {
+	type CurrencyId;
+	type Balance;
+	type OrderId;
+	/// Swap tokens buying a `buy_amount` of `currency_in` using the
+	/// `currency_out` tokens. The implementator of this method should know
+	/// the current market rate between those two currencies.
+	/// `sell_price_limit` defines the lowest price acceptable for
```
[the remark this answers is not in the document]
lemunozm: I understand `sell_price_limit` as a way of protecting my order against a big conversion ratio. 

So, if I want to buy 10 selling 20, it will be ok if `sell_price_limit` is 30, but if the conversion drops to 0.25, then I would not sell 40 to buy 10. (double-check this @thea-leake)

> Would this be the minimum amount of currency_in a user wants to get for 1 unit of currency_out?

I understand define it as the maximum amount of `currency_out` I would give for getting the `buy_amount` of `currency_in`

lemunozm:
From my understanding of this:
- I would define it as a `Rate` and name it `sell_rate_limit` if it is defined as: $\frac{CurrencyOut}{CurrencyIn}$
- I would define it as a `Balance` and name it `sell_price_limit` (as it is) if it is defined as: $x\frac{CurrencyOut}{CurrencyIn}$

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: One pallet doing reserves. Either variable or with key values as for fees. For each account. Making it easy to follow for which reasons the reserve happened if we add an identifier to the different amounts or so.

libs/traits/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: as we use it as `buy_amount.ensure_mul(sell_price_limit)` I think it is not as you described it above, right? In your example it would rather be
* `buy_amount = 10.00`
* `sell_price_limit = 0.5`

But I think we are having a logical bug here.

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
lemunozm: So just what `pallet_fees` does, but for reserves, right? 
`pallet_fees` has methods as `fee_to_author/fee_to_treasury`/... You may say something like: `fee_to_reserve(account, fee_key)`?

pallets/order-book/src/lib.rs
thea-leake: Yeah, I think you're right, as we're dealing with the overall balances and not the adjusted precision from the registry.

thea-leake:
Yeah, I agree it should be a rate; I think having the prices as Fixed Points would be much easier to reason about from a usage perspective.

lemunozm:
The `Rate/Balance` issue greets again 😆. Yes, I thought about that too, maybe from the perspective of this pallet, prices should be `Rate`, but the issue is that the current accounts for the currencies are `Balance`, so you are tied to use `Balance` and add the conversions inside this pallet...

pallets/order-book/src/lib.rs
lemunozm: Regarding the conversions, in the Oracle PR, I add a module to perform that kind of conversions. Maybe you can extend that module @thea-leake 
 with a `balance_to_balance()` conversion: https://github.com/centrifuge/centrifuge-chain/pull/1474/files#diff-833d286c0e9655fa1b12da1bceabc7ff8988d1c74b7149be553320af61b78b54
 
Hopefully this PR is merged today or tomorrow

Issue #1277, opened 2023-03-22
Normalize `Moment` generic/type


lemunozm:
Looking at the `cfg_traits` module, we can see that some traits take `Moment` as a generic and others as `cfg_primitive::Moment`. Mixing both `Moment`s we have the worst of both worlds. We depend on a fixed moment type, and we need to pass a generic bound over the code that needs and will end to `cfg_primitive::Moment`.

I propose:
- Option 1 (most scalable): Make everything generic, to take real advance of the generic and be able to configure the runtime with different `Moment` types. Even other chains in the future can use us easily.
- ~~Option 2 (most pragmatic): Use everywhere the `cfg_primitive::Moment` and get rid of all generic Moments that will end up in the `cfg_primitive::Moment`.~~ => It doesn't work with this proposal: #1380

lemunozm:
Knowing that most of our traits use the generic version, I'll go by Option 1

PR #1434, opened 2023-06-30
Change principal overdue trigger to seconds based

offerijns:
# Description

Change principal overdue trigger to seconds based. There's no good reason for it being days based, any frontend can easily just show the time difference in days and set the policy based on days. But basing it on seconds makes it more flexible for potential future use cases.

Migration is not needed as the format of the storage doesn't change, the triggers of any testnet pools will just change from x days overdue to x seconds overdue.

# Checklist:


pallets/loans/src/migrations/v1.rs
```diff
@@ -16,7 +16,7 @@ mod v0 {
 
 	#[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeDebug, MaxEncodedLen)]
 	pub struct WriteOffState<Rate> {
-		pub overdue_days: u32,
+		pub overdue_secs: u32,
```
lemunozm, line 19: Now we are in the second's magnitude, I would set the type to `Moment` that matches with what `maturity_date` is.

pallets/loans/src/entities/loans.rs
```diff
@@ -225,9 +224,8 @@ impl<T: Config> ActiveLoan<T> {
 	) -> Result<bool, DispatchError> {
 		let now = T::Time::now().as_secs();
 		match trigger {
-			WriteOffTrigger::PrincipalOverdueDays(days) => {
-				let overdue_secs = SECONDS_PER_DAY.ensure_mul(days.ensure_into()?)?;
-				Ok(now >= self.maturity_date().ensure_add(overdue_secs)?)
+			WriteOffTrigger::PrincipalOverdue(overdue_secs) => {
+				Ok(now >= self.maturity_date().ensure_add((*overdue_secs).into())?)
```
lemunozm, line 228: Then, you do not need the `into()` conversion here

pallets/loans/src/tests/util.rs
```diff
@@ -45,7 +47,9 @@ pub fn set_up_policy(percentage: f64, penalty: f64) {
 	MockChangeGuard::mock_released(move |_, _| {
 		Ok(Change::Policy(
 			vec![WriteOffRule::new(
-				[WriteOffTrigger::PrincipalOverdueDays(1)],
+				[WriteOffTrigger::PrincipalOverdue(
+					SECONDS_PER_DAY.try_into().unwrap(),
```
lemunozm, line 51: And this will be simply `SECONDS_PER_DAY`

lemunozm (commented):
Thanks for this PR! Only one minor comment

pallets/loans/src/migrations/v1.rs
offerijns: If we do that, I think we'll need to write a migration, since Moment is a u64 rather than a u32. But beyond that, a time difference of u32 can afaik represent 4294967295 seconds = 49710 days, which I think suffices.

lemunozm: But for now, we are avoiding migrations because it's not used, isn't it? (i.e. we are avoiding migrations in the refactor v3).

Yeah, in terms of precision, we're safe, although I think it seems weird from a reader's perspective to have seconds represented in the same pallet as `u64` and `u32` (Actually, this should be generalized and let the runtime choose the precision: #1277).

lemunozm (approved):
My previous comment is not blocking. LGTM!

pallets/loans/src/migrations/v1.rs
lemunozm: Also curious if a migration is needed to go from `u32` to `u64`. Maybe because no information is lost, it's not an issue 🤔 . Do you know about this @wischli?

lemunozm (dismissed):
It's perfect! 🙌🏻

lemunozm (approved):

PR #1440, opened 2023-07-05
refactor: `TrancheTokenPrice` trait

NunoAlexandre:
Addresses this bit of the Connectors v2 spec:

> Currently the trait PoolInspect also contains the get_tranche_token_price. A few arguments for refactoring
> 
> - Naming is misleading. Actually, it is computing the price
> - The trait should not be bloated. It is not inspecting here.
> - Implement something like trait TranchePricer or so
> - Maybe even store calculate prices, BUT NOT really sure if this is useful

Note: Leaving the last point since it's an optimisation that we don't need at the moment.

### DIFF
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -119,15 +119,7 @@ pub trait PoolNAV<PoolId, Amount> {
 /// A trait that support pool inspection operations such as pool existence
 /// checks and pool admin of permission set.
 pub trait PoolInspect<AccountId, CurrencyId> {
-	type PoolId: Parameter
-		+ Member
-		+ Debug
-		+ Copy
-		+ Default
-		+ TypeInfo
-		+ Encode
-		+ Decode
-		+ MaxEncodedLen;
+	type PoolId;
 	type TrancheId;
 	type Rate;
 	type Moment;
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -128,7 +128,7 @@ pub trait PoolInspect<AccountId, CurrencyId> {
 		+ Encode
 		+ Decode
 		+ MaxEncodedLen;
-	type TrancheId: Parameter + Member + Debug + Copy + Default + TypeInfo + MaxEncodedLen;
+	type TrancheId;
 	type Rate;
 	type Moment;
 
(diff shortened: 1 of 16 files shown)


libs/traits/src/lib.rs
```diff
@@ -135,18 +135,35 @@ pub trait PoolInspect<AccountId, CurrencyId> {
...
+	type PoolId: Parameter
+		+ Member
+		+ Debug
+		+ Copy
+		+ Default
+		+ TypeInfo
+		+ Encode
+		+ Decode
+		+ MaxEncodedLen;
+	type TrancheId: Parameter + Member + Debug + Copy + Default + TypeInfo + MaxEncodedLen;
```
lemunozm, line 157: Open a discussion because I always have the same dilemma. Should we add these bounds here? They seem to be required only for a specific implementation (for example, an implementation where `PoolId` is encoded in a database). On the other hand, I know that adding these bounds here is easy to handle on the pallet side because you do not need to add some `where` bounds.

NunoAlexandre: Thanks for raising this point. I had the same thought to be honest. For me, the least amount of "things" the best since that means less clutter, less inter dependencies, and less uneccessary code to maintain. I'll have a further look at this and how we can reduce its footprint

wischli: > Thanks for raising this point. I had the same thought to be honest.

Same here! I should have raised that question earlier. I am on the same page as Nuno.

libs/traits/src/lib.rs
```diff
@@ -135,18 +135,35 @@ pub trait PoolInspect<AccountId, CurrencyId> {
...
-	fn get_tranche_token_price(
-		pool_id: Self::PoolId,
-		tranche_id: Self::TrancheId,
-	) -> Option<PriceValue<CurrencyId, Self::Rate, Self::Moment>>;
 
 	/// Get the account used for the given `pool_id`.
 	fn account_for(pool_id: Self::PoolId) -> AccountId;
 
-	// Get the currency used for the given `pool_id`.
+	/// Get the currency used for the given `pool_id`.
```
wischli, line 142: Thanks for cleaning up my dirt 🧹

wischli (approved):
LGTM! Not sure if we need to remove the `PoolIdOf` and `TrancheIdOf` types but I suppose this way is cleaner.

libs/traits/src/lib.rs
NunoAlexandre: Done 👍

pallets/loans/src/lib.rs
```diff
@@ -198,19 +194,27 @@ pub mod pallet {
 		type NonFungible: Transfer<Self::AccountId>
 			+ Inspect<Self::AccountId, CollectionId = Self::CollectionId, ItemId = Self::ItemId>;
 
+		/// The PoolId type
+		type PoolId: Member + Parameter + Default + Copy + HasCompact + MaxEncodedLen;
```
lemunozm, line 198: Ok, so the downside of leaving the trait cleaned, is the addition of an associated type to the config pallet that wants to use such a trait. 👍🏻

Ideally would be great to add those bounds in the `Config` where clause. I've tried before, but it requires to be propagated to all impl blocks that uses `Config`, and it comes with a lot of boilerplate and the errors are quite obfuscated. So I think this is the less painful way 👍🏻 

Thanks to do the test Nuno!

lemunozm (approved):
LGTM!

Issue #1097, opened 2022-11-28
Use more flexible interest rate type


offerijns:
### Which part of the code is the issue addressing?

* [ ] Addressing a bug
* [x] Refactoring
* [ ] Feature

### Description
- [ ] Add new `InterestRateType` to `cfg-types`:
```rust
pub enum InterestRate<Rate> {
  Fixed(Rate)
}
```
- [ ] Use new `InterestRate` enum for associated type of `interest-accrual` pallet: https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/interest-accrual/src/lib.rs#L143 and adapt methods to accept this new data type
- [ ] Adapt https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/loans/src/types.rs#LL153C36-L153C40
- [ ] Write migration for loans
- [ ] Adapt https://github.com/centrifuge/centrifuge-chain/blob/main/libs/types/src/tranches.rs#L96
- [ ] Write migration for tranches

lemunozm:
@offerijns I don't fully understand this change. Why it's needed?

PR #1443, opened 2023-07-05
InterestAccrual: use more flexible interest rate type

offerijns:
# Description

Address https://github.com/centrifuge/centrifuge-chain/issues/1097


libs/traits/src/lib.rs
```diff
@@ -281,56 +285,103 @@ pub trait CurrencyPrice<CurrencyId> {
...
+			} => {
+				let new_rate = rate_per_year.ensure_sub(rate)?;
+				Ok(InterestRate::Fixed {
+					rate_per_year: new_rate,
+					compounding,
+				})
+			}
+		}
+	}
+}
```
lemunozm, line 333: I think could be great to add a `map/try_map` function to `InterestRate` and let the caller call the operation they want over the rate. Similar to https://github.com/centrifuge/centrifuge-chain/pull/1435/files#diff-14aad6e50ab4015efc0d5939359c1676112c04751b9a825c15e1a588677da0b6R42

libs/traits/src/lib.rs
```diff
@@ -281,56 +285,103 @@ pub trait CurrencyPrice<CurrencyId> {
...
 }
 
 /// A collection of pre-computed interest rates for performing interest accrual
-pub trait RateCollection<InterestRate, Balance, NormalizedDebt> {
+pub trait RateCollection<Rate, Balance, NormalizedDebt> {
 	/// Calculate the current debt using normalized debt * cumulative rate
 	fn current_debt(
 		&self,
-		interest_rate_per_sec: InterestRate,
+		interest_rate_per_sec: Rate,
```
lemunozm, line 384: I think your previous question come because the name is not updated to `interest_rate_per_year` here.

libs/traits/src/lib.rs
```diff
@@ -281,56 +285,103 @@ pub trait CurrencyPrice<CurrencyId> {
...
+					compounding,
+				})
+			}
+		}
+	}
+}
+
 /// A trait that can be used to calculate interest accrual for debt
-pub trait InterestAccrual<InterestRate, Balance, Adjustment> {
+pub trait InterestAccrual<Rate, Balance, Adjustment> {
```
lemunozm, line 336: Definitely, I would move now this trait and the `InterestRate` related types to their own module `cfg_traits::interest_rate`.

lemunozm (commented):
Very good and clean and readable change, as a surgeon! 🙌🏻

Left a couple of minor changes that could be done in the future too.

I've seen the PR is over the other PR of principal & interest separation. Could we merge this in main once the other is merged? Just to not merge a lot of concepts in one PR.

pallets/interest-accrual/src/lib.rs
```diff
@@ -437,118 +436,137 @@ pub mod pallet {
...
-							reference_count: 1,
-						};
-
-						rates
-							.try_push(new_rate)
-							.map_err(|_| Error::<T>::TooManyRates)?;
-
-						Ok(())
-					}
+		pub fn reference_interest_rate(
```
branan, line 439: This method was already pretty deeply nested and it's worse now. Might be worth trying to break out the guts of it into its own function that just mutates the vector or something?

pallets/interest-accrual/src/lib.rs
```diff
@@ -437,118 +436,137 @@ pub mod pallet {
...
+								Ok(())
+							}
+						}
+					})
 				}
-			})
+			}
 		}
 
 		pub fn unreference_interest_rate(
```
branan, line 473: As above

offerijns:
> I've seen the PR is over the other PR of principal & interest separation. Could we merge this in main once the other is merged? Just to not merge a lot of concepts in one PR.

Yeah I think that's fine! Just based it on that PR because it contained relevant changes, but we can just rebase this one after merging that one.

pallets/interest-accrual/src/lib.rs
offerijns: I guess it makes sense, but let's note that as a future clean up? Doesn't seem relevant particularly to this PR and this one is somewhat urgent.

libs/traits/src/lib.rs
offerijns: I actually prefer ensure add impl, because this is slightly more specific. The advantage is, say at some point there is also a floating rate variant next to the fixed variant. Then with map, it's unclear what field this should touch, and the implementor would need to handle both cases. Versus with ensure add the impl on interest rate can decide that adding always means increasing the inner rate.

offerijns: But we can still definitely also add `map/try_map` in the future, it can't hurt

libs/traits/src/lib.rs
```diff
@@ -281,56 +285,111 @@ pub trait CurrencyPrice<CurrencyId> {
...
+
+impl<Rate> InterestRate<Rate> {
+	pub fn rate_per_year(&self) -> &Rate {
+		match self {
+			InterestRate::Fixed { rate_per_year, .. } => rate_per_year,
+		}
+	}
+}
+
+impl<Rate: EnsureAdd + EnsureSub> InterestRate<Rate> {
```
mustermeiszer, line 311: Any reason why we are not implementeing the trait itself on the Type?

pallets/interest-accrual/src/lib.rs
```diff
@@ -437,118 +436,137 @@ pub mod pallet {
...
-					&& interest_rate_per_year >= Zero::zero()
-					&& (interest_rate_per_year.saturating_mul(four_decimals)).frac()
-						== Zero::zero(),
-				Error::<T>::InvalidRate
-			);
-			Ok(())
+			match interest_rate_per_year {
+				InterestRate::Fixed { rate_per_year, .. } => {
+					let four_decimals = T::Rate::saturating_from_integer(10000);
+					let maximum = T::Rate::saturating_from_integer(MAX_INTEREST_RATE);
```
mustermeiszer, line 516: Rather for @branan and @lemunozm. Why are we using saturating math here?

libs/traits/src/lib.rs
lemunozm: You "cannot", this trait is implemented automatically for each T that already is `CheckedAdd`. You could implement `CheckedAdd` for this type, having automatically an `EnsureAdd` impl.

pallets/interest-accrual/src/lib.rs
lemunozm: Because `MAX_INTEREST_RATE` is a const, we can ensure it's ok at compile time. But we could use `check_from_integer()` too, to be sure the constant value is not changed after the conversion

libs/traits/src/lib.rs
lemunozm: `map/try_map` usually only acts over the most important property. i.e. `map` only maps the ok branch in a `Result` type. For this type, the core field is the inner rate, so I think it makes sense to only map over it. Other properties are secondary. 

Anyways, it can makes sense to have an `ensure_add()` method directly in `InterestRate`. In that case, it can be built using `try_map` to factorize most of the code as follows:

```rust
fn ensure_add(&self, rate: Rate) -> Result<Self, DispatchError> {
    self.try_map(|rate_per_year| rate_per_year.ensure_add(rate))
}

fn ensure_sub(&self, rate: Rate) -> Result<Self, DispatchError> {
    self.try_map(|rate_per_year| rate_per_year.ensure_sub(rate))
}
```

lemunozm: Of course, any of this is required for this PR, and could be added in the future 👍🏻

libs/traits/src/lib.rs
```diff
@@ -281,56 +285,111 @@ pub trait CurrencyPrice<CurrencyId> {
...
 }
 
+#[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeDebug, MaxEncodedLen)]
+pub enum CompoundingSchedule {
+	/// Interest compounds every second
+	Secondly,
+}
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeDebug, MaxEncodedLen)]
+pub enum InterestRate<Rate> {
```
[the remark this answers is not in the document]
offerijns: I'm sure you're right, but for my understanding, why?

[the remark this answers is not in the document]
lemunozm: For small types, such as simple enums, numbers, or pointers, making them `Copy` makes it easy to handle them later i.e. in iterators and vectors. As a rule of thumb, anything under 128bits can (should?) be copied. Things bigger can get worse performance because of the bytes copied, and a reference is preferred. For those cases is better to not add `Copy` and let the caller pass it as a reference or explicitly clone if needed.

An important detail is that Copy only can be added if all internal types are already `Copy` before. So, adding a non-copiable field to a `Copy` struct can lead to a big refactor of the code base.

Given said this, I would add `Copy` to `CompoundingSchedule` because it's like a simple mark that adds an information unit. But not sure about adding it to `InterestRate`.

libs/traits/src/lib.rs
mustermeiszer: Ah, right. Okay. Well not mandatory, but nice to have `CheckedAdd` then ^^

lemunozm: Although I would like to do it. It's is required to be `PartialOrd` + `CheckedAdd` to be `EnsureAdd`, and it's difficult to determine a `PartialOrd` with arithmetic meaning for a type with several fields.

This is something that now, after a lot of usage, I would design differently. Instead of making this trait automatic for any `T`, I would implement it for basic & substrate types, allowing then any custom implementation for more complex custom time.

Maybe I do a new PR to Substrate this change. I think it can be beneficial, although it will be a huge breaking change, and I do not know if they would approve it.

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -45,8 +46,8 @@ impl<T: Config> ExternalPricing<T> {
 	pub fn validate(&self) -> DispatchResult {
 		if let MaxBorrowAmount::Quantity(quantity) = self.max_borrow_amount {
 			ensure!(
-				quantity.frac().is_zero(),
-				Error::<T>::AmountNotMultipleOfPrice
+				quantity.frac().is_zero() && quantity > T::Rate::zero(),
+				Error::<T>::AmountNotNaturalNumber
 			)
 		}
```
lemunozm, line 53: Tech debt from #1450 here

mustermeiszer (commented):
> From the issue, Jeroen says that we should also change the rate from pool's Tranches. Nevertheless, changing things in the pool will require a migration (or maybe not?), and by now we were not touching other tnings from the pool, and could obstaculize more the release. I think we could leave this for a future pool's PR although it will lead to a migration.

Completely agree, that we should not touch pools right now. Although, this will make migrating later on hellish, but I guess we can not prevent that. Refactoring the pool will be a big future task

Regarding the rest. 

* I do not really like using the `enum InterestRate` as a fixed type in the pallet and trait interface. Does not feel like the right change.


Wdyt about creating a a trait that bounds the parameter to satisfy all needs. Something like

```rust
trait Rate {
     type Interest: FixedPointNumberExt;

     fn steps_per_year() -> impl FixedPointOperand;

     fn interest_rate_per_year(&self) -> Self::Interest;

     fn steps(&self, last: Moment, now: Moment) -> usize;
}
```

So `validate` becomes:

```rust
pub(crate) fn validate_interest_rate(
			rate: T::Rate,
		) -> DispatchResult {
			let max_decimals = T::InterestRate::saturating_from_integer(T::MaxDecimals::get());
			let maximum = T::InterestRate::saturating_from_integer(MAX_INTEREST_RATE);
			ensure!(
				rate.interest_rate_per_year() <= maximum
					&& rate.interest_rate_per_year() >= Zero::zero()
					&&(rate.interest_rate_per_year().saturating_mul(max_decimals)).frac()
						== Zero::zero(),
				Error::<T>::InvalidRate
			);
			Ok(())
		}
```
Furthermore, we would need to adapt
* finding `Rates` based on steps per year and interest per step
* `on_initialize` calculation
* other ?

Is just my first thought, but we should be able to generailze the accrual pallet like that and we are not forced to do a migration once we want montly accrual and so on.

pallets/interest-accrual/src/lib.rs
```diff
@@ -162,19 +158,19 @@ pub use pallet::*;
...
+type RateDetailsOf<T> = RateDetails<<T as Config>::Rate>;
 
 // Storage types
 #[derive(Encode, Decode, Default, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo)]
-pub struct RateDetailsV1<InterestRate> {
-	pub accumulated_rate: InterestRate,
+pub struct RateDetailsV1<Rate> {
+	pub accumulated_rate: Rate,
 	pub reference_count: u32,
 }
```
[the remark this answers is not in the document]
lemunozm: Actually, we do not need it. We can safely remove it.

lemunozm:
Really like your idea of extracting the behavior of an interest rate into a trait!

Only two things:
- Does it make sense another implementation for the `Rate` trait, rather than the own `InterestRate`, right now I'm not able to think of another possible future implementation. If only one exists, not sure if generalizing helps here (although it's cool a lot). As a note, the idea behind the enum in `InterestRate` is that you can combine `Fixed` & `Variable` (or others) at the same time. So any implementation of `Rate` trait would need that "enum" or similar.
- In case of doing it, because it would not affect to the storage, could we leave it for a future PR after the next release?

lemunozm:
> we are not forced to do a migration once we want montly accrual and so on.

Why are we not forced to do it with the trait?

mustermeiszer:
> Why are we not forced to do it with the trait?

As we would also adapt right now what we store in storage under `RateDetails`. If not, then must migrate of course.

mustermeiszer:
> Does it make sense another implementation for the Rate trait, rather than the own InterestRate, right now I'm not able to think of another possible future implementation. If only one exists, not sure if generalizing helps here (although it's cool a lot). As a note, the idea behind the enum in InterestRate is that you can combine Fixed & Variable (or others) at the same time. So any implementation of Rate trait would need that "enum" or similar.

Yes, every impl would need that. I was more thinking of already preparing interest accrual for `foss3`. But again it is not mandatory. It just felt like using an "incomplete" enum for the interface was a fix rather than a solution. 

I think adapting this to a trait will require a migration as we can no longer rely on `interest_rate_per_sec`. The same goes for the current change. 

As soon as we want another compounding schedule, we will need to migrate to something that reflects that rate and is able to determine how many compounding steps happened between `last_updated` and `now`

lemunozm:
> It just felt like using an "incomplete" enum for the interface was a fix rather than a solution.

Yes I the idea from Jeroen was to prepare this for future work, adding this now before because of the release.

> As soon as we want another compounding schedule, we will need to migrate to something that reflects that rate and is able to determine how many compounding steps happened between `last_updated` and `now`

Although it would require migration in the `pallet-interest-accrual`, having this type yet in `pallet-loans` would avoid migrating it later there if we already have that "enum" in the type

lemunozm:
Agree then in merging this now?

mustermeiszer (approved):

PR #1444, opened 2023-07-06
Fix individual crate compilation issues & add script for running all tests

### LINKED ISSUE #1441: Script to check and test all crates separately
### Description

When testing the entire workspace at once, rust chooses the crates with more features on them. This means that individually our crates can be broken, but because we are compiling all together it just works. 

We can avoid this situation and be able to work faster compiling and testing each crate separately with any combination of features as follows:

```sh
cargo test -p <my-crate> -F runtime-benchmarks
```

Also, having them isolated compiling helps to find general compilation issues due to features in the future.

### Propose
- Leave all crates without any compilation issues.
- Add a job to check automatically that all crates work separately with all feature combinations. This can take time, so I would propose creating a nightly job to do this.

lemunozm:
# Description

Fixes #1441 

## Changes and Descriptions

- Fixed all compilation issues running all crates isolated 🎉. Now you can run without errors each crate as:
  ```sh
  cargo test -p <crate> -F <feature1,feature2,...>
  ``` 
- Added script to run all workspace crate tests with all feature combinations.
 
  To run the script for all crates:
  ```sh
  ./scripts/tests.sh
  ```

  To run the script for all crates from the selected one in the workspace list (used mainly to recover the script execution point after a crate failure):
  ```sh
  ./scripts/tests.sh <from-crate>
  ```


runtime/common/Cargo.toml
```diff
@@ -40,6 +40,11 @@ pallet-evm = { git = "https://github.com/PureStake/frontier", default-features =
 pallet-evm-chain-id = { git = "https://github.com/PureStake/frontier", default-features = false, branch = "moonbeam-polkadot-v0.9.38" }
 pallet-evm-precompile-dispatch = { git = "https://github.com/PureStake/frontier", default-features = false, branch = "moonbeam-polkadot-v0.9.38" }
 
+# There is a bug in the frontier repo that adds pallet-ethereum without a try-runtime dependency
+# for this crate which makes our compilation fail with the i_know_what_i_am_doing error.
+# It seem fixed in 0.9.39, and this dependency can be removed from this file safely.
+fp-self-contained = { git = "https://github.com/PureStake/frontier", default-features = false, branch = "moonbeam-polkadot-v0.9.38" }
+
```
lemunozm, line 47: I'm pretty sure this was the issue that happened in https://github.com/centrifuge/centrifuge-chain/pull/1406 (not tested yet), so it unlocks that PR.

pallets/migration/src/benchmarking.rs
```diff
@@ -189,7 +190,9 @@ fn inject_total_issuance() {
 }
 
 fn inject_system_accounts() {
-	let accounts = test_data::system_account::SYSTEM_ACCOUNT;
+	// We need to put the array into the stack (with a vec) to avoid stack overflow
+	// here
+	let accounts = test_data::system_account::SYSTEM_ACCOUNT.to_vec();
```
lemunozm, line 195: This allows us to remove the stack limit increment in the CI scripts

lemunozm (commented):
A couple of "relevant" fixes comes with this PR:

lemunozm:
It would be great to add this script as a CI job, but it surely exceeds the maximum size available and will take considerably more time than other jobs. Any idea @gpmayorga? My suggestion is to clean artifacts after testing each crate (which will take even more time to run all crates) and create a weekend job for it.

runtime/common/Cargo.toml
NunoAlexandre: That could very well be it. But I tried using my fork of frontier fixing all the cargo features and the issue still persisted 🤔

scripts/tests.sh
```diff
@@ -0,0 +1,73 @@
...
+
+ESC="\033"
+testing_prompt="$ESC[1;36m     Testing$ESC[0m"
+go=0
+
+if [[ -z "$start_from" ]]; then
+    go=1
+fi
+
+cargo install cargo-workspaces
```
NunoAlexandre, line 36: Can we have this as part of a `setup` script? Having this here means that every time we want to run this (locally) we will have cargo checking for updates and trying to update this crate, which will slow this process even further. wdyt?

runtime/common/Cargo.toml
lemunozm: Have you added `"fp-self-contained/try-runtime"` too to your try?

scripts/tests.sh
lemunozm: Sure! Make sense. What's the `setup` script you're referring?

runtime/common/Cargo.toml
NunoAlexandre: Not sure, but I ran subalfred on the whole repo to make sure I wasn't missing any feature anywhere

scripts/tests.sh
NunoAlexandre: It doesn't quite exist atm, but we could create one or adapt the current install-toolchain to be more general-purpose and install tools we need for development

runtime/common/Cargo.toml
lemunozm: Ah, but before this PR I also run `subalfred` for all crates, and everything went "well" 😆, and things weren't well. I guess we can not trust 100% in `subalfred`. Some things are not checked.

scripts/tests.sh
lemunozm: Ok, I understand. By now, f you agree, I can print a message if the command is not found instead of fetching every time if it exists. I do not want to add things into `install_toolchain.sh` because it affects all CI job time/space. Only the required job should fetch its required tools, I think.

lemunozm: We can prepare a good script in another PR, with taplo, subalfred, cargo-workspaces, ...

NunoAlexandre: Yeah that's fine :)

NunoAlexandre (approved):
LGTM  but it would be nice to have another reviewer 👍🏻

lemunozm:
Thanks Nuno! I think other review is even required 🤣

mustermeiszer (approved):
LGTM!

pallets/pool-registry/src/mock.rs
```diff
@@ -222,6 +223,7 @@ impl<
 		_currency: <T as pallet_pool_registry::Config>::CurrencyId,
 		_max_reserve: <T as pallet_pool_registry::Config>::Balance,
 	) -> DispatchResult {
+		#[cfg(feature = "runtime-benchmarks")]
```
branan, line 226: Are we sure the only code paths for this can be thru the runtime-benchmarks feature? I'm a little concerned about having the mocks behave differently between benches and tests...

branan (approved):
One question / minor concern, but it's not a blocker. Leaving my approval so merging isn't blocked on the discussion

wischli (approved):
LGTM!

pallets/pool-registry/src/mock.rs
lemunozm: Good point! It's true that it's used in tests even when `runtime-benchmarks` is not enabled, so in that case, you're not creating a pool. Anyways, it is compiling and passing with and without `runtime-benchmarks` feature, so I suppose it should behave as expected.

Looking more deeply creating a pool seems to doesn't matter for the testing, but yes for the benchmarks because it adds a lot of weight to the call, so makes sense to only call `create_pool()` on benchmaking.

pallets/migration/src/benchmarking.rs
mikiquantum: I think we should remove this pallet all-together. We will not use it again.

mikiquantum (approved):
I would remove the migrations pallet, but in any case I am not blocking this pr 👍

pallets/migration/src/benchmarking.rs
lemunozm: Good to know!

lemunozm: https://github.com/centrifuge/centrifuge-chain/issues/1448

Issue #1360, opened 2023-05-24
Traits: Add `TimeAsSecs` trait


lemunozm:
### Description

We use the trait `UnixTime` everywhere to get the time as secs. We can use a trait wrapper over it (`TimeAsSecs`) that transforms the returned `Duration` time directly to seconds, avoiding confusion about what the time unit should be (`Moment` should be secs? millis? mins?) in some places.

Implementation draft:
```rust
trait TimeAsSecs: UnixTime {
    type Seconds: From<u64>; //u64 is the representation for Duration::as_secs()

    fn now() -> Self::Seconds {
        UnixTime::now().as_secs().into()
    }
}

// Auto implementations for any UnixTime type
impl<T: UnixTime> TimeAsSecs for T {}
```

mustermeiszer:
Thanks for creating this! Should be trivial to clean up the codebase with this.

lemunozm:
Wondering if makes sense to have a `Second` generic when the own `Duration` type from `UnixTime` is `u64`.

mustermeiszer:
Probablz not ^^

PR #1445, opened 2023-07-06
Loans: maturity extension support

### LINKED ISSUE #1405: Loans:  Fixed maturity with extension
### Description
Add an option to include a # of extension days in a fixed maturity asset, which enables extending the maturity by those # of days without going through the change guard.

The benefit is that for internally priced assets, the valuation can depend on the correct maturity date. 

- [ ] Add
```rust
enum MaturityDate {
  Fixed {
    date: Moment,
    extension: Moment,
  }
}
```
- [ ] Maturity date can be modified as long as
    ```rust
    new_maturity_date = old_maturity_date + (prev_extension_days - remaining_extension_days)
    remaining_extension_days = prev_extension_days - (new_maturity_date - old_maturity_date)`
    ```
  Basically the extension days are reduced with the same amount as the maturity date is increased.

lemunozm:
# Description

Fixes #1405

## Changes and Descriptions

- Add `extension` field to `Maturity::Fixed`
- Add new loan mutation: `MaturityExtension`


pallets/loans/src/types/mod.rs
```diff
@@ -85,25 +85,45 @@ pub enum MutationError {
...
 /// Specify the expected repayments date
 #[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeDebug, MaxEncodedLen)]
 pub enum Maturity {
 	/// Fixed point in time, in secs
-	Fixed(Moment),
+	Fixed {
+		/// Secs when maturity ends
+		date: Moment,
+		/// Date extension without special permissions
+		extension: Moment,
```
lemunozm, line 100: We definitely need better types to differentiate different time units. I think for loans, we would need a `Seconds` type with a bound of `From<Moment>`. Or use this trait instead: #1360. I think this is even more important than differentiating `Instant` from `Duration`, and more expressive for the UI to know what the unit is.

runtime/common/src/lib.rs
```diff
@@ -424,6 +424,7 @@ pub mod changes {
 				// <https://docs.google.com/spreadsheets/d/1RJ5RLobAdumXUK7k_ugxy2eDAwI5akvtuqUM2Tyn5ts>
 				LoansChangeOf::<T>::Loan(_, loan_mutation) => match loan_mutation {
 					LoanMutation::Maturity(_) => vec![week, blocked],
+					LoanMutation::MaturityExtension(_) => vec![],
```
lemunozm, line 427: We have no tests in the `runtime/common` module. It can only be checked properly by `integration-tests` (that we do not currently have)

pallets/loans/src/types/mod.rs
```diff
@@ -85,25 +85,45 @@ pub enum MutationError {
 	DiscountedCashFlowExpected,
 	/// Emits when a modification expect the loan to have an iternal pricing.
 	InternalPricingExpected,
+	/// Maturity extensions exceed max extension allowed.
+	MaturityExtendedToMuch,
```
offerijns, line 89:
```suggestion
	MaturityExtendedTooMuch,
```

pallets/loans/src/types/mod.rs
```diff
@@ -85,25 +85,45 @@ pub enum MutationError {
...
-			Maturity::Fixed(moment) => *moment > now,
+			Maturity::Fixed { date, .. } => *date > now,
+		}
+	}
+
+	pub fn extends(&mut self, value: Moment) -> Result<(), ArithmeticError> {
+		match self {
+			Maturity::Fixed { date, extension } => {
+				date.ensure_add_assign(value)?;
+				extension.ensure_sub_assign(value)
```
offerijns, line 125: This is really clean, I like it :)

pallets/loans/src/types/mod.rs
lemunozm: Added a test too to check this is error is correctly generated 👍🏻

offerijns (approved):

lemunozm:
I've rebased this already approved by Jeroen, it would be awesome if you could give a re-approve it @mustermeiszer 🙌🏻

mustermeiszer:
@lemunozm does this need a migration? Because I can not spot the migration code ^^

Just am worried we forget about it if we add it later.

pallets/loans/src/types/mod.rs
```diff
@@ -85,25 +85,45 @@ pub enum MutationError {
...
 pub enum Maturity {
 	/// Fixed point in time, in secs
-	Fixed(Moment),
+	Fixed {
+		/// Secs when maturity ends
+		date: Moment,
+		/// Extension in secs, without special permissions
+		extension: Moment,
+	},
 }
```
lemunozm, line 102: Here is the migration. But it needs migration "theoretically". Right now we can purge all `pallet-loans` storage. It only requires a migration if it is used before we can deploy these changes. The same applies to all pending PRs

lemunozm: Ah ok, I read you wrong 😆, I pointed to the code that causes the migration. No, there is no migration done.

PR #1452, opened 2023-07-12
feat: make fixed point flexible

mustermeiszer:
# Description
This is a non breaking change of our fixed point integer logic. It allows to use our logic with variable decimals. This is especially useful for places where we want to reduce the max decimals compared to 27 decimals. 

While 27 decimals are great for interest calculation in some place they are limiting to the maximum value that can be represented. I.e. roughly 1 billion in the 27 decimal  case.

## Changes and Descriptions
* change `struct Rate` to `struct FixedU128<const DIV: u128>`
* create `type Rate = FixedU128<DECIMAL_27>` 

# Checklist:


libs/types/src/fixed_point.rs
```diff
@@ -1,4 +1,4 @@
-// This file is part of Substrate.
+// This file is part of SubstFixedU128.
```
wischli, line 1:
```suggestion
// This file is part of Substrate.
```

lemunozm (commented):
I like the generalization!

A note about maintaining our `fixed_point` file: I think this can be a sucurity issue in the way that we are not merging updates in this section from Substrate. What if tomorrow Substrate fixes an issue in the fixed point multiplication or similar? How can we receive these changes and update accordinly? Is there no way to reuse the Substrate `FixedU128` type and add the rounding capabilities on top of that type? 
Adding traits as extensions or wrapping the `FixedU128` from Substrate internally should work.

I mean, I think there is two things here.
1. We need a Rate with 27 decimals for accruing interest. (This could lead to having our own type)
2. We need rounding capabilities. (This not need a new type)

So for a `Rate` with rounding capabilities with 18 decimals, I think we could use the Substrate `FixedU128` because using the `FixedPointNumberExtension` it inherents to those capabilities.

libs/types/src/fixed_point.rs
```diff
@@ -293,15 +293,16 @@ pub trait FixedPointNumberExtension: FixedPointNumber {
 	/// rounding preference. Returns None if out of bounds.
 	///
 	/// Note:  This assumes that the FP accuracy has been adjusted to match
-	/// the accuracy of the FP extended type in question (Rate in this case).
-	/// For example:
-	/// Rate::saturating_from_rational(2).checked_div_with_rounding(2,
+	/// the accuracy of the FP extended type in question (FixedU128 in this
+	/// case). For example:
+	/// FixedU128::saturating_from_rational(2).checked_div_with_rounding(2,
```
wischli, line 298: I suppose this also needs some tweaking?

wischli (changes requested):
Code changes LGTM. Let's hope all tests succeed. 

There are bunch of undesired replacements of `rate` to `FixedU128` in the comments, e.g. 
 * `Substrate` --> `SubstFixedU128`. 
 * `Saturates` --> ~SatuFixedU128s`

libs/types/src/fixed_point.rs
```diff
@@ -412,20 +418,38 @@ pub trait FixedPointNumberExtension: FixedPointNumber {
...
+
+/// A fixed point number representation with 18 decimals. Used for
+/// representing a ratio between two things.
+#[doc = "_Fixed Point 128 bits unsigned with 18 decimals precision"]
+pub type Ratio = FixedU128<DECIMALS_18>;
+
+/// A fixed point number representation with 18 decimals. Used for
+/// representing a quantity of something.
+#[doc = "_Fixed Point 128 bits unsigned with 18 decimals precision"]
+pub type Quantity = FixedU128<DECIMALS_18>;
```
wischli, line 449: Out of curiosity, where is this expected to be used?

lemunozm: When you want to express an absolute number value with decimals. `Ratio` should represent instead a proportion of a value. Both are represented in the same way and are the same.

Maybe these "aliases" could be moved to our runtime definition, because they are how we are using them.

mustermeiszer:
> A note about maintaining our fixed_point file: I think this can be a sucurity issue in the way that we are not merging updates in this section from Substrate. What if tomorrow Substrate fixes an issue in the fixed point multiplication or similar? How can we receive these changes and update accordinly? Is there no way to reuse the Substrate FixedU128 type and add the rounding capabilities on top of that type?

I do not think this will be an issue. We are using the core methods which are not provided by the trait but by the arithmetics of Substrate. If they fix something there, we will receive the change.

The problem with rounding is not solvable by that as `checked_from_rational` and `checked_mul_int` are the core problems here and we must override them to get the right stuff.

libs/types/src/fixed_point.rs
mustermeiszer: I am not keen on moving them now as this means adapting all imports of `Rate`. Lets do in a follow up if wanted.

libs/types/src/fixed_point.rs
mustermeiszer: Why?

libs/types/src/fixed_point.rs
```diff
@@ -293,15 +293,16 @@ pub trait FixedPointNumberExtension: FixedPointNumber {
...
-	/// Whereas Rate::saturating_from_rational(2).checked_div_with_rounding(2 *
-	/// Rate::accuracy)would be equivalent to               2 * Rate::accuracy *
-	/// (Rate::accuracy / 2 * Rate::accuracy)               Which would be 1 *
-	/// Rate::accuracy
+	/// FixedU128::accuracy) * (FixedU128::accuracy / 2) instead of 2 * 1/2
+	/// Whereas FixedU128::saturating_from_rational(2).
+	/// checked_div_with_rounding(2 * FixedU128::accuracy)would be equivalent to
+	/// 2 * FixedU128::accuracy * (FixedU128::accuracy / 2 *
+	/// FixedU128::accuracy)               Which would be 1 *
+	/// FixedU128::accuracy
```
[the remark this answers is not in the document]
mustermeiszer: No.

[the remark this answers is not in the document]
mustermeiszer: Ah

mustermeiszer:
@wischli and @lemunozm udpated all

lemunozm:
> The problem with rounding is not solvable by that as checked_from_rational and checked_mul_int are the core problems here and we must override them to get the right stuff.

And we could not create a `FixedU128(sp_runtime::FixedU128)` that implements `FixedPointNumber` calling the inner fixed rate except for `checked_from_rational` and `checked_mul_int` methods where we add our custom logic?

wischli (dismissed):
Thanks, LGTM!

mustermeiszer:
> And we could not create a FixedU128(sp_runtime::FixedU128) that implements FixedPointNumber calling the inner fixed rate except for checked_from_rational and checked_mul_int methods where we add our custom logic?

Do not see the advantage. We are essentialy doing just that ^^

lemunozm:
The advantage I see is we are less error-prone in our code. Instead of adding simple calculations by us, we call the underlying type to do it. But it's true that the layer is quite  quite slim 👍🏻, so maybe not needed. Ideally should be Substrate who maintains all this 😞

lemunozm (dismissed):
LGTM!

PR #1455, opened 2023-07-14
Loans: borrow & repay method using pricing-based principal

lemunozm:
# Description

Jeroen's Slack introduction to this feature is [here](https://centrifugedao.slack.com/archives/C05DX0RH1T3/p1688128534804769)

Basically, we want to write the principal in different ways depending on the pricing method. As a schema, instead of always given a `Balance` as principal in `borrow()` and `repay()` methods, we want:
```rust
enum Principal {
    Internal {
        amount: Balance
    },
    External {
        quantity: Rate,
        settlement_price: Balance,
    }
}
```

The core idea is that for external pricing, the amount that is transferred to the pool is `quantity * settlement_price`, and the `outstanding_quantity` is reduced by `quantity`.

So, with this schema:
- Amount transferred: `quantity * settlement_price`
- Amount used for calculating interest: `quantity * notional`
- Amount used for valuating the loan: `quantity * current_price`

## Changes
- [x] Implementation
- [x] Adapt testing
- [x] Adapt benchmarks

### DIFF
--- a/pallets/loans/src/entities/pricing/external.rs
+++ b/pallets/loans/src/entities/pricing/external.rs
@@ -26,6 +26,20 @@ pub struct ExternalAmount<T: Config> {
 }
 
 impl<T: Config> ExternalAmount<T> {
+	pub fn new(quantity: T::Rate, price: T::Balance) -> Self {
+		Self {
+			quantity,
+			settlement_price: price,
+		}
+	}
+
+	pub fn empty() -> Self {
+		Self {
+			quantity: T::Rate::zero(),
+			settlement_price: T::Balance::zero(),
+		}
+	}
+
 	pub fn balance(&self) -> Result<T::Balance, ArithmeticError> {
 		self.quantity.ensure_mul_int(self.settlement_price)
 	}
@@ -59,7 +73,7 @@ impl<T: Config> ExternalPricing<T> {
 	pub fn validate(&self) -> DispatchResult {
 		if let MaxBorrowAmount::Quantity(quantity) = self.max_borrow_amount {
 			ensure!(
-				quantity.frac().is_zero() && quantity > T::Rate::zero(),
+				quantity.frac().is_zero() && quantity >= T::Rate::zero(),
 				Error::<T>::AmountNotNaturalNumber
 			)
 		}
@@ -165,7 +179,7 @@ impl<T: Config> ExternalActivePricing<T> {
 
 		let interest_adj = quantity_adj.try_map(|quantity| -> Result<_, DispatchError> {
 			ensure!(
-				quantity.frac().is_zero() && quantity > T::Rate::zero(),
+				quantity.frac().is_zero() && quantity >= T::Rate::zero(),
 				Error::<T>::AmountNotNaturalNumber
 			);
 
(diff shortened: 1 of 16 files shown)


lemunozm:
@mustermeiszer Because we are in a hurry, the main implementation can already be reviewed (it's finished) while I adapt/add tests. 🙌🏻

lemunozm:
Now it's ready for a final review 👐🏻

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -10,14 +10,41 @@ use frame_support::{self, ensure, RuntimeDebug, RuntimeDebugNoBound};
...
 use crate::{
 	entities::interest::ActiveInterestRate,
 	pallet::{Config, Error, PriceOf},
 };
 
+#[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeDebugNoBound, MaxEncodedLen)]
+#[scale_info(skip_type_params(T))]
+pub struct ExternalAmount<T: Config> {
+	pub quantity: T::Rate,
+	pub settlement_price: T::Balance,
```
mustermeiszer, line 25: Nit pick. Not sure if `settlement_price` is the right tearm for when repaying? cc @denniswell

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -127,42 +150,42 @@ impl<T: Config> ExternalActivePricing<T> {
...
-			let price = self.current_price()?;
-
-			let quantity = T::Rate::ensure_from_rational(principal, price)?;
+		self.outstanding_quantity = quantity_adj.ensure_add(self.outstanding_quantity)?;
 
+		let interest_adj = quantity_adj.try_map(|quantity| -> Result<_, DispatchError> {
 			ensure!(
-				quantity.frac().is_zero(),
-				Error::<T>::AmountNotMultipleOfPrice
+				quantity.frac().is_zero() && quantity >= T::Rate::zero(),
```
mustermeiszer, line 182: Again, why the change to allow zero

mustermeiszer (approved):
I am approving as I spot nothing bad. Just would want to get feedback from Dennis regarding naming and am curious for the relaxation of the `ensure!`-clauses.

mustermeiszer:
Code change of course looks good and clean! :-)

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -46,7 +73,7 @@ impl<T: Config> ExternalPricing<T> {
 	pub fn validate(&self) -> DispatchResult {
 		if let MaxBorrowAmount::Quantity(quantity) = self.max_borrow_amount {
 			ensure!(
-				quantity.frac().is_zero() && quantity > T::Rate::zero(),
+				quantity.frac().is_zero() && quantity >= T::Rate::zero(),
```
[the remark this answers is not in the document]
lemunozm: It's totally fine to repay with 0 of quantity if you want, for example, to pay only interest.

Or you can create a loan that can not be borrowed only to repay later unchecked amounts.

pallets/loans/src/entities/pricing/external.rs
lemunozm: It's totally fine to repay with 0 of quantity if you want, for example, to pay only interest.

pallets/loans/src/entities/pricing/external.rs
lemunozm: As a note: `ExternalAmount` is used for both: borrow & repay. So if changed there, it's changed for both.

lemunozm: From my understanding of what this means, I think the name fits with its purpose.

lemunozm:
Merged, if finally disagree with something. I'll add a PR with that tech debt

Issue #1174, opened 2023-01-17
RtApi - Portfolio Composition of a pool


mustermeiszer:
In the near future a pool -- if being an index -- will hold value in other pools and maybe also have loans being originated from it. In order to have a non-changing API for the Apps team we need a runtime API for this. 

### Tasks
- [ ] Sync with Apps team

### Draft 
```rust
fn portfolio(pool_id: PoolId) -> Vec<Holding> {
   ...
}

pub enum Holding {
    Loan {
        // TODO: Define with Apps what is needed here
    },
    Tranche {
          token: Currency,
          amount: Balance, 
          value: Balance
    }
}
```

mustermeiszer:
I think this is really misleading. At a first thought at least. I have to think about it in more detail. 

Why do you think this is helpful?

lemunozm:
This feature is done for the loans part. Check https://github.com/centrifuge/centrifuge-chain/pull/1457 description for a discussion about why keep this separate.

mustermeiszer:
I will close this here. Once we have actual tranche holdings we should revisit

PR #1457, opened 2023-07-17
Loans: Add Runtime API

lemunozm:
# Description

Fix partially #1174 for loans. I have not wanted to add the pool part because I'm not pretty involved in the index pool feature, and things have changed a bit from the last specs regarding this. 

Also, not sure what we gain by merging pool & loan things in the same RuntimeAPI call. The user can make two API requests, and we maintain things isolated. If making two calls comes with some performance cost (that, as a first view, doesn't, because both calls can be done concurrently), then I think we can add an API wrapper that performs both API requests, keeping things isolated.

## Changes and Descriptions
- [x] Add an RPC to get `ActiveLoan` data and computed values as `present_value` and `interest_accrued`. 
- [x] ~~Integration tests~~ => Can be done in another PR => #1458


src/rpc/loans.rs
```diff
@@ -0,0 +1,94 @@
...
+	#[method(name = "loans_portfolioLoan")]
+	fn portfolio_loan(
+		&self,
+		pool_id: PoolId,
+		loan_id: LoanId,
+		at: Option<BlockHash>,
+	) -> RpcResult<ActiveLoanInfo<T>>;
+}
+
+pub struct Loans<C, P> {
```
cdamian, line 31: nit - rename second generic param to `Loans<C, Block>`?

cdamian (dismissed):

pallets/loans/src/entities/loans.rs
```diff
@@ -489,3 +491,47 @@ impl<T: Config> ActiveLoan<T> {
 		self.schedule.maturity = crate::types::Maturity::fixed(duration);
 	}
 }
+
+/// Data containing an active loan with extra computed.
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebugNoBound, TypeInfo, MaxEncodedLen)]
+#[scale_info(skip_type_params(T))]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "std", serde(bound = ""))]
+pub struct ActiveLoanInfo<T: Config> {
```
lemunozm, line 500: This is a bit tricky and probably useful for futures runtime APIs.
The line with `bound = ""` is mandatory for cases where you have both, the `scale_info` and `derive(serde)` lines. Serde defaults to adding `Serialize` and `Deserialize` bounds to `T`,  but T, in this case, does not need to be serializable, it's only used to fetch information from associated types. With `bounds = ""`, we remove those implicit bounds from serde.

cdamian:
@lemunozm Can you also add an integration test for this similar to the one that we have for rewards?

lemunozm:
Yeah, I wanted to do it, but looking into it, I've realized it would be quite time-consuming (creating a pool, creating assets, registering permissions, creating investors, investing in the pool, borrowing from loans to make the "active" loan, ...) just for being able to make the runtime API call, and right now I need to put efforts in order task for the release 😓.

Also, I would need to create integration tests for all loans related things. Right now, it's well-tested at pallet level but not at runtime level. Issue to track this work here: #1458

pallets/loans/src/entities/loans.rs
wischli: Thanks for clarifying this. So this is the `serde` equivalent of `scale_info(skip_type_params(T)` more or less?

pallets/loans/Cargo.toml
```diff
@@ -63,6 +64,7 @@ std = [
   "sp-io/std",
   "strum/std",
   "orml-traits/std",
+  "serde/std",
```
wischli, line 67: Since it's optional, shouldn't we apply `?`
```suggestion
  "serde?/std",
```

libs/traits/Cargo.toml
```diff
@@ -43,6 +44,7 @@ std = [
   "sp-std/std",
   "cfg-primitives/std",
   "scale-info/std",
+  "serde/std",
```
wischli, line 47: Since it's optional, shouldn't we apply `?`
```suggestion
  "serde?/std",
```

wischli (commented):
LGTM apart from my minor nit and the supposedly missing `#[serde(rename_all = "camelCase")]` for all call input arguments based on [this comment](https://github.com/polkadot-js/api/issues/5529#issuecomment-1465271895).

pallets/loans/Cargo.toml
lemunozm: Good one!

pallets/loans/src/entities/loans.rs
lemunozm: Yeah! More or less. Also, if there were other bounds, and you do not want to "reset them" you should add `bound = "K: MyOtherBound"`. That way you've reset the `Serialize` bound to `T` and `K` but not the `MyOtherBound` to `K`.

libs/traits/Cargo.toml
wischli: Wondering why adding this broke tests?

lemunozm: I think:
- `serde?/std` adds `serde/std` in case `serde` is enable
- `serde/srd` enables `serde` and adds `serde/std`

lemunozm:
Much much simpler PR after removing the annoying RPC part. Thanks @wischli!

wischli (approved):
All hail the runtime API!

Issue #1461, opened 2023-07-20
Loans: prepare Loans for release


lemunozm:
### Description

Some tasks are required to be able to release loans correctly:
- Check Centrifuge & Altair chain storages and add a migration to nuke `pallet-loans`
- Check Centrifuge & Altair chain storages and add a migration to nuke `interest-accrual`
- Add missing methods to `ProxyType` in runtimes: https://github.com/centrifuge/centrifuge-chain/pull/1463

PR #1462, opened 2023-07-21
Algol runtime 0.10.28

branan:


lemunozm:
Wondering if, before this, we should do https://github.com/centrifuge/centrifuge-chain/issues/1461

mikiquantum:
I believe we need to get our demo environment up. We can always do another upgrade with the tasks mentioned in https://github.com/centrifuge/centrifuge-chain/issues/1461 . 
@lemunozm Is your comment blocking?

mikiquantum (approved):

lemunozm:
If it's used only to get our demo environment up I think migrations are not needed, but maybe this is needed: https://github.com/centrifuge/centrifuge-chain/pull/1463 (which can be already merged)

branan:
Basic functionality should still work without #1463, right? If so I think we should still just push this runtime as it's already tagged, just to unblock the apps team. 

/cc @sophialittlejohn on if the missing write-off and loan modification bits will be a blocker to you.

branan:
Given this is already tagged and built, I'm gonna hit the merge button - if it ends up that we need to include #1463 we'll just have to tag an updated version

PR #1465, opened 2023-07-24
Loans: transfer_debt implementation

### LINKED ISSUE #1378: Loans: add transfer_debt() extrinsic
### Description

The main description of this issue can be found in the [slack thread](https://kflabs.slack.com/archives/C050E3ZN2JU/p1685518818663599).

As an overview, the new extrinsic will transfer debt from one loan to another, repaying the first loan and borrowing the second one with the same amount:

```rust
fn transfer_debt(pool_id, from_loan_id, to_loan_id, amount_taken_from_loan_debt);
```

lemunozm:
# Description

Fixes #1378 

### Changes

- [x] Base implementation
- [x] Tests
- [x] Benchmarks


pallets/loans/src/lib.rs
```diff
@@ -860,30 +885,6 @@ pub mod pallet {
 				.map_err(|_| Error::<T>::NoLoanChangeId.into())
 		}
 
-		fn update_portfolio_valuation_for_pool(
```
lemunozm, line 863: No changes here simply moved in the file

pallets/loans/src/lib.rs
```diff
@@ -781,13 +766,53 @@ pub mod pallet {
...
+			to_loan_id: T::LoanId,
+			repaid_amount: RepaidPricingAmount<T>,
+			borrow_amount: PricingAmount<T>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			ensure!(
+				from_loan_id != to_loan_id,
+				Error::<T>::TransferDebtToSameLoan
+			);
```
lemunozm, line 790: Initially, I thought it was needed, but I think it's not and can be removed.
cc @offerijns

lemunozm: I mean, a borrower can always borrow from the same loan to repay it. So, why forbid it here?

pallets/loans/src/lib.rs
```diff
@@ -781,13 +774,61 @@ pub mod pallet {
...
+						Error::<T>::TransferDebtAmountMismatched
+					)
+				}
+				PricingAmount::External(_external) => {
+					// TODO (1): handle it once slippage is added.
+					// TODO (2): if quantity is measured as a rate,
+					// then instead of using slippage we can use quantity with
+					// decimals.
+				}
+			}
```
lemunozm, line 815: If we have the implicit invariant that says that repayment and borrowing ideally should withdraw/deposit the same exact amount from the pool, then it works fine for Internal pricing but would only work fine for External pricing if we chose case 2: Decimals with quantity. 

In that case, we should add that feature before merging this not to open a security hole allowing the transfer of debt amounts that are not equivalent to doing the extrinsics separately: `repay() + borrow()`.

pallets/loans/src/lib.rs
```diff
@@ -781,13 +774,61 @@ pub mod pallet {
...
+
+			let _count = Self::borrow_action(&who, pool_id, to_loan_id, &borrow_amount)?;
+
+			Self::deposit_event(Event::<T>::TransferDebt {
+				pool_id,
+				from_loan_id,
+				to_loan_id,
+				amount: repaid_amount.repaid_amount()?.total()?,
```
lemunozm, line 823: Supposing we fix the previous issue, then, no matter if we choose here `repaid_amount` or `borrow_amount`, the total amount will always be the same.

pallets/loans/src/lib.rs
lemunozm: i.e. 
borrower does:
- repay with (`quantity * settlement_price_1`)
- borrow with (`quantity * settlement_price_2`)

Then debt has been transferred but they have more balance in their account if `settlement_price_2 > settlement_price_1`.

Instead, if the borrower does:
- transfer debt (repay: `quantity * settlement_price_1`, borrow: `quantity * settlement_price_2`)

Then debt has been transferred but their account balance remains the same when not.

lemunozm: cc @offerijns @denniswell

lemunozm:
This is ready for a final review

pallets/loans/src/lib.rs
lemunozm: @hieronx I've seen you've resolved this, does it mean you agree to remove it?

pallets/loans/src/lib.rs
lemunozm: Solved once we use decimals for quantities

pallets/loans/src/tests/transfer_debt.rs
```diff
@@ -0,0 +1,425 @@
...
+				loan_2,
+				repay.clone(),
+				borrow.clone(),
+			))
+		}
+	});
+}
+
+#[test]
+fn with_wrong_borrower() {
```
hieronx, line 48: Maybe would be good to check this in both directions, that the origin needs to be a borrower of both the `from` and `to` loans

pallets/loans/src/lib.rs
```diff
@@ -770,6 +765,174 @@ pub mod pallet {
...
+		/// The repaid and borrow amount must match.
+		#[pallet::weight(T::WeightInfo::propose_transfer_debt(T::MaxActiveLoansPerPool::get()))]
+		#[pallet::call_index(11)]
+		pub fn propose_transfer_debt(
+			origin: OriginFor<T>,
+			pool_id: T::PoolId,
+			from_loan_id: T::LoanId,
+			to_loan_id: T::LoanId,
+			repaid_amount: RepaidPricingAmount<T>,
+			borrow_amount: PricingAmount<T>,
```
hieronx, line 782: I think this naming is confusing. Generally pricing refers to setting loan parameters (e.g. updating the interest rate).

What about `BorrowInput` and `RepayInput`?

pallets/loans/src/tests/transfer_debt.rs
lemunozm: If I understand you well, I think it works just like you say now, because both loans are from another borrower different than the origin used in `propose_transfer_debt()`. But I think it could be interesting to check what happens if only one of the loans is different from the origin used. 

Added a couple of extra tests regarding this situation.

pallets/loans/src/lib.rs
lemunozm: These "Pricing" names come because the `pricing` module defines them as follows:
```rust
pub enum PricingAmount<T: Config> {
	Internal(T::Balance),
	External(external::ExternalAmount<T>),
}
```
So, it carries the "pricing mode" information in the struct.

But it's also true that these structs are also used for introducing values into the system and are never stored, so I think `Input` is a good name for those too 👍🏻

hieronx: But the information included in that `PricingAmount` struct is not pricing information. It's just that the borrow/repay input is dependent on the pricing type. I would even say the `PricingAmount` struct semantically should not be in the `pricing.rs` file for that reason.

lemunozm: Mmm... good point. I'll rethink if it should be moved to a better place. And definitely `Input` suffix fits here 👍🏻

lemunozm: Ok, looking deeply into it, I think the problem here is that `RepaidPricingAmount` contains into it a `PricingAmount`, so we can not change them to `BorrowInput/RepayInput` names. That would mean that for composing a `RepaidInput`, the user needs to create a `BorrowInput`.

I like the `Input` suffix, so I'd opt for these new names:

- `PrincipalInput`
- `RepaidInput` (which is composed using a `PricipalInput`).

And those types can be moved to a new module `/entities/input.rs`

WDYT?

hieronx: Yeah that's a great solution 💯

hieronx (approved):
Nice! Maybe good if @mustermeiszer also reviews but looks good to me

pallets/loans/src/lib.rs
```diff
@@ -770,6 +766,174 @@ pub mod pallet {
...
+			from_loan_id: T::LoanId,
+			to_loan_id: T::LoanId,
+			repaid_amount: RepaidInput<T>,
+			borrow_amount: PrincipalInput<T>,
+			permissionless: bool,
+		) -> Result<(T::Balance, u32), DispatchError> {
+			ensure!(
+				from_loan_id != to_loan_id,
+				Error::<T>::TransferDebtToSameLoan
+			);
```
lemunozm, line 912: Just to double-check @hieronx, could it make sense to allow transferring debt to the same loan? i.e. for paying the interest with the amount borrowed from the principal.

Maybe I'm being too strict with this restriction

mustermeiszer (approved):
Regarding the logic here. 

Is it okay if
* `transfer_debt_action` returns `Ok(())` during proposal - borrower has enough funds to repay
* `transfer_debt_action` returns `Err(..)` during execution - borrower has NOT enough funds to repay

Otherwise it looks good.

lemunozm:
I would say yes.

Although the borrower account is not considered here, the borrower repays with the amount they borrow, so the account funds are not moved. But it's also true that the borrower can propose a `transfer_debt`, later do `borrow()`/`repay()`, and when the `transfer_debt` is applied, it will fail because maybe the loan is already fully repaid or even closed.

mustermeiszer:
It can also fail if in the meantime the pool itself has not enough funds to borrow. So it can fail for various reasons. But I can not see any misbehaviour that could occur if this action has been released but never executed. Do you see some?

lemunozm:
> It can also fail if in the meantime the pool itself has not enough funds to borrow

Actually, this can not happen, because `trasnfer_debt()` does not transfer funds to/from the pool. It's one of the main reasons to implement this.

> Do you see some?

If not executed, it does not change/mutate anything, so it takes no effect. Nevertheless from a game theory point of view, some investors could have redeemed/invested money because they knew about some "transfer_debt" actions that were supposed to occur but did not. So false positives could affect how investors behave, but not sure how really it affects the system or if it could be an attack vector. By now, I think there's no issue with this

mustermeiszer (approved):
Re-approve

PR #1466, opened 2023-07-26
Runtime: Remove week delay for fast runtimes in RuntimeChange

lemunozm:
# Description

Currently, Apps need to wait 1 entire week to test some `RuntimeChange`s. This is undesirable.
If `development` node is compiled with `fast-runtime`, this limitation is removed and can be done instantly.

## Changes and Descriptions

- Enable `fast-runtime` for `runtime-common`
- Disable week delay for `RuntimeChange`.

### DIFF
--- a/runtime/common/Cargo.toml
+++ b/runtime/common/Cargo.toml
@@ -144,3 +144,6 @@ try-runtime = [
   "pallet-evm-chain-id/try-runtime",
   "fp-self-contained/try-runtime",
 ]
+
+# Set timing constants (e.g. session period) to faster versions to speed up testing.
+fast-runtime = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -267,6 +267,7 @@ try-runtime = [
 ]
 
 fast-runtime = [
+  "runtime-common/fast-runtime",
   "altair-runtime/fast-runtime",
   "centrifuge-runtime/fast-runtime",
   "development-runtime/fast-runtime",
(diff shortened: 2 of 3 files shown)


runtime/common/Cargo.toml
```diff
@@ -144,3 +144,6 @@ try-runtime = [
   "pallet-evm-chain-id/try-runtime",
   "fp-self-contained/try-runtime",
 ]
+
+# Set timing constants (e.g. session period) to faster versions to speed up testing.
+fast-runtime = []
```
lemunozm, line 149: Not sure TBH, when `fast-runtime` and when `testnet-runtime` is used. Seems like this should be under `fast-runtime`. I understand real production nodes will not use `fast-runtime`. 

Why both features are not combined into just one?

wischli: > Why both features are not combined into just one?

I think the main reason is lazyness from our side as the `fast-runtime` is "inherited" from Substrate. It comes with macro support, i.e. `prod_or_fast!` for setting parameter types. 

IMO, we should merge both features. Additionally, we have been building `fast-runtime` Altair WASMs over the past months even though there is NO use for them, we only need it for Centrifuge runtime as our Catalyst test network runs with it. 

So most ideally, we should only have a `testnet-runtime` feature such that Altair@testnet-runtime is the Algol WASM and Centrigue@testnet-runtime is the Catalyst WASM.

lemunozm: Ok, I agree with you! Thanks @wischli for all your explanations regarding this topic

lemunozm:
Closed in favor of #1467 which solves this particular case without touching features/scripts/deployment. Anyways, we should think about this for future cases.

PR #1467, opened 2023-07-26
Runtime: add `RuntimeChange` version for fast runtimes

lemunozm:
# Description

Another solution for #1466 that it does not imply changes in deployment/scripts.

## Changes and Descriptions

- Add a `RuntimeChange` wrapper that modifies delay times for "fast" runtime versions. Each runtime can choose between using `RuntimeChange` or `fast::RuntimeChange` depending on their configuration.

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -472,4 +472,44 @@ pub mod changes {
 			Ok(loan_change)
 		}
 	}
+
+	pub mod fast {
+		use pallet_pool_system::pool_types::changes::Requirement;
+
+		use super::*;
+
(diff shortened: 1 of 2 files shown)


lemunozm:
@onnovisser Once merged this commit and deployed to Dev, you should be able to configure policies waiting 60 secs. Tell me if you have the same issue when testing in Argol, and I will fix it there. By now, it's only fixed in development.

wischli (approved):
LGTM! Will probably need to be applied to Algol as well. But then we could look into merging `fast-runtime` and `testnet-runtime` all at once.

lemunozm:
Perfect @wischli 👍🏻

PR #1469, opened 2023-07-27
ScheduleUpgrade connectors message

AStox:
# Description

Adds the ScheduleRely message, intended for governance to rely and execute spells on EVM domain gateway contracts

## Changes and Descriptions
- Adds a `ScheduleRely(user: Address)` message.
- Adds a `schedule_rely` function to lib.rs

# Checklist:


pallets/connectors/src/lib.rs
```diff
@@ -688,6 +688,19 @@ pub mod pallet {
 		}
 	}
 
+	puv fn schedule_rely(
```
NunoAlexandre, line 691:
```suggestion
	pub fn schedule_rely(
```

NunoAlexandre (commented):
One typo that's breaking the build (if I don't miss anything else) and apart from that we will then also need to run `cargo fmt` and `taplo fmt`.

Regarding the test, once the build is fixed the test will fail since it will try to compare the resulting hex the message encoded to against `""`; we can then use that hex-encoded call and replace the `""` with that.

offerijns:
We renamed this to `AddAdmin` on the Solidity side, so we should do the same here.

AStox:
Actually we mostly (but not fully) renamed this set of features in solidity to "scheduling relies". I've just pushed a [PR](https://github.com/centrifuge/connectors/pull/73) to rename the rest of the `addAdmin` mentions in solidity to scheduleRely @offerijns

pallets/connectors/src/lib.rs
```diff
@@ -686,6 +686,20 @@ pub mod pallet {
 
 			Ok(())
 		}
+		#[pallet::weight(10_000)]
+		#[pallet::call_index(10)]
+		pub fn schedule_rely(
+			origin: OriginFor<T>,
+			usr: DomainAddress,
```
NunoAlexandre, line 693: this looks correct! :) can we tho rename `usr` to `user`? 🙏

pallets/connectors/src/lib.rs
```diff
@@ -686,6 +686,20 @@ pub mod pallet {
...
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+	
+			T::OutboundQueue::submit(
```
NunoAlexandre, line 697: and here we should ensure that the the `DomainAddress` is a `DomainAddress::EVM`, i.e, this should fail with an error when it's `DomainAddress::Centrifuge` since we don't want to be sending this message to the chain itself.

pallets/connectors/src/message.rs
```diff
@@ -290,8 +290,8 @@ where
 	/// Schedules an EVM address to become rely-able by the gateway. Intended to be used via governance to execute EVM spells.
 	///
 	/// Directionality: Centrifuge -> EVM Domain.
-	ScheduleRely {
-		usr: Address,
+	ScheduleUpgrade {
+		user: Address,
```
NunoAlexandre, line 294: maybe given that this not actually a `user` address afterall but a spell contract address, we could make name this either `spell` or `contract`. wdyt?

AStox: Yah good point. Contract makes the most sense I think

pallets/connectors/src/lib.rs
```diff
@@ -706,6 +706,25 @@ pub mod pallet {
...
+					T::OutboundQueue::submit(
+						who,
+						contract.domain(),
+						Message::ScheduleUpgrade {
+							contract: contract.address(),
+						},
+					)
+				}
+				_ => Err(DispatchError::Other("Not an EVM domain address")),
+			}
```
NunoAlexandre, line 726: This is not wrong but a more idiomatic way to implement this is to have an `ensures!(` block where we would ensure the `contract` domain is not `Centrifuge` and fail otherwise with `Error::<T>::InvalidDomain`. Something like:

```suggestion
			let who = ensure_signed(origin)?;
			ensure!(contract.domain() != Domain::Centrifuge, Error::<T>::InvalidDomain);
      
			T::OutboundQueue::submit(
			  who,
			  contract.domain(),
			  Message::ScheduleUpgrade {
			    contract: contract.address(),
			  },
			)
```

pallets/connectors/src/lib.rs
```diff
@@ -706,6 +706,25 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		#[pallet::weight(10_000)]
+		#[pallet::call_index(10)]
+		pub fn schedule_upgrade(origin: OriginFor<T>, contract: DomainAddress) -> DispatchResult {
+			match contract {
+				DomainAddress::EVM(_, _) => {
+					let who = ensure_signed(origin)?;
```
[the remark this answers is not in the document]
NunoAlexandre: Absolute good catch ☝️

pallets/liquidity-pools/src/lib.rs
```diff
@@ -693,20 +697,18 @@ pub mod pallet {
...
-						contract.domain(),
-						Message::ScheduleUpgrade {
-							contract: contract.address(),
-						},
-					)
-				}
-				_ => Err(DispatchError::Other("Not an EVM domain address")),
-			}
+			ensure_root(origin)?;
+			// get account id of soverign account
```
NunoAlexandre, line 701:
```suggestion
			// get account id of this pallet
```

NunoAlexandre: @offerijns with that said, this would mean that this pallet itself would be paying for the fees for executing this call. So we would need to first transfer sufficient CFG to the account derived from the pallet id before we could call this extrinsic

NunoAlexandre (commented):
A nit 👀

pallets/liquidity-pools/src/lib.rs
NunoAlexandre: cc @mustermeiszer

NunoAlexandre: you all forget what I said above, this extrinsic is called by the root origin, I was mixing stuff up. All good!

pallets/liquidity-pools/src/lib.rs
```diff
@@ -689,6 +692,26 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		#[pallet::weight(10_000)]
+		#[pallet::call_index(10)]
+		pub fn schedule_upgrade(origin: OriginFor<T>, contract: DomainAddress) -> DispatchResult {
```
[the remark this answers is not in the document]
NunoAlexandre: Note that `Message::ScheduleUpgrade` holds a `contract: Address` where `Address` is a 32 byte array. In light of what I said above, I am of the opinion we should make this a 20-byte array, since we don't have any plans to use this message for other domains and don't even know if 32 bytes would work for that hypotetical new use case. TL;DR make it specific for EVM since that's what this message is here to do.

If we don't, then the code suggestion above needs to be adjusted to: `Message::ScheduleUpgrade { contract: vec_to_fixed_array(contract.to_vec()) },`

cc @offerijns

NunoAlexandre (commented):
I promise we are almost there 🙏

pallets/liquidity-pools/src/lib.rs
```diff
@@ -689,6 +692,26 @@ pub mod pallet {
...
+			ensure_root(origin)?;
+			// get account id of this pallet
+			let who = T::PalletId::get().into_account_truncating();
```
mustermeiszer, line 701: If we start using system internal accounts, we maybe think about adapting the `T::OutboundQueue` implementation to use something like 

```rust
enum Account<AccountId> {
   Interal,
   External(AccountId)
}
```

and let the queue handle paying for fees, I.e. take them straight from the treasury or so. Otherwise, we must fund the individuel pallets with unknown high amounts, which will be insane overhead in the long run.

NunoAlexandre: @mustermeiszer that's also what I was wondering in a thread above but note that this extrinsic can only be called by the `Origin::Root` so it's not the pallet account that will be paying for this extrinsic, we just use the pallet-derived-account as the `who` for `OutboundQueue::submit`.

mustermeiszer: But the outbound queue makes use of the account and triggers (possibly) ethereum tx afterwards. And the EVM will use the given account for fee paying. IIRC.

NunoAlexandre: @mustermeiszer thanks for the deeper insight! I've written a basic test to cover the `EthereumXcmRouter` case and it fails due to "UnableToWithdrawFunds`, despite me funding the `LiquidityPools` pallet account, which is unexpected. https://github.com/centrifuge/centrifuge-chain/pull/1469/commits/6804dc1a502a3b7b7108c183b06a20c8a4af8b2c

In any case, we do have to look further into this since it doesn't work as is, at least not trivially (unless if I am making some weird mistake in that test)

cc @AStox

NunoAlexandre: Update: I was indeed making a mistake. It's now fixed. I was minting CFG instead of GLMR to the `LiquidityPools` account 🙉 

@mustermeiszer on your remark:

> we must fund the individuel pallets with unknown high amounts, which will be insane overhead in the long run.

Note that this is the only pallet and the only edge case where this scenario applies. And we don't plan on running `schedule_upgrade` every day, so I don't think this is a problem we actually have to solve.

NunoAlexandre: @mustermeiszer or we drop `PalletId` in favour of `TreasuryAccount` and use that as the `who` passed down to `Outbound::queue`. wdyt?

mustermeiszer: Yes, that I think is a great idea! it is anyways always democracy calling that, so the network should also pay!

NunoAlexandre: Sweet, that makes a lot of sense. @AStox sorry buddy! We can just rename the new `PalletId` type param you added to the `Config` to `type TreasuryAccount: Get<Self::AccountId>;` and then pass `T::TreasuryAccount::get()` as the `who` of `submit` 🙏

pallets/liquidity-pools/src/lib.rs
```diff
@@ -689,6 +692,22 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		#[pallet::weight(10_000)]
+		#[pallet::call_index(10)]
+		pub fn schedule_upgrade(
+			origin: OriginFor<T>,
+			evm_chain_id: EVMChainId,
```
AStox, line 700: @mustermeiszer would you be able to show me how to solve this `cannot find type `EVMChainId` in this scope` error here? I'm not exactly sure how to get this type in scope.

NunoAlexandre (commented):
Since that Adam is OOO I will polish the last corners and make sure this gets in

pallets/liquidity-pools/src/lib.rs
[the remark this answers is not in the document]
offerijns: I think making it a 20 byte array makes sense. If we later want to support other contract types, since this is the last message argument, that would anyways be a non-breaking change :)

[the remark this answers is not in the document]
NunoAlexandre: Ah right, didn't think about that :) thanks 👌

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools.rs
```diff
@@ -1150,6 +1151,52 @@ fn allow_pool_should_fail() {
...
+		assert_noop!(
+			LiquidityPools::schedule_upgrade(
+				RuntimeOrigin::signed(BOB.into()),
+				MOONBEAM_EVM_CHAIN_ID,
+				[7; 20]
+			),
+			pallet_xcm_transactor::Error::<DevelopmentRuntime>::UnableToWithdrawAsset
+		);
+
+		// The LiquidityPools pallet account needs GLRM to cover the fees of sending
```
offerijns, line 1183:
```suggestion
		// The treasury needs GLRM to cover the fees of sending
```

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools.rs
```diff
@@ -1150,6 +1151,52 @@ fn allow_pool_should_fail() {
...
+		assert_noop!(
+			LiquidityPools::schedule_upgrade(
+				RuntimeOrigin::signed(BOB.into()),
+				MOONBEAM_EVM_CHAIN_ID,
+				[7; 20]
+			),
+			BadOrigin
+		);
+
+		// Failing because Root (?) or the LiquidityPools pallet account has no funds
```
offerijns, line 1173:
```suggestion
		// Failing because the treasury has no funds
```

offerijns (dismissed):
LGTM!

mustermeiszer (dismissed):

mustermeiszer:
What will be triggered on the `Solidity` if this message is received?

mustermeiszer (dismissed):

NunoAlexandre:
> What will be triggered on the `Solidity` if this message is received?

@offerijns will explain this better but, afaik, this will schedule a contract upgrade (Solidity "spell") on the EVM side. the `contract` field of this new `ScheduleUpgrade` points to that upgrade/spell.

I guess a fair analogy would be to schedule a runtime upgrade and pass on the `BLAKE2_256` of the new wasm (if the wasm was living somewhere in the network like a contract does)

offerijns:
> I guess a fair analogy would be to schedule a runtime upgrade and pass on the BLAKE2_256 of the new wasm (if the wasm was living somewhere in the network like a contract does)

Yes basically this, except the hash being passed is not the new code version itself, but rather a hash of the runtime upgrade that is executing the changes.

mustermeiszer (dismissed):
Re-approve

NunoAlexandre:
@offerijns @mustermeiszer can we get another re-approval here? I had forgotten to include https://github.com/centrifuge/centrifuge-chain/pull/1469/commits/49f68f9c5b27809c8856304ffb051277015987ef in the last run

mustermeiszer (approved):

mustermeiszer (approved):

wischli:
It would be great if we did not bump dependencies silently as part of new features. They can introduce silent errors and thus should ideally only be packaged with Substrate updates.

PR #1471, opened 2023-07-28
Loans: Quantity as fixed point with 18 decimals

lemunozm:
# Description

The current `Rate` type used from `cfg_types` uses 27 decimals. Using an underlying 128bits number means that we only ca reach values up to `10^9`, which can not be enough for quantity purposes. `Rate` should only be used for interest rate computations. Other rates should use 18 decimals.

This PR fixes this, using a rate with 18 decimals.


thea-leake (approved):
LGTM!

thea-leake (approved):

thea-leake (approved):

PR #1472, opened 2023-07-28
connectors: Restrict `AddPool` and `AddTranche` to `PoolAdmin`

### LINKED ISSUE #1470: connectors: Restrict `AddPool` to `PoolAdmin`
> two reasons:

>  1. say coinbase wants to launch a pool on base just for their users, nothing stops a troll from deploying it elsewhere which they might not like
> 2. if we use different bridges the security of a pool is worsened if you deploy it through a less secure bridge

NunoAlexandre:
Closes #1470 

Note: we also want to introduce a fee at least for `AddPool`, that's under discussion. We can choose to open a follow up PR for that other bit or wait on this one.


wischli (approved):
LGTM! Could you update the spec as well after merging this?

PR #1475, opened 2023-08-01
connectors-gateway: Add gateway routers

cdamian:
# Description

Part 3 of - https://github.com/centrifuge/centrifuge-chain/pull/1376

The routers added in this PR will be used by the `Connectors Gateway` pallet when sending outgoing messages via EVM or XCM.

Currently there are 3 routers:
- Axelar EVM - used to call the Axelar Gateway contract which in turn calls the Connectors contract.
- Axelar XCM - used to call the Axelar Gateway contract which in turn calls the Connectors contract via Moonbeam XCM.
- Ethereum XCM - used to call the Connectors contract via Moonbeam XCM.

## Changes and Descriptions

- Add `Connectors Gateway` routers

# Checklist:


pallets/connectors-gateway/connectors-gateway-routers/Cargo.toml
```diff
@@ -0,0 +1,100 @@
+[package]
+authors = ["Centrifuge <admin@centrifuge.io>"]
+description = "Centrifuge Connectors Gateway Routers"
+edition = "2021"
+license = "LGPL-3.0"
+name = "connectors-gateway-routers"
```
NunoAlexandre, line 6: We don't need to do this now but the current directory structure for all things connectors is getting a bit out of hand. For this PR we could just rename this crate directory to become `pallets/connectors-gateway/routers/`and keep the crate name itself as is, `connectors-gateway-routers`.

pallets/connectors-gateway/connectors-gateway-routers/src/axelar_evm.rs
```diff
@@ -0,0 +1,191 @@
...
+		self.router.do_send(sender, eth_msg)
+	}
+}
+
+/// Encodes the provided message into the format required for submitting it
+/// to the Axelar contract which in turn submits it to the Connectors
+/// contract.
+pub(crate) fn get_axelar_encoded_msg(
+	serialized_msg: Vec<u8>,
+	target_chain: String,
```
NunoAlexandre, line 92: For type-safety reasons, we should `target_chain` a `EVMChain` param, to ensure we can't pass random strings to this function. Then, within the function we apply the `to_string()` conversion.

pallets/connectors-gateway/connectors-gateway-routers/src/axelar_evm.rs
```diff
@@ -0,0 +1,191 @@
...
+	//     bytes calldata payload,
+	// ) external {
+	//     emit ContractCall(
+	// 			msg.sender,
+	// 			destinationChain,
+	// 			destinationContractAddress,
+	// 			keccak256(payload),
+	// 			payload,
+	// 	   );
+	// }
```
NunoAlexandre, line 145: Was this here to aid your Rust coding and or do you still want to keep it?

pallets/connectors-gateway/connectors-gateway-routers/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+		+ pallet_xcm_transactor::Config
+		+ pallet_connectors_gateway::Config
+		+ pallet_ethereum_transaction::Config
+		+ pallet_evm::Config,
+	T::AccountId: AsRef<[u8; 32]>,
+{
+	EthereumXCM(EthereumXCMRouter<T>),
+	AxelarEVM(AxelarEVMRouter<T>),
+	AxelarXCM(AxelarXCMRouter<T>),
+}
```
NunoAlexandre, line 62: I don't think we benefit from having this enum. What we want is a type that implements the `Router` trait. I see that we use this enum so that we offer this polymorpgic api with `do_init()` and `do_send()`, but the first is still unclear to me if we even want that instead of making that a governance proposal setting that initial state or even a migration, and the latter is just a duplicate of `Router::send`.

NunoAlexandre (commented):
First Round 🏄‍♂️  A higher level review first mostly addressing the structure we are adopting here.

1. `connectors-gateway-routers` is not an actual pallet - it has no `Config`, no storage, no extrinsics, etc. We can and imo should just make this a sub directory of the `connectors-gateway` pallet.

2. Having `connectors-gateway-routers/routers/` hosting the "primitives" of the actual routers in the root dir (i.e, `connectors-gateway-routers/axelar_evm.rs`, `connectors-gateway-routers/ethereum_xcm.rs`, etc) is counter-intuitive. I suggest that we do:

```
./connectors-gateway/
  |- ./routers/
     |- ./mod.rs # contains all primitives used by the actual routers; if useful, can group them in `mod`s
     |- axelar_evm.rs
     |- axelar_xcm.rs
     |- ethereum_xcm.rs
```

4. the `do_init` pattern is something that I personally have never seen in Substate. Happy to change my mind, but as of now I'd rather drop that and write a migration instead that sets whatever state we may want to set once we are ready to do that (we know the right weights, chain ids, etc)

pallets/connectors-gateway/connectors-gateway-routers/Cargo.toml
cdamian: Agree, gonna move some of these around until I get a better structure.

cdamian:
> First Round surfing_man A higher level review first mostly addressing the structure we are adopting here.
> 
>     1. `connectors-gateway-routers` is not an actual pallet - it has no `Config`, no storage, no extrinsics, etc. We can and imo should just make this a sub directory of the `connectors-gateway` pallet.
> 
>     2. Having `connectors-gateway-routers/routers/` hosting the "primitives" of the actual routers in the root dir (i.e, `connectors-gateway-routers/axelar_evm.rs`, `connectors-gateway-routers/ethereum_xcm.rs`, etc) is counter-intuitive. I suggest that we do:
> 
> 
> ```
> ./connectors-gateway/
>   |- ./routers/
>      |- ./mod.rs # contains all primitives used by the actual routers; if useful, can group them in `mod`s
>      |- axelar_evm.rs
>      |- axelar_xcm.rs
>      |- ethereum_xcm.rs
> ```
> 
>     4. the `do_init` pattern is something that I personally have never seen in Substate. Happy to change my mind, but as of now I'd rather drop that and write a migration instead that sets whatever state we may want to set once we are ready to do that (we know the right weights, chain ids, etc)

Regarding the last point - there are some pallets with init logic based on the genesis config, the session-info pallet would be one example IIRC. 

As for the second part regarding migration/setting of required states, I think it makes sense to have it like this since we would only need one governance step for adding a router, otherwise, in the case of XCM, we would need 2 calls to set the storages + 1 call to set the router, and I don't see a good reason for that.

runtime/development/Cargo.toml
```diff
@@ -189,6 +190,7 @@ std = [
   "pallet-collective/std",
   "pallet-connectors/std",
   "pallet-connectors-gateway/std",
+  "pallet-connectors-gateway/std",
```
lemunozm, line 193: This is duplicated (probably some git issue)

runtime/development/Cargo.toml
```diff
@@ -314,6 +316,7 @@ runtime-benchmarks = [
   "pallet-bridge/runtime-benchmarks",
   "pallet-connectors/runtime-benchmarks",
   "pallet-connectors-gateway/runtime-benchmarks",
+  "pallet-connectors-gateway/runtime-benchmarks",
```
lemunozm, line 319: Same

pallets/connectors-gateway/connectors-gateway-routers/src/lib.rs
cdamian: I think that this enum will come in handy when setting a domain router - polkadotJS, for example, will present all 3 available options which IMO would make things easier.

I'm still not sold on the idea of having a separate step for setting the initial state, especially since not all routers require an initial step, we would then have to over-complicate and over-document the steps for specific routers.

pallets/connectors-gateway/connectors-gateway-routers/src/mock.rs
```diff
@@ -0,0 +1,475 @@
+use std::str::FromStr;
+
```
lemunozm, line 2: Wondering if we need a runtime to test this if it's not a pallet. Not being a pallet means that we do not need to store anything, so theoretically, these unit tests can be done without a Runtime 🤔

cdamian: Given that some of these routers actually use pallets during init/send, I think it doesn't hurt to have this.

runtime/development/Cargo.toml
cdamian: Weird, thanks for catching this!

pallets/connectors-gateway/connectors-gateway-routers/src/axelar_evm.rs
NunoAlexandre: Yeah sounds good. Maybe leave it as a more obvious comment and less like forgotten commented code or leave a link to the actual definition of that type/call in the original repo?

pallets/connectors-gateway/connectors-gateway-routers/src/lib.rs
NunoAlexandre: Ah yeah you are absolutely right, I forgot about that part 🤦🏻 and about the second bit, not a blocker for me, if you think that's the way to go let's do that

pallets/connectors-gateway/connectors-gateway-routers/src/mock.rs
lemunozm: Then, what you're testing is the correct functionality of those pallets routers use, isn't it? Could those pallets change in the future and break the runtime mock/tests?

I am asking this from the point of view of having the minimum needed to ensure (ideally) 100% coverage of unitary testing without unneeded dependencies.  (pedantic mode on)

cdamian: > Then, what you're testing is the correct functionality of those pallets routers use, isn't it? Could those pallets change in the future and break the runtime mock/tests?
Yes, the tests also rely on the underlying pallets behaving correctly, I don't think I have a way of mocking those out or am I mistaken?

lemunozm: You're right then! Sorry for the noise

pallets/connectors-gateway/connectors-gateway-routers/src/axelar_evm.rs
cdamian: OK, updated the function comment with links to both contracts that are handled here.

pallets/connectors-gateway/connectors-gateway-routers/Cargo.toml
cdamian: Changed, please let me know what you think.

NunoAlexandre (commented):
Looking better already, thanks for taking the comments in consideration 🙏 

A few more comments. I will do another round later going through the details of the messages encoding since it's important to check those outs more carefully

pallets/connectors-gateway/routers/src/routers.rs
```diff
@@ -0,0 +1,281 @@
...
+use scale_info::TypeInfo;
+use sp_core::{bounded::BoundedVec, ConstU32, H160, H256, U256};
+use sp_runtime::traits::{BlakeTwo256, Hash};
+use sp_std::{boxed::Box, marker::PhantomData, vec::Vec};
+use xcm::{
+	latest::{MultiLocation, OriginKind},
+	VersionedMultiLocation,
+};
+
+/// A generic router used for executing EVM calls.
```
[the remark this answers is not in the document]
cdamian: It seemed weird to me to have a `routers` dir inside the `routers` dir, but then again, it's the same with the `routers.rs` file =D I'll change that.

pallets/connectors-gateway/routers/src/axelar_evm.rs
```diff
@@ -0,0 +1,172 @@
...
+use sp_core::H160;
+use sp_std::{collections::btree_map::BTreeMap, marker::PhantomData, vec, vec::Vec};
+
+use crate::{
+	AccountIdOf, EVMRouter, MessageOf, AXELAR_DESTINATION_CHAIN_PARAM,
+	AXELAR_DESTINATION_CONTRACT_ADDRESS_PARAM, AXELAR_FUNCTION_NAME, AXELAR_PAYLOAD_PARAM,
+	CONNECTORS_FUNCTION_NAME, CONNECTORS_MESSAGE_PARAM,
+};
+
+/// EVMChain holds all supported EVM chains.
```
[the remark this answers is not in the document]
cdamian: Hmm, not sure how that would work given that we need these EVM chain identifiers as strings. If we do this, we might have to map those strings in a storage and use our `EVMChainId` and then just use those strings everywhere? I mainly picked this enum due to the type safety that comes with it...

pallets/connectors-gateway/routers/src/routers.rs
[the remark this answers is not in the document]
cdamian: Done.

NunoAlexandre (approved):
LGTM! We will probably find some quirks when testing this on Moonbase Alpha but since it's all in development and appears correct we are good merging this imo.

The only thing I would still say is whether there really is a need for a separate crate for these routers over having them as a sub-dir of `connectors-gateway`. Not a blocker for me!

Thanks a lot for all the work and all the positivity in taking in all the review comments 🙏

cdamian:
> LGTM! We will probably find some quirks when testing this on Moonbase Alpha but since it's all in development and appears correct we are good merging this imo.
> 
> The only thing I would still say is whether there really is a need for a separate crate for these routers over having them as a sub-dir of `connectors-gateway`. Not a blocker for me!
> 
> Thanks a lot for all the work and all the positivity in taking in all the review comments pray

@mustermeiszer came with the idea of having routers in a separate crate, and given the xcm/evm dependencies that we need there I think it makes sense to have it like that and not to bloat the actual gateway.

NunoAlexandre:
> > LGTM! We will probably find some quirks when testing this on Moonbase Alpha but since it's all in development and appears correct we are good merging this imo.
> > The only thing I would still say is whether there really is a need for a separate crate for these routers over having them as a sub-dir of `connectors-gateway`. Not a blocker for me!
> > Thanks a lot for all the work and all the positivity in taking in all the review comments pray
> 
> @mustermeiszer came with the idea of having routers in a separate crate, and given the xcm/evm dependencies that we need there I think it makes sense to have it like that and not to bloat the actual gateway.

Fair points, this is fine 👍

pallets/connectors-gateway/routers/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+	/// Sets the weight information for the provided XCM domain location, and
+	/// the fee per second for the provided fee asset location.
+	pub fn do_init(&self) -> DispatchResult {
+		pallet_xcm_transactor::Pallet::<T>::set_transact_info(
+			<T as frame_system::Config>::RuntimeOrigin::root(),
+			self.xcm_domain.location.clone(),
+			self.xcm_domain.transact_info.transact_extra_weight,
+			self.xcm_domain.transact_info.max_weight,
+			self.xcm_domain.transact_info.transact_extra_weight_signed,
+		)?;
```
mustermeiszer, line 206: Could we run into troubles if 

* `RouterA` sets these values
* time passes and we add a new `RouterB`
* `RouterB` overwrites values with a new value?

pallets/connectors-gateway/routers/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+		pallet_xcm_transactor::Pallet::<T>::transact_through_sovereign(
+			<T as frame_system::Config>::RuntimeOrigin::root(),
+			// The destination to which the message should be sent.
+			self.xcm_domain.location.clone(),
+			// The sender will pay for this transaction.
+			sender,
+			// The currency in which we want to pay fees.
+			CurrencyPayment {
+				currency: Currency::AsCurrencyId(self.xcm_domain.fee_currency.clone()),
+				fee_amount: None,
```
mustermeiszer, line 230: Why is this `None`?

mustermeiszer (commented):
Nice setup! Pretty flexible!

pallets/connectors-gateway/routers/src/lib.rs
cdamian: I don't think so. In theory, we should only have one router for a specific location, so if at some point we overwrite the router I think it makes sense to overwrite the xcm transactor info that is stored.

pallets/connectors-gateway/routers/src/lib.rs
cdamian: This was as per the original connectors implementation found here - https://github.com/centrifuge/centrifuge-chain/blob/ac7ee491cb805d99e97122a47d86c3951ce7b03b/pallets/connectors/src/lib.rs#L541

Maybe @NunoAlexandre can provide more specific info?

Issue #1409, opened 2023-06-20
Loans: add `remark` to loan transactions


offerijns:
### Description

Add `remark: IpfsHash` (where `IpfsHash = [u8; 46]`) as an extrinsic argument, which is emitted in the events but never stored, to all loan transactions. This would enable borrowers to add documentation about the transactions to these in off-chain storage (IPFS).

mustermeiszer:
@offerijns I was wondering whether we should rather provide a `Remark` pallet that takes a single `Extrinsic` or an `BoundedVec<Extrinsic>`. Then the event is emitted alongside the called extrinsic. Would allow issuers to extend this behaviour to any extrinsic/batch being called and we must not take care of refactoring loans or other pallets when changing/extending this.

offerijns:
Interesting idea! We should investigate how much work this would be to index in the subquery. By default it indexes specific events (e.g. listens to `loans.Created`), so it would need to listen to all `remark.Remarked` events and tie this back to the asset tx. But it would indeed be nice to have this implicitly for all events across all pools.

mustermeiszer:
I think the main advantage of having a slim pallet would be 
* batch is part if the extrinsic input &rarr; we can include the called extrinsic in the remark event
* we can use an enum type that not only provides ipfs hash but also string values or so

From a business perspective it would be:

* issuer wants to tie some action `A` to some documenting string/hash `Doc` &rarr; uses `Remark(Doc, A)`

lemunozm:
Is this feature required for August's release?

lemunozm:
Specs for a future fix: https://centrifuge.hackmd.io/koHrwp3URbuai__g3dJGtQ

lemunozm:
This issue does not require implementation by our side, and can be achieved using the `pallet_utilities` and `pallet_remark` (or `frame_system::remark_with_event`). More in this [slack thread](https://centrifugedao.slack.com/archives/C05DX0RH1T3/p1689323303521659).

lemunozm:
I think could be closed then

PR #1479, opened 2023-08-04
Add call-remarked pallet

lemunozm:
# Description

Proof of concept for #1409 where we do not need to copy `pallet-utility`.

[Slack explanation](https://centrifugedao.slack.com/archives/C05DX0RH1T3/p1691147844607099)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6953,6 +6953,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "pallet-call-remarked"
+version = "1.0.0"
+dependencies = [
+ "frame-support",
+ "frame-system",
+ "pallet-utility",
+ "parity-scale-codec 3.4.0",
+ "scale-info",
+ "sp-arithmetic",
+ "sp-runtime",
+ "sp-std",
+]
+
+[[package]]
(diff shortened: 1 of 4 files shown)


lemunozm:
I think even we could not need to call `batch_all`, and implement the function as:
```rust
#[pallet::weight(call.get_dispatch_info().weight)]
pub fn remark_call(origin: OriginFor<T>, call: T::RuntimeCall, remark: T::Remark) -> DispatchResultWithPostInfo {
    let info = call.dispatch(origin);
    Self::deposit_event(Event::Remark { remark });
    info
}
```

The own `call` could be a `Call::batch_all` if the user wants to tie several calls with the same remark.

lemunozm:
@mustermeiszer would it make sense to avoid using the `pallet-utility` as shown above? ^^

mustermeiszer:
Yes, that looks really good to be honest.

I would say we add the hash of the call to the event. WDYT?

mustermeiszer:
> A bit lost about where/how to obtain a hash or a call, any idea?

just use a type that is bound by `trait Hashing`.

But I think given we want to use system remark we can close this, right?

PR #1482, opened 2023-08-08
Loans: add outstanding principal & interest to runtime API

lemunozm:
# Description

~~Add `outstanding_debt` loan's property to runtime API~~
Add `outstanding_principal` and `outstanding_interest` to the runtime API.

### DIFF
--- a/pallets/loans/src/entities/loans.rs
+++ b/pallets/loans/src/entities/loans.rs
@@ -352,13 +352,13 @@ impl<T: Config> ActiveLoan<T> {
 					.total_borrowed
 					.ensure_sub(self.total_repaid.principal)?;
 
-				(inner.current_interest(principal)?, principal)
+				(inner.interest_accrued(principal)?, principal)
 			}
 			ActivePricing::External(inner) => {
 				let external_amount = amount.principal.external()?;
 				let max_repay_principal = inner.max_repay_principal(external_amount)?;
 
-				(inner.current_interest()?, max_repay_principal)
+				(inner.interest_accrued()?, max_repay_principal)
 			}
 		};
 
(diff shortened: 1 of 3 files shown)


pallets/loans/src/entities/pricing/external.rs
```diff
@@ -126,7 +126,7 @@ impl<T: Config> ExternalActivePricing<T> {
 		Ok(T::PriceRegistry::get(&self.info.price_id)?.1)
 	}
 
-	pub fn current_interest(&self) -> Result<T::Balance, DispatchError> {
+	pub fn interest_accrued(&self) -> Result<T::Balance, DispatchError> {
```
lemunozm, line 129: I've changed this name because `current_interest` seems to return something like "2%",  instead of an interest accrued during the time.

mustermeiszer (commented):
Why add the debt to the type itself instead of just computing it on the fly in the API? Api is still missing, right?

lemunozm:
> the debt to the type itself instead of just computing it on the fly in the API

This is just what is happening, the type where I added interest/present_value/... is a type only used for the Runtime API, and is there where things are computed.

BTW, This PR needs a refactor after a last talk with Jeroen.

pallets/loans/src/entities/pricing/external.rs
lemunozm: No longer apply

pallets/loans/src/entities/loans.rs
```diff
@@ -344,25 +344,25 @@ impl<T: Config> ActiveLoan<T> {
 		&self,
 		mut amount: RepaidPricingAmount<T>,
 	) -> Result<RepaidPricingAmount<T>, DispatchError> {
-		let (interest_accrued, max_repay_principal) = match &self.pricing {
+		let (outstanding_interest, max_repay_principal) = match &self.pricing {
```
offerijns, line 347: Tiny detail but I think it would make sense to swap the tuple order here, as in all other cases principal comes before interest.

offerijns (dismissed):
LGTM!

pallets/loans/src/entities/loans.rs
lemunozm: Totally agree!

offerijns (approved):

PR #1484, opened 2023-08-09
Feature/precompile v3

mustermeiszer:
# Description
Includes the necessary changes for adding a precompile to our codebase that acts as a glue between the axelar evm logic and our substrate based connectors logic. 

## Changes and Descriptions
* provide a pallet that is at the same time a precompile
  * needed as we need capability to dynamically set the gateway address of the axelar contracts
* provides a new precompile set for development containing the precompile
* adds all pallets and changes to dev runtime
* actually use connectors-gateway as `OutboundQueue`


Cargo.toml
```diff
@@ -41,6 +41,7 @@ members = [
   "pallets/connectors",
   "pallets/connectors-gateway",
   "pallets/connectors-gateway/routers",
+  "pallets/connectors-gateway/connectors-gateway-axelar-precompile",
```
cdamian, line 44: Nit - maybe rename that folder to `precompile`?

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+		///    - Second string: Defines `sourceAddress`
+		type SourceConverter: sp_runtime::traits::Convert<
+			(Vec<u8>, Vec<u8>),
+			Result<cfg_types::domain_address::DomainAddress, DispatchError>,
+		>;
+	}
+
+	#[pallet::storage]
+	#[pallet::getter(fn get_pre_commit)]
+	pub(super) type Gateway<T: Config> = StorageValue<_, H160, ValueQuery>;
```
cdamian, line 70: Nit - rename to `GatewayAddress` or something more specific?

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+		///    - First string: Defines `sourceChain`
+		///    - Second string: Defines `sourceAddress`
+		type SourceConverter: sp_runtime::traits::Convert<
+			(Vec<u8>, Vec<u8>),
+			Result<cfg_types::domain_address::DomainAddress, DispatchError>,
+		>;
+	}
+
+	#[pallet::storage]
+	#[pallet::getter(fn get_pre_commit)]
```
cdamian, line 69: This getter func doesn't seem right :)

runtime/development/src/lib.rs
```diff
@@ -1571,22 +1570,6 @@ impl orml_asset_registry::Config for Runtime {
 	type WeightInfo = ();
 }
 
-pub struct DummyOutboundQueue;
```
cdamian, line 1574: This shouldn't be part of latest main, maybe a re-base might fix this?

runtime/common/src/evm/precompile.rs
```diff
@@ -54,6 +63,39 @@ where
...
+			))
+		} else if handle.code_address() == CONNECTORS_AXELAR_GATEWAY {
+			Some(<connectors_gateway_axelar_precompile::Pallet<R> as Precompile>::execute(handle))
+		} else {
+			None
+		}
+	}
+
+	fn is_precompile(&self, address: H160) -> bool {
+		address == DISPATCH_ADDR
```
cdamian, line 95: Not sure if we need to check for `CONNECTORS_AXELAR_GATEWAY` here as well.

mustermeiszer: Definitely

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+	#[pallet::pallet]
+	#[pallet::generate_store(pub (super) trait Store)]
+	pub struct Pallet<T>(_);
+
+	#[pallet::config]
+	pub trait Config:
+		frame_system::Config + pallet_evm::Config + pallet_connectors_gateway::Config
+	{
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
```
cdamian, line 53: Should we also add a type with a trait that the connectors gateway implements? This way we'll be able to mock that and also add some tests here and ensure that the right stuff is being call and so forth.

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
NunoAlexandre: @cdamian in Solidity usually contracts are identified by their address so it's not common to suffix these names with *Address. Not a strong opinion on my end either way tho

NunoAlexandre: But it could make sense to be specific whether this means `AxelarGateway` or `ConnectorsGateway`

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
mustermeiszer: We can not unfortunately, as we will need to call the extrinsic explictily. Or is it trait abstracted somwhere?

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
mustermeiszer: rm thanks ^^

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	// In Solidity, a boolean value (bool) is stored as a single byte (8 bits) in
+	// contract storage. The byte value 0x01 represents true, and the byte value
+	// 0x00 represents false.
+	//
+	// When you declare a boolean variable within a contract and store its value in
+	// storage, the contract reserves one storage slot, which is 32 bytes (256 bits)
+	// in size. However, only the first byte (8 bits) of that storage slot is used
+	// to store the boolean value. The remaining 31 bytes are left unused.
```
NunoAlexandre: Thanks for adding this 👌

Cargo.toml
mustermeiszer: named to axelar-gatway-precompile

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
mustermeiszer: suggestions?

runtime/common/src/evm/precompile.rs
```diff
@@ -54,6 +63,39 @@ where
 	}
 }
 
+pub struct DevelopmentPrecompiles<R>(PhantomData<R>);
```
NunoAlexandre, line 66: Can we document what this is? I will look further around to understand its application but going through it alone is not clear to me

NunoAlexandre: Plus, can we name this just `Development`, given it's defined in the `precompile.rs` module? That way we use it as `precompiles::Development` 👍

runtime/development/src/lib.rs
```diff
@@ -1873,6 +1856,26 @@ impl pallet_order_book::Config for Runtime {
...
+impl
+	sp_runtime::traits::Convert<
+		(Vec<u8>, Vec<u8>),
+		Result<cfg_types::domain_address::DomainAddress, DispatchError>,
+	> for SourceConverter
+{
+	fn convert(
+		_source: (Vec<u8>, Vec<u8>),
+	) -> Result<cfg_types::domain_address::DomainAddress, DispatchError> {
+		Err(DispatchError::Other("No supported domains yet!"))
```
NunoAlexandre, line 1869: This is what you mentioned that you would change still right?

runtime/development/src/lib.rs
```diff
@@ -1966,6 +1969,7 @@ construct_runtime!(
 		BaseFee: pallet_base_fee::{Pallet, Call, Config<T>, Storage, Event} = 162,
 		Ethereum: pallet_ethereum::{Pallet, Config, Call, Storage, Event, Origin} = 163,
 		EthereumTransaction: pallet_ethereum_transaction::{Pallet, Storage, Event<T>} = 164,
+		ConnectorsAxelarGateway: connectors_gateway_axelar_precompile::{Pallet, Call, Storage, Event<T>} = 165,
```
NunoAlexandre: Is it no longer possible to place this together with the other connectors pallet?

NunoAlexandre (commented):
Left some comments but all small stuff or some questions 🧗‍♂️

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
```diff
@@ -31,13 +32,63 @@ pub type Bytes<const U32: u32> = BoundedBytes<ConstU32<U32>>;
...
+		match self.domain {
+			Domain::Centrifuge => Some(DomainAddress::Centrifuge(Self::try_into_32bytes(
+				&maybe_address,
+			)?)),
+			Domain::EVM(id) => Some(DomainAddress::EVM(
+				id,
+				Self::try_into_20bytes(&maybe_address)?,
+			)),
+		}
+	}
```
NunoAlexandre: This makes sense! Can't we implement `TryInto` or `TryFrom` instead tho?

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+	// in size. However, only the first byte (8 bits) of that storage slot is used
+	// to store the boolean value. The remaining 31 bytes are left unused.
+	fn bool_to_h256(value: bool) -> H256 {
+		let mut bytes: [u8; 32] = [0u8; 32];
+
+		if value {
+			bytes[0] = 1;
+		}
+
+		H256::from(bytes)
```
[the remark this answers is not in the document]
mustermeiszer: Are we sure the byte sits at the right direction? If yes, then looks good!

runtime/development/src/lib.rs
mustermeiszer: Changed in latest version

runtime/development/src/lib.rs
mustermeiszer: I thought it belongs to EVM stuff. But can move

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
mustermeiszer: Both of these take only into account `self` but we need an external input. Like `TryConvert` but this is not available on our Subsrtate version yet. Or am I missing some>

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
+		fn build(&self) {
+			Gateway::<T>::set(self.gateway)
+		}
+	}
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		SetGateway { address: H160 },
```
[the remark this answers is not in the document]
mustermeiszer: I had it this way and wondered if it was the other way and changed it. Will change

runtime/development/src/lib.rs
mustermeiszer: I am on latest main. I hope at least ^^

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+			Token::FixedBytes(PREFIX_CONTRACT_CALL_APPROVED.into()),
+			Token::FixedBytes(command_id.as_bytes().into()),
+			Token::String(source_chain.clone().try_into().map_err(|_| {
+				RevertReason::read_out_of_bounds("utf-8 encoding failing".to_string())
+			})?),
+			Token::String(source_address.clone().try_into().map_err(|_| {
+				RevertReason::read_out_of_bounds("utf-8 encoding failing".to_string())
+			})?),
+			Token::Address(handle.context().address),
+			Token::FixedBytes(payload_hash.as_bytes().into()),
```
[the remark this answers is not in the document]
mustermeiszer: I just followed the logic from Axelar. ^^ 

I guess this gives them a static-ish size. but not sure

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
NunoAlexandre: We could always include that as having the conversion being between a (Domain, Address) and DomainAddress but this is fine 👍🏻

runtime/development/src/lib.rs
NunoAlexandre: Sorry, I misread it as "ConnectorsGateway" 🙃  all good, sorry for the noise!

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: so if `value` is a bool, then this is "if true { 1 } else { 0 }, which seems to match what your current implementation does, which in other words is:

```rust
match value {
  True => 1,
  False => 0
}
```

pallets/connectors-gateway/axelar-gateway-precompile/src/lib.rs
```diff
@@ -0,0 +1,397 @@
+// Copyright 2021 Centrifuge Foundation (centrifuge.io).
```
NunoAlexandre, line 1: @mustermeiszer do you think we may have more precompiles in the future within connectors? I see that Moonbeam has a whole `precompiles` directory at the root of their repo. I don't think we need that now but maybe having a `precompiles` dir here now containing the `axelar-gateway` precompile would make sense.

not a blocker whatsoever tho!

pallets/connectors-gateway/axelar-gateway-precompile/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+		if maybe_address.len() == 20 {
+			let mut address: [u8; 20] = [0u8; 20];
+			address.copy_from_slice(maybe_address);
+
+			Some(address)
+		} else {
+			None
+		}
+	}
+}
```
NunoAlexandre, line 82: We could use `vec_to_fixed_array` as the implementation logic of these functions aside from the `len()` check

NunoAlexandre (approved):
Looks correct to me! left two comments but no blockers 👍

pallets/connectors-gateway/axelar-gateway-precompile/src/lib.rs
mustermeiszer: Not sure tbh. Would keep it like that for now and refactor later.

pallets/connectors-gateway/axelar-gateway-precompile/src/lib.rs
mustermeiszer: ah I knew you had this problem already once. I will keep it as is fro now, as I need the len check.

mustermeiszer:
@NunoAlexandre do you know why the tests are failing?

cdamian:
@mustermeiszer Ah, this is most likely due to the fact that you are removing the `OutboundQueue` then the integration tests are using the actual gateway in the tests, causing failures. This should be fixed as soon as https://github.com/centrifuge/centrifuge-chain/pull/1478/files is merged and you do a small rebase ^^

pallets/connectors-gateway/axelar-gateway-precompile/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+	//        _execute(sourceChain, sourceAddress, payload);
+	// }
+	//
+	// Note: The _execute logic in this case will forward all calls to the
+	//       pallet-connectors-gateway with a special runtime local origin
+	#[precompile::public("execute(bytes32,string,string,bytes)")]
+	fn execute(
+		handle: &mut impl PrecompileHandle,
+		command_id: H256,
+		source_chain: String<MAX_SOURCE_CHAIN_BYTES>,
```
branan, line 226: Since all we care about is that these are buckets of bytes that we can toss into the hasher, and don't rely on them being printable UTF-8 strings, I don't think we necessarily need a different `String` type from `Bytes` here. That would also avoid the possible failure on a non-utf-8 source (but would that ever really happen?).

On the other hand, it's nice to match the solidity signature. :shrug:

Simplifying the hashing code by treating these as just bytes might be a touch cleaner, but I think this is fine

runtime/common/src/evm/precompile.rs
```diff
@@ -94,11 +99,75 @@ where
...
+
+impl<R> PrecompileSet for Development<R>
+where
+	R: pallet_evm::Config + axelar_gateway_precompile::Config + frame_system::Config,
+	R::RuntimeCall: Dispatchable<PostInfo = PostDispatchInfo> + GetDispatchInfo + Decode,
+	<R::RuntimeCall as Dispatchable>::RuntimeOrigin: From<Option<R::AccountId>>,
+	axelar_gateway_precompile::Pallet<R>: Precompile,
+{
+	fn execute(&self, handle: &mut impl PrecompileHandle) -> Option<PrecompileResult> {
+		match handle.code_address().0 {
```
branan, line 124: ...

Which would allow writing this as:
```
self.0.execute(handle).or_else(|| {
  match handle.code_address().0 {
    CONNECTORS_AXELAR_GATEWAY => ...,
    // Additional development-specific ones
    _ => None
  }
})
```

runtime/common/src/evm/precompile.rs
```diff
@@ -94,11 +99,75 @@ where
 	}
 }
 
+<<<<<<< HEAD
+/// A set of precompiles. This set might contain
+/// not yet mainnet ready precompiles in order to test
+/// those in development or staging environment without touching
+/// the mainnet set.
+pub struct Development<R>(PhantomData<R>);
```
branan, line 107: I'd probably implement this struct in terms of `CentrifugePrecompiles` so we can DRY up the list of common precompiles:

`pub struct Development<R>(CentrifugePrecompiles<R>);`
...

runtime/common/src/evm/precompile.rs
```diff
@@ -94,11 +99,75 @@ where
...
+			SHA3FIPS256_ADDR => Some(Sha3FIPS256::execute(handle)),
+			DISPATCH_ADDR => Some(Dispatch::<R>::execute(handle)),
+			ECRECOVERPUBLICKEY_ADDR => Some(ECRecoverPublicKey::execute(handle)),
+			CONNECTORS_AXELAR_GATEWAY => Some(<axelar_gateway_precompile::Pallet<R> as Precompile>::execute(handle)),
+			_ => None,
+		}
+	}
+
+	fn is_precompile(&self, address: H160) -> bool {
+		[
```
branan, line 143: As above. Bools have the same `.or_else` combinator as Options now, so the logic would be nearly identical.

branan (approved):
A few style comments, but this overall looks good

pallets/connectors-gateway/axelar-gateway-precompile/src/lib.rs
mustermeiszer: I will stick with compliant parameters here.

runtime/common/src/evm/precompile.rs
mustermeiszer: Good idea. Will do that!

pallets/connectors-gateway/connectors-gateway-axelar-precompile/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I also think bytes would be in different positions if memory is big-endian/little-endian.
Maybe @Nuno, you want to say:
```rust
H256::from(if value { [1u32] } else { [0u8] } );
```

NunoAlexandre (approved):

runtime/common/src/evm/precompile.rs
```diff
@@ -76,21 +82,52 @@ where
...
+				| DISPATCH_ADDR | ECRECOVERPUBLICKEY_ADDR
+		)
+	}
+}
+
+/// A set of precompiles. This set might contain
+/// not yet mainnet ready precompiles in order to test
+/// those in development or staging environment without touching
+/// the mainnet set.
+pub struct Development<R>(CentrifugePrecompiles<R>);
```
NunoAlexandre, line 102: @mustermeiszer how do you envision merging with with https://github.com/centrifuge/centrifuge-chain/pull/1490? like do want to add a custom `Altair` precompiles struct a la this `Development` or how do you see us doing that?

mustermeiszer: I would add an Altair type maybe? Whatever you find cleaner

PR #1487, opened 2023-08-10
Loans: add slippage for external pricing

### LINKED ISSUE #1464: Loans: Add slippage for settlement_price
### Description

After merging https://github.com/centrifuge/centrifuge-chain/pull/1455 we leave a tech debt for the future of adding a `slippage` value to limit the values the borrower can choose regarding `settlement_price`. More in the [slack thread](https://centrifugedao.slack.com/archives/C05DX0RH1T3/p1688128534804769) about this.

It still requires a bit of design before implementing it

lemunozm:
# Description

Fixes #1464 

## Changes and Descriptions

- Implementation
- Test fixed and updated with slippage feature
- **IMPORTANT**: Fixed bug in runtime API where the `present_value` didn't taking into account the write off.


pallets/loans/src/entities/pricing/external.rs
```diff
@@ -67,6 +67,11 @@ pub struct ExternalPricing<T: Config> {
 
 	/// Reference price used to calculate the interest
 	pub notional: T::Balance,
+
+	/// Maximum variation between the settlement price chosen for
+	/// borrow/repay and the current oracle price.
+	/// Represented as: Oracle price +/- oracle price * max_variation_price
+	pub max_variation_price: T::PerThing,
```
lemunozm, line 74: Finally, chosen variation, because *delta* usually means a "variation" but in the same units, i.e: $x + \Delta x$, and what we have is measured in %

pallets/loans/docs/types.md
```diff
@@ -156,7 +156,7 @@ package pricing {
             price_id: PriceId,
             max_borrow_quantity: MaxBorrowAmount,
             notional: Balance,
-            pool_id: PoolId
+            max_variation_price: Perthing,
```
offerijns, line 159: I think this name is worse than `max_price_delta`: variation technically is a change/delta over time (as in, the variance of a series of data points, rather than the difference of one data point to another)

pallets/loans/src/entities/pricing/external.rs
offerijns: Nit but then I think `max_price_variation` is more correct still :)

lemunozm: BTW, in the chat when you wrote:
*"I think “max delta price” is the more correct English order"*
did you want to say the opposite?

Regarding the "time" meaning for variation, maybe "proportion" could fit better

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -67,6 +67,11 @@ pub struct ExternalPricing<T: Config> {
 
 	/// Reference price used to calculate the interest
 	pub notional: T::Balance,
+
+	/// Maximum variation between the settlement price chosen for
+	/// borrow/repay and the current oracle price.
+	/// Represented as: Oracle price +/- oracle price * max_variation_price
```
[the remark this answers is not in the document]
lemunozm: I think this is ok, but the equation is not well expressed

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -67,6 +72,11 @@ pub struct ExternalPricing<T: Config> {
 
 	/// Reference price used to calculate the interest
 	pub notional: T::Balance,
+
+	/// Maximum variation between the settlement price chosen for
+	/// borrow/repay and the current oracle price.
+	/// See [`ExternalAmount::settlement_price`].
+	pub max_price_variation: T::PerThing,
```
mustermeiszer, line 79: This will limit us to a variation between `0-100%`. Is this wanted?

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -148,10 +158,36 @@ impl<T: Config> ExternalActivePricing<T> {
...
+		amount: &ExternalAmount<T>,
+		pool_id: T::PoolId,
+	) -> Result<(), DispatchError> {
+		let price = T::PriceRegistry::get(&self.info.price_id, &pool_id)?.0;
+		let delta = if amount.settlement_price > price {
+			amount.settlement_price.ensure_sub(price)?
+		} else {
+			price.ensure_sub(amount.settlement_price)?
+		};
+		let variation = T::PerThing::from_rational(delta.ensure_into()?, price.ensure_into()?);
```
mustermeiszer, line 172: This is not correct IMO. 

```rust
let th = Perquintill::from_rational(100u64, 1u64);
assert_eq!(th, Perquintill::one())
```
If `max_price_variation` is `1`
* you can easily create a $\Delta > price$ 
* but the variation will saturate at `1`

I think we must use a `FixedPointNumber` here.

mustermeiszer (changes requested):
If I understand the feature request correctly I would request changing `PerThing`.

pallets/loans/src/entities/pricing/external.rs
lemunozm: Is this wanted (and required), just for that, we've chosen that type.

pallets/loans/src/entities/pricing/external.rs
lemunozm: Good catch! it saturates. We still want a `PerThing`, but we need to return an error in case of saturating here.

lemunozm: Fixed now!

pallets/loans/src/entities/pricing/external.rs
offerijns: I think I had the wrong mental model. In my mind the variation couldn't be more than 100%, but I guess it can (if oracle price is 1.0 and settlement price is 3.0, the variation is 200%). So then maybe it is better to use a `Rate`

pallets/loans/src/entities/pricing/external.rs
lemunozm: BTW, not sure why there is not a `try_from_rational` or so

pallets/loans/src/entities/pricing/external.rs
lemunozm: But does that use case make sense? I thought that variation should be a minor correction over the oracle price

lemunozm: Well, even though that use case is "not real" maybe it makes more sense a `Rate`...

offerijns: Yeah I think it's unlikely but theoretically possible

lemunozm:
Changed to `Rate` type. I've left the `PerThing` associated config type because it will be needed when we change the write_off_percentage type from `Rate` to `PerThing`.

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -67,6 +72,11 @@ pub struct ExternalPricing<T: Config> {
 
 	/// Reference price used to calculate the interest
 	pub notional: T::Balance,
+
+	/// Maximum variation between the settlement price chosen for
+	/// borrow/repay and the current oracle price.
+	/// See [`ExternalAmount::settlement_price`].
+	pub max_price_variation: T::Rate,
```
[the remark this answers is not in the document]
lemunozm: Would we want asymmetric variations? I think usually those variations are symmetric (until saturating with zero, as it happens now).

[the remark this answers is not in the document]
mustermeiszer: I am not sure they are. I might be happy to have a variation to increase, but will always want to ensure I am not selling lower than market price. But could also be premature understanding of the markets and usage.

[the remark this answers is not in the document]
mustermeiszer: @offerijns and @denniswell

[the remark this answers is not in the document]
offerijns: I think we don't really need an upper variation as anyways repaying more than market value is always possible through unscheduled repayments. So I would keep it as a single `Rate` variation.

offerijns (approved):

lemunozm:
@mustermeiszer, I think we need your approval to get this merged

mustermeiszer (approved):

PR #1488, opened 2023-08-10
ethereum-allowlist: Add allowlist pallet for Ethereum contracts

cdamian:
# Description

Adds a new pallet that serves as an allowlist for contracts that can be added by an Ethereum account.

This pallet will be used when processing an EVM `create` call to ensure that no contract is created if the account and the hash of the contract code is not in the allowlist.

# Checklist:

### DIFF
--- a/pallets/ethereum-allowlist/src/lib.rs
+++ b/pallets/ethereum-allowlist/src/lib.rs
@@ -1,0 +1,118 @@
+// Copyright 2021 Centrifuge Foundation (centrifuge.io).
+// This file is part of Centrifuge chain project.
+
+// Centrifuge is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version (see http://www.gnu.org/licenses).
+
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use frame_support::{dispatch::DispatchResult, pallet_prelude::*};
+use frame_system::pallet_prelude::*;
+pub use pallet::*;
+use sp_core::{H160, H256};
+pub use weights::*;
+
+#[cfg(test)]
+mod mock;
+#[cfg(test)]
+mod tests;
+
+pub mod weights;
+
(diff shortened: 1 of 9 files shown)


mustermeiszer (commented):
Looks good! Just minor improvement regarding storage and logic to allow multiple calls of same hash. 

I would impl those 
* https://github.com/paritytech/frontier/blob/master/frame/ethereum/src/lib.rs#L108-L167
methods on the pallet level as a proxy and check when it is create if somebody is allowed to call it. 

Concern: 
* How do we remove the allowance once the extrinsic is through? I currently think the only way is to also proxy `pallet_ethereum::Call::Transact{..}` and force users to use this one as a self-contained call when they want to create stuff. 

Not sure how this works with the EVM side though ...

mustermeiszer:
cc @branan regarding proxying this

pallets/ethereum-allowlist/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+	#[pallet::pallet]
+	#[pallet::generate_store(pub (super) trait Store)]
+	pub struct Pallet<T>(_);
+
+	/// Storage for the hashes of the Ethereum contract byte codes that an
+	/// Ethereum account is allowed to create.
+	#[pallet::storage]
+	#[pallet::getter(fn account_code_hashes)]
+	pub type AccountCodeHashes<T: Config> =
+		StorageDoubleMap<_, Blake2_128Concat, H160, Blake2_128Concat, H256, (), ValueQuery>;
```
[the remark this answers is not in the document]
cdamian: Hmm, not sure how that would work. I thought this is just an allowlist for a contract that can be created on our EVM. Do we want to keep track of how many times a contract with the same code is created or what would be the purpose of that?

runtime/development/Cargo.toml
```diff
@@ -124,6 +124,7 @@ pallet-connectors-gateway = { path = "../../pallets/connectors-gateway", default
 pallet-crowdloan-claim = { path = "../../pallets/crowdloan-claim", default-features = false }
 pallet-crowdloan-reward = { path = "../../pallets/crowdloan-reward", default-features = false }
 pallet-data-collector = { path = "../../pallets/data-collector", default-features = false }
+pallet-ethereum-allowlist = { path = "../../pallets/ethereum-allowlist", default-features = false }
```
lemunozm, line 127: Here is the issue I think, you need to add this pallet to std/runtime-benchmarks/try-runtime

cdamian: Yup, I already have those changes locally and it's fine, I'm tackling the clippy issue now.

mustermeiszer:
closing as we currently do not need that.

Issue #1485, opened 2023-08-09
altair: Add `connectors`


NunoAlexandre:
We want to bring the connectors feature to the Altair runtime and have that available soon on Algol for testing our integration with Axelar.

This includes all the latest revision of connectors, including the gateway and latest routers.

PR #1490, opened 2023-08-10
altair: Add `liquidityPools` and `AxelarGateway` precompiles

NunoAlexandre:
Closes https://github.com/centrifuge/centrifuge-chain/issues/1485; To be merged after https://github.com/centrifuge/centrifuge-chain/issues/1489

### Changes

**Altair:**
- [x] Add `pallet_liquidity_pools`
- [x] Add `pallet_liquidity_pools_gateway`
- [x] Add `pallet_ethereum_transaction`
- [x] Add `pallet_xcm_transactor`
- [x] Add `axelar_gateway_precompile`

### Out of scope

- [Rename `connectors` to `liquidity-pools`](https://github.com/centrifuge/centrifuge-chain/issues/1489)


NunoAlexandre:
@mustermeiszer @branan regarding these two points:

> - [ ] Ensure algol genesis provides sufficient ALGL balance to dev accounts
> - [ ] Ensure algol genesis provides enough ETH balance to a dev account (passphrase to be generated)


Usually for other chains we have a `_local` genesis config where we specify these properties. However, we don't have one for `algol` (yet). Looking through the `res/algol-spec.rs` I see that we do set balance for some accounts, but as I tried to figure out where those properties are set or inherited from, I find the code around this rather obscure.

After adding the new pallets to the `altair` rutnime, I have re-exported the `algol` spec like so:

``` bash
./target/release/centrifuge-chain build-spec --chain algol --disable-default-bootnode > res/algol-spec.json
```

Following the code for the `BuildSpec` command, I see that it calls `load_spec` which, for algol, as expected, loads `res/algol-spec.json` but it's unclear to me where it's setting things like the algol evm chain.

Could you explain how this works? And I guess that we need a separate chain spec like "algol_local"?

branan:
* Regarding Algol genesis accounts: The funded accounts are all in lastpass, so developers should be able to claim them. One of the funded accounts is also the Demo environment faucet. Funding an EVM account from your claimed developer account is just a normal substrate transfer to the derived address, so there's no need to fund EVM accounts directly in the genesis.

* Regarding local development: Algol's distinguishing feature from Altair is that it's built with a different feature flag (`testnet_runtime`), so we having an `algol-local` spec would be confusing at best and just plain wrong at worst. The right way to do local Algol testing is to just run `altair-local` with the feature flag.

* Regarding re-generating the Algol spec, that would imply we're blocking the re-deployment of Algol on this PR landing. If that's not the case we shouldn't be re-generating the spec, and instead should do a runtime upgrade of Algol once this PR lands. The Algol spec needs to be static just like all other chain specs, so that future nodes built from the repo can recognize the genesis codehash and genesis blockhash.

runtime/altair/src/lib.rs
```diff
@@ -1302,6 +1303,89 @@ impl pallet_data_collector::Config for Runtime {
 	type Moment = Moment;
 }
 
+#[derive(Clone, Eq, Debug, PartialEq, Ord, PartialOrd, Encode, Decode, TypeInfo)]
+pub struct NullTransactor {}
+
+impl UtilityEncodeCall for NullTransactor {
+	fn encode_call(self, _call: UtilityAvailableCalls) -> Vec<u8> {
+		unimplemented!("XcmTransactor feature not used")
```
mustermeiszer, line 1311: When would one hit this logic? Just am worried about `unimplemented` in our logic.

NunoAlexandre: Good question! The only extrinsic from `xcm_transactor` we use is `pallet_xcm_transactor::Pallet::<T>::transact_through_sovereign(`, which doesn't use this config type:

```rust
		// XcmTransact needs to be implemented. This type needs to implement
		// utility call encoding and multilocation gathering
		type Transactor: Parameter + Member + Clone + XcmTransact;
```

`Transactor` is used in `transact_through_derivative`, and this extrinsic requires static knowledge on how to encode a `AsDerivative` call on a given destination chain. Again, we don't use this.

What is clear by going through this after your prompt is that we must make `xcm_transactor::Call` hidden to stop users from calling `transact_through_derivative` which would call this `unimplemented`. Thanks for the catch 👍

NunoAlexandre: @mustermeiszer actually just tested this scenario on our node on the Moonbase Alpha and `transact_through_derivative` fails early on because we'd need to register a derivative index account (`IndexToAccount` storage, which requires `EnsureRoot`).

So we _can_ make `impl UtilityEncodeCall` return `vec![]` but for `impl xcm_primitives::XcmTransact` which requires `fn destination(self) -> MultiLocation` we must return some value, for example `MultiLocation { 0, HERE }` which is a sort of a `dev/null` in this context.

The later is safer but in reality none of the options ever get hit. wdyt?

@wischli would be cool to get your thoughts on this as well

NunoAlexandre: I have DRYed the `NullTransactor` into `runtime_common` and made it a safer implementation: https://github.com/centrifuge/centrifuge-chain/pull/1490/commits/1e76620268ba07205882db02446b793d0954f497

I think this is a safer and cleaner approach for us to move forward with without loosing sleep :)

wischli: > So we _can_ make `impl UtilityEncodeCall` return `vec![]` but for `impl xcm_primitives::XcmTransact` which requires `fn destination(self) -> MultiLocation` we must return some value, for example `MultiLocation { 0, HERE }` which is a sort of a `dev/null` in this context.
>
> The later is safer but in reality none of the options ever get hit. wdyt?

I would also feel much better going this path with a proper comment for both impls more or less citing your comment above. IMO, we should never add `unimplemented` or `todo` code to Altair/Centrifuge runtimes if not feature gated.

We should also add `transact_through_derivative` to the `BaseCallFilter` to ensure the call cannot be made. We cannot guarantee the calls behaviour will stay is forever. This at least guarantees, it cannot be called such that we don't need to rely on the assumption of erroring out earlier.

mustermeiszer: Maybe lets add a note to the `construct_runtime!` section where we add the `pallet_xcm_transactor` that we never want to add the call to our runtime call unless someone takes care of the `Transactor`

NunoAlexandre: @wischli @mustermeiszer I forgot to correct something I said above; we do need the `xcmTransactor` `Call` to set some storage state. I agree that given we can't guarantee future behaviour for the xcm_transactor `transact_through_derivative` call we should just block it in the call filter.

NunoAlexandre: Done in https://github.com/centrifuge/centrifuge-chain/pull/1490/commits/3d1f7b8e546a7cc33da0f6eba68bf1d82ada1368

mustermeiszer (approved):

wischli (approved):
Thanks for adhering to my suggestion! LGTM 🚀

NunoAlexandre:
@mustermeiszer @wischli thanks a ton for the reviews 🚀  I am working on renaming connectors to liquidityPools and that should be in before we add ~~connectors~~ liquidityPools to Altair to avoid - even if so slight the risk - the need for storage migrations

mustermeiszer:
@NunoAlexandre this must include #1484.

NunoAlexandre:
> @NunoAlexandre this must include #1484.

Do you mean that #1484 must also be applied to Altair? or how do you mean? Because, if the former, it makes more sense if you rebase your branch based on this one and just add the precompile to altair as well, no?

Let me know if you want me to do that 👍

mustermeiszer:
I would prefer to merge #1484 and then add it as part of this PR. As this is part of the `connectors` logic

NunoAlexandre:
> I would prefer to merge #1484 and then add it as part of this PR. As this is part of the `connectors` logic

That's fine with me 👍 feel free to merge your PR and I take over from there

runtime/development/src/lib.rs
```diff
@@ -1964,7 +1949,7 @@ construct_runtime!(
 		BaseFee: pallet_base_fee::{Pallet, Call, Config<T>, Storage, Event} = 162,
 		Ethereum: pallet_ethereum::{Pallet, Config, Call, Storage, Event, Origin} = 163,
 		EthereumTransaction: pallet_ethereum_transaction::{Pallet, Storage, Event<T>} = 164,
-		ConnectorsAxelarGateway: axelar_gateway_precompile::{Pallet, Call, Storage, Event<T>} = 165,
+		LiquidityPoolsAxelarGateway: axelar_gateway_precompile::{Pallet, Call, Storage, Event<T>} = 165,
```
NunoAlexandre, line 1952: Renamed this here as well

runtime/development/src/evm.rs
```diff
@@ -61,30 +62,35 @@ impl pallet_evm::Config for crate::Runtime {
...
 	type StateRoot = pallet_ethereum::IntermediateStateRoot<Self>;
 }
 
-impl pallet_ethereum_transaction::Config for crate::Runtime {
-	type RuntimeEvent = crate::RuntimeEvent;
+impl pallet_ethereum_transaction::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+}
+
+impl axelar_gateway_precompile::Config for Runtime {
```
NunoAlexandre, line 93: Moved this here since it falls into the evm-related-pallets category and I did that right away on Altair, so it's consistent across the two runtimes

runtime/development/src/evm.rs
```diff
@@ -42,12 +43,12 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for FindAuthorTruncated<F> {
 
 parameter_types! {
 	pub BlockGasLimit: U256 = U256::from(NORMAL_DISPATCH_RATIO * MAXIMUM_BLOCK_WEIGHT.ref_time() / WEIGHT_PER_GAS);
-	pub PrecompilesValue: Development<crate::Runtime> = Development::<_>::new();
+	pub PrecompilesValue: Development<Runtime> = Development::<_>::new();
 	pub WeightPerGas: Weight = Weight::from_ref_time(WEIGHT_PER_GAS);
 }
 
-impl pallet_evm::Config for crate::Runtime {
```
NunoAlexandre, line 49: Nit but given the 10+ references we just import it above for simplicity.

runtime/common/src/evm/precompile.rs
```diff
@@ -95,6 +95,10 @@ where
 	}
 }
 
+/// Altair's precompiles
+/// For now, Altair uses the exact same set of precompiles used in Development.
+pub type Altair<R> = Development<R>;
```
NunoAlexandre, line 100: Given that they meant to be the exact same set for now, I made this an alias; shall we need a different set for Altair, we just need to change this file alone in the future.

NunoAlexandre (commented):
Some notes for reviewers

mustermeiszer (approved):
Thanks! Looks good! 

We can cut out a release afterwards, wdyt?

NunoAlexandre:
> Thanks! Looks good!
> 
> We can cut out a release afterwards, wdyt?

if we are confident about upgrading Altair, that's fine with me. I am not 100% confident tho 🤔

wischli (commented):
It appears to me that the `LiquidityPoolsGateway` runtime index is incorrect. Additionally, I suppose weight types will be updated as part of the release? 

The rest LGTM!

runtime/altair/src/lib.rs
```diff
@@ -1302,6 +1309,74 @@ impl pallet_data_collector::Config for Runtime {
...
+	type HrmpManipulatorOrigin = EnsureRootOr<HalfOfCouncil>;
+	type MaxHrmpFee = xcm_builder::Case<MaxHrmpRelayFee>;
+	type ReserveProvider = xcm_primitives::AbsoluteAndRelativeReserve<SelfLocation>;
+	type RuntimeEvent = RuntimeEvent;
+	type SelfLocation = SelfLocation;
+	type SovereignAccountDispatcherOrigin = EnsureRoot<AccountId>;
+	type Transactor = xcm_transactor::NullTransactor;
+	type UniversalLocation = UniversalLocation;
+	type Weigher = XcmWeigher;
+	type WeightInfo = ();
```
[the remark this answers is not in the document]
NunoAlexandre: Yap we should. Or did you have different ideas, @mustermeiszer?

runtime/altair/src/lib.rs
```diff
@@ -1666,13 +1741,16 @@ construct_runtime!(
 		BlockRewards: pallet_block_rewards::{Pallet, Call, Storage, Event<T>, Config<T>} = 105,
 		Keystore: pallet_keystore::{Pallet, Call, Storage, Event<T>} = 106,
 		PriceCollector: pallet_data_collector::{Pallet, Storage} = 107,
+		LiquidityPools: pallet_liquidity_pools::{Pallet, Call, Storage, Event<T>} = 108,
+		LiquidityPoolsGateway: pallet_liquidity_pools_gateway::{Pallet, Call, Storage, Event<T>, Origin } = 115,
```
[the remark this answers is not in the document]
NunoAlexandre: Note that I kept it `115` like on development since we had spoken, in the last few weeks, about trying to keep pallet indexes the same across the three runtimes. I am fine making it `109` so that's sorted within the `altair` runtime but then it won't be consistent cross runtimes. wdyt? cc @mustermeiszer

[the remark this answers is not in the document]
mustermeiszer: Can we even make the ordering the same on all three? If so, I would like to keep that, if not, lets have sequential.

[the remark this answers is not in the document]
NunoAlexandre: We sure could try but let's not spend time on this now. Made it `109` to be sequential 👍

wischli (approved):
LGTM!

mustermeiszer (approved):

PR #1491, opened 2023-08-11
Github codeowners/templates changes

lemunozm:
## Changes and Descriptions

- Removed `*.toml` matching to avoid having all of us in all PRs.
- (opinionated) Removed more template sections. I think most of all, do not use those sections, and in a lot of cases, it does not apply to the issue/PR purpose, and it only puts another mind barrier to open new issues. I think open issues should be something fast/easy/kind to incentivize tracking our task.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,15 +1,2 @@
-### Description
+# Description
 
-[Description of the issue or feature]
-
-### Research/based on
-
-[Which codebases, pallets or designs did you base your design on]
-
-### How will this affect the code base
-
-[Does it improve readability, performance? Will it introduce new complexity?]
-
-### What are foreseen obstacles or hurdles to overcome?
-
-[Are migrations needed, structural changes around testing etc?]
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,2 +1,15 @@
-# Description
+### Description
 
+[Description of the issue or feature]
+
+### Research/based on
+
+[Which codebases, pallets or designs did you base your design on]
+
+### How will this affect the code base
+
+[Does it improve readability, performance? Will it introduce new complexity?]
+
+### What are foreseen obstacles or hurdles to overcome?
+
+[Are migrations needed, structural changes around testing etc?]
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,16 +1,9 @@
 # Description
 
-Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.
+[description]
 
-Fixes # (issue)
+Fixes #(issue)
 
 ## Changes and Descriptions
 
 [List your changes here]
-
-# Checklist:
-
-- [ ] I have added Rust doc comments to structs, enums, traits and functions
-- [ ] I have made corresponding changes to the documentation
-- [ ] I have performed a self-review of my code
-- [ ] I have added tests that prove my fix is effective or that my feature works
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,3 +7,10 @@ Fixes #(issue)
 ## Changes and Descriptions
 
 [List your changes here]
+
+# Checklist:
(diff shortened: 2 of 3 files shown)


NunoAlexandre (dismissed):
Agreed and approved 🤝

.github/pull_request_template.md
```diff
@@ -1,16 +1,9 @@
...
 ## Changes and Descriptions
 
 [List your changes here]
-
-# Checklist:
-
-- [ ] I have added Rust doc comments to structs, enums, traits and functions
-- [ ] I have made corresponding changes to the documentation
-- [ ] I have performed a self-review of my code
-- [ ] I have added tests that prove my fix is effective or that my feature works
```
wischli, line 16: I am not sure whether I can agree with this removal. Personally, I go through this checklist for every PR but I suppose the reality is, that it is rather ignored and many PRs do not include any ticks.

wischli (commented):
Apart from the PR checklist removal, I fully agree with the changes. I also think the issue template is overhead it most occasions such that many chose to ignore it.

.github/issue_template.md
```diff
@@ -1,15 +1,2 @@
...
-
-[Which codebases, pallets or designs did you base your design on]
-
-### How will this affect the code base
-
-[Does it improve readability, performance? Will it introduce new complexity?]
-
-### What are foreseen obstacles or hurdles to overcome?
-
-[Are migrations needed, structural changes around testing etc?]
```
mustermeiszer, line 15: I like that sections TBH and found it helpful to have the structure.

.github/pull_request_template.md
cdamian: Same here, I like having this checklist for chores that might be forgotten.

.github/issue_template.md
NunoAlexandre: my POV is that these are often good questions to ask oneself when creating an issue but I personally don't find value is having to try and fit this structure per se. my 2cents, whatever is more helpful to the team is good with me 👍

cdamian: IMO a section for description and probably another one for extra notes might be more than enough.

.github/pull_request_template.md
lemunozm: Yes, In some PRs, I also found myself reading it to check mentally if things were done.

But, I found 2 points against having this there (maybe some other in place?).
- Most of the time, some points do not apply to the PR meaning.
- Most of the time, PRs are merged without clicking done in all those points.

From my point of view, I think each PR has its own checklist and should be thought of specifically for the PR purpose. So I agree with adding a checklist, but I think should be done explicitly at that moment of creation for the content of that PR.

.github/issue_template.md
mustermeiszer: I am good with what the matority wants

lemunozm: I do not have a strong opinion of these sections. It's true that sometimes they do not apply to the issue or can be handled by tags (as "migration needed"), but they also help to explain the issue better

lemunozm:
I knew it would open discussions 😆

mustermeiszer (approved):

lemunozm:
I'll leave this PR open until/during next week, to get a chance for others to give arguments

.github/issue_template.md
wischli: Personally, I would like to keep this but it seems like the majority choses to ignore it. I don't have a strong opinion.

NunoAlexandre: @wischli maybe we can find a middle ground that makes everyone happy?

lemunozm:
I think, given most of you want to keep these template parts, I'm going to leave it as it is. If in some months you find it less usable, then can remove it. Is everyone ok with it?

lemunozm:
@wischli @NunoAlexandre @cdamian @mustermeiszer would you want to add yourself to the CODEOWNERS file with the new liquidity-pools-related content we added last month?

Feel free to add a commit in this branch

PR #1492, opened 2023-08-11
Generic nuke migrations

### LINKED ISSUE #1461: Loans: prepare Loans for release

lemunozm:
# Description

This PR adds a nuke migration utility agnostic from any pallet to be used when we need to restore the storage of a pallet. Located in `runtime-common`.

To nuke the migration of a pallet you need to add the following line (supposing we want to nuke `Loans` from on-chain version `1`):
```rust
runtime_common::migrations::nuke::Migration<Loans, RocksDbWeight, 1>
```

**This PR do not remove/add any runtime migration**

Partial_fix #1461

## Testing
Tested in Altair (previously modifying the runtime by adding this migration) with the following:
```sh
RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace cargo run --release --features try-runtime try-runtime --runtime target/release/wbuild/altair-runtime/altair_runtime.wasm --chain centrifuge on-runtime-upgrade live --uri wss://fullnode.altair.centrifuge.io:443
```

Done positive (when it works) & negative (when it shouldn't work) testing


runtime/common/src/migrations/nuke.rs
```diff
@@ -0,0 +1,127 @@
...
+				"Nuke-{}: nuke aborted. This upgrade must be removed!",
+				Pallet::name()
+			);
+			return Weight::zero();
+		}
+
+		if Pallet::on_chain_storage_version() < Pallet::current_storage_version() {
+			log::info!("Nuke-{}: nuking pallet...", Pallet::name());
+
+			// TODO: How we can set the maximum for this? Currently hardcode
```
mustermeiszer, line 74: I would set no Maximum TBH. If you want to nuke it, you need to ierate everything. WDYT?

lemunozm: How can be protected against that, or even know that limit well taking into account that this will be done in the same block after the upgrade with other migrations?

mustermeiszer: Migrations anyways block a whole block...

If it is too big, it is too big. That is currently the sad truth I would say.

lemunozm: Mmm... so the following could not halt the chain:
- Upgrade runtime version.
  - Runtime is upgraded.
  - Each migration is done...
  - Huge nuke migration is done (suppose it would take 10 blocks of time)
- The node appears as "halted" so another node is chosen to create the runtime upgrade block, which later would appear halted too, etc...

mustermeiszer: If the realy-chain can not produce the block in timely fashion or the parachain can`t we are stuck and must revert through relay democracy.

wischli: > If [either] the [assigned] relay-chain [validators] or the parachain [collators] can not produce the block in timely fashion, we are stuck and must revert through relay democracy.

Made some minor additions. If the block cannot be produced, it will be tried over and over again as the upgrade is part of `on_initialize`. Therefore, if we either know the number of migrating items is sufficiently large or `try-runtime` returns a weight close enough to the maximum block weight, another approach needs to be chosen. 

Parity is [currently working on multi block migrations](https://github.com/paritytech/substrate/pull/14275). Until it is merged and tested and our chain updated to the supported substrate version, we should go with a lazy migration instead, if necessary. I provided some food for thoughts [here](https://github.com/centrifuge/centrifuge-chain/pull/1343#issuecomment-1591333784).

lemunozm: Removed the maximum limit then 👍🏻

wischli (commented):
The code LGTM! I was wondering though whether it was desired not to apply the migration to any runtime? Just realized this after I could not get any logs when running against Altair or Centrifuge.

mustermeiszer:
I would merge after #1493

lemunozm:
> I was wondering though whether it was desired not to apply the migration to any runtime? Just realized this after I could not get any logs when running against Altair or Centrifuge.

@wischli, yeah, it all depends on the chain state at the moment of the runtime upgrade (at least for `pallet-loans` and `pallet-interest-accrual`). By now, it's added as a utility. On the release day, we should check those storages and add the nuke migration for them. Or maybe it would be more safe adding them now and remove whatever we found on the runtime upgrade day 🤔

wischli:
> @wischli, yeah, it all depends on the chain state at the moment of the runtime upgrade (at least for `pallet-loans` and `pallet-interest-accrual`). By now, it's added as a utility. On the release day, we should check those storages and add the nuke migration for them. Or maybe it would be more safe adding them now and remove whatever we found on the runtime upgrade day 🤔

If they are planned to be part of the next release, I would like to add them now as otherwise the verification of the correctness of the migration will only be done during release and not as part of this PR. I don't have a strong opinion about this.

lemunozm:
For me, the purpose of this PR is only to add this utility for future usage in a common place, without affecting any behavior. 

I can add a next PR with the required changes just after merging this in altair/centrifuge runtimes to remove old migrations and add this nuke migration to the required pallets for the next upgrade. WDYT? Just for splitting concerns in different PRs

wischli (approved):
Thanks for the conversation! Definitely in favor of splitting up PRs such that the addition to corresponding runtimes can be done in a subsequent PR.

wischli (approved):
Re-approving

PR #1496, opened 2023-08-14
liquidity-pool: Drop `KnownInstances` storage

NunoAlexandre:
This storage should have been removed the gateway was added and integrated into the mix, since it's the gateway's responsibility to handle these authorisation matters.

No need for migrations since this is only on `development` as of now.


pallets/liquidity-pools/src/lib.rs
```diff
@@ -664,7 +647,7 @@ pub mod pallet {
 
 		/// Allow a currency to be used as a pool currency and to invest in a
 		/// pool on the domain derived from the given currency.
-		#[pallet::call_index(90)]
+		#[pallet::call_index(9)]
```
NunoAlexandre, line 650: This seems to have been a typo which I am fixing here.

wischli (approved):
LGTM! Thanks for handling this separately!

mustermeiszer (dismissed):

PR #1500, opened 2023-08-16
Fix missing cargo features

NunoAlexandre:
This should fix the [failed build on main](https://github.com/centrifuge/centrifuge-chain/actions/runs/5876657996/job/15935260717).

`NOTE`: Still verifying that's the case 🚧


lemunozm:
If still failing, there is a `scripts/tests.sh` script that tries to check all crates with all features. Probably running it will fail at some point giving information about the failing features.

NunoAlexandre:
> If still failing, there is a `scripts/tests.sh` script that tries to check all crates with all features. Probably running it will fail at some point giving information about the failing features.

It seems fine locally but I will run `scripts/tests.sh` for completeness 👍  thanks

offerijns (approved):

PR #1502, opened 2023-08-16
Altair spec 1030 for Algol

branan:


NunoAlexandre (approved):
LG 🚀

branan:
@NunoAlexandre is this integration test failure the one you mentioned in Slack you were working on?

wischli (commented):
I would like `UpgradeAltair1028` to be bumped to `UpgradeAltair1029` as part of this PR as well 🙃

branan:
Given we might have more algol upgrades before we finally release on Altair, I think it would be better to ticket that we need to fix-up the migrations and do that work as part of the final Altair release.

wischli (dismissed):
Withdrawing my previous comment, you are right @branan 🚀

branan:
Watch, now that we agreed on that, we won't end up doing another Algol release before Altair :smile:

NunoAlexandre:
> @NunoAlexandre is this integration test failure the one you mentioned in Slack you were working on?

No. I worked on fixing the runtime builds and that's fixed now. Not sure what this integration issue is; if it's a CI error we can try to rerun in, otherwise we need to look into that 👀

branan:
Looks like it's an out of space issue in the cargo caching step. Yay.

mustermeiszer (dismissed):
LGTM!

NunoAlexandre:
@branan @mustermeiszer are we planning on keeping this as a side-branch for now? I thought we wanted to merge this?

mustermeiszer:
I was also under the impression of merging it. I think we should merge it, but create no new tag.

mustermeiszer:
@branan can you take care of this today? ^^

wischli:
Since we have added new migrations since the 1029 upgrade on Algol, I took the liberality to bump to 1030 and fix the `CrossChainTransferabilityMigration` short circuit.

runtime/altair/src/migrations.rs
```diff
@@ -64,7 +66,7 @@ mod asset_registry {
 
 	impl OnRuntimeUpgrade for CrossChainTransferabilityMigration {
 		fn on_runtime_upgrade() -> Weight {
-			if VERSION.spec_version > 1028 {
+			if VERSION.spec_version > 1030 {
```
mustermeiszer, line 69: This will run on `Algol` then? 

@NunoAlexandre IIRC you already tested that another run does not harm, right?

NunoAlexandre: I have an idea 👍

NunoAlexandre: @mustermeiszer https://github.com/centrifuge/centrifuge-chain/pull/1510 👀

mustermeiszer: Nice!

mustermeiszer (approved):
re-approve

PR #1503, opened 2023-08-16
Only validate Frontier logs  from after the EVM was enabled.

branan:
The Frontier block importer layer does an early filter for bad blocks by ignoring any which do not have EVM consensus logs. However, since we did not have the EVM at genesis, we have thousands of blocks which don't have those consensus logs.

This duplicates the Frontier block importer and filters blocks only after a new `first_evm_block` parameter. This parameter is passed through from the chain spec file as a new extension.

# Checklist:


mustermeiszer (approved):
LGTM! Nice tweak @branan

mustermeiszer:
@branan I will rebase this one and get it in.

mustermeiszer (approved):

NunoAlexandre:
We need more PRs with descriptions like this one. This is literally embedded historical documentation 👌

wischli (approved):
I suppose this has higher priority than #1502 ?!

PR #1507, opened 2023-08-17
altair: Add `LiquidityRewards` & `LiquidityRewardsBase` pallets

NunoAlexandre:


mustermeiszer (approved):
Changes look good. I think for algol we should be good to go.

NunoAlexandre:
> Changes look good. I think for algol we should be good to go.

Cool. Based on the discussion on Slack I have feature-gatted the new pallets but if we think we don't need to, a simple `git revert ` https://github.com/centrifuge/centrifuge-chain/pull/1507/commits/2fdd646eb4b098bfc7ab6ab65a2f9ffa92c75b0a will do 👍

wischli:
I would still like this to be feature gated. Adding it to Altair with the argument we wont release feels like merging `unimplented` code into a production state. Say we need to patch, we need to remember feature gating. It think it does not hurt to be defensive and safe.

src/chain_spec.rs
```diff
@@ -776,6 +776,10 @@ fn altair_genesis(
 		},
 		ethereum: Default::default(),
 		evm: Default::default(),
+		liquidity_rewards_base: altair_runtime::LiquidityRewardsBaseConfig {
```
branan, line 779: I think this will break if `testnet_runtime` isn't specified, since then this wouldn't exist in the Genesis?

I'd rather remove that feature gate anyway - making algol more-different from Altair is just going to be even more of a maintenance headache. Since we intend to land this feature in the Centrifuge runtime Soon:tm:, I think we should just add it to Altair properly.

branan (dismissed):
One comment about the feature gating, but otherwise LGTM.

We'll need to benchmark this properly before we can release it to Altair, but I think it's fine to use the stubbed weights in Algol :shrug:

src/chain_spec.rs
branan: Oh, yup. The test failure is exactly the compile error I thought would happen

runtime/altair/src/lib.rs
```diff
@@ -1377,6 +1392,35 @@ impl pallet_liquidity_pools_gateway::Config for Runtime {
...
+	type Weight = u64;
+	type WeightInfo = ();
+}
+
+#[cfg(feature = "testnet-runtime")]
+impl pallet_rewards::Config<pallet_rewards::Instance2> for Runtime {
+	type Currency = Tokens;
+	type CurrencyId = CurrencyId;
+	type GroupId = u32;
+	type PalletId = RewardsPalletId;
```
branan, line 1416: Should this be a different pallet ID?

runtime/altair/src/lib.rs
```diff
@@ -1377,6 +1392,35 @@ impl pallet_liquidity_pools_gateway::Config for Runtime {
 	type WeightInfo = ();
 }
 
+#[cfg(feature = "testnet-runtime")]
+impl pallet_liquidity_rewards::Config for Runtime {
+	type AdminOrigin = EnsureRootOr<HalfOfCouncil>;
```
lemunozm, line 1397: One of the reasons it was not still in `altair` & `centrifuge` was to choose this value properly.

All functions to configure reward groups should be relatively fast and done by some kind of administrator (maybe). Those are: increasing the distributed reward, setting epochs, creating groups, etc.

Ideally, these functions will be handled by governance, but maybe the governance process is too slow for this. i.e. If a new pool is created with N tranches, we should configure then some rewards groups or associate those tranches to an existing group. Doing by governance would imply some period where the reward system is not still working for that pool.

runtime/altair/src/lib.rs
```diff
@@ -1377,6 +1392,35 @@ impl pallet_liquidity_pools_gateway::Config for Runtime {
...
+	type CurrencyId = CurrencyId;
+	type GroupId = u32;
+	type InitialEpochDuration = InitialEpochDuration;
+	type MaxChangesPerEpoch = MaxChangesPerEpoch;
+	type MaxGroups = MaxGroups;
+	type Rewards = LiquidityRewardsBase;
+	type RuntimeEvent = RuntimeEvent;
+	type Timer = Timestamp;
+	type Weight = u64;
+	type WeightInfo = ();
```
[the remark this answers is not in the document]
lemunozm: It's true that for Algol we do not need weights, but I think adding the file in the correct location and pointing to it doesn't hurt and protects us against a future oversight when we use it in Altair

runtime/altair/src/lib.rs
mustermeiszer: Couldn't we do it as part of the pool creation process?

lemunozm: I think it could work. Just a question:
- Would adding a pool also increase the total distributed reward or just would take some amount of the current distribution reward?

lemunozm: Once a pool is created, from liquidity reward PoV:
- It would require (or not) to create some new group for some new % rate. All groups could be created at the beginning.
- For each tranche, you need to move each PoolXTrancheX currency into the reward group with the corresponding %.
- Decide if we need to increase the minted reward (I think this can be done always in another extrinsic handle by economics)

offerijns: For now I think this needs to be governance. I would imagine what will happen initially is that soon the first 1 or 2 pools will launch, and then there will be a first governance proposal that sets the rate, creates the groups, and assigns the existing tranches from those 1 or 2 pools to those reward groups. Then afterwards it definitely makes sense to batch this with the pool creation process, but the reward groups need to be known for this.

> Would adding a pool also increase the total distributed reward or just would take some amount of the current distribution reward? I would say the second choice, because in the creation moment, the stake remains the same

Yes the total distributed reward should not change with more pools.

src/chain_spec.rs
```diff
@@ -776,6 +776,10 @@ fn altair_genesis(
...
+			currency_id: CurrencyId::Native,
+			amount: altair_runtime::ExistentialDeposit::get(),
```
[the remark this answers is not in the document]
wischli: It must be at least the `ExistentialDeposit` because else the `liquidity_rewards_base` sovereign account does not hold any balance on its own such that users cannot claim below the ED. We could also add this type to the `pallet_rewards::Config` and use it for the `GenesisBuild` but IMO the current approach is cleaner.

Adding this to the chain spec will only become effective if we need to update the Algol spec at some point upon restarting the network. Else we need to add the funds [via the pallets/rewards/src/migrations/new_instance.rs](https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/rewards/src/migrations/new_instance.rs).

Quoting myself from https://github.com/centrifuge/centrifuge-chain/pull/1342

> While testing block rewards on a local chain environment, I stumbled into a minor issue which needs to be fixed: When claiming rewards, we [currently transfer](https://github.com/centrifuge/centrifuge-chain/blob/bd4a96bf6e9bb6606f6b285cfe0493b7415a8734/pallets/rewards/src/lib.rs#L370) from the pallet id of the  base `pallet_rewards` instance to the recipient with `keep_alive` enabled. This blocks any reward claims which would reduce the balance of the pallet's sovereign account below the existential deposit (ED). The `keep_alive` should stay enabled. Thus, we need to fund the `pallet_rewards` sovereign account (at least) with the ED which is done by the migration [added here](https://github.com/centrifuge/centrifuge-chain/pull/1342/files#diff-0946c08ffa10c8e8b0c1113bc058b68fda4c798a1724cad4c5b7e6241ce6cd6cR55).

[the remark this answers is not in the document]
wischli: Anything that gets at least the `ED` in the native currency into the pallet's sovereign account does the trick.

runtime/altair/src/lib.rs
```diff
@@ -1192,14 +1192,20 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
...
 	pub const SingleCurrencyMovement: u32 = 1;
 	#[derive(scale_info::TypeInfo, Debug, PartialEq, Eq, Clone)]
 	pub const MaxChangesPerEpoch: u32 = 50;
 	pub const RewardsPalletId: PalletId = cfg_types::ids::BLOCK_REWARDS_PALLET_ID;
 	pub const RewardCurrency: CurrencyId = CurrencyId::Native;
+
+	#[derive(scale_info::TypeInfo)]
+	pub const MaxCurrencyMovements: u32 = 50;
+	#[derive(scale_info::TypeInfo)]
+	pub const MaxGroups: u32 = 20;
```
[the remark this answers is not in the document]
lemunozm: Re-thinking this, I think it's enough for now 👍🏻

src/chain_spec.rs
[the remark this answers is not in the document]
lemunozm: If you agree, I'll do a new PR with this in order to double-check with you if it would work leaving this as it is right now.

lemunozm:
Because Nuno is out, I'll take the ownership of this PR

src/chain_spec.rs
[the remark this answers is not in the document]
lemunozm: Done: [38c8f51](https://github.com/centrifuge/centrifuge-chain/pull/1507/commits/38c8f5173d7f4ca5ba371fd966b6e17e6abd840c)

runtime/altair/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Done

runtime/altair/src/lib.rs
lemunozm: Done

src/chain_spec.rs
lemunozm: Done

mustermeiszer (approved):

wischli (commented):
This PR is missing adding the funds [via the pallets/rewards/src/migrations/new_instance.rs](https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/rewards/src/migrations/new_instance.rs) for running networks such as Algol or Altair. 

Can you add this in a subsequent one @lemunozm ?

lemunozm:
Good catch! I'll add it!

PR #1508, opened 2023-08-19
Rewards: Add deposit for reward account for Altair. Enable liquidity rewards for all chains based on Altair

lemunozm:
# Description

This PR adds:
- A migration to fund the distributed reward account with `ED`.
- Enable liquidity rewards for all chains based on `altair`. After this [slack conversation](https://kflabs.slack.com/archives/C04H9H7EVEE/p1692286840844139) resolution

### DIFF
--- a/runtime/altair/src/lib.rs
+++ b/runtime/altair/src/lib.rs
@@ -1249,7 +1249,6 @@ impl pallet_block_rewards::Config for Runtime {
 
 // Liquidity rewards
 
-#[cfg(feature = "testnet-runtime")]
 parameter_types! {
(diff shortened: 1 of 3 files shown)


lemunozm:
The main reason to enable liquidity rewards for all chains (removing the `testnet` check feature) is that I think if we start adding migrations feature-based, it could become a real nightmare to determine in future versions what migrations were added for what chains. Because adding liquidity rewards is not a risk, I think it's better to add it for all chains under `altair` runtime.

wischli (approved):
Fully agreeing with your comment at this point (and thus withdrawing my previous request to feature gate it for Algol). Changes LGTM!

PR #1473, opened 2023-07-28
feat: foreign investments

wischli:
# Description

Fixes #https://github.com/centrifuge/centrifuge-chain/issues/1418
Spec: https://centrifuge.hackmd.io/IPtRlOrOSrOF9MHjEY48BA

In case anyone wants to do perform a soft review, I recommend going through the non-exhaustive spec first for a high level understanding of the approach and the pallet structure, especially the [diagrams](https://centrifuge.hackmd.io/IPtRlOrOSrOF9MHjEY48BA#Invest-with-DAI-into-USDC-Pool) for increasing/decreasing investments/redemptions. I have added rustdocs for anything new, so the code should be self explanatory.

## Changes and Descriptions

This section will be finalized once the PR is reviewable. Please check the linked spec.\

* Adds `pallet-foreign-investments`
* Adds `ForeignInvestments` trait
* Adds `StatusNotificationHook` trait required for async handling
* Refactors return type of some investment functions such as `do_collect_{invest, redeem}`
* Moves all investment related traits from `cfg_traits` to `cfg_traits::investments` mod
* Moves some types from `pallet/investment/src/types.rs` to `libs/types/investments.rs`
* Adds `pallet-foreign-investments` to Develop and Altair runtime
* Adds `pallet-order-book` to Altair runtime

## TODO

- [x]  Invest transitions
    - [x]  Increase
    - [x]  Decrease
    - [x]  Order fulfilled
    - [x]  Storage update
    - [x]  Epoch
- [x]  Redeem transitions
    - [x]  Increase
    - [x]  Decrease
    - [x]  Order fulfilled
    - [x]  Epoch
    - [x]  Collect
    - [x]  Storage update
- [x]  Swap orders
    - [x]  Impl `StatusNotificationHook` for acting upon (partially) fulfilled orders
    - [x]  Handle concurrent (potentially conflicting) order updates
- [x]  Collect invest
- [x]  Collect redeem
- [x] Add `CancelInvestOrder` message
- [x] Add `CancelRedeemOrder` message
- [x]  Use `pallet-foreign-investments` as `(Foreign)Investment` provider of `pallet-connectors` instead of `pallet-investments`
   - [x] Add `pallet-foreign-investments` to dev runtime
   - [x] Add `pallet-foreign-investments` to altair runtime
   - [x] Add `pallet-orderbook` to altair runtime (required for foreign investments)
[...]

PR #1509, opened 2023-08-20
ForeignInvestments: Unitary tests & fuzzer

lemunozm:
# Description

Unitary tests for foreign investments

## Changes and Descriptions

- [x] Mock files
- [x] ~~UTs with mocks.~~ It seems like this path is never-ending and collides with integration tests.
- [x] ~~UTs for states.~~ It seems like this path is never-ending and very difficult to maintain if future changes are made.
- [x] Fuzzer. Fix the above problem with programmatic testing in order to check all cases at once with fewer lines.


libs/mock-builder/src/lib.rs
```diff
@@ -337,3 +337,19 @@ macro_rules! execute_call {
...
+	}};
+}
+
+/// Same as [`execute_call`] but using an instanciable storage
+#[macro_export]
+macro_rules! execute_call_i {
+	($input:expr) => {{
+		$crate::execute::<CallIds<T, I>, _, _, _>(|| (), $input)
+	}};
+}
```
lemunozm, line 355: Shall be copied to: https://github.com/foss3/runtime-pallet-library/tree/main/mock-builder

lemunozm:
@wischli I've needed to tweak a bit the `InvestStateConfig` trait here: [c12b203](https://github.com/centrifuge/centrifuge-chain/pull/1509/commits/c12b203570ce088e992794ceba37624b36865358). 

I missed the fact that auto implementations for any `T` do not allow creating a different implementation later for any other type, so we would still need `Config` in tests, and then we are not taking advantage of such a new trait. So, I've added a new type `Of<T>` (that emulates the jargon we use when a type uses `T`), that will be used only for production, which implements `InvestStateConfig` instead of doing it for any `T`. That way we can create other types with other implementations for testing.

Tell me WDYT and if you prefer this change in your PR

wischli:
> @wischli I've needed to tweak a bit the `InvestStateConfig` trait here: [c12b203](https://github.com/centrifuge/centrifuge-chain/pull/1509/commits/c12b203570ce088e992794ceba37624b36865358).
> 
> I missed the fact that auto implementations for any `T` do not allow creating a different implementation later for any other type, so we would still need `Config` in tests, and then we are not taking advantage of such a new trait. So, I've added a new type `Of<T>` (that emulates the jargon we use when a type uses `T`), that will be used only for production and implements `InvestStateConfig` instead of doing it for any `T`. That way, we can create other types with other implementations for testing.
> 
> Tell me WDYT and if you prefer this change in your PR

I suppose it will make maintaining sync between this PR and mine much easier for you. In that case, I would add to to mine as well. Otherwise, we can also leave it untouched in #1473.

lemunozm:
If the intention is to merge your PR today, and no changes there are expected, we can leave it as it is. If you find that you need to make some "big" change, then I think we can put this change just before. By now, we can leave it here 👍🏻

lemunozm:
Just as a thought (mostly to myself). 

Given the number of state combinations we have, maybe the best way to test this is to extract the invariant laws of this (i.e. the amount of money in the system must remain the same, independently of how many swaps are done) and create a fuzzier ensuring the invariant is always fulfilled for all combinations.

pallets/foreign-investments/src/impls/invest.rs
```diff
@@ -1085,3 +1085,279 @@ where
...
+	impl Checker {
+		fn feed_with_step(&mut self, transition: &InvestTransition) {
+			// TODO: mutate checker state to track the happen transition into
+			// something with meaning for the invariant computation
+		}
+
+		fn check_invariant(&self, state: &InvestState) {
+			// TODO: evaluate the invariant for the current state and check with
+			// an assert
+		}
```
lemunozm, line 1286: @wischli could you help me by suggesting a good invariant here? Or with some basic rule I could compute from a list of `transactions` and comparable to the final `state` after applying those transactions?

pallets/foreign-investments/src/impls/redeem.rs
```diff
@@ -584,3 +584,158 @@ where
...
+					swap.amount
+				}
+				Self::RedeemingAndActiveSwapIntoForeignCurrency { swap, .. } => swap.amount,
+				Self::RedeemingAndActiveSwapIntoForeignCurrencyAndSwapIntoForeignDone {
+					swap,
+					..
+				} => swap.amount,
+				_ => 0,
+			}
+		}
```
wischli, line 642: Could make use of `RedeemState::get_active_swap`
```suggestion
		fn get_swap_amount(&self) -> u128 {
			self.get_active_swap().map(|swap| swap.amount).unwrap_or(0)
		}
```

pallets/foreign-investments/src/impls/redeem.rs
```diff
@@ -584,3 +584,158 @@ where
...
+
+		let transitions = [
+			RedeemTransition::IncreaseRedeemOrder(120),
+			RedeemTransition::IncreaseRedeemOrder(60),
+			RedeemTransition::DecreaseRedeemOrder(120),
+			RedeemTransition::DecreaseRedeemOrder(60),
+			RedeemTransition::FulfillSwapOrder(foreign_swap_big),
+			RedeemTransition::FulfillSwapOrder(foreign_swap_small),
+			//RedeemTransition::CollectInvestment(60),
+			//RedeemTransition::CollectInvestment(120),
```
wischli, line 718: I suppose we also want to check `CollectRedemption`.

wischli (dismissed):
Super clean and elegant code as always! Thanks for complementing the existing integration tests with valuable fuzz tests plus providing mock skeletons.

I have three super minor nits which can be ignored. We could merge this PR right now or wait until `RedeemState` fuzzing is implemented. Totally fine with doing this in a subsequent PR.

pallets/foreign-investments/src/impls/invest.rs
```diff
@@ -1117,3 +1117,348 @@ where
...
+				swap: foreign_swap,
+				done_amount,
+			}
+			.transition(increase.clone()),
+			InvestState::ActiveSwapIntoForeignCurrencyAndSwapIntoForeignDoneAndInvestmentOngoing {
+				swap: Swap {
+					amount: foreign_swap.amount - to_foreign(pool_swap.amount),
+					..foreign_swap
+				},
+				done_amount: 100,
```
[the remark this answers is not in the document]
lemunozm: Sure! I was not sure about where that 100 came from at that point, and I left so.

pallets/foreign-investments/src/impls/redeem.rs
lemunozm: Functional and elegant!

pallets/foreign-investments/src/impls/invest.rs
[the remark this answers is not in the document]
wischli: I'd say we should also make the `40` a variable or just replace it with `to_foreign(pool_swap.amount)` as it should be exactly that.

pallets/foreign-investments/src/impls/redeem.rs
lemunozm: Right! I didn't know what I did there to have them commented 😅. 
Updated and working!

lemunozm:
> plus providing mock skeletons.

My first intention was to create individual tests using mocks, but I realized that testing for all transitions could take me months, because I need to evaluate each parameter of each mocked call, and most of what I would check is already covered by integration tests. Anyway, it still can be seen as a future work (or used by other pallets), that's why I leave the mock skeletons there.

pallets/foreign-investments/Cargo.toml
```diff
@@ -30,9 +30,12 @@ sp-std = { git = "https://github.com/paritytech/substrate", default-features = f
 frame-benchmarking = { git = "https://github.com/paritytech/substrate", default-features = false, optional = true, branch = "polkadot-v0.9.38" }
 
 [dev-dependencies]
+rand = "0.8"
```
NunoAlexandre, line 33: No `*/std` needed for either of these 2 pallets?

pallets/foreign-investments/src/impls/invest.rs
```diff
@@ -1117,3 +1117,348 @@ where
...
+
+		fn stable_to_stable(
+			currency_in: Self::Currency,
+			currency_out: Self::Currency,
+			amount: Self::Balance,
+		) -> Result<Self::Balance, Self::Error> {
+			match (currency_out, currency_in) {
+				(CurrencyId::Foreign, CurrencyId::Pool) => Ok(to_pool(amount)),
+				(CurrencyId::Pool, CurrencyId::Foreign) => Ok(to_foreign(amount)),
+				_ => panic!("Same currency"),
```
NunoAlexandre, line 1160: Why panic here instead of returning an Err?

pallets/foreign-investments/src/impls/invest.rs
```diff
@@ -1117,3 +1117,348 @@ where
...
+mod tests {
+	use cfg_traits::IdentityCurrencyConversion;
+	use frame_support::{assert_err, assert_ok};
+	use rand::{rngs::StdRng, seq::SliceRandom, SeedableRng};
+
+	use super::*;
+
+	#[derive(Clone, Copy, PartialEq, Debug)]
+	enum CurrencyId {
+		Foreign,
```
NunoAlexandre, line 1131: `nit` I understand why we call this "Foreign" but the idea that we want this to represent AFAIK is that it should be a stable coin. Not all Foreign assets will be stable coins / pool currencies. So I'd say the code and the mental model one would build from it would be more sensible to name this `StableCoin`.

pallets/foreign-investments/src/impls/invest.rs
```diff
@@ -1117,3 +1117,348 @@ where
...
+		type Balance = u128;
+		type CurrencyConverter = TestCurrencyConverter;
+		type CurrencyId = CurrencyId;
+	}
+
+	type InvestState = super::InvestState<TestConfig>;
+	type InvestTransition = super::InvestTransition<u128, CurrencyId>;
+
+	#[test]
+	fn increase_with_pool_swap() {
```
NunoAlexandre, line 1178: Could we add a comment documentation to these? They are pretty large and for example `let done_amount` is not used until line 1264 which requires a lot of reading up and down to understand what's going on.

pallets/foreign-investments/src/impls/invest.rs
```diff
@@ -1117,3 +1117,348 @@ where
...
+							self.old_state.get_swap_pool_amount(),
+						);
+					}
+				}
+			}
+		}
+	}
+
+	#[test]
+	fn fuzzer() {
```
NunoAlexandre, line 1409: 👌

pallets/foreign-investments/src/tests.rs
```diff
@@ -0,0 +1,328 @@
...
+		MockTokenSwaps::mock_cancel_order(|_| Ok(()));
+		MockTokenSwaps::mock_is_active(|_| true);
+		MockCurrencyConversion::mock_stable_to_stable(move |_, _, _| Ok(amount) /* 1:1 */);
+
+		FulfilledSwapOrderHook::<Runtime>::notify_status_change(
+			order_id,
+			Swap {
+				currency_out: USER_CURR,
+				currency_in: POOL_CURR,
+				amount: amount,
```
NunoAlexandre, line 57:
```suggestion
				amount,
```

Also, _super_ `nit` but for me it would read better by ordering these fields more narratively: `currency_in, `amount_in`, currency_out`.

pallets/foreign-investments/Cargo.toml
lemunozm: Good point! I think so

lemunozm: I usually do not think about `std` feature if the crate doesn't come from substrate 😅

pallets/foreign-investments/src/impls/invest.rs
lemunozm: Because it's a programming error if happens (and it is test code)

pallets/foreign-investments/src/impls/invest.rs
lemunozm: But the pool's currency is also a stable coin, right? I mean, both would be stable coins, the foreign and the pool.

pallets/foreign-investments/src/impls/invest.rs
lemunozm: It's long because of setting mocks. But actually, it only tests one call. I leave those tests "abandoned" because they overlap with integration tests too much and I feel like I would never end up mocking everything. If we want to continue, I think it needs a refactor to factorize the mock configuration part. So I would refactor instead of documenting it (but in case we chose to go in that path)

pallets/foreign-investments/src/impls/invest.rs
lemunozm: Not sure if it should be fuzzier by the way haha 🤔

pallets/foreign-investments/src/tests.rs
lemunozm: I think it's `amount_out`, but I agree with your point 👍🏻

pallets/foreign-investments/src/impls/invest.rs
lemunozm: Wanted to know the opinion of @wischli because I think this was an open topic. Right now, I've chosen those names because everything in the implementation is called as them.

NunoAlexandre: Yeah you got a good point! A pool currency is pegged to a stable coin so by definition it is indeed a stable coin afaik. All good for me 👍

pallets/foreign-investments/src/tests.rs
wischli: The amount of a swap should always be the `amount_in` inherited from the `pallet-orderbook` where an order is placed with the `amount_in`. There have been some discussions around this, but this was the decision.

pallets/foreign-investments/Cargo.toml
wischli: As they are `dev-dependencies`, we don't need to disable default features. If we did, we would have to configure `std` for these of course.

pallets/foreign-investments/src/impls/redeem.rs
```diff
@@ -584,3 +584,137 @@ where
...
+		fn get_swap_amount(&self) -> u128 {
+			self.get_active_swap().map(|swap| swap.amount).unwrap_or(0)
+		}
+
+		fn total(&self) -> u128 {
+			self.get_redeeming_amount() + self.get_done_amount() + self.get_swap_amount()
+		}
+	}
+
+	struct Checker {
```
cdamian, line 628: This looks nice, maybe in the future we can make this a bit more generic so we can use it in other places as well.

pallets/foreign-investments/src/impls/redeem.rs
```diff
@@ -584,3 +584,137 @@ where
...
+					let swap_diff = new_state.get_swap_amount() - self.old_state.get_swap_amount();
+					assert_eq!(swap_diff, swap.amount)
+				}
+			};
+		}
+	}
+
+	#[test]
+	fn fuzzer() {
+		let foreign_swap_big = Swap {
```
cdamian, line 678: I think that at some point we should go one step further and figure out how to fuzz these inputs as well.

cdamian:
Awesome stuff @lemunozm!

pallets/foreign-investments/src/impls/redeem.rs
wischli: For sure! I suppose we should investigate srlabs OS fuzzer: https://github.com/srlabs/substrate-runtime-fuzzer

pallets/foreign-investments/src/impls/redeem.rs
lemunozm: Yeah! As a starting point can probably be factorized between `Invest` & `Checker` types. 
BTW, referring to generic fuzzer. SRLabs gives a talk in Sub0 about its fuzzer which is now open: https://github.com/srlabs/substrate-runtime-fuzzer

It could be interesting to use it. I liked the talk if you have the opportunity to see it.

pallets/foreign-investments/src/impls/redeem.rs
lemunozm: Just reading it after writing my previous message 🥲

pallets/foreign-investments/src/impls/redeem.rs
wischli: I need to rewatch it with full attention once the VODs are out but it was super interesting!

pallets/foreign-investments/src/impls/invest.rs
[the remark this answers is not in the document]
lemunozm: Great!

pallets/foreign-investments/src/impls/invest.rs
lemunozm: It actually is under a `mod test` with #[cfg(tests)] flag. Some lines above.

pallets/foreign-investments/src/impls/invest.rs
lemunozm: Confirmed, it's fuzzer :)

pallets/foreign-investments/src/tests.rs
lemunozm: Just to confirm. From the docs of `Swap` type, currently there are the following commented lines:
```rust
	/// The incoming currency, i.e. the desired one.
	pub currency_in: Currency,
	/// The outgoing currency, i.e. the one which should be replaced.
	pub currency_out: Currency,
	/// The amount of outgoing currency which shall be exchanged.
	pub amount: Balance,
```
Seems like `amount` is the `amount_out` (?)

wischli: Ouch, documentation bug! It should definitely be the amount of incoming currency. Will fix soon. I assumed the swap amounts to be outgoing currency in the early implementation and then realised that the `OrderBook` requires `buy_amount` in incoming currency which we need to follow in foreign ivnestments.

Here's the conversation in the DAO Slack as reference: https://centrifugedao.slack.com/archives/C04HCK98KLL/p1693205602696709

libs/mock-builder/src/lib.rs
lemunozm: Done inf foss3 here: [86e6e69190ae8fdba28d714a6763aabbfc502d44](https://github.com/foss3/runtime-pallet-library/commit/86e6e69190ae8fdba28d714a6763aabbfc502d44)

wischli (approved):
Re-approving!

PR #1510, opened 2023-08-21
altair: feature-gate migrations

NunoAlexandre:
### Context

> This will run on Algol then?
>
>@NunoAlexandre IIRC you already tested that another run does not harm, right?
>
> https://github.com/centrifuge/centrifuge-chain/pull/1502#discussion_r1299732775

Currently, `Algol` is at `1029` going for `2030` while Altair is still at `1027` _maybe_ going to `2030`. This causes the migrations for each runtime to differ since the migration set meant for Altair 1028 was already executed on `Algol` but not yet on Altair.

Instead of relying on version checks to avoid migrations from re-runing and hopefully not break, here we feature gate the migration set to have a dedicated one for `Altair` (**not** `testnet-runtime`) which contains all the migrations applicable since `Altair 2017`, and one for `Algol` with only the migrations applicable since `Algol 1029`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,7 +190,7 @@ dependencies = [
 
 [[package]]
 name = "altair-runtime"
-version = "0.10.29"
+version = "0.10.30"
 dependencies = [
  "axelar-gateway-precompile",
  "cfg-primitives",
--- a/runtime/altair/src/migrations.rs
+++ b/runtime/altair/src/migrations.rs
@@ -37,7 +37,7 @@ pub type UpgradeAltair1030 = (
 	>,
 );
 
-/// The Upgrade set for Algol - it exludes the migrations already executed in
+/// The Upgrade set for Algol - it excludes the migrations already executed in
 /// the side releases that only landed on Algol (1028 & 1029) but not yet on
 /// Altair.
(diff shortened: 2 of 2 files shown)


wischli:
> `1029` going for `2030`

Small typo :) 1029 going for 1030

wischli (approved):
Thanks for fixing this! We could also rename the testnet feature gated `UpgradeAltair1030` to `UpgradeAlgol1030` and either apply `UpgradeAltair1030` or UpgradeAlgol1030` depending on the feature in the `altair` runtime to make it even clearer.

NunoAlexandre:
> Thanks for fixing this! We could also rename the testnet feature gated `UpgradeAltair1030` to `UpgradeAlgol1030` and either apply `UpgradeAltair1030` or `UpgradeAlgol1030` depending on the feature in the `altair` runtime to make it even clearer.

Yeah I thought of doing that but that kinda breaks the flow we initially wanted of having only one "Upgrade*" being exported by the migrations module and used in the lib.

If others also prefer that setup that's fine for me, although I think that given that we are not planning on keeping separate releases for Algol and Altair this should not be an issue after the next Altair release.

mustermeiszer (approved):

PR #1511, opened 2023-08-21
Allow `estimate_gas` for contract creation

branan:
The `estimate_gas` RPC uses the same pathway as a `call` RPC, so we need to allow contract creation in this path so that `estimate_gas` of contract creation will work.

This does not affect any allow/deny-listing of actual contract creation, since that filtering will happen on the transaction itself rather than the `estimate_gas` call.

### DIFF
--- a/runtime/altair/src/lib.rs
+++ b/runtime/altair/src/lib.rs
@@ -2248,17 +2248,40 @@ impl_runtime_apis! {
 		}
 
 		fn create(
-			_from: H160,
-			_data: Vec<u8>,
-			_value: U256,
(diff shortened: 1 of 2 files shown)


mustermeiszer (approved):
Thanks for catching that! Can we upgrade Algol with this today? 

Cc @NunoAlexandre @wischli 

So that this is live when USA wakes up?

wischli (approved):

mustermeiszer (approved):

NunoAlexandre (approved):

PR #1512, opened 2023-08-22
Use latest settlement price by default

offerijns:
# Description

Adds behaviour that external assets can use latest settlement price instead of the oracle price, if no price has been fed. Also adds a borrow restriction that blocks this behaviour.

# Checklist:

### DIFF
--- a/pallets/loans/src/entities/loans.rs
+++ b/pallets/loans/src/entities/loans.rs
@@ -192,7 +192,6 @@ impl<T: Config> ActiveLoan<T> {
 		Ok(ActiveLoan {
 			schedule: info.schedule,
 			collateral: info.collateral,
-			restrictions: info.restrictions,
 			borrower,
 			write_off_percentage: T::Rate::zero(),
 			origination_date: now,
@@ -206,9 +205,11 @@ impl<T: Config> ActiveLoan<T> {
 						info.interest_rate,
 						pool_id,
 						initial_amount.external()?,
+						info.restrictions.borrows == BorrowRestrictions::OraclePriceRequired,
 					)?)
 				}
 			},
+			restrictions: info.restrictions,
 			total_borrowed: T::Balance::zero(),
 			total_repaid: RepaidAmount::default(),
 			repayments_on_schedule_until: now,
--- a/pallets/loans/src/entities/loans.rs
+++ b/pallets/loans/src/entities/loans.rs
@@ -304,6 +304,12 @@ impl<T: Config> ActiveLoan<T> {
 				BorrowRestrictions::FullOnce => {
 					self.total_borrowed.is_zero() && amount.balance()? == max_borrow_amount
 				}
+				BorrowRestrictions::OraclePriceRequired => {
+					match &self.pricing {
+						ActivePricing::Internal(_) => true,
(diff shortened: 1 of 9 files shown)


pallets/loans/src/tests/borrow_loan.rs
```diff
@@ -19,10 +21,7 @@ fn config_mocks(withdraw_amount: Balance) {
 	});
 	MockPrices::mock_register_id(|id, pool_id| {
 		assert_eq!(*pool_id, POOL_A);
-		match *id {
-			REGISTER_PRICE_ID => Ok(()),
-			_ => Err(PRICE_ID_NO_FOUND),
-		}
```
lemunozm, line 25: Why this removal? I think this should work for the case of the price is not found, which means that you've correctly used the `latest_settlement_price` instead

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -126,7 +130,12 @@ impl<T: Config> ExternalActivePricing<T> {
 	}
 
 	pub fn outstanding_principal(&self, pool_id: T::PoolId) -> Result<T::Balance, DispatchError> {
-		let price = T::PriceRegistry::get(&self.info.price_id, &pool_id)?.0;
+		let price = match T::PriceRegistry::get(&self.info.price_id, &pool_id) {
+			Ok(data) => Ok(data.0),
+			Err(_) => self
+				.latest_settlement_price
+				.ok_or(DispatchError::Other("No price found")),
```
lemunozm, line 137: Theoretically, this can never happen, isn't it? That's the reason not to use a defined error, right? If that's the case, I would append to the message that "this can not happen" or something like that.

lemunozm (commented):
The changes looks good!

Just one thing:
- `present_value_changed()` should be modified too in order to skip reading from the oracle and read from `latest_settlement_price` instead.

pallets/loans/src/entities/loans.rs
```diff
@@ -304,6 +304,12 @@ impl<T: Config> ActiveLoan<T> {
 				BorrowRestrictions::FullOnce => {
 					self.total_borrowed.is_zero() && amount.balance()? == max_borrow_amount
 				}
+				BorrowRestrictions::OraclePriceRequired => {
+					match &self.pricing {
+						ActivePricing::Internal(_) => true,
```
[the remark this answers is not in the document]
lemunozm: I think maybe my previous comment is not really required, and returning `true` here is enough. If the issuer creates an internal loan with `OraclePriceRequired` this requirement does not do anything, and we are fine with that.

pallets/loans/src/entities/pricing/external.rs
lemunozm: Actually, I'm pretty sure we can not have an active loan without previously having set a `latest_settlement_price` so I think this variable is not needed to be an `Option` if we initialize the `ExternalActivePricing` struct with the `settlement_price`

lemunozm: Done here: [ae20e0c](https://github.com/centrifuge/centrifuge-chain/pull/1512/commits/ae20e0cbd83437e140b49181cb9af5642d2541cc)

mustermeiszer (approved):
LGTM

PR #1513, opened 2023-08-22
fix: nuke migrations

mustermeiszer:
# Description
Nuke migrations check the storage for a certain prefix. As the `on_chain_strorage_version` is also stored under the prefix, we need to ommit this specific key if we check whether the storage was cleared successfully.

### DIFF
--- a/runtime/common/src/migrations/nuke.rs
+++ b/runtime/common/src/migrations/nuke.rs
@@ -113,7 +113,7 @@ where
 		);
 
 		ensure!(
-			!unhashed::contains_prefixed_key(&pallet_prefix::<Pallet>()),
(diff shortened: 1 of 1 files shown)


wischli (approved):
Thanks for the fix! Happy that the loop is only necessary for the `try-tuntime` part.

NunoAlexandre (approved):
Just to be sure, did you test this locally building with the try-runtime,testnet-runtime features and then run try-runtime against Algol?

mustermeiszer:
See logs I send on slack

NunoAlexandre:
> See logs I send on slack

yeah sorry I saw the PR first 👍

PR #1514, opened 2023-08-22
ci: Publish Algol wasm to GCS

NunoAlexandre:
We cut a release for `Altair 1030 for Algol`, I realised after an `Unauthorised` runtime upgrade attempt, that we don't publish the `Algol` wasm runtime to GCS. 

So what I used for my attempted runtime upgrade was the output wasm of `build-runtime-fast` for Altair instead.

Here we do handle the `Publish to GCS` job for `target == 'build-runtime-testnet' && package == 'altair-runtime'`; in other words, for `Algol`. I also named the wasm stored on GCS `algol-{hash}` to avoid having multiple names to address it by and have it in different directories, etc.


wischli (approved):
LGTM, thanks for fixing!

mustermeiszer (approved):

PR #1517, opened 2023-08-24
Feat: xcm router weights

mustermeiszer:
# Description
Allows to set the weights of the xcm router more granulary. This is especially importatnt to get a realistic max weight for the pov.

### DIFF
--- a/pallets/liquidity-pools-gateway/routers/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/routers/src/lib.rs
@@ -11,6 +11,19 @@
 // GNU General Public License for more details.
 #![cfg_attr(not(feature = "std"), no_std)]
 
+// polkadot/blob/19f6665a6162e68cd2651f5fe3615d6676821f90/xcm/src/v3/mod.rs#
+// L1193 Defensively we increase this value to allow UMP fragments through
+// xcm-transactor to prepare our runtime for a relay upgrade where the xcm
+// instruction weights are not ZERO hardcoded. If that happens stuff will break
+// in our side. Rationale behind the value: e.g. staking unbond will go above
+// 64kb and thus required_weight_at_most must be below overall weight but still
+// above whatever value we decide to set. For this reason we set here a value
+// that makes sense for the overall weight.
+pub const DEFAULT_PROOF_SIZE: u64 = 256 * 1024;
+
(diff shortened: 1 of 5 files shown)


NunoAlexandre (dismissed):
LGTM! Overwriting the router on dev is fine with me 👍

wischli (dismissed):
AFAICT, a migration for upgrading DEV is not required as we can apply the changes of `XcmDomain` to the router by calling `LiquidityPoolsGateway::set_domain_router`. 

If we want to update existing storage without manually overwriting, we would need a migration of course as the newly added `max_pov_size` will default to `0` otherwise.

mustermeiszer:
@cdamian I did adapt our approach. I think we need to use specific fees and weights everywhere instead of setting defaults. LMKWYT.

pallets/liquidity-pools-gateway/routers/src/lib.rs
```diff
@@ -197,19 +210,7 @@ where
 	/// Sets the weight information for the provided XCM domain location, and
 	/// the fee per second for the provided fee asset location.
 	pub fn do_init(&self) -> DispatchResult {
-		pallet_xcm_transactor::Pallet::<T>::set_transact_info(
```
cdamian, line 200: Did we test to confirm that none of these are needed or do we manually need to set these storages after creating a router?

mustermeiszer: Non are needed. The `pallet-xcm-transactor` can work with user provided values and only fills in the `Option` weights and fees if they are not provided with the defaults.

mustermeiszer: I tested that I still sends with `Okay`

mustermeiszer: There is a test for that I mean

cdamian:
I will pick this up and fix the integration tests, will update this branch when ready.

mustermeiszer:
I can not approve. this one.. But looks good.

mustermeiszer:
I guess you can approve @cdamian

cdamian:
@mustermeiszer I'll re-base, approve and enable auto-merge.

cdamian (approved):

PR #1518, opened 2023-08-28
Order book extension

mustermeiszer:
# Description

* Adds admin capabilites for setting valid trading pairs and updating the minimum order amounts for those.
* Adapts `trait TokenSwaps` to allow querying for valid pairs


mustermeiszer:
@wischli is this sufficient?

pallets/order-book/src/lib.rs
```diff
@@ -233,6 +227,40 @@ pub mod pallet {
...
+		ResultQuery<Error<T>::InvalidTradingPair>,
+	>;
+
+	/// Storage of valid order pairs.
+	/// Stores:
+	///  - key1 -> AssetOut
+	///  - key2 -> AssetIn
+	// NOTE: Mirror storage for easier querability.
+	#[pallet::storage]
+	pub type TradingPairOutIn<T: Config> = StorageDoubleMap<
```
wischli, line 254: I am failing to see when we would need this mirrored storage. IMO it can be replaced by calling `TraidngPairInOut::<T>::(out, in)`?

pallets/order-book/src/lib.rs
```diff
@@ -420,6 +466,76 @@ pub mod pallet {
...
+		pub fn rm_trading_pair(
+			origin: OriginFor<T>,
+			asset_in: T::AssetCurrencyId,
+			asset_out: T::AssetCurrencyId,
+		) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
+
+			// We do not check, we just remove as this is an admin action.
+			TradingPairInOut::<T>::remove(asset_in.clone(), asset_out.clone());
+			TradingPairOutIn::<T>::remove(asset_out.clone(), asset_in.clone());
```
wischli, line 504: If we remove a trading pair, we should also cancel all order for these pairs IMO. Removing the trading pair disables placing new orders but still allows old ones to be fulfilled.

pallets/order-book/src/lib.rs
```diff
@@ -420,6 +466,76 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			asset_in: T::AssetCurrencyId,
+			asset_out: T::AssetCurrencyId,
+			min_order: T::Balance,
+		) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
+
+			// try_mutate is a pain with the direct error return. But we do want to only
+			// update if the pair exists.
+			let _old_min_order = TradingPairInOut::<T>::get(&asset_in, &asset_out)?;
```
wischli, line 528: Nit: Can't we use `try_mutate` for that?

pallets/order-book/src/lib.rs
```diff
@@ -681,5 +797,9 @@ pub mod pallet {
 		fn is_active(order: Self::OrderId) -> bool {
 			<Orders<T>>::contains_key(order)
 		}
+
+		fn valid_pair(currency_out: Self::CurrencyId, currency_in: Self::CurrencyId) -> bool {
+			TradingPairInOut::<T>::get(currency_in, currency_out).is_ok()
+		}
```
wischli, line 803: Thanks, this is exactly what I envisioned.

pallets/order-book/src/tests.rs
```diff
@@ -11,13 +11,161 @@
...
+			DispatchError::BadOrigin
+		);
+		assert_eq!(
+			TradingPairInOut::<Runtime>::get(DEV_AUSD_CURRENCY_ID, DEV_USDT_CURRENCY_ID),
+			Err(Error::<Runtime>::InvalidTradingPair)
+		);
+		assert_eq!(
+			TradingPairOutIn::<Runtime>::get(DEV_USDT_CURRENCY_ID, DEV_AUSD_CURRENCY_ID),
+			Err(Error::<Runtime>::InvalidTradingPair)
+		);
```
wischli, line 60: Nit
```suggestion
		assert_noop!(
			TradingPairInOut::<Runtime>::get(DEV_AUSD_CURRENCY_ID, DEV_USDT_CURRENCY_ID),
			Error::<Runtime>::InvalidTradingPair
		);
		assert_noop!(
			TradingPairOutIn::<Runtime>::get(DEV_USDT_CURRENCY_ID, DEV_AUSD_CURRENCY_ID),
			Error::<Runtime>::InvalidTradingPair
		);
```

pallets/order-book/src/tests.rs
```diff
@@ -11,13 +11,161 @@
...
+			DEV_USDT_CURRENCY_ID,
+		));
+		assert_eq!(
+			TradingPairInOut::<Runtime>::get(DEV_AUSD_CURRENCY_ID, DEV_USDT_CURRENCY_ID),
+			Err(Error::<Runtime>::InvalidTradingPair)
+		);
+		assert_eq!(
+			TradingPairOutIn::<Runtime>::get(DEV_USDT_CURRENCY_ID, DEV_AUSD_CURRENCY_ID),
+			Err(Error::<Runtime>::InvalidTradingPair)
+		);
```
wischli, line 79: Nit
```suggestion
		assert_noop!(
			TradingPairInOut::<Runtime>::get(DEV_AUSD_CURRENCY_ID, DEV_USDT_CURRENCY_ID),
			Error::<Runtime>::InvalidTradingPair
		);
		assert_noop!(
			TradingPairOutIn::<Runtime>::get(DEV_USDT_CURRENCY_ID, DEV_AUSD_CURRENCY_ID),
			Error::<Runtime>::InvalidTradingPair
		);
```

pallets/order-book/src/tests.rs
```diff
@@ -11,13 +11,161 @@
...
+}
+
+#[test]
+fn updating_min_order_fails() {
+	new_test_ext().execute_with(|| {
+		assert_noop!(
+			OrderBook::rm_trading_pair(
+				RuntimeOrigin::signed(ACCOUNT_0),
+				DEV_AUSD_CURRENCY_ID,
+				DEV_USDT_CURRENCY_ID,
```
wischli, line 132: The call does not match the title of the test. 
```suggestion
			OrderBook::update_min_order(
				RuntimeOrigin::signed(ACCOUNT_0),
				DEV_AUSD_CURRENCY_ID,
				DEV_USDT_CURRENCY_ID,
				1
```

pallets/order-book/src/tests.rs
```diff
@@ -11,13 +11,161 @@
...
+}
+
+#[test]
+fn updating_min_order_fails_if_not_set() {
+	new_test_ext_no_pair().execute_with(|| {
+		assert_noop!(
+			OrderBook::rm_trading_pair(
+				RawOrigin::Root.into(),
+				DEV_AUSD_CURRENCY_ID,
+				DEV_USDT_CURRENCY_ID,
```
wischli, line 154: The call does not match the title of the test
```suggestion
			OrderBook::update_min_order(
				RuntimeOrigin::signed(ACCOUNT_0),
				DEV_AUSD_CURRENCY_ID,
				DEV_USDT_CURRENCY_ID,
				1
```

wischli (commented):
Thanks for implementing this 🫶! It suffices the OrderBook adjustments needed for #1473 
🚀

IMO, the mirrored storage is not needed. Some tests use the incorrect call (which probably is a result of you working late) and the rest of my comments are nits which can be ignored.

Unfortunately, we also need to bench the new three extrinsics which was the main reason for me not to kick off these changes.

mustermeiszer:
I would not bench them. They are admin extrinsics, and root does not pay. So some basic values for weight should be sufficient.

pallets/order-book/src/lib.rs
mustermeiszer: Frontend. But most not be used. Can also remove

pallets/order-book/src/lib.rs
mustermeiszer: Aren't we also fine with just letting them live till fulfillment or cancel? This would make benching this really hard

pallets/order-book/src/lib.rs
mustermeiszer: See comment above

pallets/order-book/src/lib.rs
wischli: > Aren't we also fine with just letting them live till fulfillment or cancel?

IMO, for now that's totally fine but in the future we need to weight the user experience as well. I just realized that trading pairs are not necessarily undirected edges of a graph, e.g., killing one trading pair `A -> B` does not imply counter orders `B -> A` can still be placed. 

Otherwise (neither `A -> B` nor `B -> A` can be placed), we can assume none of these orders can be fulfilled and the user has to cancel them. This made me realize that we also need a hook for cancelled orders, other wise Foreign Investments are stuck or corresponding `InvestState`/`RedeemState`s even buggy.

pallets/order-book/src/lib.rs
wischli: If we plan on treating trading pairs as directed edges of a graph where each node is a currency id, then we actually should keep this storage. When I commented, I wrongly assumed the graph to be undirected.

pallets/order-book/src/lib.rs
mustermeiszer: Ah, you are right... 

Then I remove the other mapping. But I for now we can assume that there is rather a `placeOrder(a, b)` and a `fulfillOrder(a, b)` flow. So for swapping from `A -> B`, only `(A,B)` needs to be a valid pair.

mustermeiszer: How should the hook look liike? I can add it to this one here.

pallets/order-book/src/lib.rs
mustermeiszer: You are right. Removing it.

wischli (dismissed):
LGTM! 🚀

mustermeiszer:
Also included benches now. will trigger new weights generation if runs through locally.

mustermeiszer:
@wischli I have updated the weights only for dev. Can you take another look once its green?

wischli (dismissed):
Re-approving despite all dev benchmarks having been updated 😅

mustermeiszer:
I needed to rebuild the weights as the auto-build was broken for some. Comparing the weights it does not make much of a difference compared to the cloud machine and it is only dev. So I think it is fine.

pallets/crowdloan-reward/src/lib.rs
```diff
@@ -398,8 +398,9 @@ where
 				.unwrap_or_default()
 				.len() < pallet_vesting::MaxVestingSchedulesGet::<T>::get()
 				.try_into()
-				.unwrap_or(0), /* This is currently a u32, but in case it changes, we will
-			                * fail-safe to zero. */
+				// This is currently a u32, but in case it changes, we will
+				// fail-safe to zero.
+				.unwrap_or(0),
```
mustermeiszer, line 403: Needed for clippy

scripts/tests.sh
```diff
@@ -73,4 +73,4 @@ do
 done
 
 # Run all tests all
-cargo test -workspace -F runtime-benchmarks,try-runtime
+cargo test --workspace -F runtime-benchmarks,try-runtime
```
mustermeiszer, line 76: Was failing locally

wischli: cc @lemunozm

lemunozm: I should have added a typo after testing by myself it worked 😆

mustermeiszer:
@wischli cleaner now ^^

wischli (approved):
Re-re-approving! Thanks for 🧹

PR #1520, opened 2023-08-30
Pools: use `Rate` represented as 18 decimals

lemunozm:
# Description

This PR changes the internal `Rate` representation for pool_system (and derivatives)

## Changes and Descriptions

- Update `Rate` by `Ratio` or `Quantity` where it should be used.
- `Rate` of `pool_system` split into `BalanceRatio` (18 decimals for prices) and `Rate` (27 decimals for interest rates).
- Add official `StorageVersion` for `pool_system` and `investments` and update the value to `1`.
- Nuke `pool_system` storages for `altair` & `centrifuge` runtimes.
- Nuke `investments` storages for `altair` & `centrifuge` runtimes.

--> **IMPORTANT**: Should we nuke also `liquidity_pools` and `liquidity_pools_gateway`? It seems the storage is not still used.

### `algol`
Comand: 
```sh
RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace cargo run --release --features try-runtime try-runtime --runtime target/release/wbuild/altair-runtime/altair_runtime.wasm --chain centrifuge on-runtime-upgrade live --uri wss://fullnode.algol.cntrfg.com:443
```
Output:
```
Nuke-PoolSystem: nuking pallet...
Nuke-PoolSystem: storage cleared successful
Nuke-PoolSystem: iteration result. backend: 7 unique: 7 loops: 7  
Nuke-Investments: nuking pallet...
Nuke-Investments: storage cleared successful
Nuke-Investments: iteration result. backend: 1 unique: 1 loops: 1 
```

### `altair`
Comand: 
```sh
RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace cargo run --release --features try-runtime try-runtime --runtime target/release/wbuild/altair-runtime/altair_runtime.wasm --chain centrifuge on-runtime-upgrade live --uri wss://fullnode.altair.centrifuge.io:443
```
Output:
```
Nuke-PoolSystem: nuking pallet...
Nuke-PoolSystem: storage cleared successful
Nuke-PoolSystem: iteration result. backend: 4 unique: 4 loops: 4  
Nuke-Investments: nuking pallet...
Nuke-Investments: storage cleared successful
Nuke-Investments: iteration result. backend: 0 unique: 0 loops: 0 
```

### `catalyst`
Command:
```sh
[...]

PR #1523, opened 2023-08-31
Add Liquidity Pools related pallets to centrifuge runtime

cdamian:
# Description

Adding the Liquidity Pools pallets to the centrifuge runtime.

## Changes and Descriptions

- [x] Add Liquidity Pool Gateway Pallet to runtime
    - [x] Dummy Liquidity Pool `InboundQueue` that triggers event with info (Runtime struct)
      - NOTE - No event being triggered at the moment
    - [x] Include Liquidity Pools to runtime
        - NOT use LPs as InboundQueue
    - [x] Wrapper `OutboundQueue` that LPs consumes
        - Only allow (`AddCurrency`, `UpdateMember`, `AddPool`, `AddTranche`, `UpdateTrancheTokenPrice`)
        - Uses gateway as real outbound 

# Checklist:

### DIFF
--- a/runtime/centrifuge/Cargo.toml
+++ b/runtime/centrifuge/Cargo.toml
@@ -101,10 +101,10 @@ pallet-evm-chain-id = { git = "https://github.com/PureStake/frontier", default-f
 pallet-evm-precompile-dispatch = { git = "https://github.com/PureStake/frontier", default-features = false, branch = "moonbeam-polkadot-v0.9.38" }
 
 # Our pallets and modules
-liquidity-pools-gateway-routers = { path = "../../pallets/liquidity-pools-gateway/routers", default-features = false }
 cfg-primitives = { path = "../../libs/primitives", default-features = false }
 cfg-traits = { path = "../../libs/traits", default-features = false }
 cfg-types = { path = "../../libs/types", default-features = false }
+liquidity-pools-gateway-routers = { path = "../../pallets/liquidity-pools-gateway/routers", default-features = false }
 pallet-anchors = { path = "../../pallets/anchors", default-features = false }
 pallet-block-rewards = { path = "../../pallets/block-rewards", default-features = false }
 pallet-bridge = { path = "../../pallets/bridge", default-features = false }
@@ -132,8 +132,8 @@ pallet-rewards = { path = "../../pallets/rewards", default-features = false }
 runtime-common = { path = "../common", default-features = false }
 
 # LiquidityPools 3rd-party dependencies
-pallet-xcm-transactor = { git = "https://github.com/PureStake/moonbeam", default-features = false, rev = "00b3e3d97806e889b02e1bcb4b69e65433dd805d" }
 moonbeam-relay-encoder = { git = "https://github.com/PureStake/moonbeam", default-features = false, rev = "00b3e3d97806e889b02e1bcb4b69e65433dd805d" }
+pallet-xcm-transactor = { git = "https://github.com/PureStake/moonbeam", default-features = false, rev = "00b3e3d97806e889b02e1bcb4b69e65433dd805d" }
 xcm-primitives = { git = "https://github.com/PureStake/moonbeam", default-features = false, rev = "00b3e3d97806e889b02e1bcb4b69e65433dd805d" }
 
 # bridge pallets
--- a/runtime/centrifuge/Cargo.toml
+++ b/runtime/centrifuge/Cargo.toml
@@ -101,6 +101,7 @@ pallet-evm-chain-id = { git = "https://github.com/PureStake/frontier", default-f
 pallet-evm-precompile-dispatch = { git = "https://github.com/PureStake/frontier", default-features = false, branch = "moonbeam-polkadot-v0.9.38" }
 
 # Our pallets and modules
+axelar-gateway-precompile = { path = "../../pallets/liquidity-pools-gateway/axelar-gateway-precompile", default-features = false }
 cfg-primitives = { path = "../../libs/primitives", default-features = false }
 cfg-traits = { path = "../../libs/traits", default-features = false }
 cfg-types = { path = "../../libs/types", default-features = false }
@@ -253,6 +254,7 @@ std = [
   "xcm-executor/std",
   "xcm-primitives/std",
   "xcm/std",
+  "axelar-gateway-precompile/std",
 ]
 
 runtime-benchmarks = [
@@ -323,6 +325,7 @@ runtime-benchmarks = [
   "xcm-primitives/runtime-benchmarks",
   "pallet-claims/runtime-benchmarks",
   "pallet-rewards/runtime-benchmarks",
+  "axelar-gateway-precompile/runtime-benchmarks",
 ]
 
 try-runtime = [
@@ -401,6 +404,7 @@ try-runtime = [
   "polkadot-runtime-common/try-runtime",
   "runtime-common/try-runtime",
   "sp-runtime/try-runtime",
+  "axelar-gateway-precompile/try-runtime",
 ]
 
 # A feature that should be enabled when the runtime should be build for on-chain
--- a/runtime/centrifuge/Cargo.toml
+++ b/runtime/centrifuge/Cargo.toml
@@ -55,7 +55,7 @@ sp-version = { git = "https://github.com/paritytech/substrate", default-features
 # frame dependencies
 frame-benchmarking = { git = "https://github.com/paritytech/substrate", default-features = false, optional = true, branch = "polkadot-v0.9.38" }
 frame-executive = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.38" }
-frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.38" }
+frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, features = ["tuples-96"], branch = "polkadot-v0.9.38" }
 frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.38" }
 frame-system-benchmarking = { git = "https://github.com/paritytech/substrate", default-features = false, optional = true, branch = "polkadot-v0.9.38" }
 frame-system-rpc-runtime-api = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.38" }
@@ -101,6 +101,7 @@ pallet-evm-chain-id = { git = "https://github.com/PureStake/frontier", default-f
 pallet-evm-precompile-dispatch = { git = "https://github.com/PureStake/frontier", default-features = false, branch = "moonbeam-polkadot-v0.9.38" }
 
 # Our pallets and modules
+liquidity-pools-gateway-routers = { path = "../../pallets/liquidity-pools-gateway/routers", default-features = false }
 cfg-primitives = { path = "../../libs/primitives", default-features = false }
 cfg-traits = { path = "../../libs/traits", default-features = false }
 cfg-types = { path = "../../libs/types", default-features = false }
@@ -112,10 +113,13 @@ pallet-collator-allowlist = { path = "../../pallets/collator-allowlist", default
 pallet-crowdloan-claim = { path = "../../pallets/crowdloan-claim", default-features = false }
 pallet-crowdloan-reward = { path = "../../pallets/crowdloan-reward", default-features = false }
 pallet-data-collector = { path = "../../pallets/data-collector", default-features = false }
+pallet-ethereum-transaction = { path = "../../pallets/ethereum-transaction", default-features = false }
 pallet-fees = { path = "../../pallets/fees", default-features = false }
 pallet-interest-accrual = { path = "../../pallets/interest-accrual", default-features = false }
 pallet-investments = { path = "../../pallets/investments", default-features = false }
 pallet-keystore = { path = "../../pallets/keystore", default-features = false }
+pallet-liquidity-pools = { path = "../../pallets/liquidity-pools", default-features = false }
+pallet-liquidity-pools-gateway = { path = "../../pallets/liquidity-pools-gateway", default-features = false }
 pallet-liquidity-rewards = { path = "../../pallets/liquidity-rewards", default-features = false }
 pallet-loans = { path = "../../pallets/loans", default-features = false }
 pallet-migration-manager = { path = "../../pallets/migration", default-features = false }
@@ -126,6 +130,11 @@ pallet-pool-system = { path = "../../pallets/pool-system", default-features = fa
 pallet-restricted-tokens = { path = "../../pallets/restricted-tokens", default-features = false }
 pallet-rewards = { path = "../../pallets/rewards", default-features = false }
 runtime-common = { path = "../common", default-features = false }
+
+# LiquidityPools 3rd-party dependencies
+pallet-xcm-transactor = { git = "https://github.com/PureStake/moonbeam", default-features = false, rev = "00b3e3d97806e889b02e1bcb4b69e65433dd805d" }
+moonbeam-relay-encoder = { git = "https://github.com/PureStake/moonbeam", default-features = false, rev = "00b3e3d97806e889b02e1bcb4b69e65433dd805d" }
+xcm-primitives = { git = "https://github.com/PureStake/moonbeam", default-features = false, rev = "00b3e3d97806e889b02e1bcb4b69e65433dd805d" }
 
 # bridge pallets
 chainbridge = { git = "https://github.com/centrifuge/chainbridge-substrate.git", default-features = false, branch = "polkadot-v0.9.38" }
@@ -160,6 +169,8 @@ std = [
   "frame-system-rpc-runtime-api/std",
   "frame-system/std",
   "frame-try-runtime/std",
+  "liquidity-pools-gateway-routers/std",
+  "moonbeam-relay-encoder/std",
   "orml-asset-registry/std",
   "orml-tokens/std",
   "orml-traits/std",
@@ -184,6 +195,7 @@ std = [
   "pallet-democracy/std",
   "pallet-elections-phragmen/std",
   "pallet-ethereum/std",
+  "pallet-ethereum-transaction/std",
   "pallet-evm/std",
   "pallet-evm-precompile-dispatch/std",
   "pallet-evm-chain-id/std",
@@ -191,6 +203,8 @@ std = [
   "pallet-identity/std",
   "pallet-interest-accrual/std",
   "pallet-investments/std",
+  "pallet-liquidity-pools/std",
+  "pallet-liquidity-pools-gateway/std",
   "pallet-liquidity-rewards/std",
   "pallet-loans/std",
   "pallet-keystore/std",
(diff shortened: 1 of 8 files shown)


runtime/centrifuge/src/lib.rs
```diff
@@ -19,13 +19,16 @@
 // Allow things like `1 * CFG`
 #![allow(clippy::identity_op)]
 
+// TODO(cdamian): Use V3?
+use ::xcm::latest::{MultiAsset, MultiLocation};
```
cdamian, line 23: Should we use v3 here as per the update that we have planned?

runtime/centrifuge/src/lib.rs
```diff
@@ -258,7 +262,14 @@ impl Contains<RuntimeCall> for BaseCallFilter {
 				| pallet_xcm::Call::force_subscribe_version_notify { .. }
 				| pallet_xcm::Call::force_unsubscribe_version_notify { .. } => true,
 			},
-			RuntimeCall::EVM(_) => false, // Disable all non-root EVM access
+			// Disable all non-root EVM access
+			RuntimeCall::EVM(_) => false,
+			// We block this call since it includes Moonbeam trait implementations such
+			// as UtilityEncodeCall and XcmTransact that we don't implement and don't want
+			// arbitrary users calling it.
+			RuntimeCall::XcmTransactor(
```
cdamian, line 270: Copied over from the development runtime, I assume this point still makes sense here, please let me know otherwise.

runtime/centrifuge/src/lib.rs
```diff
@@ -418,6 +429,92 @@ impl orml_asset_registry::Config for Runtime {
 	type WeightInfo = ();
 }
 
+parameter_types! {
+	pub LiquidityPoolsPalletIndex: PalletIndex = <LiquidityPools as PalletInfoAccess>::index() as u8;
```
cdamian, line 433: On the development runtime, this is used in the genesis spec to register assets. This won't be needed here since we are going through governance for that right?

runtime/centrifuge/src/lib.rs
```diff
@@ -418,6 +429,92 @@ impl orml_asset_registry::Config for Runtime {
...
+			| LiquidityPoolsMessage::UpdateTrancheTokenPrice { .. } => {
+				<LiquidityPoolsGateway as OutboundQueue>::submit(sender, destination, msg)
+			}
+			_ => Err(DispatchError::Other("unsupported outbound message")),
+		}
+	}
+}
+
+parameter_types! {
+	pub const MaxIncomingMessageSize: u32 = 1024;
```
cdamian, line 490: Should we increase this?

runtime/centrifuge/src/lib.rs
```diff
@@ -1180,6 +1277,37 @@ impl pallet_collator_selection::Config for Runtime {
 	type WeightInfo = weights::pallet_collator_selection::WeightInfo<Self>;
 }
 
+/// Xcm Weigher shared between multiple Xcm-related configs.
+pub type XcmWeigher = xcm_builder::FixedWeightBounds<UnitWeightCost, RuntimeCall, MaxInstructions>;
+
+parameter_types! {
+	// TODO(cdamian): is this correct?
+	// 1 DOT should be enough to cover for fees opening/accepting hrmp channels
+	pub MaxHrmpRelayFee: MultiAsset = (MultiLocation::parent(), 1_000_000_000_000u128).into();
```
cdamian, line 1286: Are we OK with this value?

runtime/centrifuge/src/lib.rs
```diff
@@ -418,6 +429,92 @@ impl orml_asset_registry::Config for Runtime {
...
+
+/// DummyInboundQueue will be used in the first phase of testing in order to
+/// ensure that no incoming messages will be processed.
+pub struct DummyInboundQueue;
+
+impl InboundQueue for DummyInboundQueue {
+	type Message = pallet_liquidity_pools::Message<Domain, PoolId, TrancheId, Balance, Rate>;
+	type Sender = DomainAddress;
+
+	fn submit(_: Self::Sender, _: Self::Message) -> DispatchResult {
```
cdamian, line 513: Given that this dummy queue is an intermediary solution defined here, does it make sense to also publish events here? (Not sure if that's even possible tbh)

runtime/centrifuge/src/lib.rs
```diff
@@ -418,6 +430,92 @@ impl orml_asset_registry::Config for Runtime {
...
+	type AssetRegistry = OrmlAssetRegistry;
+	type Balance = Balance;
+	type CurrencyId = CurrencyId;
+	type ForeignInvestment = Investments;
+	type GeneralCurrencyPrefix = cfg_primitives::liquidity_pools::GeneralCurrencyPrefix;
+	type OutboundQueue = WrapperOutboundQueue;
+	type Permission = Permissions;
+	type PoolId = PoolId;
+	type PoolInspect = PoolSystem;
+	type Rate = Rate;
```
lemunozm, line 449: Very important here to change `Rate` by `Quantity` in case https://github.com/centrifuge/centrifuge-chain/pull/1520 is finally merged:

cdamian: OK, I'm not sure we wanna have that part of this release so let's keep this open until we decide that and then I'll adjust as needed.

Thank you for the catch @lemunozm! <3

cc @mustermeiszer

runtime/centrifuge/src/lib.rs
wischli: IIU, this is the length bound for the number bytes an LP message can have? Right now, the max LP message has [113 bytes](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#TransferTrancheTokens). It appears to me 1024 is a reasonable max bound length we should not reach.

WDYT @mustermeiszer?

runtime/centrifuge/Cargo.toml
mustermeiszer, line 1: We should also add the `axelar-gateway-precompile`. I am using that one for the `(source_chain, source_address)` to `DomainAddress` conversion on the XCM level

runtime/centrifuge/src/lib.rs
mustermeiszer: Yes. I would go away from using latest anywhere if possible. Then we recognize breaking changes.

runtime/centrifuge/src/lib.rs
mustermeiszer: As long as our EVM codes do not call this it is fine.

runtime/centrifuge/src/lib.rs
wischli: If we throw independent of the sender or message, I don't see a reason to dispatch an event. I think it would be helpful to make the error message about more explicit, e.g. `InboundQueue not supported yet`.

runtime/centrifuge/src/lib.rs
```diff
@@ -1845,6 +1969,7 @@ construct_runtime!(
 		EVMChainId: pallet_evm_chain_id::{Pallet, Config, Storage} = 161,
 		BaseFee: pallet_base_fee::{Pallet, Call, Config<T>, Storage, Event} = 162,
 		Ethereum: pallet_ethereum::{Pallet, Config, Call, Storage, Event, Origin} = 163,
+		EthereumTransaction: pallet_ethereum_transaction::{Pallet, Storage, Event<T>} = 164,
```
mustermeiszer, line 1972: This si not 3rd party but ours right?

runtime/centrifuge/Cargo.toml
```diff
@@ -127,6 +131,11 @@ pallet-restricted-tokens = { path = "../../pallets/restricted-tokens", default-f
 pallet-rewards = { path = "../../pallets/rewards", default-features = false }
 runtime-common = { path = "../common", default-features = false }
 
+# LiquidityPools 3rd-party dependencies
+moonbeam-relay-encoder = { git = "https://github.com/PureStake/moonbeam", default-features = false, rev = "00b3e3d97806e889b02e1bcb4b69e65433dd805d" }
```
mustermeiszer, line 135: why do we need that?

runtime/centrifuge/src/lib.rs
mustermeiszer: Nuno wanted to have the registration as part of the chain spec for local chains. Which I think makes sense. Maybe lets leave a comment here for the future ready to use that then.

runtime/centrifuge/src/lib.rs
wischli: AFAICT this is way too high: DOT has [10 decimals](https://wiki.polkadot.network/docs/learn-DOT#polkadot), so this value is actually 100 DOT.

wischli (commented):
I think the DOT decimals are off and we must use another HRMP Encoder.

runtime/centrifuge/src/lib.rs
```diff
@@ -418,6 +430,92 @@ impl orml_asset_registry::Config for Runtime {
...
+	) -> DispatchResult {
+		match msg {
+			LiquidityPoolsMessage::AddCurrency { .. }
+			| LiquidityPoolsMessage::UpdateMember { .. }
+			| LiquidityPoolsMessage::AddPool { .. }
+			| LiquidityPoolsMessage::AddTranche { .. }
+			| LiquidityPoolsMessage::UpdateTrancheTokenPrice { .. } => {
+				<LiquidityPoolsGateway as OutboundQueue>::submit(sender, destination, msg)
+			}
+			_ => Err(DispatchError::Other("unsupported outbound message")),
```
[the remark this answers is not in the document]
mustermeiszer: I think we should not yet allow transfer of tokens untill solidity is safe to receive them correctly.

runtime/centrifuge/src/lib.rs
mustermeiszer: We had time overlap here. Yes, I think it is fine. It should leave enough room for further `20` bytes of an address, `2 * 4` bytes for length and a string that contains the axelar identifier for an domain.

runtime/centrifuge/src/lib.rs
```diff
@@ -1180,6 +1278,37 @@ impl pallet_collator_selection::Config for Runtime {
...
+
+impl pallet_xcm_transactor::Config for Runtime {
+	type AccountIdToMultiLocation = AccountIdToMultiLocation<AccountId>;
+	type AssetTransactor = FungiblesTransactor;
+	type Balance = Balance;
+	type BaseXcmWeight = BaseXcmWeight;
+	type CurrencyId = CurrencyId;
+	type CurrencyIdToMultiLocation = xcm::CurrencyIdConvert;
+	type DerivativeAddressRegistrationOrigin = EnsureRootOr<HalfOfCouncil>;
+	type HrmpEncoder = moonbeam_relay_encoder::westend::WestendEncoder;
```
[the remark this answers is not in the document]
mustermeiszer: Good catch! We should take a dedicated look at `pallet_xcm_transactor` again. I have this as part of our release list. That we check the XCM setup. I am almost certain, that we currently do not allow to transfer tokens out from our chain ^^

runtime/centrifuge/Cargo.toml
cdamian: For the HRMP encoder here - https://github.com/centrifuge/centrifuge-chain/blob/feat/add-lp-to-centrifuge-runtime/runtime/centrifuge/src/lib.rs#L1298, which we might also have to change and use the Polkadot on as William suggested.

runtime/centrifuge/Cargo.toml
cdamian: Will do.

runtime/centrifuge/src/lib.rs
cdamian: Should we go with 1 DOT or less?

runtime/centrifuge/src/lib.rs
cdamian: Yes, this is ours. It's what we added when adding the gateway.

runtime/centrifuge/Cargo.toml
cdamian: Added and updated the Centrifuge precompile as well. PTAL

runtime/centrifuge/src/lib.rs
cdamian: Got it, thank you!

runtime/centrifuge/src/lib.rs
cdamian: Not sure what we could add without bloating the message tbh.

cdamian: Can we just keep it like this since we will remove this eventually?

runtime/centrifuge/src/lib.rs
cdamian: Gonna leave it as it is then, thank you for double-checking!

runtime/centrifuge/src/lib.rs
```diff
@@ -1180,6 +1277,36 @@ impl pallet_collator_selection::Config for Runtime {
 	type WeightInfo = weights::pallet_collator_selection::WeightInfo<Self>;
 }
 
+/// Xcm Weigher shared between multiple Xcm-related configs.
+pub type XcmWeigher = xcm_builder::FixedWeightBounds<UnitWeightCost, RuntimeCall, MaxInstructions>;
+
+parameter_types! {
+	// 1 DOT, which has 10 decimals, should be enough to cover for fees opening/accepting hrmp channels.
+	pub MaxHrmpRelayFee: MultiAsset = (MultiLocation::parent(), 10_000_000_000u128).into();
```
cdamian, line 1285: Updated to 1 DOT, please let me know if we want to change this.

runtime/centrifuge/src/lib.rs
mustermeiszer: Just as this is o=under 3rd party pallets. But no blocker

runtime/centrifuge/src/lib.rs
mustermeiszer: I think it is fine.

mustermeiszer (approved):

runtime/centrifuge/src/lib.rs
```diff
@@ -62,7 +64,7 @@ use orml_traits::{currency::MutationHooks, parameter_type_with_key};
 use pallet_anchors::AnchorData;
 pub use pallet_balances::Call as BalancesCall;
 use pallet_collective::{EnsureMember, EnsureProportionAtLeast, EnsureProportionMoreThan};
-use pallet_ethereum::Transaction as EthereumTransaction;
+use pallet_ethereum::Transaction as EthTransaction;
```
NunoAlexandre, line 67: I would prefer that we just literally refer to it with the pull path instead of aliasing the import. That's also what we do on Altair. It's only referred to in 3 places so it's fine and we save the indirection.

runtime/centrifuge/src/lib.rs
```diff
@@ -418,6 +428,93 @@ impl orml_asset_registry::Config for Runtime {
...
+	type TreasuryAccount = TreasuryAccount;
+	type WeightInfo = ();
+}
+
+type LiquidityPoolsMessage =
+	pallet_liquidity_pools::Message<Domain, PoolId, TrancheId, Balance, Rate>;
+
+/// The WrapperOutboundQueue serves as a filter for outbound LP messages that we
+/// want to allow initially.
+pub struct WrapperOutboundQueue;
```
NunoAlexandre, line 464: Not a blocker, just a suggestion: 

```suggestion
pub struct FilteredOutboundQueue;
```

runtime/centrifuge/src/lib.rs
```diff
@@ -418,6 +428,93 @@ impl orml_asset_registry::Config for Runtime {
...
+		sender: Self::Sender,
+		destination: Self::Destination,
+		msg: Self::Message,
+	) -> DispatchResult {
+		match msg {
+			LiquidityPoolsMessage::AddCurrency { .. }
+			| LiquidityPoolsMessage::UpdateMember { .. }
+			| LiquidityPoolsMessage::AddPool { .. }
+			| LiquidityPoolsMessage::AddTranche { .. }
+			| LiquidityPoolsMessage::UpdateTrancheTokenPrice { .. } => {
```
NunoAlexandre, line 481: So we are not allowing `Executed*` messages? Neither `ScheduleUpgrade`? 🤔

runtime/centrifuge/src/lib.rs
```diff
@@ -418,6 +428,93 @@ impl orml_asset_registry::Config for Runtime {
...
+	type Message = LiquidityPoolsMessage;
+	type Router = liquidity_pools_gateway_routers::DomainRouter<Runtime>;
+	type RuntimeEvent = RuntimeEvent;
+	type RuntimeOrigin = RuntimeOrigin;
+	type WeightInfo = ();
+}
+
+/// DummyInboundQueue will be used in the first phase of testing in order to
+/// ensure that no incoming messages will be processed.
+pub struct DummyInboundQueue;
```
NunoAlexandre, line 507: Why would we want this on Centrifuge?

runtime/centrifuge/src/lib.rs
NunoAlexandre: @cdamian yeah this is fine and we don't plan to use this feature of `xcm-transactor` for now anywyays 👍

runtime/centrifuge/src/lib.rs
```diff
@@ -1819,18 +1940,19 @@ construct_runtime!(
 		BlockRewardsBase: pallet_rewards::<Instance1>::{Pallet, Storage, Event<T>, Config<T>} = 100,
 		BlockRewards: pallet_block_rewards::{Pallet, Call, Storage, Event<T>, Config<T>} = 101,
 		PriceCollector: pallet_data_collector::{Pallet, Storage} = 102,
-		// RESERVED 103 index for LiquidityPools
-		// LiquidityPools: pallet_liquidity_pools::{Pallet, Call, Storage, Event<T>} = 103,
+		LiquidityPools: pallet_liquidity_pools::{Pallet, Call, Storage, Event<T>} = 103,
```
NunoAlexandre, line 1943: 👌

runtime/centrifuge/src/migrations.rs
```diff
@@ -234,12 +234,11 @@ mod asset_registry {
...
 
-			let assets_to_migrate;
-			if is_centrifuge {
-				assets_to_migrate = get_centrifuge_assets();
+			let assets_to_migrate = if is_centrifuge {
+				get_centrifuge_assets()
 			} else {
-				assets_to_migrate = get_catalyst_assets();
-			}
+				get_catalyst_assets()
```
NunoAlexandre, line 240: nice touch 👌

NunoAlexandre (commented):
Some comments, the only blocker afaik is the use of `DummyInboundQueue` on Centrifuge 👀

runtime/centrifuge/src/lib.rs
mustermeiszer: Incoming queue errors out.

runtime/centrifuge/src/lib.rs
mustermeiszer: We do want to check the gatway, but do not want to hit the connectors logic yet.

mustermeiszer: FI is not in yet.

NunoAlexandre: Ah ok I see.

runtime/centrifuge/src/lib.rs
cdamian: I did this because `EthereumTransaction` is how we reference the actual `pallet-ethereum-transaction` that we added.

PR #1524, opened 2023-09-04
fix: max orders per trading pair

mustermeiszer:
# Description
The runtime configuration had an maximum order amounts of `1_000_000` which creates a maximum POV weight of over `8_000_000` and the allowed maximum we have is `5_242_880` so the extrinsics are too heavy for our chain although they are not really consuming that much weight. 

This fixes the maximum allowed orders to `1_000`


runtime/development/src/lib.rs
```diff
@@ -1844,33 +1844,7 @@ impl pallet_transfer_allowlist::Config for Runtime {
...
-const DEV_AUSD_CURRENCY_ID: CurrencyId = CurrencyId::ForeignAsset(2);
-const DEV_USDT_DECIMALS: u128 = 1_000_000;
-const DEV_AUSD_DECIMALS: u128 = 1_000_000_000_000;
-const DEFAULT_DEV_MIN_ORDER: u128 = 5;
-const MIN_DEV_USDT_ORDER: u128 = DEFAULT_DEV_MIN_ORDER * DEV_USDT_DECIMALS;
-const MIN_DEV_AUSD_ORDER: u128 = DEFAULT_DEV_MIN_ORDER * DEV_AUSD_DECIMALS;
-const MIN_DEV_NATIVE_ORDER: u128 = DEFAULT_DEV_MIN_ORDER * CFG;
-
-parameter_type_with_key! {
-		pub MinimumOrderAmount: |pair: (CurrencyId, CurrencyId)| -> Option<Balance> {
```
NunoAlexandre, line 1863: Where was `MinimumOrderAmount` being used? Will check the code but leaving the code here 👀

mustermeiszer: The pallet now supports admin actions that allow to set those values.

NunoAlexandre (approved):
🧹 ✅

wischli (approved):
Thanks for fixing!

PR #1526, opened 2023-09-04
feat: safe XCM version migration + extend SafeCallFilter

wischli:
For both the Centrifuge as well as the Altair runtimes adds:

* Migration to set safe XCM version for Centrifuge and Altair chains to v2 (mirroring behaviour of relay and other parachains)
* `pallet_liquidity_pools_gateway::Call::process_msg` to `SafeCallFilter` which enables this to be executed by xcm `transact`

### DIFF
--- a/runtime/altair/src/migrations.rs
+++ b/runtime/altair/src/migrations.rs
@@ -901,18 +901,15 @@ mod xcm_v2_to_v3 {
 
 	impl OnRuntimeUpgrade for SetSafeXcmVersion {
 		fn on_runtime_upgrade() -> Weight {
-			if VERSION.spec_version > 1030 {
-				return Weight::zero();
-			}
-
-			// Unfortunately, SafeXcmVersion storage is not leaked to runtime
+			// Unfortunately, SafeXcmVersion storage is not leaked to runtime, so we can't
+			// do any pre- or post-upgrade checks
 			PolkadotXcm::force_default_xcm_version(
 				RuntimeOrigin::root(),
 				Some(cfg_primitives::SAFE_XCM_VERSION),
 			)
 			.unwrap_or_else(|_| log::error!("Failed to set safe XCM version on runtime upgrade, requires manual call via governance"));
 
-			RocksDbWeight::get().reads_writes(1, 1)
+			RocksDbWeight::get().writes(1)
 		}
 	}
 }
--- a/runtime/altair/src/migrations.rs
+++ b/runtime/altair/src/migrations.rs
@@ -49,6 +49,8 @@ pub type UpgradeAltair1031 = (
 	crate::XcmpQueue,
 	// Low weight, bumps uninitialized storage version from v0 to v1
 	pallet_xcm::migration::v1::MigrateToV1<crate::Runtime>,
+	// Sets currently unset safe XCM version to v2
+	xcm_v2_to_v3::SetSafeXcmVersion,
 );
 
 /// The Upgrade set for Algol - it excludes the migrations already executed in
@@ -887,6 +889,30 @@ mod pool_system {
 			}
 
 			Ok(())
+		}
+	}
+}
+
+mod xcm_v2_to_v3 {
+	use super::*;
+	use crate::{PolkadotXcm, RuntimeOrigin, VERSION};
+
+	pub struct SetSafeXcmVersion;
+
+	impl OnRuntimeUpgrade for SetSafeXcmVersion {
+		fn on_runtime_upgrade() -> Weight {
+			if VERSION.spec_version > 1030 {
+				return Weight::zero();
+			}
+
+			// Unfortunately, SafeXcmVersion storage is not leaked to runtime
+			PolkadotXcm::force_default_xcm_version(
+				RuntimeOrigin::root(),
+				Some(cfg_primitives::SAFE_XCM_VERSION),
+			)
+			.unwrap_or_else(|_| log::error!("Failed to set safe XCM version on runtime upgrade, requires manual call via governance"));
+
+			RocksDbWeight::get().reads_writes(1, 1)
 		}
 	}
 }
--- a/runtime/altair/src/migrations.rs
+++ b/runtime/altair/src/migrations.rs
@@ -895,7 +895,7 @@ mod pool_system {
 
 mod xcm_v2_to_v3 {
 	use super::*;
-	use crate::{PolkadotXcm, RuntimeOrigin, VERSION};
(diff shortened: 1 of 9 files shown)


NunoAlexandre (approved):
LGTM 👌

cdamian (approved):

runtime/altair/src/migrations.rs
```diff
@@ -890,3 +892,27 @@ mod pool_system {
...
+			// Unfortunately, SafeXcmVersion storage is not leaked to runtime
+			PolkadotXcm::force_default_xcm_version(
+				RuntimeOrigin::root(),
+				Some(cfg_primitives::SAFE_XCM_VERSION),
+			)
+			.unwrap_or_else(|_| log::error!("Failed to set safe XCM version on runtime upgrade, requires manual call via governance"));
+
+			RocksDbWeight::get().reads_writes(1, 1)
+		}
+	}
```
mustermeiszer, line 917: No `before` and `after`?

runtime/altair/src/xcm.rs
```diff
@@ -87,6 +87,9 @@ impl frame_support::traits::Contains<RuntimeCall> for SafeCallFilter {
 			) | RuntimeCall::XcmpQueue(..)
 				| RuntimeCall::DmpQueue(..)
 				| RuntimeCall::Proxy(..)
+				| RuntimeCall::LiquidityPoolsGateway(
+					pallet_liquidity_pools_gateway::Call::process_msg { .. }
```
mustermeiszer, line 91: I am not sure what we are filtering here. All these calls will be allowed, right?

* We need to allow `pallet_liquidity_pools_gateway::Call::process_msg { .. }`
* But I would rather not allow all of the above calls. Only some.

Why do we have them here and why did we choose them? ^^

wischli: These are the allowed calls, exactly. The above ones are mirroring the allowed ones on [Polkadot](https://github.com/paritytech/polkadot/blob/52209dcfe546ff39cc031b92d64e787e7e8264d4/runtime/polkadot/src/xcm_config.rs#L182) and [System parachains](https://github.com/paritytech/cumulus/blob/d4bb2215bb28ee05159c4c7df1b3435177b5bf4e/parachains/runtimes/assets/asset-hub-polkadot/src/xcm_config.rs#L198). We can definitely be more restrictive.

runtime/altair/src/migrations.rs
wischli: See the comment before setting:
> Unfortunately, SafeXcmVersion storage is not leaked to runtime

So we can't check this storage in `pre` and `post` upgrade unfortunately. Since we have no `PolkadotXcm` storage at the moment, we can't batch the runtime upgrade with the execution of `force_default_xcm_version` either unfortunately. 

Either we live with the fact it cannot be checked with `try-runtime` or need to create a post-upgrade referendum to set it. Given that we only do a single write here, I think it is sane to keep the migration. I have tested it locally against `centrifuge-local` @ `1019`.

runtime/altair/src/xcm.rs
mustermeiszer: ok. As we are not a system parachain and do not want to be "controlled" by the relay I think we should revisit that. But lets keep it as is for now.

wischli: I fully agree. In the end, it is much more restrictive than the current setting of allowing everything. I am happy to make another proposal and cut even more. My approach was to make the most open proposal first and then adjust based on your input.

runtime/altair/src/migrations.rs
mustermeiszer: Sorry, yeah, thats true. Lets keep it as is

runtime/altair/src/xcm.rs
mustermeiszer: I still would opt for changing this. I would say remove everything and only add
* calls of `pallet-investments`
* `batch` and `batch-all` of `pallet-utility`
* calls of `pallet-order-book`


Do see anything else that would be needed to generate an xcm message that would allow somebody to invest `natively` from other parachains?

wischli: We must not add `batch` and `batch-all` per definition! The `SafeCallFilter` should only allow calls which have 
> /// 1. Have a fixed weight;
/// 2. Cannot lead to another call being made;
/// 3. Have a defined proof size weight, e.g. no unbounded vecs in call
/// parameters. 

Moreover, as `pallet-investments` is not benchmarked, we must not allow that either.

We can add `pallet-orderbook` calls though.

mustermeiszer: Makes sense. Then lets just allow the one that are safe, add the order book. But we must add `pallet-investments` soon in order to allow somebody from asset hub to invest directly. 

I still would like to remove the system extrinsics though. Although, the origin is not correct. Not a blocker.

wischli: [cea48d4](https://github.com/centrifuge/centrifuge-chain/pull/1526/commits/cea48d46ff5ee057f18c873591c60eb34a934357)

runtime/altair/src/migrations.rs
```diff
@@ -890,3 +892,27 @@ mod pool_system {
...
+	use super::*;
+	use crate::{PolkadotXcm, RuntimeOrigin, VERSION};
+
+	pub struct SetSafeXcmVersion;
+
+	impl OnRuntimeUpgrade for SetSafeXcmVersion {
+		fn on_runtime_upgrade() -> Weight {
+			if VERSION.spec_version > 1030 {
+				return Weight::zero();
+			}
```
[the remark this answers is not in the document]
wischli: [deb9fa3](https://github.com/centrifuge/centrifuge-chain/pull/1526/commits/deb9fa32643fc757dd9cb8e0d85830a7631b11a1)

runtime/altair/src/xcm.rs
wischli: Had to remove the `OrderBook` call again as its not implemented for Altair or Centrifuge Runtime.

mustermeiszer: >We can do this ones we have reasonable upper PoV bounds, most ideally real benchmarks. Beforehand, we are vulnerable to PoV excess and thus stalling.

Yeah, good catch. Thanks for being so thorough!

mustermeiszer (approved):
Will reaprove if lints are in. But I would also be fine with merging and fixing lints locally and merging directly.

mustermeiszer (approved):
Re-approve

PR #1522, opened 2023-08-31
Upgrade 1020

mikiquantum:
# Description

Last items to complete before code freeze:

- [x] XCM v3 Migration
    - [x] Asset Registry (Migrate current registry location entry of `CurrencyId::Native`)
    - [x] ...
- [x] Add Liquidity Pool Gateway Pallet to runtime [PR](https://github.com/centrifuge/centrifuge-chain/pull/1523)
    - [x] Dummy Liquidity Pool `InboundQueue` that triggers event with info (Runtime struct)
    - [x] Include Liquidity Pools to runtime
        - NOT use LPs as InboundQueue
    - [x] Wrapper `OutboundQueue` that LPs consumes
        - Only allow (`AddCurrency`, `UpdateMember`, `AddPool`, `AddTranche`, `UpdateTrancheTokenPrice`)
        - Uses gateway as real outbound 
- [x] Custom LP XCM Converter to GatewayOrigin #1521 
  *NOTE: This one will be merged to main in order to allow upgrading DEV earlier*
  - Will be added with [this PR](https://github.com/centrifuge/centrifuge-chain/pull/1525/files)
- [x] Adapt Liquidity Pools Weights [PR](https://github.com/centrifuge/centrifuge-chain/pull/1521)
- [x] Bump versions

PR #1527, opened 2023-09-04
xcm: Align config across all runtimes

NunoAlexandre:
### Problem

We noticed that the xcm config of the `development` runtime differs in some details from config we use on `Centrifuge` and `Altair`, which currently doesn't handle incoming CFG transfers correctly.

### Changes

- [x] `development`: Fix incoming CFG transfer handling
- [x] `development`: Test outgoing and incoming CFG transfers (e2e tests)
- [x] Align xcm config across all runtimes
- [x] Improve un-anchoring of Centrifuge-native tokens on `CurrencyIdConvert`
   - This allows us to use the entries on the AssetRegistry for both sending and receiving Centrifuge-native tokens without any custom rule for each separate token. On Centrifuge and Altair we supported this but assumed we only had to deal with relative tokens with a `X1` interior; the changes here support any `Xx` interior, making it much more robust and future proof.
 - [x] Clean ups (drop unused traders, etc)


runtime/altair/src/xcm.rs
```diff
@@ -301,6 +279,9 @@ pub type LocationToAccountId = (
 	SiblingParachainConvertsVia<Sibling, AccountId>,
 	// Straight up local `AccountId32` origins just alias directly to `AccountId`.
 	AccountId32Aliases<RelayNetwork, AccountId>,
+	// A wildcard MultiLocation to AccountId conversion for all the other MultiLocations
+	// within the same Relay network.
+	Account32Hash<RelayNetwork, AccountId>,
```
NunoAlexandre, line 284: This is needed to handle origins with a `DescendOrigin` composed multilocation. We added it a while back to `development` to support `Moonbase -> Development` Liquidity Pools messages being sent through XCM. Adding that support here now for Altair and also for Centrifuge.

mustermeiszer: We need to use the compatible solution in the ecosystem for this. 

* https://github.com/moonbeam-foundation/polkadot/blob/d83bb6cc7d7c93ead2fd3cafce0e268fd3f6b9bc/xcm/xcm-builder/src/location_conversion.rs#L25C1-L68C2

I wanted to wait till we can actually use the one provided by parity but given that we need this now lets mimic the one from moonbeam. 

Tests can be copi pasta from the same repo.

runtime/centrifuge/src/xcm.rs
```diff
@@ -300,6 +281,9 @@ pub type LocationToAccountId = (
 	SiblingParachainConvertsVia<Sibling, AccountId>,
 	// Straight up local `AccountId32` origins just alias directly to `AccountId`.
 	AccountId32Aliases<RelayNetwork, AccountId>,
+	// A wildcard MultiLocation to AccountId conversion for all the other MultiLocations
+	// within the same Relay network.
+	Account32Hash<RelayNetwork, AccountId>,
```
mustermeiszer, line 286: See Altair comment

runtime/integration-tests/src/liquidity_pools/pallet/development/transfers.rs
```diff
@@ -0,0 +1,250 @@
...
+
+		assert_ok!(OrmlAssetRegistry::register_asset(
+			RuntimeOrigin::root(),
+			meta.clone(),
+			Some(CurrencyId::Native),
+		));
+	});
+
+	Moonbeam::execute_with(|| {
+		assert_eq!(OrmlTokens::free_balance(cfg_in_sibling, &BOB.into()), 0);
```
mustermeiszer, line 104:
```suggestion
		assert_eq!(Tokens::free_balance(cfg_in_sibling, &BOB.into()), 0);
```

I think we should get used to using `Tokens` everywhere instead of `OrmlTokens`. But just for the future. Need to change to many tests for getting this cleanly in now.

runtime/altair/src/xcm.rs
NunoAlexandre: Thanks @mustermeiszer, TIL.

That's on Moonbeam's fork of Polkadot tho. To use that we would need to copy-paste it and share it across runtimes, which I guess is fine but runs the risk of getting out of sync.

> Tests can be copi pasta from the same repo.

what do you mean?

wischli (commented):
I could not find anything new that stuck out to me.

runtime/altair/src/xcm.rs
mustermeiszer: > That's on Moonbeam's fork of Polkadot tho. To use that we would need to copy-paste it and share it across runtimes, which I guess is fine but runs the risk of getting out of sync.

We can switch to the maintained version once we are on `> polkadot-v0.9.43`.

mustermeiszer: >> Tests can be copi pasta from the same repo.

>what do you mean?

We can copy paste the tests ^^

mustermeiszer: The reason for this is that once users transfer money into their derived accounts we must support this. Moonbeam almost had this issue as Parity switch the approach on short notice...

Hope this explains the reasoning.

NunoAlexandre: Thanks @mustermeiszer, definitely :) it's done now 👍 The CI is acting up but it's looking good on my machine.

libs/primitives/src/lib.rs
```diff
@@ -336,3 +336,162 @@ pub mod liquidity_pools {
...
+pub mod xcm {
+	use codec::{Compact, Encode};
+	use sp_io::hashing::blake2_256;
+	use sp_std::{borrow::Borrow, marker::PhantomData, vec::Vec};
+	use xcm::prelude::{
+		AccountId32, AccountKey20, Here, MultiLocation, PalletInstance, Parachain, X1,
+	};
+	use xcm_executor::traits::Convert;
+
+	/// NOTE: Copied from https://github.com/moonbeam-foundation/polkadot/blob/d83bb6cc7d7c93ead2fd3cafce0e268fd3f6b9bc/xcm/xcm-builder/src/location_conversion.rs#L25C1-L68C2
```
wischli, line 349: I suppose clippy will complain about the url
```suggestion
	/// NOTE: Copied from <https://github.com/moonbeam-foundation/polkadot/blob/d83bb6cc7d7c93ead2fd3cafce0e268fd3f6b9bc/xcm/xcm-builder/src/location_conversion.rs#L25C1-L68C2>
```

wischli (commented):
We should use the same `SafeCallFilter` across all runtimes, i.e. change the `dev` one from `Everything` to the one matching prod runtimes. 

It would be awesome if we could define this (and potentially other XCM configurations) in one place. However, I feel like that's either not simple or even not possible due to the hard coupling with `RuntimeCall`. Moreover, we might diverge in the future, at least between `dev` and prod runtimes.

NunoAlexandre:
> We should use the same `SafeCallFilter` across all runtimes, i.e. change the `dev` one from `Everything` to the one matching prod runtimes.

I guess I would need to rebase this branch off yours? Since here they are all set to `Everything`. If you are doing that change in your branch for Altair & Centrifuge could you also do that for development there? It would offer less friction.
 
> It would be awesome if we could define this (and potentially other XCM configurations) in one place. However, I feel like that's either not simple or even not possible due to the hard coupling with `RuntimeCall`. Moreover, we might diverge in the future, at least between `dev` and prod runtimes.

Yes, indeed. It's been on my list for a while but as you say there are a lot of hard coupling issues that make that end not so trivial to achieve. If you agree, let's get this in soon to get this upgrade going and once things calm down after the summit we work on that.

mustermeiszer (dismissed):
Lets merge without waiting for CI. Then we can run the final CI job once Williams and Cosmins changes are in. 

We should try update Algol and Catalyst today in the evening.

wischli (approved):
LGTM! 

Sorry about the `SafeCallFilter` request, I forgot this PR is based against `main`. Of course, we can normalize that as part of #1522.

We can worry about the XCM unifications later for sure. Great that you have it on your list 😊

wischli:
> Lets merge without waiting for CI. Then we can run the final CI job once Williams and Cosmins changes are in.

Since this is based against `main`, we shouldn't merge with breaking CI. IMO we should merge this PR into #1522 and then can totally ignore CI here.

wischli (approved):
Re-approving

NunoAlexandre:
@mustermeiszer @wischli the only CI job failing is `cargo docs` so imho we can just merge it in. I've fixed the warning in the last commit anyways.

mustermeiszer (approved):
LGTM!

PR #1528, opened 2023-09-04
fix: asset metadata migration xcm v3

wischli:
* Refactors and fixes asset metadata migration to xcm v3

## TODO

- [ ] Apply to Altair runtime


runtime/centrifuge/src/migrations.rs
```diff
@@ -195,404 +208,316 @@ mod asset_registry {
...
+							X3(Parachain(1000), PalletInstance(50), GeneralIndex(1337)),
+						))),
+						additional: CustomMetadata {
+							mintable: false,
+							permissioned: false,
+							pool_currency: true,
+							// TODO(@review): Should probably be LiquidityPoolsTransferability?
+							transferability: CrossChainTransferability::Xcm(XcmMetadata {
+								fee_per_second: None,
+							}),
```
mustermeiszer, line 396: No, that is perfect as is. We only want to receive that via xcm into accounts for now. I think we first need to go over the implications of having both transferabilities enabled for a single asset.

runtime/common/src/migrations/asset_registry_xcmv3.rs
```diff
@@ -0,0 +1,232 @@
...
+	Assets: AssetsToMigrate,
+{
+	fn on_runtime_upgrade() -> Weight {
+		log::info!("AssetRegistryMultilocationToXCMV3: on_runtime_upgrade: started");
+		// Complexity: 2 reads
+		let (loc_count, meta_count) = Self::get_key_counts();
+
+		// Complexity: O(loc_count) writes
+		let result =
+			orml_asset_registry::LocationToAssetId::<T>::clear(loc_count.saturating_add(2), None);
```
mustermeiszer, line 68: Why the plus 2?

runtime/common/src/migrations/asset_registry_xcmv3.rs
```diff
@@ -0,0 +1,232 @@
...
+		}
+		log::info!(
+            "AssetRegistryMultilocationToXCMV3: LocationToAssetId clearing iteration result. backend: {} unique: {} loops: {}",
+            result.backend,
+            result.unique,
+            result.loops,
+        );
+
+		// Complexity: O(meta_count) writes
+		let result = orml_asset_registry::Metadata::<T>::clear(meta_count.saturating_add(1), None);
```
mustermeiszer, line 86: Why the plus 1?

runtime/common/src/migrations/asset_registry_xcmv3.rs
mustermeiszer, line 1: Sick!

mustermeiszer (commented):
Will take a final look once Altair is in! Migrations looks like it was really tricky, thanks that looks great now!

runtime/centrifuge/src/migrations.rs
```diff
@@ -195,404 +208,316 @@ mod asset_registry {
...
+						name: b"Centrifuge".to_vec(),
+						symbol: b"CFG".to_vec(),
+						existential_deposit: 1_000_000_000_000u128,
+						location: Some(VersionedMultiLocation::V3(MultiLocation::new(
+							0,
+							Junctions::X1(GeneralKey {
+								length: 2,
+								data: gk,
+							}),
+						))),
```
mustermeiszer, line 259: I think this must be 

```suggestion
						location: Some(VersionedMultiLocation::V3(MultiLocation::new(
							1,
							Junctions::X2(
							    Parachain(ParachainInfo::parachain_id().into()),
							    GeneralKey {
								length: 2,
								data: gk,
							}),
						))),
```

mustermeiszer: cc @NunoAlexandre as I think in your PR you came to the conclusion that we can better go with the absolut locations in the registry.

runtime/centrifuge/src/migrations.rs
```diff
@@ -195,404 +208,316 @@ mod asset_registry {
...
+
+			// 35fd988a3d77251b19d5d379a4775321
+			let tranche_id_bytes = &[
+				53u8, 253u8, 152u8, 138u8, 61u8, 119u8, 37u8, 27u8, 25u8, 213u8, 211u8, 121u8,
+				164u8, 119u8, 83u8, 33u8,
+			];
+			let mut tranche_id = [0u8; 16];
+			tranche_id[..tranche_id_bytes.len()].copy_from_slice(tranche_id_bytes);
+
+			vec![
```
mustermeiszer, line 432: Can we also register `Native` here? Wdyt? No hard requirement.

mustermeiszer (changes requested):
Native currency needs absolut path.

runtime/common/src/migrations/asset_registry_xcmv3.rs
wischli: Sorry, this can be removed. It was needed when we read the counts via the `iter_keys` API due to key decoding failures. However, the current approach ignores decoding and thus catches all keys.

runtime/common/src/migrations/asset_registry_xcmv3.rs
wischli: See [above](https://github.com/centrifuge/centrifuge-chain/pull/1528/files#r1315426451).

runtime/centrifuge/src/migrations.rs
wischli: Sure! I suppose we want to register all Centrifuge assets actually?

runtime/altair/src/xcm.rs
```diff
@@ -69,24 +69,17 @@ impl frame_support::traits::Contains<RuntimeCall> for SafeCallFilter {
...
-				) | RuntimeCall::CollatorSelection(
-				pallet_collator_selection::Call::set_desired_candidates { .. }
-					| pallet_collator_selection::Call::set_candidacy_bond { .. }
-					| pallet_collator_selection::Call::register_as_candidate { .. }
-					| pallet_collator_selection::Call::leave_intent { .. },
-			) | RuntimeCall::XcmpQueue(..)
+				) | RuntimeCall::XcmpQueue(..)
 				| RuntimeCall::DmpQueue(..)
 				| RuntimeCall::Proxy(..)
+				| RuntimeCall::LiquidityPoolsGateway(
```
NunoAlexandre, line 80: These changes to the other calls are unrelated to this PR and could introduce undesired behaviour given that we don't have tests in place to properly assess this. Not a blocker tho

runtime/centrifuge/src/migrations.rs
NunoAlexandre: I think `Native` is sufficient, no?

mustermeiszer: Not sure TBH. Catalyst is not connected to anything and we control Council. So I would say we can also do it later. Are there any other reasons I am missing? Whatever you think is best and easiest.

mustermeiszer (approved):
Removing change request as it is addressed. Just so that I am not the blocking person anymore.

runtime/altair/src/xcm.rs
wischli: I honestly don't know why this diff still exists. This change was merged yesterday already with #1526. I previously had this PR based against the #1526 branch.

NunoAlexandre:
> @NunoAlexandre
> 
> > I was thinking that we would "just" basically call omrl_asset_registry::update_asset for each known CurrencyId with the updated, xcm v3 multilocation, instead of clearing all the different storages and registering the tokens anew. That still seems much simpler but now we have this already in place so that's not a problem for me 👍
> 
> Same here. I took over this approach from @mikiquantum and wanted to change as little as possible apart from making it runtime agnostic.
> 
> I suppose we don't need to run the `CrossChainTransferabilityMigration` then as we hardcode all metadata. Sorry for the wasted time on that beatiful migration which is superseeded by this ugly hardcoded one.

Ah, that makes sense! :) 

And no worries, what is important is that we get this right 🏄‍♂️

wischli:
FYI, I cleaned up and checked all migrations against Algol, Altair, Catalyst and Centrifuge with `try-runtime`.

runtime/centrifuge/src/migrations.rs
```diff
@@ -9,36 +9,34 @@
...
+	// At minimum, bumps storage version from 0 to 1
+	runtime_common::migrations::nuke::Migration<crate::Investments, crate::RocksDbWeight, 0>,
+	// Funds pallet_rewards::Instance2 account with existential deposit
 	pallet_rewards::migrations::new_instance::FundExistentialDeposit<
 		crate::Runtime,
 		pallet_rewards::Instance2,
 		crate::NativeToken,
 		crate::ExistentialDeposit,
 	>,
-	asset_registry::AssetRegistryMultilocationToXCMV3<crate::Runtime>,
```
wischli, line 41: Note: Replaces with `runtime_common::migrations::asset_registry_xcmv3::Migration`

runtime/centrifuge/src/migrations.rs
```diff
@@ -9,36 +9,34 @@
...
-	weights::Weight,
-};
-use sp_std::{vec, vec::Vec};
-use xcm::{v3::prelude::*, VersionedMultiLocation};
-
-use crate::{LiquidityPoolsPalletIndex, RocksDbWeight, Runtime};
+use frame_support::{traits::OnRuntimeUpgrade, weights::Weight};
 
 pub type UpgradeCentrifuge1020 = (
-	asset_registry::CrossChainTransferabilityMigration,
```
wischli, line 30: Note: Not necessary due to hardcoded metadata from `runtime_common::migrations::asset_registry_xcmv3::Migration`

runtime/altair/src/migrations.rs
```diff
@@ -9,40 +9,40 @@
...
 
 /// The migration set for Altair 1031 @ Kusama. It includes all the migrations
 /// that have to be applied on that chain, which includes migrations that have
 /// already been executed on Algol (1028 & 1029).
 #[cfg(not(feature = "testnet-runtime"))]
 pub type UpgradeAltair1031 = (
-	asset_registry::CrossChainTransferabilityMigration,
-	// TODO: This migration errors out against Altair
+	// FIXME: This migration fails to decode 4 entries against Altair
 	// orml_tokens_migration::CurrencyIdRefactorMigration,
```
wischli, line 20: Note: Needs fixing before Altair upgrade!

runtime/altair/src/migrations.rs
```diff
@@ -9,40 +9,40 @@
...
-	pool_system::MigrateAUSDPools,
```
wischli, line 34: Note: Superseeded by `runtime_common::migrations::nuke::Migration<crate::PoolSystem, crate::RocksDbWeight, 0>`

runtime/altair/src/migrations.rs
```diff
@@ -9,40 +9,40 @@
...
+	// At minimum, bumps storage version from 0 to 1
+	runtime_common::migrations::nuke::Migration<crate::Investments, crate::RocksDbWeight, 0>,
+	// Funds pallet_rewards::Instance2 account with existential deposit
 	pallet_rewards::migrations::new_instance::FundExistentialDeposit<
 		crate::Runtime,
 		pallet_rewards::Instance2,
 		crate::NativeToken,
 		crate::ExistentialDeposit,
 	>,
-	asset_registry::AssetRegistryMultilocationToXCMV3<crate::Runtime>,
```
wischli, line 45: Note: Replaced with `runtime_common::migrations::asset_registry_xcmv3::Migration`

runtime/altair/src/migrations.rs
```diff
@@ -9,40 +9,40 @@
...
-	runtime_common::migrations::nuke::Migration<crate::Investments, RocksDbWeight, 0>,
+	// At minimum, bumps storage version from 1 to 2
+	runtime_common::migrations::nuke::Migration<crate::Loans, crate::RocksDbWeight, 1>,
+	// At minimum, bumps storage version from 0 to 3
+	runtime_common::migrations::nuke::Migration<crate::InterestAccrual, crate::RocksDbWeight, 0>,
+	// At minimum, bumps storage version from 0 to 1
+	runtime_common::migrations::nuke::Migration<crate::PoolSystem, crate::RocksDbWeight, 0>,
+	// At minimum, bumps storage version from 0 to 1
+	runtime_common::migrations::nuke::Migration<crate::Investments, crate::RocksDbWeight, 0>,
+	// Funds pallet_rewards::Instance2 account with existential deposit
```
wischli, line 29: Note: Storage is empty but this nuke still bumps `StorageVersion`.

runtime/altair/src/migrations.rs
```diff
@@ -58,676 +58,427 @@ pub type UpgradeAltair1031 = (
 /// Altair.
 #[cfg(feature = "testnet-runtime")]
 pub type UpgradeAltair1031 = (
-	// TODO: Verify that these ones can be removed from Algol upgrade, since the first 2 fail
-	// runtime_common::migrations::nuke::Migration<crate::Loans, RocksDbWeight, 1>,
-	// runtime_common::migrations::nuke::Migration<crate::InterestAccrual, RocksDbWeight, 0>,
```
wischli, line 63: Note: No nuke required on `Algol` for `Loans` and `InterestAccrual` as these are already at the latest `StorageVersion`

runtime/centrifuge/src/migrations.rs
```diff
@@ -9,36 +9,34 @@
...
+		3,
+	>,
+	// At minimum, bumps storage version from 1 to 2
+	runtime_common::migrations::nuke::Migration<crate::Loans, crate::RocksDbWeight, 1>,
+	// At minimum, bumps storage version from 0 to 3
+	runtime_common::migrations::nuke::Migration<crate::InterestAccrual, crate::RocksDbWeight, 0>,
+	// At minimum, bumps storage version from 0 to 1
+	runtime_common::migrations::nuke::Migration<crate::PoolSystem, crate::RocksDbWeight, 0>,
+	// At minimum, bumps storage version from 0 to 1
+	runtime_common::migrations::nuke::Migration<crate::Investments, crate::RocksDbWeight, 0>,
```
wischli, line 32: Note: Storage is empty but this nuke still bumps `StorageVersion`.

runtime/altair/src/migrations.rs
```diff
@@ -58,676 +58,427 @@ pub type UpgradeAltair1031 = (
...
+						decimals: 18,
+						name: b"Altair".to_vec(),
+						symbol: b"AIR".to_vec(),
+						existential_deposit: 1_000_000_000_000u128,
+						location: Some(VersionedMultiLocation::V3(MultiLocation::new(
+							0,
+							Junctions::X1(GeneralKey {
+								length: 2,
+								data: gk,
+							}),
```
mustermeiszer, line 145: Absolut path needed

runtime/altair/src/migrations.rs
```diff
@@ -58,676 +58,427 @@ pub type UpgradeAltair1031 = (
...
-		match sp_io::storage::next_key(&next_key) {
-			Some(key) if !key.starts_with(prefix) => break count,
-			Some(key) => {
-				next_key = key;
-				count += 1;
-			}
-			None => {
-				break count;
-			}
+		pub fn get_algol_assets() -> Vec<(
```
mustermeiszer, line 229: missing native.

mustermeiszer (changes requested):
Altair native wrong location. Otherwise fine.

runtime/altair/src/migrations.rs
wischli: [a9caf16](https://github.com/centrifuge/centrifuge-chain/pull/1528/commits/a9caf160523db0249e71a6437a036820e7e5741f)

runtime/altair/src/migrations.rs
wischli: [1e8c597](https://github.com/centrifuge/centrifuge-chain/pull/1528/commits/1e8c597f69e5ea76aca4179e6250867687a21f76)

mustermeiszer (approved):
LGTM! Thanks for the effort, that was quite a pice ^^

wischli:
> Thanks for the effort, that was quite a pice ^^

Indeed... Migrations always multiply the usual required time by orders of magnitude. At least we can be sure all migrations are now properly configured - 🤞 it works out in practice.

PR #1531, opened 2023-09-05
Upgrade1020 gateway changes

cdamian:
# Description

Env prep discussion follow-ups:
- Drop the EVMChain enum and replace it with BoundedVec.
- Add Sender type to LP gateway config.

# Checklist:


pallets/liquidity-pools-gateway/routers/src/lib.rs
```diff
@@ -60,6 +60,9 @@ type CurrencyIdOf<T> = <T as pallet_xcm_transactor::Config>::CurrencyId;
 type MessageOf<T> = <T as pallet_liquidity_pools_gateway::Config>::Message;
 type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
 
+/// Maximum size allowed for a byte representation of an EVM chain string.
+pub const MAX_EVM_CHAIN_SIZE: u32 = 16;
```
cdamian, line 64: Given that we should support the strings that Axelar is using, I think this might suffice, please let me know otherwise.

runtime/common/src/gateway.rs
```diff
@@ -0,0 +1,36 @@
...
+use crate::account_conversion::AccountConverter;
+
+pub struct GatewayAccountProvider<T>(core::marker::PhantomData<T>);
+
+impl<T> GatewayAccountProvider<T>
+where
+	T: pallet_evm_chain_id::Config + parachain_info::Config,
+{
+	pub fn get_gateway_account() -> AccountId {
+		let sender_account: AccountId =
```
wischli, line 28: IIUC, that's the parachain's sovereign account?

wischli (approved):
LGTM but I am missing the context, so another review would be much appreciated.

runtime/common/src/gateway.rs
cdamian: Yes, after talking to @mustermeiszer we figured out that that's the account that we want to derive and then use as the Gateway sender account. We'll have to ensure that both this account and the truncated version of it have enough funds for interacting with the XCM or EVM routers.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -96,6 +96,11 @@ pub mod pallet {
 		/// Maximum size of an incoming message.
 		#[pallet::constant]
 		type MaxIncomingMessageSize: Get<u32>;
+
+		/// The sender account that will be used in the OutboundQueue
+		/// implementation.
+		#[pallet::constant]
+		type Sender: Get<Self::AccountId>;
```
NunoAlexandre, line 103: I think the name and the intention are somewhat ambiguous. This is used to specify an account that will be paying for fees, a `FeePayer`. It's not specifying who is sending the message per se. This is a naming detail but getting these small details slightly tweaked and accumulated make a difference in the big picture mental model we make out of these dozens and complex pallets 🙏

runtime/common/src/gateway.rs
NunoAlexandre: Yeah it looks like a mimic of that. I don't think we want that, given a recent discussion we had for the `schedule_upgrade` extrinsic of the `LiquidityPools` pallet, where we concluded the treasury account would cover for those fees.

pallets/liquidity-pools-gateway/routers/src/lib.rs
NunoAlexandre: Note that this is a string that _only_ makes sense for Axelar, AFAIK. I see value in being more specific regarding that, either in the naming of this const or in the place where it is defined.

cdamian: Will change, thank you!

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: In the case of XCM - yes, it's a fee payer, in the EVM case - it's both the fee payer and the sender of the message. As far as routers are concerned, it's also sender. IMO calling it a `FeePayer` is even more confusing.

runtime/common/src/gateway.rs
NunoAlexandre: > Yes, after talking to @mustermeiszer we figured out that that's the account that we want to derive and then use as the Gateway sender account. We'll have to ensure that both this account and the truncated version of it have enough funds for interacting with the XCM or EVM routers.

@cdamian but then shouldn't we make it a pallet derived account, instead of this mimic of the sovereign account?

pallets/liquidity-pools-gateway/routers/src/mock.rs
```diff
@@ -129,6 +129,10 @@ impl TryConvert<(Vec<u8>, Vec<u8>), DomainAddress> for MockOriginRecovery {
 	}
 }
 
+parameter_types! {
+	pub Sender: AccountId32 = AccountId32::from(H256::from_low_u64_be(1).to_fixed_bytes());
```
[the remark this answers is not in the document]
mustermeiszer: We will use the `AccountKey20` derived account of our parachains sovereign account. Need to fund that. The reason is that with this approach, but the xcm sender as also the EVM sender are the same.

mustermeiszer (commented):
Need to agree in final gatway sender. But otherwise looks good.

runtime/common/src/gateway.rs
NunoAlexandre: Ah I see, good point 👍

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -423,7 +432,7 @@ pub mod pallet {
 
 			let router = DomainRouters::<T>::get(destination).ok_or(Error::<T>::RouterNotFound)?;
 
-			router.send(sender, msg)
+			router.send(T::Sender::get(), msg)
```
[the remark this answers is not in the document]
cdamian: Will do.

mustermeiszer (approved):

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -115,6 +121,13 @@ pub mod pallet {
 
 		/// A relayer was removed.
 		RelayerRemoved { relayer: DomainAddress },
+
+		/// An outbound message has been submitted.
+		OutboundMessageSubmitted {
+			sender: T::AccountId,
+			message: Vec<u8>,
```
mustermeiszer, line 128: Why not use a type here? Easier to read in the explorer.

mustermeiszer (approved):
Nit. Otherwise looks great! Can merge that

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Like `Bytes`?

PR #1532, opened 2023-09-05
Feat: upgrade1020 misc 2

mustermeiszer:
# Description
Miscellaneous changes for the upgrade. Details see changes

## Changes and Descriptions
* new `EnsureAccoutOrRootOr` struct - allows to define an admin account for certain calls
  * we need that to block the gatway logic faster than council could do (temporary only until we reach stability on the Centrifuge and Soidity side)
* Bumb `Development` version to `10.21` 
* Adapt weights of pallets 

# Checklist:


runtime/common/src/lib.rs
```diff
@@ -419,3 +419,162 @@ pub mod xcm_transactor {
...
+		dispatch::RawOrigin,
+		traits::{EitherOfDiverse, EnsureOrigin},
+	};
+	use frame_system::EnsureRoot;
+	use sp_core::Get;
+
+	type EnsureAccountOrRootOr<Account, O> =
+		EitherOfDiverse<EitherOfDiverse<EnsureAccount<Account>, EnsureRoot<AccountId>>, O>;
+
+	pub struct EnsureAccount<Account>(PhantomData<Account>);
```
NunoAlexandre, line 437: @mustermeiszer Can't we just use `EnsureSignedBy` instead of building this custom `EnsureAccount`?

mustermeiszer: Ah nicel Changed.

mustermeiszer:
This builds with `--release` and with `--release --features runtime-benchmarks`.  I will trigger the benchmarks from this branch here as it contains all the latest changes and we must not wait for benchmarks then tomorrow.

pallets/liquidity-pools-gateway/src/weights.rs
```diff
@@ -21,28 +21,80 @@ pub trait WeightInfo {
...
+		//       * This weight      { ref_time: 4333558693, proof_size:   91070 }
+		//       * Maximum weight { ref_time: 500000000000, proof_size: 5242880 }
+		//
+		Weight::from_parts(78_019_565, 19974)
+			.saturating_add(Weight::from_ref_time(38_884_782).saturating_mul(N))
+			.saturating_add(RocksDbWeight::get().reads(8))
+			.saturating_add(RocksDbWeight::get().reads((7_u64).saturating_mul(N)))
+			.saturating_add(RocksDbWeight::get().writes(8))
+			.saturating_add(RocksDbWeight::get().writes((6_u64).saturating_mul(N)))
+			.saturating_add(Weight::from_proof_size(17774).saturating_mul(N))
```
wischli, line 98: Note: I made sure thus default weight and PoV size is still executable. It is:
Screenshot 2023-09-06 at 08 24 39

wischli (approved):
LGTM! 🚀

pallets/liquidity-pools-gateway/src/weights.rs
mustermeiszer: I did too ^^

mustermeiszer: No, I think we should increase it.

mustermeiszer:
Integration tests and test go through locally I will merge that now. We can run final CI on feature branch then.

PR #1534, opened 2023-09-07
chore: remove Centrifuge 1020 & Algol 1031 migrations

wischli:
# Description

Part of the release process is cleaning up past migrations. Since Altair still needs to be migrated, we have to keep the `common/src/migrations/asset_registry_xcmv3.rs` migration.

* Removes all Centrifuge migrations as both Catalyst as well as Centrifuge have been upgraded
  * Bumps migration tuple to `UpgradeCentrifuge1021`
* Removes Algol `1031` migrations and bumps migration tuple to `UpgradeAltair1032`


NunoAlexandre (commented):
LGTM but I would prefer to not merge this yet until Centrifuge is upgraded, shall we find some issue in this time in between and need to push some hotfix. We can also merge it an git revert it but that might bite our 🍑 .

wdyt?

wischli:
> LGTM but I would prefer to not merge this yet until Centrifuge is upgraded, shall we find some issue in this time in between and need to push some hotfix. We can also merge it an git revert it but that might bite our 🍑 .

IMO, a hotfix should always be built on top of the release head and then be merged into `main`. Otherwise your argument could also be applied new features which should be held back. It will be quite common that new features are merged into main while the Centrifuge upgrade is not close to being executed. Take the 0.10.27 as an example: We merged the 0.9.38 upgrade immediately after releasing.

NunoAlexandre:
> > LGTM but I would prefer to not merge this yet until Centrifuge is upgraded, shall we find some issue in this time in between and need to push some hotfix. We can also merge it an git revert it but that might bite our 🍑 .
> 
> IMO, a hotfix should always be built on top of the release head and then be merged into `main`. Otherwise your argument could also be applied new features which should be held back. It will be quite common that new features are merged into main while the Centrifuge upgrade is not close to being executed. Take the 0.10.27 as an example: We merged the 0.9.38 upgrade immediately after releasing.

I get your point! Adding new features doesn't affect the correctness of the storage like in this case but your suggestion to apply any hotfix deemed necessary on top of the release head is fine and suffices 👍

NunoAlexandre (approved):

PR #1535, opened 2023-09-07
liquidityPools: Clean up orphan storage and events

NunoAlexandre:
We forgot to remove these when we introduced the new gateway, deleting them here. No storage migrations needed.


cdamian (approved):
Totally forgot about this, thank you very much for taking care! <3

wischli (approved):
LGTM! Thanks for cleaning up.

PR #1536, opened 2023-09-08
fix: ethereum transaction pallet must forward to pallet-ethereum::transact

mustermeiszer:
# Description
Axelar depends on the `pallet_ethereum::Event<T>::Transact{..}` and we were previsoulsy shortcircuting the logic, which lead to the event missing. This change fixes that behaviour and just generates the same origin as the runtime and forwards directly to `pallet-ethereum`.

## Changes and Descriptions
* forward ethereum calls directly to `pallet-ethereum::Pallet::<T>::transact(..)`
# Checklist:


pallets/ethereum-transaction/src/lib.rs
```diff
@@ -32,106 +29,45 @@ mod tests;
...
+		OriginFor<T>:
+			From<pallet_ethereum::Origin> + Into<Result<pallet_ethereum::Origin, OriginFor<T>>>,
+	{
 		fn get_transaction_signature() -> Option<TransactionSignature> {
-			//TODO(cdamian): Same signature as the one in ethereum-xcm.
 			TransactionSignature::new(
 				TRANSACTION_RECOVERY_ID,
-				H256::from_low_u64_be(1u64),
-				H256::from_low_u64_be(1u64),
+				H256::from_low_u64_be(2u64),
```
wischli, line 60: Just want to make sure this bump is intended? If you have the time, would be interested what this number reflects as the type is quite opaque:

```rust
pub struct TransactionSignature {
	v: TransactionRecoveryId,
	r: H256,
	s: H256,
}
```

pallets/liquidity-pools/src/lib.rs
```diff
@@ -243,6 +243,19 @@ pub mod pallet {
...
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// An incoming LP message was
+		/// detected and is further processed
+		IncomingMessage {
+			sender: DomainAddress,
+			msg: MessageOf<T>,
+		},
```
wischli, line 258: On a note: I would like to introduce the opposite `OutgoingMessage` enum event in #1473

wischli (dismissed):
I am missing expertise to evaluate the changes for the router and gateway but given that this is the result of a long debugging session, I am optimistically approving. Rest LGTM!

pallets/liquidity-pools/src/lib.rs
mustermeiszer: Should I also add it here do you mean? In this PR?

pallets/ethereum-transaction/src/lib.rs
mustermeiszer: I just wanted the signature to not match the one from `pallet-xcm-transactor`. I have to be honest, no clue whether this makes snese. The signature is not checked, so I guess it is not really relevant.

pallets/liquidity-pools/src/lib.rs
wischli: No, not at all. I should have been more explicit. I wanted to express I am happy the opposite of my planned `Event` impl for Liquidity Pools will be implemented by this PR as this is a good signal for my plans.

pallets/ethereum-transaction/src/lib.rs
wischli: All right, thanks nevertheless!

cdamian (approved):

PR #1537, opened 2023-09-08
Fix lp-gateway routers' weight and fee calculations

NunoAlexandre:
### Changes

Drop the `XcmRouter`.`fee_per_second` field and add the 2 following new ones

```
pub transact_weight_at_most: Weight,
pub overall_weight: Weight,
pub fee_amount: u128,
```

That way we have complete control over those two values being passed down to `XcmTransactor`, instead of storing values which then have harcoded calculations in the runtime which we can't change without a runtime upgrade.


NunoAlexandre:
We have to keep the `max_gas_limit` since we needed that for the encoded evm call function. We could derive that from the transact max weight but then we are again stucking ourselves with hardcoded logic.

NunoAlexandre:
@cdamian @wischli @mustermeiszer this should be ready now. The only thing we need to discuss is migrations, which would only be needed for `Algol`. do we want to do that? 🙃

mustermeiszer:
Nope we just set a new router for algol

mustermeiszer (approved):
Can we merge this?

NunoAlexandre:
> Can we merge this?

yes please :) if there are no review remarks ofc.

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools.rs
```diff
@@ -814,10 +814,28 @@ fn add_currency() {
 			})
 		));
 
+		assert_eq!(
```
cdamian, line 817: Thank you for adding these checks! <3

pallets/liquidity-pools-gateway/routers/src/tests.rs
```diff
@@ -315,29 +304,6 @@ mod xcm_router {
 				assert_ok!(router.do_send(test_data.sender, test_data.msg));
 			});
 		}
-
-		#[test]
-		fn transactor_info_not_set() {
```
cdamian, line 320: Thanks for removing this, it was a left-over from the time we had init logic in the router IIRC.

cdamian (approved):
Thank you @NunoAlexandre !

pallets/liquidity-pools-gateway/routers/src/tests.rs
mustermeiszer: My artifact. just thought it is still valid to test whether it works if global info is not set. But is implicitly anyways tested ^^

mustermeiszer (approved):

PR #1546, opened 2023-09-12
OrderBook: Use ratio instead of rate

lemunozm:
# Description

Ideally, `order-book` should use an 18-decimal fixed point number too.

## Changes and Descriptions
- Change type in dev runtime.
- Remove dependency from `Rate` type in tests/benchs.

### DIFF
--- a/pallets/order-book/src/benchmarking.rs
+++ b/pallets/order-book/src/benchmarking.rs
@@ -12,10 +12,7 @@
 
 #![cfg(feature = "runtime-benchmarks")]
 
-use cfg_types::{
-	fixed_point::Rate,
-	tokens::{CurrencyId, CustomMetadata},
-};
+use cfg_types::tokens::{CurrencyId, CustomMetadata};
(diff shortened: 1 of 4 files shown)


mustermeiszer (commented):
Why ideally? Changes look fine though

lemunozm:
As I see it, the `SellRate` is used to denominate prices in units of other prices. So, they are specifying quantities or ratios, not 27-decimal rates (whose purpose is to be used for accruing interest per second). Actually, if some currency is 10^9 more valued than others, with 27-decimal, we can no longer use order-book for those.

wischli (approved):
Thanks for picking this up! LGTM

wischli:
> Actually, if some currency is 10^9 more valued than others, with 27-decimal, we can no longer use order-book for those.

Thanks so much for making this explicit! Given that `DAI` has 18 decimals and `USDT` having only 6, this fix is a must-have.

PR #1547, opened 2023-09-12
fix: algol relay v3 spec

wischli:
* Updates existing `rococo-local.json` relay spec used for Algol
* Sets distinct spec id `rococo_algol_v3` instead of default `rococo_local_testnet`
* Adds two more initial authorities (now 4 instead of 2)
* Based on `Polkadot v0.9.38`


wischli:
Closing because we drop Algol

PR #1550, opened 2023-09-14
lp-gateway: Emit more events when processing an incoming message

cdamian:
# Description

Added more events in LP gateway `process_msg` to allow for better debugging of incoming XCM transaction.

# Checklist:

### DIFF
--- a/pallets/liquidity-pools-gateway/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/src/lib.rs
@@ -57,6 +57,8 @@ pub mod pallet {
 	const BYTES_U32: usize = 4;
 	const BYTES_ACCOUNT_20: usize = 20;
 
+	use frame_support::traits::OriginTrait;
+
 	use super::*;
 	use crate::RelayerMessageDecodingError::{
 		MalformedMessage, MalformedSourceAddress, MalformedSourceAddressLength,
@@ -148,6 +150,12 @@ pub mod pallet {
 			message: Vec<u8>,
 			domain: Domain,
 		},
+
+		/// An incoming message has been received.
+		IncomingMessageReceived { message: Vec<u8> },
+
+		/// An invalid origin was used for an incoming message.
+		InvalidIncomingMessageOrigin,
 	}
 
 	/// Storage for domain routers.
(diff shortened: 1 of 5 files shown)


pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -328,7 +336,17 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			msg: BoundedVec<u8, T::MaxIncomingMessageSize>,
 		) -> DispatchResult {
-			let (domain_address, incoming_msg) = match T::LocalEVMOrigin::ensure_origin(origin)? {
+			Self::deposit_event(Event::IncomingMessageReceived {
+				message: msg.to_vec(),
+			});
+
+			let origin = T::LocalEVMOrigin::ensure_origin(origin).map_err(|e| {
+				Self::deposit_event(Event::InvalidIncomingMessageOrigin);
```
mustermeiszer, line 344: We will never see that. Upon error the state will be discarded. We would need to make this method `non_transactional` in order to see events of failures.

cdamian: Ah ok, but making it `non-transactional` would not be OK given that we're calling the LP pallet right?

I can also remove this, the one thing that is important here would be the test that confirms that an incoming xcm call is executed accordingly.

wischli: > Ah ok, but making it non-transactional would not be OK given that we're calling the LP pallet right?

If we removed the `transactional` component, we would have to ensure no storage is mutated until errors cannot occur anymore. However AFAIK, we can have transactional sub-scope within `process_msg` such as `submit`.

cdamian: The inbound queue implemented by the LP pallet is marked as transactional, does that mean that we can mark `process_msg` as non-transactional then?

https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/liquidity-pools/src/lib.rs#L937

wischli: > The inbound queue implemented by the LP pallet is marked as transactional, does that mean that we can mark `process_msg` as non-transactional then?

Yes, that would be possible. However, we should keep extrinsics `transactional` and rather check, whether we can remove that layer from the `InboundQueue::submit` impl of `pallet-liquidity-pools`. 

I had to ensure myself via some local tests: If a transactional call `c` executes nested non-transactional calls `nested_c` which mutate storage, then all storage mutations of `c` including `nested_c` are still rolled back if any error is thrown by `c`. 

Transactional layers [come with minor overhead](https://substrate.stackexchange.com/questions/1184/whats-the-overhead-associated-with-the-transactional-macro):
* 50 ns for the layer itself
* ~146 ns writing storage item from `transactional` layer to db which is only ~0.15% overhead compared to weight for one `RocksDb` write

mustermeiszer:
I close this for now. Am cleaning up a bit. Feel free to re-open if still needed.

PR #1556, opened 2023-09-18
feat: enable full foreign collect flow via extrinsic

wischli:
# Description

## Collect

* Adds `InvestmentPaymentCurrency` storage for the `foreign payment currency` upon receiving the first IncreaseInvestOrder message (similar to redemptions)
  * `InvestmentPaymentCurrency` will be cleared if the `InvestState` is cleared, i.e. if neither an unprocessed swap nor investment is happening in case of reaching `NoState` or `SwapIntoForeignDone`
  * `InvestmentPaymentCurrency` will be overwritten upon receiving `IncreaseInvestOrder` or `CollectInvestOrder` to be less restrictive to investors
  * We will read from that storage after the investment was collected through `pallet_investments::collect_invest_for` such that the dispatch of `ExecutedCollectInvest` happens automatically
* This enables us to not have to add a `collect_foreign_investment_for` extrinsic.
* For collecting redemptions, this flow is already implemented. However, there is an async gap in between collecting the redemption (into `pool `currency) and fully fulfilling the swap order (to `foreign` currency).
  * When that swap is fulfilled, the dispatch of `ExecutedCollectRedeem` happens automatically
* Of course, triggering collections through LP messages from EVM domains is also supported. The flows would be the same as for `pallet_investments::collect_{invest, redeem}_for`.

## Misc

* Renames `allow_pool_currency` extrinsic to `allow_investment_currency`
* Slightly reduces `foreign_investments` integration test boilerplate

# Checklist:

### DIFF
--- a/libs/types/src/investments.rs
+++ b/libs/types/src/investments.rs
@@ -185,7 +185,7 @@ pub struct Swap<
 	pub currency_in: Currency,
 	/// The outgoing currency, i.e. the one which should be replaced.
 	pub currency_out: Currency,
-	/// The amount of outgoing currency which shall be exchanged.
+	/// The amount of incoming currency which shall be bought.
 	pub amount: Balance,
 }
 
--- a/libs/types/src/investments.rs
+++ b/libs/types/src/investments.rs
@@ -237,30 +237,29 @@ pub struct ExecutedForeignDecreaseInvest<Balance, Currency> {
 	pub amount_remaining: Balance,
 }
 
-/// A representation of an executed collected investment.
+/// A representation of an executed collected foreign investment or redemption.
 #[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, Default, TypeInfo, MaxEncodedLen)]
 
-pub struct ExecutedForeignCollectInvest<Balance> {
-	/// The amount that was actually collected
-	pub amount_currency_payout: Balance,
-	/// The amount of tranche tokens received for the investment made
-	pub amount_tranche_tokens_payout: Balance,
-	/// The unprocessed plus processed but not yet collected investment amount
-	/// denominated in foreign currency
-	pub amount_remaining_invest: Balance,
-}
-
-/// A representation of an executed collected redemption.
-#[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, Default, TypeInfo, MaxEncodedLen)]
-
-pub struct ExecutedForeignCollectRedeem<Balance, Currency> {
-	/// The foreign currency in which the payout takes place
+pub struct ExecutedForeignCollect<Balance, Currency> {
+	/// The foreign currency in which ...
+	/// * If investment: the payment took place
+	/// * If redemption: the payout takes place
 	pub currency: Currency,
-	/// The amount of `currency` being paid out to the investor
+
+	/// The amount of `currency`...
+	/// * If investment: that was collected
+	/// * If redemption: paid out to the investor
 	pub amount_currency_payout: Balance,
-	/// How many tranche tokens were actually redeemed
+
+	/// The amount of tranche tokens...
+	/// * If investment: received for the investment made
+	/// * If redemption: which were actually redeemed
 	pub amount_tranche_tokens_payout: Balance,
-	/// The unprocessed plus processed but not yet collected redemption amount
-	/// of tranche tokens
-	pub amount_remaining_redeem: Balance,
+
+	/// The unprocessed ...
+	/// * If investment: investment amount of `currency` (denominated in foreign
+	///   currency)
+	/// * If redemption: redemption amount of tranche tokens (denominated in
+	///   pool currency)
+	pub amount_remaining: Balance,
 }
--- a/pallets/foreign-investments/src/impls/mod.rs
+++ b/pallets/foreign-investments/src/impls/mod.rs
@@ -55,9 +55,27 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 			Error::<T>::InvestError(InvestError::CollectRequired)
 		);
 
-		// NOTE: For the MVP, we simply overwrite any existing payment currency. In a
-		// later version, we might want to store a `StorageDoubleMap` instead.
-		InvestmentPaymentCurrency::<T>::insert(who, investment_id, foreign_currency);
+		// NOTE: For the MVP, we restrict the investment to the payment currency of the
+		// one from the initial increase. Once the `InvestmentState` has been cleared,
+		// another payment currency can be introduced.
+		let currency_matches = InvestmentPaymentCurrency::<T>::try_mutate_exists(
+			who,
+			investment_id,
+			|maybe_currency| {
+				if let Some(currency) = maybe_currency {
+					Ok::<bool, DispatchError>(currency == &foreign_currency)
+				} else {
+					*maybe_currency = Some(foreign_currency);
+					Ok::<bool, DispatchError>(true)
+				}
+			},
+		)
+		// An error reflects the payment currency has not been set yet
+		.unwrap_or(true);
+		ensure!(
+			currency_matches,
+			Error::<T>::InvestError(InvestError::InvalidPaymentCurrency)
+		);
 
 		let amount_pool_denominated =
 			T::CurrencyConverter::stable_to_stable(pool_currency, foreign_currency, amount)?;
@@ -89,6 +107,11 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 			!T::Investment::investment_requires_collect(who, investment_id),
 			Error::<T>::InvestError(InvestError::CollectRequired)
 		);
+		let payment_currency = InvestmentPaymentCurrency::<T>::get(who, investment_id)?;
+		ensure!(
+			payment_currency == foreign_currency,
+			Error::<T>::InvestError(InvestError::InvalidPaymentCurrency)
+		);
 
 		let pre_state = InvestmentState::<T>::get(who, investment_id);
 		ensure!(
@@ -119,18 +142,23 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 		amount: T::Balance,
 		payout_currency: T::CurrencyId,
 	) -> Result<(), DispatchError> {
-		let currency_matches =
-			RedemptionPayoutCurrency::<T>::mutate(who, investment_id, |maybe_currency| {
+		let currency_matches = RedemptionPayoutCurrency::<T>::try_mutate_exists(
+			who,
+			investment_id,
+			|maybe_currency| {
 				if let Some(currency) = maybe_currency {
-					currency == &payout_currency
+					Ok::<bool, DispatchError>(currency == &payout_currency)
 				} else {
 					*maybe_currency = Some(payout_currency);
-					true
+					Ok::<bool, DispatchError>(true)
 				}
-			});
+			},
+		)
+		// An error reflects the payout currency has not been set yet
+		.unwrap_or(true);
 		ensure!(
 			currency_matches,
-			Error::<T>::InvalidRedemptionPayoutCurrency
+			Error::<T>::RedeemError(RedeemError::InvalidPayoutCurrency)
 		);
 		ensure!(
 			!T::Investment::redemption_requires_collect(who, investment_id),
@@ -157,14 +185,10 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 		amount: T::Balance,
 		payout_currency: T::CurrencyId,
 	) -> Result<(T::Balance, T::Balance), DispatchError> {
+		let stored_payout_currency = RedemptionPayoutCurrency::<T>::get(who, investment_id)?;
 		ensure!(
-			RedemptionPayoutCurrency::<T>::get(who, investment_id)
-				.map(|currency| currency == payout_currency)
-				.unwrap_or_else(|| {
-					log::debug!("Redemption payout currency missing when calling decrease. Should never occur if redemption has been increased beforehand");
-					false
-				}),
-			Error::<T>::InvalidRedemptionPayoutCurrency
+			stored_payout_currency == payout_currency,
+			Error::<T>::RedeemError(RedeemError::InvalidPayoutCurrency)
 		);
 		ensure!(
 			!T::Investment::redemption_requires_collect(who, investment_id),
@@ -189,11 +213,13 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 	fn collect_foreign_investment(
 		who: &T::AccountId,
 		investment_id: T::InvestmentId,
-		foreign_payout_currency: T::CurrencyId,
+		foreign_payment_currency: T::CurrencyId,
 	) -> DispatchResult {
-		// Simply overwrite any existing payment currency to enable the investor to have
-		// full control over the payment currency
-		InvestmentPaymentCurrency::<T>::insert(who, investment_id, foreign_payout_currency);
+		let payment_currency = InvestmentPaymentCurrency::<T>::get(who, investment_id)?;
+		ensure!(
+			payment_currency == foreign_payment_currency,
+			Error::<T>::InvestError(InvestError::InvalidPaymentCurrency)
+		);
 
 		// Note: We assume the configured Investment trait to notify about the collected
 		// amounts via the `CollectedInvestmentHook` which handles incrementing the
@@ -212,14 +238,10 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 		foreign_payout_currency: T::CurrencyId,
 		pool_currency: T::CurrencyId,
 	) -> Result<(), DispatchError> {
+		let payout_currency = RedemptionPayoutCurrency::<T>::get(who, investment_id)?;
 		ensure!(
-			RedemptionPayoutCurrency::<T>::get(who, investment_id)
-				.map(|currency| currency == foreign_payout_currency)
-				.unwrap_or_else(|| {
-					log::debug!("Corruption: Redemption payout currency missing when calling decrease. Should never occur if redemption has been increased beforehand");
-					false
-				}),
-			Error::<T>::InvalidRedemptionPayoutCurrency
+			payout_currency == foreign_payout_currency,
+			Error::<T>::RedeemError(RedeemError::InvalidPayoutCurrency)
 		);
 		ensure!(T::PoolInspect::currency_for(investment_id.of_pool())
 			.map(|currency| currency == pool_currency)
@@ -1071,10 +1093,7 @@ impl<T: Config> Pallet<T> {
 		investment_id: T::InvestmentId,
 		collected: CollectedAmount<T::Balance>,
 	) -> DispatchResult {
-		let foreign_payout_currency = RedemptionPayoutCurrency::<T>::get(who, investment_id)
-			.ok_or(Error::<T>::RedeemError(
-				RedeemError::CollectPayoutCurrencyNotFound,
-			))?;
+		let foreign_payout_currency = RedemptionPayoutCurrency::<T>::get(who, investment_id)?;
 		let pool_currency = T::PoolInspect::currency_for(investment_id.of_pool())
 			.expect("Impossible to collect redemption for non existing pool at this point");
 
@@ -1162,8 +1181,7 @@ impl<T: Config> Pallet<T> {
 		who: &T::AccountId,
 		investment_id: T::InvestmentId,
 	) -> DispatchResult {
-		let foreign_payout_currency = InvestmentPaymentCurrency::<T>::get(who, investment_id)
-			.ok_or(Error::<T>::InvestmentPaymentCurrencyNotFound)?;
+		let foreign_payout_currency = InvestmentPaymentCurrency::<T>::get(who, investment_id)?;
 		let pool_currency = T::PoolInspect::currency_for(investment_id.of_pool())
 			.ok_or(Error::<T>::PoolNotFound)?;
 		let collected = CollectedInvestment::<T>::take(who, investment_id);
--- a/pallets/foreign-investments/src/impls/mod.rs
+++ b/pallets/foreign-investments/src/impls/mod.rs
@@ -1054,15 +1054,18 @@ impl<T: Config> Pallet<T> {
 		collected: CollectedAmount<T::Balance>,
 	) -> DispatchResult {
 		// Increment by previously stored amounts (via `CollectedInvestmentHook`)
-		CollectedInvestment::<T>::mutate(who, investment_id, |collected_before| {
-			collected_before
-				.amount_collected
+		let nothing_collect = CollectedInvestment::<T>::mutate(who, investment_id, |c| {
+			c.amount_collected
 				.ensure_add_assign(collected.amount_collected)?;
-			collected_before
-				.amount_payment
+			c.amount_payment
 				.ensure_add_assign(collected.amount_payment)?;
-			Ok::<(), DispatchError>(())
+			Ok::<bool, DispatchError>(c.amount_collected.is_zero() && c.amount_payment.is_zero())
 		})?;
+
+		// No need to transition if nothing was collected
+		if nothing_collect {
+			return Ok(());
+		}
 
 		// Update invest state to decrease the unprocessed investing amount
 		let investing_amount = T::Investment::investment(who, investment_id)?;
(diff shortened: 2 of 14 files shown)


pallets/foreign-investments/src/impls/mod.rs
```diff
@@ -56,6 +54,11 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 			!T::Investment::investment_requires_collect(who, investment_id),
 			Error::<T>::InvestError(InvestError::CollectRequired)
 		);
+
+		// NOTE: For the MVP, we simply overwrite any existing payment currency. In a
+		// later version, we might want to store a `BoundedVec` instead.
```
NunoAlexandre, line 59: I almost always find it better to use a double map instead with `()` as the value type since that makes storage reads/writes a lot cheaper and keeps the code simple as it doesn't need the Max for the bounded vec. If this is something we may need in the future, it's low effort to implement it already and save the migration work later on.

pallets/foreign-investments/src/impls/mod.rs
```diff
@@ -325,6 +328,10 @@ impl<T: Config> Pallet<T> {
 		match state {
 			InvestState::NoState => {
 				InvestmentState::<T>::remove(who, investment_id);
+				#[cfg(feature = "std")] {
+					println!("Clearing InvestState");
+				}
```
NunoAlexandre, line 333: Of curiosity, is it possible to use the `log` instead so that we can have these logs both on std and wasm?

pallets/foreign-investments/src/lib.rs
```diff
@@ -412,17 +437,12 @@ pub mod pallet {
...
-	// 		fn from(error: InvestError) -> Self {
-	// 			Error::<T>::InvestError(error)
-	// 		}
-	// 	}
-
-	// 	impl<T> From<RedeemError> for Error<T> {
-	// 		fn from(error: RedeemError) -> Self {
-	// 			Error::<T>::RedeemError(error)
-	// 		}
-	// 	}
```
NunoAlexandre, line 427: 🧹 ✨

pallets/liquidity-pools/src/hooks.rs
```diff
@@ -101,7 +104,57 @@ where
...
+		};
+
+		T::OutboundQueue::submit(T::TreasuryAccount::get(), domain_address.domain(), message)?;
+
+		Ok(())
+	}
+}
+
+/// The hook struct which acts upon a finalized investment collection.
+
```
NunoAlexandre, line 117:
```suggestion
```

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
```diff
@@ -634,7 +647,8 @@ mod same_currencies {
 				1
 			);
 
-			// User collects through foreign investments
+			// Should not mutate any state if user collects through foreign investments
+			// after collecting through Investments
```
NunoAlexandre, line 651: these comments are really helpful 🙏

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
```diff
@@ -1450,10 +1477,96 @@ mod mismatching_currencies {
...
+			let pool_id = DEFAULT_POOL_ID;
+			let investor: AccountId =
+				AccountConverter::<DevelopmentRuntime, LocationToAccountId>::convert((
+					DOMAIN_MOONBEAM,
+					BOB,
+				));
+			let pool_currency: CurrencyId = AUSD_CURRENCY_ID;
+			let foreign_currency: CurrencyId = USDT_CURRENCY_ID;
+			let pool_currency_decimals = currency_decimals::AUSD;
+			let invest_amount_pool_denominated: u128 = 6_000_000_000_000_000;
```
NunoAlexandre, line 1503: can we use the `dollar(<decimals>)` function here? So this would become `6 * dollar(18)` (where 18 is the AUSD decimals. Even better if we have a constant for that). Just to facilitate reading these amounts and to avoid a situation where for example we may be moving amounts that are too high or too low and not quite testing things properly.

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
```diff
@@ -1450,10 +1477,96 @@ mod mismatching_currencies {
...
+			let sending_domain_locator = Domain::convert(DEFAULT_DOMAIN_ADDRESS_MOONBEAM.domain());
+			create_currency_pool(pool_id, pool_currency, pool_currency_decimals.into());
+			let invest_amount_foreign_denominated: u128 =
+				enable_usdt_trading(pool_currency, invest_amount_pool_denominated, true, || {});
+			do_initial_increase_investment(
```
NunoAlexandre, line 1508: Nit but personally as a reader I like to have a line break between all the `lets` and an actual call 🙏 

```suggestion
			
			do_initial_increase_investment(
```

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
```diff
@@ -1450,10 +1477,96 @@ mod mismatching_currencies {
...
+				),
+				Some(foreign_currency)
+			);
+			assert!(Tokens::balance(default_investment_id().into(), &investor).is_zero());
+			assert_eq!(
+				Tokens::balance(default_investment_id().into(), &sending_domain_locator),
+				invest_amount_pool_denominated * 2
+			);
+
+			// Should not be cleared as invest state is swapping into pool currency
```
NunoAlexandre, line 1559: Again, super nice to have these checks and these comments 👌

NunoAlexandre (dismissed):
Very nice work, like always 💯 

A couple of small comments but no blockers for me.

pallets/foreign-investments/src/impls/mod.rs
wischli: Sorry, this is an unwanted remnant from debugging integration tests. I will remove it!

pallets/foreign-investments/src/impls/mod.rs
wischli: I think both options have separate use cases. When using a `DoubleStorageMap`, you don't want to limit the key-value pairs which you want when using a `StorageMap` with `BoundedVec` values. 

Nevertheless, I agree a double map might be more suitable here. Given that we currently limit `(investor, investment_id)` pairs to have at most one swap at the same time, adding true multicurrency support (both in EVM as well as Centrifuge Chain) will be a bigger feature. It was discussed this is something we need mid- to long-term (rather 2024) but not for this MVP.

NunoAlexandre: > [It was discussed ](https://centrifuge.hackmd.io/tKGS5CwqSQeeI3bU1dKUlw#20230815)this is something we need mid- to long-term (rather 2024) but not for this MVP.


Then I say we do this now, it takes very little effort compared to the significant effort and error-proneness that writing a migration for this in the future will entail.

NunoAlexandre: I see it's done already, thanks for doing that :)

NunoAlexandre (approved):
👌  nice

pallets/foreign-investments/src/impls/mod.rs
wischli: I just updated the comment but did not alter any storage. Migrating to double maps will touch multiple storages and complicates swap order handling much more. It definitely should not be part of this PR.

NunoAlexandre: Oh, I looked at the `InvestmentPaymentCurrency` which is a double map, my bad. I don't agree with the decision here but it's not a blocker for me.

libs/types/src/investments.rs
```diff
@@ -185,7 +185,7 @@ pub struct Swap<
 	pub currency_in: Currency,
 	/// The outgoing currency, i.e. the one which should be replaced.
 	pub currency_out: Currency,
-	/// The amount of outgoing currency which shall be exchanged.
+	/// The amount of incoming currency which shall be bought.
```
wischli, line 188: See [this comment](https://github.com/centrifuge/centrifuge-chain/pull/1509#discussion_r1331346309) for explanation.

mustermeiszer:
Checking something from the description:

> InvestmentPaymentCurrency will be overwritten upon receiving IncreaseInvestOrder or CollectInvestOrder to be less restrictive to investors

Wouldn't that also mean, that a user could overwrite this storage with another currency mid-flight and for example, change the actual currency that is signaled via `ExecutedCollectInvest`?

pallets/foreign-investments/src/impls/mod.rs
```diff
@@ -187,19 +190,19 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 		who: &T::AccountId,
 		investment_id: T::InvestmentId,
 		foreign_payout_currency: T::CurrencyId,
-		pool_currency: T::CurrencyId,
-	) -> Result<ExecutedForeignCollectInvest<T::Balance>, DispatchError> {
+	) -> DispatchResult {
+		// Simply overwrite any existing payment currency to enable the investor to have
+		// full control over the payment currency
+		InvestmentPaymentCurrency::<T>::insert(who, investment_id, foreign_payout_currency);
```
mustermeiszer, line 196: Isn't that criticial if something is happening currently? E.g.  swapping due to an `DecreaseInvestOrder`?

wischli:
> Checking something from the description:
> 
> > InvestmentPaymentCurrency will be overwritten upon receiving IncreaseInvestOrder or CollectInvestOrder to be less restrictive to investors
> 
> Wouldn't that also mean, that a user could overwrite this storage with another currency mid-flight and for example, change the actual currency that is signaled via `ExecutedCollectInvest`?

You are right, shoot. Then, I don't see another way of limiting an `(investor, investment_id)` pair to the same foreign currency as long as the foreign investment state is not cleared which happens if neither an active (invest) swap nor an unprocessed amount exist. WDYT?

pallets/foreign-investments/src/impls/mod.rs
mustermeiszer: Or a `IncreaseInvestOrder` with `payment_currency_a` followed by an `CollectInvest` with `payment_currency_b`

pallets/foreign-investments/src/impls/mod.rs
```diff
@@ -1171,9 +1151,56 @@ impl<T: Config> Pallet<T> {
...
+	/// currency via the `CurrencyConverter` trait.
+	pub(crate) fn notify_executed_collect_invest(
+		who: &T::AccountId,
+		investment_id: T::InvestmentId,
+	) -> DispatchResult {
+		let foreign_payout_currency = InvestmentPaymentCurrency::<T>::get(who, investment_id)
+			.ok_or(Error::<T>::InvestmentPaymentCurrencyNotFound)?;
+		let pool_currency = T::PoolInspect::currency_for(investment_id.of_pool())
+			.ok_or(Error::<T>::PoolNotFound)?;
+		let collected = CollectedInvestment::<T>::take(who, investment_id);
```
mustermeiszer, line 1169: Due to being a `ValueQuery` this will always hit. Are we sure the solidity side wants to be informed if nothing was acutally transferred? 

Due to the need of bridging this seems like a big overhead for no information gain.

pallets/foreign-investments/src/impls/mod.rs
```diff
@@ -1171,9 +1151,56 @@ impl<T: Config> Pallet<T> {
...
+	/// `ExecutedCollectInvest`.
+	///
+	/// NOTE: Converts the collected pool currency payment amount to foreign
+	/// currency via the `CurrencyConverter` trait.
+	pub(crate) fn notify_executed_collect_invest(
+		who: &T::AccountId,
+		investment_id: T::InvestmentId,
+	) -> DispatchResult {
+		let foreign_payout_currency = InvestmentPaymentCurrency::<T>::get(who, investment_id)
+			.ok_or(Error::<T>::InvestmentPaymentCurrencyNotFound)?;
```
mustermeiszer, line 1166: FYI: All these errors can be directly merged into the storage type now.

wischli:
> You are right, shoot. Then, I don't see another way of limiting an `(investor, investment_id)` pair to the same foreign currency as long as the foreign investment state is not cleared which happens if neither an active (invest) swap nor an unprocessed amount exist. WDYT?

EDIT: Thinking about this more I actually think that's not really possible since collecting a redemption is synchronous. So it could only occur if within the same block an investor changes the `foreign_currency` by either increasing their investment or calling `collect` and this transaction is included before the `investment::collect_redeem_for` one. I don't see any issues with that as this signal's the investor's intend.

pallets/liquidity-pools/src/inbound.rs
```diff
@@ -305,44 +303,13 @@ where
 		tranche_id: T::TrancheId,
 		investor: T::AccountId,
 		currency_index: GeneralCurrencyIndexOf<T>,
-		destination: DomainAddress,
 	) -> DispatchResult {
 		let invest_id: T::TrancheCurrency = Self::derive_invest_id(pool_id, tranche_id)?;
-		let currency_index_u128 = currency_index.index;
 		let payout_currency = Self::try_get_payout_currency(invest_id.clone(), currency_index)?;
```
mustermeiszer, line 308: **CR**

Looking at the code of `try_get_payout_currency` I think this is not working. 

The check that is being used `T::ForeignInvestment::accepted_payment_currency(invest_id,currency)` will check the wrong trading pair. I.e. the wrong trading pair direction. As trading pairs are not ensured to work in both ways, this might return ok also it will fail when trying to swap.

mustermeiszer (changes requested):
1. Unsure whether overwriting the invest currency is not potentially dangerous.
2. Quite sure that the check for `payout_currency` is currently a bug

mustermeiszer:
> Thinking about this more I actually think that's not really an issue since collecting an investment is synchronous. So it could only occur if within the same block

But I think given the current code complexity, this is quite an assumption for the future. I mean it being syncrhonous all the time. 

And IIRC we wanted that for the Solidity side and if IIRC this does not handle it well

Example:
* a user invests from multiple `LP`s  (`Lp(currency_a)`, `LP(currency_b)`)
* user triggers `IncreaseInvestOrder(currency_a)` &rarr; `msg(3)`
* user triggers `CollectInvest(currency_a)` &rarr; `msg(2)`
* user triggers `IncreaseInvestOrder(currency_b)` &rarr; `msg(3)`

Message execution on our side due to bridge or mem-pool is: 
* `msg(1)`
* `msg(3)`
* `msg(2)`

this would mean the wrong `LP` on the Solidity side is informed about the collect and the right one never sees it. But maybe I am also not getting the Solidity side correctly. cc @hieronx and @ilinzweilin

pallets/foreign-investments/src/impls/mod.rs
wischli: I am confused, in #1473 [you requested](https://github.com/centrifuge/centrifuge-chain/pull/1473#discussion_r1323224830) to allow collecting zero amounts, so I removed the barrier. IIUC, now you contradict your previous request?

pallets/liquidity-pools/src/inbound.rs
wischli: Great catch! Will fix `fn try_get_payout_currency`.

wischli:
> But I think given the current code complexity, this is quite an assumption for the future. I mean it being syncrhonous all the time.

When would it not be synchronous? Collecting an investment does not require a swap. It just performs `T::Investment::collect_invest_for` and puts `CollectInvest` into the `OutboundQueue` based on the information stored during the `collect_invest_for` and the `InvestmentPaymentCurrency`.

Regarding the example, your message id declaration is off. I suppose you wanted the first message to have index 1.

> And IIRC we wanted that for the Solidity side and if IIRC this does not handle it well
> 
> Example:
> 
> * a user invests from multiple `LP`s  (`Lp(currency_a)`, `LP(currency_b)`)
> * user triggers `IncreaseInvestOrder(currency_a)` → ~~`msg(3)`~~ `msg(1)`
> * user triggers `CollectInvest(currency_a)` → `msg(2)`
> * user triggers `IncreaseInvestOrder(currency_b)` → `msg(3)` 

If the user invests from multiple Liquidity Pools, each investment has their separate `investment_id` such that these are considered separate investments `(investor, LP(currency_a))` and `(investor, LP(currency_b))`.

If we had that message flow for the same `(investor, investment_id)` (i.e. same tranche id of the same LP), then you are right. I blanked out on the fact that we only need to do this so we can map the message to the correct contract address. I only thought about the response we give to the investor. And if that used `currency_b` instead of `currency_a`, it would just translate the payment amount into `currency_b` instead of `currency_a`.

TLDR: We need to restrict an investor to the same payment currency until the foreign investment state is cleared for now on Centrifuge chain IF we want to enable the full `collect_foreign_investment` flow by just calling `collect_invest_for` in `pallet-investment`. If that restriction is plausible, then I will make the appropriate adjustments to this PR. Please let me know what you think @mustermeiszer @hieronx

pallets/foreign-investments/src/impls/mod.rs
mustermeiszer: No, I am not. I am just saying that if we trigger a return message, we should be sure Solidity actually needs it. If not we should collect and not trigger a return message. But lets just keep it as is. Its just optimisation.

pallets/liquidity-pools/src/inbound.rs
mustermeiszer: I can not, but it might still happen by accident as no logic in the order book ensures bidirectionality/

mustermeiszer:
@hieronx need to answer if this is sufficient. I would for now maybe check on-chain that if an `InvestmentPaymentCurrency` is set, all new updates must be of the same currency, wdyt? Then the chain is save from wrongly generated message on the Solidity or frontend side.

pallets/foreign-investments/src/impls/mod.rs
wischli: I have been favouring blocking a return message if the amount is zero. I will implement this.

pallets/liquidity-pools/src/inbound.rs
wischli: On that note: If bidirectionality is not given, an investor cannot decrease their foreign investment if it creates an orderbook swap from `pool` to `foreign` because then `TradingPair::<T>::get(pool_currency, foreign_currency)`  does not exist. Moreover, we block redemptions for that `foreign` currency as well.

AFAICT, we should check in `fn allow_investment_currency` both if the currency is accepted as payment as well as payout, not only the former. WDYT?

pallets/foreign-investments/src/impls/mod.rs
hieronx: I think there is a possible scenario in which someone submitted a redemption order, and then the price ended up going to 0, and the redemption is fully executed but with 0 currency payout. In that case, the value of sending `ExecutedCollectRedeem` to the other domains is that they will update the local `remaining_redeem_order` state to 0.

hieronx:
> TLDR: We need to restrict an investor to the same payment currency until the foreign investment state is cleared for now on Centrifuge chain IF we want to enable the full collect_foreign_investment flow by just calling collect_invest_for in pallet-investment.
> I would for now maybe check on-chain that if an InvestmentPaymentCurrency is set, all new updates must be of the same currency, wdyt? 

Seems totally fine to me 👍

pallets/liquidity-pools/src/inbound.rs
wischli: > AFAICT, we should check in fn allow_investment_currency both if the currency is accepted as payment as well as payout, not only the former. WDYT?

cc @hieronx

hieronx: Yes 100%. I think in practice governance should ensure that always both directions of an LP token trading pair are added.

pallets/foreign-investments/src/impls/mod.rs
wischli: > I think there is a possible scenario in which someone submitted a redemption order, and then the price ended up going to 0, and the redemption is fully executed but with 0 currency payout. In that case, the value of sending `ExecutedCollectRedeem` to the other domains is that they will update the local `remaining_redeem_order` state to 0.

IIUC, the payment amount should then not be zero. I added a short circuit if the both the accumulated payment amount as well as collected amount are zero: [0b6924d](https://github.com/centrifuge/centrifuge-chain/pull/1556/commits/0b6924df3d14d6c18a15f760746259baca6b47c2)

pallets/foreign-investments/src/impls/mod.rs
wischli: [2768b4b](https://github.com/centrifuge/centrifuge-chain/pull/1556/commits/2768b4bdd81be4f3d025881252b75f6f0b538d08)

pallets/foreign-investments/src/impls/mod.rs
```diff
@@ -186,20 +213,22 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 	fn collect_foreign_investment(
 		who: &T::AccountId,
 		investment_id: T::InvestmentId,
-		foreign_payout_currency: T::CurrencyId,
-		pool_currency: T::CurrencyId,
-	) -> Result<ExecutedForeignCollectInvest<T::Balance>, DispatchError> {
+		foreign_payment_currency: T::CurrencyId,
+	) -> DispatchResult {
+		let payment_currency = InvestmentPaymentCurrency::<T>::get(who, investment_id)?;
```
mustermeiszer, line 218: Just wandering regarding the states of this. Wouldn't it be possible for the initial `IncreaseInvestOrder` to be fully "done" before a `CollectInvest` is actually triggered? 

Meaning, the `InvestmentPaymentCurrency` state would be `None`.

pallets/foreign-investments/src/impls/mod.rs
```diff
@@ -209,14 +238,10 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 		foreign_payout_currency: T::CurrencyId,
 		pool_currency: T::CurrencyId,
 	) -> Result<(), DispatchError> {
+		let payout_currency = RedemptionPayoutCurrency::<T>::get(who, investment_id)?;
```
mustermeiszer, line 241: Not sure if the same thinking applies to redemptions.

mustermeiszer (commented):
All looks good. I got a minor logical worry regarding the need to have a valid `InvestmentPaymentCurrency` when collecting. If this is unjustified it is good to go.

pallets/foreign-investments/src/impls/mod.rs
wischli: A `RedemptionState` is only cleared if 
1. either the entire unprocessed amount is decreased without an active/done swap (`Redeeming` --> `NoState`) or
2. the entire swap is fulfilled without unprocessed redeeming amount (`SwapIntoForeignDone` --> `NoState`)

The first one is irrelevant here because it requires the `RedemptionState` to be re-initialized by increasing the redemption which sets the payout currency.

The second one triggers the redemption collection transition before/when clearing:

https://github.com/centrifuge/centrifuge-chain/blob/0b6924df3d14d6c18a15f760746259baca6b47c2/pallets/foreign-investments/src/impls/mod.rs#L503-L509

https://github.com/centrifuge/centrifuge-chain/blob/0b6924df3d14d6c18a15f760746259baca6b47c2/pallets/foreign-investments/src/impls/mod.rs#L706-L714

pallets/foreign-investments/src/impls/mod.rs
wischli: An `InvestmentState` is only cleared if
1. either the entire unprocessed amount was decreased and that swap is fulfilled (`SwapIntoForeignDone` --> `NoState`)
2. The state only holds an unprocessed investment amount which was fully processed (`InvestmentOngoing` --> `NoState`)

The first one is irrelevant here because it requires the `InvestmentState` to be re-initialized by increasing the investment which sets the payment currency.

The second one is only known if the investment is collected. Therefore, we trigger the foreign collect flow before clearing the `InvestmentState` in `apply_invest_state_transition`:

https://github.com/centrifuge/centrifuge-chain/blob/0b6924df3d14d6c18a15f760746259baca6b47c2/pallets/foreign-investments/src/impls/mod.rs#L1076-L1083

mustermeiszer: Thanks for the details!

mustermeiszer (approved):
Thanks for the patience! looks good to go.

PR #1558, opened 2023-09-19
feat: orderbook bench helper

wischli:
Adds an benchmark helper to the `pallet-order-book` which exposes the necessary setup to register a trading pair for arbitrary currencies. 

It is inspired by the the existing `PoolBenchmarkHelper` which was moved to a new `benchmarking` module of `cfg_traits`.

Required for benchmarking LPs, especially foreign investments

### DIFF
--- a/libs/mocks/src/pools.rs
+++ b/libs/mocks/src/pools.rs
@@ -23,6 +23,7 @@ pub mod pallet {
 		type Balance;
 		type BalanceRatio;
 		type CurrencyId;
+		type TrancheCurrency;
 	}
 
 	#[pallet::pallet]
@@ -66,12 +67,18 @@ pub mod pallet {
 			register_call!(move |(a, b, c)| f(a, b, c));
 		}
 
-		pub fn mock_benchmark_create_pool(f: impl Fn(T::PoolId, &T::AccountId) + 'static) {
+		pub fn mock_bench_create_pool(f: impl Fn(T::PoolId, &T::AccountId) + 'static) {
 			register_call!(move |(a, b)| f(a, b));
 		}
 
-		pub fn mock_benchmark_give_ausd(f: impl Fn(&T::AccountId, T::Balance) + 'static) {
-			register_call!(move |(a, b)| f(a, b));
+		pub fn mock_bench_investor_setup(
+			f: impl Fn(T::PoolId, T::AccountId, T::Balance) + 'static,
+		) {
+			register_call!(move |(a, b, c)| f(a, b, c));
+		}
+
+		pub fn mock_bench_default_investment_id(f: impl Fn(T::TrancheCurrency) + 'static) {
+			register_call!(f);
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


pallets/pool-system/src/impls.rs
```diff
@@ -456,7 +456,7 @@ mod benchmarks_utils {
 
 	use super::*;
 
-	const AUSD_CURRENCY_ID: CurrencyId = CurrencyId::ForeignAsset(1);
+	const POOL_CURRENCY: CurrencyId = CurrencyId::ForeignAsset(1);
```
lemunozm, line 459: 👌🏻

pallets/pool-system/src/impls.rs
```diff
@@ -553,8 +553,8 @@ mod benchmarks_utils {
 			Pallet::<T>::close_epoch(RawOrigin::Signed(admin.clone()).into(), pool_id).unwrap();
 		}
 
-		fn benchmark_give_ausd(account: &T::AccountId, balance: T::Balance) {
-			T::Tokens::mint_into(AUSD_CURRENCY_ID, account, balance).unwrap();
+		fn bench_mint_pool_currency_into(account: &T::AccountId, balance: T::Balance) {
```
lemunozm, line 556: Thinking on the name of this function, WDYM: `bench_pool_user()`?.

I say that because what the function does should not be relevant to the user of it. As a last resort, the API user would want to configure a *pool user*. No matter what it implies. That way if tomorrow, it implies more things than "minting" we do not need to change the name.

lemunozm (commented):
Like a lot! Good to see this schema is growing to other pallets ❤️

pallets/pool-system/src/impls.rs
wischli: I partially agree. The question is whether the definition of _a pool user_ is clear. To me, it is an account with sufficient funds to invest and the `Investor` permission. What does it mean to you? Maybe we want to have finer granularity and have one call to fund an account and separate calls for preparing certain roles.

wischli: For now, I went with `bench_investor_setup`. Depending on your response, I might introduce finer granularity.

lemunozm: Not sure what direction it will grow, TBH, but I like a lot the new name `bench_investor_setup()`, more than `bench_pool_user` 👍🏻

mustermeiszer (approved):

libs/mocks/src/pools.rs
```diff
@@ -66,12 +67,18 @@ pub mod pallet {
...
 
-		pub fn mock_benchmark_give_ausd(f: impl Fn(&T::AccountId, T::Balance) + 'static) {
-			register_call!(move |(a, b)| f(a, b));
+		pub fn mock_bench_investor_setup(
+			f: impl Fn(T::PoolId, T::AccountId, T::Balance) + 'static,
+		) {
+			register_call!(move |(a, b, c)| f(a, b, c));
+		}
+
+		pub fn mock_bench_default_investment_id(f: impl Fn(T::TrancheCurrency) + 'static) {
```
lemunozm, line 80: I think the closure of this method should be `impl Fn(T::PoolId) -> T::InvestmentId + 'static` instead to match the trait method. 

This matching can only be ensured in runtime, there is no way to get the compiler warming about this. We'd need the procedural macros to avoid these problems.

lemunozm:
I would need this PR for: https://github.com/centrifuge/centrifuge-chain/pull/1563, is this ready to be merged?

wischli:
> I would need this PR for: #1563, is this ready to be merged?

Let me resolve the issues after rebasing and then let's merge this!

libs/mocks/src/pools.rs
wischli: Great catch! You are right. Looking forward to having procedural macro support 🤘

lemunozm (approved):

PR #1559, opened 2023-09-19
Feat/order book add fulfill order partially

cdamian:
# Description

Added a new extrinsic in the `order-book` pallet that allows for a partial order fulfillment.

# Checklist:

### DIFF
--- a/pallets/order-book/src/lib.rs
+++ b/pallets/order-book/src/lib.rs
@@ -127,9 +127,7 @@ pub mod pallet {
 			+ EnsureMul
 			+ EnsureDiv
 			+ TypeInfo
-			+ MaxEncodedLen
-			+ From<u64>
-			+ sp_arithmetic::traits::Unsigned;
+			+ MaxEncodedLen;
 
 		/// Type for currency orders can be made for
 		type TradeableAsset: AssetInspect<Self::AccountId, Balance = Self::Balance, AssetId = Self::AssetCurrencyId>
@@ -355,9 +353,8 @@ pub mod pallet {
 		/// Error when unable to convert fee balance to asset balance when asset
 		/// out matches fee currency
 		BalanceConversionErr,
-		/// Error when unable to calculate the remaining buy amount for a
-		/// partial order.
-		RemainingBuyAmountError,
+		/// Error when the provided partial buy amount is too large.
+		BuyAmountTooLarge,
 	}
 
 	#[pallet::call]
@@ -464,50 +461,9 @@ pub mod pallet {
 		pub fn fill_order_full(origin: OriginFor<T>, order_id: T::OrderIdNonce) -> DispatchResult {
 			let account_id = ensure_signed(origin)?;
 			let order = <Orders<T>>::get(order_id)?;
+			let buy_amount = order.buy_amount;
 
-			ensure!(
-				T::TradeableAsset::can_hold(order.asset_in_id, &account_id, order.buy_amount),
-				Error::<T>::InsufficientAssetFunds,
-			);
-
-			Self::unreserve_order(&order)?;
-			T::TradeableAsset::transfer(
-				order.asset_in_id,
-				&account_id,
-				&order.placing_account,
-				order.buy_amount,
-				false,
-			)?;
-			T::TradeableAsset::transfer(
-				order.asset_out_id,
-				&order.placing_account,
-				&account_id,
-				order.max_sell_amount,
-				false,
-			)?;
-			Self::remove_order(order.order_id)?;
-
-			T::FulfilledOrderHook::notify_status_change(
-				order_id,
-				Swap {
-					amount: order.buy_amount,
-					currency_in: order.asset_in_id,
-					currency_out: order.asset_out_id,
-				},
-			)?;
-
-			Self::deposit_event(Event::OrderFulfillment {
-				order_id,
-				placing_account: order.placing_account,
-				fulfilling_account: account_id,
-				partial_fulfillment: false,
-				currency_in: order.asset_in_id,
-				currency_out: order.asset_out_id,
-				fulfillment_amount: order.buy_amount,
-				sell_rate_limit: order.max_sell_rate,
-			});
-
-			Ok(())
+			Self::fulfill_order_with_amount(order, buy_amount, account_id)
 		}
 
 		/// Adds a valid trading pair.
@@ -588,7 +544,7 @@ pub mod pallet {
 			Ok(())
 		}
 
-		/// Fill an existing order, based on the provided ratio.
+		/// Fill an existing order, based on the provided partial buy amount.
 		#[pallet::call_index(7)]
 		#[pallet::weight(T::Weights::fill_order_partial())]
 		pub fn fill_order_partial(
@@ -599,6 +555,16 @@ pub mod pallet {
 			let account_id = ensure_signed(origin)?;
 			let order = <Orders<T>>::get(order_id)?;
 
+			Self::fulfill_order_with_amount(order, buy_amount, account_id)
+		}
+	}
+
+	impl<T: Config> Pallet<T> {
+		pub fn fulfill_order_with_amount(
+			order: OrderOf<T>,
+			buy_amount: T::Balance,
+			account_id: T::AccountId,
+		) -> DispatchResult {
 			ensure!(
 				buy_amount >= order.min_fulfillment_amount,
 				Error::<T>::InsufficientOrderSize,
@@ -618,13 +584,13 @@ pub mod pallet {
 			let remaining_buy_amount = order
 				.buy_amount
 				.checked_sub(&buy_amount)
-				.ok_or(Error::<T>::RemainingBuyAmountError)?;
+				.ok_or(Error::<T>::BuyAmountTooLarge)?;
 			let partial_fulfillment = !remaining_buy_amount.is_zero();
 
 			if partial_fulfillment {
 				Self::update_order(
 					order.placing_account.clone(),
-					order_id,
+					order.order_id,
 					remaining_buy_amount,
 					order.max_sell_rate,
 					remaining_buy_amount.min(order.min_fulfillment_amount),
@@ -647,7 +613,6 @@ pub mod pallet {
 				buy_amount,
 				false,
 			)?;
-
 			T::TradeableAsset::transfer(
 				order.asset_out_id,
 				&order.placing_account,
@@ -657,7 +622,7 @@ pub mod pallet {
 			)?;
 
 			T::FulfilledOrderHook::notify_status_change(
-				order_id,
+				order.order_id,
 				Swap {
 					amount: buy_amount,
 					currency_in: order.asset_in_id,
@@ -666,7 +631,7 @@ pub mod pallet {
 			)?;
 
 			Self::deposit_event(Event::OrderFulfillment {
-				order_id,
+				order_id: order.order_id,
 				placing_account: order.placing_account,
 				fulfilling_account: account_id,
 				partial_fulfillment,
@@ -678,9 +643,7 @@ pub mod pallet {
 
 			Ok(())
 		}
-	}
 
-	impl<T: Config> Pallet<T> {
 		/// Remove an order from storage
 		pub fn remove_order(order_id: T::OrderIdNonce) -> DispatchResult {
 			let order = <Orders<T>>::get(order_id)?;
--- a/pallets/order-book/src/lib.rs
+++ b/pallets/order-book/src/lib.rs
@@ -54,10 +54,7 @@ pub mod pallet {
 	use frame_system::pallet_prelude::{OriginFor, *};
 	use orml_traits::asset_registry::{self, Inspect as _};
 	use scale_info::TypeInfo;
-	use sp_arithmetic::{
-		traits::{BaseArithmetic, CheckedSub},
-		Perquintill,
-	};
+	use sp_arithmetic::traits::{BaseArithmetic, CheckedSub};
 	use sp_runtime::{
 		traits::{
 			AtLeast32BitUnsigned, EnsureAdd, EnsureDiv, EnsureFixedPointNumber, EnsureMul,
@@ -361,8 +358,6 @@ pub mod pallet {
 		/// Error when unable to calculate the remaining buy amount for a
 		/// partial order.
 		RemainingBuyAmountError,
-		/// Error when the ratio provided for a partial error is 0.
-		InvalidPartialOrderRatio,
 	}
 
 	#[pallet::call]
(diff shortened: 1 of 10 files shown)


runtime/altair/src/weights/pallet_order_book.rs
```diff
@@ -140,4 +140,14 @@ impl<T: frame_system::Config> pallet_order_book::WeightInfo for WeightInfo<T> {
 			.saturating_add(T::DbWeight::get().reads(1))
 			.saturating_add(T::DbWeight::get().writes(1))
 	}
+	// TODO(cdamian): This is copied from `fill_order_full`, don't merge until we get the real weight.
```
cdamian, line 143: Reminders gonna remind.

pallets/order-book/src/lib.rs
```diff
@@ -582,6 +592,98 @@ pub mod pallet {
...
+				Error::<T>::InsufficientOrderSize,
+			);
+
+			ensure!(
+				T::TradeableAsset::can_hold(order.asset_in_id, &account_id, partial_buy_amount),
+				Error::<T>::InsufficientAssetFunds,
+			);
+
+			if partial_fulfillment {
+				Self::update_order(
```
cdamian, line 631: Doing an update here, in this case, prior to the transfer since we are also releasing funds there.

pallets/order-book/src/lib.rs
```diff
@@ -582,6 +592,98 @@ pub mod pallet {
...
+					order.placing_account.clone(),
+					order_id,
+					remaining_buy_amount,
+					order.max_sell_rate,
+					remaining_buy_amount.min(order.min_fulfillment_amount),
```
cdamian, line 636: Note that we're also updating this if the `remaining_buy_amount` is lower than the `order.min_fulfillment_amount`.

pallets/order-book/src/lib.rs
```diff
@@ -582,6 +592,98 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		/// Fill an existing order, based on the provided ratio.
+		#[pallet::call_index(7)]
+		#[pallet::weight(T::Weights::fill_order_partial())]
+		pub fn fill_order_partial(
+			origin: OriginFor<T>,
+			order_id: T::OrderIdNonce,
+			fulfillment_ratio: Perquintill,
```
mustermeiszer, line 602: ~~On a second thought (I know I mentioned using `Perquintill`, so my bad for changing sides now) I think it would be easier to simply use `Balance` here. No rounding, easier for UI for get the "right" value.~~

Forget what I said. It is easier to handle both `sell_amount` and `buy_amount` with that, right? Then lets keep it.

pallets/order-book/src/lib.rs
```diff
@@ -582,6 +592,98 @@ pub mod pallet {
...
+			let partial_buy_amount = fulfillment_ratio.mul_floor(order.buy_amount);
+			let partial_sell_amount = fulfillment_ratio.mul_floor(order.max_sell_amount);
+			let remaining_buy_amount = order
+				.buy_amount
+				.checked_sub(&partial_buy_amount)
+				.ok_or(Error::<T>::RemainingBuyAmountError)?;
+			let partial_fulfillment = !remaining_buy_amount.is_zero();
+
+			ensure!(
+				partial_buy_amount >= order.min_fulfillment_amount,
```
mustermeiszer, line 621: This would mean the `min_fullfillment_amount` is not a global min but rather a `min_per_buy`. I am okay with that, just wanted to channel this by you @hieronx .

pallets/order-book/src/lib.rs
mustermeiszer: EDIT: I would actually still use `Balance` and manually convert the `buy_amount` into a `sell_amount` like we are already doing in `fn inner_update_order`. My main reasoning is that we have no rounding and that the UI can show exactly how much a user will need to "pay".

I.e. using `fn convert_with_ratio`. Its more expensive as we need two additional storage calls, but I hope its worth having not to deal with rounding. 

But of course make a case against that if you think it should be different.

mustermeiszer (commented):
Clean and nice and fast as always. Got some thoughts about the parameters of the call. But otherwise good.

pallets/order-book/src/lib.rs
cdamian: On second thought, I also agree that's the better option. This way we can ditch the `min_fulfillment_amount` update and also be super specific about the exact balance that we are buying, the UI could in turn have a feature to use `Perquintill` go get the actual balance, if needed.

Gonna update this, thank you for checking!

pallets/order-book/src/lib.rs
cdamian: Please let me know if we are OK with this, otherwise we can add a check for the remaining buy amount to ensure that it doesn't go too low, even though I'm not sure if that makes sense ^^

pallets/order-book/src/tests.rs
```diff
@@ -380,6 +389,325 @@ fn fill_order_full_works() {
...
+			let (order_id, _) = get_account_orders(ACCOUNT_0).unwrap()[0];
+
+			let total_balance = OrmlTokens::balance(DEV_AUSD_CURRENCY_ID, &ACCOUNT_1);
+			assert_ok!(OrmlTokens::hold(
+				DEV_AUSD_CURRENCY_ID,
+				&ACCOUNT_1,
+				total_balance
+			));
+
+			// TODO(cdamian): No way of triggering `InsufficientAssetFunds` given that
```
cdamian, line 667: Reminder to double-check this.

pallets/order-book/src/lib.rs
cdamian: Done.

pallets/order-book/src/tests.rs
```diff
@@ -380,6 +389,325 @@ fn fill_order_full_works() {
...
+				RuntimeEvent::OrmlTokens(orml_tokens::Event::Transfer {
+					currency_id: DEV_USDT_CURRENCY_ID,
+					to: ACCOUNT_1,
+					from: ACCOUNT_0,
+					amount: max_sell_amount
+				})
+			);
+			assert_eq!(
+				System::events()[5].event,
+				RuntimeEvent::OrderBook(Event::OrderFulfillment {
```
wischli, line 556: Nit: Could we also ensure the order was cleared from storage? Checking for the event does this implicitly but since clearing an order and dispatching the _full_ fulfillment event are decoupled, we need to check the former as well.

pallets/order-book/src/tests.rs
```diff
@@ -380,6 +389,325 @@ fn fill_order_full_works() {
...
+			assert_noop!(
+				OrderBook::fill_order_partial(
+					RawOrigin::None.into(),
+					order_id,
+					min_fulfillment_amount,
+				),
+				BadOrigin
+			);
+		});
+	}
```
wischli, line 597: 🤘 Thanks for adding this test as well, very thorough! 

Nit: We don't need to set up an order, as the origin check is performed first
```suggestion
	#[test]
	fn fill_order_partial_bad_origin() {
		new_test_ext().execute_with(|| {
			assert_noop!(
				OrderBook::fill_order_partial(
					RawOrigin::None.into(),
					Default::default(),
					Default::default(),
				),
				BadOrigin
			);
		});
	}
```

pallets/order-book/src/tests.rs
wischli: I also ran into this _issue_ during integration tests. AFAIR, I always encountered `BalanceTooLow`. Would be great if we fixed this for better debugging.

pallets/order-book/src/lib.rs
```diff
@@ -582,6 +587,97 @@ pub mod pallet {
...
 		}
+
+		/// Fill an existing order, based on the provided ratio.
+		#[pallet::call_index(7)]
+		#[pallet::weight(T::Weights::fill_order_partial())]
+		pub fn fill_order_partial(
+			origin: OriginFor<T>,
+			order_id: T::OrderIdNonce,
+			buy_amount: T::Balance,
+		) -> DispatchResult {
```
wischli, line 598: Change request: 

There's a lot of code duplication between `fill_order_full` and `fill_order_partial`. We should move this extrinsic's scope to a helper function which is called both from `fill_order_full` and `fill_order_partial`.

wischli (changes requested):
I really appreciate your thoroughness, especially with the unit tests! The logic LGTM but I would like to reduce code duplication. Apart from that a few minor nits which are no blocking.

pallets/order-book/src/tests.rs
cdamian: Yes, I guess there is a bug in the `can_hold` logic that we have, gonna check it out.

pallets/order-book/src/lib.rs
```diff
@@ -127,7 +127,9 @@ pub mod pallet {
 			+ EnsureMul
 			+ EnsureDiv
 			+ TypeInfo
-			+ MaxEncodedLen;
+			+ MaxEncodedLen
+			+ From<u64>
+			+ sp_arithmetic::traits::Unsigned;
```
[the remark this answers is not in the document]
cdamian: Leftover from the initial implementation with `Perquintill`.

pallets/order-book/src/tests.rs
cdamian: Ah, yes that makes sense. Copy/pasta error.

pallets/order-book/src/lib.rs
cdamian: Yup, already was looking into that.

pallets/order-book/src/lib.rs
mustermeiszer: No, I was rather thinking that it is a `one-time` amount. Like, I wanna get at least `x`. But it is fine to go with the given solution.

pallets/order-book/src/lib.rs
```diff
@@ -582,6 +587,97 @@ pub mod pallet {
...
+			let sell_amount = Self::convert_with_ratio(
+				order.asset_in_id,
+				order.asset_out_id,
+				order.max_sell_rate,
+				buy_amount,
+			)?;
+			let remaining_buy_amount = order
+				.buy_amount
+				.checked_sub(&buy_amount)
+				.ok_or(Error::<T>::RemainingBuyAmountError)?;
```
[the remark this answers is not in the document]
mustermeiszer: I disagree here. I would rename the error. But `ensure_sub`-errors are mostly errors that are not happening. IMO in this case we should signal that the user provided the wrong `buy_amount`. Something like, `Error::BuyAmountTooLarge`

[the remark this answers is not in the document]
wischli: Sorry for causing noise. The current approach and @mustermeiszer seems superior indeed.

pallets/order-book/src/tests.rs
```diff
@@ -380,6 +389,325 @@ fn fill_order_full_works() {
...
+						amount: expected_sell_amount
+					})
+				);
+				assert_eq!(
+					System::events()[6].event,
+					RuntimeEvent::OrderBook(Event::OrderFulfillment {
+						order_id,
+						placing_account: order.placing_account,
+						fulfilling_account: ACCOUNT_1,
+						partial_fulfillment: true,
```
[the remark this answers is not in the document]
cdamian: Yes, but it doesn't get to 100 in this test. The 100 percent case is handled separately given the different expected events.

pallets/order-book/src/tests.rs
```diff
@@ -380,6 +389,325 @@ fn fill_order_full_works() {
...
+					})
+				);
+				assert_eq!(
+					System::events()[3].event,
+					RuntimeEvent::OrderBook(Event::OrderUpdated {
+						order_id,
+						account: order.placing_account,
+						buy_amount: remaining_buy_amount,
+						sell_rate_limit: order.max_sell_rate,
+						min_fulfillment_amount: order.min_fulfillment_amount,
```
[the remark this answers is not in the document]
cdamian: Yup, see my comment above.

pallets/order-book/src/tests.rs
cdamian: Sure, adding that check now. Thank you!

pallets/order-book/src/tests.rs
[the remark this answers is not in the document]
wischli: Ah dumb me forgetting that `0..100 != 0..=100` in Rust 🙈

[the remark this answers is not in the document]
cdamian: Nah, that's on me since I did not find a proper way to have a separate test for each percentage using a macro. If we get a failure here now, one would have to debug and figure out which percentage is problematic, which is awkward to say the least.

Maybe @lemunozm can help ^^

pallets/order-book/src/tests.rs
cdamian: Fixed this cc @mustermeiszer

mustermeiszer: Where? ^^ Can not find it.

cdamian: Just pushed, sorry.

pallets/order-book/src/lib.rs
cdamian: Done.

pallets/order-book/src/tests.rs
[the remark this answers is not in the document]
wischli: IMO it's the current test is more than sane. Theoretically, you could add a debug message to the assertion(s) to populate the failure `fulfillment_ratio`.

pallets/restricted-tokens/src/impl_fungibles.rs
```diff
@@ -151,12 +151,15 @@ impl<T: Config> InspectHold<T::AccountId> for Pallet<T> {
...
+				<T::Fungibles as InspectHold<T::AccountId>>::can_hold(asset, who, amount);
+
 			T::PreFungiblesInspectHold::check(FungiblesInspectHoldEffects::CanHold(
 				asset,
 				who.clone(),
 				amount,
-				<T::Fungibles as InspectHold<T::AccountId>>::can_hold(asset, who, amount),
-			))
+				can_hold,
+			)) && can_hold
```
wischli, line 162: Much thanks for fixing! Questioning the thoroughness of our existing `restricted-tokens` test if this was uncaught but that's out of scope for this PR.

runtime/integration-tests/src/evm/precompile.rs
```diff
@@ -156,6 +156,7 @@ async fn axelar_precompile_execute() {
 
 	let command_id = H256::from_low_u64_be(5678);
 
+	#[allow(deprecated)]
```
wischli, line 159: Thanks for the cleanup!

wischli (approved):
LGTM! Great work 🚀

pallets/restricted-tokens/src/impl_fungibles.rs
cdamian: Yup, I also ran the tests locally expecting a failure but everything is green...

mustermeiszer (approved):
LGTM!

PR #1561, opened 2023-09-21
Loans: Remove portfolio entry when removing a loan +  portfolio UT

lemunozm:
# Description

Optional addition: I think it could be beneficial to leave the portfolio values representing the exact loans found in `ActiveLoans`. Anyways, the storage is reset for each NAV computation, but having during some time different elements in the portfolio and active loans could lead to errors.

PROS: Data more structured.
CONS: 1 read/write DB for `close_loan()` action.

## Changes and Descriptions

- add `portfolio.remove_elem()`
- add Unitary tests to the portfolio module.


hieronx (approved):
Definitely makes sense!

mustermeiszer (approved):
Reasonable

PR #1562, opened 2023-09-22
Fix/unweighted process msg

mustermeiszer:
# Description
Weight is not applied for `fn process_msg`. This is the only entry into the really heavy `FI` logic and we should at least apply some weight. 

Fixes https://github.com/centrifuge/centrifuge-chain-internal/issues/49

## Changes and Descriptions
* Make use of fixed weight in `trait WeightInfo` from the `pallet-liquidity-pools-gateway`.

# Checklist:


wischli (approved):

mustermeiszer:
Clsoing in favor of #1565

PR #1565, opened 2023-09-22
feat: benchmark foreign investments worst case

wischli:
# Description

* Adds `ForeignInvestmentBenchmarkHelper` which provides the setup required for benchmarking the heaviest foreign investment call
* Adds `benchmarking` mod to `pallet-liquidity-pools` which benches an incoming `CollectRedeem` message by executing `handle_collect_redemption`
  * NOTE: Skipped the `InboundQueue` since this complicates the investor setup: Right now, the `investor` account is created during the pool creation of `pool_system::PoolBenchmarkHelper` and thus just some Substrate account id. However, for the `InboundQueue` it would have to be a `AccountConverter<Runtime, LocationToAccountId>::convert((sender_domain, investor_bytes))`.
* Feature creep: Renames `pallet_foreign_investments::<T>::Config::Rate` to `BalanceRatio` 

## TODO

- [ ] Discuss how to move forward: Do we want to defensively stick to the hardcoded weights?
  * If so, we should not add `LiquidityPools` to the runtimes' benchmark list 
  * Else, I would continue writing benchmarks for existing extrinsics such that we have everything covered properly: an inbound messages by the FI worst case (`handle_collect_redemption`) and extrinsics by their corresponding benchmarks
- [ ] Maybe: Right now, the `pallet-liquidity-pools` benchmarks cannot be unit tested, because the existing mock is insufficient. If running unit tests is desired, I propose to enable this in a subsequent PR.

## Justification about heaviest FI call

* In the context of `pallet-foreign-investments`, the heaviest action (measured by amount of reads and writes) is collecting a redemption when there exists an `InvestmentState` which swaps from `foreign` to `pool`
* By collecting the redemption, a swap from `pool` to `foreign` emerges which needs to be resolved in `handle_concurrent_swap_orders` by comparing with the existing swap from the investment

## About the weight

For presentation purposes, I temporarily added the generated weight file. This must be removed before merging as otherwise executing benchmarks for `pallet-liquidity-pools` would remove all hardcoded values.

When looking at the results of executing the benchmark on my local machine, the total weight is ~55% of the current hardcoded one. The PoV size is ~80% of the hardcoded one. For details, see the [WIP Weights Sheet](https://docs.google.com/spreadsheets/d/1X0N-TJg1R3Lsr1ILqW0iVWdl5etlXYXPMobzFWpxdCw/edit#gid=0). I suppose the locally generated weights are close enough to our benchmark runner, since the results are similar.

```
		Weight::from_parts(236_000_000, 71940)
			.saturating_add(T::DbWeight::get().reads(20))
			.saturating_add(T::DbWeight::get().writes(17))
```

# Checklist:


pallets/foreign-investments/src/lib.rs
```diff
@@ -152,7 +152,7 @@ pub mod pallet {
 
 		/// Type for price ratio for cost of incoming currency relative to
 		/// outgoing
-		type Rate: Parameter
+		type BalanceRatio: Parameter
```
wischli, line 155: Sorry for feature creeping. This change does not have any functional impact.

pallets/liquidity-pools/src/weights_generated.rs
```diff
@@ -0,0 +1,78 @@
+
+//! Autogenerated weights for `pallet_liquidity_pools`
```
wischli, line 2: NOTE: Committed solely for demonstration purpose, will be removed before merging!

pallets/foreign-investments/src/impls/benchmark_utils.rs
```diff
@@ -0,0 +1,164 @@
...
+	fn bench_prepare_foreign_investments_setup() -> (
+		T::AccountId,
+		T::InvestmentId,
+		T::CurrencyId,
+		T::CurrencyId,
+		T::AccountId,
+	) {
+		let pool_id = Default::default();
+		let pool_admin: T::AccountId = frame_benchmarking::account("pool_admin", 0, 0);
+		<T::PoolInspect as PoolBenchmarkHelper>::bench_create_pool(pool_id, &pool_admin);
```
lemunozm, line 73: Comment: I really like how the `bench_*` methods are called inside `bench_*` methods, leaving a clean API to use from the benchmark side with 0 knowledge about what's happening under the hood. i.e. the `pool_id` used is totally transparent here ❤️

pallets/pool-system/src/impls.rs
```diff
@@ -521,6 +522,14 @@ mod benchmarks_utils {
 				FUNDS.into(),
 			));
 
+			// Fund pool account
+			let pool_account = PoolLocator { pool_id }.into_account_truncating();
+			frame_support::assert_ok!(T::Tokens::mint_into(
+				POOL_CURRENCY,
+				&pool_account,
+				POOL_ACCOUNT_BALANCE.into()
+			));
```
lemunozm, line 531: Just curious, why is this minting now needed?

lemunozm (commented):
Love how clean are you getting this given that is quite difficult benchmark. 💯

mustermeiszer (dismissed):
Sick! Whats your final feeling regarding the adaption of the `process_msg` weight? 

I am personally leaning torwards keeping it as is an heavier. But might be paranoia... 

@wischli could you include the changes from #1562, then we do not need to wait for CI 2 times.

libs/traits/src/benchmarking.rs
```diff
@@ -58,3 +58,45 @@ pub trait OrderBookBenchmarkHelper {
...
+	///  * The foreign currency id
+	///  * A trading account which can bidirectionally fulfill swap orders for
+	///    the (foreign, pool) currency pair
+	fn bench_prepare_foreign_investments_setup() -> (
+		Self::AccountId,
+		Self::InvestmentId,
+		Self::CurrencyId,
+		Self::CurrencyId,
+		Self::AccountId,
+	);
```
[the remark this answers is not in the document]
wischli: [239fa89](https://github.com/centrifuge/centrifuge-chain/pull/1565/commits/239fa899c9f38c61e0dd0003f0fb5a38aa01d46e)

libs/traits/src/benchmarking.rs
```diff
@@ -67,6 +67,7 @@ pub trait ForeignInvestmentBenchmarkHelper {
 	type Balance;
 	type CurrencyId;
 	type InvestmentId;
+	type SetupInfo;
```
lemunozm, line 70: I think it doesn't need to be generalized here. The new struct can just be used directly and even can be in the traits module (i.e. just before the trait). I mean, it's not a type the user needs to configure in runtime or similar. It's just an aggregation of fields to avoid using a nameless tuple. WDYT?

wischli: That was my initial thought but then I assumed you'd prefer it to be generic. Will follow your proposal then to make it absolutely explicit!

pallets/pool-system/src/impls.rs
wischli: This was taken from the integration tests when I was debugging `BalanceTooLow` errors. So it might not be necessary but I believe it still is.

When we process a redemption of tranche tokens, the pool account needs to have native balance which is exchanged for these tranche tokens. 

In contrast, we don't need to fund the pool account with tranche tokens as these are minted when processing an investment.

mustermeiszer (approved):

PR #1566, opened 2023-09-22
centrifuge: Anemoy pool currency migration

NunoAlexandre:
# Description

We want to migrate the `currency` of the `Anemoy` pool from `LpEthUSDC` to (Polkadot-native) `USDC` on Centrifuge.

- Anemoy has pool id 4,139,607,887 (the only pool currently registered on Centrifuge)
- We will migrate the currency of that pool from `ForeignAsset(100,001)` (`LpEthUSDC`) to `ForeignAsset(6)`  (`Circle’s USDC native on Polkadot`)
- Both currencies have 6 decimals so no extra considerations in that regard

## To Do

- [x] You can just add a sanity check to the migration that the pool value is indeed 0 and there are no locked orders
  - [x] For all tranches within the Anemoy pool, verify that are no entries on pallet_investments' `InvestOrders` , `RedeemOrders` , `ActiveInvestOrders` and `ActiveRedeemOrders`.
  - [x] Check that pool value is `0`

### DIFF
--- a/libs/types/src/tokens.rs
+++ b/libs/types/src/tokens.rs
@@ -187,8 +187,8 @@ where
 )]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 pub struct TrancheCurrency {
-	pub(crate) pool_id: PoolId,
-	pub(crate) tranche_id: TrancheId,
+	pub pool_id: PoolId,
+	pub tranche_id: TrancheId,
 }
 
 impl From<TrancheCurrency> for CurrencyId {
--- a/pallets/investments/src/lib.rs
+++ b/pallets/investments/src/lib.rs
@@ -224,7 +224,7 @@ pub mod pallet {
 		StorageMap<_, Blake2_128Concat, T::InvestmentId, OrderId, ValueQuery>;
 
 	#[pallet::storage]
-	pub(crate) type InvestOrders<T: Config> = StorageDoubleMap<
+	pub type InvestOrders<T: Config> = StorageDoubleMap<
 		_,
 		Blake2_128Concat,
 		T::AccountId,
@@ -234,7 +234,7 @@ pub mod pallet {
 	>;
 
 	#[pallet::storage]
-	pub(crate) type RedeemOrders<T: Config> = StorageDoubleMap<
+	pub type RedeemOrders<T: Config> = StorageDoubleMap<
 		_,
 		Blake2_128Concat,
 		T::AccountId,
@@ -245,7 +245,7 @@ pub mod pallet {
 
 	#[pallet::storage]
 	#[pallet::getter(fn acc_active_invest_order)]
-	pub(crate) type ActiveInvestOrders<T: Config> =
+	pub type ActiveInvestOrders<T: Config> =
 		StorageMap<_, Blake2_128Concat, T::InvestmentId, TotalOrder<T::Amount>, ValueQuery>;
 
 	#[pallet::storage]
--- a/pallets/investments/src/lib.rs
+++ b/pallets/investments/src/lib.rs
@@ -250,7 +250,7 @@ pub mod pallet {
 
 	#[pallet::storage]
 	#[pallet::getter(fn acc_active_redeem_order)]
-	pub(crate) type ActiveRedeemOrders<T: Config> =
+	pub type ActiveRedeemOrders<T: Config> =
 		StorageMap<_, Blake2_128Concat, T::InvestmentId, TotalOrder<T::Amount>, ValueQuery>;
 
 	#[pallet::storage]
--- a/runtime/centrifuge/src/migrations.rs
+++ b/runtime/centrifuge/src/migrations.rs
@@ -9,7 +9,7 @@
 // but WITHOUT ANY WARRANTY; without even the implied warranty of
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU General Public License for more details.
-use crate::{Investments, PoolSystem, Runtime, Weight};
+use crate::{Runtime, Weight};
 
 pub type UpgradeCentrifuge1021 = anemoy_pool::Migration;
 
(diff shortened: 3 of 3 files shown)


libs/types/src/tokens.rs
```diff
@@ -187,8 +187,8 @@ where
 )]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 pub struct TrancheCurrency {
-	pub(crate) pool_id: PoolId,
-	pub(crate) tranche_id: TrancheId,
+	pub pool_id: PoolId,
```
NunoAlexandre, line 190: any objections to open these fields visibility? by having them `pub(crate)` I would have to import he `TrancheCurrencyT` trait to call `generate` to build an instance of this type which is quite an overkill imo. let me know if I am not overseeing any requirement 👍

runtime/centrifuge/src/migrations.rs
```diff
@@ -9,5 +9,92 @@
...
+
+			log::info!("anemoy_pool::Migration: post_upgrade succeeded");
+			Ok(())
+		}
+	}
+
+	// todo(nuno): also check that pool value is 0 and check also that Investments::InvestOrders and
+	// Investments::RedeemOrder have no entries from Anemoy; the latter ones seem tricky at first
+	// sight since they are double maps first keyed by an AccountId, meaning we need to transverse
+	// that first which is more costly.
```
wischli, line 93: Should suffice to check `StorageDoubleMap::<T>::iter_keys().count().is_zero()`.

NunoAlexandre: Can you expand on that? Because we need to iter keys on the account (first storage key) and then filter by the investment id (pool and tranche); if we just do iter_keys we are of taking any other pool into account.

If the idea is to use the assumption that anemoy is the only pool, then it makes sense. But it sort of defeats or weakens the point of this sanity check in the first place (tho, sure, it's less likely there will be another pool being added than a new order for anemoy).

Wtdyt? If everyone is fine with that I'm not blocking the approach

NunoAlexandre: I'll just filter the iter_keys involving anemoy to make this 100% robust.

NunoAlexandre:
Implemented the sanity checks for orders under the Anemoy pool and tested again with `try-runtime`. Looking good:

```
2023-09-23 14:15:57.471  INFO main centrifuge_runtime::migrations::anemoy_pool: anemoy_pool::Migration: currency set to USDC ✓
2023-09-23 14:15:57.471  INFO main centrifuge_runtime::migrations::anemoy_pool: anemoy_pool::Migration: post_upgrade succeeded ✓
```

Still have to implement the check for the pool value being 0.

runtime/centrifuge/src/migrations.rs
```diff
@@ -9,5 +9,126 @@
...
+			pallet_investments::ActiveInvestOrders::<Runtime>::iter_keys()
+				.filter(|investment| investment.pool_id == ANEMOY_POOL_ID)
+				.count() == 0 && pallet_investments::ActiveInvestOrders::<Runtime>::iter_keys()
+				.filter(|investment| investment.pool_id == ANEMOY_POOL_ID)
+				.count() == 0 && pallet_investments::InvestOrders::<Runtime>::iter_keys()
+				.filter(|(_, investment)| investment.pool_id == ANEMOY_POOL_ID)
+				.count() == 0 && pallet_investments::RedeemOrders::<Runtime>::iter_keys()
+				.filter(|(_, investment)| investment.pool_id == ANEMOY_POOL_ID)
+				.count() == 0;
+
```
mustermeiszer, line 118: That looks soo clean! nice!

mustermeiszer (dismissed):
Looks good!

runtime/centrifuge/src/migrations.rs
NunoAlexandre: thanks 🙏 just pushed the check for pool value == 0 and realised that we were checking `ActiveInvestOrders` twice instead of checking `ActiveRedeemOrders` :) fixed

NunoAlexandre:
This PR is now feature-complete. Just testing it one last time 👍

pallets/investments/src/lib.rs
```diff
@@ -224,7 +224,7 @@ pub mod pallet {
 		StorageMap<_, Blake2_128Concat, T::InvestmentId, OrderId, ValueQuery>;
 
 	#[pallet::storage]
-	pub(crate) type InvestOrders<T: Config> = StorageDoubleMap<
+	pub type InvestOrders<T: Config> = StorageDoubleMap<
```
wischli, line 227: This enables writing these storages from outside which is not desired. We should at least add a note to revert this change after removing the migration. 

Since we only read, it appears to me that it would be beneficial on using existing /adding `getters`.

runtime/centrifuge/src/migrations.rs
```diff
@@ -9,5 +9,127 @@
...
+				log::info!("anemoy_pool::Migration: currency set to USDC ✓");
+			});
+
+			weight.saturating_add(
+				<Runtime as frame_system::Config>::DbWeight::get().reads(
+					pallet_pool_system::Pool::<Runtime>::iter_keys()
+						.count()
+						.saturating_mul(2) as u64,
+				),
+			)
```
wischli, line 83: Why are we adding two reads multiplied with the number of pool keys? Shouldn't it just add single read and a write?

wischli (commented):
The migration LGTM! I would like to raise my concerns about making the `pallet-investment` storages public.

runtime/centrifuge/src/migrations.rs
NunoAlexandre: fixed. I guess the result of moving coding around and changes to the implementation. Made this add 1 read and 1 write here, for the pool lookup and mutate, respectively. thanks!

pallets/investments/src/lib.rs
NunoAlexandre: Opened an issue to take care of this and other things after the upgrade: rhttps://github.com/centrifuge/centrifuge-chain/issues/1567

We have full control of what code makes it's way to production so it's not that opening up this visibility puts us at danger that ill-intended users or ourselves corrupting those storages. Adding `getters` imo is just adding code that's not needed aside from this migration that's to be executed once and deleted after so I'd prefer to keep this as is.

wischli (approved):
LGTM! Thanks for adding the post-merge issue!

PR #1568, opened 2023-09-25
fix: typos

omahs:
fix: typos


lemunozm (dismissed):
Thanks for the corrections! 🙌🏻

docs/runtime-upgrade.md
```diff
@@ -43,15 +43,15 @@ This must be done before releasing a new version of any of our runtimes to ensur
 
     - Click on **propose**
 
-1. Voting phase. You should make 3 votes in favor with 3 diferent people (modify the CFG amount before to a low number).
+1. Voting phase. You should make 3 votes in favor of 3 different people (modify the CFG amount before to a low number).
```
NunoAlexandre, line 46: That correction is not actually correct, since the desired semantic here is that 3 different accounts vote `Aey`, note that someone "should make 3 votes in favour of 3 different people".

```suggestion
1. Voting phase. Three different account should vote in favour of the proposal (modify the CFG amount before to a low number).
```

NunoAlexandre: I'm also not sure we need 3 separate 👍 , 1 should be enough? 🤔  cc @wischli

NunoAlexandre (changes requested):
Thanks for contributing! One thing to fix

docs/runtime-upgrade.md
wischli: > I'm also not sure we need 3 separate 👍 , 1 should be enough? 🤔 cc @wischli

Our local chain specs setup 4 genesis council members

https://github.com/centrifuge/centrifuge-chain/blob/0387385954349e3ca8b2c4d84e4e17b16bbbc5db/src/chain_spec.rs#L541-L543

Since both proposing an external referendum as well as fast-tracking requires at least half of the council, two votes would suffice. I suppose this is a remnant of previously requiring at least 75% of the council.

On that note, our runtime upgrade scripts votes with three members:

https://github.com/centrifuge/centrifuge-chain/blob/0387385954349e3ca8b2c4d84e4e17b16bbbc5db/scripts/js/upgrade/index.js#L87-L93

wischli: The number of votes must match the threshold set when proposing. The document currently states 3 (75%) which is outdated
```diff
1. In Governance -> Council -> Motions, click on **propose motion**.
-    - At **umbral**, put `3` (a 75% of the total which is 4).
+    - At **umbral**, put `2` (a 50% of the total which is 4).
```

wischli (dismissed):
Thanks for fixing our typos! As stated, some content of the readme should be updated as well  most ideally. However, I am fine with merging these fixes as the content is not incorrect and fixing the details is out of scope of this PR in my opinion.

docs/runtime-upgrade.md
NunoAlexandre, line 46:
```suggestion
1. Voting phase. The appropriate threshold of votes should be placed in favour of the proposal.
```

NunoAlexandre: Following the discussion at https://github.com/centrifuge/centrifuge-chain/pull/1568/files#r1336927624, let's fix this to a more generic sentence that is correct for any of the runtimes and is semantically correct.

NunoAlexandre (changes requested):

NunoAlexandre (approved):
Thanks for your contribution 🙏

mustermeiszer:
@omahs still interested in bringing this in?

omahs:
@mustermeiszer Hey, yes, should I edit something?

mustermeiszer (approved):
LGTM! Thanks for your contribution.

wischli (approved):
Thanks for your contribution!

PR #1570, opened 2023-09-26
fix: min fulfillment amount for partial fulfillment

wischli:
# Description

This PR mainly fixes issues for partial orderbook fulfillments of foreign investments as a successor of #1559. It also enables partial investments for orders created via extrinsics:

* Fixes issues for partial swap order fulfillments in `pallet-foreign-investments`
* Enables partial swaps for orders created through user extrinsics `pallet_order_book::create_order` and `pallet_order_book::user_update_order` which had the `min_fulfillment_amount` set to the `buy_amount` making partial fulfillments impossible
* Removes `min_fulfillment_amount` from `TokenSwaps` trait
  * IMO, it should solely be handled by the orderbook
* Adds `NativeBalanceDecimalConverter` which provides means of converting between native and asset balance (i.e. re-denomination)
* Adds `MinFulfillmentAmountNative` to `pallet_order_book::Config` which provides a global threshold which gets converted into the incoming asset precision via the also new `pallet_order_book::Config::DecimalConverter` type

## Integration tests

* Extends FI integration tests to check for correct `Executed*` dispatches
* Adds FI integration test for partial order fulfillment in edge case of having an order from pool to foreign as a result of decreasing an investment and collecting a redemption

# Checklist:

### DIFF
--- a/pallets/foreign-investments/src/hooks.rs
+++ b/pallets/foreign-investments/src/hooks.rs
@@ -79,18 +79,28 @@ impl<T: Config> StatusNotificationHook for FulfilledSwapOrderHook<T> {
 					Error::<T>::FulfilledTokenSwapAmountOverflow
 				);
 
-				let invest_swap = SwapOf::<T> {
-					amount: active_invest_swap_amount,
-					..status
-				};
-				let redeem_swap = SwapOf::<T> {
-					amount: status.amount.ensure_sub(active_invest_swap_amount)?,
-					..status
-				};
+				// Order was fulfilled at least for invest swap amount
+				if status.amount > active_invest_swap_amount {
+					let invest_swap = SwapOf::<T> {
+						amount: active_invest_swap_amount,
+						..status
+					};
+					let redeem_swap = SwapOf::<T> {
+						amount: status.amount.ensure_sub(active_invest_swap_amount)?,
+						..status
+					};
 
-				// NOTE: Fulfillment of invest swap before redeem one for no particular reason
-				Self::fulfill_invest_swap_order(&info.owner, info.id, invest_swap, false)?;
-				Self::fulfill_redeem_swap_order(&info.owner, info.id, redeem_swap)
+					// NOTE: Fulfillment of invest swap before redeem one for no particular reason.
+					// If we wanted to fulfill the min swap amount, we would have to add support for
+					// oppression of for swap updates to `fulfill_redeem_swap_order` as well in case
+					// redeem_swap.amount < status.amount < invest_swap.amount
+					Self::fulfill_invest_swap_order(&info.owner, info.id, invest_swap, false)?;
+					Self::fulfill_redeem_swap_order(&info.owner, info.id, redeem_swap)
+				}
+				// Order was fulfilled below invest swap amount
+				else {
+					Self::fulfill_invest_swap_order(&info.owner, info.id, status, true)
+				}
 			}
 			_ => {
 				log::debug!("Fulfilled token swap order id {:?} without advancing foreign investment because swap reason does not exist", id);
--- a/pallets/foreign-investments/src/impls/mod.rs
+++ b/pallets/foreign-investments/src/impls/mod.rs
@@ -14,7 +14,8 @@
 
 use cfg_traits::{
 	investments::{ForeignInvestment, Investment, InvestmentCollector, TrancheCurrency},
-	IdentityCurrencyConversion, PoolInspect, StatusNotificationHook, TokenSwaps,
+	ConversionToAssetBalance, IdentityCurrencyConversion, PoolInspect, StatusNotificationHook,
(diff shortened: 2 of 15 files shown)


runtime/common/src/lib.rs
```diff
@@ -484,6 +486,80 @@ pub mod foreign_investments {
...
+					convert_balance_decimals(
+						cfg_primitives::currency_decimals::NATIVE,
+						to_decimals,
+						balance,
+					)
+					.map_err(DispatchError::from)
+				}
+				_ => Err(DispatchError::Token(sp_runtime::TokenError::Unsupported)),
+			}
+		}
```
lemunozm, line 528: Question 1: as I read this method, it seems like I want to transform a `currency_in` into an asset balance, but the implementation seems to do the opposite, transforming an asset balance to a native one. (Sure, I'm missing something). Is `currency_in` actually `currency_out`?

Question 2: Does it make sense to call this method with a native balance? Or should instead be an error?

lemunozm (commented):
Awesome how fast you're adding new features 🙌🏻. Just a couple of questions, but overall LGTM!

runtime/common/src/lib.rs
wischli: 1. I followed the naming convention from the orderbook and foreign investments such that `currency_in` refers to the _incoming_ asset for which we want the balance translation. I agree it is not optimal, probably `currency_to` much better. We could also just remove the direction and call it `currency_id`. WDYTH?

2. Since calling it with native balance does not introduce any overhead, I think it should not error out because that makes it more flexible. However, I do not have a strong opinion here.

pallets/foreign-investments/src/impls/mod.rs
```diff
@@ -827,8 +825,7 @@ impl<T: Config> Pallet<T> {
 					swap.amount,
 					// The max accepted sell rate is independent of the asset type for now
 					T::DefaultTokenSellRatio::get(),
-					// The minimum fulfillment must be everything
-					swap.amount,
+					// Convert default min fulfillment amount from native to incoming currency
```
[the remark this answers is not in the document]
wischli: Great catch, it was a remnant from [0ae5314](https://github.com/centrifuge/centrifuge-chain/pull/1570/commits/0ae5314b10450bdb68d08a9b164b81627db4d90c).

pallets/foreign-investments/src/hooks.rs
```diff
@@ -79,18 +79,28 @@ impl<T: Config> StatusNotificationHook for FulfilledSwapOrderHook<T> {
...
+					// If we wanted to fulfill the min swap amount, we would have to add support for
+					// oppression of for swap updates to `fulfill_redeem_swap_order` as well in case
+					// redeem_swap.amount < status.amount < invest_swap.amount
+					Self::fulfill_invest_swap_order(&info.owner, info.id, invest_swap, false)?;
+					Self::fulfill_redeem_swap_order(&info.owner, info.id, redeem_swap)
+				}
+				// Order was fulfilled below invest swap amount
+				else {
+					Self::fulfill_invest_swap_order(&info.owner, info.id, status, true)
+				}
```
mustermeiszer, line 103: Do I read this correctly, that we are favouring `Investment`-swaps over `Redepemtion`-swaps? I.e. we take care of fulfilling investments first.

mustermeiszer (approved):
That was fast!

pallets/foreign-investments/src/hooks.rs
wischli: Yes. Last night I started refactoring to prioritize lower orders instead. However, this requires adding support for oppressing updating swaps as part of `fulfill_redeem_swap_order` (i.e. `apply_redeem_state_transition`) similar to `fulfill_invest_swap_order`. This is quite tedious and creates more edge cases and should be tested thoroughly. Given that we would like to cut the release ASAP and me feeling confident with the current MVP, I propose to handle this in a subsequent PR.

wischli: On a note: This favor of `investment`-swaps only occurs if both the invest state as well as redeem state are swapping into foreign such that the orderbook swap is "sum" of both.

runtime/common/src/lib.rs
lemunozm: 1. Probably it's me that I feel dyslexic as with right or left 😆. I think I asked about this like 10 times haha. Given that the convention in other places is the same, we can leave it as it is. `currency_id` also fits 👍🏻 

Ok, if semantically makes sense that I could call `to_asset_balance()` using a native balance expecting no change on it, then I'm fine with the current implementation.

pallets/foreign-investments/src/hooks.rs
mustermeiszer: Thanks for the context.

PR #1571, opened 2023-09-27
Investments: simplify constraints and remove InvestmentProperties trait

lemunozm:
# Description

Some simplifications regarding investment info/properties. [Slack conversation](https://kflabs.slack.com/archives/C04GJQAM9P0/p1695383388075129)

## Changes and Descriptions

- Removed the `InvestmentProperties` trait. Now, the type `InvestmentInfo` is used directly (which I think simplifies some parts)
- Removed the need for the `where` bound everywhere the investment pallets is used to determine the correct `InvestmentInfo` (this simplifies things for future integration tests)

### DIFF
--- a/pallets/investments/src/lib.rs
+++ b/pallets/investments/src/lib.rs
@@ -16,8 +16,7 @@
 use cfg_primitives::OrderId;
 use cfg_traits::{
 	investments::{
-		Investment, InvestmentAccountant, InvestmentCollector, InvestmentProperties,
-		InvestmentsPortfolio, OrderManager,
+		Investment, InvestmentAccountant, InvestmentCollector, InvestmentsPortfolio, OrderManager,
 	},
 	PreConditions, StatusNotificationHook,
 };
@@ -109,7 +108,7 @@ pub enum CollectType {
 
 #[frame_support::pallet]
 pub mod pallet {
-	use cfg_types::investments::ForeignInvestmentInfo;
+	use cfg_types::investments::{ForeignInvestmentInfo, InvestmentInfo};
 	use sp_runtime::{traits::AtLeast32BitUnsigned, FixedPointNumber, FixedPointOperand};
 
 	use super::*;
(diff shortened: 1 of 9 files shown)


pallets/investments/src/lib.rs
```diff
@@ -595,7 +572,7 @@ where
 			amount,
 		})?;
 
-		let info = T::Accountant::info(investment_id).map_err(|_| Error::<T>::UnknownInvestment)?;
+		let _ = T::Accountant::info(investment_id).map_err(|_| Error::<T>::UnknownInvestment)?;
```
lemunozm, line 575: Actually, this line could be removed and saved as read access to the database, but there is a test that checks for the error `UnknownInvestment`. If removed a future `Accountant::transfer()` fails with a different error.

pallets/investments/src/lib.rs
```diff
@@ -662,7 +638,7 @@ where
 		who: T::AccountId,
 		investment_id: T::InvestmentId,
 	) -> DispatchResultWithPostInfo {
-		let info = T::Accountant::info(investment_id).map_err(|_| Error::<T>::UnknownInvestment)?;
+		let _ = T::Accountant::info(investment_id).map_err(|_| Error::<T>::UnknownInvestment)?;
```
lemunozm, line 641: Same here

pallets/investments/src/lib.rs
```diff
@@ -955,7 +928,7 @@ where
 			&mut total_order.amount,
 		)?;
 
-		T::Accountant::transfer(info.id(), send, recv, transfer_amount)
+		T::Accountant::transfer(investment_id, send, recv, transfer_amount)
```
lemunozm, line 931: `info.id()` and `investment_id` always match with the same value

mustermeiszer (approved):
Changes look sane. But I am still thinking whether it is worth the refactor. Just am wondering whether we need to have that abstraction for `pallet-linked-pools`. 

But I we can go back and change it again if we want to.

lemunozm:
Ok! Happy to revert some parts if needed in the future.

Note that the `InvestmentAccountant` trait is still generic over `InvestmentInfo`. So it is not fixed to always use the `InvestmentInfo` type.

Issue #1499, opened 2023-08-16
Pallet-pool-registry: add test for write-off policy creation


offerijns:
I would add a test for checking they are added validly here.

_Originally posted by @mustermeiszer in https://github.com/centrifuge/centrifuge-chain/pull/1347#discussion_r1195093960_

Issue #1375, opened 2023-05-30
Loans: Oracle integration tests


lemunozm:
### Description

Add an end-to-end test for a loan with an Oracle price. 

The steps are:
- Add an Oracle member.
- Price a price id with that Oracle member
- Create a loan with external valuation.
- Update the price id.
- Update the portfolio and check if it has changed.

PR #1574, opened 2023-09-28
Loans: integration tests (experimental runtime tests framework)

### LINKED ISSUE #1458: Loans: add integration tests
### Description

Currently, there is no specific integration test for loan use cases. We need to add it and also to tests runtime APIs.

Loan use-cases logic is very tied to other pallets. So it involves setting up pools, assets, registries, investors, etc...

lemunozm:
# Description

Fixes #1458 #1375 #1499

This PR adds a "new way" of doing integration tests, but only when the purpose is to test the interaction between different pallets in runtimes. No node, no fudge, no 3parties, no xcm, just pallet interactions in real runtimes. [Slack thread](https://kflabs.slack.com/archives/C04GJQAM9P0/p1695725295059549) with the initial discussion.

In order to improve compilation times and testing feedback loop, I've created a new crate under `runtime/tests` where tests can be compiled much faster. cc @mustermeiszer

### Motivation
Create a way to test end-to-end paths in a runtime context independently of the runtime used.

Example. The following parameterized test: 

```rust
fn test_foo<T: Config>() {
    // Test over a T runtime configuration common to all our runtimes,
    // with some common instantiations to improve the usage:
    // Balance as u128, CurrencyId, AccountId32, ...
    
    // ...
    
    // If the test requires different choices for each runtime, you can dispatch as follows:
    match T::KIND {
        RuntimeKind::Development => ...,
        RuntimeKind::Altair => ...,
        RuntimeKind::Centrifuge => ...,
    }
}

// Inject the tests for all runtimes using the `genesis()` to setup the initial stage.
crate::test_with_all_runtimes!(genesis, test_foo);
```

will create the following output:
```
running 3 tests
test test_foo::altair ... ok
test test_foo::centrifuge ... ok
test test_foo::development ... ok
```

## Changes and Descriptions

- [x] Runtime tests framework
- [ ] Loan integration tests (with oracles and runtime changes)


lemunozm:
Close in favor of: #1583

PR #1497, opened 2023-08-15
feat: apply polkadot v0.9.43 to docker 

wischli:
# Description

* Updates rococo local to polkadot v0.9.43 obtained via
```
docker run --platform=linux/amd64 parity/polkadot:v0.9.43 build-spec --disable-default-bootnode --chain rococo-local --raw > res/rococo-local.json
```
* Fixes docker compose files after [breaking changes CLI from polkadot v0.9.43](https://github.com/paritytech/polkadot/releases/tag/v0.9.43):
  * Removes `ws-port` which now is just the RPC port
  * Maps outer docker rpc port to `9944` for relay alice, `9945` for relay bob and `9946` for para charlie

## How to test
onboard.sh
```
cc_docker_image_tag="${PARA_DOCKER_IMAGE_TAG:-test-main-latest}"
parachain="${PARA_CHAIN_SPEC:-centrifuge-local}"

export PARA_DOCKER_IMAGE_TAG=$cc_docker_image_tag
export CC_DOCKER_TAG=$cc_docker_image_tag
export PARA_CHAIN_SPEC=$parachain
export DOCKER_ONBOARD=true

./scripts/init.sh start-relay-chain
./scripts/init.sh start-parachain-docker
./scripts/init.sh onboard-parachain
```

```
$ PARA_CHAIN_SPEC=altair-local ./scripts/onboard.sh
```

# Checklist:

- [ ] I have added Rust doc comments to structs, enums, traits and functions
- [ ] I have made corresponding changes to the documentation
- [ ] I have performed a self-review of my code
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1577, opened 2023-09-29
Polkadot v0.9.43

NunoAlexandre:
Polkadot v0.9.43


wischli:
Thanks for tackling this beast. Once the compiler is satisfied, we should not forget to pull in changes from https://github.com/centrifuge/centrifuge-chain/pull/1497

runtime/altair/src/lib.rs
```diff
@@ -999,9 +999,9 @@ impl pallet_crowdloan_claim::Config for Runtime {
 parameter_types! {
 	pub const PotId: PalletId = cfg_types::ids::STAKE_POT_PALLET_ID;
 	#[derive(scale_info::TypeInfo, Debug, PartialEq, Eq, Clone)]
-	pub const MaxCandidates: u32 = 100;
+	pub const MaxCandidates: u32 = 20;
 	pub const MinCandidates: u32 = 5;
-	pub const MaxVoters: u32 = 1000;
+	pub const MaxVoters: u32 = 100;
```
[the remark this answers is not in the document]
NunoAlexandre: these values were extremely high, leading to this failing test:

```
failures:

---- __construct_runtime_integrity_test::runtime_integrity_tests stdout ----
thread '__construct_runtime_integrity_test::runtime_integrity_tests' panicked at /Users/n/.cargo/git/checkouts/substrate-ec832a859c3205cb/548a8e1/frame/elections-phragmen/src/lib.rs:319:13:
election weight Weight(ref_time: 71432120000, proof_size: 6747840)s (0.07143212) will exceed a 0.5s chain's block weight (Weight { ref_time: 500000000000, proof_size: 5242880 }) (MaxCandidates 100, MaxVoters 1000, MaxVotesPerVoter 5 -- tweak these parameters)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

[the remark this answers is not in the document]
lemunozm: Thanks!

pallets/block-rewards/src/tests.rs
```diff
@@ -5,7 +5,7 @@ use sp_runtime::traits::BadOrigin;
 use super::*;
 use crate::mock::*;
 
-const REWARD: u128 = 100;
+const REWARD: u128 = 100 * crate::mock::ExistentialDeposit::get();
```
lemunozm, line 8: Maybe it should be an addition instead?

NunoAlexandre: It doesn't really matter, as long as it's more than the ED and the genesis balance is sufficient. I made it 100 * ED since in some tests we set the total reward to 2 or 10 * REWARD or claim twice, so 100+ would work.

lemunozm: That weird. It works for me using `+` there. Maybe it was another error that is already fixed?

I really think having a `+` there improves understanding of what's going on in the tests.

NunoAlexandre: Maybe, I don't know. And I don't share the feeling that + improves the understanding over, both operations basically set the reward to a value higher than the ED itself. We should just leave a comment saying that the Reward should be bigger than the ED and maybe make the reward some more reasonable value like 1 CFG for example.

pallets/loans/src/util.rs
```diff
@@ -57,7 +57,7 @@ impl<T: Config> DataProvider<T::PriceId, T::Rate> for NoPriceRegistry<T> {
 }
 
 impl<T: Config> DataFeeder<T::PriceId, T::Rate, T::AccountId> for NoPriceRegistry<T> {
-	fn feed_value(_: T::AccountId, _: T::PriceId, _: T::Rate) -> DispatchResult {
+	fn feed_value(_: Option<T::AccountId>, _: T::PriceId, _: T::Rate) -> DispatchResult {
```
lemunozm, line 60: This is the issue related to `pallet-loans` benchmarks. Quite funny BTW: The `account()` benchmark method takes the `Account` type for deserializing from the destination of its usage. Because now it has changed to `Option<Account>`, it deserializes wrongly.

lemunozm: Written down a tech-debt to myself: Now orml is updated this part can be simplified.

pallets/block-rewards/src/tests.rs
NunoAlexandre: Sorry if I am not understanding your point of view, but I really don't follow what we are discussing here. 

> you are setting more total rewards than you need

Why do I need to set it to a bare minimum? Does it even matter?

> With `+`, you set exactly the amount used in the tests.

Isn't this the amount used in the tests? Isn't that exactly the purpose of this constant? 🤔 what am I missing here? Keep in mind I've gone through hundreds of failing tests so I might be forgetting some detail here.

> Actually, multiplying Balance by itself is an operation that should not "exist". EUR * EUR does not exist.

Really don't follow what the discussion is here. By setting the mock test reward value to 100 times the ED, we are not "multiplying Balance by itself", we are multiplying ED (a value of type Balance) by a magnitude of 100. 100 is the multiplier in this case, but it's not quite that relevant; what's relevant is that `REWARD` needs to be a value higher than ED, and `100 * ED` accomplishes that just fine, as well as `100 +`, `99 +`, `42 *` or whatever other option would as long as there's no overflow :) As I mentioned above, we could also set the reward to 1 CFG; as long as it's > ED and the accounts used in the tests have sufficient balance, it all serves the purpose of covering what these tests need to cover.

pallets/loans/src/util.rs
NunoAlexandre: Ah awesome, thanks for the catch @lemunozm 🙏

lemunozm:
Updated with the latest main again. I'm not sure why, but fudge takes a looooot of time to run.

runtime/integration-tests/src/generic/utils/mod.rs
```diff
@@ -53,7 +52,12 @@ pub fn give_nft<T: Runtime>(dest: AccountId, (collection_id, item_id): (Collecti
 
 pub fn give_balance<T: Runtime>(dest: AccountId, amount: Balance) {
 	let data = pallet_balances::Account::<T>::get(dest.clone());
-	pallet_balances::Pallet::<T>::set_balance(&dest, data.free + amount);
+	pallet_balances::Pallet::<T>::force_set_balance(
+		RawOrigin::Root.into(),
+		T::Lookup::unlookup(dest),
+		data.free + amount,
+	)
+	.unwrap();
```
lemunozm, line 60: Probably this is a bit Nit from my side @wischli 

The old `set_balance()` has been mapped to `force_set_balance()`. I feel like if any state change in the runtime is done through "extrinsic" calls, it is easier to replicate later in development for the Apps team. They would only need to go through `state_mut` or `sumbit()` to know which extrinsic they should call to replicate the same actions

pallets/block-rewards/src/tests.rs
lemunozm: Yeah, I understand that any value over some limit will fulfill the test's requirement. My point is that setting the exact value you need to fulfill those requirements helps the reader to know exactly what's going on in the tests.

If I read `100 + ED`, I know somehow that `100` will be handled in the tests, and the `ED` is the part required to fulfill the existential deposit requirement.

Reading `100 * ED` seems like that amount will be distributed to 100 accounts or something. Now, `100` is no longer an amount but a multiplier of `ED`, which leads the reader to think this.

lemunozm: Anyways, just a Nit point 😄 . I'll revert this change because it was added along with other changes in the fixes PR I've pushed over this

NunoAlexandre: There was no need to revert the change, I just wanted to understand your point of view :) I honestly don't share the feeling you share but that's ok. I also think that keeping the value just right above the minimum it could be not only is not mapping the reality in production but it also makes the tests more brittle.

In any case, feel free to set this back to your `100 +`, specially if some other peers shares your feeling here 🙏

wischli: I personally favor the `+`.

NunoAlexandre: Fixed and discussion closed in https://github.com/centrifuge/centrifuge-chain/pull/1577/commits/1ba6cddd527a57e86f03ffe3954f13ef9ddacf63 where I've also added a comment. Thanks for your input 👍

runtime/centrifuge/src/weights/frame_system.rs
```diff
@@ -64,6 +64,11 @@ impl<T: frame_system::Config> frame_system::WeightInfo for WeightInfo<T> {
 			.saturating_add(T::DbWeight::get().reads(1))
 			.saturating_add(T::DbWeight::get().writes(2))
 	}
+
+	fn set_code() -> Weight {
+		todo!()
+	}
```
wischli, line 70: Unfortunately, the `frame_system` bench fix is [not included](https://github.com/paritytech/cumulus/pull/2766#issuecomment-1690303367) in Polkadot v0.9.43. It is included in v1.0.0. I propose to disable benchmarks for `frame_system` in the script until we upgrade to v1.0.0. The `set_code` call can only be called via root, so I don't think it imposes risks.

WDYT @NunoAlexandre?

NunoAlexandre: Agreed 1000% 👍

pallets/restricted-tokens/src/lib.rs
```diff
@@ -509,20 +530,24 @@ pub mod pallet {
...
+						&(),
 						&who,
 					);
 				<T::Fungibles as fungibles::MutateHold<T::AccountId>>::release(
 					currency_id,
+					&(),
 					&who,
 					old_reserved,
-					false,
+					Precision::BestEffort,
```
wischli, line 541: @NunoAlexandre Any particular reason why we go with `BestEffort` here whereas for the rest we have `Precision::Exact`? Asking because benchmarks are failing with `Account cannot be created for recording amount on hold`.

NunoAlexandre: @wischli it may have been a mistake in this case. Before we passed a `bool` where `true` meant `best_effort`, so if this was `false` before this should probably be `Exact`. Do the tests succeed with that option?

pallets/investments/src/lib.rs
```diff
@@ -591,19 +591,22 @@ impl<T: Config> Pallet<T> {
...
-						)?;
-
 						order.update_submitted_at(cur_order_id);
 
 						// Remove order from storage if empty
-						if amount == T::Amount::zero() {
+						if amount.is_zero() {
 							*maybe_order = None;
+						} else {
+							// nuno: check that this is ok. Amount == 0 removes the order, so
```
cdamian, line 600: Just a note here - `cancel_redeem_order` failed with this change. I reverted it and it worked.

Going to run the other tests now to see if they are affected.

wischli: > Just a note here - `cancel_redeem_order` failed with this change. I reverted it and it worked.

Had to revert it as well for a bunch of benchmarks in registry and psystem.

runtime/altair/src/evm.rs
```diff
@@ -43,7 +47,13 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for FindAuthorTruncated<F> {
...
 	pub BlockGasLimit: U256 = U256::from(NORMAL_DISPATCH_RATIO * MAXIMUM_BLOCK_WEIGHT.ref_time() / WEIGHT_PER_GAS);
 	pub PrecompilesValue: Altair<Runtime> = Altair::<_>::new();
-	pub WeightPerGas: Weight = Weight::from_ref_time(WEIGHT_PER_GAS);
+	pub WeightPerGas: Weight = Weight::from_parts(WEIGHT_PER_GAS, 0);
+	pub GasLimitPovSizeRatio: u64 = {
+		let block_gas_limit = BlockGasLimit::get().min(u64::MAX.into()).low_u64();
+		block_gas_limit.saturating_div(MAX_POV_SIZE)
+	};
+	pub GasLimitStorageGrowthRatio: u64 =
+		BlockGasLimit::get().min(u64::MAX.into()).low_u64().saturating_div(BLOCK_STORAGE_LIMIT);
```
mustermeiszer, line 56: **CLARIFICATION**
@NunoAlexandre could you give me some details on where these values come from?

runtime/altair/src/evm.rs
```diff
@@ -55,13 +65,17 @@ impl pallet_evm::Config for Runtime {
...
+	type GasLimitStorageGrowthRatio = GasLimitStorageGrowthRatio;
 	type GasWeightMapping = pallet_evm::FixedGasWeightMapping<Self>;
 	type OnChargeTransaction = ();
 	type OnCreate = ();
 	type PrecompilesType = Altair<Self>;
 	type PrecompilesValue = PrecompilesValue;
 	type Runner = pallet_evm::runner::stack::Runner<Self>;
 	type RuntimeEvent = RuntimeEvent;
+	type Timestamp = crate::Timestamp;
+	type WeightInfo = ();
```
mustermeiszer, line 78: **CR**

Have we tried benching `pallet-evm`?

runtime/altair/src/evm.rs
```diff
@@ -80,7 +94,14 @@ impl pallet_base_fee::Config for Runtime {
 	type Threshold = BaseFeeThreshold;
 }
 
+parameter_types! {
+	pub const PostBlockAndTxnHashes: PostLogContent = PostLogContent::BlockAndTxnHashes;
+	pub const ExtraDataLength: u32 = 30;
```
mustermeiszer, line 99: **CLARIFICATION**

@NunoAlexandre where does this value come from?

runtime/altair/src/lib.rs
```diff
@@ -329,6 +325,10 @@ impl pallet_balances::Config for Runtime {
 	type DustRemoval = ();
 	/// The minimum amount required to keep an account open.
 	type ExistentialDeposit = ExistentialDeposit;
+	type FreezeIdentifier = ();
+	type HoldIdentifier = ();
+	type MaxFreezes = ();
```
mustermeiszer, line 330: **MAYBE CR**

What does this mean? Wording makes it sound like it should be an integer...

runtime/altair/src/lib.rs
```diff
@@ -652,16 +652,19 @@ parameter_types! {
 	pub CouncilMotionDuration: BlockNumber = prod_or_fast!(5 * DAYS, 1 * MINUTES, "AIR_MOTION_DURATION");
 	pub const CouncilMaxProposals: u32 = 100;
 	pub const CouncilMaxMembers: u32 = 100;
+	pub MaxProposalWeight: Weight = Perbill::from_percent(50) * RuntimeBlockWeights::get().max_block;
```
mustermeiszer, line 655: **CLARIFICATION**

Does this also hold for runtime-upgrade proposals?

runtime/altair/src/lib.rs
```diff
@@ -1324,8 +1332,13 @@ impl pallet_membership::Config for Runtime {
 	type WeightInfo = pallet_membership::weights::SubstrateWeight<Self>;
 }
 
+parameter_types! {
+	pub const MaxFeedValues: u32 = 10;
```
mustermeiszer, line 1336: @lemunozm this is sufficient for our needs?

runtime/altair/src/lib.rs
```diff
@@ -1965,6 +1977,9 @@ impl_runtime_apis! {
 		fn metadata() -> OpaqueMetadata {
 			OpaqueMetadata::new(Runtime::metadata().into())
 		}
+
+		fn metadata_at_version(_: u32) -> Option<sp_core::OpaqueMetadata> { todo!("nuno") }
+		fn metadata_versions() -> frame_benchmarking::Vec<u32> { todo!("nuno") }
```
mustermeiszer, line 1982: **CR**

We must have something here. This would otherwise allow to kill nodes by calling this api.

runtime/altair/src/lib.rs
```diff
@@ -2206,8 +2221,41 @@ impl_runtime_apis! {
...
+				match <Runtime as pallet_evm::Config>::GasWeightMapping::gas_to_weight(
+					gas_limit,
+					without_base_extrinsic_weight
+				) {
+					weight_limit if weight_limit.proof_size() > 0 => {
+						(Some(weight_limit), Some(estimated_transaction_len as u64))
+					}
+					_ => (None, None),
+				};
+
```
mustermeiszer, line 2258: **CLARIFICATION**
@NunoAlexandre where do these lines come from?

runtime/altair/src/xcm.rs
```diff
@@ -268,13 +269,16 @@ parameter_types! {
 /// Pallet Xcm offers a lot of out-of-the-box functionality and features to
 /// configure and handle XCM messages.
 impl pallet_xcm::Config for Runtime {
+	type AdminOrigin = EnsureRoot<AccountId>;
```
mustermeiszer, line 272: **MAYBE CR**

I would opt in here for `EnsureRootOr<AccountId, HalfOfCouncil>`

runtime/centrifuge/src/evm.rs
```diff
@@ -44,7 +48,13 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for FindAuthorTruncated<F> {
...
 	pub BlockGasLimit: U256 = U256::from(NORMAL_DISPATCH_RATIO * MAXIMUM_BLOCK_WEIGHT.ref_time() / WEIGHT_PER_GAS);
 	pub PrecompilesValue: CentrifugePrecompiles<crate::Runtime> = CentrifugePrecompiles::<_>::new();
-	pub WeightPerGas: Weight = Weight::from_ref_time(WEIGHT_PER_GAS);
+	pub WeightPerGas: Weight = Weight::from_parts(WEIGHT_PER_GAS, 0);
+	pub GasLimitPovSizeRatio: u64 = {
+		let block_gas_limit = BlockGasLimit::get().min(u64::MAX.into()).low_u64();
+		block_gas_limit.saturating_div(MAX_POV_SIZE)
+	};
+	pub GasLimitStorageGrowthRatio: u64 =
+		BlockGasLimit::get().min(u64::MAX.into()).low_u64().saturating_div(BLOCK_STORAGE_LIMIT);
```
mustermeiszer, line 57: **CLARIFICATION**

Same as in `altiar/lib.rs`

runtime/centrifuge/src/evm.rs
```diff
@@ -56,13 +66,17 @@ impl pallet_evm::Config for crate::Runtime {
...
+	type GasLimitStorageGrowthRatio = GasLimitStorageGrowthRatio;
 	type GasWeightMapping = pallet_evm::FixedGasWeightMapping<Self>;
 	type OnChargeTransaction = ();
 	type OnCreate = ();
 	type PrecompilesType = CentrifugePrecompiles<Self>;
 	type PrecompilesValue = PrecompilesValue;
 	type Runner = pallet_evm::runner::stack::Runner<Self>;
 	type RuntimeEvent = crate::RuntimeEvent;
+	type Timestamp = crate::Timestamp;
+	type WeightInfo = ();
```
mustermeiszer, line 79: **CR**

Same as in `altiar/lib.rs`

runtime/centrifuge/src/evm.rs
```diff
@@ -81,7 +95,14 @@ impl pallet_base_fee::Config for crate::Runtime {
 	type Threshold = BaseFeeThreshold;
 }
 
+parameter_types! {
+	pub const PostBlockAndTxnHashes: PostLogContent = PostLogContent::BlockAndTxnHashes;
+	pub const ExtraDataLength: u32 = 30;
```
mustermeiszer, line 100: **CLARIFICATION**

Same as in `altiar/lib.rs`

runtime/centrifuge/src/lib.rs
```diff
@@ -472,6 +469,10 @@ impl pallet_balances::Config for Runtime {
 	type DustRemoval = ();
 	/// The minimum amount required to keep an account open.
 	type ExistentialDeposit = ExistentialDeposit;
+	type FreezeIdentifier = ();
+	type HoldIdentifier = ();
+	type MaxFreezes = ();
```
mustermeiszer, line 474: **CLARIFICATION**

Same as in `altiar/lib.rs`

runtime/centrifuge/src/lib.rs
```diff
@@ -1673,8 +1682,13 @@ impl pallet_membership::Config for Runtime {
 	type WeightInfo = pallet_membership::weights::SubstrateWeight<Self>;
 }
 
+parameter_types! {
+	pub const MaxFeedValues: u32 = 10;
```
mustermeiszer, line 1686: **CLARIFICATION**

Same as in `altiar/lib.rs`

runtime/centrifuge/src/lib.rs
```diff
@@ -2002,6 +2014,9 @@ impl_runtime_apis! {
 		fn metadata() -> OpaqueMetadata {
 			OpaqueMetadata::new(Runtime::metadata().into())
 		}
+
+		fn metadata_at_version(_: u32) -> Option<sp_core::OpaqueMetadata> { todo!("nuno") }
+		fn metadata_versions() -> frame_benchmarking::Vec<u32> { todo!("nuno") }
```
mustermeiszer, line 2019: **CR**

Same as in `altiar/lib.rs`

runtime/centrifuge/src/xcm.rs
```diff
@@ -270,13 +271,16 @@ parameter_types! {
 /// Pallet Xcm offers a lot of out-of-the-box functionality and features to
 /// configure and handle XCM messages.
 impl pallet_xcm::Config for Runtime {
+	type AdminOrigin = EnsureRoot<AccountId>;
```
mustermeiszer, line 274: **MAYBE CR**

Same as in `altiar/lib.rs`

runtime/development/src/evm.rs
```diff
@@ -44,7 +47,13 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for FindAuthorTruncated<F> {
...
 	pub BlockGasLimit: U256 = U256::from(NORMAL_DISPATCH_RATIO * MAXIMUM_BLOCK_WEIGHT.ref_time() / WEIGHT_PER_GAS);
 	pub PrecompilesValue: Development<Runtime> = Development::<_>::new();
-	pub WeightPerGas: Weight = Weight::from_ref_time(WEIGHT_PER_GAS);
+	pub WeightPerGas: Weight = Weight::from_parts(WEIGHT_PER_GAS, 0);
+	pub GasLimitPovSizeRatio: u64 = {
+		let block_gas_limit = BlockGasLimit::get().min(u64::MAX.into()).low_u64();
+		block_gas_limit.saturating_div(MAX_POV_SIZE)
+	};
+	pub GasLimitStorageGrowthRatio: u64 =
+		BlockGasLimit::get().min(u64::MAX.into()).low_u64().saturating_div(BLOCK_STORAGE_LIMIT);
```
mustermeiszer, line 56: **CLARIFICATION**

Same as in `altiar/lib.rs`

runtime/development/src/evm.rs
```diff
@@ -81,7 +94,14 @@ impl pallet_base_fee::Config for Runtime {
 	type Threshold = BaseFeeThreshold;
 }
 
+parameter_types! {
+	pub const PostBlockAndTxnHashes: PostLogContent = PostLogContent::BlockAndTxnHashes;
+	pub const ExtraDataLength: u32 = 30;
```
mustermeiszer, line 99: **CLARIFICATION**

Same as in `altiar/lib.rs`

runtime/development/src/lib.rs
```diff
@@ -338,6 +336,10 @@ impl pallet_balances::Config for Runtime {
 	type DustRemoval = ();
 	/// The minimum amount required to keep an account open.
 	type ExistentialDeposit = ExistentialDeposit;
+	type FreezeIdentifier = ();
+	type HoldIdentifier = ();
+	type MaxFreezes = ();
```
mustermeiszer, line 341: **CLARIFICATION**

Same as in `altiar/lib.rs`

runtime/development/src/lib.rs
```diff
@@ -1307,8 +1315,13 @@ impl pallet_membership::Config for Runtime {
 	type WeightInfo = pallet_membership::weights::SubstrateWeight<Self>;
 }
 
+parameter_types! {
+	pub const MaxFeedValues: u32 = 10;
```
mustermeiszer, line 1319: **CLARIFICATION**

Same as in `altiar/lib.rs`

runtime/development/src/evm.rs
```diff
@@ -56,13 +65,17 @@ impl pallet_evm::Config for Runtime {
...
+	type GasLimitStorageGrowthRatio = GasLimitStorageGrowthRatio;
 	type GasWeightMapping = pallet_evm::FixedGasWeightMapping<Self>;
 	type OnChargeTransaction = ();
 	type OnCreate = ();
 	type PrecompilesType = Development<Self>;
 	type PrecompilesValue = PrecompilesValue;
 	type Runner = pallet_evm::runner::stack::Runner<Self>;
 	type RuntimeEvent = RuntimeEvent;
+	type Timestamp = crate::Timestamp;
+	type WeightInfo = ();
```
mustermeiszer, line 78: **CR**

Same as in `altiar/lib.rs`

runtime/development/src/lib.rs
```diff
@@ -2089,6 +2104,9 @@ impl_runtime_apis! {
 		fn metadata() -> OpaqueMetadata {
 			OpaqueMetadata::new(Runtime::metadata().into())
 		}
+
+		fn metadata_at_version(_: u32) -> Option<sp_core::OpaqueMetadata> { todo!("nuno") }
+		fn metadata_versions() -> frame_benchmarking::Vec<u32> { todo!("nuno") }
```
mustermeiszer, line 2109: **CR**

Same as in `altiar/lib.rs`

runtime/development/src/xcm.rs
```diff
@@ -164,7 +165,8 @@ impl TakeRevenue for ToTreasury {
 /// executed.
 pub type Barrier = (
 	TakeWeightCredit,
-	xcm_primitives::AllowTopLevelPaidExecutionDescendOriginFirst<Everything>,
+	//todo(nuno): revisit this
+	// xcm_primitives::AllowTopLevelPaidExecutionDescendOriginFirst<Everything>,
```
mustermeiszer, line 169: **CLARIFICATION**

I have not seen that in change in the other runtimes. What is the difference between the two?

scripts/init.sh
```diff
@@ -51,20 +52,19 @@ start-parachain)
...
     --wasm-execution=compiled \
     --execution=wasm \
     --port $((30355 + $para_id)) \
     --rpc-port $((9936 + $para_id)) \
-    --ws-port $((9946 + $para_id)) \
     --rpc-external \
     --rpc-cors all \
-    --ws-external \
     --rpc-methods=Unsafe \
-    --state-cache-size 0 \
```
mustermeiszer, line 66: **NIT**

We might wanna consider a `trie-cache size` value, but not sure if needed.

src/command.rs
```diff
@@ -436,6 +443,11 @@ pub fn run() -> Result<()> {
 			let collator_options = cli.run.collator_options();
 
 			runner.run_node_until_exit(|config| async move {
+				let hwbench = (!cli.no_hardware_benchmarks).then_some(
+					config.database.path().map(|database_path| {
+						let _ = std::fs::create_dir_all(database_path);
```
mustermeiszer, line 448: **MAYBE CR**

Usually the db takes care of creating the folders on the path to the db. 
* Why do we need to do that here too?
*  Isn't the client already instantiated at this point? 
* We do not care for the result, as?

mustermeiszer: Nevermind. Client is not there yet. See why we need to create that path here. What a hack... Is this copied from parity?

src/command.rs
```diff
@@ -531,14 +546,10 @@ impl DefaultConfigurationValues for RelayChainCli {
 		30334
 	}
 
-	fn rpc_ws_listen_port() -> u16 {
+	fn rpc_listen_port() -> u16 {
```
mustermeiszer, line 549: **MAYBE CR**

Shouldn't we get those values from the config? Ahh not possible... okay...

src/service.rs
```diff
@@ -36,21 +36,13 @@ use crate::rpc::{
...
-
-#[cfg(feature = "runtime-benchmarks")]
-type HostFunctions = (
-	sp_io::SubstrateHostFunctions,
-	frame_benchmarking::benchmarking::HostFunctions,
-);
-
-type FullClient<RuntimeApi> = TFullClient<Block, RuntimeApi, WasmExecutor<HostFunctions>>;
+type FullClient<RuntimeApi, Executor> =
+	TFullClient<Block, RuntimeApi, NativeElseWasmExecutor<Executor>>;
```
mustermeiszer, line 40: **CR**

Why the change to `NativeElseWasmExecutor`?

src/service/evm.rs
```diff
@@ -465,14 +474,16 @@ where
...
+		keystore: params.keystore_container.keystore(),
 		backend: backend.clone(),
 		network: network.clone(),
 		system_rpc_tx,
 		tx_handler_controller,
+		sync_service: sync_service.clone(),
 		telemetry: telemetry.as_mut(),
 	})?;
 
+	// do we need this at all?
```
mustermeiszer, line 486: **CLARIFICATION**

@wischli What makes think we do not need them?

mustermeiszer (commented):
Generally: I would like to get rid of any `moonbeam-foundation/{frontier, substrate, polkadot}` dependency. Do you think this is possible? 

NOTE: I will do that in the `polkadot-v.1.1.0` upgrade.

At everybody involved in this upgrade: Thanks and kudos! That was a nightmare...

mustermeiszer:
NOTE: I did not yet review integration tests at this point as we will bring in #1590 which will change a lot there.

pallets/investments/src/lib.rs
```diff
@@ -591,6 +591,8 @@ impl<T: Config> Pallet<T> {
 							Error::<T>::CollectRequired
 						);
 
+						order.update_submitted_at(cur_order_id);
+
```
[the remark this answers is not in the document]
mustermeiszer: Removed. This is not doing anything useful. The generation of the default order takes already care to make the submitted equal to the current order id. If it is not equal, then we early error out anyayws.

pallets/restricted-tokens/src/impl_fungible.rs
```diff
@@ -165,57 +250,100 @@ pub enum FungibleMutateHoldEffects<AccountId, Balance> {
...
 	) -> Result<Self::Balance, DispatchError> {
 		ensure!(
 			T::PreFungibleMutateHold::check(FungibleMutateHoldEffects::TransferHeld(
 				source.clone(),
 				dest.clone(),
 				amount,
-				best_effort,
-				on_held
+				precision == Precision::BestEffort,
+				mode == Restriction::OnHold,
```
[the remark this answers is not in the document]
mustermeiszer: Keeping

runtime/altair/src/evm.rs
mustermeiszer: Leaving this for a post mortem review. But will keep it for now.

runtime/altair/src/lib.rs
mustermeiszer: @wischli leaving this for a post mortem review

src/service.rs
mustermeiszer: Keeping it for now as I am tired of changing these things back. @wischli would appreciate a post mortem clarification.

runtime/altair/src/lib.rs
mustermeiszer: Made it consistent with Polkadot/Moonbeam.

runtime/altair/src/evm.rs
mustermeiszer: Not possible right now

runtime/altair/src/lib.rs
mustermeiszer: Done

runtime/altair/src/xcm.rs
mustermeiszer: Democracy okay. Leaving that.

runtime/centrifuge/src/lib.rs
mustermeiszer: Changed

runtime/development/src/evm.rs
mustermeiszer: Done

runtime/development/src/lib.rs
mustermeiszer: Changed.

runtime/development/src/lib.rs
mustermeiszer: Done

runtime/development/src/xcm.rs
mustermeiszer: leaving for later

mustermeiszer:
benchmark altair

src/service.rs
wischli: Quick reply before Monday: This is taken from the cumulus parachain client.

docker-compose-local-relay.yml
```diff
@@ -1,15 +1,14 @@
...
 # Node Key Bob: `subkey inspect //Bob`
-version: '3'
+version: '3.4'
 services:
   node_alice:
     container_name: alice
-    image: "parity/polkadot:v0.9.38"
-    platform: "linux/amd64"
+    image: "parity/polkadot:v1.0.0"
+    platform: "linux/x86_64"
```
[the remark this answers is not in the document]
wischli: I think that unfolds to the same, doesnt it? But good catch!

pallets/restricted-tokens/src/impl_fungible.rs
```diff
@@ -56,20 +63,47 @@ impl<T: Config> Inspect<T::AccountId> for Pallet<T> {
...
+		#[cfg(feature = "std")]
+		{
+			dbg!(
+				<T::NativeFungible as Inspect<T::AccountId>>::reducible_balance(
+					who,
+					preservation,
+					force,
+				)
+			);
+		}
```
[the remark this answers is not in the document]
wischli: Yes, should be removed. Artifact from my debug session.

runtime/altair/src/lib.rs
wischli: Will check on Monday!

scripts/init.sh
wischli: The trie cache size is set by default. Should be fine for now.

src/service/evm.rs
wischli: I think this comment belonged to codepath which I eventually removed. Will remove the comment on Monday.

mustermeiszer (approved):
Approving from my side. I will most likely force merge once green and then go ahead to upgrade to polkadot-v1.1.0 using the polkadot-sdk monorepo. 

Note: We should not cut a release for now.

runtime/altair/src/evm.rs
mustermeiszer: From moonbeam. I looked it up and will keep it as is.

runtime/altair/src/lib.rs
mustermeiszer: This is sufficient. It is an upper boundary on the weight of an proposal. As runtime ugprades usually have the `external_propose_majority` extrinsics as the thing voted on in the proposal the weight should be sufficient.

runtime/altair/src/lib.rs
mustermeiszer: Moonbeam.

src/service.rs
mustermeiszer: Fine for me.

PR #1578, opened 2023-10-02
PoC: Integration tests with generic fudge environment

lemunozm:
# Description

Proof of concept to convert the `env` fudge module of `integration-test` in a `Runtime` independent way. So after this, test cases will be based on `<T: Config>` where `T` is the current runtime.


runtime/integration-tests/src/experimental/env.rs
```diff
@@ -0,0 +1,317 @@
...
+	generic::BlockId,
+	traits::{BlakeTwo256, Block, Extrinsic},
+	AccountId32, DigestItem, Storage,
+};
+use sp_transaction_pool::runtime_api::TaggedTransactionQueue;
+use tokio::runtime::Handle;
+
+use crate::utils::{logs, time::START_DATE};
+
+pub trait Config:
```
lemunozm, line 43: Common configuration between all our runtimes

runtime/integration-tests/src/experimental/env.rs
```diff
@@ -0,0 +1,317 @@
...
+type RelayDp = Box<dyn DigestCreator<RelayBlock> + Send + Sync>;
+
+pub enum EventRange {
+	All,
+	One(BlockNumber),
+	Range(BlockNumber, BlockNumber),
+	Latest,
+}
+
+pub trait Environment {
```
lemunozm, line 99: The environment wraps different `fudge::companion` structs to be used independently of the runtime

runtime/integration-tests/src/experimental/runtimes.rs
```diff
@@ -0,0 +1,82 @@
...
+	// TODO...
+	// implement non-default methods of Environment trait
+}
+
+impl Config for altair_runtime::Runtime {
+	//type RelayApi = kusama_runtime::RuntimeApi;
+	//type RelayRuntime = kusama_runtime::Runtime;
+
+	const KIND: RuntimeKind = RuntimeKind::Altair;
+}
```
lemunozm, line 53: Each runtime should implement the `Environment` and the `Config`

lemunozm:
@mustermeiszer maybe the ideas from this PoC can be used to build the `Option A` of your thoughts [here](https://kflabs.slack.com/archives/C04GJQAM9P0/p1696233663801349)

mustermeiszer:
I think my main problem is that `test_with_all_runtimes` assumes that the setup and so on is the same for all runtimes. Which I think is mostly true, but one essential breaking point are the differing `AdminOrigin` that we are using. Although, injecting via root should still work.

And tests would be generic over `T: Config` then?

lemunozm:
Yes, the test would be generic over `T: Config`. I think procedural macros could be pretty obscure, and they are difficult to update if you do not have the required knowledge (actually, only you have such a power!). For me, as a rule of thumb, if you can avoid macros (decl or proc), then avoid them.

> I think my main problem is that test_with_all_runtimes assumes that the setup and so on is the same for all runtimes.

Regarding this. The setup is the same only for those types you explicitly instantiate in `pallet_xxx::Config`. I mean, if I put in the `Config` trait, a bound with `pallet_balances::Config<Balance = u128>`, then I assume all runtimes use `u128`, but if I leave it without the restriction, so just `pallet_balances::Config`, then each runtime could have their own `Balance` types. So, the idea is only restricted when you know all of our runtimes use the same types, if not, the types are not specified.

For things like `AdminOrigin`, where maybe the call in the test changes because of that, you can manually dispatch it with an `if/match` using `T::KIND`, for such border cases.

lemunozm:
But I think, the big point of this PR (and challenge), is that the `fudge::companion` structs and associated methods and macros, can be abstracted by an `Environtment` trait. Also if they're abstracted, I think we no longer need macros as `env::run` or similar, they could be implemented in actual rust code.

lemunozm:
Closed in favor of: https://github.com/centrifuge/centrifuge-chain/pull/1588

PR #1583, opened 2023-10-06
Integration test: Generic runtime and environment

lemunozm:
# Description

This PR adds a testing framework with the following features:
- Tests are now parameterized by `T: Runtime` where `T` contains all expected types from a runtime as Config, Block, Events, Calls, Apis... (check `env.rs` file).
- Tests are generic over an environment. The `Env` API abstracts between:
   - fudge (pending work for a future PR)
   - a more simple/low-level environment close to the runtime itself
- No more macros in the test cases (easier to read/debug)

The framework and tests cases are under a `generic` folder. Check the [example](runtime/integration-tests/src/generic/cases/example.rs) to get an introduction of this.

### Pending extra changes for this PR:
- [x] Fix the fees calculation method.
- [x] Macro to run tests parameterized with the runtimes to test.


runtime/integration-tests/src/generic/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+
+			let dispatch_info = match runtime_event.try_into() {
+				Ok(frame_system::Event::<T>::ExtrinsicSuccess { dispatch_info }) => dispatch_info,
+				_ => panic!("expected to be called after a successful extrinsic"),
+			};
+
+			match dispatch_info.pays_fee {
+				Pays::Yes => WeightToFee::weight_to_fee(&dispatch_info.weight),
+				Pays::No => 0,
+			}
```
lemunozm, line 106: This is not exactly the fees the last executed extrinsic has consumed. Not sure why or what I'm missing.

runtime/integration-tests/src/generic/cases/example.rs
```diff
@@ -0,0 +1,86 @@
...
+	env.state(|| {
+		// Call to Core::version() API.
+		// It's automatically implemented by the runtime T, so you can easily do:
+		// T::version()
+		assert_eq!(T::version(), <T as frame_system::Config>::Version::get());
+	})
+}
+
+// Generate tests for all runtimes
+crate::test_with_all_runtimes!(transfer_balance);
```
lemunozm, line 76: I wanted to modify this macro to:
```rust
crate::test_for!((development, altair, centrifuge), transfer_balance);
```
where the user can specify which runtimes they want to run

lemunozm (commented):
Pending things:

runtime/development/src/lib.rs
```diff
@@ -922,7 +922,7 @@ impl pallet_treasury::Config for Runtime {
 	type Burn = Burn;
 	// we burn and dont handle the unbalance
 	type BurnDestination = ();
-	type Currency = Tokens;
```
lemunozm, line 925: BTW, It's not a breaking change, and we align all runtimes with this.

runtime/integration-tests/src/generic/cases/example.rs
lemunozm: Solved

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -0,0 +1,222 @@
...
+		let digest = Digest {
+			logs: vec![DigestItem::PreRuntime(AURA_ENGINE_ID, slot.encode())],
+		};
+
+		let header = Header {
+			number: i,
+			digest,
+			state_root: H256::default(),
+			extrinsics_root: H256::default(),
+			parent_hash: [69u8; 32].into(),
```
wischli, line 162: I suppose the `69` is an arbitrary number. Maybe we want to take default instead to reduce unexpected behaviour likelihood?

runtime/integration-tests/src/generic/runtime.rs
```diff
@@ -0,0 +1,151 @@
...
+	+ pallet_timestamp::Config<Moment = Moment>
+	+ pallet_aura::Config<Moment = Moment, AuthorityId = AuraId>
+	+ pallet_authorship::Config
+	+ pallet_treasury::Config<Currency = pallet_balances::Pallet<Self>>
+	+ pallet_transaction_payment::Config<
+		WeightToFee = WeightToFee,
+		OnChargeTransaction = CurrencyAdapter<pallet_balances::Pallet<Self>, DealWithFees<Self>>,
+	> + pallet_restricted_tokens::Config<
+		Balance = Balance,
+		NativeFungible = pallet_balances::Pallet<Self>,
```
wischli, line 91: IIUC, we should add all pallet's here which we want to use in integration tests. I wonder if the compiler would not throw if runtimes diverge in the Configs they implement or would this be solved by [your efforts to pick the runtime(s) on the test granularity](https://github.com/centrifuge/centrifuge-chain/pull/1583/files#r1348506233)?

wischli (commented):
Some comments. Really looking forward to getting hands on with this new approach! Thanks for your great work, as always. Super clean and a pleasure to read.

runtime/integration-tests/src/generic/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+
+	/// Pass a number of blocks proportional to these seconds
+	BySeconds(Moment),
+
+	/// Pass a number of block until find an event or reach the limit
+	UntilEvent {
+		event: T::RuntimeEventExt,
+		limit: BlockNumber,
+	},
+}
```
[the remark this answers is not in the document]
lemunozm: Good point!

It rounds up. But now you've left me thinking if we really want that 🤔. We must definitely comment out this

runtime/integration-tests/src/generic/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+
+/// Define an environment behavior
+pub trait Env<T: Runtime> {
+	/// Loan the environment from a genesis
```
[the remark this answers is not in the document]
lemunozm: To much time with with loans 😆

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: Yeah, I think it is an arbitrary number. It's the number used to configure the `frame_system` when genesis is the default. I need to check if I can actually put whatever there, or if not, I should link to the number in `frame_system`. 

Good point!

runtime/integration-tests/src/generic/runtime.rs
lemunozm: > IIUC, we should add all pallet's here which we want to use in integration tests

Yes, if you want to use *explicitly* a pallet, then you should add it there.

> I wonder if the compiler would not throw if runtimes diverge in the Configs they implement

Configs could diverge, but it's not an issue until you don't use something explicitly. So, the idea is to keep this `Config` as minimal and generic as possible. 

Because all of our runtimes are expected to be the "same" application, most architectural changes should be the same (pallets depending on the same pallets and things like that), which are the most important, and only minor config properties differ, which call be accessed using T::Property in most cases.

These tests also help to determine structural incoherences among our runtimes, so I think it's another good point of having them. This is how I found the line you commented on. Only properties that use other pallets are the ones that, sooner or later, should be mostly aligned.

In the worst case, where you need explicitly something different for a runtime, you can add a bound in your test:

```rust
test_foo<T: Runtime>() where T: ... {

}
```

But if you need this, probably you can test directly with the original runtime because such runtime behaves differently, and you expect something different.

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: Confirmed, it was a magic number. Nothing to do with the default from `frame_system`. Thanks for comment this!

runtime/development/src/lib.rs
mustermeiszer: Why? We should use Tokens everywhere in order to keep restrictions on place.

lemunozm: Then we should change instead from `Balance` to `Tokens` in `altair` and `centrifuge`

runtime/integration-tests/src/generic/environment.rs
lemunozm: They're different, not sure why, but the way of calculating them comes from different paths. The difference is really subtle anyway:
- From `TransactionFeePaid`: 
    - 2004154679487179
- From `ExtrinsicSuccess`: 
    - 2003074679487179
    
Still investigating, thanks for addressing me to the `TransactionFeePaid` event!

runtime/integration-tests/src/generic/envs/runtime_env.rs
NunoAlexandre: > it was a magic number


No pun intended?

runtime/development/src/lib.rs
lemunozm: Done

runtime/integration-tests/src/generic/environment.rs
lemunozm: Not sure why the difference. But as far as I am concerned, using the value from `TransactionFeePaid` works so... I'll give this as solved.

runtime/integration-tests/src/generic/envs/fudge_env.rs
```diff
@@ -0,0 +1 @@
+// TODO: implement generic::env::Env for an environment using fudge
```
NunoAlexandre, line 1: Do we still want to do this in this PR?

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: A magic pun then 😆

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -0,0 +1,200 @@
...
+
+			T::apply_extrinsic(extrinsic).unwrap()
+		})
+	}
+
+	fn pass(&mut self, blocks: Blocks<T>) {
+		self.ext.borrow_mut().execute_with(|| {
+			let next = frame_system::Pallet::<T>::block_number() + 1;
+
+			let last_block = match blocks {
```
NunoAlexandre, line 82: This name can be a bit confusing since it's not getting the last block but the block at which we want to stop at. Within context it makes sense but I had to think that twice. Just a thought 👍

NunoAlexandre (commented):
Overall this looks fantastic 👏  awesome work, Luis.

I am wondering how we plan on handling certain behaviour deviations between runtimes; for example, DOT and KSM have different decimals so when testing XCM transfers involving such a case, we need to take those differences in consideration. Or we might have certain restrictions on `centrifuge` that we don't have on `development` which would make it hard to build generic tests over the two.

Either way, this is a beautiful step further ❤️

runtime/integration-tests/src/generic/environment.rs
NunoAlexandre: Maybe there's some extra fees for dispatching the transaction? That happens with xcm-related transactions, maybe there's a similar mechanism here involved?

runtime/integration-tests/src/generic/envs/fudge_env.rs
lemunozm: No, this will be a future PR not to block this. That file is just a placeholder.

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: Maybe `block_to_stop` is more readable?

lemunozm:
> I am wondering how we plan on handling certain behaviour deviations between runtimes; for example, DOT and KSM have different decimals so when testing XCM transfers involving such a case, we need to take those differences in consideration. Or we might have certain restrictions on centrifuge that we don't have on development which would make it hard to build generic tests over the two.

For such things, the `Runtime` trait has a `T::KIND` where you can match and distinguish your behavior. i.e:
```rust
match T::KIND {
   KindRuntime::Altair => //do X,
   KindRuntime::Centrifuge => //do Y
}
```

But I think most of the time differentiations as decimals or other properties can be used under some `T::property::get()` without looking into what it really is.

lemunozm:
Thanks for all your comments and feedback!

I think the PR is ready. The `Cargo.toml` from `integration-tests` can be simplify a lot (removing some features, runtime-benchmarks block which I think it's not needed, `default-features = false` parts, etc). But I think it's better to do this once @NunoAlexandre finalizes with the upgrade to avoid creating many conflicts.

runtime/integration-tests/src/generic/envs/runtime_env.rs
NunoAlexandre: yeah or `end_block`. something like that 👍

NunoAlexandre:
> Thanks for all your comments and feedback!
> 
> I think the PR is ready. The `Cargo.toml` from `integration-tests` can be simplify a lot (removing some features, runtime-benchmarks block which I think it's not needed, `default-features = false` parts, etc). But I think it's better to do this once @NunoAlexandre finalizes with the upgrade to avoid creating many conflicts.

I think it's better if you move ahead, it's no problem for me to deal with the conflicts. The world needs this PR merged asap 😄

runtime/integration-tests/src/generic/envs/fudge_env.rs
lemunozm: I should create an issue for this

lemunozm: https://github.com/centrifuge/centrifuge-chain/issues/1586

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: I like it more!

lemunozm:
There would be a lot of line changes. I think we can merge it without touching `Cargo.toml`, and in another PR after the upgrade, simplify the dependencies. WDYT?

runtime/integration-tests/src/generic/environment.rs
lemunozm: Yeah, there must be some extra fees that do not depend on the weight of the call, so when I do `WeightToFee::weight_to_fee(&dispatch_info.weight)` I'm leaving some fees behind.

lemunozm:
Ready for a final review/merge

lemunozm:
I consider this PR done, although some minor parts will be modified in a next PR #1588 to be able to support a `FudgeEnv`

wischli (approved):
Looks great! We should streamline the setup process by adding setter functions for common tasks such as setting the authorities, see comment below. Not a blocker for sure.

runtime/integration-tests/src/generic/cases/example.rs
```diff
@@ -0,0 +1,118 @@
...
+		assert_eq!(T::version(), <T as frame_system::Config>::Version::get());
+	})
+}
+
+fn check_fee<T: Runtime>() {
+	let mut env = RuntimeEnv::<T>::from_storage(
+		Genesis::default()
+			.add(pallet_aura::GenesisConfig::<T> {
+				authorities: vec![AuraId::from(Keyring::Charlie.public())],
+			})
```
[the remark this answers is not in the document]
lemunozm: Yeah! The idea is to add this kind of thing. By now, in https://github.com/centrifuge/centrifuge-chain/pull/1588 such genesis are moved to the environment because it's mandatory needed to make aura works.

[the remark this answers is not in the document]
lemunozm: It still need a lot of utilities for common stuff as creating assets, etc

runtime/integration-tests/src/generic/environment.rs
```diff
@@ -0,0 +1,85 @@
...
+	/// Retrieve the fees used in the last submit call
+	fn last_fee(&self) -> Balance {
+		self.find_event(|e| match e {
+			pallet_transaction_payment::Event::TransactionFeePaid { actual_fee, .. } => {
+				Some(actual_fee)
+			}
+			_ => None,
+		})
+		.expect("Expected transaction")
+	}
```
mustermeiszer, line 84: This seems pretty detailed for the trait itself...

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -0,0 +1,200 @@
...
+		let mut inherent_data = InherentData::default();
+
+		let sproof_builder = RelayStateSproofBuilder::default();
+		let (relay_parent_storage_root, relay_chain_state) =
+			sproof_builder.into_state_root_and_proof();
+
+		let cumulus_inherent = ParachainInherentData {
+			validation_data: PersistedValidationData {
+				parent_head: vec![].into(),
+				relay_parent_number: i,
```
mustermeiszer, line 166: Nit: I think this might become problemativ once we rely on the relay chain block number in some code locations. But not the case right now. But we will probably switch to the relay-chain as a our clock sooner than later.

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -0,0 +1,200 @@
...
+				max_pov_size: Default::default(),
+				relay_parent_storage_root,
+			},
+			relay_chain_state,
+			downward_messages: Default::default(),
+			horizontal_messages: Default::default(),
```
mustermeiszer, line 172: We could inject XCM messages here. That would be great.

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: >  once we rely on the relay chain block number in some code locations.

Makes sense, taken into account!

runtime/integration-tests/src/generic/environment.rs
lemunozm: Yes, it is, and I've removed it in https://github.com/centrifuge/centrifuge-chain/pull/1588

Now, `submit_now()` returns the fee it has cost. In order to later check in tests the exact balance required for some actions

Thanks for your post-morten review!

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: @cdamian

cdamian: This seems like a standalone env - no relay nor other parachains being involved. So, even if we populate the `horizontal_messages`, there's no receiving end for it afaict.

lemunozm: Could we mock the receiving end somehow?

cdamian: The receiving end would have to be an actual parachain that's onboarded etc. Not sure if that's something that we want to add at this point.

I think that for the purpose of most integration tests, this setup is fine and we can leave it as it is. For other tests, such as XCM ones, we might need to use a fudge companion with multiple paras. So, I wouldn't mingle with this, for now.

lemunozm: Ok Cool!

BTW, I think for the `FudgeEnv` is quite easy to add new parachains, just adding it in the `fudge::companion` struct. Something to think about in the future, but I think it's feasible.

PR #1580, opened 2023-10-04
Fix/precompile account codes

mustermeiszer:
# Description
* We need to set `pallet_evm::AccountCodes` storage for ALL precompiles.

The problem is the following. During making a call to another contract the asm code contains a `EXTCODESIZE` instruction. This fetches the given code for an address. Precompiles have no code and hence we get a `Revert`. Moonbeam has also solved it by setting the `AccountCodes` storage for all precompiles.
They set it to `0x600600fd` which I think are instructions to push two zero bytes into the stack and then revert. 

## Changes and Descriptions
* add migration to push `0x600600fd` to all precompile `AccountCodes` storage
* bumb runtime version Centrifuge to 2022
* add migration to `Altair`, `Centrifuge`, `Development`
* add account codes as default for precompiles to chain-spec

# Checklist:

- [x] I have added Rust doc comments to structs, enums, traits and functions
- [x] I have made corresponding changes to the documentation
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or that my feature works

PR #1584, opened 2023-10-06
fix: remove spec_version check for Centrifuge 1023 migrations

wischli:
# Description

Removes `spec_version` checks for Anemoy, USDC and EVM Precompile Accountcode migrations. 

Apparently, we cannot rely on the result of `try-runtime` which [succeeded for all these migrations](https://github.com/centrifuge/centrifuge-chain/pull/1580#issuecomment-1747015476). However, in practice neither Dev/Demo nor Catalyst ran the EVM Accountcode migrations (Anemoy & USDC migrations ran on Catalyst in a previous upgrade from 1020 to 1021). My understanding is that at the time of executing `on_runtime_upgrade()`, the spec_version has been bumped already and `LastRuntimeUpgrade` updated to the current `spec_version`. I could not find any way of reading the pre upgrade spec_version.

None of the migrations performs many reads or writes. Moreover, they were adjusted such that no written storage is mutated. Thus, they cannot do harm.

# Checklist:

### DIFF
--- a/runtime/centrifuge/Cargo.toml
+++ b/runtime/centrifuge/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "centrifuge-runtime"
-version = "0.10.22"
+version = "0.10.23"
 authors = ["Centrifuge <admin@centrifuge.io>"]
 edition = "2021"
 build = "build.rs"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1141,7 +1141,7 @@ dependencies = [
 
 [[package]]
 name = "centrifuge-runtime"
-version = "0.10.22"
+version = "0.10.23"
 dependencies = [
  "axelar-gateway-precompile",
  "cfg-primitives",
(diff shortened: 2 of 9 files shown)


runtime/centrifuge/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "centrifuge-runtime"
-version = "0.10.22"
```
lemunozm, line 3: I understand this has not been changed for `altair` because we are using development as demo now.

lemunozm (approved):
Not an expert in this topic, but changes look good!

cdamian (approved):
Thank you @wischli!

runtime/centrifuge/Cargo.toml
wischli: The correct reason would be that we had not upgraded Algol or Altair yet. In contrast, Catalyst received two upgrades already from 1020 (current Centrifuge version) to 1022.

lemunozm: Thanks for extend!

NunoAlexandre (approved):

PR #1588, opened 2023-10-10
Integration tests: fudge support for the generic environment

lemunozm:
# Description

This PR adds support for `fudge` to be used as a generic environment over https://github.com/centrifuge/centrifuge-chain/pull/1583

### Moving from runtime environment to fudge environment

If you had a test as the following:
```rust
fn foo<T: Runtime>() {
    let env = RuntimeEnv::<T>::from_storage(...);
    //... unaltered rest of the test...
}

crate::test_for_runtimes!([development, altair, centrifuge], foo);
```

you can upgrade the test to use fudge as follows:

```rust
fn foo<T: Runtime + FudgeSupport>() {
    let env = FudgeEnv::<T>::from_storage(...);
    //... unaltered rest of the test...
}

crate::test_for_runtimes!([development, altair, centrifuge], foo);
```

The `FudgeSupport` trait allows all features from fudge, and `FudgeEnv` implements an environment that uses them under the hood.

### How does it work?

A fudge struct (`fudge::companion` struct) for each runtime implements the runtime agnostic trait `FudgeHandle` that acts as a wrapper and unlocks all fudge features in a generic way. The `FudgeEnv` type implements the `Env` trait but makes use of the `FudgeHandle` that is only available from the tests if the `FudgeSupport` marker trait is added to the `Runtime`.

## Changes and Descriptions

- [x] `FudgeHandle`
- [ ] `FudgeEnv` implementation
- [ ] Support for all runtimes


runtime/integration-tests/src/generic/environment.rs
```diff
@@ -26,11 +31,75 @@ pub trait Env<T: Runtime> {
...
+	/// Creates an extrinsic, used by mainly by the own environment.
+	/// To create and submit an extrinsic, see `submit()`
+	fn create_extrinsic(
+		&self,
+		who: Keyring,
+		call: impl Into<T::RuntimeCall>,
+	) -> <T::Block as Block>::Extrinsic {
+		self.state(|| {
+			let runtime_call = call.into();
+			let nonce = frame_system::Pallet::<T>::account(who.to_account_id()).nonce;
```
lemunozm, line 43: Seems like extracting the expected `nonce` directly from `frame_system` allows to not manage an `nonce` in the `fudge::companion` struct

lemunozm:
I wanted to expand this point from the description a bit more:
 
- *Consolidate `submit()` behavior.*
   - *`RuntimeEnv` modifies the state after `submit(extrinsic)` and events can be read directly.*
   - *`FudgeEnv` required one block to modify the state after `submit(extrinsic)` and be able to read events.*

Basically, we have different behaviors for the same `Env` API depending of the environment used. `RuntimeEnv` applies change instantly, and `FudgeEnv` requires one block to make them effective. Options to solve this dichotomy:

- Option 1: We transform `RuntimeEnv` to work as `FudgeEnv`. This solution is easy to achieve, but I think we lose the instant feedback from `RuntimeEnv`, which is a desired behavior and makes testing much easier for many test cases.

- Option 2: We create a "temporal storage" in `FudgeEnv` when a `submit(extrinsic)` is done, emulating its application in the storage, to get instant feedback from the test ([frederic suggestion](https://kflabs.slack.com/archives/C04GJQAM9P0/p1696509873359889?thread_ts=1696233663.801349&cid=C04GJQAM9P0)). The question then is, if tests then would check against such "temporal storage", why use `FudgeEnv` instead of `RuntimeEnv`? How do we debug something that seems correct but fails later for the real fudge storage?

- Option 3: We live with this "dichotomy", some tests can be done easier with `RuntimeEnv` expecting instant feedback at reading events just after submitting extrinsic, and another test will be done by fudge where the state doesn't change until the chain evolves at least 1 block. The tester should choose the environment depending on their needs and know about this difference in behavior.

runtime/integration-tests/src/generic/environment.rs
lemunozm: *This only works if the extrinsic is applied instantly*

lemunozm:
The PR with the proposed changes is done.

I'm adding loan's tests to polish some minor part, but the main structure can be reviewed or receive feedback

NunoAlexandre:
> The PR with the proposed changes is done.
> 
> I'm adding loan's tests to polish some minor part, but the main structure can be reviewed or receive feedback

I would love to review this later today 🤟

runtime/integration-tests/src/generic/envs/fudge_env.rs
```diff
@@ -1 +1,152 @@
...
+				// More information, issue: https://github.com/centrifuge/fudge/issues/67
+			})?;
+
+		self.nonce_storage.insert(who, nonce + 1);
+
+		Ok(())
+	}
+
+	fn state_mut<R>(&mut self, f: impl FnOnce() -> R) -> R {
+		self.handle.parachain_mut().with_mut_state(f).unwrap()
```
cdamian, line 72: Should we maybe drop the `unwrap` and return the result here? Thinking that maybe the caller should then decide how to handle/unwrap/expect that further.

runtime/integration-tests/src/generic/envs/fudge_env/handle.rs
```diff
@@ -0,0 +1,263 @@
...
+
+	fn append_extrinsic(&mut self, chain: Chain, extrinsic: Vec<u8>) -> Result<(), ()>;
+
+	fn with_state<R>(&self, chain: Chain, f: impl FnOnce() -> R) -> R;
+	fn with_mut_state<R>(&mut self, chain: Chain, f: impl FnOnce() -> R) -> R;
+	fn evolve(&mut self);
+
+	fn build(relay_storage: Storage, parachain_storage: Storage) -> Self;
+
+	fn build_relay(
```
cdamian, line 127: nit - would rename to `relay_builder` since that's what we're returning.

runtime/integration-tests/src/generic/envs/fudge_env/handle.rs
```diff
@@ -0,0 +1,263 @@
...
+		storage: Storage,
+	) -> RelaychainBuilder<Self::RelayConstructApi, Self::RelayRuntime> {
+		sp_tracing::enter_span!(sp_tracing::Level::INFO, "Relay - StartUp");
+
+		let code = Self::RELAY_CODE.unwrap();
```
cdamian, line 132: I would prefer handling this and following unwraps or at least have some meaningful messages that can help our future selves in debugging this better ^^

runtime/integration-tests/src/generic/envs/fudge_env/handle.rs
```diff
@@ -0,0 +1,263 @@
...
+		let dp: DigestCreator<RelayBlock> = Box::new(move |parent, inherents| async move {
+			let babe = FudgeBabeDigest::<RelayBlock>::new();
+			let digest = babe.build_digest(&parent, &inherents).await?;
+			Ok(digest)
+		});
+
+		RelaychainBuilder::new(init, |client| (cidp(client), dp))
+	}
+
+	fn build_parachain(
```
cdamian, line 192: nit - rename to `parachain_builder`?

runtime/integration-tests/src/generic/envs/fudge_env.rs
lemunozm: Mmm.. do you know exactly which kind of events could `with_mut_state(f)` make fail?

From my point of view, if such events can not be handle by the tester before calling this call, then we should return the error, instead, if the tester could know before calling `with_mut_state()` which things they should change not to have an error, then an `unwrap()`  is ok because it acts as a precondition that any successful test must satisfy.

runtime/integration-tests/src/generic/envs/fudge_env/handle.rs
lemunozm: Agree, maybe then the above `build()` method should be also modified to `new_environment()` or something else, WDYT?

runtime/integration-tests/src/generic/envs/fudge_env/handle.rs
lemunozm: I would say it's fine not to propagate the error here, because from a tester perspective, you always want a `builder()`, if not, you're doing something wrong in using this API.

I think unwrap/expect are fine to be used when you want to force a precondition of the usage of the API. In production, we should not use them because of the security implications of reaching a `unwrap` someday after some refactoring. However, having an `unwrap` near the originator of the error helps debugging and for testing purposes, I think improves the feedback cycle.

runtime/integration-tests/src/generic/envs/fudge_env/handle.rs
cdamian: Yes, or maybe `new` would suffice.

runtime/integration-tests/src/generic/envs/fudge_env.rs
cdamian: Hmmm, only fudge-specific actions might trigger an error in this case, e.g. the backend failing to retrieve state at block X.

lemunozm: And if it fails, it's an error from the tester making a wrong configuration for its test?

cdamian: Highly unlikely, maybe my initial suggestion can be disregarded for now - we could address this later if really needed.

Sorry for the noise.

lemunozm: Absolutely not! Pretty good question to think about 👌🏻

Let's see then how it evolves keeping this in mind 👍🏻

cdamian (approved):

runtime/integration-tests/src/generic/cases/loans.rs
```diff
@@ -0,0 +1,52 @@
...
+const POOL_A: PoolId = 23;
+const NFT_A: (CollectionId, ItemId) = (1, ItemId(10));
+
+fn borrow<T: Runtime + FudgeSupport>() {
+	let mut env = RuntimeEnv::<T>::from_storage(
+		Genesis::<T>::default()
+			.add(genesis::balances(T::ExistentialDeposit::get() + FOR_FEES))
+			.add(genesis::assets(vec![Usd6::ID]))
+			.add(genesis::tokens(vec![(Usd6::ID, Usd6::ED)]))
+			.storage(),
```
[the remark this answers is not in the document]
lemunozm: The above initializes a currency in `orml_asset_registry` called `Usd6` (6 decimals) with a metadata whose ED correctly matches the `orml_tokens`. No more balance than `ED` is given to all users (ALICE, BOB...) for that currency.

runtime/integration-tests/src/generic/utils/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+	)
+	.unwrap();
+
+	// In order to later close the epoch fastly,
+	// we mofify here that requirement to significalty reduce the testing time.
+	// The only way to do it is breaking the integration tests rules mutating
+	// this state directly.
+	pallet_pool_system::Pool::<T>::mutate(pool_id, |pool| {
+		pool.as_mut().unwrap().parameters.min_epoch_time = POOL_MIN_EPOCH_TIME;
+	});
```
lemunozm, line 116: @mustermeiszer, my question about epoch time in slack came from this. If I call:
```rust
env.pass(Blocks::BySeconds(T::DefaultMinEpochTime::get()))
```
it takes around 10 min per test, even without fudge, only passing blocks to reach the point in time where I can actually closing the epoch.

If I modify the pool to set to the min bound, it takes around 60 seconds, but I think it's still not feasible if we expect to have a lot of loans tests (or any other thing that requires a funded pool).

The solution was to hardcode a really low value, but I don't like it because I should use something more high-level to deal with it. Doing this is really obscure.

lemunozm:
This is ready for a final review & merge

runtime/integration-tests/src/generic/utils/genesis.rs
```diff
@@ -18,3 +32,117 @@ impl Genesis {
...
+}
+
+pub fn assets<T: Runtime>(currency_ids: Vec<CurrencyId>) -> impl GenesisBuild<T> {
+	orml_asset_registry::GenesisConfig::<T> {
+		assets: currency_ids
+			.into_iter()
+			.map(|currency_id| (currency_id, currency::find_metadata(currency_id).encode()))
+			.collect(),
+		last_asset_id: Default::default(), // It seems deprecated
+	}
```
mustermeiszer, line 68: Why redo what was already in the non generic part?

mustermeiszer (approved):
Nit. Otherwise locks sick!

runtime/integration-tests/src/generic/utils/genesis.rs
lemunozm: If you mean, why not reuse the genesis functions from `utils/genesis.rs` is because:
- I've tried, based on them, to make them more generic. The genesis methods from that module assume some currencies and initial values that maybe are not needed for some tests. These implementations are a bit more generic. Ideally, I would need to modify the others but doing it would make a lot of tests fail. Out of the scope of this PR, but we'll end up unifying them.
- Nit. Returning `impl GenesisBuild<T>` improves the ergonomics with the `Genesis<T>` type (added in the previous PR). Passing a mutable ref to storage and later assimilating is a bit awkward, not sure why Substrate has chosen this.

lemunozm: i.e: with this:
```rust
Genesis::<T>::default()
   .add(genesis::balances(T::ExistentialDeposit::get() + FOR_FEES))
   .add(genesis::assets(vec![Usd6::ID]))
   .add(genesis::tokens(vec![(Usd6::ID, Usd6::ED)]))
   .storage()
```
You know exactly how your initial state is for your current test without looking into the genesis functions. No more balances that can you see, nor more assets.

lemunozm: You can always wrap a common initial state under some method and then call this method for a set of tests that requires that specific state. In the loan example, the common state is a function intended to be called from different loan tests that need the same pool state.

It's a very low-level foundation

lemunozm: Actually, each test domain could implement a `GenesisBuild` where its initial state is built using the previous pieces, and later:

```rust
Genesis::<T>::default()
   .add(genesis::loans_initial_state())
   .storage()
```

or 

```rust
Genesis::<T>::default()
   .add(genesis::empty_pool())
   .storage()
```

Things like those. Just drafting an idea of how it can evolve

lemunozm: My point in favor of this is that I feel quite frustrated when a test I didn't write fails, and I need to dig into the first calls to know precisely what was the initial state it was built. And it was because of some global thing I didn't consider that was related to the failure.

PR #1590, opened 2023-10-13
integration-tests: Use latest version of fudge-polkadot-v0.9.43

cdamian:
Fixing the integration test setup using the latest fudge version.


cdamian:
The xcm-related tests are still failing due to the same error thrown by the XCMP queue - `ErrorDelivering`. 

Main reason for this being the "partial use" of the `xcm-simulator`, to be more specific:

- as per the example provided [here](https://github.com/paritytech/polkadot/blob/release-v0.9.43/xcm/xcm-simulator/example/src/lib.rs#L40), the `XCMP/DMP` message handlers are mocked, whereas we use the actual `XCMP/DMP` queues.
- another requirement for using this setup would be the usage of the `ParachainXcmRouter` found [here](https://github.com/paritytech/polkadot/blob/release-v0.9.43/xcm/xcm-simulator/src/lib.rs#L382).

Given the above, in the current approach I resumed to doing the following changes on the relay side:
- manually setting the required storages for `HRMP` message handling [here](https://github.com/centrifuge/centrifuge-chain/pull/1590/files#diff-ea32e10e1dcb43c14741c9ce554b6319258f3f735988084cb9ea543d1c6dcb5cR105).
- force-opening the `HRMP` channels [here](https://github.com/centrifuge/centrifuge-chain/pull/1590/files#diff-ea32e10e1dcb43c14741c9ce554b6319258f3f735988084cb9ea543d1c6dcb5cR137). Ideally this would've been done via the genesis config, however, the relevant field in there is not private...

However, the above relay changes are not sufficient given that certain storages have to be also populated accordingly on the parachain side.

When attempting to deliver an `XCM` message, the `XCMP` pallet attempts to retrieve the max message size for a channel [here](https://github.com/paritytech/cumulus/blob/polkadot-v0.9.43/pallets/xcmp-queue/src/lib.rs#L517), which ends up querying the `RelevantMessagingState` storage of the parachain system pallet [here](https://github.com/paritytech/cumulus/blob/master/pallets/parachain-system/src/lib.rs#L1020).

The issue here is that this storage is set when `set_validation_data` is called during block building, as I found out during the work that I did on `fudge` recently.

Given that using the `xcm-simulator` might require us to use mocked components for the `XCM` messaging, I think we should re-consider the approach here and to discuss more about using `fudge` for this purpose since we already have a [WIP PR](https://github.com/centrifuge/fudge/pull/66) that addresses this topic.

cc @NunoAlexandre @mustermeiszer

Cargo.toml
```diff
@@ -293,153 +293,153 @@ testnet-runtime = [
...
-try-runtime-cli = { git = "https://github.com/NunoAlexandre//substrate", branch = "polkadot-v0.9.43" }
-substrate-test-runtime = { git = "https://github.com/NunoAlexandre//substrate", branch = "polkadot-v0.9.43" }
-substrate-test-runtime-client = { git = "https://github.com/NunoAlexandre//substrate", branch = "polkadot-v0.9.43" }
-node-primitives = { git = "https://github.com/NunoAlexandre//substrate", branch = "polkadot-v0.9.43" }
-pallet-nomination-pools = { git = "https://github.com/NunoAlexandre//substrate", branch = "polkadot-v0.9.43" }
-sc-service-test = { git = "https://github.com/NunoAlexandre//substrate", branch = "polkadot-v0.9.43" }
-sc-network-transactions = { git = "https://github.com/NunoAlexandre//substrate", branch = "polkadot-v0.9.43" }
-sp-consensus-beefy = { git = "https://github.com/NunoAlexandre//substrate", branch = "polkadot-v0.9.43" }
-sp-consensus-grandpa = { git = "https://github.com/NunoAlexandre//substrate", branch = "polkadot-v0.9.43" }
+binary-merkle-tree = { git = "https://github.com/paritytech//substrate", branch = "polkadot-v0.9.43" }
```
cdamian, line 296: @NunoAlexandre not sure why you used your fork here, however, this is one of the main reasons why I addressed these changes in this separate PR.

NunoAlexandre:
> The xcm-related tests are still failing due to the same error thrown by the XCMP queue - `ErrorDelivering`.
> 
> Main reason for this being the "partial use" of the `xcm-simulator`, to be more specific:
> 
> * as per the example provided [here](https://github.com/paritytech/polkadot/blob/release-v0.9.43/xcm/xcm-simulator/example/src/lib.rs#L40), the `XCMP/DMP` message handlers are mocked, whereas we use the actual `XCMP/DMP` queues.
> * another requirement for using this setup would be the usage of the `ParachainXcmRouter` found [here](https://github.com/paritytech/polkadot/blob/release-v0.9.43/xcm/xcm-simulator/src/lib.rs#L382).
> 
> Given the above, in the current approach I resumed to doing the following changes on the relay side:
> 
> * manually setting the required storages for `HRMP` message handling [here](https://github.com/centrifuge/centrifuge-chain/pull/1590/files#diff-ea32e10e1dcb43c14741c9ce554b6319258f3f735988084cb9ea543d1c6dcb5cR105).
> * force-opening the `HRMP` channels [here](https://github.com/centrifuge/centrifuge-chain/pull/1590/files#diff-ea32e10e1dcb43c14741c9ce554b6319258f3f735988084cb9ea543d1c6dcb5cR137). Ideally this would've been done via the genesis config, however, the relevant field in there is not private...
> 
> However, the above relay changes are not sufficient given that certain storages have to be also populated accordingly on the parachain side.
> 
> When attempting to deliver an `XCM` message, the `XCMP` pallet attempts to retrieve the max message size for a channel [here](https://github.com/paritytech/cumulus/blob/polkadot-v0.9.43/pallets/xcmp-queue/src/lib.rs#L517), which ends up querying the `RelevantMessagingState` storage of the parachain system pallet [here](https://github.com/paritytech/cumulus/blob/master/pallets/parachain-system/src/lib.rs#L1020).
> 
> The issue here is that this storage is set when `set_validation_data` is called during block building, as I found out during the work that I did on `fudge` recently.

Thanks a lot for the throughout work and update, @cdamian.


> Given that using the `xcm-simulator` might require us to use mocked components for the `XCM` messaging, I think we should re-consider the approach here and to discuss more about using `fudge` for this purpose since we already have a [WIP PR](https://github.com/centrifuge/fudge/pull/66) that addresses this topic.


I wholeheartedly agree. I noticed that Acala, for instance, followed the same route but using Chopsticks. I agree this should be the way forward. We could delete or skip these failing tests on this update but I am afraid that by doing so we won't catch actual issues that will have our xcm setup break in production. When will the Fudge version that could support XCM testing be ready and when could we switch to using that? if it's in a matter of days I think it makes sense to align these two pieces of work and wait to merge this update until then. If it's longer, we need to think of an intermediary safe approach.

Cargo.toml
NunoAlexandre: We can't compile the official version, failing with the error: 

```
warning: `pallet-block-rewards` (lib) generated 2 warnings (run `cargo fix --lib -p pallet-block-rewards` to apply 1 suggestion)
     Compiling liquidity-pools-gateway-routers v0.0.1 (/Users/n/centrifuge/centrifuge-chain/pallets/liquidity-pools-gateway/routers)
  error[E0599]: no function or associated item named `assert_last_event` found for struct `frame_system::Pallet` in the current scope
    --> /Users/n/.cargo/git/checkouts/substrate-f4423eebfa0046a3/5e49f6e/frame/system/benchmarking/src/lib.rs:55:16
     |
  55 |         System::<T>::assert_last_event(frame_system::Event::<T>::CodeUpdated.into());
     |                      ^^^^^^^^^^^^^^^^^ function or associated item not found in `Pallet<T>`
```


I tried using subalfred and featalign to fix any potential mixing feature flags and tried using different rust toolchains but couldn't get any further. So my fork basically just silences that line on `frame-system-benchmarking` so that we can continue with the work on our side. We can always get back to it at the end.

cdamian:
@NunoAlexandre I'm hoping that the fudge version with XCM support will be a matter of a couple of days, however, I have to double-check that with @mustermeiszer

mustermeiszer:
@NunoAlexandre are we actually checking against the moonbeam runtime?

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2505 @@
...
+
+			env.sibling_state_mut(|| {
+				let current_balance =
+					orml_tokens::Pallet::<T>::free_balance(air_in_sibling, &Keyring::Bob.into());
+
+				// Verify that Keyring::Bob now has (amount transferred - fee)
+				assert_eq!(current_balance, transfer_amount - fee(18));
+
+				// Sanity check for the actual amount Keyring::Bob ends up with
+				assert_eq!(current_balance, 4992960800000000000);
```
cdamian, line 503: Old value - `4991987200000000000`, as per - https://github.com/centrifuge/centrifuge-chain/blob/f46d327889fd7376c587331e8f2a795dc59b340d/runtime/integration-tests/src/liquidity_pools/pallet/kusama/tests/transfers.rs#L175C31-L175C50

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2505 @@
...
+			});
+
+			env.pass(Blocks::ByNumber(1));
+
+			env.relay_state_mut(|| {
+				assert_eq!(
+					pallet_balances::Pallet::<FudgeRelayRuntime<T>>::free_balance(
+						&Keyring::Bob.into()
+					),
+					999907996044
```
cdamian, line 804: Old value - `999909712564`. As per - https://github.com/centrifuge/centrifuge-chain/blob/f46d327889fd7376c587331e8f2a795dc59b340d/runtime/integration-tests/src/liquidity_pools/pallet/kusama/tests/transfers.rs#L410C4-L410C16

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2505 @@
...
+			});
+
+			env.pass(Blocks::ByNumber(2));
+
+			env.parachain_state_mut(|| {
+				let bob_balance =
+					orml_tokens::Pallet::<T>::free_balance(usdc_asset_id, &Keyring::Bob.into());
+
+				// Sanity check to ensure the calculated is what is expected
+				assert_eq!(bob_balance, 11992961);
```
cdamian, line 1014: Old value - `11991988`. As per - https://github.com/centrifuge/centrifuge-chain/blob/f46d327889fd7376c587331e8f2a795dc59b340d/runtime/integration-tests/src/liquidity_pools/pallet/kusama/tests/transfers.rs#L589C27-L589C35

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2505 @@
...
+
+			env.sibling_state_mut(|| {
+				let current_balance =
+					orml_tokens::Pallet::<T>::free_balance(cfg_in_sibling, &Keyring::Bob.into());
+
+				// Verify that Keyring::Bob now has (amount transferred - fee)
+				assert_eq!(current_balance, transfer_amount - fee(18));
+
+				// Sanity check for the actual amount Keyring::Bob ends up with
+				assert_eq!(current_balance, 4992960800000000000);
```
cdamian, line 1996: Old value - `4991987200000000000`. As per - https://github.com/centrifuge/centrifuge-chain/blob/f46d327889fd7376c587331e8f2a795dc59b340d/runtime/integration-tests/src/liquidity_pools/pallet/polkadot/tests/transfers.rs#L159C31-L159C50

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2505 @@
...
+			});
+
+			env.pass(Blocks::ByNumber(1));
+
+			env.relay_state_mut(|| {
+				assert_eq!(
+					pallet_balances::Pallet::<FudgeRelayRuntime<T>>::free_balance(
+						&Keyring::Alice.into()
+					),
+					79628418552
```
cdamian, line 2286: Old value - `79637471000`. As per - https://github.com/centrifuge/centrifuge-chain/blob/f46d327889fd7376c587331e8f2a795dc59b340d/runtime/integration-tests/src/liquidity_pools/pallet/polkadot/tests/transfers.rs#L381C4-L381C15

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2505 @@
...
+			});
+
+			env.pass(Blocks::ByNumber(2));
+
+			env.parachain_state_mut(|| {
+				let bob_balance =
+					orml_tokens::Pallet::<T>::free_balance(usdc_asset_id, &Keyring::Bob.into());
+
+				// Sanity check to ensure the calculated is what is expected
+				assert_eq!(bob_balance, 11992961);
```
cdamian, line 2495: Old value - `11991988`. As per - https://github.com/centrifuge/centrifuge-chain/blob/f46d327889fd7376c587331e8f2a795dc59b340d/runtime/integration-tests/src/liquidity_pools/pallet/polkadot/tests/transfers.rs#L559C27-L559C35

runtime/integration-tests/src/runtime_apis/rewards.rs
```diff
@@ -23,11 +23,11 @@ use tokio::runtime::Handle;
 use super::ApiEnv;
 use crate::utils::accounts::Keyring;
 
-#[tokio::test]
-async fn liquidity_rewards_runtime_api_works() {
-	rewards_runtime_api_works::<development_runtime::LiquidityRewardsBase>(RewardDomain::Liquidity)
-		.await;
-}
+// #[tokio::test]
+// async fn liquidity_rewards_runtime_api_works() {
```
cdamian, line 27: Currently failing since the 2nd attempt to distribute rewards is trying to mint a balance of 0 into the staker which throws a `BelowMinimum` error.

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
```diff
@@ -3045,188 +3182,26 @@ mod mismatching_currencies {
...
-					}
-					.into()
-			}));
-		});
-	}
-
 	/// Similar to [concurrent_swap_orders_same_direction] but with partial
 	/// fulfillment
-	#[test]
-	fn partial_fulfillment_concurrent_swap_orders_same_direction() {
```
mustermeiszer, line 3218: Is this test gone?

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
```diff
@@ -3045,188 +3182,26 @@ mod mismatching_currencies {
...
-		TestNet::reset();
-		Development::execute_with(|| {
-			// Increase invest setup
-			setup_pre_requirements();
+	#[tokio::test]
+	async fn fulfill_redeem_swap_order_requires_collect() {
```
mustermeiszer, line 3188: Why are the docs for this one gone?

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2505 @@
...
+								)
+							)
+							.into()
+						),
+						WeightLimit::Limited(8_000_000_000_000.into()),
+					),
+					orml_xtokens::Error::<T>::NotCrossChainTransferableCurrency
+				);
+			});
+		}
```
mustermeiszer, line 275: couldn't those be made used for all runtimes inestead of inside a `mod kusama`?

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2505 @@
...
+			env.parachain_state_mut(|| {
+				// Verify that Keyring::Alice now has initial balance + amount transferred - fee
+				assert_eq!(
+					pallet_balances::Pallet::<T>::free_balance(&Keyring::Alice.into()),
+					alice_initial_balance + transfer_amount - air_fee(),
+				);
+			});
+		}
+
+		fn transfer_ausd_to_altair<T: Runtime + FudgeSupport>() {
```
mustermeiszer, line 586: This test should best be triggered from the relay to altair. And not sibling to altair.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2505 @@
...
+			env.parachain_state_mut(|| {
+				// Verify that Keyring::Alice now has initial balance + amount transferred - fee
+				assert_eq!(
+					pallet_balances::Pallet::<T>::free_balance(&Keyring::Alice.into()),
+					alice_initial_balance + transfer_amount - cfg_fee(),
+				);
+			});
+		}
+
+		fn transfer_ausd_to_centrifuge<T: Runtime + FudgeSupport>() {
```
mustermeiszer, line 2080: @cdamian how much work would it be to have another parachain that acts as the reserve for USDC (i.e. AssetHub) and test transferring in and out from that?

mustermeiszer (commented):
LOVE IT! Just need to clarify a few things otherwise looks good!

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
mustermeiszer: EDIT: 

This is actually a good test. We should though think about having one where ausd acts as the reserve.

runtime/integration-tests/Cargo.toml
```diff
@@ -10,10 +10,12 @@ repository = "https://github.com/centrifuge/centrifuge-chain"
 [dependencies]
 codec = { package = "parity-scale-codec", version = "3.0", default-features = false, features = ["derive"] }
 fudge = { git = "https://github.com/centrifuge/fudge", branch = "polkadot-v0.9.43" }
+fudge-core = { git = "https://github.com/centrifuge/fudge", branch = "polkadot-v0.9.43" }
```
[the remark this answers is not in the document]
cdamian: Companion macro - https://github.com/centrifuge/fudge/blob/master/core/src/builder/companion/src/expand/mod.rs#L22

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
cdamian: Yes, I think that applies to a lot of these tests tbh. For this one, I wanted to have a 1 to 1 transition more or less.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
cdamian: Assuming you want to also use the `AssetHub` runtime right? We would need to change the fudge env to provide a builder for that and then use it as a para. Furthermore, adding one more parachain would make block building even slower, imo that's something that we should consider.

runtime/integration-tests/src/generic/envs/fudge_env.rs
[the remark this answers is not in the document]
cdamian: Not sure what you mean by updating the relay state. Can you be more specific please? ^^

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
cdamian: I'll have to double check this since I think William did some changes to these earlier.

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
cdamian: Same, I'll double check and get back to ya.

runtime/integration-tests/src/generic/env.rs
```diff
@@ -94,19 +98,30 @@ pub trait Env<T: Runtime> {
...
+	fn parachain_state<R>(&self, f: impl FnOnce() -> R) -> R;
+
+	/// Allows to mutate the sibling storage state through the closure.
+	fn sibling_state_mut<R>(&mut self, f: impl FnOnce() -> R) -> R;
 
-	/// Allows to read the storage state through the closure
-	/// If storage is modified, it would not be applied.
-	fn state<R>(&self, f: impl FnOnce() -> R) -> R;
+	/// Allows to read the sibling storage state through the closure.
+	fn sibling_state<R>(&self, f: impl FnOnce() -> R) -> R;
```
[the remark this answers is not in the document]
mustermeiszer: Okay. I meant to keep `parachain` but make sibling more generic. But we can always change that once we have more than one other parachain we test against.

runtime/integration-tests/src/liquidity_pools/pallet/development/transfers.rs
```diff
@@ -89,19 +109,51 @@ fn transfer_cfg_to_sibling() {
...
+				1,
+				Junctions::X1(Junction::Parachain(PARA_ID_SIBLING)),
+			)),
+			XCM_VERSION,
+		));
 	});
 
-	Moonbeam::execute_with(|| {
-		assert_eq!(OrmlTokens::free_balance(cfg_in_sibling, &BOB.into()), 0);
+	env.with_mut_state(Chain::Relay, || {
```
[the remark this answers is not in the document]
cdamian: Yes, I actually did that in the generic LP tests in `setup_xcm`. If and when we move this test there, we could just use that.

runtime/integration-tests/src/generic/envs/fudge_env.rs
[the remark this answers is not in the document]
mustermeiszer: The `update_para` logic you implemented in fudge

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
mustermeiszer: Okay. Lets keep it for now.

runtime/integration-tests/src/utils/env.rs
```diff
@@ -777,21 +775,39 @@ fn test_env(
 	// Build relay-chain builder
 	let relay = {
 		sp_tracing::enter_span!(sp_tracing::Level::INFO, "Relay - StartUp");
-		let mut state = StateProvider::new(RelayCode.expect("Wasm is build. Qed."));
 
-		// We need to HostConfiguration and use the default here.
+		//TODO(cdamian): Use RelayBlock
+		let mut state =
+			StateProvider::<TFullBackend<centrifuge::Block>, centrifuge::Block>::empty_default(
```
cdamian, line 781: @mustermeiszer Shouldn't this fail since we're not using the relay block?

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
cdamian: OK, then I'll create a separate module here called `all` and move these over + use all runtimes.

runtime/integration-tests/src/utils/env.rs
mustermeiszer: mmh. Probably we have the same as the relay.

runtime/integration-tests/src/generic/envs/fudge_env.rs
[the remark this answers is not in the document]
cdamian: You mean updating the para head in the relay after executing the extrinsic submission right?

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
cdamian: Nvm, I screwed these up, sorry about that. Fixed them, now I'm testing them locally.

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
cdamian: Will fix. Thanks for catching these!

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
cdamian: Fixed.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
cdamian: Done.

runtime/integration-tests/Cargo.toml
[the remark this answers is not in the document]
mustermeiszer: WE should re-export taht ^^

runtime/integration-tests/src/generic/mod.rs
```diff
@@ -60,6 +61,8 @@ macro_rules! test_for_runtimes {
             $(
                 #[tokio::test]
                 async fn $runtime_name() {
+                    crate::utils::logs::init_logs();
```
lemunozm, line 64: Do we really want this for all tests by default? If it's for debugging purposes, why not add it during the debugging phase for the test that is currently failing?

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -228,6 +270,7 @@ mod tests {
 					balances: vec![(Keyring::Alice.to_account_id(), 1 * CFG)],
 				})
 				.storage(),
+			Genesis::<T>::default().storage(),
```
lemunozm, line 273: NIT. I think `Storage::default()` works here.

lemunozm (commented):
Really good addition here @cdamian!

It's great to see it grow!

runtime/integration-tests/src/generic/mod.rs
cdamian: I think having it here makes sense and one can tweak the log level when debugging tests individually, if needed.

lemunozm: Ok, my comment comes mainly from [this issue](https://github.com/centrifuge/centrifuge-chain/actions/runs/6864028905/job/18664932956?pr=1606), where the amount of stdout increases the CI time until the timeout.

I think successful tests should not give any output (nobody looks logs in successful tests), only the ones that fail. Not sure how to configure the logger to act in that way and also reduce the amount of stdout in CI.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -0,0 +1,2371 @@
...
+	use altair_runtime::{CurrencyIdConvert, PoolPalletIndex};
+
+	pub const KSM_ASSET_ID: CurrencyId = CurrencyId::ForeignAsset(1000);
+
+	use super::*;
+
+	mod utils {
+		use super::*;
+
+		pub fn register_ausd<T: Runtime + FudgeSupport>() {
```
lemunozm, line 125: Just saw this. For these methods, you have a utility to define currencies in a way that later you can later refer to their properties easily in https://github.com/centrifuge/centrifuge-chain/blob/main/runtime/integration-tests/src/generic/utils/currency.rs

If you define AUSD there, for example, you can later write something like `Ausd::ED`

runtime/integration-tests/src/generic/mod.rs
cdamian: Oh, yea, that makes sense. Well, we can set the default log level to error and then the logs would be minimal, I still think it's good to have them enabled by default so that it makes or lives easier when debugging and not having to look for the init func and add it in there manually, but then again, no strong feelings on this one.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
cdamian: Oh, that's pretty awesome, I guess I missed it while dealing with the test tornado ^^.

PR #1595, opened 2023-10-17
chore: post v0.10.34 release cleanup

wischli:
# Description

* Bumps `spec_version` for chains which were upgraded (Altair was not) by one
* Removes executed migrations

# Checklist:


mustermeiszer:
I thought we bumb the spec only at the end of the release process, has this changed?

runtime/development/src/lib.rs
```diff
@@ -140,7 +140,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("centrifuge-devel"),
 	impl_name: create_runtime_str!("centrifuge-devel"),
 	authoring_version: 1,
-	spec_version: 1032,
+	spec_version: 1033,
```
lemunozm, line 143: Out of curiosity, why bumping 2 numbers?

lemunozm (approved):
Thanks for taking care of this kind of PRs!

mustermeiszer:
> However, I have no strong opinion here. WDYT?

No, I am fine with that too. Was first wondering whether it incures any danger of upgrading without notice, but shouldn't be.

PR #1600, opened 2023-10-26
Loans: Integration tests

### LINKED ISSUE #1458: Loans: add integration tests

lemunozm:
Fixes #1458 #1375

## Changes and Descriptions

- Add main loan use cases (under `generic/src/cases/loans.rs`):
  - [x] Entire flow with internal priced loans
  - [x] Entire flow with external priced loans using oracles
  - [x] Mutating a loan through a pool change guard.
  - [x] ~Check all kind of mutations based on different restriction.~ This is blocked since there is no way to pass one week without burning the computer during 15-minute computing blocks. 
    - One possibility is to mock the time, it means, pass a week in just one block. Although some state could not be "real" after this. Postpone for a future PR (related issue: https://github.com/centrifuge/centrifuge-chain/issues/1499) 
- Removed obsoleted loan integration tests and moved some env tests to their more related placed (most of the modified diff files come from this, sorry 🙏🏻).
- Reorganize generic modules under integration tests to be more readable, now:
```
runtime/integration-tests/src/generic
├── cases               // Use cases go here
│  ├── example.rs
│  └── loans.rs
├── envs                // Environments go here
│  ├── fudge_env
│  │  └── handle.rs
│  ├── fudge_env.rs
│  └── runtime_env.rs
├── utils               // Generic utilities for any use case:
│  ├── currency.rs      // Easy support for interacting with currencies as ED or its decimals
│  ├── genesis.rs       // Generic support for initializing pallets
│  └── mod.rs
├── config.rs           // Where Runtime trait and other configs are defined.
├── env.rs              // Environment trait with basic defaults
├── impls.rs            // where runtimes implement the new traits
├── mod.rs
└── README.md
```
- Fixes a bug in processing blocks `BySeconds`
- Add `Block::JumpBySeconds` see [comment](https://github.com/centrifuge/centrifuge-chain/pull/1600#discussion_r1373517636)


runtime/integration-tests/src/generic/impls.rs
```diff
@@ -0,0 +1,129 @@
...
+impl_runtime!(altair_runtime, Altair);
+impl_runtime!(centrifuge_runtime, Centrifuge);
+
+/// Implements fudge support for a runtime
+macro_rules! impl_fudge_support {
+	(
+        $fudge_companion_type:ident,
+        $relay_path:ident,
+        $parachain_path:ident,
+        $parachain_id:literal
```
cdamian, line 30: Awesome, we can easily extend this for adding other paras/siblings. Thank you!

runtime/integration-tests/src/utils/env.rs
```diff
@@ -935,3 +935,34 @@ pub fn pass_n(env: &mut TestEnv, n: u64) -> Result<(), ()> {
 
 	Ok(())
 }
+
+mod tests {
+	use super::*;
+
+	#[tokio::test]
+	async fn env_works() {
```
cdamian, line 943: Thank you for moving this, I was also thinking about moving it in my current branch but I'll skip.

runtime/integration-tests/src/utils/extrinsics.rs
```diff
@@ -218,3 +218,90 @@ where
...
+	use crate::{
+		chain::{
+			centrifuge,
+			centrifuge::{Runtime, PARA_ID},
+		},
+		utils::{accounts::Keyring, env, genesis},
+	};
+
+	#[tokio::test]
+	async fn extrinsics_works() {
```
cdamian, line 239: Same here, awesome!

runtime/integration-tests/src/generic/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Where I start?
+- Create a new file in `cases/<file.rs>` for the use case you want to test.
+- Maybe you need to update the `Runtime` trait in `config.rs` file with extra information from your new pallet.
+  This could imply:
+    - Adding bounds to the `Runtime` trait with your new pallet.
+    - Adding bounds to `T::RuntimeCallExt` to support calls from your pallet.
+    - Adding bounds to `T::EventExt` to support events from your pallet.
+    - Adding bounds to `T::Api` to support new api calls.
+- You can add `GenerisBuild` builders for setting the initial state of your pallet for others in `utils/genesis.rs`.
```
cdamian, line 19: nit - typo in `GenesisBuild` :)

cdamian:
I assume we want to get this in before the .43 update so we'll have to eventually get these changes in that branch.

runtime/integration-tests/src/generic/impls.rs
lemunozm: That should be the idea! 🙌🏻

runtime/integration-tests/src/generic/README.md
lemunozm: Such an eye! 👀 
Updated

runtime/integration-tests/src/generic/cases/example.rs
```diff
@@ -156,7 +157,21 @@ fn fudge_call_api<T: Runtime + FudgeSupport>() {
...
+	let before = env.state(|| pallet_timestamp::Pallet::<T>::get());
+
+	// Not supported in fudge
+	env.pass(Blocks::JumpBySeconds(SECONDS_PER_YEAR));
+
+	let after = env.state(|| pallet_timestamp::Pallet::<T>::get());
+
+	assert_eq!((after - before).into_seconds(), SECONDS_PER_YEAR)
+}
+
```
lemunozm, line 172: @cdamian I need your fudge wisdom on this.

What I'm doing with the `JumpBySeconds` in the `RuntimeEnv` is jumping to a block in the future where it has passed `SECONDS_PER_YEAR`. This works quite well to "emulate" a state in the future, only creating one block. It's not as accurate as passing all blocks equivalent to a year, but it's much much faster and unlocks these tests for CI. 

Do you think we can modify fudge to add this behavior, jumping to a block in the future without computing the intermediate blocks?

cc @mustermeiszer

cdamian (approved):

runtime/integration-tests/src/generic/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The aim of this module is to make integration-tests independently for all runtimes at once.
+
+You can choose the environment for each of your use cases:
+  - `RuntimeEnv`: Simple environment that acts as a wrapper over the runtime
+  - `FudgeEnv`: Advanced environment that use a client and connect the runtime to a relay chain.
+
+Both environment uses the same interface so jumping from one to the another should be something "smooth".
+
+## Where I start?
```
wischli, line 11: Nit
```suggestion
## Where do I start?
```

runtime/integration-tests/src/generic/cases/loans.rs
```diff
@@ -42,112 +47,323 @@ use crate::{
 const POOL_ADMIN: Keyring = Keyring::Admin;
 const INVESTOR: Keyring = Keyring::Alice;
 const BORROWER: Keyring = Keyring::Bob;
+const LOAN_ADMIN: Keyring = Keyring::Charlie;
+const ANY: Keyring = Keyring::Dave;
```
wischli, line 51: Nit: We should make it clear this refers to a substrate address
```suggestion
const ANYONE: Keyring = Keyring::Dave;
```

lemunozm:
Thanks for the approval @cdamian!

wischli (commented):
Shoot, I am a bit too late. Luckily, I only had non-blocking nitpicks. Great work @Luis! Thanks so much for the README, it really helps!

lemunozm:
Written down as doc-debt! 😆 
Thanks for the corrections!

PR #1601, opened 2023-10-26
fix: integration tests after ED overhaul

wischli:
# Description

* Fixes integration tests for changes of #1596 

# Checklist:

### DIFF
--- a/runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
+++ b/runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
@@ -103,7 +103,7 @@ mod same_currencies {
 		Development::execute_with(|| {
 			setup_pre_requirements();
 			let pool_id = DEFAULT_POOL_ID;
-			let amount = 100_000_000;
+			let amount = 10 * dollar(12);
 			let investor: AccountId =
 				AccountConverter::<DevelopmentRuntime, LocationToAccountId>::convert((
 					DOMAIN_MOONBEAM,
@@ -111,14 +111,6 @@ mod same_currencies {
 				));
 			let currency_id = AUSD_CURRENCY_ID;
 			let currency_decimals = currency_decimals::AUSD;
-
-			let sending_domain_locator = Domain::convert(DEFAULT_DOMAIN_ADDRESS_MOONBEAM.domain());
-			let pool_account =
-				pallet_pool_system::pool_types::PoolLocator { pool_id }.into_account_truncating();
-
-			Tokens::mint_into(AUSD_CURRENCY_ID, &sending_domain_locator, AUSD_ED).unwrap();
-			Tokens::mint_into(AUSD_CURRENCY_ID, &investor, AUSD_ED).unwrap();
-			Tokens::mint_into(AUSD_CURRENCY_ID, &pool_account, AUSD_ED).unwrap();
 
 			// Create new pool
 			create_currency_pool(pool_id, currency_id, currency_decimals.into());
@@ -159,7 +151,7 @@ mod same_currencies {
 		Development::execute_with(|| {
 			setup_pre_requirements();
 			let pool_id = DEFAULT_POOL_ID;
-			let invest_amount: u128 = 100_000_000;
+			let invest_amount: u128 = 10 * dollar(12);
 			let decrease_amount = invest_amount / 3;
 			let final_amount = invest_amount - decrease_amount;
 			let investor: AccountId =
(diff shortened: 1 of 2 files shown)


runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/transfers.rs
```diff
@@ -73,13 +73,17 @@ fn transfer_non_tranche_tokens_from_local() {
 		// Register GLMR and fund BOB
 		setup_pre_requirements();
 
-		let initial_balance = 100_000_000;
+		let initial_balance = 2 * AUSD_ED;
 		let amount = initial_balance / 2;
 		let dest_address = DEFAULT_DOMAIN_ADDRESS_MOONBEAM;
 		let currency_id = AUSD_CURRENCY_ID;
-		let source_account = BOB;
+		let source_account = CHARLIE;
```
wischli, line 80: Not sure why but apparently Bob was funded with the ED.

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
```diff
@@ -103,7 +103,7 @@ mod same_currencies {
 		Development::execute_with(|| {
 			setup_pre_requirements();
 			let pool_id = DEFAULT_POOL_ID;
-			let amount = 100_000_000;
+			let amount = 10 * dollar(12);
```
wischli, line 106: The issue we were facing was that we minted less than the ED, so `mint_into` threw.

lemunozm: Ouch!! Good catch, I thought that you could mint more that ED if the target account has more than `ED`

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/transfers.rs
lemunozm: Yes, by default, I added `ED` to Alice and Bob. We need to discuss how to initialize tests to avoid this kind of unexpected initialized state (I'm working on some experimental currency management in the generic integration tests to clarify this aspect).

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
```diff
@@ -112,14 +112,6 @@ mod same_currencies {
 			let currency_id = AUSD_CURRENCY_ID;
 			let currency_decimals = currency_decimals::AUSD;
 
-			let sending_domain_locator = Domain::convert(DEFAULT_DOMAIN_ADDRESS_MOONBEAM.domain());
-			let pool_account =
-				pallet_pool_system::pool_types::PoolLocator { pool_id }.into_account_truncating();
-
-			Tokens::mint_into(AUSD_CURRENCY_ID, &sending_domain_locator, AUSD_ED).unwrap();
-			Tokens::mint_into(AUSD_CURRENCY_ID, &investor, AUSD_ED).unwrap();
-			Tokens::mint_into(AUSD_CURRENCY_ID, &pool_account, AUSD_ED).unwrap();
```
lemunozm, line 121: Curious about why this is no longer needed

wischli: When we increase a foreign investment, the corresponding amount is minted. 
https://github.com/centrifuge/centrifuge-chain/blob/90d3cf9f2884489dafbc221925b2608ced4ace5e/pallets/liquidity-pools/src/inbound.rs#L111

As long as the minted amount is above the ED, we are fine.

lemunozm: This should be a requirement then for its usage, right? Nobody can send less than `ED`. Should this account be `Expendable` then?

wischli: Yes and yes!

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/transfers.rs
wischli: How about no default funding but instead builder functions for the test environment which enable funding a vector of account addresses? 

Example
https://github.com/centrifuge/centrifuge-chain/blob/90d3cf9f2884489dafbc221925b2608ced4ace5e/pallets/block-rewards/src/mock.rs#L327

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/foreign_investments.rs
lemunozm: Ok, then this should be a tech debt once we have everything passed and compiling

runtime/integration-tests/src/liquidity_pools/pallet/development/tests/liquidity_pools/transfers.rs
lemunozm: I'll think about it!

lemunozm:
Thanks a lot for your work on this @wischli! I was quite stuck looking in the wrong place.

PR #1273, opened 2023-03-21
WIP: Try again: Run unit tests in debug mode (#1238)

lemunozm:
Reverts centrifuge/centrifuge-chain#1272

**Be sure this PR does not break again the CI before merging** 

There is an issue with storage limitations [Slack thread](https://kflabs.slack.com/archives/C04GNE3JP6Z/p1679054894264889). Fix it before

PR #1606, opened 2023-11-14
CI Testing in debug

lemunozm:
# Description

CI workflows in debug instead of release.

Note: In the original PR modifying this, the workflow of checking benchmarks was also computed in debug to reuse artifacts (see [here](https://github.com/centrifuge/centrifuge-chain/pull/1273/files#diff-1f813070c7581975ad731d80bf757a5c032827b98b4f4bee9133eefa34015618R67)). Maybe it's still required here, but it's still not added. Let's see how it behaves.

### DIFF
--- a/ci/run-check.sh
+++ b/ci/run-check.sh
@@ -10,15 +10,15 @@ cargo --version
 
 case $TARGET in
   cargo-build)
-    cargo build --release "$@"
+    cargo build "$@"
     ;;
 
   test-general)
-    cargo test --workspace --release --features runtime-benchmarks,try-runtime --exclude runtime-integration-tests
+    cargo test --workspace --features runtime-benchmarks,try-runtime --exclude runtime-integration-tests
     ;;
 
   test-integration)
-    cargo test --release --package runtime-integration-tests --features fast-runtime
+    cargo test --package runtime-integration-tests --features fast-runtime
     ;;
 
   lint-fmt)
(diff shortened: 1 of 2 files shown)


wischli:
Can you temporarily adjust the trigger of sanity checks to run for this PR? I would like to compare the execution times.

lemunozm:
It makes total sense! Thanks for the suggestion @wischli

lemunozm:
Integration tests fail. I think it's because the amount of logging increases a lot over time. I would set `ERROR` level for CI.

Anyway, I am not sure if we want to have logging by default in integration tests. I would only add them during debugging to analyze a failing test case. cc @mustermeiszer?

.github/workflows/sanity-checks.yml
```diff
@@ -1,6 +1,6 @@
 on:
-  # push:
-  #   branches: [main, 'release-v**']
+  push:
+    branches: ['ci-rewrite-n-cache']
```
mustermeiszer, line 3: This should be changed back, right?

lemunozm: Yes, it there was there just to trigger the CI in this PR

wischli (commented):
Since integration tests take >4h in debug mode, at least these need to be run in release mode.

lemunozm:
Yes, but then we need to compile integration tests in release and workspace tests in debug. I think then it makes no sense running anything in debug 😞

lemunozm:
At least we can not until solving this: https://kflabs.slack.com/archives/C04GJQAM9P0/p1700154518905029

PR #1607, opened 2023-11-15
Release centrifuge 1024 client 1035

mustermeiszer:
# Description

[description]

Fixes #(issue)

## Changes and Descriptions

[List your changes here]

# Checklist:


wischli:
Have you checked how many accounts need to be migrated? I dont think its realistic to do this in a single block migration. Why dont we use Parity's script?

runtime/common/src/migrations/balances.rs
```diff
@@ -0,0 +1,138 @@
...
+
+		while let Some(next) = sp_io::storage::next_key(&previous_key) {
+			previous_key = next;
+
+			log::info!(
+				"Balances Migration - Processing account key - {}",
+				hex::encode(previous_key.clone())
+			);
+
+			/// The difference between the old and new account data structures
```
cdamian, line 78: Tested by running:

```
./target/debug/centrifuge-chain try-runtime --runtime target/debug/wbuild/centrifuge-runtime/centrifuge_runtime.wasm --chain centrifuge on-runtime-upgrade live --uri wss://fullnode.centrifuge.io:443
```

Some entries decode without any issues, however, the majority fail when decoding to both old and new, also with "unexpected" values, as per the checks listed below.

runtime/common/src/migrations/balances.rs
```diff
@@ -0,0 +1,138 @@
...
+			///
+			/// Old - confirm that `fee_frozen` is zero, as it shouldn't have
+			/// been used so far
+			/// New - confirm that `flags` does not have the
+			/// new logic flag set
+			match unhashed::get::<OldAccountInfo<T::Index, T::Balance>>(&previous_key) {
+				Some(old) => {
+					if !old.data.fee_frozen.is_zero() {
+						log::warn!("Balances Migration - Old account data with non zero frozen fee")
+					}
```
wischli, line 96: If we eventually apply this migration, we should return counters for frozen and reserved accounts and check against these in `post_upgrade`.

runtime/common/src/migrations/balances.rs
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		// CHECKING DECODING OLD DATASTRUCTURE WITH NEW LAYOUT WORKS:
+		// * Fetch storage from chain with NEW data structure
+		// * Check if fetched accounts matches on-chain storage entries
+
+		// TODO:
+		// * Research whether orml-pallets datastructure also changed
+		// * Research whether we need to migrate locks in orml (maybe first check if
+		//   there exist any. ^^
```
wischli, line 120: `orml-tokens` has not updated their `AccountData` yet:
https://github.com/open-web3-stack/open-runtime-module-library/blob/8f59b50ab45d15572d18cee0425966e43567a954/tokens/src/lib.rs#L142-L160

I would assume, they won't do that because to not force consumers into synchronous migrations. In Substrate, the `misc_frozen` field of `AccountData` was replaces with `flags`. However, both fields have the same size which removes the requirement of a sync migration: https://github.com/paritytech/substrate/pull/12951/files#r1140113642

runtime/common/src/migrations/balances.rs
cdamian: After running the above, using the live centrifuge chain, the total number of account keys processed was - 102015.

cc @wischli

wischli: I suppose that number reflects the decodable acocunts? I'd be interested in the accounts with `reserved` and/or `frozen` balance.

runtime/common/src/migrations/balances.rs
cdamian: Yup, just thought that it might be a good idea to have similar checks like we do for the normal accounts. No actual migrations taking place, just us having some checks before the upgrade itself. WDYT?

runtime/common/src/migrations/balances.rs
cdamian: No migration to be applied here from what I understand.

runtime/common/src/migrations/balances.rs
cdamian: We can add those as well, I'm concerned with the accounts that fail to decode into the old structure, I wasn't expecting that tbh.

runtime/common/src/migrations/balances.rs
mustermeiszer: I think that is a good idea!

runtime/common/src/migrations/balances.rs
mustermeiszer: @wischli we just have these here in order to verify our storage prior to upgrading. 

@cdamian already found some corrupted storage IIRC, and we wanted to 100% sure that the lazy migration parity is using actually works.

runtime/common/src/migrations/balances.rs
```diff
@@ -0,0 +1,269 @@
...
+		Ok(Vec::new())
+	}
+
+	fn on_runtime_upgrade() -> Weight {
+		// !!! REMOVE TEST ACCOUNT DATA !!!
+		//
+		// TEST ACCOUNT DATA - START
+		let test_account_data = get_test_account_data::<T>();
+
+		for (account_id, account_data) in test_account_data {
```
cdamian, line 145: @mustermeiszer @wischli do you think this should suffice or should we add more?

wischli: I think that suffices. Thanks a lot!

wischli: Note: `on_runtime_upgrade` already uses the new WASM sucht that we cannot write the old `AccountInfo`. Upgraded the test data to set new `AccountInfo` despite having a bit less purpose.

runtime/common/src/migrations/balances.rs
```diff
@@ -0,0 +1,301 @@
...
+	let mut total_reserved = 0;
+	let mut total_frozen = 0;
+	let mut total_misc_frozen = 0;
+	let mut total_flags = 0;
+
+	let prefix = account_prefix.to_vec();
+	let mut previous_key = prefix.clone();
+
+	while let Some(next) = sp_io::storage::next_key(&previous_key) {
+		if !next.starts_with(&prefix) {
```
cdamian, line 145: Great catch! I also had this initially but removed it since I was trying to use the `pallet_balances::Account` prefix and wasn't getting any results... fun times.

runtime/common/src/migrations/balances.rs
cdamian: I was just setting the storage with the raw encoded value of the old account data. How does the WASM version affect this?

wischli: I don't have an answer for you but none of the accounts set during migration could be decoded whereas pre-existing ones with `reserved` or `misc_frozen` could. So it didn't simulate a real scenario.

wischli: On that note: Putting the raw encoded data for the new AccountInfo also led to decoding issues. I don't have capacity to look into this. Writing via `pallet_balances::Account` worked and was sufficient to me given that this is just a testing migration.

mustermeiszer (commented):
Can not approve but looks good!

wischli (approved):
Proxy-approving for @mustermeiszer

PR #1608, opened 2023-11-15
feat: add InvestmentPortfolio Runtime API

### LINKED ISSUE #1589: Investments: RuntimeApi for redeem and collected tokens
### Description

*Output from a conversation with @onnovisser*:

Apps needs an improvement in Investment API to obtain the following:
   - remaining redeem tokens for the whole portfolio on an account
   - collectable invest tokens for the whole portfolio of an account

wischli:
# Description

Fixes #1589

## Changes and Descriptions

* Adds account based `InvestmenPortfolio` Runtime API 
* Removes existent Runtime API which has not been used by Apps
* Removes `InvestmentsPortfolio` trait due to not being used

## TODO

- [x] Integration tests
- [ ] Ensure integrity with Apps

# Checklist:


libs/traits/src/investments.rs
```diff
@@ -219,26 +219,6 @@ pub trait InvestmentAccountant<AccountId> {
 	) -> Result<(), Self::Error>;
 }
 
-/// Trait to handle Investment Portfolios for accounts
-pub trait InvestmentsPortfolio<Account> {
```
wischli, line 223: NOTE: Removed because unused.

libs/types/src/investments.rs
```diff
@@ -239,3 +239,59 @@ pub struct ExecutedForeignCollect<Balance, Currency> {
...
+pub struct InvestmentPortfolio<Balance> {
+	/// The unprocessed invest order amount in pool currency
+	pub pending_invest_currency: Balance,
+	/// The amount of tranche tokens which can be collected for an invest order
+	pub claimable_tranche_tokens: Balance,
+	/// The amount of tranche tokens which can be transferred
+	pub free_tranche_tokens: Balance,
+	/// The amount of tranche tokens which can not be used at all and get
+	/// slashed
+	pub reserved_tranche_tokens: Balance,
```
wischli, line 256: NOTE: Decided against supporting `frozen` as we don't support frozen fungibles in restricted tokens yet and are not freezing anywhere in the codebase.

runtime/altair/src/lib.rs
```diff
@@ -2164,9 +2165,9 @@ impl_runtime_apis! {
 
 
 	// Investment Runtime APIs
-	impl runtime_common::apis::InvestmentsApi<Block, AccountId, TrancheCurrency, CurrencyId, PoolId, Balance> for Runtime {
-		fn investment_portfolio(account_id: AccountId) -> Option<Vec<(PoolId, CurrencyId, TrancheCurrency, Balance)>> {
-			runtime_common::investment_portfolios::get_portfolios::<Runtime, AccountId, TrancheId, Investments, TrancheCurrency, CurrencyId, PoolId, Balance>(account_id)
+	impl runtime_common::apis::InvestmentsApi<Block, AccountId, TrancheCurrency, InvestmentPortfolio<Balance>> for Runtime {
+		fn investment_portfolio(account_id: AccountId) -> Vec<(TrancheCurrency, InvestmentPortfolio<Balance>)> {
+			runtime_common::investment_portfolios::get_account_portfolio::<Runtime, PoolSystem>(account_id)
```
wischli, line 2170: NOTE: Maybe we should bump the version to V2 but Apps has not used this API so far.

NunoAlexandre (approved):
Looks very nice, thanks for adding coverage in the integration tests and useful notes ✨

libs/types/src/investments.rs
mustermeiszer: Didn't we want to add the `CurrencyId` of the actual `pool_currency` here as a field?

libs/types/src/investments.rs
```diff
@@ -239,3 +239,59 @@ pub struct ExecutedForeignCollect<Balance, Currency> {
...
+
+pub struct InvestmentPortfolio<Balance> {
+	/// The unprocessed invest order amount in pool currency
+	pub pending_invest_currency: Balance,
+	/// The amount of tranche tokens which can be collected for an invest order
+	pub claimable_tranche_tokens: Balance,
+	/// The amount of tranche tokens which can be transferred
+	pub free_tranche_tokens: Balance,
+	/// The amount of tranche tokens which can not be used at all and get
+	/// slashed
```
mustermeiszer, line 255: I think the `and get slashed` is confusing. Wouldn't it be `and could can get clashed`.

runtime/altair/src/lib.rs
mustermeiszer: Then I would just overwrite as you did.

runtime/common/src/lib.rs
```diff
@@ -373,50 +372,136 @@ pub mod changes {
...
+			let amount = pallet_investments::RedeemOrders::<T>::get(&investor, invest_id)
+				.map(|order| order.amount())
+				.unwrap_or_default();
+			let balance_after: Balance = pool_currency
+				.map(|p_currency| {
+					orml_tokens::Accounts::<T>::get(&investor, p_currency)
+						.free
+						.into()
+				})
+				.unwrap_or_default();
```
mustermeiszer, line 487: I think we should not make this dependent on `orml_tokens` Config. I would rather use the `fungibles::Inspect` and use restricted-tokens as the input in the runtimes.

libs/types/src/investments.rs
wischli: Yeah the wording is suboptimal as it the _can_ is implicitly connected to _can not be used at all_ as well as _(can) get slashed_. I will improve this!

libs/types/src/investments.rs
wischli: IMO it's impossible to map free `pool_currency` to an `InvestmentId` because
* We will have multiple pools with the same `pool_currency`
* Once it is `free`, the root cause for that is unknown without introducing additional storage

mustermeiszer: Ah sorry. I meant, to add this in order for the UI to show that they user has pending amount of `PoolCurrency` investing. So the UI know what currencie to display. 

Does this make more sense?

runtime/common/src/lib.rs
mustermeiszer: true! Thanks!

libs/types/src/investments.rs
wischli: Sorry, I cant follow. Could you elaborate @mustermeiszer ? We already have a field for pending pool currency. Any pool currency that is not related to an investment id shouldnt be part of this portfolio API per definition (no key value mapping).

mustermeiszer: > `fn investment_portfolio(account_id: AccountId) -> Vec<(InvestmentId, InvestmentPortfolio)>`

as per this API the UI has no knowledge what is the `pool_currency`, so we should give it this information in the structure.

wischli: Now I get it, you are talking about the _currency id_ while I was still in the _amounts_ space. I will add it then to have a full portfolio without any need for further queries. On that note: The investment id key contains the pool id such that another query suffices to read the pool currency.

libs/types/src/investments.rs
wischli: [60e9add](https://github.com/centrifuge/centrifuge-chain/pull/1608/commits/60e9add23d13fb680cef7fc8f9522bac59e27765)

libs/types/src/investments.rs
wischli: [60e9add](https://github.com/centrifuge/centrifuge-chain/pull/1608/commits/60e9add23d13fb680cef7fc8f9522bac59e27765)

runtime/common/src/lib.rs
```diff
@@ -390,59 +394,88 @@ pub mod investment_portfolios {
...
-			InvestmentPortfolio<Balance>,
+			InvestmentPortfolio<Balance, CurrencyId>,
 		>::new();
 
 		// Denote current tranche token balances before dry running collecting
 		orml_tokens::Accounts::<T>::iter_key_prefix(&investor).for_each(|currency| {
 			if let CurrencyId::Tranche(pool_id, tranche_id) = CurrencyId::from(currency) {
-				let balance = orml_tokens::Accounts::<T>::get(&investor, currency);
+				let pool_currency = PoolInspector::currency_for(pool_id)
+					.expect("Pool must exist; qed")
```
wischli, line 431: I think we can safely expect here. If we want to be super duper safe, I can wrap the rest of the inner scopes into a `.map`. WDYT @mustermeiszer ?

mustermeiszer: Yeah, I would also say so. I also think zo remember that api calls are wrapped in a panic catching-something, I will double check that. 

But think the expect is fine!

mustermeiszer (approved):
Nice and clean!

Issue #1380, opened 2023-06-02
Internal dependencies organization


lemunozm:
# Description
We don't have well-defined rules about what things should be where. Here I propose some changes and rules that I think can improve the repo organization and get rid of dependency problems in the future.

Feel free to discuss!

## Current state
Our repo has the following dependencies, and we have 3 main layers. We can have *n* `chains` using the same `runtime`, and we can have a `runtime` using *n* `pallets`.

![image](https://github.com/centrifuge/centrifuge-chain/assets/15687891/a350a5a6-005b-4afe-9234-dfdc7953963c)

Ideally, we would want to remove those red upward dependencies.
Down/right directions in this diagram mean "more generic", so those upper lines involve a lot of knowledge from a part that should not be known about. We only want to remain dependencies to "more generic" and avoid that generic things know about specific things.

## Desired
By removing the red lines, redefining the purpose of each crate, and adding a new one, we can get a down/right diagram:

These arrows would be transitive. It means i.e. that a *runtime* could make use of `cfg-utils`.

![image](https://github.com/centrifuge/centrifuge-chain/assets/15687891/be81d8b2-c62c-46bc-a3b2-37a0688c2155)

### Purpose of each crate:
From more generic (widely purpose) to less generic (centrifuge-related purpose):

- `cfg-utils`: Common utilities without any specific behavior associated. 
    - i.e. `Adjustement` type or `SECONDS_PER_MINUTE`, `DAYS` constant.
- `cfg-traits`: Common interfaces to all runtimes. It implies traits and also could imply types tied to those traits. But it should not use `cfg-types`, because they are specific types for configuring in the runtimes.
    - i.e. `Permission`, `InterestAccrual`, or `PoolInvest`.
- `cfg-types`: Should know about centrifuge-related types. These are types used in runtimes to configure pallets, but should not be used direclty in the pallets. They make sense for our use cases but not for others who can use our pallets. These types, therefore, should not know about pallets or runtimes. And should not be types used by `cfg-traits`.
    - i.e. complex types: `CurrencyId`, `FeeKey`, `Location` 
    - i.e. simple types: `PoolId`, `AccountId`, `Balance`. 
    - i.e. constants: `MAXIMUM_BLOCK_WEIGHT`, `CFG`(?), `SLOT_DURATION`, `TREASURY_FEE_RATIO`
[...]

PR #1609, opened 2023-11-16
Runtimes workspace dependencies & cleaning

### LINKED ISSUE #1582: Reexport common dependencies from runtime-common
### Problem
For each pallet/runtime dependency used, we need to add such dependency to:
- `development-runtime`
- `altair-runtime`
- `centrifuge-runtime`
- `integration-tests`
- Sometimes to `runtime-common`

And configure `runtime-benchmarks` and `try-runtime` features in all crates. 

Avoiding/mixing some feature names will lead to an annoying compile error that could take time to figure out what's going on.

### Solution
We can simplify all these dependency boilerplate by adding all pallets to `runtime-common` and allowing `runtime-common` to reexport them. So, in `runtime-common` we can have:

```rust
pub mod reexport {
    pub mod pallets {
        pub use pallet_timestamp;
        // Other pallets
    }
}
```

From runtimes, on top of the file:
```rust
use runtime_common::reexport::pallets::*;
```

And the code would remain inmutable for runtimes, but with access to all pallets without modifying `Cargo.toml` files.

In the future, this would avoid `nix` hash breaks.

lemunozm:
# Description

Refactorizations and cleanings in the runtime part.

The first step to get this: https://github.com/centrifuge/centrifuge-chain/issues/1380

Fixes ~~#1582~~

## Changes and Descriptions
- [x] Reorganize the fees section in `runtime-common`
- [x] Remove `scripts/runtime-weight-template.hbs` because now we use the default from substrate.
- [x] ~~`runtime-common` export a list with all pallets used in our runtimes, to avoid importing them if all of our runtimes.~~
  - There is an issue with the autogenerated weights where the default template is not valid if we reimport the pallets, we would need to maintain a template by us (the one we currently have is not used). **Do we want to maintain the template to get this reexporting list?**
    - Finally, it seems more painful to have an exporting list than to copy the pallet list.
- [x] Using workspace dependencies in `runtime-common`
- [x] Using workspace dependencies in `development`
- [x] Using workspace dependencies in `altair`
- [x] Using workspace dependencies in `centrifuge`
- [x] Using workspace dependencies in `integration-tests`
- [ ] ~~Factorize types from `development` into `runtime-common`~~ => Other PR
- [ ] ~~Factorize types from `altair` into `runtime-common`~~=> Other PR
- [ ] ~~Factorize types from `centrifuge` into `runtime-common`~~ => Other PR


lemunozm:
Currently `cargo` complains (with a warning) about workspace inheritance with renaming packages (see the [issue](https://github.com/rust-lang/cargo/issues/12546)). We can:
- Life with the warning until it be (hopefully) solved.
- Change `codec` to `parity_scale_codec` everywhere in our database

Cargo.toml
```diff
@@ -158,7 +352,7 @@ runtime-common = { path = "runtime/common" }
 frame-benchmarking = { git = "https://github.com/paritytech/substrate", optional = true, default-features = false, branch = "polkadot-v0.9.43" }
 frame-system-benchmarking = { git = "https://github.com/paritytech/substrate", optional = true, default-features = false, branch = "polkadot-v0.9.43" }
 # integration testing
-runtime-integration-tests = { path = "runtime/integration-tests", optional = true, default-features = false }
+#runtime-integration-tests = { path = "runtime/integration-tests", optional = true, default-features = false }
```
mustermeiszer, line 355: ???

wischli, line 355: Artifact?

Cargo.toml
```diff
@@ -222,7 +416,7 @@ std = [
   "pallet-pool-system/std",
   "polkadot-primitives/std",
   "runtime-common/std",
-  "runtime-integration-tests/std",
+  #"runtime-integration-tests/std",
```
wischli, line 419: Artifact? Plus the other commented out features of `runtime-integration-tests` below

runtime/integration-tests/src/utils/extrinsics.rs
```diff
@@ -151,8 +151,6 @@ fn signed_extra_relay(nonce: RelayIndex) -> RelaySignedExtra {
 		frame_system::CheckNonce::<RelayRuntime>::from(nonce),
 		frame_system::CheckWeight::<RelayRuntime>::new(),
 		pallet_transaction_payment::ChargeTransactionPayment::<RelayRuntime>::from(0),
-		#[cfg(not(feature = "runtime-development"))]
-		polkadot_runtime_common::claims::PrevalidateAttests::<RelayRuntime>::new(),
```
mustermeiszer, line 155: Why the removal?

runtime/altair/Cargo.toml
```diff
@@ -1,179 +1,213 @@
...
-  "cumulus-pallet-session-benchmarking/std",
-  "cumulus-pallet-xcm/std",
-  "cumulus-pallet-xcmp-queue/std",
+  "getrandom/std",
+  "hex/std",
+  "scale-info/std",
+  "serde/std",
+  "log/std",
+
+  # Substrate related
```
wischli, line 156: 🚀

runtime/altair/Cargo.toml
```diff
@@ -382,29 +425,28 @@ try-runtime = [
...
-  "runtime-common/try-runtime",
-  "sp-runtime/try-runtime",
 ]
 
 # A feature that should be enabled when the runtime should be build for on-chain
 # deployment. This will disable stuff that shouldn't be part of the on-chain wasm
 # to make it smaller like logging for example.
 on-chain-release-build = [
   "sp-api/disable-logging",
+  "runtime-common/on-chain-release-build",
```
wischli, line 449: Can you give some details about this addition?

scripts/runtime-weight-template.hbs
```diff
@@ -1,47 +0,0 @@
-//! Autogenerated weights for {{pallet}}
-//!
-//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION {{version}}
-//! DATE: {{date}}, STEPS: `{{cmd.steps}}`, REPEAT: {{cmd.repeat}}, LOW RANGE: `{{cmd.lowest_range_values}}`, HIGH RANGE: `{{cmd.highest_range_values}}`
-//! EXECUTION: {{cmd.execution}}, WASM-EXECUTION: {{cmd.wasm_execution}}, CHAIN: {{cmd.chain}}, DB CACHE: {{cmd.db_cache}}
```
wischli, line 5: Why can we remove the bench template?

wischli (commented):
Thanks for cleaning up and making our future lives much easier! I have a few questions, overall LGTM.

Cargo.toml
lemunozm: Why were we linking integration tests as a dependency of our node before?

runtime/integration-tests/src/utils/extrinsics.rs
lemunozm: Nongeneric tests are always run in CI under `runtime-development,` so this implies that that line is never compiled. In the case of other runtimes, we have now the generic part which should take care of that. Actually, we no longer have `runtime-x` features.

Please correct me if I'm missing something here, but CI will always tests against the development.

Cargo.toml
```diff
@@ -70,9 +70,203 @@ members = [
   "runtime/integration-tests",
 ]
 
+[workspace.package]
+authors = ["Centrifuge <admin@centrifuge.io>"]
+edition = "2021"
+license = "LGPL-3.0"
+homepage = "https://centrifuge.io/"
+repository = "https://github.com/centrifuge/centrifuge-chain"
+documentation = "https://github.com/centrifuge/centrifuge-chain"
```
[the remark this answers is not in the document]
lemunozm: Right! I thought I copied the correct link 🤦🏻‍♂️

Cargo.toml
lemunozm: Answer above to Frederik, we do not want integration tests as a dependency of our centrifuge node, right?

Cargo.toml
lemunozm: related to the previous answer

runtime/altair/Cargo.toml
lemunozm: Yes, I understand that if we compile altair with `on-chain-release-build` we would want the runtime common used to have that related feature also.

scripts/runtime-weight-template.hbs
lemunozm: We're not using it, if we do not put any template we're using the default. That it's how it's working right now.

Cargo.toml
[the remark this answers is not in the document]
lemunozm: Done!

Cargo.toml
mustermeiszer: I thought we needed it for some features. Fine then

NunoAlexandre: @lemunozm yeah good point but then we should remove these lines all together.

mustermeiszer (approved):

PR #1610, opened 2023-11-20
LiquidityPools: Add `restriction_set` field to `AddTranche` message

NunoAlexandre:
# Description

When adding a tranche token to a LP-compatible domain, we now need to specify which implementation of the `RestrictionManager` we want to use on a per-tranche-token basis. For that, we add this new field that serves as an identifier that will be picked up on the destination accordingly.

# Checklist:


hieronx (commented):
Should have someone else also review, but logically looks good to me!

wischli (approved):
LGTM! I am a bit hesitant on not adding it to the `add_tranche` extrinsic parameter set but it wouldn't make sense as long as its fixed to one.

PR #1615, opened 2023-11-21
Fix: Transfer debt

mustermeiszer:
# Description
Issuers need to have the possibility to transfer debt between loans without restrictions. In the future we might add restrictions back, but based on a per pool setup, so investors know for each pool what to expect.

Fixes issues that our issuers are having.

Furthermore, this change allows to deterministically derive a `ChangeId` for a given proposed change, as we exclude the time of submission from the hash. This means, that equal changes will overwrite past onces, but that is fine as it just means we are extending the time of execution for that change.

## Changes and Descriptions
* `ChangeId` now `Blake256(Change)`
* `Change::TransferDebt` has no restrictions

# Checklist:

### DIFF
--- a/pallets/pool-system/src/impls.rs
+++ b/pallets/pool-system/src/impls.rs
@@ -401,14 +401,13 @@ impl<T: Config> ChangeGuard for Pallet<T> {
 	type PoolId = T::PoolId;
 
 	fn note(pool_id: Self::PoolId, change: Self::Change) -> Result<Self::ChangeId, DispatchError> {
+		// NOTE: Essentially, this key-generation allows to override previously
+		//       submitted changes, if they are identical.
+		let change_id: Self::ChangeId = T::Hashing::hash(&change.encode());
 		let noted_change = NotedPoolChange {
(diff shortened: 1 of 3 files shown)


hieronx (approved):
@lemunozm should approve but LGTM!

pallets/pool-system/src/impls.rs
```diff
@@ -406,7 +406,9 @@ impl<T: Config> ChangeGuard for Pallet<T> {
 			change,
 		};
 
-		let change_id: Self::ChangeId = T::Hashing::hash(&noted_change.encode());
+		// NOTE: Essentially, this key-generation allows to override previously
+		//       submitted changes, if they are identical.
+		let change_id: Self::ChangeId = T::Hashing::hash(&change.encode());
```
lemunozm, line 411: I like it!

lemunozm (dismissed):

lemunozm (dismissed):

hieronx (approved):

hieronx (dismissed):

lemunozm (approved):

PR #1616, opened 2023-11-21
Fix: Loans adjustments

lemunozm:
# Description

A couple of minor tasks before the upgrading process

- [x] Update `MaxFeedValues` to `500`
- [x] Borrow proxy type to include `transfer_debt` and token transfer capabilities


mustermeiszer:
@lemunozm I will take of this in the restricted transfers PR as we also need to add transfer capcabilities to the borrower proxy, if this is fine for you?

lemunozm:
Sure!

runtime/altair/src/lib.rs
```diff
@@ -490,6 +490,7 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
 					RuntimeCall::Loans(pallet_loans::Call::propose_write_off_policy{..}) |
 					RuntimeCall::Loans(pallet_loans::Call::apply_write_off_policy{..}) |
 					RuntimeCall::Loans(pallet_loans::Call::update_portfolio_valuation{..}) |
+                    RuntimeCall::Loans(pallet_loans::Call::apply_transfer_debt { .. }) |
```
mustermeiszer, line 493: Missing propose here

mustermeiszer (commented):
We are missing a propose in the `NonTransfer` proxz type but otherwise fine.

runtime/altair/src/lib.rs
lemunozm: I was not sure about how to interpret that. Could `propose()` imply a transfer if later somebody apply it? Is `transfer_debt()` taken into account as something that transfers?

mustermeiszer: By definricion transfer debt does juts transfer internally, right? So it think it is fine. 

If you think differently, then I would remove also the apply one.

lemunozm: If transferring the debt locally counts as a transfer, then we should not add it
If it's not a transfer, we should add it

mustermeiszer: He is off. Let's go with adding it!

transferring is just concerned for the acting account to move their funds which will not be the case here.

lemunozm: Make sense!

mustermeiszer (approved):

runtime/altair/src/lib.rs
hieronx: Agreed with @mustermeiszer, transferring debt is separate (and much less impactful from security perspective) from transferring stablecoins.

PR #1621, opened 2023-11-23
Feat/restricted transfers closure

mustermeiszer:
# Description
This changes closes the restricted transfers features. This will allow accounts to set allowances for their accounts and currencies to where funds are allowed to be sent to (not for CFG though, as we can not restrict `pallet-balances`). 

This is an opt-in restriction meaning
* No allowance set &rarr; No restrictions
* Allowance set &rarr; Only allowed receiver can receive funds from that account (on a per currency base)

NOTE: 
* `orml-xtokens` has NO storage and `pallet-restricted-xtokens` mimics the API 1-to-1, hence we can replace the location of `orml-xtokens` with `pallet-restricted-xtokens` in the `construct_runtime!`-macro

## Changes and Descriptions
* [x] Wrap `orml-xtokens` in `pallet-restricted-xtokens`
* [x] Adapt `pallet-transfer-allowlist` to reject native currency allowances
* [x] Add `pallet-transfer-allowlist` to dev
* [x] Add `pallet-transfer-allowlist` to altair
* [x] Add `pallet-transfer-allowlist` to centrifuge
* [x] Add `pallet-restricted-xtokens` to dev
* [x] Add `pallet-restricted-xtokens` to altair
* [x] Add `pallet-restricted-xtokens` to centrifuge
* [x] Implement common filter to be used in `pallet-restricted-xtokens` and `pallet-restricted-tokens`
* [x] Adds a new `ProxyType::Transfer` that can be used to extend otherwise non-transfer proxies with this ability.

# Checklist:


pallets/restricted-xtokens/Cargo.toml
```diff
@@ -0,0 +1,91 @@
+[package]
+name = "paller-restricted-xtokens"
```
NunoAlexandre, line 2:
```suggestion
name = "pallet-restricted-xtokens"
```

:)

pallets/restricted-xtokens/src/mock/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+		a.resize(32, 0);
+		let mut a1: Vec<u8> = "A1".into();
+		a1.resize(32, 0);
+		let mut b: Vec<u8> = "B".into();
+		b.resize(32, 0);
+		let mut b1: Vec<u8> = "B1".into();
+		b1.resize(32, 0);
+		if l == MultiLocation::parent() {
+			return Some(CurrencyId::R);
+		}
```
NunoAlexandre, line 144: I think this is unnecessarily clutter-y. If we would use a `GeneralIndex` (`u128`) that we map for each currency, then it's a lot simpler and we could even define that map at the enum definition level (i.e, `pub enum CurrencyId { R = 0, A = 1, etc`.)

pallets/restricted-xtokens/src/tests.rs
```diff
@@ -0,0 +1,538 @@
...
+		});
+
+		Relay::execute_with(|| {
+			assert_eq!(RelayBalances::free_balance(&para_a_account()), 500);
+			assert_eq!(RelayBalances::free_balance(&RESTRICTED_RECEIVER), 460);
+		});
+	}
+}
+
+mod para_a_to_para_a {
```
NunoAlexandre, line 120: XCM messages can never be from a parachain to itself 🤔

pallets/restricted-xtokens/src/tests.rs
```diff
@@ -0,0 +1,538 @@
...
+
+			assert_eq!(
+				ParaTokens::free_balance(CurrencyId::R, &RESTRICTED_SENDER),
+				1000
+			);
+		});
+	}
+
+	#[test]
+	fn restrict_normal_transfer_b1_currency() {
```
NunoAlexandre, line 188: I will keep reading but I wonder here what a "restricted normal transfer" means

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -257,6 +256,8 @@ pub mod pallet {
 		/// Transfer from sending account and currency not allowed to
 		/// destination
 		NoAllowanceForDestination,
+		/// Native currency can not be restricted with allowances
+		NativeCurrencyNotPossible,
```
NunoAlexandre, line 260:
```suggestion
		NativeCurrencyNotRestrictable,
```

Just a suggestion 🌱

pallets/restricted-xtokens/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+// Centrifuge is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version (see http://www.gnu.org/licenses).
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+//! # Xtokens Module
```
NunoAlexandre, line 13:
```suggestion
//! # Restricted Xtokens Module
```

NunoAlexandre: And the rest of this mod level docs need to be adjusted too 🙏

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -76,32 +84,23 @@ pub mod pallet {
...
-			+ Ord
-			+ TypeInfo
-			+ MaxEncodedLen;
+		type CurrencyId: AssetId + Parameter + Member + Copy + Default;
+
+		// We need this to block restrictions on the native chain currency as
+		// currently it is impossible to place transfer restrictions
+		// on native currencies as we simply have no way of
+		// restricting pallet-balances...
+		type NativeCurrency: Get<Self::CurrencyId>;
```
NunoAlexandre, line 93: I need to look deeper into this pallet so this is more a question for myself but doesn't our `pallet-restricted-transfers` take care of this for us? Why would we need to hope pallet-balances to offer that for us here?

pallets/transfer-allowlist/src/mock.rs
```diff
@@ -45,9 +42,8 @@ frame_support::construct_runtime!(
 		UncheckedExtrinsic = UncheckedExtrinsic,
 	  {
 			Balances: pallet_balances,
-		  Fees: pallet_mock_fees,
 			System: frame_system,
-		  TransferAllowList: transfer_allowlist,
+			  TransferAllowList: transfer_allowlist,
```
NunoAlexandre, line 46:
```suggestion
			TransferAllowList: transfer_allowlist,
```

Unintentional?

NunoAlexandre (commented):
The `restricted-xtokens` pallet looks good 🥳 

I left a couple of comments, suggestions or just questions as a first scan but will give it another look either later or tomorrow morning 👍

pallets/restricted-xtokens/src/mock/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+use crate as restricted_xtokens;
+
+pub mod para;
+pub mod relay;
+
+pub const RESTRICTED_SENDER: AccountId32 = AccountId32::new([0u8; 32]);
+pub const BOB: AccountId32 = AccountId32::new([1u8; 32]);
+pub const RESTRICTED_RECEIVER: AccountId32 = AccountId32::new([2u8; 32]);
+
+fn para_a_rreceiver_relay() -> MultiLocation {
```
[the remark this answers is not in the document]
mustermeiszer: restricted receiver, fence the r. But can also rename if wanted

libs/types/src/locations.rs
```diff
@@ -18,7 +18,7 @@ use sp_runtime::{
 	traits::{BlakeTwo256, Hash},
 	AccountId32,
 };
-use xcm::{latest::MultiLocation, VersionedMultiLocation};
+use xcm::{v3::MultiLocation, VersionedMultiLocation};
```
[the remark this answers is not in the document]
mustermeiszer: Do not want to miss v4, which implicitly would make v3 not existing. So specific versioning is always good for xcm imo

pallets/restricted-xtokens/src/mock/mod.rs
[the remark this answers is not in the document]
NunoAlexandre: ah sorry, I oversaw that!

libs/types/src/locations.rs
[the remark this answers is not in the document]
NunoAlexandre: yeah that's fair 👍

pallets/restricted-xtokens/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+			origin: OriginFor<T>,
+			currencies: Vec<(T::CurrencyId, T::Balance)>,
+			fee_item: u32,
+			dest: Box<VersionedMultiLocation>,
+			dest_weight_limit: WeightLimit,
+		) -> DispatchResult {
+			let sender = ensure_signed(origin.clone())?;
+			let destination: MultiLocation = (*dest.clone())
+				.try_into()
+				.map_err(|()| orml_xtokens::Error::<T>::BadVersion)?;
```
lemunozm, line 350: Wondering if using errors from `orml_xtokens` can confuse a future us trying to find an error thinking the error comes from `orml_xtokens` pallet

lemunozm: Maybe it can be nested under `restricted_xtokens` error?

lemunozm (commented):
Just a couple of comments, but looks good to me! 🚀

pallets/transfer-allowlist/src/lib.rs
mustermeiszer: Applied

pallets/transfer-allowlist/src/lib.rs
mustermeiszer: but users can always use `pallet-balances` to transfer CFG and we can not not exposte the `Call` for balances as this breaks all integrations with external parties. Unfortunately...

pallets/transfer-allowlist/src/lib.rs
```diff
@@ -433,10 +436,12 @@ pub mod pallet {
...
 				Some(AllowanceDetails { blocked_at, .. }) if blocked_at < current_block => {
-					T::ReserveCurrency::unreserve(
+					T::ReserveCurrency::release(
+						&T::HoldId::get(),
 						&account_id,
-						T::Fees::fee_value(T::AllowanceFeeKey::get()),
-					);
+						T::Deposit::get(),
+						Precision::BestEffort,
+					)?;
```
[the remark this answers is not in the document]
mustermeiszer: Fees is used in the runtime configuration but did not want to use it here anymore, as this is basically kind of missusing fees.

pallets/restricted-xtokens/src/lib.rs
mustermeiszer: The palllet itself is hard-coupled to `orml-xtokens` and the error will also happen in `orml_xtokens`. But I can of course make a dedicated `BadVersion` for restricted-xtokens.

pallets/transfer-allowlist/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: I think if we want to reconsider what fees is doing and use it as a general purpose pallet for depositing and fees, maybe even keeping track of them I am fine with that and already prposed that. But for now I would go with the simple solution to just query a fixed deposit. in the pallet itself I mean.

pallets/restricted-xtokens/src/tests.rs
mustermeiszer: normal is just the `transfer()` method.

pallets/transfer-allowlist/src/mock.rs
mustermeiszer: Done

pallets/restricted-xtokens/src/lib.rs
mustermeiszer: done

pallets/restricted-xtokens/src/mock/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+			network: None,
+			id: RESTRICTED_RECEIVER.into(),
+		}),
+	)
+}
+
+fn para_a_rreceiver_para_a() -> MultiLocation {
+	MultiLocation::new(
+		0,
+		X1(Junction::AccountId32 {
```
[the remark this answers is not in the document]
mustermeiszer: Why would I do that? The location is generated from the point of view of `para_a`, right?

pallets/restricted-xtokens/src/mock/mod.rs
mustermeiszer: I copy pasta that one from acala. While I agree I think for the given 5 enum variants this logic is fine and I do not expect to change it often in the futre. Wdyt?

pallets/transfer-allowlist/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I fully agree with you. I think the current fees API is not good. I've created a small PR with some improvements.

pallets/restricted-xtokens/src/lib.rs
lemunozm: Maybe something like: `restricted_xtokens::Error::XTokens(orml_xtokens::Error)` could work?

If not, I would say to make a dedicated version

mustermeiszer: Of course, but that is not the error. It would be a dedicated copy error of the one in xtokens

runtime/development/src/lib.rs
```diff
@@ -1903,8 +1907,9 @@ construct_runtime!(
 		PolkadotXcm: pallet_xcm::{Pallet, Call, Storage, Config, Event<T>, Origin} = 121,
 		CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 122,
 		DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 123,
-		XTokens: orml_xtokens::{Pallet, Storage, Call, Event<T>} = 124,
+		XTokens: pallet_restricted_xtokens::{Pallet, Call} = 124,
 		XcmTransactor: pallet_xcm_transactor::{Pallet, Call, Storage, Event<T>} = 125,
+		OrmlXTokens: orml_xtokens::{Pallet, Event<T>} = 126,
```
wischli, line 1912: Maybe CR: Same as in Altair + Centrifuge runtimes.

wischli (changes requested):
The current extrinsic weights are underestimated. Given that this is a common attack vector, this needs to be fixed. 

Apart from that, I also believe that we cannot simply overwrite the `OrmlXTokens` pallet index with `XTokens` because we drop the storage. 

There a few more nit comments which you can ignore. I have to admit that I only skimmed through the mocks and unit tests.

pallets/restricted-xtokens/src/lib.rs
lemunozm: Could we do the same for `AssetIndexNonExistent`? To align it

pallets/restricted-xtokens/Cargo.toml
```diff
@@ -0,0 +1,91 @@
...
+
+# orml
+orml-xtokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.43" }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.43" }
+
+# polkadot
+xcm = { git = "https://github.com/paritytech/polkadot", default-features = false , branch = "release-v0.9.43" }
+
+# local
+cfg-traits = { path = "../../libs/traits", default-features = false }
```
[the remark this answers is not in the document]
lemunozm: I'm following the rule of adding everything without default features

mustermeiszer:
> Apart from that, I also believe that we cannot simply overwrite the OrmlXTokens pallet index with XTokens because we drop the storage.

There is no storage, or am I missing something?

runtime/common/src/lib.rs
```diff
@@ -24,14 +24,38 @@ pub mod evm;
...
+impl<T: cfg_traits::fees::Fees<Balance = Balance, FeeKey = FeeKey>> Get<Balance>
+	for AllowanceDeposit<T>
+{
+	fn get() -> Balance {
+		T::fee_value(FeeKey::AllowanceCreation)
+	}
+}
+
+/// To be used with pallet-balances accross runtimes
+pub type HoldIdentifier = ();
```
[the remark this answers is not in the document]
mustermeiszer: Yeah, forgot to change that. Will do.

runtime/centrifuge/src/lib.rs
```diff
@@ -1994,6 +2015,7 @@ mod __runtime_api_use {
 
 #[cfg(not(feature = "disable-runtime-api"))]
 use __runtime_api_use::*;
+use cfg_types::locations::Location;
```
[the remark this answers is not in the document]
mustermeiszer: Clion. will see if can remove

pallets/restricted-xtokens/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+		/// below requirements, the execution may fail and assets wouldn't be
+		/// received.
+		///
+		/// It's a no-op if any error on local XCM execution or message sending.
+		/// Note sending assets out per se doesn't guarantee they would be
+		/// received. Receiving depends on if the XCM message could be delivered
+		/// by the network, and if the receiving chain would handle
+		/// messages correctly.
+		#[pallet::call_index(0)]
+		#[pallet::weight(orml_xtokens::XtokensWeight::< T >::weight_of_transfer(currency_id.clone(), * amount, dest))]
```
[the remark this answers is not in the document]
mustermeiszer: Right. Will add.

runtime/centrifuge/src/lib.rs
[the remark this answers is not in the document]
wischli: Since I switched to RustRover, I am running into similar issues. Still incredible time saving with the auto imports.

wischli:
> > Apart from that, I also believe that we cannot simply overwrite the OrmlXTokens pallet index with XTokens because we drop the storage.
> 
> There is no storage, or am I missing something?

You are right. Then I suppose we can keep your change. Will close the requests.

pallets/restricted-xtokens/src/tests.rs
```diff
@@ -0,0 +1,538 @@
...
+		.iter()
+		.for_each(|r| println!("> {:?}", r.event));
+}
+
+mod para_a_to_relay {
+	use super::*;
+
+	#[test]
+	fn restrict_normal_transfer_relay_currency() {
+		TestNet::reset();
```
cdamian, line 43: These are tests that are using the `xcm-simulator` from what I can tell. Given our efforts to clean up the integration test setup, I'm not sure how I feel about having these tests here in the first place, and re-using `xcm-simulator`.,,

mustermeiszer: Agreed, but I also not so keen on bringing fudge in here for unit-testing. I can remove all tests and just keep integration testm wdyt?

cdamian: Yup, I think that's a good move.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -1178,6 +1185,62 @@ mod centrifuge {
 			));
 		}
 
+		pub fn register_lp_eth_usdc<T: Runtime>() {
```
cdamian, line 1188: I think there's another better way of dealing with this, as Luis noted in a previous PR - https://github.com/centrifuge/centrifuge-chain/blob/main/runtime/integration-tests/src/generic/utils/currency.rs

pallets/restricted-xtokens/src/tests.rs
cdamian: I think we can just add some of the missing error cases to the integration tests and just remove the ones here.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
mustermeiszer: Not sure how this will work. There is no location that I need. Can you be more precise

pallets/restricted-xtokens/src/tests.rs
mustermeiszer: Not sure how possible this is given the overhead. I would be fine with having just the restrict and allow calls there. Or what cases do you see missing?

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -1616,6 +1745,225 @@ mod centrifuge {
...
+				pallet_transfer_allowlist::Pallet::<T>::add_transfer_allowance(
+					RawOrigin::Signed(account.into()).into(),
+					asset,
+					location
+				)
+			);
+		}
+
+		fn restrict_lp_eth_usdc_transfer<T: Runtime>() {
+			todo!()
```
cdamian, line 1783: I will add the missing ones here.

mustermeiszer (commented):
Minus nits from william I think it is good to go. We should address these and then get this in.

pallets/restricted-xtokens/Cargo.toml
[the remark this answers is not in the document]
cdamian: Changed.

pallets/restricted-xtokens/src/lib.rs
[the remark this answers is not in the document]
cdamian: Updated this, please check.

wischli (approved):
Thanks for adhering to all my nits and fixing the weight! Looks really good, great work @mustermeiszer and @cdamian!

runtime/altair/src/lib.rs
```diff
@@ -581,6 +582,12 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
 				RuntimeCall::Permissions(pallet_permissions::Call::add { .. })
 					| RuntimeCall::Permissions(pallet_permissions::Call::remove { .. })
 			),
+			ProxyType::Transfer => {
+				matches!(
+					c,
+					RuntimeCall::XTokens(..) | RuntimeCall::Balances(..) | RuntimeCall::Tokens(..)
```
mustermeiszer, line 588: @cdamian sit, I forgot something. We should add two more here

`RuntimeCall::LiquidityPools(pallet_liquidituy_pool::Call::Transfer{..} | pallet_liquidituy_pool::Call::TransferTrancheTokens{..})`

PR #1623, opened 2023-11-24
Fees API Improvement

lemunozm:
# Description

## Problem

The current API provided by the traits around `pallet_fees` to handle fees has two negative points:
1. In order to use it in a pallet you need to use at least 2 associated types in the `Config` trait. i.e:
  ```rust
  /// Entity used to pay fees
  type Fees: Fees<AccountId = Self::AccountId>;

  /// Key used to get the fee value.
  type ActionAFeeKey: Get<<Self::Fees as Fees>::FeeKey>;
  ```
2. The knowledge of how to consume that fee is in the pallet when the pallet should not be aware of it. This should be a runtime concern.
  ```rust
  /// Inside the pallet we decide when we should not do that
  T::Fees::fee_to_treasury(&who, Fee::Key(T::ActionAFeeKey))?;
  // or
  T::Fees::fee_to_author(&who, Fee::Balance(T::ActionAFeeValue))?;
  ```
  
## Solution

We can fix these 2 points by coding the whole fee action in different types (see PR code) under a new trait `PayFee`.

Pallet:
```rust
// In Config trait:
type ActionAPayFee: PayFee<T::Account>;

// When use it:
T::ActionAPayFee::pay_for(&who)?;
```

Runtime:
```rust
parameter_types! {
    pub const ActionAFee: Fee = Fee::Balance(10 * CFG);
}

impl pallet_example::Config for Runtime {
    // We set in one line:
    //   - How the fee is handled
    //   - The value of the fee
    type ActionAPayFee = FeeToTreasury<Fees, ActionAFee>;
    // ...
}
```

### DIFF
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -435,6 +435,13 @@ pub mod fees {
 			fee: Fee<Self::Balance, Self::FeeKey>,
 		) -> DispatchResult;
 	}
+
+	/// Trait to pay fees
+	/// This trait can be used by pallet to just pay fees without worring about
+	/// the value or where the fee goes.
+	pub trait PayFee<AccountId> {
+		fn pay_for(who: &AccountId) -> DispatchResult;
+	}
 }
 
 /// Trait to determine whether a sending account and currency have a
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -440,7 +440,7 @@ pub mod fees {
 	/// This trait can be used by pallet to just pay fees without worring about
 	/// the value or where the fee goes.
 	pub trait PayFee<AccountId> {
-		fn pay_for(who: &AccountId) -> DispatchResult;
+		fn pay(who: &AccountId) -> DispatchResult;
 	}
 }
 
--- a/libs/types/src/fee_keys.rs
+++ b/libs/types/src/fee_keys.rs
@@ -54,3 +54,5 @@ impl Default for FeeKey {
 		FeeKey::AnchorsCommit
 	}
 }
+
+pub type Fee = cfg_traits::fees::Fee<cfg_primitives::Balance, FeeKey>;
--- a/runtime/common/src/fees.rs
+++ b/runtime/common/src/fees.rs
@@ -1,9 +1,13 @@
 use cfg_primitives::{
 	constants::{CENTI_CFG, TREASURY_FEE_RATIO},
 	types::Balance,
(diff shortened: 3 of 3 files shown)


lemunozm:
This PR is completed and ready for review

cdamian (approved):

NunoAlexandre (dismissed):
Sweet 🍯 👌

libs/traits/src/lib.rs
```diff
@@ -435,6 +435,13 @@ pub mod fees {
 			fee: Fee<Self::Balance, Self::FeeKey>,
 		) -> DispatchResult;
 	}
+
+	/// Trait to pay fees
+	/// This trait can be used by pallet to just pay fees without worring about
+	/// the value or where the fee goes.
+	pub trait PayFee<AccountId> {
+		fn pay_for(who: &AccountId) -> DispatchResult;
```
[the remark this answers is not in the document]
lemunozm: Ok, makes sense!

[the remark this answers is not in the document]
lemunozm: Nah, never mind, I've updated the PR with `pay()`

libs/traits/src/lib.rs
```diff
@@ -435,6 +435,13 @@ pub mod fees {
 			fee: Fee<Self::Balance, Self::FeeKey>,
 		) -> DispatchResult;
 	}
+
+	/// Trait to pay fees
+	/// This trait can be used by pallet to just pay fees without worring about
+	/// the value or where the fee goes.
+	pub trait PayFee<AccountId> {
+		fn pay(who: &AccountId) -> DispatchResult;
```
NunoAlexandre, line 443: Given the previous discussion, I would for sure document that `who` is the account that's paying for the fees and not the account that is receiving them.

NunoAlexandre (approved):
One doc request but not blocking 🙏

libs/traits/src/lib.rs
wischli: +1 on that. Would also be great to rename `who` to `payer`.

lemunozm: Good point! Sorry for the "auto-merge", keep in mind for a future PR where I could add this 🧠

lemunozm: Here it goes: https://github.com/centrifuge/centrifuge-chain/pull/1629/commits/2901f0ec4d758715b5bbdc89feba843590496944

PR #1629, opened 2023-11-28
Oracle V2 implementation

lemunozm:
# Description

Main new requirements:
- Different feeders for different prices in different collections.
- Feeders are permissionless, and the admin chooses the feeders for a pool.
- The collection is updated under demand.
- We prioritize low PoV against DB rate access.

See the specs [here](https://centrifuge.hackmd.io/6QQSX-RMSJeRVjSlYGjJKw?view)

## Changes and Descriptions

- [x] Add `ValueProvider` trait to pass values through the oracle structure.
- [x] Add `OracleConverterBridge` to convert from `Quantity` to `Balance`
- [x] Add `PoolAdminCheck` type implementing `PreConditions` to carry the evaluation of a pool admin easily.
- `pallet-oracle-feed`: a pallet that just store feeding values
  - [x] Implementation
  - [x] Testing
  - [x] Benchmarking
- `pallet-oracle-data-collection`: a pallet that compose collections aggregating values.
  - [x] Implementation
  - [x] Testing
  - [x] Benchmarking
- [x] Add to development runtime (and compiling)
- [x] Add to altair runtime
- [x] Add to centrifuge runtime

### This PR not:
- Attach `pallet-loans` logic to these new pallets
- Remove old oracle infrastructure
- Modify oracle integration tests
- Perform any migration


pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+		MaxCollectionSize,
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Update the feeders associated to an specific key.
+		///
+		/// NOTE: In the future this should be go though change guard.
+		/// By now we trust in the admin to choose honest feeders without
+		/// invertor supervision
```
lemunozm, line 148: I think it is feasible to do it in this PR, do you agree? cc @hieronx @mustermeiszer

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+			});
+
+			Ok(())
+		}
+
+		/// Update the collection, doing the aggregation for each key in the
+		/// process
+		#[pallet::weight(1_000_000)]
+		#[pallet::call_index(3)]
+		pub fn update_collection(
```
lemunozm, line 182: The cost of this method currently is `O(Feeders * Keys)` which could be really high. Just highlight this to be aware of

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+				.into_iter()
+				.map(|feeder| T::OracleProvider::get(&(feeder, *collection_id), key))
+				.collect::<Result<Vec<_>, _>>()?;
+
+			let (mut values, mut timestamps): (Vec<_>, Vec<_>) = fed_values.into_iter().unzip();
+
+			let value = util::median(&mut values).ok_or(Error::<T>::KeyNotInCollection)?;
+			let timestamp = util::median(&mut timestamps).ok_or(Error::<T>::KeyNotInCollection)?;
+
+			Ok((*value, *timestamp))
```
lemunozm, line 230: This can probably be moved to an `AggregationProvider` trait or similar to handle this in the runtime.

I still want to think more about using the median for the timestamp, too. I think could work but not 100% sure.

pallets/oracle-feed/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Permissionles call to feed an oracle key from a source with value.
+		/// The first time the value is set, an extra fee is required for the
+		/// feeder.
+		#[pallet::weight(1_000_000)]
+		#[pallet::call_index(0)]
+		pub fn feed(
```
lemunozm, line 93: I think the extrinsic `purge()` could be possible if purged values under usage (those with its `key_info.usage_refs > 0`) are still maintained in the collection in `pallet-oracle-data-collection` until nobody uses them.

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+	#[pallet::storage]
+	pub(crate) type Keys<T: Config> = StorageDoubleMap<
+		_,
+		Blake2_128Concat,
+		T::CollectionId,
+		Blake2_128Concat,
+		T::OracleKey,
+		KeyInfo<T>,
+		ValueQuery,
+	>;
```
lemunozm, line 106: I'm quite happy because finally, it seems that we only need 2 storages in this pallet 😃

lemunozm (commented):
Highlighting the most important things:

pallets/oracle-data-collection/src/lib.rs
lemunozm: The infrastructure is already done, so not too much!

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+		/// By now we trust in the admin to choose honest feeders without
+		/// invertor supervision
+		#[pallet::weight(1_000_000)]
+		#[pallet::call_index(2)]
+		pub fn update_feeders(
+			origin: OriginFor<T>,
+			collection_id: T::CollectionId,
+			key: T::OracleKey,
+			feeders: BoundedVec<T::AccountId, T::MaxFeedersPerKey>,
+		) -> DispatchResult {
```
lemunozm, line 156: Regarding the last requirement change in [this slack conversation]( https://kflabs.slack.com/archives/C050E3ZN2JU/p1701076174699669), do you think it still makes sense to let the admin to link the feeders to certain keys, or directly associate feeders to the whole collection?

Right now, because feeders are associated with the keys, the admin still handles which keys belong to the connection indirectly.

cc @hieronx @mustermeiszer

mustermeiszer: Yes, let the admin decide and update here.

pallets/oracle-data-collection/src/lib.rs
mustermeiszer: As long as the max weights we set do not consume more than half a block and are still in the realms of enough keys, feeders, this should be fine for now.

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+		#[pallet::call_index(3)]
+		pub fn update_collection(
+			origin: OriginFor<T>,
+			collection_id: T::CollectionId,
+		) -> DispatchResult {
+			ensure_signed(origin)?;
+
+			let values = Keys::<T>::iter_key_prefix(&collection_id)
+				.map(|key| Self::get(&key, &collection_id).map(|value| (key, value)))
+				.collect::<Result<Vec<_>, _>>()?;
```
mustermeiszer, line 190: I am not sure we should do that...

mustermeiszer: The storage tree should not be unbalanced as we are using the blake2 hashing for generating the prefix but I wonder whether this could bite us...

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+			collection_id: &T::CollectionId,
+		) -> Result<Self::Data, DispatchError> {
+			let key_info = Keys::<T>::get(collection_id, key);
+			let fed_values = key_info
+				.feeders
+				.into_iter()
+				.map(|feeder| T::OracleProvider::get(&(feeder, *collection_id), key))
+				.collect::<Result<Vec<_>, _>>()?;
+
+			let (mut values, mut timestamps): (Vec<_>, Vec<_>) = fed_values.into_iter().unzip();
```
mustermeiszer, line 225: Am I understanding this correctly that we fetch the values here instead of using the cached ones?

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+
+		/// Represent an oracle value
+		type OracleValue: Parameter + Member + Copy + MaxEncodedLen + Ord;
+
+		/// Represent the a time moment
+		type Moment: Parameter + Member + Copy + MaxEncodedLen + Ord;
+
+		/// A way to obtain oracle values from feeders
+		type OracleProvider: ValueProvider<
+			(Self::AccountId, Self::CollectionId),
```
mustermeiszer, line 67: Why this? THis will enforce that oracle providers use the `CollectionId` while feeding the oracles right?

pallets/oracle-data-collection/src/lib.rs
lemunozm: Yes. `The OracleProvider::get()` is who fetches. I'm open to using the cached ones, but we would need to read the entire collection for just one value, which could be a PoV issue. What's your point of view regarding this?

pallets/oracle-data-collection/src/lib.rs
lemunozm: This is because of the OracleConverterBridge. From the data collection perspective, we need to query with the `CollectionId` in order to be able to make the conversion. But from oracle perspective, there is no `CollectionId`

pallets/oracle-data-collection/src/lib.rs
mustermeiszer: And I would also generate the aggregates here. Otherwise this is just used for collecting the feeders for keys, but in theoru this should cache the actual final results.

lemunozm: The call to 'Self::get()' already performs the aggregation

mustermeiszer: Ahh. Okay! So this is caching everything?

pallets/oracle-data-collection/src/lib.rs
mustermeiszer: > What's your point of view regarding this?

Sorry, I misunderstood. So this is jsut for getting a single value and then if you want a collection you take the cached one? This is great! And agree with the PoV issue otherwise.

pallets/oracle-data-collection/src/lib.rs
lemunozm: If we do not want to iterate though the keys, then we need to stores them under the same structure. By doing it, we would increase the PoV in `update_feeders()`, because it would need to write all keys each time a key is updated. 

But it could work because I think that call is not called frecuently and the admin could update all feeders for all keys in just one call.
If agree I go with it.

mustermeiszer: Why do we want to update all feeders at once? From a UI perspective and to simlify this api it would be easier if we could just add and remove single once.

mustermeiszer: I would estimate the same! SOunds good.

pallets/oracle-data-collection/src/lib.rs
mustermeiszer: nice! THen we aggregate in the runtime?

pallets/oracle-data-collection/src/lib.rs
lemunozm: @mustermeiszer, because there were several topics in this thread, to recap, regarding using `iter_key_prefix`, are you fine maintaining that call or do you prefer to remove it?

If we do not want that call, then the storage should be as follows: `StorageMap<CollectionId, Map<Key, Vec<Feeders>>>`, so a single addition/modification in any key or feeder will require to write/read the entire collection.

I'm not against the this because `update_collection()` is called more frequently than `update_feeders()`

mustermeiszer: @wischli wdyt regarding iterating over storage keys vs. the above solution? 

inside which logic would the above happen @lemunozm ?

lemunozm: - You would read the entire collection:
  -  each update_collection() action (more frequent).
- You would write the entire collection:
  -  each key/feeder is updated by the admin (less frecuent)
  -  each time a loan is created or closed, because the loan increases the ref counter inside of that storage. (less frequent)

lemunozm: Taking into account than now in each `update_collection()` we're reading each value of each key individually (`O(K * V)`), I would go with the `iter_key_prefix()` solution, which only adds `O(k)`. So if we have a performance limitation, will be because the first `O(k * V)` is not by iterating this `O(k)`, which is a dimension lower. WDYT?

mustermeiszer: I aggree, I am just worried that the iterator bites us. Usually, you should not use it, as there is no clear guarantee besides the hashing security that other storage does not introduce a lot of branching. 

I am not 100% sure, but I think the iteration will then have much more reads than we expect.

mustermeiszer: Would need to take a look at the iterator implementaion again.

mustermeiszer: But move ahead as is. And we see if we later change it.

lemunozm: Ok, I would also need to dig into it

pallets/oracle-feed/src/lib.rs
lemunozm: The previous message is correct, but if the user wants to get a purge value without reading the collection (that is, using `get()` instead of `collection()` from `pallet-loans`), we currently have no way to obtain that purged value. The only way is by reading the entire collection (which still contains the purged ones) which leads to PoV issues.

By now, I will go with not having `purge()` and taking an extra fee in the first feed for a key.

pallets/oracle-data-collection/src/lib.rs
lemunozm: The above means:

- If 50% of the feeders have set the correct value, we would aggregate with the correct value.
- If 50% of the feeders are updated, the price will be updated.

If we have less than 50% of well-behaved/ updated feeders, then the precision of the value and timestamp obtained will decrease.

I think we can leave under this scenario, and even more, taking into account that the admin is incentivized to select good/friend/owned feeders.

WDYT? @mustermeiszer @hieronx

lemunozm:
Ready for a final review. 

To help the reader, the amount of file changed are related to:
- 2 new pallets.
- Some runtime types required for those pallets.
- Some traits required for those pallets
- Some mocks for the new/adapted traits
- Addition in the runtimes
- Computed weights in runtimes

pallets/oracle-feed/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+		/// The weight information for this pallet extrinsics.
+		type WeightInfo: WeightInfo;
+	}
+
+	/// Store all oracle values indexed by feeder
+	#[pallet::storage]
+	pub(crate) type FedValues<T: Config> = StorageDoubleMap<
+		_,
+		Blake2_128Concat,
+		T::AccountId,
```
mustermeiszer, line 77: This should not be `T::AccountId` but rather an `T::Origin`. This allows us to have democracy provide oracles too. Either by council or by `Origin::Root` if wanted to.

pallets/oracle-feed/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+		/// The first time the value is set, an extra fee is required for the
+		/// feeder.
+		#[pallet::weight(T::WeightInfo::feed_first())]
+		#[pallet::call_index(0)]
+		pub fn feed(
+			origin: OriginFor<T>,
+			key: T::OracleKey,
+			value: T::OracleValue,
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
```
mustermeiszer, line 105: We should just ensure it is not unsigned. All other origins are fine.

mustermeiszer (commented):
First round. Sorry for stalling. Looks goood, but need to look at the details again.

pallets/oracle-feed/src/lib.rs
lemunozm: I would say there is no need to pay fees for the case of root, right?

lemunozm: Actually, I have no way to do it, since fees are changed to an `AccountId`

mustermeiszer: That is fine, only `Origin::Signed` should pay a fee.

lemunozm:
@mustermeiszer here is the root support: [29bfc7d](https://github.com/centrifuge/centrifuge-chain/pull/1629/commits/29bfc7d0ef6ae6ad0be577f75fe2f941b681aa33)

I've needed to propagate that to `pallet-oracle-data-collection` to be able to fetch values fed by root

pallets/oracle-feed/src/lib.rs
mustermeiszer: Sounds good!

pallets/oracle-feed/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+		/// The first time the value is set, an extra fee is required for the
+		/// feeder.
+		#[pallet::weight(T::WeightInfo::feed_with_fee())]
+		#[pallet::call_index(0)]
+		pub fn feed(
+			origin: OriginFor<T>,
+			key: T::OracleKey,
+			value: T::OracleValue,
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed_or_root(origin)?;
```
mustermeiszer, line 105: This needs to also allow any other origin. E.g. Bridge origin if we receive feeds from other domains or so.

pallets/oracle-feed/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+
+			FedValues::<T>::mutate(&who, key, |prev_value| {
+				let new_weight = match (&prev_value, &who) {
+					(None, Some(account_id)) => {
+						T::FirstValuePayFee::pay(account_id)?;
```
mustermeiszer, line 110: only `Origin::Signed` should pay for fees. I think here all others are paying too.

pallets/oracle-feed/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+		/// The weight information for this pallet extrinsics.
+		type WeightInfo: WeightInfo;
+	}
+
+	/// Store all oracle values indexed by feeder
+	#[pallet::storage]
+	pub(crate) type FedValues<T: Config> = StorageDoubleMap<
+		_,
+		Blake2_128Concat,
+		Option<T::AccountId>,
```
mustermeiszer, line 77: Should be `Origin` not to store different origins.

lemunozm: Actually, `None` corresponds to `Root` and the account to any Signed account. The origin type is not encodable and also contains a variant for unsigned elements that we do not want here.

Seeing that, do you agree as it is?

pallets/oracle-feed/src/lib.rs
lemunozm: Root is `None` in the `Some(account_id)` part, so it goes to the other branch

pallets/oracle-feed/src/lib.rs
lemunozm: Mmm... Ok, I understand what you want now. So, that should be checked by another `T::EnsureFeeder` type or similar the runtime later configures

pallets/oracle-feed/src/lib.rs
lemunozm: The idea here was `None` for Root and `Some(_)` for a Signed account. But if we want other origins as the bridge origin, we definitely need an origin enum type here.

mustermeiszer: I would just use `T::RuntimeOrigin`,

pallets/oracle-feed/src/lib.rs
mustermeiszer: mmh. I think all but `None` are fine. But that should be a runtime parameter probably ^^

lemunozm:
@mustermeiszer when you say `Origin` you mean something like:

```rust
#[pallet::origin]
pub type Origin = FeederOrigin;
```

?

lemunozm:
Ok, I think we do not want that, just reuse the current `RuntimeOrigin`.

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+
+		/// Collection size reached
+		NoOracleCollectionChangeId,
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Propose an update in the feeders associated to an specific key.
+		/// The collection will only be modified once
+		/// [`Pallet::apply_update_feeders`] be called.
```
wischli, line 190: Nit
```suggestion
		/// Propose an update of feeders associated to a specific key.
		/// The collection will only be modified once
		/// [`Pallet::apply_update_feeders`] is called.
```

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+					Err(err) => Some(Err(err)),
+				})
+				.collect::<Result<BTreeMap<_, _>, _>>()?;
+
+			let collection =
+				BoundedBTreeMap::try_from(values).map_err(|()| Error::<T>::MaxCollectionSize)?;
+
+			let len = collection.len();
+
+			Collection::<T>::insert(collection_id, collection);
```
wischli, line 276: Q: It is intended to overwrite any existing storage entry, correct?

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+		if items.is_empty() {
+			return None;
+		}
+
+		let mid_index = items.len() / 2;
+
+		// Won't panic as `items` ensured not empty.
+		let (_, item, _) = items.select_nth_unstable(mid_index);
+		Some(item)
+	}
```
wischli, line 495: Q: Why can't we just use `items.get(items.len() / 2)` which returns `None` by default if the index is out of bounds?

pallets/oracle-data-collection/src/tests.rs
```diff
@@ -0,0 +1,309 @@
...
+			feeders,
+		)
+		.unwrap();
+
+		OracleCollection::apply_update_feeders(
+			RuntimeOrigin::signed(ADMIN),
+			COLLECTION_ID,
+			CHANGE_ID,
+		)
+		.unwrap();
```
wischli, line 102: Nit: Would be cleaner to wrap the calls into `assert_ok!(..)`

pallets/oracle-data-collection/src/tests.rs
```diff
@@ -0,0 +1,309 @@
...
+			Error::<Runtime>::IsNotAdmin
+		);
+	});
+}
+
+#[test]
+fn register() {
+	new_test_ext().execute_with(|| {
+		assert_ok!(OracleCollection::register_id(&KEY_A, &COLLECTION_ID));
+		assert_eq!(Keys::<Runtime>::get(COLLECTION_ID, KEY_A).usage_refs, 1);
```
wischli, line 156: Nit: Could add check for `KeysAdded` event dispatch

pallets/oracle-data-collection/src/tests.rs
```diff
@@ -0,0 +1,309 @@
...
+fn unregister() {
+	new_test_ext().execute_with(|| {
+		assert_ok!(OracleCollection::register_id(&KEY_A, &COLLECTION_ID));
+		assert_ok!(OracleCollection::register_id(&KEY_A, &COLLECTION_ID));
+
+		assert_ok!(OracleCollection::unregister_id(&KEY_A, &COLLECTION_ID));
+		assert_eq!(Keys::<Runtime>::get(COLLECTION_ID, KEY_A).usage_refs, 1);
+
+		assert_ok!(OracleCollection::unregister_id(&KEY_A, &COLLECTION_ID));
+		assert_eq!(Keys::<Runtime>::get(COLLECTION_ID, KEY_A).usage_refs, 0);
```
wischli, line 173: Nit: Could add check for `KeysRemoved` event dispatch

pallets/oracle-data-collection/src/tests.rs
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	pub fn prepare_provider() {
+		MockProvider::mock_get(|(account, collection_id), key| {
+			assert_eq!(collection_id, &COLLECTION_ID);
+			match (account, key) {
+				(&FEEDER_1, &KEY_A) => Ok(Some((100, 50))),
+				(&FEEDER_2, &KEY_A) => Ok(Some((101, 45))),
+				(&FEEDER_3, &KEY_A) => Ok(Some((102, 55))),
+				(&FEEDER_1, &KEY_B) => Ok(Some((1000, 500))),
```
wischli, line 61: Nit: Since these values are used throughout the test, it would be nice to define them as consts as well to improve readability.

pallets/oracle-feed/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+//! Oracle pallet to feed values.
+//! Feeding is permissionless given an initial fee for each key.
+//! This pallet do not aggregate/validate any, it just store them by account as
+//! they come. It's expected other pallet read this storage and
+//! aggregate/validate the values.
```
wischli, line 18: Nit:
```suggestion
//!
//! ### Assumptions
//!
//! This pallet neither aggregates nor validates anything. It just stores values
//! by account as they come. It's expected that another pallet reads the storage
//! of this pallet and provides aggregation and validation methods to the
//! values.
```

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+//! Oracle pallet to collect and aggregate oracle values.
+//!
+//! The collection admin configures the collection allowing a list of feeders.
+//!
+//! Later, the updating a collection will collect all values based on the admin
+//! configuration of the collection. The resulting collection is optimized to
+//! iterate though all values in jus one read.
+//!
+//! This pallet is not fed with external values, you need to configure a
+//! `ValueProvider` to provide it with oracle values.
```
wischli, line 23: Nit: Let's make the assumption more explicit:

```suggestion
//! ### Assumptions
//!
//! This pallet is not fed with external values. You need to configure a
//! `ValueProvider` to provide it with oracle values.
```

pallets/oracle-feed/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+		Blake2_128Concat,
+		T::OracleKey,
+		(T::OracleValue, MomentOf<T>),
+	>;
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		Fed {
+			account_id: Option<T::AccountId>,
```
wischli, line 87: By changing the key of `FedValues` to `T::RuntimeOrigin`, this change should be applied here as well.

pallets/oracle-feed/src/tests.rs
```diff
@@ -0,0 +1,55 @@
...
+
+const FEEDER: AccountId = 1;
+const KEY: OracleKey = 23;
+const VALUE1: OracleValue = 42;
+const VALUE2: OracleValue = 43;
+const TIMESTAMP1: Timestamp = 1234;
+const TIMESTAMP2: Timestamp = 1235;
+
+#[test]
+fn feed() {
```
wischli, line 14: Nit: Could add assertion for event dispatch

pallets/oracle-feed/src/tests.rs
```diff
@@ -0,0 +1,55 @@
...
+fn feed_root() {
+	new_test_ext().execute_with(|| {
+		MockTime::mock_now(|| TIMESTAMP1);
+		MockPayFee::mock_pay(|_| unreachable!("Feeding from root does not require fees"));
+
+		assert_ok!(OracleFeed::feed(RuntimeOrigin::root(), KEY, VALUE1));
+		assert_ok!(OracleFeed::get(&None, &KEY), Some((VALUE1, TIMESTAMP1)));
+	});
+}
+
```
wischli, line 49: When changing the key of `FedValues` from `T::AccountId` to `T::RuntimeOrigin`, we need to add another test which ensures the `None` origin cannot feed (because it cannot be charged).

runtime/altair/src/lib.rs
```diff
@@ -1350,6 +1352,38 @@ impl pallet_membership::Config for Runtime {
 	type WeightInfo = pallet_membership::weights::SubstrateWeight<Self>;
 }
 
+parameter_types! {
+	#[derive(Clone, PartialEq, Eq, Debug, TypeInfo, Encode, Decode, MaxEncodedLen)]
+	pub const MaxFeedersPerKey: u32 = 10;
+	pub const FirstValueFee: Fee = Fee::Balance(10 * CFG); // TODO: Specify
```
wischli, line 1358: Reminder for this `TODO`. We should make use of the `deposit` function.

runtime/altair/src/lib.rs
```diff
@@ -1350,6 +1352,38 @@ impl pallet_membership::Config for Runtime {
...
+}
+
+impl pallet_oracle_feed::Config for Runtime {
+	type FirstValuePayFee = FeeToTreasury<Fees, FirstValueFee>;
+	type OracleKey = OracleKey;
+	type OracleValue = Quantity;
```
wischli, line 1364: Is it really `Quantity`? I would have imagined `Rate` for loans or `Ratio` for the orderbook

runtime/altair/src/lib.rs
```diff
@@ -1350,6 +1352,38 @@ impl pallet_membership::Config for Runtime {
...
+	type RuntimeEvent = RuntimeEvent;
+	type Time = Timestamp;
```
wischli, line 1366: Nit: Maybe we want to rename this type to `Timestamp` and the `pallet_oracle_data_collection::Config::Timestamp` to `Time`?

runtime/common/src/changes.rs
```diff
@@ -44,18 +55,17 @@ impl<T: Changeable, Options: Clone> RuntimeChange<T, Options> {
 				LoansChange::<T>::Policy(_) => vec![week, blocked],
 				LoansChange::<T>::TransferDebt(_, _, _, _) => vec![],
 			},
+			RuntimeChange::OracleCollection(change) => match change {
+				OracleCollectionChange::Feeders(_, _) => vec![], // TODO: specify
```
wischli, line 59: Reminder for this open `TODO`

wischli (commented):
First round with a lot of nits and some questions from my side. Feel free to ignore all the nits. Looks really good!

pallets/oracle-data-collection/src/benchmarking.rs
```diff
@@ -0,0 +1,153 @@
...
+			.map(|i| account::<T::AccountId>("feeder", i, 0).into())
+			.collect::<Vec<_>>()
+			.try_into()
+			.unwrap();
+
+		#[extrinsic_call]
+		apply_update_feeders(
+			RawOrigin::Signed(admin),
+			T::CollectionId::default(),
+			util::last_change_id_for::<T>(T::OracleKey::default(), &feeders),
```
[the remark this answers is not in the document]
lemunozm: Really good point @wischli!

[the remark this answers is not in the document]
lemunozm: I'm just curious about whether the framework precomputes the parameters before counting the time, and from reading the procedural macro, it seems so. But anyway, I think it is safer to move it above that line, as you say.

pallets/oracle-data-collection/src/benchmarking.rs
```diff
@@ -0,0 +1,153 @@
...
+			T::CollectionId::default(),
+			T::OracleKey::default(),
+			feeders,
+		);
+
+		Ok(())
+	}
+
+	#[benchmark]
+	fn apply_update_feeders(n: Linear<1, 10>) -> Result<(), BenchmarkError> {
```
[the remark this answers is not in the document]
lemunozm: Yes, you can do it. But I thought that if some runtime configures a really high value, the benchmark could take a lot of time to compute (especially in the other benchmark that has 2 variables). Because it's linear, `10` values should be enough to determine the correct linear function for later infer high values

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+//! Oracle pallet to collect and aggregate oracle values.
+//!
+//! The collection admin configures the collection allowing a list of feeders.
+//!
+//! Later, the updating a collection will collect all values based on the admin
+//! configuration of the collection. The resulting collection is optimized to
+//! iterate though all values in jus one read.
```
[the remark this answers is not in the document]
lemunozm: Thanks for the corrections!

runtime/altair/src/lib.rs
lemunozm: `Rate` sure not because it has 27 decimals with a few integer parts. It could be a `Ratio` which is exactly the same time as `Quantity` (both alias the same exact type). We should probably unify these types to avoid confusion.

Regarding the name, I think you're right, and `Ratio` fits better to specify a "price"

lemunozm:
Thanks for your deep review @wischli, I'll go into detail with your comments once I give the support for RuntimeOrigin to the PR

runtime/altair/src/lib.rs
mustermeiszer: @lemunozm can you make this an enum already? I think we will have non ratio types in the future. WDYT?

pallets/oracle-feed/src/lib.rs
lemunozm: To keep updated this topic for future issues.

`T::RuntimeOrigin` is not encodable and can not be stored, but from `T::RuntimeOrigin`, you can obtain its internal representation (`OriginCaller`) through `OriginTrait` that it is. Although `OriginCaller` has few traits implemented and is less comfortable to use.

pallets/oracle-feed/src/tests.rs
lemunozm: Finally, I've left it under an `EnsureOrigin` trait, which should determine what to do in each case. So, the test is not needed now because it is handled in the runtime. But thanks for this very accurate comment! 🙌🏻

pallets/oracle-feed/src/lib.rs
lemunozm: Finally by an `EnsureOrigin` parameter, which is configured as you say

lemunozm (commented):
@mustermeiszer latest changes for the RuntimeOrigin support: [72f3e09](https://github.com/centrifuge/centrifuge-chain/pull/1629/commits/72f3e09b871af52e82ebb77771e69cfea4e94c1b)

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+		/// The key is not in the collection.
+		KeyNotInCollection,
+
+		/// The key is not registered
+		KeyNotRegistered,
+
+		/// Collection size reached
+		MaxCollectionSize,
+
+		/// Collection size reached
```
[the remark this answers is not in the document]
lemunozm: Yeah, thanks!

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+				Error::<T>::IsNotAdmin
+			);
+
+			transactional::with_transaction(|| {
+				let result = Self::update_feeders(collection_id, key, feeders.clone());
+
+				// We do not want to apply the mutation,
+				// only check if there is no error in applying it
+				TransactionOutcome::Rollback(result)
+			})?;
```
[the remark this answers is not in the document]
lemunozm: Thanks!!

Yes, right now, looking at the code it's infallible; good point! The error comes from `mutate_and_remove_if_clean` which for other calls need to return an error. In this case I still need to handle the error to avoid `.unwrap()` or something dangerous.

pallets/oracle-data-collection/src/lib.rs
lemunozm: Yes, that is the intention. Thanks for checking it

pallets/oracle-data-collection/src/tests.rs
lemunozm: Yeah! I have a nitpick reason for this. 

I put `assert_ok!` in lines that the test should evaluate and `unwrap()` in lines where it's expected to work but does not belong to the evaluation of the test itself.

In this case, those lines are under `util::update_feeders()`, which is a utility whose concern is not to evaluate if it works fine but to provide some state to later use in other tests.

The well behaviour of those calls with `assert_ok!` is done in `updating_feeders()`

P.S. Quite nitpick probalbly 😆

pallets/oracle-data-collection/src/lib.rs
lemunozm: Agree!

runtime/altair/src/lib.rs
lemunozm: Make sense to avoid future migrations

runtime/common/src/changes.rs
lemunozm: I think it's ok for now to leave it empty, WDYT @mustermeiszer ?

runtime/altair/src/lib.rs
lemunozm: I'll think about this. We have complete naming chaos regarding this, haha.

Probably `Time` to `Timer` says better what it means?

I think calling `Timestamp` to `pallet_timestamp` is not the best name, because it creates a lot of confusion when you need to define a timestamp variable. But it's too late for that and the name is on-chain

pallets/oracle-data-collection/src/tests.rs
wischli: Makes sense! Thanks for the explanation :)

runtime/altair/src/lib.rs
wischli: Sorry, I thought we were utilizing this const function from Substrate already, but I was wrong: https://github.com/paritytech/substrate/blob/5e49f6e44820affccaf517fd22af564f4b495d40/bin/node/runtime/src/constants.rs#L28

For oracles, I suggest to set items to the appropriate `Max*` bound for the number of values per key and `bytes` as the max size of a value. Then, we only need to apply weights to both parameters. For that, we should be lower than Polkadot, maybe Kusama configuration would be a good inspiration.

pallets/oracle-data-collection/src/tests.rs
lemunozm: Done

runtime/altair/src/lib.rs
lemunozm: @mustermeiszer thinking about this. Do you think it would be interesting to have different pallet instances for `pallet-oracle-feed` in case of different values instead?

I say this to avoid, for example, a feeder feeding with a Key as `Isin` and a Value as a non-radio type, mixing keys with non related values.

lemunozm: I think it all depends on how the consumer wants to consume. No strong opinion on it

runtime/altair/src/lib.rs
lemunozm: Mmm... OK, I see. That represents somehow the byte cost of storing things on the chain, right?

In our case, the fee is for 1 item of `bytes(OracleKey) + bytes(OracleValue)` bytes

Do you know if our codebase currently has any function to handle this, or would I need to create one?

wischli: > Mmm... OK, I see. That represents somehow the byte cost of storing things on the chain, right?

Exactly. Anything else is arbitrary and could expose an attack vector if the deposit is too low for the occupied storage. Since oracles can be fed by anyone, we should use a similar function IMO. But it should not be a blocker for this PR and can be done in a subsequent one.

> Do you know if our codebase currently has any function to handle this, or would I need to create one?

Yes, we actually have and use it (at least for Proxy and Preimages) with the same configuration as above (with `CENTI_CFG` instead of `CENTS`): https://github.com/centrifuge/centrifuge-chain/blob/c4f05ca273ed14d38b484b2f3bdd5c0dd66dbb70/libs/primitives/src/lib.rs#L260-L262

So we could end up with 
```rust
	pub const FirstValueFee: Balance = deposit(1, 4 + 16)
```

which would be 135 `CENTI_CFG` or `1.35` CFG. To me, that appears to be a reasonable fee for occupying storage and protecting against spamming. In the end, all fees benefit our treasury and thus the longterm sustainability of the protocol 🙃

lemunozm: Thanks for expanding this, @wischli; I agree with all you've said!

lemunozm: Just one question. Does the bytes of the key represent the hash representation of such key in the map? (I think 32 bytes)

runtime/altair/src/lib.rs
wischli: > Probably `Time` to `Timer` says better what it means?

Yeah maybe. But we can also keep as is!

runtime/altair/src/lib.rs
wischli: It probably does, I am not entirely sure. I would just work with the bytes of the storage items: `OracleKey (u8)`, `OracleValue (u128)`, `Timestamp (u64)`. So I would adjust my above stated `135 CENTI_CFG` to `165 CENTI_CFG`.

For the `pallet_preimage`, one preimage byte costs exactly `deposit(0, 1)`.

runtime/altair/src/lib.rs
lemunozm: I'll leave it as it by now is until we reach some kind of consensus for time naming. Thanks for pointing this out!

pallets/oracle-feed/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+}
+
+pub mod util {
+	use crate::pallet::{Config, MomentOf};
+
+	pub const fn size_of_feed<T: Config>() -> u32 {
+		sp_std::mem::size_of::<(T::OracleKey, T::OracleValue, MomentOf<T>)>() as u32
+	}
```
wischli, line 172: Really elegant!

runtime/altair/src/lib.rs
wischli: I think instantiating the `pallet-oracle-feed` makes more sense for the mentioned reason. That can be done in a follow up PR. But @mustermeiszer should have the stronger say here.

wischli (approved):
Thanks for adhering to many of my nits and arguing against the other. AFAICS, the last [open discussion](https://github.com/centrifuge/centrifuge-chain/pull/1629/files#r1432315694) is about instantiation vs. `OracleValue` enum. As I am voting for the former, I am approving this PR already. We must not merge before Frederik's approval though.

runtime/altair/src/lib.rs
lemunozm: Agree!

Once this last point is solved, I think we can merge this PR.

lemunozm: After another thought, definitely I think it's better to use different pallet instances to maintain always well-formed key-value correlations on-chain. If some consumer needs different value types, we can always compound them into an enum using a runtime type that implements `ValueProvider` in the runtime. Similar to what we already do with `OracleConverterBridge`.

mustermeiszer: I am really undecided here. 

PRO: 
* strong typed relationship
* cleaner from a technical perspective

CON: 
* We have 70 pallets right now, the limit is 255 - not really close, but we are already having more pallets than most parachains IIRC
* External integrations - It is just harder to look at the right place but I might over-value that

I have even thought about allowing this API

```
		pub fn feed(
lemunozm marked this conversation as resolved.
			origin: OriginFor<T>,
			key: T::OracleKey,
			value: BoundedVec<T::MaxScaleSizeValue>,
```

and upon fetching/aggregating we decode into the needed type and fail if it is not correct. That would be the most flexible.

For example with currently we have needs for
* `ISIN` -> `Price`
* `(Currency, Currency)` -> `Price`

Should we really have an instance of this pallet for each combination that we need on-chain? There will be more, I can promise that ^^

runtime/common/src/oracle.rs
```diff
@@ -142,3 +148,90 @@ pub mod benchmarks_util {
...
+	))?;
+
+	Ok(metadata.decimals)
+}
+
+/// A provider bridge that transform generic quantity representation of a price
+/// into a balance denominated in a pool currency.
+pub struct OracleConverterBridge<Origin, Provider, Pools, AssetRegistry>(
+	PhantomData<(Origin, Provider, Pools, AssetRegistry)>,
+);
```
mustermeiszer, line 192: Don't we also need a converter for the `CachedCollection`, or is it implicitly using this one?

runtime/centrifuge/src/lib.rs
```diff
@@ -1704,6 +1708,40 @@ impl pallet_membership::Config for Runtime {
 	type WeightInfo = pallet_membership::weights::SubstrateWeight<Self>;
 }
 
+parameter_types! {
+	#[derive(Clone, PartialEq, Eq, Debug, TypeInfo, Encode, Decode, MaxEncodedLen)]
+	pub const MaxFeedersPerKey: u32 = 10;
+	pub const FirstValueFee: Fee = Fee::Balance(deposit(1, pallet_oracle_feed::util::size_of_feed::<Runtime>()));
```
mustermeiszer, line 1714: This means `1 CFG` per value?

mustermeiszer (commented):
Look great, just need to decide on the instance vs. enum vs. generic scale for key-value in oracle.

runtime/centrifuge/src/lib.rs
wischli: `1.65 CFG` for each first value, yes. See here for context: https://github.com/centrifuge/centrifuge-chain/pull/1629#discussion_r1432380994

runtime/altair/src/lib.rs
lemunozm: 🤔

> Should we really have an instance of this pallet for each combination that we need on-chain?

I think both relations above can be in the same pallet instance. The pallet instance can have the Key as an enum but the Value as `Price`. That way, any Key enum variant will map to a correct Value type.

The amount of instances will depend only on the amount of different kinds of values, not keys. Mostly, I can see `Price,` `Balance`, and maybe some id or similar (?).

WDYT?

runtime/common/src/oracle.rs
lemunozm: yes it is implicitly using this one

runtime/altair/src/lib.rs
mustermeiszer: The encoding is always ensured when fetching the keys. Basically anywas what we are already doing when accceotinbg an extrinsic, but just during a runtime call. It is then an invalid key. 

But lets go with instances if you think it is better. Should we already make the current `key` then an enum?

lemunozm: > Should we already make the current key then an enum?

It currently is!

mustermeiszer (approved):
Thanks! All thinks sort out from my side.

PR #1630, opened 2023-11-28
Restricted transfers follow-ups

cdamian:
Some follow-ups for the initial works that was done for restricted transfers:

- Removing xcm-simulator tests.
- Adding more integration tests.
- Minor changes to generic testing environment.


runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -1780,23 +1924,293 @@ mod centrifuge {
...
+					WeightLimit::Unlimited,
+				));
+
+				assert_eq!(
+					orml_tokens::Pallet::<T>::free_balance(USDC, &Keyring::Alice.into()),
+					alice_initial_usdc - usdc(1_000),
+				);
+			});
+
+			// NOTE - we cannot confirm that the Alice account present on the
```
cdamian, line 2200: When executing the above transfer, the xtokens pallet uses the `ToNonReserve` transfer kind. 

Given this transfer kind, xtokens generates an XCM message that has these instructions - `WithdrawAsset`, `ClearOrigin`, `BuyExecution`, `DepositReserveAsset` and sends it to `Parachain(1000)` which is the location of the USDC currency used in this test.

The `DepositReserveAsset` instruction is supposed to then trigger a transfer from Parachain(1000) to our sibling chain.

/cc @mustermeiszer @NunoAlexandre

NunoAlexandre: For context:

When a token XT is transferred from Parachain X to parachain Y, X keeps a record that XT went out to Y. Y can not send XT to X that did not leave X to Y. So this behaviour from xtokens makes sense, as otherwise, say, we would send 1) XT from X to Y, then 2) from Y to Z and finally 3) from Z to X; On step 3, X wouldn't know that Z owned XT so it can not accept that incoming transfer.

So the hop is necessary due to the current design of XCM. The goal of this test is to ensure that restricted_xtokens handles the permissions properly for USDC, and all that means is that it passes the request through to XTokens and thereafter outbounds the parachain; Meaning than an OK on the sending parachain is all we need to verify this property; we don't need to ensure that the intended destination chain/account receives the tokens.

runtime/integration-tests/src/generic/env.rs
```diff
@@ -61,7 +61,11 @@ impl<Event> Blocks<Event> {
 /// Define an environment behavior
 pub trait Env<T: Runtime> {
 	/// Load the environment from a storage
-	fn from_storage(parachain_storage: Storage, sibling_storage: Storage) -> Self;
+	fn from_storage(
+		relay_storage: Storage,
+		parachain_storage: Storage,
+		sibling_storage: Storage,
+	) -> Self;
```
lemunozm, line 68: Because in most tests, specifically in non-xcm tests, the typical usage will be initializing only the parachain storage, could we have two different environment initializations to simplify each test case? I mean:

```rust
// Just set the parachain storage
fn from_parachain_storage(storage: Storage) -> Self; 

// Set all storages
fn from_storage(
	relay_storage: Storage,
	parachain_storage: Storage,
	sibling_storage: Storage,
) -> Self;

**Review by lemunozm — COMMENTED** [review]
Just a thought

File: runtime/integration-tests/src/generic/env.rs (hunk 2 above)
Comment (cdamian) [reply runtime/integration-tests/src/generic/env.rs:68]: Yup, was also thinking to implement `Default` for this, since that will be used mainly from what I can tell.

Comment (lemunozm) [reply runtime/integration-tests/src/generic/env.rs:68]: Great! 👌🏻

File: runtime/integration-tests/src/generic/env.rs [hunk 3]
```diff
@@ -60,6 +60,9 @@ impl<Event> Blocks<Event> {
 
 /// Define an environment behavior
 pub trait Env<T: Runtime> {
```
lemunozm, line 62: Could we also add just: 
```suggestion
pub trait Env<T: Runtime>: Default {
```

To enable the possibility of creating a default without knowing the env type.

lemunozm (commented):
Thank so much for the changes @cdamian 🙌🏻 . I think it has been simplified a lot!

I left a minor optional comment regarding this.

runtime/integration-tests/src/generic/env.rs
cdamian: Will add this as soon as I figure out the issue with benchmarks. Thank you!

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
mustermeiszer: Yes, that should be fine!

mustermeiszer (commented):
Just am missing an `restrict_lp_eth_usdc_lp_transfer` that makes use of `liquidity_pools::Transfer`

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -1780,23 +1842,368 @@ mod centrifuge {
...
+					after_transfer_bob,
+					pre_transfer_bob + lp_eth_usdc(TRANSFER_AMOUNT)
+				);
+				assert_eq!(after_transfer_charlie, pre_transfer_charlie);
+			});
 		}
 
-		fn restrict_lp_eth_usdc_xcm_transfer<T: Runtime>() {
-			todo!()
+		fn restrict_lp_eth_usdc_lp_transfer<T: Runtime + FudgeSupport>() {
```
cdamian, line 1934: @mustermeiszer here's the test that you mentioned in a previous comment. I went the full way of adding a router just to confirm that the LP call goes through so that we can also check that the domain account gets the transfer amount. WDYT

runtime/integration-tests/src/generic/env.rs
cdamian: Added this and a couple of more changes to the `Env` trait. PTAL

mustermeiszer (approved):
NICE! Good to go.

PR #1632, opened 2023-11-29
Node crate to folder

### LINKED ISSUE #1278: `node` crate under its own folder

lemunozm:
# Description

Move the node and its dependencies to a `node` folder

Fixes #1278

## Changes and Descriptions

- `res` `src` `build.rs` to `node` folder
- Node using workspace dependencies
- Some feature compilation fixes
- Update scripts to now compile the node with `cargo build -p centrifuge-chain`
- Modify `scripts/tests.sh` into `scripts/cargo_split.sh` to filter by feature and reuse compilation artifacts, improving the speed. It takes around 20 min to compile all crates for one feature in my M1.
- Add `subalfred` target to `ci/run-check.sh` for a feature CI job checking this

## Current issue (when moving the node) *SOLVED*

When you compile the workspace, the same crate dependency built is used for different crates of the workspace. If different crates use the same dependency with different features, the first version in the compiling process is used for both. This leads to feature problems, such as the compiler complaining about missing features when the feature is actually correctly set in the crate.

When compiling the node along the workspace crate, all crates use dependencies that at least contain the features they need (probably because its dependencies are compiled first). Nevertheless, when moving the node out, the compiling order changes and causes some crates to use some dependencies with some features missing.

A simple `cargo check -F runtime-benchmarks` fails because some crate uses a previous artifact that was built without `runtime-benchmarks` even when in its `Cargo.toml`, the feature is correctly set.


ci/run-check.sh
```diff
@@ -10,11 +10,11 @@ cargo --version
 
 case $TARGET in
   cargo-build)
-    cargo build --release "$@"
+    cargo build -p centrifuge-chain --release "$@"
     ;;
 
   test-general)
-    cargo test --workspace --release --features runtime-benchmarks,try-runtime --exclude runtime-integration-tests
+    cargo test --release --features runtime-benchmarks,try-runtime,fast-runtime --exclude runtime-integration-tests
```
lemunozm, line 17: If we compile with `fast-runtime` here, it "ideally" should reuse the compiled runtimes later for testing

ci/run-check.sh
```diff
@@ -30,11 +30,25 @@ case $TARGET in
...
+    dirs=$(find . -name Cargo.toml -print0 | xargs -0 -n1 dirname | sort -u)
+
+    # Execute the command "subalfred check" on each directory
+    for dir in $dirs; do
+      # Avoiding cargo workspace
+      if [[ "$dir" == "." ]]; then
+        continue
+      fi
+      subalfred check features $dir
+    done
```
lemunozm, line 53: @wischli I finally added it here in order to let it prepared for a future CI job, is it ok for you?

node/Cargo.toml
```diff
@@ -0,0 +1,182 @@
...
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+documentation.workspace = true
+
+[package.metadata.wasm-pack.profile.release]
+# `wasm-opt` has some problems on linux, see
+# https://github.com/rustwasm/wasm-pack/issues/781 etc.
+wasm-opt = false
```
lemunozm, line 23: Does anyone know what was exactly the issue to check if it now works removing this? From the linked issue, it seems fixed, but I am not sure how to check it or evaluate it

pallets/bridge-mapping/Cargo.toml
```diff
@@ -1,30 +0,0 @@
-[package]
```
lemunozm, line 1: AFAIK, this module is deprecated and can be removed from this

lemunozm (commented):
Relevant points

Cargo.toml
```diff
@@ -66,6 +41,7 @@ members = [
   "pallets/rewards",
   "runtime/altair",
   "runtime/centrifuge",
+  "runtime/development",
```
lemunozm, line 44: I can not tell why, but you can compile development even if it is not on the list. Anyways, adding it here for correctness

ci/run-check.sh
lemunozm: I reject myself here. The cache is gathered from the previous merge in the same CI job, no across jobs in the same PR.

ci/run-check.sh
```diff
@@ -45,4 +45,19 @@ case $TARGET in
 
   docs-build)
     RUSTDOCFLAGS="-D warnings" cargo doc --all --no-deps
+    ;;
+  subalfred)
+    # Find all child directories containing Cargo.toml files
+    # TODO: Filter by crates found in the workspace
+    #   HINT: Use `cargo workspaces list -l" and filter by the paths
```
cdamian, line 52: Do we wanna address this or remove the hint?

docker/centrifuge-chain/Dockerfile
```diff
@@ -4,13 +4,13 @@
...
 	ENV RUST_BACKTRACE 1
-  
+
   # RustUp
 	COPY ./scripts scripts
 	COPY rust-toolchain.toml ./
 	# RUN rustup update && rustup default
 	ENV PATH=~/.cargo/bin:$PATH
-	RUN	scripts/install_toolchain.sh 
+	RUN	scripts/install_toolchain.sh
```
cdamian, line 13: Nit - remove extra spaces ^^

node/Cargo.toml
```diff
@@ -0,0 +1,182 @@
...
+fp-rpc = { workspace = true, default-features = true }
+fp-storage = { workspace = true, default-features = true }
+pallet-ethereum = { workspace = true, default-features = true }
+pallet-evm = { workspace = true, default-features = true }
+
+[build-dependencies]
+substrate-build-script-utils = { workspace = true, default-features = true }
+vergen = { workspace = true, default-features = true }
+
+[dev-dependencies]
```
cdamian, line 125: I'm not sure if we need any of these here.

ci/run-check.sh
lemunozm: I leave these comments in order to tackle them in another PR. Is it ok for you?

docker/centrifuge-chain/Dockerfile
lemunozm: Actually, I think the change is removing the space, am I wrong?

node/Cargo.toml
lemunozm: Good point! let me check it

docker/centrifuge-chain/Dockerfile
cdamian: Was referring to the space between `RUN` and `scripts/...` . It shows up weird on my diff here, nvm this if it looks OK in your IDE.

node/Cargo.toml
```diff
@@ -0,0 +1,182 @@
+[badges]
+travis-ci = { repository = "centrifuge/centrifuge-chain", branch = "master" }
```
wischli, line 2:
```suggestion
travis-ci = { repository = "centrifuge/centrifuge-chain", branch = "main" }
```

node/Cargo.toml
```diff
@@ -0,0 +1,182 @@
...
+maintenance = { status = "actively-developed" }
+is-it-maintained-issue-resolution = { repository = "centrifuge/centrifuge-chain" }
+is-it-maintained-open-issues = { repository = "centrifuge/centrifuge-chain" }
+
+[package]
+name = "centrifuge-chain"
```
wischli, line 8: I wonder whether the crate should rather be named `centrifuge-node`.

node/Cargo.toml
wischli: Shouldn't need those IMO

pallets/liquidity-pools/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Centrifuge <admin@centrifuge.io>"]
 description = "Centrifuge Liquidity Pools Pallet"
-edition = "2018"
+edition = "2021"
```
wischli, line 4: Great catch

scripts/cargo_split.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+# Usage
+# ./scripts/tests.sh check|test "-F=feature1,feature2,..." <--no-run>
+
```
wischli, line 5: This file should be named `tests.sh` right?

Cargo.toml
```diff
@@ -1,32 +1,7 @@
...
-[package.metadata.wasm-pack.profile.release]
-# `wasm-opt` has some problems on linux, see
-# https://github.com/rustwasm/wasm-pack/issues/781 etc.
-wasm-opt = false
-
-[badges]
-travis-ci = { repository = "centrifuge/centrifuge-chain", branch = "master" }
-maintenance = { status = "actively-developed" }
-is-it-maintained-issue-resolution = { repository = "centrifuge/centrifuge-chain" }
-is-it-maintained-open-issues = { repository = "centrifuge/centrifuge-chain" }
```
wischli, line 22: I think this should stay in the root Cargo.toml as it is not related to the client but this repository.

Cargo.toml
```diff
@@ -1,32 +1,7 @@
...
-name = "centrifuge-chain"
-version = "0.10.35"
-authors = ["Centrifuge <admin@centrifuge.io>"]
-description = "Centrifuge chain implementation in Rust."
-build = "build.rs"
-edition = "2021"
-license = "LGPL-3.0"
-default-run = "centrifuge-chain"
-homepage = "https://centrifuge.io/"
-repository = "https://github.com/centrifuge/centrifuge-chain"
```
wischli, line 11: All of this should stay, otherwise our `workspace.*` is empty right?!

Cargo.toml
```diff
@@ -1,32 +1,7 @@
-[package]
-name = "centrifuge-chain"
```
wischli, line 2: I would also keep this and name the node crate `centrifuge-node`

wischli (commented):
Overall LGTM but I think the root Cargo.toml was emptied too much 😅

scripts/cargo_split.sh
lemunozm: It was named previously so, but now you can execute any cargo `<command>`. That's why I changed the name to another more generic.

Thinking now that maybe even already exists such cargo tool 🤔

Cargo.toml
lemunozm: Sadly, you can not place `[badges]` in `Cargo.toml` that only contains workspace related things. They seem to be related to a specific crate

Cargo.toml
lemunozm: Yes, actually, they are under `[workspace.package]` section.

Cargo.toml
lemunozm: The workspace can not have a `[package]`. It is related to a crate and now the root `Cargo.toml` does not identify any crate

node/Cargo.toml
lemunozm: Confirmed

lemunozm: Confirmed

node/Cargo.toml
wischli: I see. This was related to my comment about the root toml. Given that the root toml seems to be finde, lets keep `centrifuge-chain` here to not introduce another breaking change.

lemunozm:
Seems the docker build command is failing. Could it be related to the last addition of https://github.com/centrifuge/centrifuge-chain/pull/1646 ?

Tested and "it works on my machine"

docker/centrifuge-chain/Dockerfile
lemunozm: Ah, this was added intentionally. Looks like all sections in the scripts are split by a space except that section.

Cargo.toml
wischli: I see, then this can be marked as resolved.

lemunozm:
As a test, I've reverted the already merged https://github.com/centrifuge/centrifuge-chain/pull/1646, and now it works again.

I'm quite lost, any suggestion @gpmayorga @wischli

wischli (commented):
There are a few outdated code paths which want to pull chain specs from the root `/res` not the moved `/node/res`:

* `./docker/docker-compose-local-chain.yml`
* `./docker/docker-compose-local-relay.yml`
* `./scripts/run_collator.sh`

docker/centrifuge-chain/Dockerfile
```diff
@@ -20,8 +20,7 @@ FROM --platform=linux/amd64 docker.io/paritytech/ci-linux:production as builder
 	WORKDIR /centrifuge-chain
 	ARG FEATURES=""
 	RUN	FEATURES=$(echo ${FEATURES} | tr -d '"') \
-		cargo build --locked --release --features=${FEATURES}
-
+		cargo build -p centrifuge-chain --locked --release --features=${FEATURES}
```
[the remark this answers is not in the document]
lemunozm: I'll try!

BTW, Thanks for notifying the above files 🙌🏻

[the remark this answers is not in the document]
lemunozm: Just playing around with the above command, I'll try a `cargo update`. Maybe the `--locked` param is what is hitting us here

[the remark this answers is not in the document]
lemunozm: I tested everything and everything failed 🥲. Suuuper weird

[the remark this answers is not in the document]
lemunozm: Tested locally and also fails.

This seems to be the issue: https://stackoverflow.com/questions/74643818/docker-buildx-gives-exit-code-137-with-cargo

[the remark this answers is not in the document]
wischli: > Tested locally and also fails.
> 
> This seems to be the issue: https://stackoverflow.com/questions/74643818/docker-buildx-gives-exit-code-137-with-cargo

IMO, the CI issue and Apple M1 docker build failures are unrelated. AFAIK, we cannot build our docker image locally on M1.

@lemunozm I deleted all cache related to this PR and re-ran the docker build manually in the actions and it succeeded 🎉 https://github.com/centrifuge/centrifuge-chain/actions/runs/7206286656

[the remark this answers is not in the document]
lemunozm: Uooo! Cool!

wischli (approved):
Super happy CI is finally bending the knee. Let's get this merged ASAP!

lemunozm:
Thanks so much for all your help on this last issue @wischli 🙏🏻

lemunozm:
@mustermeiszer @NunoAlexandre I think we need an approval from any of you 🙏🏻

mustermeiszer (approved):
I disagree with moving the specs into the node folder. But no blocker.

lemunozm:
Why not? Do they not belong to the node? Actually, those files are only referenced from the node crate.

mustermeiszer:
> Why not? Do they not belong to the node? Actually, those files are only referenced from the node crate.

The spec is general for our chain. It should be usable by any node in theory or by any project without forcing them to depend in the node itself.

lemunozm:
Ok, that makes sense!

wischli:
> I disagree with moving the specs into the node folder. But no blocker.

I was thinking about this as well. I know at least two projects, which also moved the specs into the node directory and didn't raise this concern for that reason. But I definitely agree with out point that specs and the client are independent.

lemunozm:
Thanks for sharing the research!

lemunozm:
Maybe for now, we can keep it until we find the need to move this out (?)

PR #1634, opened 2023-11-30
Fast runtime change handle by pool

lemunozm:
# Description

The fast version of `RuntimeChange` exists to avoid waiting a week in testnets to make changes through the pool. Although the correct solution works, it implies a lot of boilerplate each time a new `Change` is added to the system. The "hack" should be handled just at one point of the process instead of for each change. 

This PR moves the action to the pool system under a new config type.

**WARNING**: This PR modifies the `RuntimeChange` type used in `development`, so this would require migration or living with it. It only affects to `development`.

### DIFF
--- a/pallets/pool-system/src/lib.rs
+++ b/pallets/pool-system/src/lib.rs
@@ -358,6 +358,11 @@ pub mod pallet {
 		#[pallet::constant]
 		type PoolDeposit: Get<Self::Balance>;
 
+		/// Enable fast runtime changes, overwriting the waiting time with a
+		/// fixed value. `None` value means the normal behaviour.
+		/// Used for testnets.
+		type FastChanges: Get<Option<Seconds>>;
+
 		/// The origin permitted to create pools
 		type PoolCreateOrigin: EnsureOrigin<Self::RuntimeOrigin>;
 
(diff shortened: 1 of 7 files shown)


lemunozm:
Required by:
- https://github.com/centrifuge/centrifuge-chain/pull/1633
- https://github.com/centrifuge/centrifuge-chain/pull/1629

wischli (approved):
That was fast 🚀 Great improvement, always happy about less boilerplate and more explicitness.

lemunozm:
Thanks for the review! 

I was just thinking that maybe with the From/TryInto macros the boilerplate just disappear without touching the pool. Let's see how it would work with the macros before merging it

pallets/pool-system/src/lib.rs
```diff
@@ -358,6 +358,11 @@ pub mod pallet {
 		#[pallet::constant]
 		type PoolDeposit: Get<Self::Balance>;
 
+		/// Enable fast runtime changes, overwriting the waiting time with a
+		/// fixed value. `None` value means the normal behaviour.
+		/// Used for testnets.
+		type FastChanges: Get<Option<Seconds>>;
```
mustermeiszer, line 364: I do not think we should add this as a pallet type. can we not fix this in the runtime itself?

lemunozm: Yes, I think so that by doing some macros I have in mind, we can avoid this and let decide the runtime but without boilerplate

mustermeiszer: I would favor that over having test-based types in the core pallet here.

lemunozm: Agree

PR #1637, opened 2023-12-01
Modify `RuntimeChange` to support easily new additions

lemunozm:
# Description

Currently, adding a new change to `RuntimeChange` implies a lot of boilerplate. With `From/TryInto` methods and duplicate the work for `fast::RuntimeChange` version. This PR tries to ease the process.

- Added a macro to add new `From/TryInto` methods
- `fast::RuntimeChange` type has benn removed, and now `RuntimeChange` allows itself to be parameterized with an `Options` type that gives it a specific behavior, i.e. a fast version of `RuntimeChange`.
- Moved to its own `changes` module.

## Required by:
- https://github.com/centrifuge/centrifuge-chain/pull/1633
- https://github.com/centrifuge/centrifuge-chain/pull/1629

### DIFF
--- a/runtime/common/src/changes.rs
+++ b/runtime/common/src/changes.rs
@@ -1,0 +1,104 @@
+use cfg_primitives::SECONDS_PER_WEEK;
+use codec::{Decode, Encode, MaxEncodedLen};
+use frame_support::RuntimeDebug;
+use pallet_loans::entities::changes::{Change as LoansChange, InternalMutation, LoanMutation};
+use pallet_pool_system::pool_types::changes::{PoolChangeProposal, Requirement};
+use scale_info::TypeInfo;
+use sp_runtime::DispatchError;
+use sp_std::vec;
+
+/// Auxiliar type to carry all pallets bounds used by RuntimeChange
+pub trait Changeable: pallet_loans::Config {}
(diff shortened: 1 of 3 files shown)


runtime/common/src/changes.rs
```diff
@@ -0,0 +1,104 @@
...
+use sp_runtime::DispatchError;
+use sp_std::vec;
+
+/// Auxiliar type to carry all pallets bounds used by RuntimeChange
+pub trait Changeable: pallet_loans::Config {}
+impl<T: pallet_loans::Config> Changeable for T {}
+
+/// A change done in the runtime, shared between pallets
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub enum RuntimeChange<T: Changeable, Options: Clone = ()> {
```
lemunozm, line 16: This is the key. If none option is passed, `RuntimeChange<T>`, the behavior is the expected one. If `FastDelay` option is passed, `RuntimeChange<T, FastDelay>` the behavior is hacked to be faster.

wischli (commented):
Thanks a ton for this incredible change! The boilerplate setup difference is ridiculous now: A few lines and that's it. Awesome work @lemunozm 🎉

PR #1640, opened 2023-12-01
Fix benchmark issue in development

lemunozm:
The version of `RuntimeChange` with the `FastDelay` option was not the special case required for benchmarking

[slack thread](https://kflabs.slack.com/archives/C04GJQAM9P0/p1701440777599979)


wischli (approved):
LGTM!

PR #1644, opened 2023-12-11
Docker enable latest tag

cdamian:
# Description

Always tag our docker images with the `latest` tag.

# Checklist:


.github/workflows/build-docker.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           images: centrifugeio/centrifuge-chain
           tags: |
             type=semver,pattern={{raw}},prefix=${{ matrix.target == 'test' && 'test-' || '' }}
-            type=raw,value=latest,enable=${{ github.event_name == 'release' }}
+            type=raw,value=latest,enable=true
```
wischli, line 42: This change should not be merged. The purpose of `latest` reflects the latest releases. We don't want node operators to run our dev images.

wischli (changes requested):
Thanks for the fix. The `latest` tage event should not be changed. It should indeed only be built during the `release` trigger for docker, i.e. when we draft a release. Unfortunately, that pipeline didn't get triggered for the latest v0.10.35 release. AFAICT because of missing GHA tokens.

.github/workflows/build-docker.yml
cdamian: OK, then I'll revert this and I guess we want to debug the issue of not having the `latest` tag for the last release.

cdamian: There are no `release` events for our repo it seems:

https://github.com/centrifuge/centrifuge-chain/actions?query=event%3Arelease

Maybe we want to have something like the trigger mentioned [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release):

```
on:
  release:
    types: [published]
 ```

wischli: We do have that trigger set up: https://github.com/centrifuge/centrifuge-chain/blob/ccbc3a158068b568045e1f74b61620252696889a/.github/workflows/build-docker.yml#L6-L7

Unfortunately, I don't have capacity to look into this. It should be related to this StackOverflow question/solution: https://stackoverflow.com/questions/69063452/github-actions-on-release-created-workflow-trigger-not-working

wischli (approved):
Thanks for looking into this!

PR #1645, opened 2023-12-11
Split concern for pool benchmark helper

lemunozm:
# Description

The previous `PoolBenchmarkHelper` utility made use of investment utilities. This forces you to track investment bounds through the code, which is ok for parts that will end to use investments, but for cases where you only want to create a pool, it is boilerplate and annoying to use, carrying information out of the problem scope. This PR splits the current trait into 2 depending on the usage you require.

## Changes and Descriptions

- Renamed `PoolBenchmarkHelper` to `FundedPoolBenchmarkHelper`, for creating funded pools.
- New `PoolBenchmarkHelper` for only creating just pools.

Required for OracleV2. https://github.com/centrifuge/centrifuge-chain/pull/1629

### DIFF
--- a/libs/mocks/src/pools.rs
+++ b/libs/mocks/src/pools.rs
@@ -64,22 +64,6 @@ pub mod pallet {
 		) {
 			register_call!(move |(a, b, c)| f(a, b, c));
 		}
-
-		pub fn mock_bench_create_pool(f: impl Fn(T::PoolId, &T::AccountId) + 'static) {
-			register_call!(move |(a, b)| f(a, b));
-		}
-
-		pub fn mock_bench_investor_setup(
-			f: impl Fn(T::PoolId, T::AccountId, T::Balance) + 'static,
-		) {
-			register_call!(move |(a, b, c)| f(a, b, c));
-		}
-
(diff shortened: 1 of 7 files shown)


libs/mocks/src/pools.rs
```diff
@@ -64,22 +64,6 @@ pub mod pallet {
...
-			f: impl Fn(T::PoolId, T::AccountId, T::Balance) + 'static,
-		) {
-			register_call!(move |(a, b, c)| f(a, b, c));
-		}
-
-		pub fn mock_bench_default_investment_id(
-			f: impl Fn(T::PoolId) -> T::TrancheCurrency + 'static + 'static,
-		) {
-			register_call!(f);
-		}
```
lemunozm, line 82: No longer needed if default mock implementation is default (which makes sense because you do not want to benchmark in testing)

cdamian (dismissed):

wischli (dismissed):
Thanks for untangling! LGTM

mustermeiszer (dismissed):

cdamian (approved):

lemunozm:
❤️

PR #1646, opened 2023-12-11
Fix Docker workflow triggers

gpmayorga:
This PR:
- Adds an on-demand trigger in case a manual docker tag is needed
- Adjusts when a release triggers a docker build and push
- Auto-comments on the release notes adding the docker tags


gpmayorga:
Screenshot 2023-12-11 at 13 51 10

This workflow should be working now. 
It will never trigger for DRAFT releases but every other action in publishing a pre-release or a release, or updating it will trigger a Docker build.

gpmayorga:
Screenshot 2023-12-11 at 13 51 10

Example of release tags

wischli (approved):
Thanks for fixing this!

PR #1652, opened 2023-12-12
Substitute mock-accountant macro by a mock builder pallet

lemunozm:
# Description

From the addition of this feature (https://github.com/foss3/runtime-pallet-library/pull/1), the mock accountant macro can be removed, and a mock builder pallet can be used instead.

## Changes and Descriptions

- Replace mock_accountant macro with a mock builder pallet.
- Replace `Rate` by `Quantity` (both represent the same type, but using `Quantity` for correctness).
  - Adapt consequently computed prices hardcoded values in tests
- Remove unused mocks from `cfg-test-utils`


lemunozm:
Requires https://github.com/foss3/runtime-pallet-library/pull/15 to work

pallets/investments/src/tests.rs
```diff
@@ -622,7 +622,7 @@ fn update_redeem_fails_when_collect_needed() {
 fn fulfillment_flow_for_everything_works() {
 	TestExternalitiesBuilder::build().execute_with(|| {
 		#[allow(non_snake_case)]
-		let PRICE: Rate = price_of(1, 2, 10);
+		let PRICE: Quantity = price_of(1, 2, 10);
```
mustermeiszer, line 625: Why the change? In pallet pools we use `Rate`. It is confusing to change it here now IMO

mustermeiszer (commented):
Besides the change to the `FixedPoint` type it looks good.

pallets/investments/src/tests.rs
lemunozm: This price is not "related" to the one used in `pallet-pool-system`. Actually, the own `pallet-investment` defines such type as `Quantity` under `T::BalanceRatio` (namely price). We do not want 27 decimals for them, we want 18. Basically I forgot to modify it when I updated all rates some months ago.

The rule we are using is prices: `Quantity`; rates (used by interests where decimals are quite significant): `Rate`.

libs/mocks/src/pools.rs
```diff
@@ -64,6 +68,42 @@ pub mod pallet {
...
+		) {
+			register_call!(move |(a, b, c)| f(a, b, c));
+		}
+
+		#[allow(non_snake_case)]
+		pub fn mock_InvestmentAccountant_withdraw(
+			f: impl Fn(&T::AccountId, T::TrancheCurrency, T::Balance) -> DispatchResult + 'static,
+		) {
+			register_call!(move |(a, b, c)| f(a, b, c));
+		}
```
lemunozm, line 106: The application of https://github.com/foss3/runtime-pallet-library/pull/16, @wischli 

Basically the pallet implements `deposit()` and `withdraw()` methods from two different traits. So we need to add `mock_<trait>_method()` to one of these `deposit()/withdraw()`.

In those cases, the `T::AccountId` used as generic for `InvestmentAccountant` was wrongly internally coded.

pallets/investments/src/tests.rs
```diff
@@ -1289,11 +1289,11 @@ fn fulfillment_partially_works_low_price() {
 		{
 			assert_eq!(
 				free_balance_of(investment_account(INVESTMENT_0_0), AUSD_CURRENCY_ID),
-				207245508982035928145
+				207245508982035928140
```
lemunozm, line 1292: Rounding issues due to modifying `Rate` (27 decimals) to `Quantity` (18 decimals).

pallets/investments/src/tests.rs
```diff
@@ -2781,7 +2783,7 @@ fn collecting_fully_works() {
 				INVESTMENT_0_0
 			));
 			assert_eq!(
-				n_last_event(3),
+				n_last_event(4),
```
lemunozm, line 2786: The expected event is behind another produced event of killing an account.

I think the rounding issues put the account under the existential deposit value, and the account is killed.

pallets/investments/src/benchmarking.rs
```diff
@@ -56,6 +56,9 @@ mod benchmarks {
 
 	#[benchmark]
 	fn update_invest_order() -> Result<(), BenchmarkError> {
+		#[cfg(test)]
+		crate::mock::configure_accountant_mock();
```
lemunozm, line 60: Completely astonished by the needed for this.

Without the above line, the test crashes because the mock closure methods are not defined. Nevertheless, `configure_accountant_mock()` is already called under `TextExternatilitiesBuilder::build()`, so I would expect to have already defined the mock methods. 

It seems like for benchmarking, somehow, the thread used to build the externality is not the same as the one used to benchmark. And then, the internal `thread_local` used by the mock builder doesn't correspond to the one used.

lemunozm: Actually, for testing the 4 benchmarking, just one call to `TextExternatilitiesBuilder::build()` is done

pallets/investments/src/tests.rs
mustermeiszer: Ahh, I see. Sorry, my bad you are right of course. I did overlook that the tranches are using `Quantity` too.

pallets/investments/src/benchmarking.rs
mustermeiszer: Interesting finding.

mustermeiszer (approved):
Thanks for the clarification!

mustermeiszer (approved):

PR #1653, opened 2023-12-15
Loans: Expand `TransferDebt` event with more info

lemunozm:
# Description

From https://kflabs.slack.com/archives/C04GJQAM9P0/p1702475864702369


mustermeiszer:
Changes look good! @hieronx should we already rename or do we want to keep it for now as embrio needs to adapt then too?

hieronx (approved):

PR #1656, opened 2023-12-19
Add a readiness check to the Docker container

gpmayorga:
# Description

This is a pre-requisite to be able to run our fullnode/archive infrastructure as an auto-scalable cluster. In order for a proxy or a load balancer to know if a fullnode is ready to accept traffic it needs to check a bit more than "is port 9933 open?".
This PR:
- Includes a script that will check if the node where it's running (localhost) is synched and has peers.
- Adds the script to the path: this is very specific to K8's [Readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes) which requires the "command check" to be runnable in the Docker context.

Reviewers: if you think about something more sophisticated or consider "isSyncheched and has peers" insuficient please add to the script. The important question it needs to answer is "Am I ready to recieve RPC and WS traffic?"

Extras on this PR:
- Push to Github registry as well as dockerHub (do not push to DH for PRs). You can view pushed images [here](https://github.com/orgs/centrifuge/packages?repo_name=centrifuge-chain) and [tags here](https://github.com/centrifuge/centrifuge-chain/pkgs/container/centrifuge-chain) this way you can test the docker container in your PR and see if it works before you merge


.github/workflows/build-docker.yml
```diff
@@ -11,7 +11,8 @@ on:
   pull_request:
     paths:
       - ".github/workflows/build-docker.yml"
-      - "docker/centrifuge-chain/Dockerfile"
+      - "docker/"
```
wischli, line 14: This includes adjustments to the `docker-compose` files which we have to touch for every Substrate upgrade. I don't think this should trigger a docker build. Why is this change necessary?

.github/workflows/build-docker.yml
```diff
@@ -22,26 +23,38 @@ jobs:
...
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@f95db51fddba0c2d1ec667646a06c2ce06100226 #v3
-      - name: DockerHub Login
-        uses: docker/login-action@v3
+
+      # - name: DockerHub Registry Login
+      #   uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d
+      #   with:
+      #     username: ${{ secrets.DOCKER_HUB_USERNAME }}
+      #     password: ${{ secrets.DOCKER_HUB_TOKEN }}
```
wischli, line 40: Can we entirely remove this? We shouldn't commit commented code unless not absolutely necessary. If so, please add a top comment explaining why this is commented out and when it should be used.

docker/centrifuge-chain/Dockerfile
```diff
@@ -27,43 +27,56 @@ FROM --platform=linux/amd64 docker.io/paritytech/ci-linux:production as builder
...
 
 # checks
 RUN ldd /usr/local/bin/centrifuge-chain && \
 	/usr/local/bin/centrifuge-chain --version
+RUN set -e && command -v check-node-ready && command -v entrypoint.sh
 
 EXPOSE 30333 9933 9944
 VOLUME ["/data"]
-ENTRYPOINT ["/centrifuge/entrypoint.sh"]
+ENTRYPOINT ["entrypoint.sh"]
```
wischli, line 81: AFAICS, both scripts are not in the workdir `centrifuge` but `centrifuge/scripts`. Am I missing something here or are we not using the correct paths for the `check-node-ready` and `entrypoint.sh` scripts?

wischli (commented):
Thanks for improving our clients! I have two quick question and one remark.

gpmayorga:
Doing some research I found a better way to do healthchecks without needing to compromise the Docker image for the parachain:
https://hub.docker.com/r/paritytech/ws-health-exporter

Closing this and opening a new PR with some of the enhancements of this PR without the readiness check

docker/centrifuge-chain/Dockerfile
gpmayorga: RUN chmod +x /centrifuge/scripts/* && \
		ln -s /centrifuge/scripts/* /usr/local/bin/
This will make the scripts executable and also add them to PATH so they can be executed from anywhere

PR #1659, opened 2023-12-26
remarks: Add remarks pallet

### LINKED ISSUE #1647: Feat: On-chain Remarks
### Description

The goal is to allow issuers to connect an `Event::Remark(Remark)` to a Call so that it is known that the remark belongs to the executed call.

### Research/based on
* Spec: https://centrifuge.hackmd.io/FSjABNk2RSW3NNLq9vIINg

* Increases complexity and coupling of pallets

* Adds more (loose) coupling

cdamian:
# Description

Adds a pallet that's used for adding specific remarks to a call.

Fixes #1647 

# Checklist:


pallets/remarks/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+		#[pallet::weight({
+			let dispatch_info = call.get_dispatch_info();
+			(T::WeightInfo::remark()
+				.saturating_add(dispatch_info.weight),
+			dispatch_info.class)
+		})]
+		pub fn remark(
+			origin: OriginFor<T>,
+			remarks: BoundedVec<T::Remark, T::MaxRemarks>,
+			call: Box<<T as Config>::RuntimeCall>,
```
lemunozm, line 88: Question:

How can we ensure that `call` is not a batch with other calls, i.e containing a remark call inside?

cdamian: Hmm, we could add some more trait constraints that would allow us to cast this into whatever calls that we want to check, but I'm not sure if we wanna do that. I guess this should be pretty simple and generic, and I doubt that the utility pallet is doing that for example.

cdamian: On second thought, not allowing a remark inside a remark makes sense, gonna look into adding that. Thank you!

cdamian: I added a filter, similar to the one in the utility pallet.

cdamian:
Pallet added here - https://github.com/foss3/runtime-pallet-library/pull/18

pallets/remarks/src/lib.rs
lemunozm: Great! Thanks!!

I ask that because the reason to not have the utility pallet was precisely that you could add a remark in a batch, IIRC

lemunozm:
Should this PR be closed and only reviewed/merged the one from `foss3`?

cdamian:
> Should this PR be closed and only reviewed/merged the one from `foss3`?

No, the plan is to add the remarks pallet to the `foss3` repo and then use that pallet in the runtime here.

runtime/development/src/lib.rs
```diff
@@ -1890,6 +1903,7 @@ construct_runtime!(
 		ForeignInvestments: pallet_foreign_investments::{Pallet, Storage, Event<T>} = 117,
 		OraclePriceFeed: pallet_oracle_feed::{Pallet, Call, Storage, Event<T>} = 118,
 		OraclePriceCollection: pallet_oracle_collection::{Pallet, Call, Storage, Event<T>} = 119,
+		Remarks: pallet_remarks::{Pallet, Call, Event<T>} = 130,
```
cdamian, line 1906: Note that we have to jump to 130 now.

mustermeiszer:
Started benchmarks manually. Can we also already bring this to mainnet please. Otherwise looks good!

runtime/development/src/lib.rs
wischli: I am running into the same issue in #1633 and propose append at the end of all indices instead of taking the rather slim `130 to 149` range. This way, we only introduce one jump. For the `130 to 149` range I fear that we occupy all indices by the end of the year.

I propose an arbitrary 250 start index for the second batch of internal pallets: https://github.com/centrifuge/centrifuge-chain/blob/08f80ce49df346d8796b8a88e7c84332185f1654/runtime/development/src/lib.rs#L1950-L1951

cdamian: OK, will change that now. Thank you!

cdamian:
I manually added the weights that we generated [here](https://github.com/centrifuge/centrifuge-chain/actions/runs/7407712989#artifacts)

mustermeiszer (approved):

PR #1661, opened 2023-12-27
Oracles: Protect from configuring with duplicated feeders

lemunozm:
# Description

We should avoid configuring the feeders with a vector to prevent the caller from adding duplicated feeders. Instead of checking duplications in runtime, we can ensure this behavior in the type system using a set.

* Depends on https://github.com/centrifuge/centrifuge-chain/pull/1658


cdamian (approved):
LGTM

NunoAlexandre (dismissed):
Way to go 👌

mustermeiszer (dismissed):
LGTM!

lemunozm:
Thanks! Waiting to #1658 first

mustermeiszer (approved):

PR #1662, opened 2023-12-28
Integration tests for proxies

lemunozm:
# Description

From this [slack conversation](https://kflabs.slack.com/archives/C04GJQAM9P0/p1703688113169509)

### DIFF
--- a/runtime/integration-tests/src/generic/cases/proxy.rs
+++ b/runtime/integration-tests/src/generic/cases/proxy.rs
@@ -1,0 +1,236 @@
+use cfg_primitives::Balance;
+use frame_support::traits::Get;
+use frame_system::RawOrigin;
+use sp_runtime::{traits::StaticLookup, DispatchResult};
+use xcm::v3::{MultiLocation, WeightLimit};
+
+use crate::{
+	generic::{
+		config::Runtime,
+		env::{Blocks, Env},
+		envs::{fudge_env::FudgeSupport, runtime_env::RuntimeEnv},
+		utils::{
+			self,
+			currency::{self, cfg, usd6, CurrencyInfo, Usd6},
+			genesis::{self, Genesis},
+			POOL_MIN_EPOCH_TIME,
+		},
+	},
+	utils::{accounts::Keyring, tokens::rate_from_percent},
+};
+
+const FROM: Keyring = Keyring::Charlie;
+const PROXY: Keyring = Keyring::Alice;
+const TO: Keyring = Keyring::Bob;
+
+const FOR_FEES: Balance = cfg(1);
+const TRANSFER_AMOUNT: Balance = usd6(100);
+
+fn configure_proxy_and_transfer<T: Runtime>(proxy_type: T::ProxyType) -> DispatchResult {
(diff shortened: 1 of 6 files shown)


runtime/integration-tests/src/generic/cases/proxy.rs
```diff
@@ -0,0 +1,236 @@
...
+crate::test_for_runtimes!([altair], altair_x_transfer_with_proxy_transfer);
+crate::test_for_runtimes!([altair], altair_x_transfer_with_proxy_borrow);
+crate::test_for_runtimes!([altair], altair_x_transfer_with_proxy_invest);
+
+crate::test_for_runtimes!([centrifuge], centrifuge_transfer_with_proxy_transfer);
+crate::test_for_runtimes!([centrifuge], centrifuge_transfer_with_proxy_borrow);
+crate::test_for_runtimes!([centrifuge], centrifuge_transfer_with_proxy_invest);
+crate::test_for_runtimes!([centrifuge], centrifuge_x_transfer_with_proxy_transfer);
+crate::test_for_runtimes!([centrifuge], centrifuge_x_transfer_with_proxy_borrow);
+crate::test_for_runtimes!([centrifuge], centrifuge_x_transfer_with_proxy_invest);
```
lemunozm, line 236: Sorry for this boilerplate, but we have no choice because each `ProxyType` is defined in each runtime, so they're different types. They can not be moved to runtime-common because the implementation uses the `RuntimeCall` of each runtime.

Actually, I'm thinking of moving the [generic runtime config](https://github.com/centrifuge/centrifuge-chain/blob/main/runtime/integration-tests/src/generic/config.rs#L45) used in integration tests to `runtime-common` to allow building things like that without copy-paste in all runtime, WDYT? @mustermeiszer @cdamian

runtime/integration-tests/src/generic/cases/proxy.rs
```diff
@@ -0,0 +1,236 @@
...
+
+fn configure_proxy_and_x_transfer<T: Runtime>(proxy_type: T::ProxyType) -> DispatchResult {
+	/*
+	let env = // TODO: initialize fudge env with xcm support
+	let call = // TODO: create a 'pallet_restricted_xtokens::transfer' call
+
+	configure_proxy_and_call::<T>(env, proxy_type, call);
+	*/
+	todo!()
+}
```
lemunozm, line 55: @cdamian I have no knowledge of the setup for a xcm transfer, could you finish this test for `pallet_restricted_xtokens`? It should be exactly the same pattern as the one already done for `pallet_restricted_tokens` but using `FudgeEnv` instead. 🙏🏻

lemunozm:
Ready to be reviewed and merged!

lemunozm:
Ready for final review!

runtime/integration-tests/src/generic/cases/proxy.rs
lemunozm: Tracking issue: https://github.com/centrifuge/centrifuge-chain/issues/1664

mustermeiszer (approved):
Sick! Thanks @lemunozm and @cdamian, thats really reassuring!

PR #1663, opened 2023-12-29
integration-tests: Add setup for proxy xtransfers tests

cdamian:


runtime/integration-tests/src/generic/cases/proxy.rs
```diff
@@ -44,14 +56,68 @@ fn configure_proxy_and_transfer<T: Runtime>(proxy_type: T::ProxyType) -> Dispatc
...
+	};
 
-	configure_proxy_and_call::<T>(env, proxy_type, call);
-	*/
-	todo!()
+	assert_ok!(orml_asset_registry::Pallet::<T>::register_asset(
+		<T as frame_system::Config>::RuntimeOrigin::root(),
+		meta,
+		Some(C::ID)
+	));
```
lemunozm, line 120: Question:

Why do we need to register the currency in the `asset_registry`? is it a requirement for `xtokens`?

cdamian: It's a requirement for `orml_xtokens` since it will eventually check if the currency that we are trying to transfer is registered.

lemunozm (approved):
Thanks for helping me with this! 🙌🏻

PR #1670, opened 2024-01-02
Oracles: Time validation by a max age threshold

lemunozm:
# Description

We want, as a requirement, to have oracle values validated with a "max age" threshold. Older oracles values must be considered outdated and it's an error to use them.

This error is generated when:
- Getting a collection where one oracle is outdated.
- Getting just an oracle value outdated.
- Updating a collection when an oracle value is outdated.

The threshold is optionally set by the collection by the admin

## Tasks
- [x] Implementation
- [x] Testing
- [x] Benchmarks
- [x] Updated runtimes

### DIFF
--- a/pallets/oracle-data-collection/src/lib.rs
+++ b/pallets/oracle-data-collection/src/lib.rs
@@ -450,7 +450,7 @@ pub mod pallet {
 			collection_id: &T::CollectionId,
 			timestamp: T::Timestamp,
 		) -> DispatchResult {
-			if let Some(duration) = CollectionInfo::<T>::get(collection_id).value_duration {
+			if let Some(duration) = CollectionInfo::<T>::get(collection_id).value_lifetime {
 				ensure!(
 					T::Time::now().ensure_sub(timestamp)? <= duration,
 					Error::<T>::OracleValueOutdated,
@@ -549,8 +549,8 @@ pub mod types {
 	pub struct CollectionInfo<T: Config> {
 		/// Maximum duration to consider an oracle value non-outdated.
 		/// An oracle value is consider updated if its timestamp is higher
-		/// than `now() - value_duration`
-		pub value_duration: Option<T::Timestamp>,
+		/// than `now() - value_lifetime`
+		pub value_lifetime: Option<T::Timestamp>,
 
 		/// Minimun number of feeders to succesfully aggregate a value.
 		pub min_feeders: u32,
@@ -559,7 +559,7 @@ pub mod types {
 	impl<T: Config> Default for CollectionInfo<T> {
 		fn default() -> Self {
 			Self {
-				value_duration: None,
+				value_lifetime: None,
 				min_feeders: 0,
 			}
 		}
--- a/pallets/oracle-data-collection/src/lib.rs
+++ b/pallets/oracle-data-collection/src/lib.rs
@@ -274,7 +274,7 @@ pub mod pallet {
 		) -> DispatchResult {
 			ensure_signed(origin)?;
 
-			let mut older_timestamp = T::Time::now();
+			let mut older_value_timestamp = T::Time::now();
 
 			let values = Keys::<T>::iter_key_prefix(collection_id)
 				.filter_map(|key| {
@@ -285,8 +285,8 @@ pub mod pallet {
 
 					match value {
 						Ok((value, timestamp)) => {
-							if timestamp < older_timestamp {
-								older_timestamp = timestamp;
+							if timestamp < older_value_timestamp {
+								older_value_timestamp = timestamp;
 							}
 							Some(Ok((key, (value, timestamp))))
 						}
@@ -305,7 +305,7 @@ pub mod pallet {
 				collection_id,
 				CachedCollection {
 					content: collection,
-					older_timestamp,
+					older_value_timestamp,
 				},
 			);
 
@@ -380,7 +380,7 @@ pub mod pallet {
 
 		fn collection(collection_id: &T::CollectionId) -> Result<Self::Collection, DispatchError> {
 			let collection = Collection::<T>::get(collection_id);
-			Self::ensure_valid_timestamp(collection_id, collection.older_timestamp)?; //TODO
+			Self::ensure_valid_timestamp(collection_id, collection.older_value_timestamp)?;
 			Ok(collection)
 		}
 
@@ -496,7 +496,7 @@ pub mod pallet {
 
 			Collection::<T>::mutate(collection_id, |cached| {
 				cached.content.try_insert(*key, aggregated_value).unwrap();
-				cached.older_timestamp = T::Time::now();
+				cached.older_value_timestamp = T::Time::now();
 			});
 		}
 	}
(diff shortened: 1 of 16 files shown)


pallets/oracle-data-collection/src/lib.rs
```diff
@@ -312,12 +352,18 @@ pub mod pallet {
...
 				.collect::<Result<Vec<_>, _>>()?;
 
-			T::AggregationProvider::aggregate(fed_values)
-				.ok_or(Error::<T>::KeyNotInCollection.into())
+			let (value, timestamp) = T::AggregationProvider::aggregate(fed_values)
+				.ok_or(Error::<T>::KeyNotInCollection)?;
+
+			Self::ensure_valid_timestamp(collection_id, timestamp)?;
+
+			Ok((value, timestamp))
```
lemunozm, line 360: I'm aggregating both values and timestamps and from the aggregated timestamp, I'm checking that it's not outdated.

Another option could be to filter outdated timestamps and aggregate them later.

lemunozm: Thinking a bit more. I think makes more sense the second option. To avoid aggregating over outdated values

lemunozm: Option 3:

- **Final timestamp** is the median of the timestamps,
- If the median is not outdated then the **final value** is the median of the non-outdated Oracle values.

I see this option quite accurate, but the timestamp probably would be older than the average of used values.

@mustermeiszer

lemunozm: Final solution (after Frederic sync):
- Filter outdated oracles until a limit N.
  - If the limit is overpassed, then `OracleOutdatedError`.
- value & timestamp median of the non-outdated oracles

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -460,17 +541,58 @@ pub mod types {
...
+	#[scale_info(skip_type_params(T))]
+	pub struct CollectionInfo<T: Config> {
+		/// Maximum duration to consider an oracle value non-outdated.
+		/// An oracle value is consider updated if its timestamp is higher
+		/// than `now() - value_lifetime`
+		pub value_lifetime: Option<T::Timestamp>,
+
+		/// Minimun number of feeders to succesfully aggregate a value.
+		pub min_feeders: u32,
+	}
```
lemunozm, line 557: To catch all casuistics well, I've done a new type that is used as a parameter for the new extrinsic `set_collection_info()` to configure a collection with some properties.

`min_feeders` force to have at least `min_feeders` per key in a collection. That means that the feeder should have fed and should have fed recently if `value_lifetime` is different than `None`

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -312,12 +357,31 @@ pub mod pallet {
...
+				.ok_or(Error::<T>::KeyNotInCollection)?;
+
+			Ok((value, timestamp))
 		}
 
-		fn collection(collection_id: &T::CollectionId) -> Self::Collection {
-			CachedCollection(Collection::<T>::get(collection_id))
+		fn collection(collection_id: &T::CollectionId) -> Result<Self::Collection, DispatchError> {
+			let collection = Collection::<T>::get(collection_id);
+			Self::ensure_valid_timestamp(collection_id, collection.older_timestamp)?; //TODO
```
mustermeiszer, line 383: What is left TOdo here?

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -460,17 +541,58 @@ pub mod types {
...
+
+		/// Minimun number of feeders to succesfully aggregate a value.
+		pub min_feeders: u32,
+	}
+
+	impl<T: Config> Default for CollectionInfo<T> {
+		fn default() -> Self {
+			Self {
+				value_lifetime: None,
+				min_feeders: 0,
```
mustermeiszer, line 563: Lets use `min_feeders` = 1 here, otherwise, this will be a problem on the aggregation I would ugess, right?

pallets/oracle-data-collection/src/lib.rs
```diff
@@ -460,17 +541,58 @@ pub mod types {
...
 	/// A collection cached in memory
-	#[derive(Clone)]
-	pub struct CachedCollection<T: Config>(
-		pub BoundedBTreeMap<T::OracleKey, OracleValuePair<T>, T::MaxCollectionSize>,
-	);
+	#[derive(Encode, Decode, Clone, TypeInfo, RuntimeDebug, MaxEncodedLen)]
+	#[scale_info(skip_type_params(T))]
+	pub struct CachedCollection<T: Config> {
+		pub content: BoundedBTreeMap<T::OracleKey, OracleValuePair<T>, T::MaxCollectionSize>,
+		pub older_timestamp: T::Timestamp,
```
mustermeiszer, line 573: NIT: Naming `older_timestamp` confuses me.

pallets/oracle-data-collection/src/lib.rs
lemunozm: - If `0`, it means that the collection may not have all the expected keys. It's more general.
- If `1`, it means that the collection must have all expected keys.

The current built-in aggregation, with `0`, errors out with `KeyNotInCollection` because it can not aggregate with an empty list, but other `Aggregation` implementations could exist, setting a default value or similar if the list is empty.

For our concern, both `0` and `1` mean the same; it will give an error if there are no values for a key.

pallets/oracle-data-collection/src/lib.rs
lemunozm: Something I should have been removed 😅. It's already fixed.

pallets/oracle-data-collection/src/lib.rs
lemunozm: Any suggestion to indicate the oldest timestamp in the collection?

pallets/oracle-data-collection/src/lib.rs
lemunozm: BTW, good eye asking this!

mustermeiszer: Perfect! Then that's super fine.

pallets/oracle-data-collection/src/lib.rs
mustermeiszer: Last_updated maybe?

lemunozm: I understand `last_updated` as the timestamp of the last time the collection was updated. But we want to express that this indicates the timestamp of the older oracle value in the collection. It's difficult to find an appropriate name for it 😆

lemunozm: Maybe `older_value_timestamp` would better express that it belongs to a value of the collection instead of the whole collection?

mustermeiszer: Fine for me

lemunozm:
Should be ready then! Thanks for your fast review @mustermeiszer 🙌🏻

mustermeiszer (approved):

PR #1671, opened 2024-01-03
Rename from `pallet-oracle-data-collection` to `pallet-oracle-collection`

lemunozm:
# Description

Renaming to a more concise pallet name


lemunozm:
Incredible the amount of files one needs to touch to just change a pallet name

mustermeiszer (approved):
True, that is quite tedious...

PR #1672, opened 2024-01-04
Oracles: Move feeders to collection level

lemunozm:
# Description

Move the feeder concept by collection instead of by key.

### DIFF
--- a/pallets/oracle-collection/src/lib.rs
+++ b/pallets/oracle-collection/src/lib.rs
@@ -237,6 +237,7 @@ pub mod pallet {
 				.map_err(|_| Error::<T>::NoOracleCollectionChangeId)?;
 
 			CollectionInfo::<T>::insert(collection_id, info.clone());
+			Collection::<T>::remove(collection_id);
 
 			Self::deposit_event(Event::<T>::UpdatedCollectionInfo {
 				collection_id,
--- a/pallets/oracle-collection/src/lib.rs
+++ b/pallets/oracle-collection/src/lib.rs
@@ -44,22 +44,15 @@ pub use weights::WeightInfo;
 pub mod pallet {
 	use cfg_traits::{changes::ChangeGuard, data::DataRegistry, PreConditions, ValueProvider};
 	use frame_support::{
-		pallet_prelude::*,
-		storage::{
-			bounded_btree_map::BoundedBTreeMap, bounded_btree_set::BoundedBTreeSet, transactional,
-		},
-		traits::Time,
+		pallet_prelude::*, storage::bounded_btree_map::BoundedBTreeMap, traits::Time,
 	};
 	use frame_system::pallet_prelude::*;
-	use sp_runtime::{
-		traits::{EnsureAddAssign, EnsureSub, EnsureSubAssign, Zero},
-		TransactionOutcome,
-	};
+	use sp_runtime::traits::{EnsureAddAssign, EnsureSub, EnsureSubAssign, Zero};
 	use sp_std::{collections::btree_map::BTreeMap, vec::Vec};
 
 	use crate::{
 		traits::AggregationProvider,
-		types::{self, CachedCollection, Change, KeyInfo, OracleValuePair},
+		types::{self, CachedCollection, Change, OracleValuePair},
 		weights::WeightInfo,
 	};
 
@@ -142,7 +135,7 @@ pub mod pallet {
 		T::CollectionId,
 		Blake2_128Concat,
 		T::OracleKey,
-		KeyInfo<T>,
+		u32,
 		ValueQuery,
 	>;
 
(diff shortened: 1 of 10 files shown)


lemunozm:
Already compiling with `cargo check -p pallet-oracle-collection`

lemunozm:
It has become much more simple than before, IMO, and I would say it fits our requirements (by now).

If `update_collection()` is called and there is a key under use that has no associated feeder value for it, the method fails.

lemunozm:
This is ready @mustermeiszer  🙌🏻

pallets/oracle-collection/src/lib.rs
```diff
@@ -217,43 +208,34 @@ pub mod pallet {
...
 		) -> DispatchResult {
 			ensure_signed(origin)?;
 
-			let Change::Feeders(key, feeders) = T::ChangeGuard::released(collection_id, change_id)?
+			let Change::CollectionInfo(info) = T::ChangeGuard::released(collection_id, change_id)?
 				.try_into()
 				.map_err(|_| Error::<T>::NoOracleCollectionChangeId)?;
 
-			Self::update_feeders(collection_id, key, feeders.clone())?;
+			CollectionInfo::<T>::insert(collection_id, info.clone());
```
mustermeiszer, line 234: I think we should forbid any non-allowed feeder to feed any further values at this point. Basically, iterate over all keys in the storage and remove unknown feeders.

mustermeiszer (commented):
I have thought about the iteration of the `Keys` storage during this review and this is a possibly unbounded iteration. 

We should use `CountedStorageMap` instead OR restrict `register_id` to not store anymore keys if `MaxKeysPerCollection` is reached.

pallets/oracle-collection/src/lib.rs
lemunozm: This is a very good point

> Basically, iterate over all keys in the storage and remove unknown feeders.

I would say this could be more complex because some existing keys could have lost some of their feeders (but not all), and then the aggregated values could have changed or not been validated.

Could we consider that if we call `apply_update_collection_info()`, then you should call later `update_collection()` to reflect those changes? We can understand the current collection as the values the collection had at the moment of updating. If we see this in such a way, then there is no problem if feeders are modified because the collection at that point contained such previous feeders and values.

lemunozm: If you agree with me, it could be interesting to add a `last_updated` field to the collection to know if the content of the collection reflects a moment in the past.

mustermeiszer: > to add a last_updated field to the collection to know if the content of the collection reflects a moment in the past.

I thought that is already the case?

mustermeiszer: >Could we consider that if we call apply_update_collection_info(), then you should call later update_collection() to reflect those changes? We can understand the current collection as the values the collection had at the moment of updating. If we see this in such a way, then there is no problem if feeders are modified because the collection at that point contained such previous feeders and values.

Yes, that is a good idea. `update_collection` will filter them already, right?

mustermeiszer: Maybe, `apply_update_collection_info` should just purge the `CachedCollection` storage?

pallets/oracle-collection/src/lib.rs
```diff
@@ -385,67 +351,67 @@ pub mod pallet {
...
-			key: T::OracleKey,
-			f: impl FnOnce(&mut KeyInfo<T>) -> DispatchResult,
-		) -> DispatchResult {
-			Keys::<T>::mutate_exists(collection_id, key, |maybe_info| {
-				let info = maybe_info.get_or_insert(Default::default());
+							counter.ensure_sub_assign(1)?;
+
+							if counter.is_zero() {
+								*maybe_counter = None;
+							}
```
mustermeiszer, line 400: Is this correct? Isn't usage refs, unequal `usafe_refs`?

pallets/oracle-collection/src/lib.rs
lemunozm: I think it makes more sense to purge it 👍🏻

lemunozm: > I thought that is already the case?

The timestamp stored in the collection is the timestamp of the older Oracle value, not the timestamp of when the collection was outdated. So now we will store both

pallets/oracle-collection/src/lib.rs
lemunozm: I do not understand the question; could you extend it?

To recap:
- `usage_refs` corresponds to how many times a single key is used.
- `counter` corresponds to how many keys are used in the collection

pallets/oracle-collection/src/lib.rs
mustermeiszer: So purge or store outdated timestamp?

pallets/oracle-collection/src/lib.rs
mustermeiszer: Ahh, I thought you are removing the key, but you are removing the counter. Get it.

pallets/oracle-collection/src/lib.rs
lemunozm: I think we can use both. Store the `last_timestamp` each time we call `update_collection()` and purge when apply.

mustermeiszer (approved):
Thanks for your patience! Thats looks like a good iteration to the final version!

lemunozm:
Thanks for all your feedback! Very appreciated and accurate! 🙌🏻

PR #1674, opened 2024-01-05
OrderBook: Add market prices using Oracles

lemunozm:
# Description

Improve `pallet-order-book` to swap currencies with a market price.

## Change list
- For a full list of the breaking changes, see: https://kflabs.slack.com/archives/C04GARZ532T/p1706864698973649


lemunozm:
Compiling with `cargo check -p pallet-order-book`

pallets/order-book/src/lib.rs
```diff
@@ -601,38 +617,57 @@ pub mod pallet {
...
 
+			let market_price = T::PriceProvider::get(
+				&T::MarketFeederId::get(),
+				&(order.asset_in_id, order.asset_out_id).into(),
+			)?
+			.ok_or(Error::<T>::MarketPriceNotFound)?;
+
+			if market_price > order.max_sell_price {
+				Err(Error::<T>::HighMarketPrice)?
+			}
```
lemunozm, line 628: Will be removed if finally we remove the `max_sell_price` protection

pallets/order-book/src/lib.rs
```diff
@@ -387,7 +403,7 @@ pub mod pallet {
 			asset_in: T::AssetCurrencyId,
 			asset_out: T::AssetCurrencyId,
 			buy_amount: T::Balance,
-			price: T::SellRatio,
+			max_sell_price: T::SellRatio,
```
lemunozm, line 406: I feel the current `place_order()` API is enough by now, and there is no need to split. 

You choose "buy" or "sell" specifying the order on `currency_in` and `currency_out`. And the `max_sell_price` will be `1 / max_sell_price` for the cases where you want the opposite swap (as un Uniswap in this regard).

WDYT? @mustermeiszer @wischli

pallets/order-book/src/lib.rs
```diff
@@ -665,12 +696,35 @@ pub mod pallet {
...
+					let price = T::PriceProvider::get(
+						&T::MarketFeederId::get(),
+						&(currency_out, currency_in).into(),
+					)?
+					.ok_or(Error::<T>::MarketPriceNotFound)?;
+
+					T::SellRatio::one().ensure_div(price)?
+				}
+			})
+		}
```
lemunozm, line 726: This is the key to getting it to work with just one method, conserving the `place_order()` API.

lemunozm: Depending on how you specify the currency parameters, you'll obtain the market price or the inverse one

libs/traits/src/lib.rs
```diff
@@ -549,8 +556,8 @@ pub trait TokenSwaps<Account> {
 		account: Account,
 		currency_in: Self::CurrencyId,
 		currency_out: Self::CurrencyId,
-		buy_amount: Self::Balance,
-		sell_rate_limit: Self::SellRatio,
+		amount_out: Self::Balance,
+		price: OrderPrice<Self::SellRatio>,
```
lemunozm, line 560: These are the API changes:
- `amount_out` is the `sell_amount`. I've decided to call it `amount_out` to show that it's an amount measured in `currency_out`, but I'm open to suggestions.
- The `OrderPrice` enum defines how the price of the order will be computed.

pallets/order-book/src/lib.rs
```diff
@@ -571,87 +590,86 @@ pub mod pallet {
...
+				&order.placing_account,
+				amount_in,
 				Preservation::Expendable,
 			)?;
 
 			T::FulfilledOrderHook::notify_status_change(
 				order.order_id,
 				Swap {
-					amount: buy_amount,
+					amount: todo!(),
```
lemunozm, line 672: @wischli I'm not sure what I should put here after these changes.

Before, it was expected to be the `buy_amount`, but now we have no way to know it for market prices. Should I change it to `amount_out`? (which is `sell_amount`?)

pallets/order-book/src/lib.rs
```diff
@@ -1020,7 +1016,7 @@ pub mod pallet {
 		fn get_order_details(order: Self::OrderId) -> Option<Swap<T::Balance, T::AssetCurrencyId>> {
 			Orders::<T>::get(order)
 				.map(|order| Swap {
-					amount: order.buy_amount,
+					amount: todo!(),
```
lemunozm, line 1019: And here too

pallets/order-book/src/lib.rs
```diff
@@ -259,7 +281,7 @@ pub mod pallet {
 	///  - key1 -> AssetIn
 	///  - key2 -> AssetOut
 	///
-	/// Stores the minimum `buy_amount` of `asset_in` when buying
+	/// Stores the minimum `amount_out` of `asset_out` when buying
```
lemunozm, line 284: Just to notify, min_fulfillment amount is denominated in `currency_out`

pallets/order-book/src/lib.rs
```diff
@@ -721,123 +730,80 @@ pub mod pallet {
...
-					order.min_fulfillment_amount = min_fulfillment_amount;
-					order.max_sell_amount = max_sell_amount;
-
-					validate(order)?;
-
-					Ok(max_sell_amount)
-				},
-			)?;
+			Orders::<T>::insert(order.order_id, order.clone());
+			UserOrders::<T>::insert(order.placing_account, order.order_id, order.clone());
```
lemunozm, line 799: NIT: I feel it is much cleaner and less boilerplate if inserted twice instead of mutating both storage

pallets/order-book/src/lib.rs
```diff
@@ -175,6 +175,18 @@ pub mod pallet {
...
+
+		/// Account used to feed with market prices
+		type MarketFeederId: Get<Self::FeederId>;
+
+		/// Identification for a market price
+		type Pair: From<(Self::AssetCurrencyId, Self::AssetCurrencyId)>;
+
+		/// A way to obtain prices for market pairs
+		type PriceProvider: ValueProvider<Self::FeederId, Self::Pair, Value = Self::SellRatio>;
+
```
lemunozm, line 189: Right now, only one feeder (`MarketFeederId`) can feed. `PriceProvider` will be `pallet-oracle-feed` (which does not have validation & aggregation). 

In order to have the aggregation and validation features from `pallet-oracle-collection`, I was thinking of making `pallet-oracle-collection` instantiable. One instance will be used by `pallet-loans` where `CollectionId` is `PoolId`. And another instance will be used by `pallet-order-book`, where `CollectionId` can be something different than a pool id, like a `MarketId`. 

That way, we configure `pallet-order-book` with a `PriceProvider` which is a `pallet-oracle-collection` where exists a collection of market oracle values, already aggregated and validated. `ChangeGuard` of this instance of `pallet-oracle-collection` can be something different than `pool-system`, which always allows the change for this kind of admin.

Just a though for a next step improvement, but let me your thoughts @hieronx @mustermeiszer

libs/traits/src/lib.rs
wischli: I am supporting `amount_out` to reduce confusion.

pallets/order-book/src/lib.rs
```diff
@@ -188,23 +200,33 @@ pub mod pallet {
 	/// Contains fields relevant to order information
 	#[derive(Clone, Copy, Debug, Encode, Decode, Eq, PartialEq, MaxEncodedLen, TypeInfo)]
 	pub struct Order<OrderId, AccountId, AssetId, ForeignCurrencyBalance, SellRatio> {
+		/// Unique Id for this order
 		pub order_id: OrderId,
+
+		/// Associated account to this order
 		pub placing_account: AccountId,
+
+		/// Asset expected to receive to the account
```
wischli, line 209: Thanks for adding rustdocs!

pallets/order-book/src/lib.rs
```diff
@@ -571,87 +558,86 @@ pub mod pallet {
...
+			let amount_in =
+				Self::convert_with_ratio(order.asset_out_id, order.asset_in_id, price, amount_out)?;
 
-			let sell_amount = Self::convert_with_ratio(
-				order.asset_in_id,
-				order.asset_out_id,
-				order.max_sell_rate,
-				buy_amount,
-			)?;
 			let remaining_buy_amount = order
```
wischli, line 596: Nit: Should be name `remaining_sell_amount`
```suggestion
			let remaining_sell_amount = order
```

pallets/order-book/src/lib.rs
wischli: It should be `amount_in` because this value is needed to update investment states with pending swaps. The outgoing amount is already known.  The `amount_out` part does not help too much because then Foreign Investments would have to determine the `amount_in` from the `amount_out`.

pallets/order-book/src/lib.rs
```diff
@@ -664,13 +650,36 @@ pub mod pallet {
...
+				None => {
+					let price = T::PriceProvider::get(
+						&T::MarketFeederId::get(),
+						&(currency_out, currency_in).into(),
+					)?
+					.ok_or(Error::<T>::MarketPriceNotFound)?;
+
+					T::SellRatio::one().ensure_div(price)?
+				}
+			})
```
wischli, line 680: Nit: Could we determine final price outside the returning `Ok` scope? Seems preferable to me to `unwrap_or` the above price which is `Option<T::SellRatio>`.

pallets/order-book/src/lib.rs
```diff
@@ -721,123 +730,80 @@ pub mod pallet {
...
+			min_amount_out: T::Balance,
 		) -> DispatchResult {
-			let max_sell_amount = <Orders<T>>::try_mutate_exists(
-				order_id,
-				|maybe_order| -> Result<T::Balance, DispatchError> {
-					let order = maybe_order.as_mut().ok_or(Error::<T>::OrderNotFound)?;
+			let min_fulfillment_amount_out = T::DecimalConverter::to_asset_balance(
+				T::MinFulfillmentAmountNative::get(),
+				order.asset_out_id,
+			)?;
```
wischli, line 767: NOTE: As raised on Slack, this solely applies redenomination but does not determine whether the outgoing amount is worth 10 CFG. I suppose I did not bother with this because CFG internally is treated as 1 USD and we assume the OrderBook to solely deal with stable currencies. Not sure if we want to make it more explicit.

pallets/order-book/src/lib.rs
wischli: This way, we are dropping the existence check of the order which is otherwise performed by `try_mutate_exists`.

pallets/order-book/src/lib.rs
wischli: Right now, FI only checks the asset ids and does not require the amount. I suppose in an earlier version of FI, the amount was required. I propose to remove the amount entirely from the return type and only report back the `asset_out._d` and `asset_in_id`. If we want to keep it, it should be `amount_out`.

wischli (commented):
A few comments. Did not see anything I disagree with, great work!

pallets/order-book/src/lib.rs
lemunozm: Actually, to follow the new convention I would say: `remaining_amount_out`

pallets/order-book/src/lib.rs
lemunozm: Previously, the `buy_amount` or `amount_in` was known. But now with a variable market price, we do not know `amount_in`. It meant that you still need to fulfill `amount_in`. Now `amount_in` corresponds to the current already fulfilled amount. It's different to what it meant before.

pallets/order-book/src/lib.rs
lemunozm: Thanks for the details on this!

pallets/order-book/src/lib.rs
lemunozm: The existence is already checked when the order is previously retrived, in `Orders::<T>::get()`.

pallets/order-book/src/lib.rs
lemunozm: Actually, that was my first attempt, doing it with `unwrap_or_else(|| ..)` to avoid extra computation in case we do not take that branch. 

Nevertheless, because the inner closure would return an error, it became messy to get the value correctly.

pallets/order-book/src/lib.rs
wischli: But the `amount_in` (resp. `buy_amount`) is known here because we have fulfilled an order?! The purpose of the hook is to inform FI that a (partial) swap was fulfilled for which amount.

pallets/order-book/src/lib.rs
wischli: Yes, this is a legacy issue. In order to make a currency conversion, we would have to require conversion rates for `CFG <-> asset_out_id` which increases the overhead. Maybe adding a comment `// Implicitly assuming 1 CFG = 1 asset_out for min fulfilment amount` suffices here because implicitly treating CFG as 1 unit of stable currency is fine for now for the purpose of determining the minimum fulfilment amount?!

WDYT @mustermeiszer?

pallets/order-book/src/lib.rs
wischli: > Now `amount_in` corresponds to the current already fulfilled amount. It's different to what it meant before.

I disagree. In L625 we determine the `amount_in` from the `amount_out` which is the fulfilled amount at this point in time (which is reflected by determining the remaining order amount by reducing its previous value by `amount_out` in L628) : 
```
let amount_in =
				Self::convert_with_ratio(order.asset_out_id, order.asset_in_id, price, amount_out)?;
```

So `amount_in` is exactly the fulfilled amount of incoming asset which should be reported to the consumer of this hook (FI) to inform that a (partial) swap was fulfilled for that incoming amount.

pallets/order-book/src/lib.rs
```diff
@@ -175,6 +175,18 @@ pub mod pallet {
 			Error = DispatchError,
 		>;
 
+		/// Type for a market price feeder
+		type FeederId: Parameter + Member + Ord + MaxEncodedLen;
+
+		/// Account used to feed with market prices
+		type MarketFeederId: Get<Self::FeederId>;
```
lemunozm, line 182: Do we know the Account ID used to feed market values upfront? Or should it be the root/council who determines the account using an extrinsic?

lemunozm: @mustermeiszer

pallets/order-book/src/lib.rs
lemunozm: So, we need to choose between:
- Used 1 unit of `currency_out` as the min amount (current behavior, I think)
- Used a conversion to require an amount of `currency_out` equivalent to 1 CFG.

pallets/order-book/src/lib.rs
lemunozm: You're right! I was seeing this wrong. I thought the previous value was `order.buy_amount,`, but it's the `buy_amount` used to fulfill. So translated now is just `amount_in` as you say.

 Thanks a lot for guiding me to the correct reasoning! 🙌🏻

pallets/order-book/src/lib.rs
lemunozm: Given that `Swap` always expects the `amount` as `currenc_in`, I think it can be interesting to let this with `order.amount_in` value, which means the current fulfilled amount of this order. And makes sense for an API call called `get_order_details()`.  WDYT?

libs/traits/src/lib.rs
mustermeiszer: But I think `price` is really `buy_price_limit`, right?

mustermeiszer: Docs are not adapted.

mustermeiszer (commented):
Need to look at more depth tomorrrow, but what I can already say is that

* We need an additional extrinsic that is callable by `T::AdminOrigin` and allows to set a `T::FeederId`

libs/traits/src/lib.rs
lemunozm: Yes, I need a final self-review updating all docs, thanks for the reminder

lemunozm: I've modified the name because, as far as I can see, there is no "limit". You always buy/sell at the exact price you put there. A fixed price if `OrderPrice::Custom` is chosen, or the exact market price if `OrderPrice::market` is chosen.

If we want to add a limit order in the future, we should probably create another variant as `OrderPrice::Limit(price)`, which already holds the "limit" concept. WDYT?

lemunozm: Which I should indicate probably is the price direction, if it is a `buy_price` or a `sell_price` 👍🏻

pallets/order-book/src/lib.rs
wischli: Sounds good to me!

libs/traits/src/lib.rs
lemunozm: As another thought, I was thinking of calling it `conversion_rate` as a rate to pass from `currency_out` to `currency_in.` I feel like buy/sell concepts here only make things hard to understand. WDYT?

pallets/order-book/src/lib.rs
mustermeiszer: Sounds like a good idea for the next iteration. If you do not want to introduce the extrinsic and storage for setting the feeder dynamiclaly in this iteration, I would also be fine with using another instance of `collection` already in this run.

pallets/order-book/src/lib.rs
mustermeiszer: @lemunozm can you take care of writing a digest of the changes of the API for the apps team?

pallets/order-book/src/lib.rs
```diff
@@ -424,39 +428,22 @@ pub mod pallet {
...
-					Self::is_valid_order(
-						order.asset_in_id,
-						order.asset_out_id,
-						order.buy_amount,
-						order.max_sell_rate,
-						order.min_fulfillment_amount,
-						min_amount,
-					)
-				},
+				TradingPair::<T>::get(&order.asset_in_id, &order.asset_out_id)?,
```
mustermeiszer, line 446: We are defaulting the `min_fullfillment_amount` here to the `min_order_amount`, right? That is your idea?

pallets/order-book/src/lib.rs
```diff
@@ -847,36 +798,40 @@ pub mod pallet {
...
 				*n = n.ensure_add(T::OrderIdNonce::one())?;
 				Ok::<_, DispatchError>(())
 			})?;
 
-			let max_sell_amount =
-				Self::convert_with_ratio(currency_in, currency_out, sell_rate_limit, buy_amount)?;
+			let order_id = <OrderIdNonceStore<T>>::get();
+
+			let min_fulfillment_amount_out = T::DecimalConverter::to_asset_balance(
+				T::MinFulfillmentAmountNative::get(),
```
mustermeiszer, line 813: Wasn't there a discussion on slack regarding this? At least how I thing this is working, that should be fine, right?

pallets/order-book/src/lib.rs
lemunozm: No, I changed the `unner_updat_order()` method. `min_fulfillment_amount` is got from the inside of the function, and `min_order_amount` is passed from the outside.

There is no change in the logic of `min_fulfillment_amount` or `min_order_amount`; I just moved some things around to simplify the validation of these methods.

pallets/order-book/src/lib.rs
lemunozm: See the previous comment; it's working as before, and no changes did regarding this. I just move this call into the `inner_place_order()` function.

pallets/order-book/src/lib.rs
lemunozm: I'll evaluate if I have time to use the collection. I probably have a new PR over this if I have time to introduce it before the deadline.

pallets/order-book/src/lib.rs
lemunozm: Sure! I'll do it when I have a solid implementation of this

pallets/order-book/src/lib.rs
lemunozm: First approach: extrinsic. Second approach: oracle-collection

pallets/order-book/src/lib.rs
mustermeiszer: Cool.

pallets/order-book/src/lib.rs
mustermeiszer: Cool

libs/traits/src/lib.rs
mustermeiszer: You are right. I think keeping `price` is a good choice. No blocker from my side.

lemunozm:
Note: in case we need again the default inverse ratios here is the revert commit: [ae20e66](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/ae20e66d677e42f33848e4f3347633c1972bbc35)

mustermeiszer (approved):
Looks good! Thanks!

pallets/order-book/src/lib.rs
```diff
@@ -571,87 +559,102 @@ pub mod pallet {
...
-				sell_amount,
+				amount_in,
 				Preservation::Expendable,
 			)?;
 
 			T::FulfilledOrderHook::notify_status_change(
 				order.order_id,
 				Swap {
-					amount: buy_amount,
+					amount: amount_in,
```
wischli, line 657: Maybe we add the fulfilled `amount_out` here as well such that having `Swap` as status type is not sufficient. AFAICS, the we need the `amount_in` at least for `update_investment`. However, we also need `amount_out` for the pending swap state.

If Foreign Investment didn't store pending swaps and instead pulled these directly from the Orderbook, these changes would require less refactoring. It is what it is. One thing after the other.

pallets/order-book/src/lib.rs
```diff
@@ -571,87 +551,111 @@ pub mod pallet {
...
+				Precision::Exact,
+			)?;
+
+			if T::TradeableAsset::balance(order.asset_out_id, &order.placing_account) < amount_out {
+				Err(DispatchError::Token(TokenError::FundsUnavailable))?
+			}
+
+			if T::TradeableAsset::balance(order.asset_in_id, &fulfilling_account) < amount_in {
+				Err(DispatchError::Token(TokenError::FundsUnavailable))?
+			}
```
lemunozm, line 638: This is sadly needed to protect against `placing_account` == `fulfilling_account`.

mustermeiszer: Would there be a possible exploit here?

lemunozm: Yes if we do not add this check.

The `placing_account` could fulfill their own order even if they have no funds.

mustermeiszer (changes requested):
We are currently having quite some problems with the `MinFulfillment` usage on the trait side. 

I think, we can keep the concept on the public api side, but we have to remove the check and enforcement of it on the trait side, so that FI can update and place orders freely, just that if the `amount_out` is bigger than `MinFullfillment` the `order.min_fulfillment_amount_out` should be the `min` of both.

pallets/order-book/src/lib.rs
```diff
@@ -721,123 +724,83 @@ pub mod pallet {
...
-				|maybe_order| -> DispatchResult {
-					let order = maybe_order.as_mut().ok_or(Error::<T>::OrderNotFound)?;
-					order.buy_amount = buy_amount;
-					order.max_sell_rate = sell_rate_limit;
-					order.min_fulfillment_amount = min_fulfillment_amount;
-					order.max_sell_amount = max_sell_amount;
-					Ok(())
-				},
-			)?;
+			Self::validate_amount(amount_out, min_fulfillment_amount_out, min_amount_out)?;
```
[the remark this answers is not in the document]
lemunozm: This is automatically protected by the `min_order_amount` and `min_fulfilled_amount`. Do you think we should add this check?

pallets/order-book/src/lib.rs
```diff
@@ -721,123 +724,83 @@ pub mod pallet {
...
-					order.min_fulfillment_amount = min_fulfillment_amount;
-					order.max_sell_amount = max_sell_amount;
-
-					validate(order)?;
-
-					Ok(max_sell_amount)
-				},
+			let min_fulfillment_amount_out = T::DecimalConverter::to_asset_balance(
+				T::MinFulfillmentAmountNative::get(),
+				order.asset_out_id,
```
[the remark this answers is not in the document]
lemunozm: Ok, that makes sense. This is something new, right? I mean, previously, the check was in the trait, too.

pallets/order-book/src/lib.rs
```diff
@@ -941,40 +872,26 @@ pub mod pallet {
...
+			amount_out: T::Balance,
+			ratio: OrderRatio<T::Ratio>,
 		) -> Result<Self::OrderId, DispatchError> {
-			let min_fulfillment_amount = T::DecimalConverter::to_asset_balance(
-				T::MinFulfillmentAmountNative::get(),
-				currency_in,
-			)?;
+			// We only check if the trading pair exists not if the minimum amount is
+			// reached.
+			let _min_amount = TradingPair::<T>::get(&currency_in, &currency_out)?;
```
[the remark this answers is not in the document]
lemunozm: Regarding this, I think we also do not need to store `order.min_fulfillment_amount_out` in the order. `min_fulfilled_amount_out` could be computed as you say above directly in the fulfill method, WDYT?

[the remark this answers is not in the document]
mustermeiszer: Fine for me.

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: Yes. But we are having quite some troubles with that. 2 times investments got stuck. And now, updating an order below the minimum will also fail IIUC. So I would just remove it for now.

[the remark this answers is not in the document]
lemunozm: Done here: [7a6b430](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/7a6b43045ef4df4032f6364c3099c85fcc31abc6)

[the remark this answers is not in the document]
lemunozm: Last requirements here: [f209052](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/f2090528027de134d4291d2604374a3a1dd9226a)

lemunozm:
API renames from [this thread](https://kflabs.slack.com/archives/C050E3ZN2JU/p1704994019803859): [f151f9f](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/f151f9f204e60e3cae69e7bd7432dd925f9517db)

lemunozm:
There was a kind of a mess with the "asset" and "currency" names, using them indistinctly. I've pushed an opinionated commit ([9ea4401](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/9ea4401c88912e218e95175351a00f306c006002)) setting all names to their "currency" naming version. Let me know if you disagree with any of the changes

pallets/order-book/src/lib.rs
lemunozm: I fully agree with it 👍🏻

lemunozm:
I'm in a WIP state for this PR until I get refactorize the FI for market ratios

pallets/foreign-investments/src/entities.rs
```diff
@@ -52,16 +56,31 @@ pub struct InvestmentInfo<T: Config> {
...
+	/// Used to correlate the pool amount into foreign amount when the market
+	/// conversion is not known upfront.
+	pub invested_foreign_amount: T::Balance,
+
+	/// Total decrease pool amount pending to be swapped, denominated in foreign
+	/// currency.
+	///
+	/// Used to correlate the pool amount into foreign amount when the market
+	/// conversion is not known upfront.
+	pub decrease_pending_foreign_amount: T::Balance,
```
lemunozm, line 70: README FIRST

I've added these two new variables to be able to correlate the foreign and pool amounts. This is the key idea of this refactor, and we need them to know the other amount upfront without knowing the market rate.

Basically, the idea is: I invest `100` of foreign, and later it becomes into `200` of the pool. I save those `100` as `invest_foreign_amount` to later correlate with the real `invested_pool_amount` and be able to know how much pool amount I need to decrease if later the user calls `decrease(50)`. For this example, it will be: 
```rust
pool_investment_decrement =
= (foreign_amount / invested_foreign_amount) * invested_pool_amount =
= (50 / 100) * 200 = 100
```

So, no matter how the market ratio changes, if later the user calls `decrease(50)`, the amount of pool that is no longer invested is `100`.

pallets/foreign-investments/src/entities.rs
```diff
@@ -96,33 +106,47 @@ impl<T: Config> InvestmentInfo<T> {
...
-		)?;
+		let mut pool_investment_decrement = T::Balance::default();
+		if !foreign_investment_decrement.is_zero() {
+			let invested_pool_amount = T::Investment::investment(who, investment_id)?;
 
-		let pending_pool_amount_increment =
-			Swaps::<T>::pending_amount_for(who, investment_id, Action::Investment, pool_currency);
+			pool_investment_decrement = foreign_investment_decrement
+				.ensure_mul(invested_pool_amount)?
+				.ensure_div(self.invested_foreign_amount)
```
wischli, line 120: Dividing by 0 will panic. We need to ensure it's at least 1 via `self.invested_foreign_amount.max(1 in foreign denomination)`

pallets/foreign-investments/src/entities.rs
```diff
@@ -184,45 +216,54 @@ impl<T: Config> InvestmentInfo<T> {
...
-		)?;
+		let collected_pool_amount = collected.amount_payment;
 
-		let msg = ExecutedForeignCollect {
+		let previous_pool_amount_in_system =
+			T::Investment::investment(who, investment_id)?.ensure_add(collected_pool_amount)?;
+
+		let collected_foreign_amount = collected_pool_amount
+			.ensure_mul(self.invested_foreign_amount)?
+			.ensure_div(previous_pool_amount_in_system)?;
```
wischli, line 228: Again, let's ensure this is at least 1.

pallets/foreign-investments/src/impls.rs
```diff
@@ -132,14 +138,12 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
...
-				Action::Investment,
-				pool_currency_of::<T>(investment_id)?,
-			),
-		)?)
+		Ok(match ForeignInvestmentInfo::<T>::get(who, investment_id) {
+			Some(info) => info
+				.invested_foreign_amount
+				.ensure_add(info.pending_increase_swap(who, investment_id)?)?,
+			None => T::Balance::default(),
+		})
```
wischli, line 146: Nit: Could `.map(|info| ...).unwrap_or_default()`. Great that I didn't think of that when I proposed the change myself 😅

pallets/foreign-investments/src/lib.rs
```diff
@@ -192,20 +200,6 @@ pub mod pallet {
...
-		/// NOTE: Restricting to `IdentityCurrencyConversion` is solely a
-		/// short-term MVP solution. In the near future, this type must be
-		/// restricted to a more sophisticated trait which provides
-		/// unidirectional conversions based on an oracle, dynamic prices or at
-		/// least conversion ratios based on specific currency pairs.
-		type CurrencyConverter: IdentityCurrencyConversion<
-			Balance = Self::Balance,
-			Currency = Self::CurrencyId,
-			Error = DispatchError,
-		>;
```
wischli, line 207: Happy to see it gone 🙌

pallets/foreign-investments/src/swaps.rs
```diff
@@ -15,10 +15,10 @@ use crate::{
 /// `apply_swap()`
 #[derive(RuntimeDebugNoBound, PartialEq)]
 pub struct SwapStatus<T: Config> {
-	/// The amount (in) already swapped and available to use.
+	/// The incoming amount already swapped and available to use.
 	pub swapped: T::Balance,
 
-	/// The amount (in) pending to be swapped
+	/// The outcoming amount pending to be swapped
```
wischli, line 21: Nit
```suggestion
	/// The outgoing amount pending to be swapped
```

wischli (commented):
So far, I have only one potential CR for when we use `ensure_div` without ensuring the value is at least 1. Need to still go through the OrderBook UTs which I will do during integration test compile times.

Really awesome work!

pallets/foreign-investments/src/entities.rs
lemunozm: `ensure_div` already takes care of not panicking when it's 0

wischli: Ah nice, should have checked that first 🫠

lemunozm: No worries! Well done highlighting it

pallets/foreign-investments/src/impls.rs
lemunozm: Good point. I tried it, but the inner closure of `map` returns a result, so I could not later use `unwrap_or_default()`, and everything seemed ugly.

pallets/foreign-investments/src/swaps.rs
lemunozm: Thanks for correcting my English 🙌🏻

pallets/foreign-investments/src/entities.rs
```diff
@@ -96,33 +106,47 @@ impl<T: Config> InvestmentInfo<T> {
...
-			foreign_amount,
-		)?;
+		let mut pool_investment_decrement = T::Balance::default();
+		if !foreign_investment_decrement.is_zero() {
+			let invested_pool_amount = T::Investment::investment(who, investment_id)?;
 
-		let pending_pool_amount_increment =
-			Swaps::<T>::pending_amount_for(who, investment_id, Action::Investment, pool_currency);
+			pool_investment_decrement = foreign_investment_decrement
+				.ensure_mul(invested_pool_amount)?
```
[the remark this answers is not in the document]
lemunozm: Yes, good point.

When I commented on the idea, I wrote it as `(100A / 50A) * 2B`. So, first, I got the proportion, and later, I multiplied by it. Nevertheless, multiplying first gives us a bigger numerator for the division, reducing the possibility of decimal issues.

libs/types/src/investments.rs
```diff
@@ -208,9 +171,20 @@ impl<Balance: Clone + Copy + EnsureAdd + EnsureSub + Ord + Debug, Currency: Clon
...
 
+/// A representation of a currency swap in process.
+#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+pub struct SwapState<Balance, Currency> {
+	/// Swap not yet processed with the pending outcomming amount
+	pub remaining: Swap<Balance, Currency>,
+	/// Amount of incoming currency already swapped
+	pub swapped_in: Balance,
+	/// Amount of incoming currency already swapped denominated in outcomming
+	/// currency
```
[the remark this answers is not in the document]
lemunozm: I did it on purpose, maybe there is a better way to explain it. 

What I want to say is that the variable actually represents the incoming currency amount, but it's "transformed" to be read as outgoing. But it should represent the incoming one 😅

pallets/foreign-investments/src/tests.rs
```diff
@@ -783,20 +804,85 @@ mod investment {
...
+				);
+				Ok(())
+			});
+
+			assert_ok!(ForeignInvestment::increase_foreign_investment(
+				&USER,
+				INVESTMENT_ID,
+				AMOUNT / 2,
+				FOREIGN_CURR
+			));
```
lemunozm, line 870: There were a bug (fixed in [5c03bd2](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/5c03bd217d787dbe3ffb22aded66c43992c96cfc with a solution that I do not like at all).

Basically, there is a rare case where increasing a foreign investment can lead, ironically, into a `Decrease` message. It happens when previously a decrease swap was partially fulfilled, and the increase message cancel the pending part of the previous decrease swap. In that case, a message should be sent to solidity. A more visual example:

- Increase 100 and fully swapped.
- Decrease 100.
- Partial swap with 75 (25 pending). So, we have not sent a `Decrease` msg yet.
- Increase again with 25 (or more), canceling the previous pending swap, and sending here a `Decrease` msg.

I suppose this kind of test does not exist previously in IT, right? @wischli 

The current solution is really difficult to read 😞 . I'll think about how I can clean `pre_increase_swap` and `pre_decrease_swap` because both are getting very complex and require a lot of thinking to know what's really happening.

wischli: IMO, the `ExecutedDecrease` needs to be dispatched in this case. Otherwise, the decreased 100 are never signalled back to Solidity which relies on the state of Centrifuge Chain. This aligns with the V1 state diagram in which the transition was `ActiveSwapIntoForeignAndSwapIntoForeignDone --increase_investment--> ActiveSwapIntoPoolAndSwapIntoForeignDone --dispatch_ExecutedDecrease--> ActiveSwapIntoPool`.

pallets/foreign-investments/src/tests.rs
```diff
@@ -827,6 +827,69 @@ mod investment {
...
+
+			util::fulfill_last_swap(Action::Investment, foreign_to_pool(3 * AMOUNT / 4));
+
+			MockDecreaseInvestHook::mock_notify_status_change(|_, msg| {
+				assert_eq!(
+					msg,
+					ExecutedForeignDecreaseInvest {
+						amount_decreased: AMOUNT / 2,
+						foreign_currency: FOREIGN_CURR,
+						amount_remaining: AMOUNT / 2,
```
wischli, line 859: I don't understand why `amount_decreased` is `AMOUNT / 2`.  It should be `AMOUNT`, shouldn't it?

pallets/foreign-investments/src/tests.rs
```diff
@@ -787,7 +787,7 @@ mod investment {
 				Some(InvestmentInfo {
 					base: BaseInfo::new(FOREIGN_CURR).unwrap(),
 					invested_foreign_amount: AMOUNT / 4,
-					decrease_pending_foreign_amount: 3 * AMOUNT / 4,
+					decrease_pending_foreign_amount: AMOUNT / 2,
```
wischli: Great catch! I need to go over all FI unit tests once more very carefully and check all amounts.

pallets/foreign-investments/src/tests.rs
wischli: Unfortunately, this test does not exist in integration tests. As part of #1700, I removed a few tests which were handling V1 edge cases that are not edge cases anymore. I don't think the above one was included. Do you think, I should add more integration tests? There is much more boilerplate than with FI unit tests.

lemunozm: No worries! I have already handled the UTs.

lemunozm: > IMO, the ExecutedDecrease needs to be dispatched in this case with foreign amount 100. Otherwise, the decreased 100 are never signalled back to Solidity which relies on the state of Centrifuge Chain

Following the above example, the message will be signaled back to Solidity once it is increased by at least `25`. In that case, for `25`, solidity will receive a message with `75` amount decreased. But, if we wait until getting `100`, that could never happen.

pallets/foreign-investments/src/tests.rs
lemunozm: Too many amounts... I think the madness starts doing this kind of thing 😆

pallets/foreign-investments/src/tests.rs
lemunozm: This is very tricky and always makes me rethink it:

Supposing `AMOUNT` to 100:

| Action       | Pending to send | Pending to swap | Invested |
|--------      |-----------------|-----------------|----------|
| increase 100 | 0               | 100   (->)      | 0        |
| swap 100     | 0               | 0               | 100      |
| decrease 100 | 0               | 100   (<-)      | 0        |
| swap 100     | 75              | 25    (<-)      | 0        |
| increase 50  | 0               | 25    (->)      | 25       |

Basically, when increasing `50`, 
- The pending decrease swap of `25`, is cancelled and invested.
- `25` from "pending to be send" are no longer sent, and becomes pending to be swap.
- `50` from "pending to be send" are sent, because there is no more pending amount to wait.

lemunozm:
I made a couple of modifications to the internal FI structure:
- I've added a type `Correlation` to keep track of both symmetrical that are being correlated, forcing that they always are "correlated".
- I've split `post_increase/decrease_swap()` methods into two variants, one when the swap is actually a swap and another when the swap is actually a cancelation of another swap. Before that, there were a lot of concept mixed and with this explicit separation, I think the code is more easy to follow.

pallets/foreign-investments/src/tests.rs
lemunozm: The simpler way of thinking about this: 

If you increase `AMOUNT`, then decrease `AMOUNT`, and again increase in `AMOUNT / 2`, then, the resulting amount in the system is: `AMOUNT - AMOUNT + AMOUNT / 2 = AMOUNT / 2`

wischli: > If you increase `AMOUNT`, then decrease `AMOUNT`, and again increase in `AMOUNT / 2`, then, the resulting amount in the system is: `AMOUNT - AMOUNT + AMOUNT / 2 = AMOUNT / 2`

I agree with the remaining amount but not the decreased amount. It should not be `AMOUNT / 2` but `AMOUNT` . Otherwise, the message says: "You have decreased your investment by 50 and 50 remaining investment" which is wrong. It should be 100 decreased and 50 remaining.

pallets/foreign-investments/src/entities.rs
```diff
@@ -52,39 +56,72 @@ pub struct InvestmentInfo<T: Config> {
...
-			currency_out: self.base.foreign_currency,
-			amount_in: pool_amount,
-		})
+		let mut send_msg = false;
+		if foreign_amount >= decreasing_foreign_amount {
+			let swap_foreign_amount = foreign_amount.ensure_sub(decreasing_foreign_amount)?;
+
+			self.decrease_swapped_foreign_amount = self
+				.decrease_swapped_foreign_amount
+				.saturating_sub(swap_foreign_amount);
```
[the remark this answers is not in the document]
wischli: I think this is outdated now.

pallets/foreign-investments/src/tests.rs
lemunozm: Ouch! I read you wrongly 😓. Then, yes, I think you're right

pallets/foreign-investments/src/entities.rs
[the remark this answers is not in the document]
lemunozm: Yeah, that version you were reviewing, I think, was not ok.

Sorry for the amount of internal refactors. I know it's not easy for reviews 🙏🏻. Very grateful for your deep analysis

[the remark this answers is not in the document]
wischli: I already had a pending comment on that line but wanted to be sure first before causing turmoil 😅 I still believe it is incorrect.

pallets/foreign-investments/src/tests.rs
lemunozm: Solved

pallets/order-book/src/lib.rs
```diff
@@ -525,464 +484,346 @@ pub mod pallet {
...
+				Precision::Exact,
+			)?;
+
+			if T::Currency::balance(order.currency_out, &order.placing_account) < amount_out {
+				Err(DispatchError::Token(TokenError::FundsUnavailable))?
+			}
+
+			if T::Currency::balance(order.currency_in, &fulfilling_account) < amount_in {
+				Err(DispatchError::Token(TokenError::FundsUnavailable))?
+			}
```
mustermeiszer, line 688: Why do we need that again?

pallets/order-book/src/lib.rs
```diff
@@ -239,51 +241,55 @@ pub mod pallet {
...
 		_,
 		Twox64Concat,
-		T::AssetCurrencyId,
+		T::CurrencyId,
 		Twox64Concat,
-		T::AssetCurrencyId,
+		T::CurrencyId,
 		BoundedVec<T::OrderIdNonce, T::OrderPairVecSize>,
 		ValueQuery,
 	>;
```
mustermeiszer, line 257: Do we still need that?

mustermeiszer: Can't we make it an `NMap`? Would make it a lot easier and we do not have the size restriction anymore.

mustermeiszer (commented):
Review Part 1: Order book

pallets/order-book/src/lib.rs
lemunozm: > Do we still need that?

TBH, I do not know. Why did we need it before? I understand that for Apps, it's an easy way to look for existing orders to fill. But not sure if it used now.

> Can't we make it an NMap? 

Makes more sense to me. Also, I would add the ratio to it: `MMap<(CurrencyID, CurrencyID, OrderRatio), OrderId>` i.e. to look for orders that use market ratios.

But by now, if not used by Apps (@onnovisser), I would just remove this, and I'll add it again once we know exactly how we want to index it, WDYT?

pallets/foreign-investments/src/entities.rs
```diff
@@ -19,71 +23,128 @@ use crate::{
...
+	/// Transform any foreign amount into a pool amount
+	pub fn foreign_to_pool(&self, foreign_amount: T::Balance) -> Result<T::Balance, DispatchError> {
+		if foreign_amount.is_zero() {
+			return Ok(T::Balance::default());
+		}
+
+		foreign_amount
+			.ensure_mul(self.pool_amount)?
+			.ensure_div(self.foreign_amount)
+			.map_err(|_| Error::<T>::TooMuchDecrease.into())
```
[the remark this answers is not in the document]
lemunozm: [8f2ba86](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/8f2ba8610f9ce65c96dbf5262351d7960b37ab15)

pallets/foreign-investments/src/swaps.rs
```diff
@@ -152,13 +129,13 @@ impl<T: Config> Swaps<T> {
 					who.clone(),
 					new_swap.currency_in,
 					new_swap.currency_out,
-					new_swap.amount_in,
-					T::BalanceRatio::one(),
+					new_swap.amount_out,
+					OrderRatio::Market,
```
mustermeiszer, line 133: **CLARIFICATION**

Isn't it problematic that the `SwapOf<T>` has a directionality when reaching this method - `currency_out` &rarr; `currency_in`, but the returned `SwapStatus` might have a changed directionality. 

So depending on what happened internally here, the `SwapStatus` fields `swapped` and `pending` refer to different currencies...

mustermeiszer: From the point of view of the caller, that is "updating" the swap here:

`swapped` and `pending` refer to 
* if no order &rarr; `currency_in`
* if order same direction &rarr; `currency_in`
* if order reverse direction
   * if `reverse_amount_in` greater &rarr; `currency_out`
   * if `reverse_amount_in` equal &rarr; `currency_in`
   * if `reverse_amount_in` smaller &rarr; `currency_in`

mustermeiszer: @lemunozm assume the directionality has changed. Wouldn't [this line (same as code below)](https://github.com/centrifuge/centrifuge-chain/blob/12d66f6e1a2ec557ff943af68ded7b3d020746ae/pallets/foreign-investments/src/impls.rs#L41)

```rust
let swapped_foreign_amount = foreign_amount.ensure_sub(status.pending)?;
```

be problematic?

pallets/foreign-investments/src/swaps.rs
```diff
@@ -183,26 +155,25 @@ impl<T: Config> Swaps<T> {
...
-								who.clone(),
 								swap_id,
 								amount_to_swap,
-								T::BalanceRatio::one(),
+								OrderRatio::Market,
 							)?;
 
 							Ok(SwapStatus {
-								swapped: new_swap.amount_in,
+								swapped: new_swap_amount_in,
```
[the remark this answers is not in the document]
wischli: No, because `amount_to_swap` is the pending swap amount but the `status.swapped` returned by `apply_over_swap` refers to the amount that can be regarded as fulfilled swap because of resolving opposite swaps.

**Example**
Assume
* inverse swap is from pool to foreign with `amount_out = 100P`
* new swap is from foreign to pool with `amount_in = 30P`

Then, the decreasing swap from foreign to pool can be regarded as fulfilled such that the `status.swapped = 30P` and `amount_to_swap = 70P`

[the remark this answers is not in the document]
mustermeiszer: I was referring to the line below, that was not changed. It is `zero` in this case, as I wrote `pending: amount_to_swap`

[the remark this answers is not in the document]
lemunozm: No, because the `pending` here is seen from the perspective of whoever makes the swap. So, if you apply a swap, in this case, you do not have to wait for more "swapping" because your swap has been completely swaped. Independently, it could be any other inverse swap pending, but that doesn't concern you at this moment. You've got fully swapped your swap.

pallets/foreign-investments/src/swaps.rs
lemunozm: > - if reverse_amount_in greater → currency_out

If I'm not wrong, this is `currnecy_in`, so all cases return `currency_in` as `swapped` and `currency_out` as `pending`.

So, in the referenced line, `status.pending` from an increase swap is `foreign_currency`, and it's subtracted with `foreing_amount`, so both currencies match

pallets/foreign-investments/src/entities.rs
```diff
@@ -98,131 +158,221 @@ impl<T: Config> InvestmentInfo<T> {
...
-					.map_err(|_| Error::<T>::TooMuchDecrease)?,
-			)?;
-		}
-
 		Ok(Swap {
-			currency_in: self.base.foreign_currency,
+			currency_in: self.foreign_currency,
 			currency_out: pool_currency,
-			amount_in: foreign_amount,
+			amount_out: increasing_pool_amount.ensure_add(pool_investment_decrement)?,
```
mustermeiszer, line 189: I am looking at this code now for 4 hours. I thought multiple times I found a bug, but I always recovered later down the road.

I am now pretty certain I found a bug, as the `T::update_investment()` logic works different than the code expects here. It is updating the investment to the given value, so it expects an absolute value not a delta. So the bug is ONLY visible in integration tests.

### Example 1
1. `Increase(100 ethUSDC)`
1.1. `Swap(0 USDC, 100 ethUSDC)`
2. `FillOrder(70 USDC)`
2.1. `Swap(70 USDC, 30 ethUSDC)`
2.2. `Investment(70 USDC)`
3. `Decrease(20 ethUSDC)` 

#### `pre_decrease_swap(..)` 
* `foreign_amount` &rarr; `20`
* `increasing_foreign_amount  = self.pending_increase_swap(..)` &rarr;  `30`
* `foreign_investment_decrement = sat(foreign_amount - increasing_foreign_amount)` &rarr;  `sat(20-30) = 0`
* `self.decrease_investment(foreign_investment_decrement)` &rarr;  `Investment()` ❌  &rarr;  should be `Investment(70)`
* `increasing_pool_amount = min(foreign_amount, increasing_foreign_amount)` &rarr; `min(20, 30) = 20`
* `amount_out: increasing_pool_amount + foreign_investment_decrement` &rarr; `20 + 0 = 20` 
   
**Returned `SwapOf`**
* `amount_out: 20`  ✅   &rarr; swap will be updated as inverse with `Ordering::Smaller` &rarr;  `amount_out = 10`
* `currency_in: ethUSDC` ✅  
* `currency_out: USDC` ✅  
      
### Example 2 
1. `Increase(100 ethUSDC)`
1.1. `Swap(0 USDC, 100 ethUSDC)`
2. `FillOrder(70 USDC)`
2.1. `Swap(70 USDC, 30 ethUSDC)`
2.2. `Investment(70 USDC)`
3. `Decrease(100 ethUSDC)` 

#### `pre_decrease_swap(..)` 
* `foreign_amount` &rarr; `100`
* `increasing_foreign_amount  = self.pending_increase_swap(..)` &rarr;  `30`
* `foreign_investment_decrement = sat(foreign_amount - increasing_foreign_amount)` &rarr;  `sat(100-30) = 70`
* `self.decrease_investment(foreign_investment_decrement)` &rarr;  `Investment(70)` ❌  &rarr;  should be `Investment(0)` 
* `increasing_pool_amount = min(foreign_amount, increasing_foreign_amount)` &rarr; `min(100, 30) = 30`
* `amount_out: increasing_pool_amount + foreign_investment_decrement` &rarr; `30 + 70 = 100` 
   
**Returned `SwapOf`**
* `amount_out: 100` ✅  &rarr; swap will be updated as inverse with `Ordering::Greater` &rarr; `amount_out = 70`
* `currency_in: ethUSDC` ✅ 
* `currency_out: USDC` ✅ 

### Example 3
1. `Increase(100 ethUSDC)`
1.1. `Swap(0 USDC, 100 ethUSDC)`
2. `FillOrder(70 USDC)`
2.1. `Swap(70 USDC, 30 ethUSDC)`
2.2. `Investment(70 USDC)`
3. `Decrease(40 ethUSDC)` 

#### `pre_decrease_swap(..)` 
* `foreign_amount` &rarr; `40`
* `increasing_foreign_amount  = self.pending_increase_swap(..)` &rarr;  `30`
* `foreign_investment_decrement = sat(foreign_amount - increasing_foreign_amount)` &rarr;  `sat(40-30) = 10`
* `self.decrease_investment(foreign_investment_decrement)` &rarr;  `Investment(10)` ❌  &rarr;  should be `Investment(20)`
* `increasing_pool_amount = min(foreign_amount, increasing_foreign_amount)` &rarr; `min(40, 30) = 30`
* `amount_out: increasing_pool_amount + foreign_investment_decrement` &rarr; `30 + 10= 50` 
   
**Returned `SwapOf`**
* `amount_out: 50` ✅  &rarr; swap will be updated as inverse with `Ordering::Greater` &rarr; `amount_out = 20`
* `currency_in: ethUSDC` ✅ 
* `currency_out: USDC` ✅

pallets/foreign-investments/src/swaps.rs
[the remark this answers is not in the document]
mustermeiszer: I see. Thanks!

pallets/foreign-investments/src/swaps.rs
mustermeiszer: Yeah, that resulted in my misunderstanding of the return value. It is always the same directionality then! Thanks for claryfiuyng.

pallets/order-book/src/lib.rs
```diff
@@ -525,464 +484,346 @@ pub mod pallet {
...
-		) -> DispatchResult {
+		#[pallet::weight(T::Weights::set_market_feeder())]
+		pub fn set_market_feeder(origin: OriginFor<T>, feeder_id: T::FeederId) -> DispatchResult {
 			T::AdminOrigin::ensure_origin(origin)?;
 
-			// try_mutate is a pain with the direct error return. But we do want to only
-			// update if the pair exists.
-			let _old_min_order = TradingPair::<T>::get(&asset_in, &asset_out)?;
-			TradingPair::<T>::insert(&asset_in, &asset_out, min_order);
+			MarketFeederId::<T>::put(feeder_id);
```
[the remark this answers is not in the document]
mustermeiszer: Events are a means of saying that the state changed, so yes indexers will most likely only see that change if there was an evenrt.

pallets/order-book/src/lib.rs
mustermeiszer: If we assume we will only have a reasonable amount of orders in paralel that will be fine. The UI will most likely need to filter based on currency, but if they can just fetch all orders and sort them accordingly that should be fine I ugess.

pallets/foreign-investments/src/entities.rs
mustermeiszer: Ah fuck! Haha. Yeah, found that line! That its fine.

mustermeiszer: And nevermind regarding the complexity.

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Ok, because the indexer can not also index "calls" done, right? Because in that case, just seeing that the call has been performed, they now the state has changed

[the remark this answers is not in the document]
lemunozm: Done anyways here: [4a0f52a](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/4a0f52ab67e899da74e4b504a73020f63616c255)

pallets/order-book/src/lib.rs
onnovisser: the queries that the UI uses are `orderBook.orders` and `orderBook.tradingPair`. Because we also want to filter by outCurrency and can't otherwise do that, we just get all orders and filter them in the UI, and assume they will be a reasonable amount like Frederik said.

lemunozm: Ok, so I would remove this by now

lemunozm: In case of aggrement, here: [e579147](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/e57914785cafc142341c82d160d05e3292bb24ff), If not I revert

wischli (approved):
LGTM! Thanks for all the sweat and effort you have put into this!

mustermeiszer (approved):
That is a huge one!! Thanks to both @lemunozm and @wischli for the effort and brain juice that went into this here!

PR #1692, opened 2024-01-16
PoC: FI: Without internal states

lemunozm:
# Description

Proof of concept for a FI implementation without states. Corresponds to [Approach 2](https://kflabs.slack.com/archives/C04GJQAM9P0/p1705769643203879)


pallets/foreign-investments/src/lib.rs
```diff
@@ -437,20 +355,320 @@ pub mod pallet {
...
+				who,
+				investment_id,
+				T::Investment::redemption(who, investment_id)?.ensure_sub(amount)?,
+			)?;
+
+			// For William: Why I should return here the amounts? Should not be done after
+			// collecting the redemption?
+
+			Ok(todo!())
+		}
```
lemunozm, line 549: Wondering about this

pallets/foreign-investments/src/lib.rs
```diff
@@ -437,20 +355,320 @@ pub mod pallet {
...
+			who: &T::AccountId,
+			investment_id: T::InvestmentId,
+			_foreign_payout_currency: T::CurrencyId,
+			_pool_currency: T::CurrencyId,
+		) -> DispatchResult {
+			T::Investment::collect_redemption(who.clone(), investment_id)
+
+			// TODO: initialize a swap to transform the pool redemption into
+			// foreign redemption amount
+		}
```
lemunozm, line 569: in relation to this

lemunozm (commented):
It compiles with `cargo check -p foreign-investment`. And I've probably needs some extra checks in some places. But I think the idea is more or less shown.

For each `(account, investment_id)` pair we only handle one swap, and based on that swap and the investment state we determine how to proceed in each function.

pallets/foreign-investments/src/lib.rs
lemunozm: Gotcha, we just return the trache tokens that becomes "free" again after decreasing the redemption

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +302,473 @@ pub mod pallet {
...
+		/// - If there was a swap in the same direction, it increments it.
+		/// - If there was a swap in the opposite direction:
+		///   - If the amount is smaller, it decrements it.
+		///   - If the amount is the same, it removes the inverse swap.
+		///   - If the amount is greater, it removes the inverse swap and create
+		///     another with the excess
+		///
+		/// The returned status contains the swapped amount after this call and
+		/// the pending amount to be swapped.
+		fn apply_swap(
```
lemunozm, line 364: This is the key to the refactoring. All swapping processes happen here, independently of the direction or if it's an investment or redemption. Basically, this method extracts the swapping logic from `InvestState` and `RedeemState`. Later, the increase/decrease method uses `apply_swap()` to perform their needs.

It only uses the `TokenSwaps` trait, so mocking this effect should be enough to test all possible swap combinations isolated.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +302,473 @@ pub mod pallet {
...
+					currency_out: foreign_currency,
+					amount_in: pool_amount,
+				},
+			)?;
+
+			T::Investment::update_investment(
+				who,
+				investment_id,
+				T::Investment::investment(who, investment_id)?.ensure_add(status.swapped)?,
+			)
```
lemunozm, line 503: i.e. in `increase_foreign_investment` we call as follows ^^

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +302,473 @@ pub mod pallet {
...
+					currency_out: pool_currency,
+					amount_in: foreign_amount,
+				},
+			)?;
+
+			T::Investment::update_investment(
+				who,
+				investment_id,
+				T::Investment::investment(who, investment_id)?.ensure_sub(status.pending)?,
+			)?;
```
lemunozm, line 528: And in `decrease_foreign_investment` we call if with the inverse currencies ^^

lemunozm (commented):
An small tour over this refactor:

pallets/foreign-investments/src/lib.rs
lemunozm: Ideally, after https://github.com/centrifuge/centrifuge-chain/pull/1674, we only need to touch this method to update the FI to support new the order book implementation with `amount_out` instead of `amount_in`

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +302,473 @@ pub mod pallet {
...
+			pool_currency: T::CurrencyId,
+		) -> DispatchResult {
+			ForeignInvestmentInfo::<T>::insert(
+				who,
+				investment_id,
+				Info {
+					foreign_currency,
+					pool_currency,
+				},
+			);
```
mustermeiszer, line 480: There can be multiple `foreign_currency` variants comming in for one `AccountId-InvestmentId` combination. This would override the existing information. 

* We can rather use an `NMap` of `(AccountId, InvestmentId, CurrencyId)` - Still store `Info` there thouhg
* Error out if the existing storage `foreign_currency` does not match

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +302,473 @@ pub mod pallet {
...
+						T::TokenSwaps::update_order(
+							who.clone(),
+							swap_id,
+							swap.amount_in.ensure_add(new_swap.amount_in)?,
+							T::BalanceRatio::one(),
+						)?;
+
+						Ok(SwapStatus {
+							swapped: T::Balance::zero(),
+							pending: new_swap.amount_in,
```
mustermeiszer, line 400: Shouldn't that be `swap.amount_in.ensure_add(new_swap.amount_in)?`

mustermeiszer (commented):
IMO a lot of the complexity here could further be removed if we just removed the automatic handling of reverse swap fulfillments and move that to the order-book. 

We wanted to have matching orders there anyways. I.e. 
* create runtime api that generates a matching vectors
* submit that matching vector to the chain and match orders


Furthermore, this handling on FI assumes a 1:1 conversion in both directions which is most likely always the case, but a non-noticed break if we change that.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +302,473 @@ pub mod pallet {
...
+								})
+							}
+							Ordering::Equal => {
+								T::TokenSwaps::cancel_order(swap_id.clone())?;
+								Self::unregister_swap(&who, investment_id, kind, swap_id);
+
+								Ok(SwapStatus {
+									swapped: new_swap.amount_in,
+									pending: T::Balance::zero(),
+								})
```
mustermeiszer, line 427: Aren't we forgetting to notify the other part of the swap about the update here? the new submission gets a status, but the first submission never gets notified as the order is just canceled.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +302,473 @@ pub mod pallet {
...
+					currency_in: pool_currency,
+					currency_out: foreign_currency,
+					amount_in: pool_amount,
+				},
+			)?;
+
+			T::Investment::update_investment(
+				who,
+				investment_id,
+				T::Investment::investment(who, investment_id)?.ensure_add(status.swapped)?,
```
mustermeiszer, line 502: no change of order will fail on the investments side...

mustermeiszer: At least in the extrinsics. not sure about the trait

pallets/foreign-investments/src/lib.rs
lemunozm: Make sense, good point 👍🏻. Right now, are we using different foreign currencies, or is it something expected in the future?

pallets/foreign-investments/src/lib.rs
lemunozm: Thanks for the reminder!

pallets/foreign-investments/src/lib.rs
lemunozm: That's a really good point 👏🏻 . I think you're right. I would need to think about a clean solution for it. It seems not obvious.

Definitely, moving this to `order-book` will be the simplest solution. `order-book` will call the corresponding callbacks

pallets/foreign-investments/src/lib.rs
lemunozm: I was interpreting `pending` as the pending amount from the given swap (the one given to apply_swap()`). Not the total `pending` amount. 

Good point, by the way!

mustermeiszer:
> Can fulfilling an order be seen as placing an order with inverse currencies? If so, do we really need a fulfill_order() extrinsic?

It can, BUT we can not iterate over the order storage on-chain. So fulfilling an order would need to calls `place_order()` with the inverse, `match_order(s)()` (name is TBD) to actually let the chain know which orders to match. `fulfill_order` is in this case just a direct approach of the two above. 

I personally do not care too much about the design, but keep we should keep in mind that the UI needs to work with that. So we definitely need to have a simple matching api. 

> The currency apply_swap() logic only swaps for the same account. Will this new feature swap against any order from any account?

Yes, any account. Again, on-chain that is hard. My idea was that `FI` just places a swap - the order-book would increase, decrease accordingly if it already exists and the same `Price` is used. We then have an off-chain worker/boot/tasks(not yet available) that actually calls the runtime-api and submits the matching. The matching will trigger the notification for `FI` to further do what it wants to do.

> How should we act in the case where there is a placed order from A to B with a custom ratio of 0.5, but I'm trying to create an order with a market ratio from B to A of 0.9? Which ratio should we use to solve this swap?

As said above, I would then create two orders with different ids. `FI` currently does not care as it will use `Market`, but if we can already consider that, great.

pallets/foreign-investments/src/lib.rs
lemunozm: @mustermeiszer is this a new requirement or an expected requirement? From the current FI implementation I do not see support for different foreign currencies. Every state is defined by `(account, investment_id)`

lemunozm: If this is something new or expected but not used right now I would leave it for future improvement

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +321,563 @@ pub mod pallet {
...
+		) -> bool {
+			if T::Investment::accepted_payout_currency(investment_id, currency) {
+				true
+			} else {
+				T::PoolInspect::currency_for(investment_id.of_pool())
+					.map(|pool_currency| T::TokenSwaps::valid_pair(currency, pool_currency))
+					.unwrap_or(false)
+			}
+		}
+	}
```
lemunozm, line 790: This implementation is the same as before. Wondering why it could return `true`, even if `Investment::accepted_payout_currency()` return `false`. Why the `else` branch?

cc @wischli

pallets/foreign-investments/src/lib.rs
mustermeiszer: The problem here is that you are overriding the storage. IIRC `FI` does always check, that the currencies are matching right now, which is not done here.

pallets/foreign-investments/src/lib.rs
mustermeiszer: Else is bascially going to swap into `pool_currency`, while `T::Investment::accepted_payout_currency` does just check the if `currency` is equal `pool_currency`

pallets/foreign-investments/src/lib.rs
lemunozm: Ok, this is fixed now to emulate the FI behavior

pallets/foreign-investments/src/lib.rs
lemunozm: So, if the currency is not accepted by `T::Investment::accepted_payout_currency`, then it could be accepted for FI anyway by the else branch.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +380,565 @@ pub mod pallet {
...
+
+		fn apply_swap_with_notifications(
+			who: &T::AccountId,
+			investment_id: T::InvestmentId,
+			new_swap: SwapOf<T>,
+		) -> DispatchResult {
+			let status = Self::apply_swap(who, investment_id, new_swap.clone())?;
+
+			if !status.swapped.is_zero() {
+				Self::notify_swap_done(
```
wischli, line 559: Placeholder: TODO(william) Investigate why we notify twice for many occasions

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +380,565 @@ pub mod pallet {
...
+			let (who, investment_id) =
+				SwapidToForeignId::<T>::get(swap_id).ok_or(Error::<T>::SwapOrderNotFound)?;
+
+			let remaining_swap_amount_in = match T::TokenSwaps::get_order_details(swap_id) {
+				Some(swap) => swap.amount_in,
+				None => {
+					Pallet::<T>::unregister_swap(&who, investment_id, swap_id);
+					T::Balance::zero()
+				}
+			};
```
wischli, line 873: IMO, we should add this information to the `Status` such that we reduce the DB reads by one.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +380,565 @@ pub mod pallet {
...
+		) -> DispatchResult {
+			let info = ForeignInvestmentInfo::<T>::mutate(&who, investment_id, |maybe_info| {
+				let info = maybe_info.as_mut().ok_or(Error::<T>::InfoNotFound)?;
+				info.base.collected_amount.increase(&collected)?;
+				Ok::<_, DispatchError>(info.clone())
+			})?;
+
+			if info.remaining_foreign_amount()?.is_zero() {
+				ForeignInvestmentInfo::<T>::remove(&who, investment_id);
+			}
```
wischli, line 903: See my above remark that `remaining_pool_amount` should be used instead. Moreover, this check might be redundant because we increase `collected_amount` by `collected` such that the remaining amount can only be zero if it was zero and incremented by zero (0 + 0).

wischli (commented):
Looks super clean and couldn't spot major flaws in the first round! Great job 👏 🎉 

Need to review the usage of `fn apply_swap_with_notifications` and `fn notify_swap_done` in the second round.

Some remarks:

pallets/foreign-investments/src/lib.rs
```diff
@@ -43,43 +43,131 @@
...
+		Ok(Self {
+			base: BaseInfo::new(investment_id, foreign_currency)?,
+			total_foreign_amount: T::Balance::default(),
+		})
+	}
+
+	fn remaining_foreign_amount(&self) -> Result<T::Balance, ArithmeticError> {
+		self.total_foreign_amount
+			.ensure_sub(self.base.collected_amount.amount_collected)
+	}
```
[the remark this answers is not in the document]
lemunozm: > For investments, the amount_collected is the amount of collected tranche tokens. What we need here is amount_payment converted into foreign, because amount_payment is the amount of pool currency.

You're right! I was almost sure I had some error with those quantities. Thanks for your eyes!

> Thus, for any (foreign, pool) pair which does not swap 1:1, we can't determine the foreign amount from the pool amount upfront synchronously and deterministically 

Uff... you're right, and I do not like it this issue.

Just a thought. Why do not specify the `amount_remaining` as `pool_currency` instead? I mean, it made sense to specify it in foreign currencies before, but now, with market prices, we should specify what makes sense for the chain and is more "immutable," in this case, pool currencies.

pallets/foreign-investments/src/lib.rs
```diff
@@ -227,7 +292,7 @@ pub mod pallet {
 		/// restricted to a more sophisticated trait which provides
 		/// unidirectional conversions based on an oracle, dynamic prices or at
 		/// least conversion ratios based on specific currency pairs.
-		type CurrencyConverter: cfg_traits::IdentityCurrencyConversion<
+		type CurrencyConverter: IdentityCurrencyConversion<
```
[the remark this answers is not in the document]
lemunozm: Yes, this PR uses upfront known 1:1 swaps. 

My idea on mind is to have `TokensSwaps::convert()` or similar that handles the swaps using the correct ratios from the market.

pallets/foreign-investments/src/lib.rs
```diff
@@ -242,168 +307,61 @@ pub mod pallet {
...
-	pub type RedemptionPayoutCurrency<T: Config> = StorageDoubleMap<
-		_,
-		Blake2_128Concat,
-		T::AccountId,
-		Blake2_128Concat,
-		T::InvestmentId,
-		T::CurrencyId,
-		ResultQuery<Error<T>::RedemptionPayoutCurrencyNotFound>,
-	>;
+	pub(super) type SwapidToForeignId<T: Config> =
```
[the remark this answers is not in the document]
lemunozm: Thanks!

pallets/foreign-investments/src/lib.rs
```diff
@@ -242,168 +307,61 @@ pub mod pallet {
...
-	/// Maps an investor and their `InvestmentId` to the corresponding
-	/// `InvestState`.
+	/// Contains the information about the foreign investment process
 	///
-	/// NOTE: The lifetime of this storage starts with initializing a currency
-	/// swap into the required pool currency and ends upon fully processing the
-	/// investment after the potential swap. In case a swap is not required, the
-	/// investment starts with `InvestState::InvestmentOngoing`.
+	/// NOTE: The storage is killed once the investment is collected or
+	/// redemption process is collected and fully swapped
```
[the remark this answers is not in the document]
lemunozm: Thanks, actually, I need to check the lifetime of this storage

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +380,565 @@ pub mod pallet {
...
-		/// having increased their investment as this would store the payment
-		/// currency.
-		InvestmentPaymentCurrencyNotFound,
-		/// Failed to retrieve the foreign payout currency for a collected
-		/// redemption.
-		///
-		/// NOTE: This error can only occur, if a user tries to collect before
-		/// having increased their redemption as this would store the payout
-		/// currency.
-		RedemptionPayoutCurrencyNotFound,
```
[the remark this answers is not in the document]
lemunozm: Yes, exactly. Do you agree with that, or do you think we should be more verbose here?

The error is now handled under `info.ensure_same_foreign(foreign_currencies)` calls.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +380,565 @@ pub mod pallet {
...
+							swapped: T::Balance::zero(),
+							pending: amount_to_swap,
+							swapped_inverse: T::Balance::zero(),
+							pending_inverse: T::Balance::zero(),
+						})
+					} else {
+						let inverse_swap = swap;
+						let new_swap_amount_out = Self::amount_used_to_swap(&new_swap)?;
+
+						match inverse_swap.amount_in.cmp(&new_swap_amount_out) {
```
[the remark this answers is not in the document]
lemunozm: Thanks! Good eyes!

pallets/foreign-investments/src/lib.rs
lemunozm: I keep this in mind for a future polish over these changes once we get this working. Anyway, I think the value is already cached because Orderbook has used it just before calling `notify_status_change()`. 

But I'll give thought

pallets/foreign-investments/src/lib.rs
lemunozm: You're right, the amounts mismatch here. I'll think for a clean solution here

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +380,565 @@ pub mod pallet {
...
+									who.clone(),
+									swap_id,
+									amount_to_swap,
+									T::BalanceRatio::one(),
+								)?;
+
+								Ok(SwapStatus {
+									swapped: new_swap.amount_in,
+									pending: T::Balance::zero(),
+									swapped_inverse: new_swap.amount_in,
```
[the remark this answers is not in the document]
lemunozm: After another look, I disagree. The pending amount to swap for the inverse is the amount left in the inverse order; it is `amount_to_swap.`

lemunozm:
Closed in favor of: https://github.com/centrifuge/centrifuge-chain/pull/1698

PR #1693, opened 2024-01-16
Fix: Limitations of Transfer Allowlist

### LINKED ISSUE #1679: fix: pallet-transfer-allowlist filters
### Description
`pallet-transfer-allowlist` currently works as follows:
* no-restriction set &rarr; all receiving addresses for a transfer are valid
* restriction set for a currency &rarr; only allowed receivers are valid receivers, all other receiver will fail
    * all OTHER currencies ARE NOT restricted 

BUT it should rather be
* no-restriction set &rarr; all receiving addresses for a transfer are valid
* restriction set for a currency &rarr; only allowed receivers are valid receivers, all other receiver will fail
    * all OTHER currencies ARE restricted 

### Research/based on
This is an issue that got lost in the weeds of communicating requirements and technical specs.

* Storage is not made for this
* Need to adapt existing filters in runtime common

mustermeiszer:
# Description
See issue also. The main problem is that `pallet-transfer-allowlist` in our current setup does not allow to allow a single account to have permissions for all current and future currencies in one run. This gets problematic, as we are adding more foreign domains with currencies to the system in the future, and our pool admins are not protected to mistakenly send tokens there.

Furthermore, previously, we had not idea of how to restrict native transfers. 

Fixes #1679

## Changes and Descriptions
* Remove `NativeCurrency` from `pallet-transfer-allowlist`
* Introduces `PreBalanceTransferExtension` 
   - a `SignedExtension` that will block any calls of `pallet-balances` that transfer `CFG` if they should no go from `send` to `recv`
   - this is the only viable solution - BUT error resolution is bad, as we are not working with `DispatchErrors` on that level (see discussion [here](https://github.com/paritytech/polkadot-sdk/issues/2563))
* Introduces the `FilterCurrency` enum that represents an allowance for `All` currencies or just for `Specific` one
* Adapts filters in `runtime-common` to actually take into account the usage of `FilterCurrency`

# Checklist:


runtime/altair/src/lib.rs
```diff
@@ -1960,7 +1960,7 @@ mod __runtime_api_use {
 #[cfg(not(feature = "disable-runtime-api"))]
 use __runtime_api_use::*;
 use cfg_types::locations::Location;
-use runtime_common::transfer_filter::PreNativeTransfer;
+use runtime_common::transfer_filter::{FilterCurrency, PreNativeTransfer};
```
lemunozm, line 1963: Wondering if this could incur in any migration

runtime/common/src/transfer_filter.rs
```diff
@@ -13,32 +13,47 @@
...
+				sender.clone(),
 				Location::XCM(BlakeTwo256::hash(&destination.encode())),
-				currency,
+				FilterCurrency::Specific(currency),
 			)
+			.or(T::allowance(
+				sender,
+				Location::XCM(BlakeTwo256::hash(&destination.encode())),
+				FilterCurrency::All,
+			))
```
lemunozm, line 56: A couple of points:
 1. I'm pretty sure the right side of the `or()` is computed before checking whether the left side is true. I think you want here `or_else()` which takes a closure.
 2. If the first is false, and then we check for `FilterCurrency::All`. Is it not the same as just checking it for `FilterCurrency::All`?

runtime/common/src/transfer_filter.rs
```diff
@@ -121,24 +136,111 @@ impl<
...
+			t.send.clone(),
+			Location::Local(t.recv.clone()),
+			FilterCurrency::Specific(t.id),
+		)
+		.is_ok() || T::allowance(
+			t.send.clone(),
+			Location::Local(t.recv.clone()),
+			FilterCurrency::All,
+		)
+		.is_ok()
```
lemunozm, line 155: Same as above, point 2.

runtime/common/src/transfer_filter.rs
```diff
@@ -121,24 +136,111 @@ impl<
...
+		T::allowance(
+			sender.clone(),
+			Location::Address(receiver.clone()),
+			FilterCurrency::Specific(currency),
+		)
+		.or(T::allowance(
+			sender,
+			Location::Address(receiver),
+			FilterCurrency::All,
+		))
```
lemunozm, line 176: Same as above, points 1. and 2.

runtime/common/src/transfer_filter.rs
```diff
@@ -121,24 +136,111 @@ impl<
...
+		pallet_transfer_allowlist::pallet::Pallet::<T>::allowance(
+			who.clone(),
+			Location::Local(recv.clone()),
+			FilterCurrency::All,
+		)
+		.or(pallet_transfer_allowlist::pallet::Pallet::<T>::allowance(
+			who.clone(),
+			Location::Local(recv.clone()),
+			FilterCurrency::Specific(CurrencyId::Native),
+		))
```
lemunozm, line 243: Same as above, points 1. and 2.

runtime/altair/src/lib.rs
```diff
@@ -1715,11 +1715,10 @@ impl pallet_order_book::Config for Runtime {
 }
 
 impl pallet_transfer_allowlist::Config for Runtime {
-	type CurrencyId = CurrencyId;
+	type CurrencyId = FilterCurrency;
```
lemunozm, line 1718: `Currency` is used in `pallet_transfer_allowlist` storages, so this would need migration. Is this pallet already used in centrifuge?

lemunozm (commented):
More simple now!

Just a few comments

runtime/common/src/transfer_filter.rs
mustermeiszer: Good point. Will refactor

mustermeiszer:
Try-runtime against centrifuge: 

```
2024-01-17 16:40:28 TransferAllowlist-CurrencyMigration:  PRE UPGRADE: Starting...    
2024-01-17 16:40:28 TransferAllowlist-CurrencyMigration:  Counted 2 keys in old allowance storage.    
2024-01-17 16:40:28 TransferAllowlist-CurrencyMigration:  Counted 2 keys in old delay storage.    
2024-01-17 16:40:28 TransferAllowlist-CurrencyMigration:  Migrating currency used started...    
2024-01-17 16:40:28 TransferAllowlist-CurrencyMigration:  POST UPGRADE: Starting...    
2024-01-17 16:40:28 TransferAllowlist-CurrencyMigration:  POST UPGRADE: Finished. 
```

runtime/altair/src/lib.rs
mustermeiszer: Done in https://github.com/centrifuge/centrifuge-chain/pull/1693/commits/fc621f78c18b752c31f319e4584568ffed8a394c

runtime/common/src/transfer_filter.rs
mustermeiszer: Done in https://github.com/centrifuge/centrifuge-chain/pull/1693/commits/13906f416fa726d553fdc41c5ac7ea70124afcba

runtime/common/src/transfer_filter.rs
```diff
@@ -13,32 +13,49 @@
...
 use xcm::v3::{MultiAsset, MultiLocation};
 
+#[derive(
+	Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen,
+)]
+pub enum FilterCurrency {
+	All,
+	Specific(CurrencyId),
+}
+
```
lemunozm, line 35: I am still confusing about the second point: 

> If the first is false, and then we check for `FilterCurrency::All`. Is it not the same as just checking it for `FilterCurrency::All`?

Maybe would help adding docs here?

mustermeiszer: where did I write that line? ^^

mustermeiszer: Logically, you can set
* an allowance for an account to be allowed for ALL currencies
* an allowance for an account to be allowed to ONE currency

There might be an overlap if you set account `A` both for `All` and for `Specific(CURRENCY)`. Do you want to prevent that? 
Preventing the other way around is harder, as It would need to iterate over the whole storage of that account.

lemunozm: > where did I write that line? ^^

I wrote it 😆

Ok, I think I follow you now 👍🏻, thanks!

lemunozm (dismissed):
Good work with the migration! 🙌🏻 LGTM

runtime/common/src/migrations/transfer_allowlist_currency.rs
```diff
@@ -0,0 +1,153 @@
...
+const LOG_PREFIX: &str = "TransferAllowlist-CurrencyMigration: ";
+
+struct DelayPrefix;
+impl StorageInstance for DelayPrefix {
+	const STORAGE_PREFIX: &'static str = "AccountCurrencyTransferCountDelay";
+
+	fn pallet_prefix() -> &'static str {
+		"TransferAllowList"
+	}
+}
```
lemunozm, line 40: Out of curiosity, did you try `#[pallet::storage_prefix = "foo"]` this?

mustermeiszer: No, you wrote me too late. ^^ 

I think I will keep for now, as I am not sure, how to add the pallet-prefix with this macro. and lazy to change the code. We will remove it anyways afterwards.

runtime/common/src/transfer_filter.rs
```diff
@@ -175,23 +172,32 @@ impl<T: TransferAllowance<AccountId, CurrencyId = FilterCurrency, Location = Loc
...
+	pub fn new() -> Self {
+		PreBalanceTransferExtension(sp_std::marker::PhantomData::default())
+	}
+}
+
+impl<T: frame_system::Config> sp_std::fmt::Debug for PreBalanceTransferExtension<T> {
+	fn fmt(&self, f: &mut Formatter<'_>) -> sp_std::fmt::Result {
+		f.debug_struct("PreBalanceTransferExtension").finish()
+	}
+}
```
lemunozm, line 189: Maybe you were looking for the derive `RuntimeDebugNoBound` (?)

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -219,22 +219,22 @@ pub mod utils {
 	}
 
 	pub fn ausd(amount: Balance) -> Balance {
-		amount * dollar(currency_decimals::AUSD)
+		amount * decimals(currency_decimals::AUSD)
```
lemunozm, line 222: 👌🏻

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -8899,12 +8901,141 @@ mod centrifuge {
...
+		fn restrict_all<T: Runtime>() {
+			let mut env = RuntimeEnv::<T>::from_parachain_storage(
+				Genesis::default()
+					.add(genesis::balances::<T>(cfg(TRANSFER_AMOUNT + 10)))
+					.add(orml_tokens::GenesisConfig::<T> {
+						balances: vec![(
+							Keyring::Alice.to_account_id(),
+							USDC,
+							T::ExistentialDeposit::get() + usdc(TRANSFER_AMOUNT),
+						)],
```
lemunozm, line 8921: I think there is already a one line utility for just this in `genesis` module

lemunozm (dismissed):
Minor optional comments, but LGTM!

runtime/common/src/migrations/transfer_allowlist_currency.rs
```diff
@@ -0,0 +1,154 @@
...
+		OldAccountCurrencyTransferAllowance::translate::<AllowanceDetails<BlockNumber>, _>(
+			|(account, currency_id, location), allowance| {
+				pallet_transfer_allowlist::AccountCurrencyTransferAllowance::<T>::insert(
+					(account, FilterCurrency::Specific(currency_id), location),
+					allowance,
+				);
+
+				counter += 1;
+				// We will remove the storage here, as we are inserting the new storage above
+				None
```
wischli, line 93: NIT: Interesting approach 🙃 Dealt with this yesterday as well. A cleaner way would be to translate `pallet_transfer_allowlist::AccountCurrencyTransferAllowance::<T>::translate::<AllowanceDetails<BlockNumber>, _>` to reduce the number of writes by `counter`.

runtime/common/src/migrations/transfer_allowlist_currency.rs
```diff
@@ -0,0 +1,154 @@
...
+		T: pallet_transfer_allowlist::Config<
+			AccountId = AccountId,
+			Location = Location,
+			CurrencyId = FilterCurrency,
+			BlockNumber = BlockNumber,
+		>,
+	> OnRuntimeUpgrade for Migration<T>
+{
+	fn on_runtime_upgrade() -> Weight {
+		log::info!("{LOG_PREFIX} Migrating currency used started...");
```
wischli, line 82: NIT: Since this is a breaking storage change, we should bump the `StorageVersion` of `pallet-transfer-allowlist` and check versions. Normally, I would have flagged this as `CR` but no reason to be that pedantic here for the sake of the 1025 deadline.

wischli (commented):
Two non-blocking comments. Thanks for working on this!

runtime/common/src/migrations/transfer_allowlist_currency.rs
mustermeiszer: Curious. I thought about using `translate` but I need to change the `Key` and not the `Value`. Is this possible with translate?

runtime/common/src/migrations/transfer_allowlist_currency.rs
mustermeiszer: Will still do.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
mustermeiszer: Yeah, but that does not work well with the new `add(..)` logic here.

runtime/common/src/transfer_filter.rs
mustermeiszer: Nice.

runtime/common/src/migrations/transfer_allowlist_currency.rs
wischli: Sorry, didn't catch the key change. In this case, `translate` cannot be used as it only provides means of altering values.

runtime/common/src/migrations/transfer_allowlist_currency.rs
mustermeiszer: Changed

mustermeiszer:
I also tested signing transactions via Polkadot-js just now. Works all fine.

lemunozm (dismissed):

wischli (commented):
Unfortunately, I have a pedantic simple change request. Apart from that, looks great! Also tested the migration, works like a charm!

wischli (dismissed):

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -92,7 +96,18 @@ impl<T: Runtime> Env<T> for RuntimeEnv<T> {
 			utils::create_extrinsic::<T>(who, call, nonce)
 		});
 
-		self.parachain_state_mut(|| T::Api::apply_extrinsic(extrinsic).unwrap())?;
+		self.parachain_state_mut(|| {
+			let res = T::Api::apply_extrinsic(extrinsic);
+			// NOTE: This is our custom error that we are having in the
+			//       `PreBalanceTransferExtension` SignedExtension, so we need to
+			//        catch that here.
+			if let Err(TransactionValidityError::Invalid(InvalidTransaction::Custom(255))) = res {
```
lemunozm, line 104: Does `255` come from any constant defined somewhere else, or is it random?

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -101,7 +116,9 @@ impl<T: Runtime> Env<T> for RuntimeEnv<T> {
 				}
 				_ => None,
 			})
-			.unwrap();
+			// NOTE: This is actually not always correct. Even if there is not fee event, there is a
+			//       fee substracted if the `pre_dispatch()` errors out.
```
lemunozm, line 120: Is there any way to get that value from `pre_dispatch()`?

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -2002,7 +2002,7 @@ mod development {
 
 				env.parachain_state_mut(|| {
 					let pool_id = DEFAULT_POOL_ID;
-					let amount = 10 * dollar(12);
+					let amount = 10 * decimals(12);
```
lemunozm, line 2005: @William maybe you can cherry-pick the commit that introduces these changes to avoid a lot of conflicts in your liquidity-pools tests refactor

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -92,7 +96,18 @@ impl<T: Runtime> Env<T> for RuntimeEnv<T> {
 			utils::create_extrinsic::<T>(who, call, nonce)
 		});
 
-		self.parachain_state_mut(|| T::Api::apply_extrinsic(extrinsic).unwrap())?;
+		self.parachain_state_mut(|| {
+			let res = T::Api::apply_extrinsic(extrinsic);
+			// NOTE: This is our custom error that we are having in the
+			//       `PreBalanceTransferExtension` SignedExtension, so we need to
+			//        catch that here.
```
lemunozm, line 103: Why are we having this error? Should we configure something to avoid generating this error?

lemunozm (approved):
Some questions, but LGTM

runtime/integration-tests/src/generic/envs/runtime_env.rs
mustermeiszer: You mean the fee amount? do not think so.

runtime/integration-tests/src/generic/envs/runtime_env.rs
mustermeiszer: Random

runtime/integration-tests/src/generic/envs/runtime_env.rs
mustermeiszer: The `PreBalancesTransferExtension` is filtering out all calls that are `pallet-balances::Transfer*` alike calls, if there is a restriction set and the receiver is not correct. So i.e. the filter for native currency. 

But it is not working on the DispatchError level, but on the validity level, so we can not return the correct errors.

runtime/common/src/transfer_filter.rs
```diff
@@ -231,18 +236,53 @@ where
...
+		// There is NO allowance set for `All`
+		(Ok(None), Ok(None)) => Ok(()),
+		// There is an allowance set for `Specific(id)`, for the given recv
+		// There is NO allowance set for `All`
+		(Ok(Some(_)), Ok(None)) => Ok(()),
+		// There is NO allowance set for `Specific(id)`
+		// There is an allowance set for `All`, for the given recv
+		(Ok(None), Ok(Some(_))) => Ok(()),
+	}
+}
```
lemunozm: I think there is a common factor here. If there is no `Err` or there is some `Some`, then `Ok` else `Err`, but not sure if you prefer to leave it explicitly.

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: Ok, thanks for the explanation

lemunozm: So, even if such an error is dispatched, the extrinsic is considered submitted, right?

mustermeiszer: yes.

runtime/common/src/transfer_filter.rs
lemunozm: Sure!

lemunozm (approved):

PR #1695, opened 2024-01-17
fix: correct epoch execution with fees

mustermeiszer:
# Description
Adaption of the existing feature branch to do a correct epoch execution with the new fees.

# Tasks
- [x] Adapt `EpochExecutionInfo`
- [ ] Write migration for `EpochExecutionInfo`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2738,7 +2738,7 @@ dependencies = [
 
 [[package]]
 name = "development-runtime"
-version = "0.10.37"
+version = "0.10.38"
 dependencies = [
  "axelar-gateway-precompile",
  "cfg-primitives",
--- a/pallets/pool-system/src/lib.rs
+++ b/pallets/pool-system/src/lib.rs
@@ -389,7 +389,7 @@ pub mod pallet {
 		type WeightInfo: WeightInfo;
 	}
 
-	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(2);
 
 	#[pallet::pallet]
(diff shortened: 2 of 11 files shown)


wischli:
> Automatically create protocol-fees, when registering a pool

This is already working on my branch. On registration, you can submit vec of PoolFeesInfo which is added without waiting the 7d Changeguard time.

mustermeiszer:
Must not go into this release. Can be a follow up

wischli (approved):
Thanks for looking into this! LGTM.

The one broken test rooted in the automatic epoch execution trigger during closing because of a healthy pool state.  Beforehand, that was not the case. Since fees are paid during execution, assertions didn't match anymore. Fixed it by adding a senior tranche token redemption to the exisiting jr one. This way, the epoch needs to be closed manually such that all fee assertions hold true.

mustermeiszer:
>Not what I meant. I meant automatically creating a top level root fee for the protocol. With a fixed percentage that is determined by democracy

Nevermind that, we just ensure that the right fees are added when registering a pool via democracy. I think that will be easier, more flexible and fine. So not adding to force add the protoocl fee.

wischli:
@mustermeiszer This PR is ready to be merged. I tested the migration against dev and centrifuge chain. It wasn't easy to find a block in which `EpochExecution` was not empty 😅
```
[2024-01-23T09:29:11Z INFO  runtime_common::migrations::epoch_execution] EpochExecutionMigration:  EpochExecution count pre migration is 1
[2024-01-23T09:29:11Z INFO  runtime_common::migrations::epoch_execution] EpochExecutionMigration:  EpochExecutionV2 count post migration is 1
```

PR #1696, opened 2024-01-17
lp-gateway: Add queue for outbound messages

### LINKED ISSUE #1694: Feat: Implement Queue Logic for Outbound-Queue
### Description
The main reason we need this feature is that for some code-paths, `FI` will get triggered with the following path
* Incoming message via `EVM`
* Processing of message from `FI`
* FI sends an outbound message
* Gateway enters `EVM` to send messages

This triggers a `Error::Reentrancy` on the `pallet-evm`. Although, we do not enable this feature, a dependency does and so we are exposed to that. [See](https://kflabs.slack.com/archives/C04GJQAM9P0/p1705409380219429)

#### Idea
The idea is to use an actual queue for the outbound logic. This queue will queue outgoing messages. This in between step allows to prevent a reentrancy. 
For the MvP the queue should simply use on-idle to process messages and also provide an additional extrinsic to manually process messages.

### Research/based on
* Conversation [here](https://kflabs.slack.com/archives/C04GJQAM9P0/p1705409380219429) after problems with reducing investments

* Increase complexity as there will be an (possibly) automated step to further forward messages to another domain
* Reduce weight of single call logic for `FI`
* Fix bug as described above

None so far.

cdamian:
# Description

Fixes #1694 

## Changes and Descriptions

- Added storage for outbound messages.
- Added `on-idle` hook logic for processing outbound messages.
- Added `DispatchResultWithPostInfo` to routers for retrieving weight.
- Added storage for failed outbound messages.
- Added extrinsics for manually executing messages.

# Checklist:

### DIFF
--- a/pallets/liquidity-pools-gateway/routers/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/routers/src/lib.rs
@@ -29,7 +29,9 @@ pub const GAS_TO_WEIGHT_MULTIPLIER: u64 = 25_000;
 
 use cfg_traits::{ethereum::EthereumTransactor, liquidity_pools::Router};
 use frame_support::{
-	dispatch::{DispatchError, DispatchResult, Weight},
+	dispatch::{
+		DispatchError, DispatchResult, DispatchResultWithPostInfo, PostDispatchInfo, Weight,
+	},
 	ensure,
 	traits::OriginTrait,
 };
@@ -113,7 +115,7 @@ where
 		}
 	}
 
-	fn send(&self, sender: Self::Sender, message: Self::Message) -> DispatchResult {
+	fn send(&self, sender: Self::Sender, message: Self::Message) -> DispatchResultWithPostInfo {
 		match self {
 			DomainRouter::EthereumXCM(r) => r.do_send(sender, message),
 			DomainRouter::AxelarEVM(r) => r.do_send(sender, message),
@@ -159,12 +161,9 @@ where
 	/// pallet, this EVM address will be converted back into a substrate account
 	/// which will be charged for the transaction. This converted substrate
 	/// account is not the same as the original account.
-	pub fn do_send(&self, sender: T::AccountId, msg: Vec<u8>) -> DispatchResult {
+	pub fn do_send(&self, sender: T::AccountId, msg: Vec<u8>) -> DispatchResultWithPostInfo {
 		let sender_evm_address = H160::from_slice(&sender.as_ref()[0..20]);
 
-		// TODO(cdamian): This returns a `DispatchResultWithPostInfo`. Should we
-		// propagate that to another layer that will eventually charge for the
-		// weight in the PostDispatchInfo?
 		<pallet_ethereum_transaction::Pallet<T> as EthereumTransactor>::call(
 			sender_evm_address,
 			self.evm_domain.target_contract_address,
@@ -173,9 +172,6 @@ where
 			self.evm_domain.fee_values.gas_price,
 			self.evm_domain.fee_values.gas_limit,
 		)
-		.map_err(|e| e.error)?;
-
-		Ok(())
 	}
 }
 
@@ -231,7 +227,7 @@ where
 
 	/// Encodes the message to the required format and executes the
 	/// call via the XCM transactor pallet.
-	pub fn do_send(&self, sender: T::AccountId, msg: Vec<u8>) -> DispatchResult {
+	pub fn do_send(&self, sender: T::AccountId, msg: Vec<u8>) -> DispatchResultWithPostInfo {
 		let ethereum_xcm_call = get_encoded_ethereum_xcm_call::<T>(self.xcm_domain.clone(), msg)
 			.map_err(|_| DispatchError::Other("encoded ethereum xcm call retrieval"))?;
 
@@ -257,7 +253,10 @@ where
 			true,
 		)?;
 
-		Ok(())
+		Ok(PostDispatchInfo {
+			actual_weight: Some(self.xcm_domain.overall_weight),
+			pays_fee: Default::default(),
+		})
 	}
 }
 
--- a/pallets/liquidity-pools-gateway/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/src/lib.rs
@@ -707,7 +707,7 @@ pub mod pallet {
 
 			let router = DomainRouters::<T>::get(domain).ok_or(DispatchErrorWithPostInfo {
 				post_info: PostDispatchInfo {
-					actual_weight: Some(read_weight.clone()),
+					actual_weight: Some(read_weight),
 					pays_fee: Pays::Yes,
 				},
 				error: Error::<T>::RouterNotFound.into(),
--- a/pallets/liquidity-pools-gateway/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/src/lib.rs
@@ -19,9 +19,13 @@ use cfg_traits::{
 };
 use cfg_types::domain_address::{Domain, DomainAddress};
 use frame_support::{dispatch::DispatchResult, pallet_prelude::*, PalletError};
-use frame_system::pallet_prelude::OriginFor;
+use frame_system::{
+	ensure_signed,
+	pallet_prelude::{BlockNumberFor, OriginFor},
+};
 pub use pallet::*;
 use parity_scale_codec::{EncodeLike, FullCodec};
+use sp_runtime::traits::{AtLeast32BitUnsigned, EnsureAdd, One};
 use sp_std::{convert::TryInto, vec::Vec};
 
 use crate::weights::WeightInfo;
@@ -56,6 +60,10 @@ impl<T: Config> From<RelayerMessageDecodingError> for Error<T> {
 pub mod pallet {
 	const BYTES_U32: usize = 4;
 	const BYTES_ACCOUNT_20: usize = 20;
+	const DEFAULT_WEIGHT_REF_TIME: u64 = 5_000_000_000;
+
+	use frame_support::dispatch::PostDispatchInfo;
+	use sp_runtime::DispatchErrorWithPostInfo;
 
 	use super::*;
 	use crate::RelayerMessageDecodingError::{
@@ -94,7 +102,7 @@ pub mod pallet {
 		///
 		/// NOTE - this `Codec` trait is the Centrifuge trait for liquidity
 		/// pools' messages.
-		type Message: Codec + Clone + Debug + PartialEq;
+		type Message: Codec + Clone + Debug + PartialEq + MaxEncodedLen + TypeInfo + FullCodec;
 
 		/// The message router type that is stored for each domain.
 		type Router: DomainRouter<Sender = Self::AccountId, Message = Self::Message>
@@ -122,6 +130,17 @@ pub mod pallet {
 		/// implementation.
 		#[pallet::constant]
 		type Sender: Get<Self::AccountId>;
+
+		/// Type used for outbound message identification.
+		type OutboundMessageNonce: Parameter
+			+ Member
+			+ AtLeast32BitUnsigned
+			+ Default
+			+ Copy
+			+ EnsureAdd
+			+ MaybeSerializeDeserialize
+			+ TypeInfo
+			+ MaxEncodedLen;
 	}
 
 	#[pallet::event]
@@ -148,13 +167,31 @@ pub mod pallet {
 			domain: Domain,
 			message: T::Message,
 		},
+
+		/// Outbound message execution failure.
+		OutboundMessageExecutionFailure {
+			nonce: T::OutboundMessageNonce,
+			sender: T::AccountId,
+			domain: Domain,
+			message: T::Message,
+			error: DispatchError,
+		},
+
+		/// Outbound message execution success.
+		OutboundMessageExecutionSuccess {
+			nonce: T::OutboundMessageNonce,
+			sender: T::AccountId,
+			domain: Domain,
+			message: T::Message,
+		},
 	}
 
 	/// Storage for domain routers.
 	///
 	/// This can only be set by an admin.
 	#[pallet::storage]
-	pub(crate) type DomainRouters<T: Config> = StorageMap<_, Blake2_128Concat, Domain, T::Router>;
+	#[pallet::getter(fn domain_routers)]
+	pub type DomainRouters<T: Config> = StorageMap<_, Blake2_128Concat, Domain, T::Router>;
 
 	/// Storage that contains a limited number of whitelisted instances of
 	/// deployed liquidity pools for a particular domain.
@@ -162,7 +199,7 @@ pub mod pallet {
 	/// This can only be modified by an admin.
 	#[pallet::storage]
 	#[pallet::getter(fn allowlist)]
-	pub(crate) type Allowlist<T: Config> =
+	pub type Allowlist<T: Config> =
 		StorageDoubleMap<_, Blake2_128Concat, Domain, Blake2_128Concat, DomainAddress, ()>;
 
 	/// Storage that contains a limited number of whitelisted instances of
@@ -171,8 +208,34 @@ pub mod pallet {
 	/// This can only be modified by an admin.
 	#[pallet::storage]
 	#[pallet::getter(fn relayer)]
-	pub(crate) type RelayerList<T: Config> =
+	pub type RelayerList<T: Config> =
 		StorageDoubleMap<_, Blake2_128Concat, Domain, Blake2_128Concat, DomainAddress, ()>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn outbound_message_nonce_store)]
+	pub type OutboundMessageNonceStore<T: Config> =
+		StorageValue<_, T::OutboundMessageNonce, ValueQuery>;
+
+	/// Storage for outbound messages that will be processed during the
+	/// `on_idle` hook.
+	#[pallet::storage]
+	#[pallet::getter(fn outbound_message_queue)]
+	pub type OutboundMessageQueue<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat,
+		T::OutboundMessageNonce,
+		(Domain, T::AccountId, T::Message),
+	>;
+
+	/// Storage for failed outbound messages that can be manually re-triggered.
+	#[pallet::storage]
+	#[pallet::getter(fn failed_outbound_messages)]
+	pub type FailedOutboundMessages<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat,
+		T::OutboundMessageNonce,
+		(Domain, T::AccountId, T::Message, DispatchError),
+	>;
 
 	#[pallet::error]
 	pub enum Error<T> {
@@ -212,6 +275,19 @@ pub mod pallet {
 		/// Decoding that is essential and this error
 		/// signals malforming of the wrapping information.
 		RelayerMessageDecodingFailed { reason: RelayerMessageDecodingError },
+
+		/// Outbound message not found in storage.
+		OutboundMessageNotFound,
+
+		/// Failed outbound message not found in storage.
+		FailedOutboundMessageNotFound,
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_idle(_now: T::BlockNumber, max_weight: Weight) -> Weight {
+			Self::service_outbound_message_queue(max_weight)
+		}
 	}
 
 	#[pallet::call]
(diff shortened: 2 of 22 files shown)


pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -174,6 +192,19 @@ pub mod pallet {
...
+	pub(crate) type OutboundMessageQueue<T: Config> = StorageNMap<
+		_,
+		(
+			NMapKey<Blake2_128Concat, Domain>,
+			NMapKey<Blake2_128Concat, T::AccountId>,
+			NMapKey<Blake2_128Concat, T::Message>,
+		),
+		(),
+		ValueQuery,
+	>;
```
lemunozm, line 206: Q: Why is the message part of the key instead of the value?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +500,65 @@ pub mod pallet {
...
+				weight_used = weight_used.saturating_add(weight);
+
+				if weight_used.all_gte(max_weight) {
+					break;
+				}
+			}
+
+			for entry in processed_entries {
+				OutboundMessageQueue::<T>::remove(entry);
+			}
```
lemunozm, line 558: Good point removing out from the iteration!

lemunozm (commented):
LGTM!

Just a question

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: If we were to use a double map with k1: domain, k2: sender, we would then need to keep a vector of messages. I wanted to avoid doing that and having to mingle with that vector every time we process something. This approach is also easier to iterate on IMO.

pallets/liquidity-pools-gateway/routers/src/lib.rs
```diff
@@ -257,7 +253,10 @@ where
 			true,
 		)?;
 
-		Ok(())
+		Ok(PostDispatchInfo {
+			actual_weight: Some(self.xcm_domain.overall_weight),
```
cdamian, line 257: The above call to `transact_through_sovereign` returns a normal `DispatchResult`, meaning that we cannot get the weight of the call. If there is another way of retrieving the weight of that call, please let me know and I will adjust.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -56,6 +56,7 @@ impl<T: Config> From<RelayerMessageDecodingError> for Error<T> {
 pub mod pallet {
 	const BYTES_U32: usize = 4;
 	const BYTES_ACCOUNT_20: usize = 20;
+	const DEFAULT_WEIGHT_REF_TIME: u64 = 100_000_000;
```
cdamian, line 59: Please let me know if this value makes sense.

mustermeiszer: THat is equal to the weight of a single write. I think it should be a magnitude higher and times 5 maybe.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +501,66 @@ pub mod pallet {
...
+							error: e.error,
+						});
+
+						e.post_info
+					}
+				};
+
+				let weight = res
+					.actual_weight
+					.or(Some(Weight::from_parts(DEFAULT_WEIGHT_REF_TIME, 0)))
```
mustermeiszer, line 548: PoV weight is unfortunately not negligible here. The pov weight would at least be `MaxEnodedLen(T::Domain) + MaxEnodedLen(T::AccountId) + MaxEnodedLen(T::Message)` for each iterated item.

mustermeiszer: And this weight is always added. Not Matter if `actual_weight` is `Some` or not.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -171,9 +191,24 @@ pub mod pallet {
...
+	/// Storage for outbound messages that will be processed during the
+	/// `on_idle` hook.
+	#[pallet::storage]
+	#[pallet::getter(fn outbound_message_queue)]
+	pub type OutboundMessageQueue<T: Config> = StorageNMap<
+		_,
+		(
+			NMapKey<Blake2_128Concat, Domain>,
+			NMapKey<Blake2_128Concat, T::AccountId>,
+			NMapKey<Blake2_128Concat, T::Message>,
```
mustermeiszer, line 206: This means, the same message will overwrite past once, but it should be possible to send the same message twice. We need ids unfortunately for the queue.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -214,6 +246,13 @@ pub mod pallet {
 		RelayerMessageDecodingFailed { reason: RelayerMessageDecodingError },
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_idle(_now: T::BlockNumber, max_weight: Weight) -> Weight {
+			Self::service_outbound_message_queue(max_weight)
+		}
```
mustermeiszer, line 253: Additionally I would add an extrinsic to progress a specific message manually inc ases the chain is busy.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -485,13 +584,13 @@ pub mod pallet {
...
 			);
 
-			let router =
-				DomainRouters::<T>::get(destination.clone()).ok_or(Error::<T>::RouterNotFound)?;
-
-			router.send(T::Sender::get(), message.clone())?;
+			OutboundMessageQueue::<T>::insert(
+				(destination.clone(), T::Sender::get(), message.clone()),
+				(),
+			);
```
mustermeiszer, line 590: I would add the checks for missing routers here and fail in theses cases.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +501,66 @@ pub mod pallet {
...
+
+			for ((domain, sender, message), _) in OutboundMessageQueue::<T>::iter() {
+				processed_entries.push((domain.clone(), sender.clone(), message.clone()));
+
+				let router = match DomainRouters::<T>::get(domain.clone()) {
+					Some(router) => router,
+					None => {
+						Self::deposit_event(Event::RouterNotFound {
+							domain: domain.clone(),
+						});
```
mustermeiszer, line 518: If this error occurs, the message should be stored in an error storage with the problem that happened, or we we stop processing.

mustermeiszer: If we check the router when submitting, this would mean that intermediately, the router has been removed, but still state changes on Centrifuge Chain happened. So an error storage is the best IMO.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +501,66 @@ pub mod pallet {
...
+
+						r
+					}
+					Err(e) => {
+						Self::deposit_event(Event::RouterExecutionFailure {
+							sender: sender.clone(),
+							domain: domain.clone(),
+							message: message.clone(),
+							error: e.error,
+						});
```
mustermeiszer, line 540: Also an error storage would be best IMO.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +501,66 @@ pub mod pallet {
...
+				if weight_used.all_gte(max_weight) {
+					break;
+				}
+			}
+
+			for entry in processed_entries {
+				OutboundMessageQueue::<T>::remove(entry);
+			}
+
+			weight_used
```
mustermeiszer, line 562: The weight also includes 1 read and 1 write per entry, which need to added in the for loop.

mustermeiszer (commented):
First look.

mustermeiszer:
once this is realeased: We have to re-execute https://axelarscan.io/gmp/0x15c33b374f44725927e0025a3e3b07ebb99bc844307e2563161df8de6a331ac4:11

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Added an extrinsic for that.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Updated the weights here.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: I updated all weight calculations here.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Added a nonce for keeping track of each outbound/failed message.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Saving the error along with the message info in the `FailedOutboundMessages` storage.

mustermeiszer (commented):
Logic looks good! Need to check the weights again in more detail. Probably need @wischli eyes on this too! 

Should we add anymore test? I would like to bring this into the next RU if possible.

gpmayorga:
there was an error in the pipeline due to Gcloud permissions. It is fixed but it seems like Clippy is unhappy still

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +621,139 @@ pub mod pallet {
...
+							message,
+						});
+
+						weight = weight
+							.saturating_add(T::DbWeight::get().writes(1))
+							.saturating_add(
+								post_info
+									.actual_weight
+									.expect("expected message processing weight"),
+							);
```
wischli, line 656: Two nits:
1. Cleaner way is to use `weight.saturating_accrue` 
2. The expect message is a bit verbose. Since expecting can panic and thus should generally not be used, I'd prefer to point to the `Ok(...)` match here either via comment or the expect message itself to make it explicit.
```suggestion
						weight.saturating_accrue(T::DbWeight::get().writes(1)
							.saturating_accrue(
								post_info
									.actual_weight
									.expect("Message processing success already ensured; qed"),
							);
```

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +621,139 @@ pub mod pallet {
...
+							(domain, sender, message, e.error),
+						);
+
+						weight = weight
+							.saturating_add(T::DbWeight::get().writes(2))
+							.saturating_add(
+								e.post_info
+									.actual_weight
+									.expect("expected message processing weight"),
+							);
```
wischli, line 678: Nit: see above

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +621,139 @@ pub mod pallet {
...
+					}
+				}
+
+				weight_used = weight_used.saturating_add(weight);
```
wischli, line 682: Nit: The match could directly return the processing `weight` such that we don't need to handle two mutable `weights`.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +621,139 @@ pub mod pallet {
...
+					Ok(post_info) => {
+						Self::deposit_event(Event::OutboundMessageExecutionSuccess {
+							nonce,
+							sender,
+							domain,
+							message,
+						});
+
+						weight = weight
+							.saturating_add(T::DbWeight::get().writes(1))
```
wischli, line 651: CR: I suppose the `write` here refers to the removal of the entry. We also need to account for the read operation
```suggestion
							.saturating_add(T::DbWeight::get().reads_writes(1, 1))
```

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +621,139 @@ pub mod pallet {
...
+							error: e.error,
+						});
+
+						FailedOutboundMessages::<T>::insert(
+							nonce,
+							(domain, sender, message, e.error),
+						);
+
+						weight = weight
+							.saturating_add(T::DbWeight::get().writes(2))
```
wischli, line 673: CR: Should factor in the read as well.

pallets/liquidity-pools-gateway/src/lib.rs
wischli: The weight for reading and removing should be already applied in the iteration. Assume we reach the max weight in the iteration, then we cannot remove the entry anymore. Wondering why we need to accrue by 1 write per entry here again?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -423,6 +499,89 @@ pub mod pallet {
...
+		pub fn process_outbound_message(
+			origin: OriginFor<T>,
+			nonce: T::OutboundMessageNonce,
+		) -> DispatchResult {
+			ensure_signed(origin)?;
+
+			let (domain, sender, message) =
+				OutboundMessageQueue::<T>::get(nonce).ok_or(Error::<T>::OutboundMessageNotFound)?;
+
+			OutboundMessageQueue::<T>::remove(nonce);
```
wischli, line 518: Nit: Just `take`

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -423,6 +499,89 @@ pub mod pallet {
...
+			match Self::process_message(domain.clone(), sender.clone(), message.clone()) {
+				Ok(_) => {
+					Self::deposit_event(Event::<T>::OutboundMessageExecutionSuccess {
+						nonce,
+						domain,
+						sender,
+						message,
+					});
+
+					FailedOutboundMessages::<T>::remove(nonce);
```
wischli, line 568: Q: AFAICS, fault messages which fail to succeed will congest our `FailedOutboundMessages` storage because they cannot be removed. Is that desired for now to keep things simple?

cc @mustermeiszer

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +621,139 @@ pub mod pallet {
...
+		) -> Weight {
+			let pov_weight: u64 = (Domain::max_encoded_len()
+				+ T::AccountId::max_encoded_len()
+				+ T::Message::max_encoded_len())
+			.try_into()
+			.expect("can calculate outbound message POV weight");
+
+			router_call_weight
+				.or(Some(Weight::from_parts(DEFAULT_WEIGHT_REF_TIME, 0)))
+				.unwrap()
```
wischli, line 753: Nit: 
```suggestion
				.unwrap_or(Weight::from_parts(DEFAULT_WEIGHT_REF_TIME, 0))
```

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -485,13 +777,23 @@ pub mod pallet {
...
-			let router =
-				DomainRouters::<T>::get(destination.clone()).ok_or(Error::<T>::RouterNotFound)?;
+			ensure!(
+				DomainRouters::<T>::contains_key(destination.clone()),
+				Error::<T>::RouterNotFound
+			);
 
-			router.send(T::Sender::get(), message.clone())?;
+			<OutboundMessageNonceStore<T>>::try_mutate(|n| {
+				*n = n.ensure_add(T::OutboundMessageNonce::one())?;
```
wischli, line 786: Nit: 
```suggestion
				*n.ensure_add_assign(T::OutboundMessageNonce::one())?;
```

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -485,13 +777,23 @@ pub mod pallet {
...
+				Ok::<_, DispatchError>(())
```
wischli, line 787: Nit: Could return `*n` here to not have to read the value again below

```suggestion
				Ok::<T:: OutboundMessageNonce, DispatchError>(*n)
```

wischli (commented):
Looks really good and clean! I have one minor, pedantic change request for the weight calculation because we are introducing an inherent here were being correct is important.

Apart from that, sorry about the large list of nits which can be ignored 🫠

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -56,6 +60,10 @@ impl<T: Config> From<RelayerMessageDecodingError> for Error<T> {
 pub mod pallet {
 	const BYTES_U32: usize = 4;
 	const BYTES_ACCOUNT_20: usize = 20;
+	const DEFAULT_WEIGHT_REF_TIME: u64 = 5_000_000_000;
```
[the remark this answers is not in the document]
cdamian: This conversation here - https://github.com/centrifuge/centrifuge-chain/pull/1696#discussion_r1456370592

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: No, this `write` is for the event deposit operation. Any other weights should be added accordingly in `process_message`.

cdamian: I did add the missing 1 read tho, thanks for catching that!

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: `saturating_accrue` cannot be used like that since it's return is `()`. I suggest we keep the `add` as it is since it's easier to read imo.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done. Thank you!

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: I changed the error message slightly since this should never happen anyway.

runtime/altair/src/liquidity_pools.rs
```diff
@@ -100,6 +100,7 @@ impl pallet_liquidity_pools_gateway::Config for Runtime {
 	type MaxIncomingMessageSize = MaxIncomingMessageSize;
 	type Message = LiquidityPoolsMessage;
 	type OriginRecovery = crate::LiquidityPoolsAxelarGateway;
+	type OutboundMessageNonce = u64;
```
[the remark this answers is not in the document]
cdamian: Completely forgot about that, thank you!

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: That extra write was for the removal of the entry which was not accounted for in the iteration. That was changed now, PTAL.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: We could also add some extrinsics for removing these as long as there's an admin doing it. Please let us know what you think @mustermeiszer

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
wischli: Nit: Could add that as a comment to the code :)

pallets/liquidity-pools-gateway/src/lib.rs
wischli: Whatever you prefer. Of course we don't reassign when using `accrue` but just alter the mutable `weight` variable.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -462,6 +619,140 @@ pub mod pallet {
...
+								message,
+							});
+
+							post_info
+								.actual_weight
+								.expect("Message processing success already ensured")
+								// Extra weight breakdown:
+								//
+								// 1 read for the outbound message
+								// 1 write for the event
```
wischli, line 653: Dispatching an event is much cheaper than performing a `write` operation. But we can certainly keep the `1 write` here to be defensive.

wischli (dismissed):
Thanks a lot for adhering to my pedantic nit picks 🫶. We only need to discuss the missing cleanup possibility of failure-safe messages, which IMO does not have to be part of this PR.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -423,6 +504,87 @@ pub mod pallet {
...
+
+					Ok(())
+				}
+			}
+		}
+
+		/// Manually process a failed outbound message.
+		#[pallet::weight(T::WeightInfo::process_failed_outbound_message())]
+		#[pallet::call_index(7)]
+		pub fn process_failed_outbound_message(
```
cdamian, line 553: Note how we always return `Ok(())` here so that we can preserve the storage changes in case of errors.

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
wischli: Thanks a lot for adding the context 🫶

wischli (approved):
Great test additions!

pallets/liquidity-pools-gateway/src/tests.rs
```diff
@@ -835,26 +955,162 @@ mod outbound_queue_impl {
...
+			assert_ok!(LiquidityPoolsGateway::submit(
+				sender.clone(),
+				domain.clone(),
+				msg.clone()
+			));
+
+			let expected_nonce = OutboundMessageNonce::one();
+
+			let queue_entry = OutboundMessageQueue::<Runtime>::take(expected_nonce)
+				.expect("an entry is added to the queue");
```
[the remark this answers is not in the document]
cdamian: No reason in particular, I'll change it to `get` to make this less confusing.

wischli (approved):

PR #1698, opened 2024-01-22
FI v2: refactor without states

lemunozm:
# Description

The main idea of this PR is to remove states to easier the adoption of https://github.com/centrifuge/centrifuge-chain/pull/1674.

### Crate organization

- `swaps.rs`: It only focuses on making swaps. Doesn't know about any `Info` type or info Storage. It only reads the storage related to the swap ids. Actually, this file could be moved to its own pallet with the storage Ids if we need swapping for other purposes in the future.
- `entities.rs`: Defines the `InvestmentInfo` and `RedemptionInfo` with some methods that mutate their fields. Only these methods can mutate the internal states of the `Info` types. Higher layers should only read the content and never mutate `Info` fields directly. This is the low-level logic.
- `impls.rs`: The interface to handle FI, through the pallet or through the Hooks. The higher logic level "easier to follow"

### Pending work

- [x] Fix tests from other crates that uses FI stuff in their tests
- [x] Fix `integration-tests`: https://github.com/centrifuge/centrifuge-chain/pull/1700
- [x] Fix benchmarking process


lemunozm:
Currently compiling with `cargo check -p pallet-foreign-investments`

pallets/foreign-investments/src/lib.rs
```diff
@@ -154,20 +249,8 @@ pub mod pallet {
...
-		/// swaps created/updated through Foreign Investments.
-		///
-		/// Example: Say this rate is set to 3/2, then the incoming currency
-		/// should never cost more than 1.5 of the outgoing currency.
-		///
-		/// NOTE: Can be removed once we implement a
-		/// more sophisticated swap price discovery. For now, this should be set
-		/// to one.
-		#[pallet::constant]
-		type DefaultTokenSellRatio: Get<Self::BalanceRatio>;
```
lemunozm, line 167: BTW, I've removed this thinking that later we'll use a market ratio, and for now, all swaps are 1:1. Just for the sake of simplifying things now.

pallets/foreign-investments/src/lib.rs
```diff
@@ -43,43 +43,137 @@
...
+	fn new(
+		investment_id: T::InvestmentId,
+		foreign_currency: T::CurrencyId,
+	) -> Result<Self, DispatchError> {
+		Ok(Self {
+			base: BaseInfo::new(investment_id, foreign_currency)?,
+			swap_id: None,
+			total_tranche_tokens: T::Balance::default(),
+			swapped_amount: T::Balance::default(),
+			tranche_tokens_sent: T::Balance::default(),
```
wischli, line 142: I am having hard times to follow this whole new syntax, probably because it is a bit mixed here and there. Of course, I understand this is all not set in stone and thus not perfect. What does `sent` here refer to? The amount of tranche tokens which were collected during epoch fulfillment?

pallets/foreign-investments/src/lib.rs
```diff
@@ -43,43 +43,137 @@
...
+
+impl<T: Config> RedemptionInfo<T> {
+	fn new(
+		investment_id: T::InvestmentId,
+		foreign_currency: T::CurrencyId,
+	) -> Result<Self, DispatchError> {
+		Ok(Self {
+			base: BaseInfo::new(investment_id, foreign_currency)?,
+			swap_id: None,
+			total_tranche_tokens: T::Balance::default(),
```
wischli, line 140: What does `total` here refer to? The amount of tranche tokens which are pending for fulfillment?

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +372,422 @@ pub mod pallet {
...
+
+							T::CollectedForeignRedemptionHook::notify_status_change(
+								(who.clone(), investment_id),
+								ExecutedForeignCollect {
+									currency: info.base.foreign_currency,
+									amount_currency_payout: last_swap.amount_in,
+									amount_tranche_tokens_payout: tranche_tokens,
+									amount_remaining: info.remaining_tranche_tokens()?,
+								},
+							)?;
```
wischli, line 711: I need to come back to this entire block tomorrow. Can't fully follow. 

For now, I would like to ask the question why we don't decrement `info.total_tranche_tokens`. AFAICS, this is missing such that `remaining_tranche_tokens` would never be zero right now.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +372,422 @@ pub mod pallet {
...
+
+						if info.base.collected.amount_collected == info.swapped_amount {
+							let tranche_tokens = info.pending_tranche_tokens_to_send()?;
+							info.tranche_tokens_sent.ensure_add_assign(tranche_tokens)?;
+
+							T::CollectedForeignRedemptionHook::notify_status_change(
+								(who.clone(), investment_id),
+								ExecutedForeignCollect {
+									currency: info.base.foreign_currency,
+									amount_currency_payout: last_swap.amount_in,
```
wischli, line 707: Why do we increment the `info.swapped_amount` but then only use `last_swap.amount_in` here? Shouldn't we use `info.swapped_amount`?

wischli (commented):
Definitely preferred over Approach2! A few questions and comments. Will resume with the `ForeignRedemptionInfo::<T>::mutate_exists` and both `Collected*Hook`s blocks tomorrow which are still missing.

pallets/foreign-investments/src/lib.rs
```diff
@@ -43,43 +43,137 @@
...
+	}
+}
+
+/// Hold the information of a foreign investment
+#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+#[scale_info(skip_type_params(T))]
+pub struct InvestmentInfo<T: Config> {
+	base: BaseInfo<T>,
+	increase_swap_id: Option<T::SwapId>,
+	decrease_swap_id: Option<T::SwapId>,
```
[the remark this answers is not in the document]
lemunozm: Indirectly, yes. Now, each swap an account does over the same `(AccountId, InvestmentId)` is not mixed. So, the swapping actions would be handled separately. In this case, a foreign investment can have 2 kinds of swappings: when increments and when decrements. And the `InvestmentInfo` handles that information.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +372,422 @@ pub mod pallet {
...
+						}
+					} else if info.decrease_swap_id == Some(swap_id) {
+						T::DecreasedForeignInvestOrderHook::notify_status_change(
+							(who.clone(), investment_id),
+							ExecutedForeignDecreaseInvest {
+								amount_decreased: last_swap.amount_in,
+								foreign_currency: info.base.foreign_currency,
+								amount_remaining: pending_amount,
+							},
+						)?;
```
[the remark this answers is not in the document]
lemunozm: Ah, ok, so we want to minimize the number of msgs sent to just 1 (per partial collection, I guess), right?

pallets/foreign-investments/src/lib.rs
lemunozm: Sorry, the naming for these variables may not be the best. I am open to suggestions. And I should add docs to them, too.

`tranche_tokens_sent` is the number of tranche tokens already notified by an `ExecutedForeignCollected.` I need to keep this information so I can set the correct value for the next `ExecutedForeignCollected` event, it is, the unnotified amount.

pallets/foreign-investments/src/lib.rs
lemunozm: It refers to the number of tranche tokens the user has decided to redeem via increase/decrease. The expected amount. Nothing to do with collected or swapped.

pallets/foreign-investments/src/lib.rs
lemunozm: Yeah, this is the most tricky part. I'll try to re-thing if I could leave it more readable.

`total_tranche_tokens` is the amount the user expects to redeem. Incremented and decremented in `increase/decrease` methods.

When the redemption is collected [here](https://github.com/centrifuge/centrifuge-chain/pull/1698/files#diff-3d5d99980cfa0ce4827916a65ca67084aaee0f5b61eb6c07d6e086431604f93bR777) the tranche amount is aggregated in `info.collected.amount_payment`. So here in line `709`, we are actually doing: `amount_remaining = info.total_tranche_tokens - info.base.collected.amount_payment`

pallets/foreign-investments/src/lib.rs
lemunozm: `info.swapped_amount` is the total amount of all swaps done until this moment. It's only used to determine whether we've swapped the same amount as we've collected, in order to dispatch the `ExecutedForeignCollect` event.

But I think you're right regarding the fact that we should not use `last_swap.amount_in` because it is only the amount of the last swap. We should use the amount from the last `ExecutedForeignCollect`

pallets/foreign-investments/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I'm a bit out of this scope. The idea is to send the event each time we collect but not each time we swap, isn't it?

pallets/foreign-investments/src/lib.rs
lemunozm: Maybe `total_` to `expected_` clarifies this?

lemunozm:
Thanks for the review @wischli, I'll try to think if I can have a better model to store the information required to generate the notifications

pallets/foreign-investments/src/lib.rs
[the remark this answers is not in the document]
wischli: Exactly. For investments, we want to send the `ExecutedDecreaseInvest` message once the entire swap from pool to foreign is completed. For redemptions, we do the same for `ExecutedCollectInvest`.

pallets/foreign-investments/src/lib.rs
wischli: I propose pending because they are `pending` for redemption. That also aligns with the language we settled with for the `InvestmentPortfolio`.

pallets/foreign-investments/src/lib.rs
wischli: Thanks for the explanation! Then I propose to call this `redeemed_` or `notified_tranche_tokens`.

pallets/foreign-investments/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Thanks for this clarification!

pallets/foreign-investments/src/lib.rs
lemunozm: The "problem" I see with `pending` is that it can be confused with `pending` to collect or `pending` to swap. Actually, I think it's not `pending` as it works right now because after the user redeems part of it, this variable does not change.

Let me consider if I can rename/refactor this to better understand it 🤔

pallets/foreign-investments/src/lib.rs
lemunozm: I have managed to eliminate that variable in this simplification: [80f5fa9](https://github.com/centrifuge/centrifuge-chain/pull/1698/commits/80f5fa91ecc4788083b7f213ec182988aa83560c)

pallets/foreign-investments/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Regarding this, and looking at your code, I think `amount_remaining` should be also changed to be the current investment at that moment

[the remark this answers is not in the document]
lemunozm: BTW, that amount coming from the call of `T::Investment::investment()` should be transformed to `foreign_currency`, isn't it?

pallets/foreign-investments/src/lib.rs
lemunozm: Fixed

pallets/foreign-investments/src/lib.rs
lemunozm: Refactored here [38347ff](https://github.com/centrifuge/centrifuge-chain/pull/1698/commits/38347ff110bd287b652bddea40e0eed147f4c121) following your suggestion

pallets/foreign-investments/src/lib.rs
lemunozm: No longer apply since [38347ff](https://github.com/centrifuge/centrifuge-chain/pull/1698/commits/38347ff110bd287b652bddea40e0eed147f4c121)

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +351,446 @@ pub mod pallet {
...
+							T::DecreasedForeignInvestOrderHook::notify_status_change(
+								(who.clone(), investment_id),
+								ExecutedForeignDecreaseInvest {
+									amount_decreased: info.decrease_swapped_amount,
+									foreign_currency: info.base.foreign_currency,
+									amount_remaining: remaining_foreign_amount,
+								},
+							)?;
+
+							info.decrease_swapped_amount = T::Balance::default();
```
wischli, line 676: We are mutating and using `decrease_swapped_amount` only in this scope. Right now, it is redundant because:

1. We increment it by `last_swap.amount_in` (L656)
2. We use it for `ExecutedForeignDecreaseInvest.amount_decreased` (L670)
3. Then, we always reset it to zero  (L676)

So anytime we hit this scope, `info.decrease_swapped_amount` is zero and we could just use `last_swap.amount_in` for L670

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +351,446 @@ pub mod pallet {
...
+							info.decrease_swap_id = None;
```
wischli, line 677: IIUC, we reset all `swap` data here except for the `foreign_currency` such that any `(investor_address, investment_id)` pair right now is limited to their initial `foreign_currency` which is never reset, even if all swaps, investments and redemptions are cleared. 

There should be some condition which clears `ForeignInvestmentInfo`. IMO, this should be the case if there is neither a swap, nor a pending investment.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +351,446 @@ pub mod pallet {
...
+
+				// NOTE: How make this works with market ratios?
+				let remaining_foreign_amount = T::CurrencyConverter::stable_to_stable(
+					info.base.foreign_currency,
+					Pallet::<T>::pool_currency_of(investment_id)?,
+					T::Investment::investment(&who, investment_id)?,
+				)?;
+
+				// NOTE: How make this works with market ratios?
+				let foreign_amount = T::CurrencyConverter::stable_to_stable(
```
wischli, line 743: Nit: To improve readability, could we rename this to 
```suggestion
				let collected_foreign_amount = T::CurrencyConverter::stable_to_stable(
```

wischli (commented):
Great improvements since last evening, much cleaner and clearer now to me! Overall looks really good! Two comments and one nit from me.

pallets/foreign-investments/src/lib.rs
lemunozm: Totally agree!

pallets/foreign-investments/src/lib.rs
lemunozm: Supposing the swap was a partial swap, then `pending_amount` is not 0. So in this case it's not redundant. The purpose of `decrease_swapped_amount` is to accumulate the partially swapped amounts

pallets/foreign-investments/src/lib.rs
lemunozm: Good point! I think you're right. I was clearing line `759` under the investment collected, but the user could decrease everything without collecting. In this case I should clean where you say.

Thanks for your insightful comments!

lemunozm: Fixed. I think this is the correct check: [eb43c1f](https://github.com/centrifuge/centrifuge-chain/pull/1698/commits/eb43c1fa36394997bc5854594c5718c069152860), ensuring that we only remove if there is no more pending investment collects.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +380,673 @@ pub mod pallet {
...
+			)?;
+
+			ForeignInvestmentInfo::<T>::mutate(&who, investment_id, |info| {
+				let info = info.get_or_insert(InvestmentInfo::new(foreign_currency)?);
+
+				info.base.ensure_same_foreign(foreign_currency)?;
+				info.total_pool_amount.ensure_add_assign(pool_amount)?;
+
+				Ok::<_, DispatchError>(())
+			})?;
```
mustermeiszer, line 761: couldn't we make the `foreign_currency` part of the key? Easily allow to invest with multiple currencies in paralel while ommitting the check and storage of the `foreign_currency` alltogether.

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +380,673 @@ pub mod pallet {
...
+				info.total_pool_amount.ensure_sub_assign(pool_amount)?;
+
+				Ok::<_, DispatchError>(())
+			})?;
+
+			T::Investment::update_investment(
+				who,
+				investment_id,
+				T::Investment::investment(who, investment_id)?.ensure_sub(pool_amount)?,
+			)?;
```
mustermeiszer, line 801: That looks odd. Why would you direclty call `update_investment`? Where does the knowledge come from what has already been swapped and fulfilled? 

I the above `increase_` logic, the `total_pool_amount` is just increased, and here it is just decreased.

mustermeiszer: Shouldn't this logic come after the `Self::apply_swap_and_notify` part and use the returned `status`?

mustermeiszer: I think this will not work, as this assumes at all times that `pool_amount` before mutate has already been locked in the investments pallet. 

## Example
* Investor increases by `100 FRAX`, assume 1:1 conversion for simplicity
* FI creates swap for `100 FRAX > USDC`
* Order gets fulfilled with `50 USDC`
* `FI` updates the order of the investor to `50 USDC` in the `Investments`
* Investors decreases by `100 FRAX`

Now here is the problem IMO. This code will now try to reduce the order by `100 USDC` while it can at most decrease by `50 USDC`. Furthermore, if we are using market rates, then this becomes even harder, as even if all has been swapped, the rate is "better" now, that you want to decrease the order by too much.

lemunozm: You're right. It should be done after `apply_swap_and_notify()` reading the status.

I understood this in a way that `decrease_foreign_investment()` only decreases for amounts already swapped. In such a case, the current investment already has the amount invested and swapped. But it's true that we're skipping the cases where I want to decrease a previous increase that was not swapped.

pallets/foreign-investments/src/lib.rs
lemunozm: Not trivial, the `StatusNotificationHook` from collected investment and redemption does not identify with a foreign currency to know how to proceed.

lemunozm: I would start this as a new feature for later

pallets/foreign-investments/src/lib.rs
wischli: Increasing and decreasing a foreign investment is asynchronous:

1. User signals increase
2. Apply swap
3. ASYNC: When the swap order (partially) fulfils, the swapped ...
  a. If increased swap: `pool_amount` can be invested
  b. If decreased swap: `foreign_amount` can be transferred back

lemunozm: But we need to reduce the investment before swapping if we are decrementing. So, in case of decrement, we can not apply swap directly.

wischli: > But we need to reduce the investment before swapping if we are decrementing. So, in case of decrement, we can not apply swap directly.

That depends on the current pending swap:
* If the swap is `pool -> foreign` with `swap_amount >= decrease_amount`, we don't need to decrease the investment
* If the swap is `pool -> foreign` with `swap_amount < decrease_amount`, we need to reduce the investment by `decrease_amount - swap_amount`
* Else we need to decrease the investment by `decrease_amount`

lemunozm: Here is the solution: [eca018c](https://github.com/centrifuge/centrifuge-chain/pull/1698/commits/eca018cb1db5bd038e5ac9ff3e8189d678a7bb3d)

pallets/foreign-investments/src/lib.rs
```diff
@@ -422,35 +380,673 @@ pub mod pallet {
...
+									who.clone(),
+									swap_id,
+									amount_to_swap,
+									T::BalanceRatio::one(),
+								)?;
+
+								Ok(SwapStatus {
+									swapped: new_swap.amount_in,
+									pending: T::Balance::zero(),
+									swapped_inverse: new_swap.amount_in,
```
wischli, line 526: Continuing with the discussion from #1692 (i.e. this [comment](https://github.com/centrifuge/centrifuge-chain/pull/1692/files/32d2297f731da83ff4d66e8c3e1ba60cd7ee6728#r1461601534)): The `pending` amount is definitely correct. However, I still believe we cannot set `swapped_inverse` amount to `new_swap.amount_in` because it is denoted in the opposite currency of `new_swap`, not the denomination of `inverse_swap`. 

So this should be `inverse_swap.amount_in.ensure_sub(amount_to_swap)` which equals `new_swap_amount_out`. This aligns with the `Ordering::Greater` implementation in which we set `swapped: inverse_swap_amount_out` (L563).

lemunozm: You're right. I was paying attention to the wrong variable 🤦🏻‍♂️. Thanks for pushing this even though I disagree.

pallets/foreign-investments/src/lib.rs
```diff
@@ -242,168 +313,56 @@ pub mod pallet {
...
-		T::CurrencyId,
-		ResultQuery<Error<T>::RedemptionPayoutCurrencyNotFound>,
-	>;
+	pub(super) type ForeignIdToSwapId<T: Config> =
+		StorageMap<_, Blake2_128Concat, ForeignId<T>, T::SwapId>;
 
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		ForeignInvestmentUpdated {
```
lemunozm, line 362: @wischli I'm not sure if these events apply now. I think not because `pallet-order-book` and `pallet-investments` will already emit their corresponding events when the "state" changes.

wischli: Yes, you can certainly can drop all events which are already handled by the Orderbook and Investments! AFAIK, none of the events are used right now. I would add events for mutations of the `ForeignInvestmentInfo` and `ForeignRedemptionInfo`.

pallets/foreign-investments/src/lib.rs
wischli: That LGTM!

lemunozm:
@mustermeiszer @wischli, to keep you updated with the PR changes. Finally, the inverse swap notifications are not needed. Actually, we want to avoid them. The inverse swaps are canceled swaps that must not generate events or notify nothing. 

Commit: [c8be732](https://github.com/centrifuge/centrifuge-chain/pull/1698/commits/c8be732a478a719ac71880bbf7f8160de438d576)

lemunozm:
I've split the pallet into some files to understand the dependencies among different methods better:
- `swaps.rs`: It only focuses on making swaps. Doesn't know about any `Info` type or info Storage. It only reads the storage related to the swap ids. Actually, this file could be moved to its own pallet with the storage Ids if we need swapping for other purposes in the future.
- `entities.rs`: Defines the `InvestmentInfo` and `RedemptionInfo` with some methods that mutate their fields. Only these methods can mutate the internal states of the `Info` types. Higher layers should only read the content and never mutate `Info` fields directly.
- `impls.rs`: The interface to handle FI, through the pallet or through the Hooks. The higher logic level.

wischli (approved):
Nothing to complain, not even a single nit. Looks really good and clean. Thanks so much for all your great work!

wischli (approved):

lemunozm:
Thanks again @wischli for all your work, help, support, and eyes here, and for taking care of all border cases in the previous implementation! 🙌🏻

lemunozm:
Let's keeping the previous branch by now 🙏🏻

PR #1702, opened 2024-01-26
Fix: transfer filters

mustermeiszer:
# Description
The filter previously just filtered out direct balance calls. As you can generate dispatching a call without going through `pre_dispatch` again easily, this new filter actually deconstructs the call itself and checks the inner ways.

* Better fix would be https://github.com/paritytech/polkadot-sdk/issues/2563

## Changes and Descriptions
* Extended `PreBalanceTransferExtension`
* Integration tests for all runtimes

# Checklist:


runtime/integration-tests/src/generic/cases/restricted_transfers.rs
```diff
@@ -0,0 +1,362 @@
...
+				assert_eq!(after_transfer_bob, pre_transfer_bob);
+				assert_eq!(after_transfer_charlie, pre_transfer_charlie);
+			});
+		}
+	}
+
+	fn validate_ok<T: Runtime>(who: Keyring, call: impl Into<T::RuntimeCallExt> + Clone) {
+		// With FilterCurrency::Specific(CurrencyId::Native)
+		{
+			let mut env = setup::<T>(FilterCurrency::All);
```
wischli, line 142: Comment and state mismatch

wischli (dismissed):
LGTM! 🚀 A few style/naming nits which can be ignored.

wischli (approved):
Thanks for adhering to my suggestions!

PR #1703, opened 2024-01-26
feat: add NAV runtime API

wischli:
# Description

* Exposes new NAV as result of #1633 calculation via runtime API :
```rust
pub struct PoolNav<Balance> {
	pub nav_aum: Balance,
	pub nav_fees: Balance,
	pub reserve: Balance,
	pub total: Balance,
}

fn nav(pool_id: PoolId) -> Option<PoolNav<Balance>>;
```

# Checklist:


runtime/common/src/apis/pools.rs
```diff
@@ -46,5 +46,7 @@ decl_runtime_apis! {
 		fn tranche_id(pool_id: PoolId, tranche_index: TrancheIndex) -> Option<TrancheId>;
 
 		fn tranche_currency(pool_id: PoolId, tranche_loc: TrancheLoc<TrancheId>) -> Option<Currency>;
+
+		fn nav(pool_id: PoolId) -> Option<(Balance, Nav<Balance>)>;
```
wischli, line 50: @mustermeiszer Not sure if we would rather want to expose all data in a new type to be more explicit. Here we return `(nav.total(pool.reserve), nav)`.

mustermeiszer (commented):
Looks good to me. @hieronx would the indexer be able to work with that?

runtime/common/src/apis/pools.rs
hieronx: Would it be possible to use a structured type for this? Just because it's quite opaque now what the data returned here is (hard to know that the first element is the reserve and the second element is the NAV)

wischli: Not at all. I wanted to propose the most simple version and then iterate on that based on the feedback.

runtime/centrifuge/src/lib.rs
```diff
@@ -2220,6 +2220,16 @@ impl_runtime_apis! {
 			let pool = pallet_pool_system::Pool::<Runtime>::get(pool_id)?;
 			pool.tranches.tranche_currency(tranche_loc).map(Into::into)
 		}
+
+		fn nav(pool_id: PoolId) -> Option<(Balance, pallet_pool_system::Nav<Balance>)> {
+			let pool = pallet_pool_system::Pool::<Runtime>::get(pool_id)?;
+			let nav_loans = Loans::update_nav(pool_id).ok()?;
```
lemunozm, line 2226: Just to be sure about this. Do we want the NAV value used to compute the loan's portfolio, or is the NAV value just for the moment we call the RuntimeAPI?

wischli: I was not sure and asked @mustermeiszer [who said that we want the current NAV](https://kflabs.slack.com/archives/C04H9H7EVEE/p1706279505349109?thread_ts=1706028509.406749&cid=C04H9H7EVEE), even if the onchain one is outdated.

lemunozm: Thanks to share the double check!

mustermeiszer: Definitely current one, as the UI needs to have the latest data!

mustermeiszer (approved):
LGTM!

PR #1706, opened 2024-01-30
refactor: assign pool fee id during propose

wischli:
# Description

 Adding a pool fee to an existing pool is a two step process of proposing and applying (after success of the `ChangeGuard`). Right now, the pool fee id is assigned as part of the second step such that it is unknown in between proposing and applying. This PR fixes this by assigning an id already during the proposing step. The id is exposed via the `Proposed` event dispatch.

Another benefit from this refactor is that two same `PoolFeeInfo`s are now encoded as a unique `Change`. Beforehand, they would have been the same which is an uncaught error. 

# Checklist:


mustermeiszer (approved):
Looks good!

lemunozm (approved):
LGTM!

PR #1713, opened 2024-02-01
Feat: `Pallet-Token-Mux`

### LINKED ISSUE #1714: `Pallet-Token-Mux`
### Description
A pallet that is able to lock variants of the same foreign asset and mints the respective local representation of these variants. The reverse process would be burn local representation and release variant token.

Variants and local representations must be recognizable by the registry metadata.

### Research/based on
* Needed for Anemoy and BT

* Reduce Swapping overhead

* Migration for `AssetMetadata` needed

mustermeiszer:
# Description
Adds a new pallet that is able to lock variants of the same foreign asset and mints the respective local representation of these variants. The reverse process would be burn local representation and release variant token. 

This is needed for increasing utility and decreasing swap burdens during investments. 

Closes #1714
## Changes and Descriptions
* Extend `trait TokenSwaps` with `fulfill_order()`-function
* Cut out `fn get_order_details()` into its own trait to allow different details to be provided
* `Pallet-Token-Mux` that provides a public api for
  * `fn deposit()` - lock and mint
  * `fn burn()` - burn and release
  * `fn match_swap` - match swap orders with newly minted local representations or by releasing variants

### Centrifuge
* Register local currency id
* Update custom metadata of all assets to include `local_representation`
  * Re-enables decoding of asset metadata
* Set local USDC as representation of all 6 registered USDC variants: Polkadot, Axelar, Ethereum, Base, Arbitrum, Celo
* Add bidirectional trading pairs for all USDC variants to/from local USDC

### Altair
* Update metadata
* Fixed: Migration of deprecated `CurrencyId::AUSD` 
 
## Tasks
- [x] Pallet functionality
- [x] Pallet UT
- [x] Pallet benchmarking
- [x] Migration `AssetMetadata` for all runtimes
  - [x] Register `LocalAsset(USDC)` on Centrifuge
  - [x] Set `LocalAsset(USDC)` for all `USDC` variants on Centrifuge
        NOTE: For dev and demo we can set those with sudo
- [x] Integrate Pallet into runtimes 

# Checklist:


wischli (commented):
It would not be a review from me, if there weren't nits. Overall, the current scope looks good to me! 

For benchmarking, let's please use v2: https://paritytech.github.io/polkadot-sdk/master/frame_benchmarking/trait.Benchmark.html?search=benchmarking.

 But we might want to do that in a separate PR. Happy to help with UTs, ITs or benches!

libs/types/src/tokens.rs
```diff
@@ -335,6 +363,62 @@ impl From<CurrencyId> for FilterCurrency {
 	}
 }
 
+pub mod altair {
```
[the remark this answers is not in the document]
mustermeiszer: See removed `AUSD` variant, for which the migration on Altair did not yet run.

libs/types/src/orders.rs
```diff
@@ -126,3 +126,15 @@ impl<Balance: Zero + Copy, OrderId: Copy + Ord + PartialOrd> Order<Balance, Orde
...
 }
+
+/// Swap details that the multiplexing pallet needs in order to match orders
+pub struct MuxSwap<Currency, Balance, Ratio> {
+	/// The incoming currency, i.e. the desired one.
+	pub currency_in: Currency,
+	/// The outgoing currency, i.e. the one which should be replaced.
+	pub currency_out: Currency,
+	/// The amount of incoming currency which shall be bought.
+	pub amount_in: Balance,
```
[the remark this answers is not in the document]
mustermeiszer: Already done.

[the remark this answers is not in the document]
wischli: Sorry, thought this was automatically discarded as I didn't see it in the Github review UI anymore after refreshing to your latest changes.

lemunozm:
> It would not be a review from me, if there weren't nits

@wischli your NITs are the best! 🙌🏻

pallets/token-mux/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
+
+	#[pallet::pallet]
+	#[pallet::storage_version(STORAGE_VERSION)]
+	pub struct Pallet<T>(_);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config
+	where
+		CurrencyFor<Self>: From<LocalAssetId> + TryInto<LocalAssetId>,
```
lemunozm, line 73: Currently, all related with currencies are a bit mixed in the codebase, but theoretically, `LocalAssetId` is a runtime type. If we're abstracting the `CurrencyId,` we should also abstract `LocalAssetId`. Maybe makes sense to have this type under another associated type in `Config`:

```rust
trait Config {
    type LocalAssetId;
    type CurrencyId: From<Self::LocalAssetId> + TryInto<Self::LocalAssetId>
}
```

pallets/token-mux/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+			let meta_local =
+				T::AssetRegistry::metadata(&local).ok_or(Error::<T>::MetadataNotFound)?;
+
+			// NOTE: We could also think about making conversion between local
+			//       representations and variants but I fear that we then have problems with
+			//       SUM(locked variants) = local. Hence, this restriction.
+			ensure!(
+				meta_local.decimals == meta_variant.decimals,
+				Error::<T>::DecimalMismatch
+			);
```
lemunozm, line 323: I feel like this restriction is hard. I would expect to have different decimal representation of currencies that has a 1:1 relation. But sure @mustermeiszer can say more here

pallets/token-mux/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+		pub fn deposit(
+			origin: OriginFor<T>,
+			currency_out: CurrencyFor<T>,
+			amount_out: BalanceFor<T>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			let local = Self::try_local(&currency_out)?;
+
+			Self::mint_route(&who, local.clone(), currency_out.clone(), amount_out)?;
```
lemunozm, line 158: Is `who` account coming from `liquidity-pools`? If so, I think nobody could call this.

lemunozm (commented):
Amazing work of both of you!

I left some questions and suggestions

pallets/token-mux/src/lib.rs
wischli: This call is permissionless, i.e. open to any signed origin, such that a bot can match those local swaps.

pallets/token-mux/src/lib.rs
wischli: The purpose of this pallet is to aggregate variants of the same currency into a CFG-local representation. So IMO it's feasible to assume all these variants as well as the local variant to have the exact same number of decimals.

pallets/token-mux/src/mock.rs
```diff
@@ -0,0 +1,189 @@
...
+		OrmlTokens: orml_tokens,
+	}
+);
+
+cfg_test_utils::mocks::orml_asset_registry::impl_mock_registry! {
+	MockRegistry,
+	CurrencyId,
+	Balance,
+	CustomMetadata
+}
```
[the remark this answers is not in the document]
wischli: I agree! I was looking for it yesterday without success 😄

pallets/order-book/src/lib.rs
```diff
@@ -762,10 +765,37 @@ pub mod pallet {
...
+		fn get_order_details(order: Self::OrderId) -> Option<MuxSwap<T::CurrencyId, T::Ratio>> {
+			Orders::<T>::get(order)
+				.map(|order| MuxSwap {
+					ratio: order.ratio,
+					currency_in: order.currency_in,
+					currency_out: order.currency_out,
+				})
+				.ok()
+		}
+	}
```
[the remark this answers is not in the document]
mustermeiszer: Was to lazy to change all locations where the normal infos are fetched ^^

pallets/token-mux/src/lib.rs
mustermeiszer: Definitely need to keep that for now. Otherwise, we possibly can get mismatches of amounts.

pallets/token-mux/src/mock.rs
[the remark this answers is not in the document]
lemunozm: I can do it and merge it in another PR fastly if you want

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Then I think could be helpful to have them under the same `Swap` (also for future purposes)

pallets/token-mux/src/lib.rs
wischli: The LP and FI flow is untouched by this pallet. It solely provides means of on- and off-ramping local stable coin variants by exposing the `match_swap(origin, order_id, amount)` extrinsic. This allows any signed extrinsic to fulfil any orderbook swap into/from `LocalAssetId`. The caller origin and the recipient of the swap are independent of each other.

Let's look at the current LP flow: An investor locks ETH USDC on Ethereum to invest into a DOT USDC pool (= `ForeignAsset(6)` on Centrifuge chain). When we receive the inbound message, we mint `LpEthUSDC` as foreign asset and create a swap for `LpEthUSDC -> ForeignAsset(6)`. Both currencies are publicly known and in order to fulfill the swap, someone needs to trade `ForeignAsset(6)` for `LpWrappedEthUsdc`.

The new flow will be adjusted slightly:
* USDC pools will have `CurrencyId::LocalAsset(LocalAssetId(1))` (:= `Local USDC`) as pool currency instead of DOT USDC
* The orderbook swap of `LpEthUSDC -> Local USDC` can be fulfilled by anyone who holds sufficient CFG to pay the transaction fee because this pallet transfers the `LpEthUSDC` amount to sovereign account of the token-mux pallet and mints the same amount as `Local USDC`
* For offramping (`Local USDC -> LpEthUSDC`), the corresponding amount needs to be in the sovereign account. Then, again anyone can fulfill the swap by transferring from sovereign to the swap order destination and burning the `Local USDC`

pallets/token-mux/src/mock.rs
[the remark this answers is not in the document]
wischli: Doesn't need to be fast. I have the utility set up already. Thanks for the offer though!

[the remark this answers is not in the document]
lemunozm: Added here: https://github.com/centrifuge/centrifuge-chain/pull/1721 

I realized that the current mock for the asset registry (`impl_mock_registry!` macro) works pretty well in most scenarios. It's probably not necessary to mock it with the mock builder.

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
wischli: @lemunozm The issue I see with merging `MuxSwap` and `Swap` by adding the `ratio` field to `Swap` is that in foreign investments, we often construct a `Swap` without any knowledge or need for a ratio. 

IMO, it's better to keep both separate. Another solution would be making the `ratio` optional, i.e. `ratio: Option<OrderRatio<Ratio>>`. WDYT?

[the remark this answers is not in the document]
wischli: Originally, I wanted to impl `From<Order<T>>` for `Swap` and `MuxSwap` such that we can keep a single `OrderDetails` impl in pallet-order-book:
```
impl<T: Config, S> OrderDetails<S> for Pallet<T> where S: From<Order<T>> { ... }
```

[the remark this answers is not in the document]
lemunozm: > The issue I see with merging MuxSwap and Swap by adding the ratio field to Swap is that in foreign investments, we often construct a Swap without any knowledge or need for a ratio.

Mmm... are we having the same type to represent 2 different kinds of things?
1. Specifying the intent of making a swap.
2. See the order details

Maybe each thing deserves its own type. Right now, they represent the same amount of information, but it seems that they can evolve differently.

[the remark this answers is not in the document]
lemunozm: I see this as:
```rust
/// Specifies an intent of making a swap
Swap {
    currency_in,
    currency_out,
    amount_out,
}

/// Information about an order
OrderInfo { 
    swap: Swap,
    ratio: OrderRatio,
    // Future attributes order related, i.e:
    // initial_amount, swapped_amount, etc
} 
```

pallets/token-mux/src/lib.rs
wischli: Since the `cfg_types::tokens::LocalAssetId` is baked into the `CustomMetadata`, we still need to restrict the associated `LocalAssetId` Config type to impl `From<cfg_types::tokens::LocalAssetId>`. Not sure if this makes sense to be refactored then.

lemunozm: Mmm... I see 🤔. Our `CustomMetadata` forces us to bring a lot of types from the runtime into testing. Maybe we should abstract the `CustomeMetadata` somehow. But sure, it's not work for this PR.

Then, as you said, it does not make sense to refactor it. Use `LocalAssetId` directly.

pallets/token-mux/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+		/// The local asset type
+		type LocalAssetId: From<cfg_types::tokens::LocalAssetId>;
+
+		/// The type for handling transfers, burning and minting of
+		/// multi-assets.
+		type Tokens: fungibles::Inspect<Self::AccountId, AssetId = Self::CurrencyId>
+			+ Mutate<Self::AccountId, AssetId = Self::CurrencyId>;
+
+		/// The type for retrieving and fulfilling swap orders.
+		type Swaps: TokenSwaps<
```
lemunozm, line 130: Super NIT, I would call this `OrderBook` (actually, I would call `TokenSwaps` as `OrderBook`, but that for another PR). This does not perform "swaps", it performs actions over an order book. The difference is now more clear with `pallet-swaps`, which does use swaps.

pallets/token-mux/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// The given currency has no metadata set.
+		MetadataNotFound,
+		/// The given currency has no local representation and can hence not be
+		/// deposited to receive a local representation.
+		NoLocalRepresentation,
+		/// The given currency is not a local currency
+		NotLocalCurrency,
```
lemunozm, line 172: I think this is not use

pallets/token-mux/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+		/// The provided local currency does not match the local representation
+		/// of the currency to be unlocked
+		LocalCurrencyMismatch,
```
lemunozm, line 175: I think this is not use

libs/traits/src/lib.rs
```diff
@@ -543,16 +557,21 @@ pub trait TokenSwaps<Account> {
...
 	) -> Result<Self::BalanceIn, DispatchError>;
+
+	#[cfg(feature = "runtime-benchmarks")]
+	/// Adds a valid trading pair.
+	fn add_trading_pair(
+		currency_in: Self::CurrencyId,
+		currency_out: Self::CurrencyId,
+		min_order: Self::BalanceOut,
+	) -> DispatchResult;
 }
```
lemunozm, line 575: 👌🏻

pallets/token-mux/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+			AssetId = Self::CurrencyId,
+		>;
+
+		/// Balance type for incoming values
+		type BalanceIn: Member
+			+ Parameter
+			+ FixedPointOperand
+			+ AtLeast32BitUnsigned
+			+ EnsureMul
+			+ EnsureDiv
```
lemunozm, line 86: NIT. IIRC, the `Ensure*` bounds already are in `AlLeast32BitUnsigned` (several ocurrences)

pallets/token-mux/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+			+ MaxEncodedLen
+			+ Into<BalanceOf<Self>>
+			+ From<BalanceOf<Self>>;
+
+		/// Type for price ratio for cost of incoming currency relative to
+		/// outgoing
+		type BalanceRatio: Parameter
+			+ Member
+			+ sp_runtime::FixedPointNumber
+			+ MaybeSerializeDeserialize
```
lemunozm, line 107: `MaybeSerializeDeserialize` I think is never used (several occurrences)

pallets/token-mux/src/mock.rs
```diff
@@ -0,0 +1,241 @@
...
+use sp_io::TestExternalities;
+use sp_runtime::{
+	testing::Header,
+	traits::{BlakeTwo256, IdentityLookup},
+	FixedU128,
+};
+
+use crate::{
+	pallet as pallet_token_mux,
+	tests::{
```
lemunozm, line 34: NIT. Right now, `mock` uses the `tests` module, and the `tests` module uses `mock`. If the mock uses these constants to initialize the testing state, could it then own them?

pallets/token-mux/src/tests.rs
```diff
@@ -0,0 +1,531 @@
+// Copyright 2023 Centrifuge Foundation (centrifuge.io).
```
lemunozm, line 1: Very well-organized tests! It was really easy to follow them! 🚀

lemunozm (commented):
Good work! Organized and easy to follow!

Just some NITs

pallets/token-mux/src/mock.rs
wischli: I agree. I first had them in `mock` and then wanted to follow your example. But its cleaner to keep them in mock. In general, I prefer having them there.

lemunozm: Yeah, I always have the dichotomy of where to put them... depends a lot if the mock uses them directly or not.

Cargo.toml
```diff
@@ -297,6 +297,7 @@ cfg-types = { path = "libs/types", default-features = false }
 cfg-utils = { path = "libs/utils", default-features = false }
 cfg-mocks = { path = "libs/mocks", default-features = false }
 cfg-test-utils = { path = "libs/test-utils", default-features = false }
+proofs = { path = "libs/proofs", default-features = false }
```
lemunozm, line 300: Out of curiosity, where are you using this? I'm not able to see it

wischli: Not sure how it got into there, removed it.

pallets/token-mux/src/mock.rs
wischli: Moved all constants back to the mock because that uses all of them

runtime/common/src/migrations/local_currency.rs
```diff
@@ -0,0 +1,440 @@
...
+	const LOG_PREFIX: &str = "AddBidirectionalTradingPair";
+
+	pub struct Migration<T, CurrencyOne, CurrencyTwo, MinOrderAmount>(
+		sp_std::marker::PhantomData<(T, CurrencyOne, CurrencyTwo, MinOrderAmount)>,
+	);
+
+	impl<T, CurrencyOne, CurrencyTwo, MinOrderAmount> OnRuntimeUpgrade
+		for Migration<T, CurrencyOne, CurrencyTwo, MinOrderAmount>
+	where
+		T: pallet_order_book::Config<
```
lemunozm, line 365: If IIRC, `order-book` can be reset. So maybe this isn't needed. If can not be reset, then we should make another big and painful migration with the previous order-book refactor

lemunozm: In `centrifuge`, only one order exists (with `buy_amount` 0, so no balance is held). In `altair`,  `order-book` does not exist.

runtime/altair/src/migrations.rs
```diff
@@ -268,29 +481,21 @@ mod orml_tokens_migration {
...
-							"Failed to call burn_from({:?}, {:?}, {balance}): {:?}",
-							DEPRECATED_AUSD_CURRENCY_ID,
-							account,
-							e
-						)
-					})
-					.ok();
+
+					// Remove account data and reduce total issuance manually at the end
+					Accounts::<T>::remove(account.clone(), DEPRECATED_AUSD_CURRENCY_ID);
```
lemunozm, line 495: Better than burning 👍🏻

lemunozm (commented):
Thanks for your awesome work with the migrations!

pallets/token-mux/src/lib.rs
lemunozm: @wischli not sure if you read this or it has become too many collapsed 😆

node/src/chain_spec.rs
```diff
@@ -1007,6 +1007,24 @@ fn asset_registry_assets() -> Vec<(CurrencyId, Vec<u8>)> {
 			}
 			.encode(),
 		),
+		(
+			CURRENCY_ID_LOCAL,
+			AssetMetadata::<Balance, CustomMetadata> {
+				decimals: 6,
+				name: b"Local USDC".to_vec(),
+				symbol: b"localUSDC".to_vec(),
```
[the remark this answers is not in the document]
wischli: Could be mistaken for `Liquid` prefix which is common in DeFi, ex. LDOT, LSKM.

runtime/common/src/migrations/local_currency.rs
```diff
@@ -0,0 +1,440 @@
...
+	{
+		fn on_runtime_upgrade() -> Weight {
+			<orml_asset_registry::Pallet<T> as Mutate>::register_asset(
+				Some(LocalCurrency::get()),
+				AssetMetadata {
+					decimals: 6,
+					// TODO: Ask others
+					name: "Local USDC".as_bytes().to_vec(),
+					symbol: "localUSDC".as_bytes().to_vec(),
+					existential_deposit: Zero::zero(),
```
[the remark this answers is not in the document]
wischli: Good point, should not be zero. Will take `1000` similar to other USDC variants.

runtime/common/src/migrations/local_currency.rs
wischli: I added this migration as convenience to immediately be able to trade USDC variants with our local one. Normally, this should be done through Governance but that can only happen after RU 1025 because the `CurrencyId::LocalAsset` variant does not exist yet. Thus, this migration.

runtime/altair/src/migrations.rs
wischli: Here, this approach is a bit more lightweight. Moreover, burning was not possible without many "hacks" because the runtime does not know the deprecated currency variant anymore. For storages, we can mitigate this by declaring a storage alias. However, abstraction layers such as minting or burning are not affected by this.

pallets/token-mux/src/lib.rs
wischli: Thanks for the reminder. I agree and will change it!

node/src/chain_spec.rs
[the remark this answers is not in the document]
lemunozm: Mmm... that's true 🤔

runtime/common/src/migrations/local_currency.rs
lemunozm: Ah, then it makes a lot of sense, thanks!

runtime/altair/src/migrations.rs
lemunozm: And we also have free resources for an account that will never be used, so great!

runtime/centrifuge/src/migrations.rs
```diff
@@ -10,9 +10,29 @@
...
+	pub const LocalAssetIdUsdc: LocalAssetId = LOCAL_ASSET_ID;
+	pub const LocalCurrencyIdUsdc: CurrencyId = CURRENCY_ID_LOCAL;
+	pub const PoolIdAnemoy: PoolId = 4_139_607_887;
+	pub const PoolCurrencyAnemoy: CurrencyId = CURRENCY_ID_DOT_NATIVE;
+	pub const UsdcDot: CurrencyId = CURRENCY_ID_DOT_NATIVE;
+	pub const UsdcEth: CurrencyId = CURRENCY_ID_LP_ETH;
+	pub const UsdcBase: CurrencyId = CURRENCY_ID_LP_BASE;
+	pub const UsdcArb: CurrencyId = CURRENCY_ID_LP_ARB;
+	pub const UsdcCelo: CurrencyId = CURRENCY_ID_LP_CELO;
+	pub const MinOrderAmount: Balance = 10 * CFG;
```
lemunozm, line 35: `MinOrderAmount` is measured in `amount_out` used in each `TraidingPair`. In this case, we're adding a lot of decimals from `CFG` and `USDC`, and we're only using 6 decimals, so we can never place a swap.

runtime/common/src/migrations/local_currency.rs
```diff
@@ -0,0 +1,440 @@
...
+		CurrencyOne: Get<CurrencyId>,
+		CurrencyTwo: Get<CurrencyId>,
+		MinOrderAmount: Get<Balance>,
+	{
+		fn on_runtime_upgrade() -> Weight {
+			pallet_order_book::Pallet::<T>::add_trading_pair(
+				RawOrigin::Root.into(),
+				CurrencyTwo::get(),
+				CurrencyOne::get(),
+				MinOrderAmount::get(),
```
lemunozm, line 379: In this case `MinOrderAmount` is denominated in `CurrencyOne`

runtime/common/src/migrations/local_currency.rs
```diff
@@ -0,0 +1,440 @@
...
+					CurrencyOne::get(),
+					e
+				);
+			})
+			.ok();
+			pallet_order_book::Pallet::<T>::add_trading_pair(
+				RawOrigin::Root.into(),
+				CurrencyOne::get(),
+				CurrencyTwo::get(),
+				MinOrderAmount::get(),
```
lemunozm, line 394: In this case `MinOrderAmount` is denominated in `CurrencyTwo`

runtime/centrifuge/src/migrations.rs
lemunozm: @mustermeiszer do we really need a `TradingPair` min amount if we already have a fulfilled min amount? Can not the fulfilled minimum amount also be used as the min order amount? Do we need to be very precise here?

lemunozm (changes requested):

runtime/centrifuge/src/migrations.rs
wischli: Great catch @lemunozm! I intended to go with `10 * USDC`, e.g. `10_000_000`. Was also wondering about the need of the min order amount.

runtime/common/src/migrations/local_currency.rs
wischli: In this particular case, it does not matter as both currencies have the same denomination.

runtime/common/src/migrations/local_currency.rs
```diff
@@ -0,0 +1,444 @@
...
+					}
+				});
+				weight.saturating_accrue(T::DbWeight::get().writes(1));
+				log::info!("{LOG_PREFIX} Migrated pool currency");
+			} else {
+				log::info!("{LOG_PREFIX} Skipping pool currency migration");
+			}
+
+			weight
+		}
```
mustermeiszer, line 284: We need to error out here if there are outstanding invest orders. Redemptions are fine, as they can be collected in the new local representation.

mustermeiszer: CR

mustermeiszer (commented):
Looks good! Thanks, those we quite some migrations to write! 

* CR: Migration of pool
* Clarification: @hieronx does Anemoy want to migrate for sure to local?
* Feat: We should add a converter that wraps `OracleRatioProvider` or extends it so that if we trade `variant` <> `local` we return `One::one()`

lemunozm:
> Feat: We should add a converter that wraps OracleRatioProvider or extends it so that if we trade variant <> local we return One::one()

@mustermeiszer, regarding this, could make sense to dispatch an error if `get_order_details()` returns an order with `OrderRation::Market`? That way we force only orders with `OrderRatio::Custom(One::one())` can be used for `token-mux`

mustermeiszer:
But that beaks the usage of this pallet with foreign investments as this one uses market. 

I would rather use the wrapper unless you have a better idea

runtime/centrifuge/src/migrations.rs
lemunozm: IIRC, it is there to avoid creating orders with 0.0000001 amount and then opening an attack vector, wasting chain resources very cheaply. But having already a min fulfillment amount, I think this issue is automatically solved for both placing an order and filling an order.

But looking forward to what @mustermeiszer says about this

runtime/common/src/migrations/local_currency.rs
lemunozm: That's right 👍🏻

runtime/development/src/migrations.rs
```diff
@@ -10,7 +10,31 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU General Public License for more details.
 
-pub type UpgradeDevelopment1038 = (
-	runtime_common::migrations::epoch_execution::Migration<super::Runtime>,
-	runtime_common::migrations::transfer_allowlist_currency::Migration<crate::Runtime>,
+use cfg_types::tokens::{
+	usdc::{
+		CURRENCY_ID_AXELAR, CURRENCY_ID_DOT_NATIVE, CURRENCY_ID_LOCAL, CURRENCY_ID_LP_ARB,
+		CURRENCY_ID_LP_BASE, CURRENCY_ID_LP_CELO, CURRENCY_ID_LP_ETH, LOCAL_ASSET_ID,
```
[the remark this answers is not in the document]
wischli: Actually only `CURRENCY_ID_LP_ETH`. I have the fix on another branch and will cherry pick!

node/src/chain_spec.rs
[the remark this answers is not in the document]
wischli: Any opinion about the symbol/name @mustermeiszer ?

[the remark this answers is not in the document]
mustermeiszer: Local IMO

lemunozm:
I meant, inside `pallet_swaps::match_swaps()`, if the ratio returned by `get_order_details()` there is `Market`, then error out. Foreign investment should not be affected by that.

I say this because some oracle can still add a price for a variant <> local price. And it would be awkward if the price used is wired to `one()` later.

Maybe it would even be interesting to add a `T::Check` associated type to `pallet-feed` in order to evaluate if some key is correctly fed. In this case, that `Check` will be implemented as: any local currency can only have `one()` as a market ratio with its associated currency.

mustermeiszer:
Mmmh. I see your points from a technical perspective.

But using oracles would mean that we need to set and control oracles for all `variant <> localrepresentation` combinations. It is a set once, but we would need to watch them and it is a potential exposure. 

I see no use-case for a non 1:1 conversion for `localrepresenation <> variant` and hence I would argue that intercepting those oracle fetches is okay. The `pallet-feed` IMO should have no opinion on the validity of oracles. I mean, anybody can feed anything.

lemunozm:
> The pallet-feed IMO should have no opinion on the validity of oracles. I mean, anybody can feed anything.

That is also true. There is no perfect solution here. No strong opinion then 👍🏻

mustermeiszer:
@wischli what do you think? ^^

wischli:
> @wischli what do you think? ^^

Sorry for staying quiet. I prefer your proposal @mustermeiszer and am working on a implementation right now. I can imagine I won't implement it exactly as desired but we can iterate on that in a bit when I pushed it.

runtime/centrifuge/src/migrations.rs
lemunozm: You know I'm always in favor of removing things 😆. Now, seriously, I think removing this can simplify many things: usage, migrations, maintenance...

wischli: Can we do this in a subsequent PR though please?

runtime/common/src/migrations/local_currency.rs
```diff
@@ -0,0 +1,440 @@
...
+				AssetMetadata {
+					decimals: 6,
+					// TODO: Ask others
+					name: "Local USDC".as_bytes().to_vec(),
+					symbol: "localUSDC".as_bytes().to_vec(),
+					existential_deposit: Zero::zero(),
+					location: None,
+					additional: CustomMetadata {
+						transferability: CrossChainTransferability::None,
+						mintable: true,
```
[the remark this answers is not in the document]
wischli: [77290a1](https://github.com/centrifuge/centrifuge-chain/pull/1713/commits/77290a1af83f1fbd4efa0b93e9a4f82cced10a49)

runtime/development/src/migrations.rs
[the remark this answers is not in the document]
wischli: 2221f23185d43590d67d76f05c3aff53746af706

runtime/common/src/migrations/local_currency.rs
wischli: [70587ae](https://github.com/centrifuge/centrifuge-chain/pull/1713/commits/70587ae68ba0fa741c6dcaac6c6adb6df5e34230)

wischli:
@mustermeiszer @lemunozm Please check out my proposal for the extension: [e8b617d](https://github.com/centrifuge/centrifuge-chain/pull/1713/commits/e8b617d96ad1236730c7750f4e4c99bd3299c6bb)

Looking forward to your change requests 😄

runtime/centrifuge/src/migrations.rs
wischli: If that is fine, I changed the amount to 10 USDC here: [0be2e66](https://github.com/centrifuge/centrifuge-chain/pull/1713/commits/0be2e66118a52153e5a34e9fae5d969d0adbc236)

lemunozm: Make sense!

I can own that in the new PR if we want to go with it.

mustermeiszer (commented):
My initial plan was to make this a pure runtime configuration and wrap `OracleRatioProvider` and not adapt the `pallet-order-bool`, but I am happy with this change. Nothing to complain from my side.

lemunozm:
Agree with @mustermeiszer, I think it can be done by wrapping the current `OracleRatioProvider` and just modifying the returned `ratio` with `one()` in case of local currencies. That way, `order-book` is totally agnostic from this.

Even instead of wrapping, the current OracleRationProvided can be modified with this "filter"

wischli:
> Agree with @mustermeiszer, I think it can be done by wrapping the current `OracleRatioProvider` and just modifying the returned `ratio` with `one()` in case of local currencies. That way, `order-book` is totally agnostic from this.
> 
> 
> 
> Even instead of wrapping, the current `OracleRationProvided` can be modified with this "filter"

Could you please elaborate? IIUC, I we want to wrap the `ValueProvider::get(feeder, key) ` call here. I didnt go with your proposed way because the pre check for local coupling must not require a feeder. So the solution would be a get call with a callback to retrieve the feeder if the pre check is false. IMO, thats not more decoupled from the orderbook than the current code and uglier. I am sure to be missing something here.

lemunozm:
Yes, forget about my proposal about `Check`. What I says is what Frederik says, which is modifying the current Provider to provide a ratio with `one()`. This can be done by wrapping the current one or by mutating the current one.

Mutating the current one, I think all of this is a 5 line change or so [here](https://github.com/centrifuge/centrifuge-chain/blob/main/runtime/common/src/oracle.rs#L148) we need to perform the check you've done [here](https://github.com/centrifuge/centrifuge-chain/pull/1713/commits/e8b617d96ad1236730c7750f4e4c99bd3299c6bb#diff-52a45a281a0f1b38a94309fd57b8004e802c6af340f03989276345672c0576e1R182). 

So `get_try_local()` will be just `ValueProvider::get()` without the need for another trait.

lemunozm:
The wrapping version is implementing a new `ValueProvider`, which calls `OracleRatioProvider` under the hood. Later, `order-book` is configured with the new `ValueProvider` types who wires to `one()` for local currencies.

wischli:
> Mutating the current one, I think all of this is a 5 line change or so, just here we need to perform the check you've done here.

But then we still need to check for the market feeder existence beforehand which does not necessarily have to exist, does it?

lemunozm:
Ok, I see now your point.

So, the problem is that to obtain a market price of `one()` for a local currency, `order-book` needs a feeder configured to be able to call `ValueProvider::get()` right? But ideally you could obtain that value without adding a feeder.

I would say that if we are asking for a market price, `order-book` should be configured with a feeder. Even if it makes no sense because it is always `one()`, it should be a precondition to call `convert_by_market()`

lemunozm:
The reasoning is: 

If the prices are handled by `OracleRatioProvider`, then from `order-book` perspective those values are fed somehow by an oracle. Even though we're wiring those to `one()`. Consequently, from `order-book` to be able to use the provider, it needs first to configure a feeder.

wischli:
@lemunozm @mustermeiszer I hope this refactor (cd0da2b80ddb98cdd03885ef6c9d5e751eaaad66) now fulfils your desires. Sorry about the bumpy road, your vison was a bit cryptic to me.

mustermeiszer:
Nothing I could wish for left! Sorry for the complaints, I TBH forgot about the market feeder needed to be set. ^^ 

Should we also add a migration for that one or can we use council for setting it?

lemunozm:
Super clean @wischli!

wischli:
> Nothing I could wish for left! Sorry for the complaints, I TBH forgot about the market feeder needed to be set. ^^
> 
> Should we also add a migration for that one or can we use council for setting it?

Neither the new oracle pallets nor the FI refactor are not live on Altair or Centrifuge, so no need for a migration.

libs/traits/src/lib.rs
```diff
@@ -539,6 +539,23 @@ impl<Amount, Currency: PartialEq> Swap<Amount, Currency> {
...
+}
+
+// impl<Balance, Currency, Ratio> From<OrderInfo<Balance, Currency, Ratio>>
+// 	for Swap<Balance, Currency>
+// {
+// 	fn from(&self) -> Swap<Balance, Currency> {
+// 		self.swap
+// 	}
+// }
+
```
lemunozm, line 558: NIT: Could be removed if not needed?

runtime/altair/src/lib.rs
```diff
@@ -1716,6 +1722,26 @@ impl pallet_swaps::Config for Runtime {
...
+	type BalanceOut = Balance;
+	type BalanceRatio = Ratio;
+	type CurrencyId = CurrencyId;
+	type LocalAssetId = LocalAssetId;
+	type OrderBook = OrderBook;
+	type OrderId = OrderId;
+	type PalletId = TokenMuxPalletId;
+	type RuntimeEvent = RuntimeEvent;
+	type Tokens = OrmlTokens;
+	// TODO(william): Change to weights once they exist
```
lemunozm, line 1741: Can be in another PR, but we should not forget this!

lemunozm (dismissed):
Amazing work William! LGTM! 💯 

Just a minor comment. Nothing blocking

libs/traits/src/lib.rs
wischli: Good catch! I thought I had removed it but couldn't find it anymore.

runtime/altair/src/lib.rs
wischli: Yes, will fix this as part of the RC1

wischli (approved):

PR #1717, opened 2024-02-01
fix: integration tests for OrderBook market prices

wischli:
# Description

* Fixes integration tests for #1674
* Adds shortcut to orderbook for same currencies to not require market feed
* Adds integration test for edge case: `Increase -> Fulfill -> Decrease -> Fulfill partially -> Increase more than pending`

## TODO

- [x] Discuss and fix `increase_fulfill_decrease_fulfill_partial_increase` 

# Checklist:

### DIFF
--- a/pallets/order-book/src/lib.rs
+++ b/pallets/order-book/src/lib.rs
@@ -868,6 +868,10 @@ pub mod pallet {
 			currency_out: Self::CurrencyId,
 			amount_out: Self::Balance,
 		) -> Result<Self::Balance, DispatchError> {
+			if currency_in == currency_out {
+				return Ok(amount_out);
+			}
+
 			let ratio = Self::market_ratio(currency_out, currency_in)?;
 			Self::convert_with_ratio(currency_out, currency_in, ratio, amount_out)
 		}
--- a/runtime/integration-tests/src/generic/cases/liquidity_pools.rs
+++ b/runtime/integration-tests/src/generic/cases/liquidity_pools.rs
@@ -4232,8 +4232,6 @@ mod development {
 					));
 
 					dbg!(frame_system::Pallet::<T>::events());
-					dbg!(invest_amount_foreign_denominated);
-
 					assert!(frame_system::Pallet::<T>::events().iter().any(|e| {
 						e.event
 							== pallet_liquidity_pools_gateway::Event::<T>::OutboundMessageSubmitted {
@@ -4244,8 +4242,7 @@ mod development {
 								tranche_id: default_tranche_id::<T>(pool_id),
 								investor: investor.clone().into(),
 								currency: general_currency_index::<T>(foreign_currency),
-								// FIXME: Expects invest_amount_foreign_denominated / 2 but should be invest_amount_foreign_denominated 
-								currency_payout: invest_amount_foreign_denominated / 2,
+								currency_payout: invest_amount_foreign_denominated,
 								// FIXME: Expects invest_amount_foreign_denominated / 4 but should be invest_amount_foreign_denominated / 2
 								remaining_invest_amount: invest_amount_foreign_denominated / 4,
 							},
--- a/runtime/integration-tests/src/generic/cases/liquidity_pools.rs
+++ b/runtime/integration-tests/src/generic/cases/liquidity_pools.rs
@@ -4215,7 +4215,7 @@ mod development {
 					assert_ok!(pallet_order_book::Pallet::<T>::fill_order(
 						RawOrigin::Signed(trader.clone()).into(),
 						swap_order_id,
-						invest_amount_pool_denominated / 4 * 3
+						3 * invest_amount_pool_denominated / 4
 					));
 
 					// Increase more than pending swap (pool -> foreign) amount from decrease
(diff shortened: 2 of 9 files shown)


pallets/order-book/src/lib.rs
```diff
@@ -868,6 +868,10 @@ pub mod pallet {
 			currency_out: Self::CurrencyId,
 			amount_out: Self::Balance,
 		) -> Result<Self::Balance, DispatchError> {
+			if currency_in == currency_out {
+				return Ok(amount_out);
+			}
+
```
lemunozm, line 874: I fully agree with this addition. Thanks!

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -3889,48 +3865,65 @@ mod development {
...
 						pool_id,
 						invest_amount_foreign_denominated,
 						investor.clone(),
 						foreign_currency,
 					);
+					let swap_order_id = pallet_foreign_investments::Swaps::<T>::swap_id_from(
+						&investor,
+						default_investment_id::<T>(),
+						pallet_foreign_investments::Action::Investment,
+					)
```
lemunozm, line 3903: Not sure if it could make sense to expose this under a `RuntimeAPI`. How do Apps know which order placed by FI needs to be fulfilled?

@cc @mustermeiszer

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -3942,9 +3935,24 @@ mod development {
 						DEFAULT_DOMAIN_ADDRESS_MOONBEAM,
 						decrease_msg_pool_swap_amount
 					));
+					assert!(frame_system::Pallet::<T>::events().iter().any(|e| {
```
lemunozm, line 3938: Super NIT. There is a `find_event()` in `generic/utils/mod.rs`

lemunozm (commented):
Thanks for these update! I like what I see. 

I'll check wrong use case. 🙌🏻

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -4102,8 +4137,132 @@ mod development {
...
+					let swap_order_id = pallet_foreign_investments::Swaps::<T>::swap_id_from(
+						&investor,
+						default_investment_id::<T>(),
+						pallet_foreign_investments::Action::Investment,
+					)
+					.expect("Swap order exists; qed");
+					assert_ok!(pallet_order_book::Pallet::<T>::fill_order(
+						RawOrigin::Signed(trader.clone()).into(),
+						swap_order_id,
+						invest_amount_pool_denominated / 4 * 3
```
lemunozm, line 4218: I think this is wrong. The multiplication has priority. So it is: `AMOUNT / 12`

wischli: It should not, rust math is distributive such that `x * y  / z = (x * y) / z`. Safety first, I checked by doing `3 * invest_amount / 4` and it didn't change anything.

wischli: I think you are mixing it up with the order of operations for addition/subtraction and multiplication/division: `a + b + c = a + (b * c)`

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
wischli: I think it's fine to query `ForeignIdToSwapId::<T>::get((account, investment_id, action))` for Apps which would be `api.query.foreignInvestments.foreignIdToSwapId([account, investmentId, action['Investment'])`.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
wischli: I don't find this API to be more helpful here because all foreign investment integration tests operate in a single closure of `env` such that for multiple of these events it is better to directly look for the exact same one as we are asserting at least two values here, not only a single one.

But I will definitely remember this for the next integration tests.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
lemunozm: Make sense!

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
lemunozm: Yeah, in the same way `*` takes more priority over `+`, I thought it was for  `*` and `/`. Thanks for enlighten me 🙌🏻

lemunozm (approved):
Approved! Amazing work with integration-tests and debugging @wischli!

Feel free to merge it. I've also added a UT for the swap ids case: [5536efe](https://github.com/centrifuge/centrifuge-chain/pull/1674/commits/5536efe134efad923d01f116a9554468f919dca6)

Right now, it fails (as expected) and your commit should fix it.

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
wischli: Thanks for the explanation. Next time, I touch the integration tests, I will re-check whether it makes sense to switch to that API :)

PR #1719, opened 2024-02-06
Add typed balances to FI and OrderBook.

lemunozm:
# Description

Distinguish at the type level different kinds of balances used in FI and OrderBook, to force never mix them at compile time.

## Changes and Descriptions

FI:
 - `ForeignBalance`
 - `PoolBalance`
 - `TrancheBalance`
 - `SwapBalance` (that actually can be `ForeignBalance` or `PoolBalance` in the swapping system).
 
 OrderBook:
 - `BalanceIn`
 - `BalanceOut`
 
 Still struggle with how to add a type balance level to the swapping system because swaps are used in both directions. Is it possible for other PR if I find a clean way to do it?
 
 ## Future improvements can be:
 - `pallet-investments` with `Balance` and `TrancheBalance`.
 - `liquidity-pools` with `ForeignBalance` and `TrancheBalance`.
 - `pallet-pool-system` with `Balance` and `TrancheBalance`


wischli (approved):
Super straight forward and clean - great improvement! 

IMO, it makes most sense for FI because in the Orderbook, the meaning of `BalanceIn` and `BalanceOut` gets a bit fuzzy since we can have swaps into both directions such that `BalanceIn == BalanceOut` is implied more or less.

lemunozm:
> the meaning of BalanceIn and BalanceOut gets a bit fuzzy since we can have swaps into both directions such that BalanceIn == BalanceOut is implied more or less.

Aggre. My idea for the swapping part is that this generic should be in the methods, so those `BalanceOut` and `BalanceIn` depend on how the caller uses it. i.e. `convert_by_market<In, Out>(amount_out)`.

From OrderBook's perspective, I think it make sense to have the implementation of the pallet organized by `BalanceOut` and `BalanceIn`, the problem is how to reprensent later the `TokenSwaps` trait.

wischli (approved):

PR #1720, opened 2024-02-06
Feat: LP integration tests

mustermeiszer:
# Description
Currently, we can only test EVM <> Substrate interactions in the UI/testnetworks. This is cumbersome and not reliable for testing changes both on the Substrate as also on the EVM side. 
This PR enables to write real e2e tests for both codebases.

## Changes and Descriptions
* builds https://github.com/centrifuge/liquidity-pools and returns all needed `*.json` files for creating contracts
* provides an additional `trait EvmEnv` that provides interaction with EVM in the test codebase

## How to test

1. [Install](https://book.getfoundry.sh/getting-started/installation) Foundry Forge CLI
2. Run Tests
```
cargo test -p runtime-integration-tests 
```
If you encounter an error with an unset `LP_SOL_SOURCES` env var, the script did not build the solidity contracts in `target/debug/build/runtime-integration-tests-*/out/liquidity-pools` properly. Try removing all `runtime-integration-tests` directories in `target/debug/build/` and `runtime_integration_tests` ones in `target/debug/debs/`. If that doesn't help, clear the entire `target` directory.

# Checklist:


runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -46,6 +55,258 @@ impl<T: Runtime> Default for RuntimeEnv<T> {
...
+			.contract
+			.function(function.into().as_ref())
+			.expect(ESSENTIAL)
+			.encode_input(args.unwrap_or_default())
+			.expect(ESSENTIAL);
+
+		// TODO: Needs to imutable actually, but the current state implementation does
+		//       not rollback but error out upon changes, which is not ideal if you want
+		// to       test stuff without altering your state.
+		self.parachain_state_mut(|| {
```
lemunozm, line 284: Interesting. Maybe we should change the `*_state()` behavior to make a rollback instead

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
```diff
@@ -0,0 +1,102 @@
...
+		config::Runtime,
+		env::{Env, EvmEnv},
+	},
+	utils::accounts::Keyring,
+};
+
+#[test]
+fn _test() {
+	add_currency::<centrifuge_runtime::Runtime>()
+}
```
lemunozm, line 34: I guess this can be removed because it's already tested by `test_for_runtimes` line

runtime/integration-tests/src/generic/cases/lp/mod.rs
```diff
@@ -0,0 +1,912 @@
...
+pub mod pool_management;
+
+const DEFAULT_BALANCE: Balance = 1_000_000;
+
+pub const DECIMALS_6: Balance = 1_000_000;
+pub const DECIMALS_18: Balance = 1_000_000_000_000_000_000;
+
+pub const USDC: CurrencyId = CurrencyId::ForeignAsset(100_001);
+pub const DAI: CurrencyId = CurrencyId::ForeignAsset(100_002);
+pub const FRAX: CurrencyId = CurrencyId::ForeignAsset(100_003);
```
lemunozm, line 61: Maybe you can register these currencies with a `CurrencyInfo` in `generic/utils/currency.rs` to later use them more easily.

runtime/integration-tests/src/generic/cases/lp/mod.rs
```diff
@@ -0,0 +1,912 @@
...
+
+	// Deploy LP and more for both pools and all currencies
+
+	// TODO: Does panic... Have we ever tested that? Building a block would be super
+	// useful to flush the pending of the evm side
+	// env.__priv_build_block(1);
+	env
+}
+
+pub fn register_asset<T: Runtime>(
```
lemunozm, line 868: Regarding the above comment, see `register_currency()` from `generic/utils/currency.rs`

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
```diff
@@ -0,0 +1,102 @@
...
+
+fn add_currency<T: Runtime>() {
```
lemunozm, line 36: I really like this!!

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -46,6 +55,258 @@ impl<T: Runtime> Default for RuntimeEnv<T> {
...
+			.expect("Contract not deployed")
+			.clone();
+		let input = contract_info
+			.contract
+			.functions_by_name(function.into().as_ref())
+			.expect(ESSENTIAL)
+			.iter()
+			.filter_map(|f| f.encode_input(args.unwrap_or_default()).ok())
+			.collect::<Vec<_>>()
+			.pop()
```
lemunozm, line 230:
```suggestion
			.last()
```
Super NIT

lemunozm (commented):
Really cool testing feature and added very clean! 🚀 

Left some minor questions, suggestions and NITs. But everything is minor. Good job!

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -46,6 +55,258 @@ impl<T: Runtime> Default for RuntimeEnv<T> {
...
+		caller: Keyring,
+		value: U256,
+		contract: impl Into<String>,
+		function: impl Into<String>,
+		args: Option<&[Token]>,
+	) -> Result<CallInfo, DispatchError> {
+		let contract_info = self
+			.deployed_contracts
+			.get(&contract.into())
+			.expect("Contract not deployed")
```
cdamian, line 221: Nit - return a `DispatchError::Other`.

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -46,6 +55,258 @@ impl<T: Runtime> Default for RuntimeEnv<T> {
...
+			.clone();
+		let input = contract_info
+			.contract
+			.functions_by_name(function.into().as_ref())
+			.expect(ESSENTIAL)
```
cdamian, line 226: Nit - same here, return an actual error.

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -46,6 +55,258 @@ impl<T: Runtime> Default for RuntimeEnv<T> {
 	}
 }
 
+const GAS_LIMIT: u64 = 5_000_000;
+const VALIDATE: bool = true;
+const TRANSACTIONAL: bool = true;
+
+impl<T: Runtime> EvmEnv<T> for RuntimeEnv<T> {
```
cdamian, line 62: Do we also want to implement this for the `FudgeEnv` at some point? ^^

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: I think the idea here is to return errors only from EVM side. Note that this is test code, and you want to panic as close to the origin of the error as possible if this method is called wrongly.

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: The same applies here

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "runtime/integration-tests/submodules/liquidity-pools"]
+	path = runtime/integration-tests/submodules/liquidity-pools
+	url = git@github.com:centrifuge/liquidity-pools.git
+	branch = feat/local-router
```
cdamian, line 4: Eventually, we want to use the `main` branch here right?

cdamian:
Looks awesome so far @mustermeiszer! Thanks for taking care!

runtime/integration-tests/build.rs
```diff
@@ -0,0 +1,114 @@
...
+				"cargo:warning=Git fetch failed with: \n  - status: {}\n   -stderr: {}",
+				o.status,
+				String::from_utf8(o.stderr).expect("stderr is utf-8 encoded. qed.")
+			);
+		}
+		Err(err) => {
+			println!("cargo:warning=Failed to execute git command: {}", err);
+		}
+	}
+	 */
```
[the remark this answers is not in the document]
wischli: +1. AFAICT, we should solve this via CI which regularly pulls the latest contracts

runtime/integration-tests/src/generic/envs/runtime_env.rs
cdamian: If we do want to have a mix of panics and returns, then we should at least adapt the error message to make it easier while debugging.

.gitmodules
mustermeiszer: Feature branch for the local router we need for routing to local evm. Will be master once this is merged

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: I would say the choice is (for testing):
- If you want to handle the error as a possible result, that should be handled in another part of the code, then `DispatchError`. In this case, the error coming from EVM is something we could want to inspect/check in some use cases.
- If the method is supposed to work in a particular manner, and the error is a programming error (i.e., not fulfilling a precondition of it), then `unwrap()` or `expect(),` as soon as the error is originated to find fast the root of the problem.

runtime/integration-tests/build.rs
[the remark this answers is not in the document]
lemunozm: > we should solve this via CI which regularly pulls the latest contracts

One concern regarding this is that an update in `liquidity-pools` can break our CI. I suppose it is `centrifuge-chain` who should choose the commit used from `liquidity-pools`. When liquidity-pools is updated, then we should create a commit in `centrifuge-chain` just updating the dependency.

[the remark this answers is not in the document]
mustermeiszer: I am not 100% sure on the process here, tbh. @hieronx wdyt?

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -46,6 +55,258 @@ impl<T: Runtime> Default for RuntimeEnv<T> {
...
+			.event(Into::<String>::into(event).as_ref())
+			.unwrap();
+
+		self.parachain_state(|| {
+			pallet_ethereum::Pending::<T>::get()
+				.into_iter()
+				.map(|(_, status, _)| status.logs)
+				.flatten()
+				.collect::<Vec<_>>()
+				.into_iter()
```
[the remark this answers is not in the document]
mustermeiszer: What is the suggested change here? Looks the same ^^

runtime/integration-tests/src/generic/envs/runtime_env.rs
mustermeiszer: We should definitely. There should be state changes to be allowed in `parachain_state` that get just rolled back.

mustermeiszer: Where can I add that?

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
mustermeiszer: Just for my internal testing, to easily call a test method. will be removed.

runtime/integration-tests/src/generic/cases/lp/mod.rs
mustermeiszer: Is this a genesis registration or just more easily calling registration?

runtime/integration-tests/build.rs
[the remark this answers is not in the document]
hieronx: > When liquidity-pools is updated, then we should create a commit in centrifuge-chain just updating the dependency.

I think this makes sense. Also because we probably want to be able to run the main tests against the release commit from liquidity pools (what is live now), but also want to be able to run the tests in a branch against a more recent development commit. Having the commit be chosen manually would make this possible.

runtime/integration-tests/src/generic/envs/runtime_env.rs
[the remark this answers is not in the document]
lemunozm: Those lines can be removed

runtime/integration-tests/src/generic/envs/runtime_env.rs
mustermeiszer: I want an owned value not the reference though.

runtime/integration-tests/src/generic/envs/runtime_env.rs
lemunozm: We can wrap the `f()` call inside `RuntimeEnv::parachain_state()` with a transaction and revert in case of an error. For reference: https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/loans/src/lib.rs#L635

runtime/integration-tests/src/generic/cases/lp/mod.rs
lemunozm: A more easy calling registration. The point with `CurrencyInfo` is that you have all related information associated with a currency and tied to that currency. So you do not need to have extra constants like `DECIMALS_X`, you can do `Frax::DECIMALS` or `Frax::ID` and things like that. Just less error prone in testing

runtime/integration-tests/src/generic/cases/lp/mod.rs
```diff
@@ -0,0 +1,1071 @@
...
+		);
+	});
+
+	additional(&mut env);
+
+	env.pass(Blocks::ByNumber(1));
+	env
+}
+
+pub fn setup_investment_currencies<T: Runtime>(env: &mut impl EvmEnv<T>) {
```
wischli, line 792:
```suggestion
pub fn setup_investment_currencies<T: Runtime>(env: &mut impl EvmEnv<T>) {
	unimplemented!("")
```

runtime/integration-tests/src/generic/cases/lp/mod.rs
```diff
@@ -0,0 +1,1071 @@
...
+	// AllowInvestmentCurrency 2
+	// AllowInvestmentCurrency 3
+
+	// Pool 2, Tranche 2
+	// AllowInvestmentCurrency 1
+	// AllowInvestmentCurrency 2
+	// AllowInvestmentCurrency 3
+}
+
+pub fn setup_deploy_lps<T: Runtime>(env: &mut impl EvmEnv<T>) {
```
wischli, line 804:
```suggestion
pub fn setup_deploy_lps<T: Runtime>(env: &mut impl EvmEnv<T>) {
	unimplemented!("")
```

runtime/integration-tests/src/generic/cases/lp/mod.rs
```diff
@@ -0,0 +1,1071 @@
...
+	// ------------------ EVM Side ----------------------- //
+	// Deploy LP and more for both pools and all currencies
+}
+
+pub fn setup_tranches<T: Runtime>(env: &mut impl EvmEnv<T>) {
```
wischli, line 809:
```suggestion
pub fn setup_tranches<T: Runtime>(env: &mut impl EvmEnv<T>) {
	unimplemented!("")
```

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
```diff
@@ -0,0 +1,249 @@
...
+			OriginFor::<T>::signed(Keyring::Admin.into()),
+			POOL,
+			Domain::EVM(EVM_DOMAIN_CHAIN_ID)
+		));
+
+		utils::process_outbound::<T>()
+	});
+
+	// TODO: Actually find the revert event here.
+	//
```
wischli, line 159: Reminder for this TODO

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
```diff
@@ -0,0 +1,249 @@
...
+	// NOTE: That is really relevant for the router too.
+	//     We need to check `Pending` and see for errors
+	//     unfortuntately and then error in the router instead of returning success.
+
```
wischli, line 163: Could you elaborate on the comment? I cannot follow because I don't know which event you are referring to.

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
```diff
@@ -0,0 +1,249 @@
...
+		utils::process_outbound::<T>()
+	});
+}
+
+fn allow_investment_currency<T: Runtime>() {
+	let mut env = super::setup::<T>(|env| {
+		super::setup_currencies(env);
+		super::setup_pools(env);
+		super::setup_tranches(env);
+	});
```
wischli, line 209: Missing the purpose of this call

```suggestion
fn allow_investment_currency<T: Runtime>() {
	let mut env = super::setup::<T>(|env| {
		super::setup_currencies(env);
		super::setup_pools(env);
		super::setup_tranches(env);
	});
	todo!("allow investment currency");
```

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
```diff
@@ -0,0 +1,249 @@
...
+
+fn disallow_investment_currency<T: Runtime>() {
+	let mut env = super::setup::<T>(|env| {
+		super::setup_currencies(env);
+		super::setup_pools(env);
+		super::setup_tranches(env);
+		super::setup_investment_currencies(env);
+		super::setup_deploy_lps(env);
+	});
+}
```
wischli, line 220: Missing the purpose of this call

```suggestion
fn disallow_investment_currency<T: Runtime>() {
	let mut env = super::setup::<T>(|env| {
		super::setup_currencies(env);
		super::setup_pools(env);
		super::setup_tranches(env);
		super::setup_investment_currencies(env);
		super::setup_deploy_lps(env);
	});
	todo!("disallow investment currency");
}
```

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
```diff
@@ -0,0 +1,249 @@
...
+fn update_member<T: Runtime>() {
+	todo!()
+}
+
+fn update_tranche_token_metadata<T: Runtime>() {
+	let mut env = super::setup::<T>(|env| {
+		super::setup_currencies(env);
+		super::setup_pools(env);
+		super::setup_tranches(env);
+	});
```
wischli, line 231: Missing the purpose of this call

```suggestion
fn update_tranche_token_metadata<T: Runtime>() {
	let mut env = super::setup::<T>(|env| {
		super::setup_currencies(env);
		super::setup_pools(env);
		super::setup_tranches(env);
	});
	todo!("update tranche token metdata");
```

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+fn update_tranche_token_price<T: Runtime>() {
+	let mut env = super::setup::<T>(|env| {
+		super::setup_currencies(env);
+		super::setup_pools(env);
+		super::setup_tranches(env);
+	});
```
wischli, line 239: Missing the purpose of this call
```suggestion
fn update_tranche_token_price<T: Runtime>() {
	let mut env = super::setup::<T>(|env| {
		super::setup_currencies(env);
		super::setup_pools(env);
		super::setup_tranches(env);
	});
	todo!("update tranche token price");
```

wischli (commented):
Just pointing out a few open todos.

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
mustermeiszer: The problem is that the EVM is never returning an error. It is always Ok, even if reverted. We need to check the core return type of the evm. It is stored in the Pending storage of the evm on the frontier side.

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
mustermeiszer: Not sure what you mean?

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
mustermeiszer: Calling dissalow invesmtent currency.

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
mustermeiszer: updating the tranche token metadata. THat is just skeleoton.

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
mustermeiszer: Skeleton to know what should be set up for the test to work

runtime/integration-tests/src/generic/cases/lp/mod.rs
mustermeiszer: The setups are stackable. Depending on what you want to be set up. This allows to easily test situations. FOr the operations mod I wanted to test that setting up the LP worked at each step that involces messages to be transferred. 

But isolated.

wischli: Yeah I wanted to point out to you/myself, that the implementation is still missing.

runtime/integration-tests/src/generic/cases/lp/pool_management.rs
wischli: If I am not mistaken here, the current implementation only includes the setup but not the actual allowing of the investment currency. I am trying to locate the missing pieces of this PR. Please correct me if I am wrong about any of the pinged functions, i.e. if you believe they are complete in the current state.

runtime/integration-tests/src/generic/cases/lp/mod.rs
wischli: https://github.com/moonbeam-foundation/frontier/blob/c5d6daa9ffd46c0a85915526aa26d200fd635e30/frame/ethereum/src/lib.rs#L678-L679

.gitmodules
wischli: Update: Main branch cannot be used as the `LocalRouter` feat is required. We have been using the `centrifuge-chain/release-v1.0` branch: https://github.com/centrifuge/liquidity-pools/pull/260

runtime/integration-tests/build.rs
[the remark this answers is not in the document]
wischli: IMO, the Liquidity Pools repo should include a CI which runs LP integration tests from Centrifuge Chain and notify Protocol / Centrifuge Chain repo when it is failing such that we know a breaking change has been introduced on the EVM side.

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -82,14 +83,24 @@ impl<T: Runtime> Env<T> for RuntimeEnv<T> {
 		.assimilate_storage(&mut parachain_storage)
 		.unwrap();
 
+		// Precompiles need to have code-set
+		pallet_evm::GenesisConfig::<T> {
+			accounts: runtime_common::evm::precompile::utils::precompile_account_genesis::<
+				centrifuge_runtime::CentrifugePrecompiles,
```
lemunozm, line 89: This should be:

```suggestion
				T::PrecompilesType
```

Although I think the correct place for this is `utils/genesis.rs` to use it only for those tests that need `pallet_evm`.

mustermeiszer: No, that is a general setup. Took us 3 weeks to discover it again, we had it already discovered 6 month ago. So this should be default for all setups.

mustermeiszer: `T::RecompilesType` does not work 
image

I do not want to carry that to the `trait Runtime` as a where clause, this will need to be carried to every location otherwise IIRC.

mustermeiszer: Fixed it here: [125bc47](https://github.com/centrifuge/centrifuge-chain/pull/1720/commits/125bc47dff38b6355b73b107ec5040501a845e37)

lemunozm: Ok, if we only use it there, we can leave it as it is.

In case we need to use it in more places, then you can add a new associated type to `trait Runtime` as `type PrecompilesTypeExt: H160Addresses` and then instead of adding the pallet as `pallet_evm::Config`, adding it as: `pallet_evm::Config<PrecompileType = Self::PrecompileTypeExt>`

Later in the above code you can use `T::PrecompileTypeExt`

mustermeiszer: Yeah, thought about that too, but seemed to much work for the one location. Maybe will do later.

lemunozm: Not mandatory at all!

lemunozm:
BTW, I think most of the conflicts with `main` comes from `to_account_id()` changes in removed files. I think should be easy to fix them. But happy to help if you need support for the rebase.

mustermeiszer:
@wischli I think testing wise this is fine for now. We should try to close this, especially CI and then work on that from time to time. Will try to generate test in a "fuzzy" way, so that random orders of invest,cancel, etc. are possible. Otherwise, we will need to write too many edge-case scenarios. WDYT?

wischli:
> @wischli I think testing wise this is fine for now. We should try to close this, especially CI and then work on that from time to time. Will try to generate test in a "fuzzy" way, so that random orders of invest,cancel, etc. are possible. Otherwise, we will need to write too many edge-case scenarios. WDYT?

Absolutely agree. We need @gpmayorga's help unfortunately. Could you help us out here when you have time? It's not urgent. This PR adds a git submodule `liquidity-pools` which is linked to [our solidity contracts](https://github.com/centrifuge/liquidity-pools). They need to be part of the debug build, i.e. `target/debug/build/runtime-integration*/out/liquidity-pools` needs to exist. Sometimes, I also had issues getting the submodule build in the debug target directory. The quick fix was to 

```
rm -rf target/debug/deps/runtime_integration_tests-*
rm -rf target/debug/build/runtime-integration*

# Sometimes it was also necessary to 
cargo clean
```

I have tried everything I could think of to get the LP contracts built but they never made it into the `target/*/build/runtime-integration*` directory. This roots to us using SCCACHE which does not have the LP contracts in the cache. 

As a digest, I tried the following two approaches
1. Disabling cache in `sanity-checks.yml` with and without `SCCACHE_RECACHE=true` in `run-check.sh`. This did not suffice because SCCACHE was still required somewhere ([error](https://github.com/centrifuge/centrifuge-chain/actions/runs/8686906472/job/23819334905))
2. Removing all cache files in the repository's `[actions/caches](https://github.com/centrifuge/centrifuge-chain/actions/caches)`. This did not result in building the contracts unfortunately.

wischli:
@mustermeiszer Can you rebase to `main` before we get @gpmayorga involved?

mustermeiszer:
Will try. I already tried once, but the changes from main where quite confusing.

lemunozm:
This PR is 100% tests, right? Can we add it after doing the RU in case we lack time?

lemunozm:
The problem I see is that this PR will create a big amount of conflicts with `polkadot-1.7.2` I guess, because it's over `main`. So maybe it puts a big delay to have this merged today

mustermeiszer:
@lemunozm then lets merge Tranche tokens as is and we do the testing in dev. Wdyt?

runtime/integration-tests/src/generic/cases/liquidity_pools_transfers.rs
```diff
@@ -4327,166 +4324,6 @@ mod development {
 			});
 		}
 
-		#[test_runtimes([development])]
-		fn transfer_tranche_tokens_to_local<T: Runtime + FudgeSupport>() {
```
mustermeiszer, line 4331: Is now represented in the `/lp/transfers` test space using the evm.

runtime/integration-tests/src/generic/cases/liquidity_pools_transfers.rs
```diff
@@ -4327,166 +4324,6 @@ mod development {
...
-					&sending_domain_locator
-				)
-				.is_zero());
-			});
-		}
-
-		/// Try to transfer tranches for non-existing pools or invalid tranche
-		/// ids for existing pools.
-		#[test_runtimes([development])]
-		fn transferring_invalid_tranche_tokens_should_fail<T: Runtime + FudgeSupport>() {
```
mustermeiszer, line 4422: Is now represented in the `/lp/transfers` test space using the evm.

runtime/integration-tests/src/generic/cases/lp/transfers.rs
```diff
@@ -0,0 +1,392 @@
...
+				.unwrap()
+				.value,
+			),
+			AMOUNT
+		);
+	});
+}
+
+#[test_runtimes(all)]
+fn transfer_tranche_tokens_domain_to_local_to_domain<T: Runtime>() {
```
mustermeiszer, line 233: Testing https://github.com/centrifuge/centrifuge-chain/pull/1860

runtime/integration-tests/src/generic/cases/lp/transfers.rs
```diff
@@ -0,0 +1,392 @@
...
+				.unwrap()
+				.value,
+			),
+			AMOUNT
+		);
+	});
+}
+
+#[test_runtimes(all)]
+fn transfer_tranche_tokens_domain_to_local<T: Runtime>() {
```
mustermeiszer, line 306: Testing https://github.com/centrifuge/centrifuge-chain/pull/1860

runtime/integration-tests/build.rs
```diff
@@ -0,0 +1,114 @@
...
+			.to_str()
+			.expect("OsStr is utf-8. qed");
+
+		out_dir_build.push(parent);
+		let out_dir_build = out_dir_build
+			.as_os_str()
+			.to_str()
+			.expect("OsStr is utf-8. qed");
+
+		match Command::new("forge")
```
cdamian, line 73: Should we add a small script for downloading a specific version of this or at least mention the version that is known to work. I guess we want to avoid having failures here by using a version that might behave differently.

wischli: IMO, there definitely needs to be some improvement to this. Maybe it is sufficient to add `forge` install to our `init.sh` script for local development. For CI, forge is already installed manually. Eventually, there should be one codepath setting up requirements both for local development. as well as CI.

runtime/integration-tests/src/generic/cases/lp/mod.rs
```diff
@@ -0,0 +1,1953 @@
...
+		// Fund gateway sender
+		give_balance::<T>(
+			<T as pallet_liquidity_pools_gateway::Config>::Sender::get(),
+			DEFAULT_BALANCE * CFG,
+		);
+
+		// Register general local pool-currency
+		register_currency::<T>(LocalUSDC, |_| {});
+
+		/* TODO: Use that but index needed contracts afterwards
```
cdamian, line 558: Nit - do we want this now or should we have a follow-up for this?

runtime/integration-tests/build.rs
```diff
@@ -0,0 +1,114 @@
...
+			.args(["build", "--out", out_dir_build])
+			.output()
+		{
+			Ok(o) if o.status.success() => {
+				let source = match parent {
+					"liquidity-pools" => {
+						println!(
+							"cargo:info=Build liquidity-pools Solidity contracts. Stored at {} ",
```
cdamian, line 81: `cargo:info` doesn't seem to be supported.

runtime/integration-tests/build.rs
```diff
@@ -0,0 +1,114 @@
...
+							LP_SOL_SOURCES
+						);
+
+						LP_SOL_SOURCES
+					}
+					_ => {
+						println!(
+							"cargo:warning=Unknown solidity source build. Name: {}",
```
cdamian, line 89: The correct format is `cargo::warning`.

runtime/integration-tests/build.rs
```diff
@@ -0,0 +1,114 @@
...
+const LP_SOL_SOURCES: &str = "LP_SOL_SOURCES";
+
+fn main() {
+	let paths = fs::read_dir("./submodules/")
+		.expect("Submodules directory must exist for integration-tests");
+	let out_dir = env::var("OUT_DIR").expect("Cargo sets OUT_DIR environment variable. qed.");
+
+	/*
+	match Command::new("git")
+		.args(&["pull", "--all", "--recurse-submodules=yes"])
```
cdamian, line 23: Note - Locally, I had to run a `git submodule update --init --recursive` first.

runtime/integration-tests/build.rs
```diff
@@ -0,0 +1,114 @@
...
+			}
+			Ok(o) => {
+				println!(
+					"cargo:warning=forge build failed with: \n  - status: {}\n   -stderr: {}",
+					o.status,
+					String::from_utf8(o.stderr).expect("stderr is utf-8 encoded. qed.")
+				);
+			}
+			Err(err) => {
+				eprintln!("cargo:warning=Failed to execute git command: {}", err);
```
cdamian, line 110: Can you please change to `println!` here?

runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
```diff
@@ -40,9 +40,12 @@ fn call<T: Runtime>() {
 		utils::evm::mint_balance_into_derived_account::<T>(creator_address, 1 * CFG);
 		utils::evm::mint_balance_into_derived_account::<T>(sender_address, 1 * CFG);
 
-		let contract_address = utils::evm::deploy_contract::<T>(
-			creator_address,
-			hex::decode(TEST_CONTRACT_CODE).unwrap(),
+		let contract_address = sp_core::H160::from(
+			utils::evm::deploy_contract::<T>(
```
lemunozm, line 44: NIT: `deploy_contact()` already returns a `H160`

runtime/integration-tests/src/generic/env.rs
```diff
@@ -166,6 +175,58 @@ pub trait Env<T: Runtime>: Default {
 	fn __priv_build_block(&mut self, i: BlockNumber);
 }
 
+pub trait EnvEvmExtension<T: Runtime>: Env<T> {
+	type EvmEnv: EvmEnv<T>;
+
+	/// Allows to mutate the parachain storage state through the closure.
+	fn state_mut<R>(&mut self, f: impl FnOnce(&mut Self::EvmEnv) -> R) -> R;
```
lemunozm, line 182: Because we have a bunch of these methods, I would call it `parachain_state_mut_with_evm()` or similar. 

Same with `state()`

runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
```diff
@@ -25,9 +25,9 @@ use crate::generic::{
 // 	}
 pub const TEST_CONTRACT_CODE: &str = "608060405234801561001057600080fd5b50610113806100206000396000f3fe6080604052348015600f57600080fd5b506004361060325760003560e01c8063c2985578146037578063febb0f7e146057575b600080fd5b603d605f565b604051808215151515815260200191505060405180910390f35b605d6068565b005b60006001905090565b600060db576040517f08c379a00000000000000000000000000000000000000000000000000000000081526004018080602001828103825260098152602001807f6572726f725f6d7367000000000000000000000000000000000000000000000081525060200191505060405180910390fd5b56fea2646970667358221220fde68a3968e0e99b16fabf9b2997a78218b32214031f8e07e2c502daf603a69e64736f6c63430006060033";
 
-#[test_runtimes([development])]
+#[test_runtimes(all)]
 fn call<T: Runtime>() {
-	RuntimeEnv::<T>::default().parachain_state_mut(|| {
```
lemunozm, line 30: Is this change really needed?

runtime/integration-tests/src/generic/envs/evm_env.rs
```diff
@@ -0,0 +1,218 @@
...
+	},
+	utils::accounts::Keyring,
+};
+
+const GAS_LIMIT: u64 = 105_000_000;
+const VALIDATE: bool = false;
+const TRANSACTIONAL: bool = true;
+
+pub struct EvmEnv<T: Runtime> {
+	sol_contracts: Option<HashMap<String, ContractInfo>>,
```
lemunozm, line 26: NIT: Could it directly `HashMap`? Is there a difference between `None` and an empty `HashMap`?

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -171,6 +191,31 @@ impl<T: Runtime> RuntimeEnv<T> {
 		let mut ext = sp_io::TestExternalities::new(storage);
 
 		ext.execute_with(|| {
+			let accounts = match T::KIND {
+				RuntimeKind::Centrifuge => {
+					runtime_common::evm::precompile::utils::precompile_account_genesis::<
+						centrifuge_runtime::CentrifugePrecompiles,
+					>()
```
lemunozm, line 198: Not sure if asked before, the following doesn't work?

```suggestion
					runtime_common::evm::precompile::utils::precompile_account_genesis::<
						T::PrecompilesType,
					>()
```

for all runtimes?

runtime/integration-tests/src/generic/cases/lp/mod.rs
mustermeiszer: Follow up. It requires setting up quite a few hacky contracts for foundry. Possible, but hard. Would of course make it possible for us to not have to refactor the huge list of deployments below all the time, but something for the next iteration.

runtime/integration-tests/build.rs
mustermeiszer: Do we want that in or should everybody pull the deps themselves?

runtime/integration-tests/build.rs
mustermeiszer: Yes!

lemunozm (commented):
First review pass. Looks pretty clean despite the amount of code in the PR 💯 

Just some Nits and question. Nothing important. In my second round I wanted to review the cases. Hopefully today or tomorrow

runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
lemunozm: The diff seem off. The new line is:

```rust
RuntimeEnv::<T>::default().state_mut(|_evm| {
```

runtime/integration-tests/build.rs
mustermeiszer: I am unsure how to proceed here. I think for now, erroring out if `forge` is not installed is fine, or should we install it during the build script - i mean locally in the build directory?

runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
mustermeiszer: tech debt. Before your changes for 1.7.2 the version of `sp_core::H160` and `ethereum_types::H160` pointed to different versions of the same `H160`, making it necessary to convert them all the time. I will try to clean that up ^^

runtime/integration-tests/src/generic/env.rs
mustermeiszer: puh. Not sure this does not complicate it more than it needs to be tbh.

runtime/integration-tests/src/generic/envs/runtime_env.rs
mustermeiszer: Will try.

runtime/integration-tests/src/generic/mod.rs
```diff
@@ -14,8 +15,9 @@ mod cases {
 	mod ethereum_transaction;
 	mod example;
 	mod investments;
-	mod liquidity_pools;
+	mod liquidity_pools_transfers;
```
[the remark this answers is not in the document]
mustermeiszer: Will revert for now

runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
lemunozm: Uggh, I think that has been solved after cleaning the dependency tree

runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
lemunozm: I referred the line wrong. But what I wanted to ask is, that change is needed if we do not need the `evm` variable?

runtime/integration-tests/src/generic/env.rs
lemunozm: From a reader's perspective, it was confusing at first for me, what exactly it mutates. Not super strong opinion anyways.

runtime/integration-tests/src/generic/envs/runtime_env.rs
mustermeiszer:
```
197 |                         T::PrecompilesType,
    |                         ^^^^^^^^^^^^^^^^^^ the trait `H160Addresses` is not implemented for `<T as pallet_evm::Config>::PrecompilesType`
```

We had this already fix IIRC. But I am not really motived atm to put work into this.

runtime/integration-tests/src/generic/envs/evm_env.rs
mustermeiszer:
```
sol_contracts
			.as_ref()
			.expect("Need to load_contracts first")
```

Better error resolution IMO.

runtime/integration-tests/src/generic/env.rs
mustermeiszer: I mean, we are always mutating the parachain state, right? In 99% of cases.

mustermeiszer: Also, I think this makes using `trait Env` and `trait EnvEvmExtension` in the same namespace quite annoying, right? Will keep for now. We can revisit later, I anyways wanted to make the `EVM` a global, like Substrate storage. Then we can remove all of that here anyways.

runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
mustermeiszer: Tried fixing it here: https://github.com/centrifuge/centrifuge-chain/pull/1720/commits/e4e048a574e393061f0169e88d2d66b7e1970760

runtime/integration-tests/build.rs
mustermeiszer: using warning now. Maybe confusing, but I think helpful to know where they are stored, wdyt?

runtime/integration-tests/src/generic/env.rs
lemunozm: Makes sense the latter 👍🏻

runtime/integration-tests/src/generic/cases/lp/mod.rs
```diff
@@ -0,0 +1,1956 @@
...
+pub fn setup<T: Runtime, F: FnOnce(&mut <RuntimeEnv<T> as EnvEvmExtension<T>>::EvmEnv)>(
+	additional: F,
+) -> impl EnvEvmExtension<T> {
+	let mut env = RuntimeEnv::<T>::from_parachain_storage(
+		Genesis::default()
+			.add(genesis::balances::<T>(DEFAULT_BALANCE * CFG))
+			.storage(),
+	);
+	env.state_mut(|evm| {
+		evm_balances::<T>(DEFAULT_BALANCE * CFG);
```
lemunozm, line 547: I think once the `genesis::balance()` has been modified to take into account the evm accounts this call is no longer needed

lemunozm (approved):
LGTM! Let's merge this beast!

runtime/integration-tests/src/generic/cases/lp/mod.rs
mustermeiszer: The problem with `gensesis::balance()` is that we need to know the `ParaId` for the given chain in order to derive the evm based accounts. For this we are accessing the `ParaId` storage and this is with our current setup not possible by using `add` as it acts outside of an externalities provided environment.

mustermeiszer: tl;dr: I tried it already and we need a bigger change for that ^^

gpmayorga (approved):

runtime/integration-tests/src/generic/cases/lp/mod.rs
lemunozm: 🤔 Mmm... makes sense. Thanks for trying it!

lemunozm:
@cdamian I think we need your approval to get this in 💪🏻

cdamian (approved):

runtime/integration-tests/build.rs
lemunozm: > or should we install it during the build script - i mean locally in the build directory?

I think this is the best option. The script should fetch locally everything it needs to work properly.

runtime/integration-tests/build.rs
lemunozm: Agree with @cdamian that we should modify the message.

runtime/integration-tests/build.rs
lemunozm: Although compiling this implies compiling 900 crates 😅

runtime/integration-tests/build.rs
lemunozm: Maybe executing `echo` and printing the info message can work to avoid the warning?

PR #1721, opened 2024-02-06
Add asset_registry mock

lemunozm:
# Description

The current `cfg_test_utils::impl_mock_registry!` macro works pretty well for most of the cases, establishing the base functionality for an asset registry. But added a mock builder version for cases where we wanted to test some border cases related to the asset registry.

This PR is not mandatory at all and can be left unmerged here until the mock is required by any pallet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1158,6 +1158,7 @@ dependencies = [
  "sp-io",
  "sp-runtime",
  "sp-std",
+ "xcm",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


mustermeiszer (approved):
Would already bring it in, otherwise, we will forget it I guess ^^

wischli (approved):
This will come in handy for #1713! I have the tests already set up and will refactor with this mock.

lemunozm:
Great then!

PR #1723, opened 2024-02-07
Fix: FI compute orders correclty for asymmetric ratios

lemunozm:
# Description

The issue is described in this [thread](https://kflabs.slack.com/archives/C04GJQAM9P0/p1707298087515729)

> When we decrease, and there is an increasing pending swap,  we use the `convert_by_market()` to know upfront the amount we need to cancel to the inverse swap and distinguish from the amount we need to decrease the investment.
Nevertheless, later, inside the swap, `convert_by_market()` is called again in the inverse direction to perform the cancellation of the swap. This works fine. However, with asymmetric ratios (A -> B, ratio: 0.9; B -> A, ratio 0.95), it can lead to obtaining a different swap than expected. In the worst case, it would lead to a NoFund error if the amounts are very fit.

This is a non-pretty fix that works. Nevertheless, I would wait until I have the `pallet-swaps` refactor, where I think it can be fixed much more cleanly.

### DIFF
--- a/pallets/foreign-investments/src/entities.rs
+++ b/pallets/foreign-investments/src/entities.rs
@@ -163,7 +163,7 @@ impl<T: Config> InvestmentInfo<T> {
 		who: &T::AccountId,
 		investment_id: T::InvestmentId,
 		foreign_amount: T::ForeignBalance,
-	) -> Result<SwapOf<T>, DispatchError> {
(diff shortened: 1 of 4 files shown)


lemunozm:
Close in favor of: https://github.com/centrifuge/centrifuge-chain/pull/1746

PR #1724, opened 2024-02-07
Refactor: Extract swapping part from FI into `pallet-swaps`

lemunozm:
# Description

Remove the machinery to make swaps (with canceling inverse support) from FI and move it to its own pallet. This helps to reason easier from FI, leaving the swapping part complexity in `pallet_swaps` and making it reusable for future utilities. 

The interface to let FI access the swapping logic is by using the `Swaps` trait.

## Changes and Descriptions

- The swap module and storage related to swapping moved to `pallet_swaps`.
- `SwapId` is no longer an `OrderId`. `SwapId` is now the identification of the swap in `pallet_swaps`, which currently, for FI it is `(InvesmentId, Action)`. FI no longer knows about `OrderId`.
  - In the future, for multiple foreign currencies, it will be `(InvestmentId, Action, CurrencyId)` instead.
- When a fulfill notification is generated in OrderBook, it notifies now `pallet-swaps`, who filter the notification and pass then to FI.

# Checklist:
- [x] pallet-swaps implementation and testing
- [x] adapt FI
- [x] adapt Runtimes
- [x] adapts `integration-tests`

Pending a good fix for this issue (https://github.com/centrifuge/centrifuge-chain/pull/1723), but probably in another PR.


libs/traits/src/lib.rs
```diff
@@ -544,7 +572,7 @@ pub trait TokenSwaps<Account> {
 	fn cancel_order(order: Self::OrderId) -> DispatchResult;
 
 	/// Retrieve the details of the order if it exists.
-	fn get_order_details(order: Self::OrderId) -> Option<Self::OrderDetails>;
+	fn get_order_details(order: Self::OrderId) -> Option<Swap<Self::BalanceOut, Self::CurrencyId>>;
```
lemunozm, line 575: @wischli. I needed to move `Swap` to the interface section, because now it's used in the `Swaps` trait. For that, I use the `Swap` type directly here. But I suppose this will be refactorized by you with a new type. Sorry for the conflicts on that! 🙏🏻

libs/traits/src/lib.rs
```diff
@@ -555,6 +583,78 @@ pub trait TokenSwaps<Account> {
...
+	fn valid_pair(currency_in: Self::CurrencyId, currency_out: Self::CurrencyId) -> bool;
+
+	/// Makes a conversion between 2 currencies using the market ratio between
+	/// them
+	// TODO: Should be removed after #1723
+	fn convert_by_market(
+		currency_in: Self::CurrencyId,
+		currency_out: Self::CurrencyId,
+		amount_out: Self::Amount,
+	) -> Result<Self::Amount, DispatchError>;
```
lemunozm, line 655: This should not be in the trait. Once https://github.com/centrifuge/centrifuge-chain/pull/1723 is fixed. Getting the market rate is very low level for this trait, I guess.

pallets/foreign-investments/src/lib.rs
```diff
@@ -72,9 +68,8 @@ pub type ForeignId<T> = (
 /// Swap alias
 pub type SwapOf<T> = Swap<<T as Config>::SwapBalance, <T as Config>::CurrencyId>;
 
-/// Swap state alias
-pub type SwapStateOf<T> =
-	SwapState<<T as Config>::SwapBalance, <T as Config>::SwapBalance, <T as Config>::CurrencyId>;
+/// Identification of a swap from foreing-investment perspective
+pub type SwapId<T> = (<T as Config>::InvestmentId, Action);
```
lemunozm, line 72: Note that the meaning of `SwapId` has changed regarding the previous implementation (see the description of the PR)

lemunozm (commented):
Relevant notes

pallets/foreign-investments/src/mock.rs
```diff
@@ -131,6 +129,15 @@ impl pallet_mock_pools::Config for Runtime {
 	type TrancheId = TrancheId;
 }
 
+impl pallet_swaps::Config for Runtime {
+	type Balance = Balance;
```
wischli, line 133: Since we have `BalanceIn` and `BalanceOut` in the Orderbook, shouldn't we also have it in the swaps pallet?

pallets/swaps/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version (see http://www.gnu.org/licenses).
+
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+//! # Swaps pallet: Enables applying swaps independiently of previous swaps in the same or opposite
```
wischli, line 14: Nit
```suggestion
//! # Swaps pallet: Enables applying swaps independently of previous swaps in the same or opposite
```

pallets/swaps/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	/// Configure the pallet by specifying the parameters and types on which it
+	/// depends.
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		/// Represents an amount that can be swapped
+		type Balance: Parameter + Member + AtLeast32BitUnsigned + Default + Copy + MaxEncodedLen;
```
wischli, line 46: See above for my question of whether we should have typed balance here as well.

runtime/altair/src/lib.rs
```diff
@@ -1810,6 +1819,7 @@ construct_runtime!(
 		// our pallets
 		Fees: pallet_fees::{Pallet, Call, Storage, Config<T>, Event<T>} = 90,
 		Anchor: pallet_anchors::{Pallet, Call, Storage} = 91,
+		Swaps: pallet_swaps::{Pallet, Storage} = 92,
```
wischli, line 1822: Maybe CR: IMO we shouldn't re-use indices of removed pallets. In this case, it does not matter. But in general we could receive an XCM pointing to the incorrect pallet index.

runtime/centrifuge/src/lib.rs
```diff
@@ -1900,7 +1909,7 @@ construct_runtime!(
 		// our pallets
 		Fees: pallet_fees::{Pallet, Call, Storage, Config<T>, Event<T>} = 90,
 		Anchor: pallet_anchors::{Pallet, Call, Storage} = 91,
-		// Removed: Claims = 92
+		Swaps: pallet_swaps::{Pallet, Storage} = 92,
```
wischli, line 1912: See above

wischli (commented):
Another great improvement for our foreign investments! A have a small mix of nits, questions and a potential CR.

libs/traits/src/lib.rs
lemunozm: I guess, previously, it was generic because `Swap` was in `cfg-types` and `cfg-traits` has no access to it. But it is expected to always have that value so I would say yes.

pallets/foreign-investments/src/mock.rs
lemunozm: That is a good question that I always ask to myself 😆. I'll give another thought.  My last thought was that for the `Swaps` trait has no meaning for it, because the incoming balance is unknown. But for the `pallet-swaps` implementation it can make sense.

runtime/altair/src/lib.rs
lemunozm: I didn't know that. Thanks for pointing it out!

Given that these indexes were not used by XCM, maybe we can reuse this as an exception? And fill the gap? If not I revert it

pallets/foreign-investments/src/mock.rs
lemunozm: Okay, I was trying to refactor the pallet using `BalanceIn` and `BalanceOut`, and I think it's not very useful in this case (now). The problem is that `OrderBook` is used in both directions, and we end up with many `.into()` everywhere, losing the "static" benefits of dividing the balances.

So, I would say that in this case, we are making the code "dirty" for almost no static benefit.

Anyway, I still need to think about the `TokenSwaps` trait, which ideally should use the generic in its methods instead of in the trait. Thus, its usage will be much better.  Maybe we can even start using the GATs Rust feature here 🤔. But a task for the future.

lemunozm: Wanted to play with GATs, and it seems to work pretty well for this issue: https://github.com/centrifuge/centrifuge-chain/pull/1727

runtime/altair/src/lib.rs
wischli: Would be interested in @mustermeiszer's thoughts. I think it would be fine to do an exception here because wallets require the latest metadata which will apply this.

mustermeiszer: I am not 100% sure tbh. The fn-signature will not match so the tx would be invalid I guess. In general I think we can re-use indices but given that this is blockchain maybe lets not do until we ran out of indices ^^.

lemunozm: Ok, then I move out to a new index

pallets/foreign-investments/src/impls.rs
```diff
@@ -218,55 +218,49 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
...
-			Ok((who, investment_id, action)) => {
-				let pool_currency = pool_currency_of::<T>(investment_id)?;
-				let swapped_amount_in = swap_state.swapped_in;
-				let swapped_amount_out = swap_state.swapped_out;
-				let pending_amount = swap_state.remaining.amount_out;
-
-				if pending_amount.is_zero() {
-					Swaps::<T>::update_id(&who, investment_id, action, None)?;
-				}
+	type Id = (T::AccountId, SwapId<T>);
```
mustermeiszer, line 230: What prevents us from using Currency already? Is there more work to be done besides extending the key?

pallets/foreign-investments/src/lib.rs
mustermeiszer: **MAYBE CR**

Aren't we missing a migration for the existing storage to the one used in this case here?

pallets/swaps/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+			Error = DispatchError,
+		>;
+	}
+
+	/// Maps a `OrderId` to its corresponding `AccountId` and `SwapId`
+	///
+	/// NOTE: The storage is killed when the swap order no longer exists
+	#[pallet::storage]
+	pub(super) type OrderIdToSwapId<T: Config> =
+		StorageMap<_, Blake2_128Concat, T::OrderId, (T::AccountId, T::SwapId)>;
```
mustermeiszer, line 79: At least this storage needs to get populated, right?

mustermeiszer (changes requested):
Code looks good, but IIUC we are missing a migration.

pallets/foreign-investments/src/impls.rs
lemunozm: Yes, i.e. right now, we can not know in which currency we should collect when a redemption collection when it's notified. `pallet-investment` just notify with the `investment_id`, and having more than one foreign currency we do not know in which currency we should transform that. Even worst, we can collect an amount that comes from a mix of redemptions in different foreign currencies.

We should specify how to proceed in this case. That would imply a new storage with all currencies associated with an `investment_id`(maybe a `Vec` or a `Set`), and iterate them in some order, splitting the pool collected amount into different foreign currencies. Not trivial 😓

pallets/swaps/src/lib.rs
lemunozm: The previous storage does not exist yet in `altair` and `centrifuge`. Just Dev, but I guess we will reset Dev with the new RU.

mustermeiszer: But `FI` already exists, right? And we are changing that quite a lot. Don't we need a migration there, and if so, shouldn't this migration be adapted to the changes made here? Maybe @wischli can help me? If no migration needed not CR from my side.

lemunozm: But the big storage refactor happened in the previous merged PR. I mean, in case of a need for a migration, we should have done that already. IIRC, we said that we could purge the previous FI (?)

Anyways, whether or not migration occurs, this PR is independent of that.

lemunozm: IIRC, the point was that if there are no pending investments at the RU moment, we can reset the storage

wischli (approved):
Approving because code LGTM. If we decide on writing a migration, that can happen in a subsequent PR because the migration is a result of #1698

mustermeiszer (approved):
All of the above fine for me.

PR #1727, opened 2024-02-09
Refactor (PoC): Use GATs for token swaps traits

lemunozm:
# Description

`TokenSwaps` & `Swaps` traits can be more static if they define `AmoutOut` and `AmountIn` to not mismatch different kinds of amounts. Nevertheless, the "direction" usage of this trait depends on the moment when they are called, and from the pallet perspective we can not define upfront what is "in" or "out", because the direction of their usage is different in each part of the pallet implementation, when the methods are called.

Nevertheless, the GAT Rust feature comes to the rescue, allowing parameterization at the moment of the call, allowing a much cleaner and static API and implementation.

### DIFF
--- a/pallets/swaps/src/lib.rs
+++ b/pallets/swaps/src/lib.rs
@@ -42,8 +42,16 @@ pub mod pallet {
 	/// depends.
 	#[pallet::config]
 	pub trait Config: frame_system::Config {
+		/// Represents an amount that has been swapped
+		type BalanceIn: Parameter + Member + AtLeast32BitUnsigned + Copy + MaxEncodedLen;
+
 		/// Represents an amount that can be swapped
-		type Balance: Parameter + Member + AtLeast32BitUnsigned + Default + Copy + MaxEncodedLen;
+		type BalanceOut: Parameter
+			+ Member
+			+ AtLeast32BitUnsigned
+			+ Copy
+			+ MaxEncodedLen
+			+ Into<Self::BalanceIn>;
 
 		/// An identification for a swap
 		type SwapId: Parameter + Member + Copy + Ord + MaxEncodedLen;
@@ -55,18 +63,18 @@ pub mod pallet {
 		type CurrencyId: Parameter + Member + Copy + MaxEncodedLen;
 
 		/// The type which exposes token swap order functionality
-		type OrderBook: TokenSwaps<
+		type OrderBook<BalanceFrom, BalanceTo>: TokenSwaps<
 			Self::AccountId,
+			BalanceFrom,
+			BalanceTo,
 			CurrencyId = Self::CurrencyId,
-			BalanceIn = Self::Balance,
-			BalanceOut = Self::Balance,
 			OrderId = Self::OrderId,
 		>;
 
 		/// The hook which acts upon a (partially) fulfilled the swap
 		type FulfilledSwap: StatusNotificationHook<
 			Id = (Self::AccountId, Self::SwapId),
-			Status = SwapState<Self::Balance, Self::Balance, Self::CurrencyId>,
+			Status = SwapState<Self::BalanceIn, Self::BalanceOut, Self::CurrencyId>,
 			Error = DispatchError,
 		>;
 	}
@@ -131,12 +139,12 @@ pub mod pallet {
 		#[allow(clippy::type_complexity)]
 		fn apply_over_swap(
 			who: &T::AccountId,
-			new_swap: Swap<T::Balance, T::CurrencyId>,
+			new_swap: Swap<T::BalanceOut, T::CurrencyId>,
 			over_swap_id: Option<T::OrderId>,
-		) -> Result<(SwapStatus<T::Balance>, Option<T::OrderId>), DispatchError> {
+		) -> Result<(SwapStatus<T::BalanceIn, T::BalanceOut>, Option<T::OrderId>), DispatchError> {
 			match over_swap_id {
 				None => {
-					let order_id = T::OrderBook::place_order(
+					let order_id = T::OrderBook::<T::BalanceIn, T::BalanceOut>::place_order(
 						who.clone(),
 						new_swap.currency_in,
 						new_swap.currency_out,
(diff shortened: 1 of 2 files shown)


lemunozm:
PoC. Actually compiling with `cargo check -p pallet-swaps`

pallets/swaps/src/lib.rs
```diff
@@ -55,18 +63,18 @@ pub mod pallet {
...
-		type OrderBook: TokenSwaps<
+		type OrderBook<BalanceFrom, BalanceTo>: TokenSwaps<
 			Self::AccountId,
+			BalanceFrom,
+			BalanceTo,
 			CurrencyId = Self::CurrencyId,
-			BalanceIn = Self::Balance,
-			BalanceOut = Self::Balance,
 			OrderId = Self::OrderId,
 		>;
```
lemunozm, line 72: GAT definition

pallets/swaps/src/lib.rs
```diff
@@ -146,29 +154,34 @@ pub mod pallet {
...
+							.ok_or(Error::<T>::OrderNotFound)?;
 
 					if swap.is_same_direction(&new_swap)? {
 						let amount_to_swap = swap.amount_out.ensure_add(new_swap.amount_out)?;
-						T::OrderBook::update_order(order_id, amount_to_swap, OrderRatio::Market)?;
+						T::OrderBook::<T::BalanceIn, T::BalanceOut>::update_order(
+							order_id,
+							amount_to_swap,
+							OrderRatio::Market,
+						)?;
```
lemunozm, line 174: GAT usage, allowing to specify the direction of the order

pallets/swaps/src/lib.rs
```diff
@@ -55,18 +63,18 @@ pub mod pallet {
...
-		type OrderBook: TokenSwaps<
+		type OrderBook<BalanceTo, BalanceFrom>: TokenSwaps<
 			Self::AccountId,
+			BalanceTo,
+			BalanceFrom,
 			CurrencyId = Self::CurrencyId,
-			BalanceIn = Self::Balance,
-			BalanceOut = Self::Balance,
 			OrderId = Self::OrderId,
 		>;
```
lemunozm, line 72: GAT definition

pallets/swaps/src/lib.rs
```diff
@@ -146,29 +154,34 @@ pub mod pallet {
...
-					let swap = T::OrderBook::get_order_details(order_id)
-						.ok_or(Error::<T>::OrderNotFound)?;
+					let swap =
+						T::OrderBook::<T::BalanceIn, T::BalanceOut>::get_order_details(order_id)
+							.ok_or(Error::<T>::OrderNotFound)?;
 
 					if swap.is_same_direction(&new_swap)? {
 						let amount_to_swap = swap.amount_out.ensure_add(new_swap.amount_out)?;
-						T::OrderBook::update_order(order_id, amount_to_swap, OrderRatio::Market)?;
+						T::OrderBook::<T::BalanceIn, T::BalanceOut>::update_order(
```
lemunozm, line 170: GAT usage, specifying the direction of the order

pallets/swaps/src/lib.rs
```diff
@@ -181,7 +194,7 @@ pub mod pallet {
 								let amount_to_swap =
 									inverse_swap.amount_out.ensure_sub(new_swap_amount_in)?;
 
-								T::OrderBook::update_order(
+								T::OrderBook::<T::BalanceOut, T::BalanceIn>::update_order(
```
lemunozm, line 197: Or the inverse direction

pallets/swaps/src/lib.rs
```diff
@@ -190,24 +203,28 @@ pub mod pallet {
 								Ok((
 									SwapStatus {
 										swapped: new_swap_amount_in,
-										pending: T::Balance::zero(),
+										pending: T::BalanceOut::zero(),
```
wischli, line 206: Shouldn't this be `BalanceIn`? In this match arm, the inverse swap (`OrderBook<Out, In>` instead of `<In, Out>`) is larger. So the swapped amount is `Out` but the pending should be `In`, right?

wischli (commented):
Really awesome application of a Rust feature I haven't used myself so far 🚀 Thanks for constantly pushing me forward and expanding my mind 🫶

I know this is an early draft. I have two questions.

pallets/swaps/src/lib.rs
```diff
@@ -273,36 +290,33 @@ pub mod pallet {
...
+				.and_then(T::OrderBook::<T::BalanceIn, T::BalanceOut>::get_order_details)
 				.filter(|swap| swap.currency_out == from_currency)
 				.map(|swap| swap.amount_out)
-				.unwrap_or_default())
+				.unwrap_or(Zero::zero()))
 		}
 
 		fn valid_pair(currency_in: Self::CurrencyId, currency_out: Self::CurrencyId) -> bool {
-			T::OrderBook::valid_pair(currency_in, currency_out)
+			T::OrderBook::<T::BalanceIn, T::BalanceOut>::valid_pair(currency_in, currency_out)
```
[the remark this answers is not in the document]
lemunozm: Good question! Because `valid_pair()` does not receive or return any balance (it's independent of the balance), you can basically put there what you want and it will compile and work.

Actually, because it's independent, the resulting patter indirectly "suggests" that a better design might be to move `valid_pair()` out of the trait, which semantically makes sense, I think.

pallets/swaps/src/lib.rs
lemunozm: It is seen from the swap perspective, where swapped is always "in" and pending is always "out".

Note that `new_swap_amount_in` comes from `convert_by_market()` which returns `BalanceIn`. I didn't specified the bounds there because it's automatically inferred by the parameters.

lemunozm:
Sadly I think this can not be used 😢. When configuring the GATs, they require a type to be parameterized. Nevertheless, our pallets are only parameterized by the Config trait through associated types. I can not parameterize a Pallet as `Pallet<Config, BalanceIn, BalanceOut>,` and then I can not initialize this later in the runtime. Maybe with an intermediate type, it can be achieved, but I think it's too much boilerplate.

wischli:
@lemunozm Do we want to close this? If GATs become in option in the future, we can revisit this.

lemunozm:
Yeah! We can close it!

PR #1737, opened 2024-02-19
Cleaning: Repo organization before Polkadot 1.1.0

lemunozm:
# Description

Some refactors to leave the codebase more "organized" with a view to the next Polkadot update.

## Changes and Descriptions

- [x] `cfg_traits::fees` module to its own file.
- [x] `cfg_traits::swaps` to wrap swapping related modules.
- [x] Removed unused dependencies
- [x] Remove some runtime reexports and cyclic dependencies that made runtime files a bit messy.
- [x] Remove some non-prelude runtime imports as `*`
- [x] `<runtime>/liquidity_pools.rs` move to `<runtime>/lib.rs` and common parts to `runtime_common`
- [x] `<runtime>/evm.rs` move to `<runtime>/lib.rs` and common parts to `runtime_common`
- [x] ~~`<runtime>/xcm.rs` move to `<runtime>/lib.rs` and common parts to `runtime_common`~~ => Too many complicated to factorize given the amount of runtime types.
  - Factorized only some types


runtime/centrifuge/src/xcm.rs
```diff
@@ -82,7 +76,7 @@ impl frame_support::traits::Contains<RuntimeCall> for SafeCallFilter {
 				| RuntimeCall::Proxy(..)
 				| RuntimeCall::LiquidityPoolsGateway(
 					pallet_liquidity_pools_gateway::Call::process_msg { .. }
-				) // TODO: Enable later: | RuntimeCall::OrderBook(..)
+				) | RuntimeCall::OrderBook(..)
```
lemunozm, line 79: Already fixed in this PR

node/src/chain_spec.rs
```diff
@@ -680,26 +680,29 @@ fn centrifuge_genesis(
...
 ) -> altair_runtime::GenesisConfig {
 	let chain_id: u32 = id.into();
 
 	endowed_accounts.extend(endowed_evm_accounts.into_iter().map(|(addr, id)| {
 		let chain_id = id.unwrap_or_else(|| chain_id.into());
-		AccountConverter::<centrifuge_runtime::Runtime, centrifuge_runtime::LocationToAccountId>::convert_evm_address(chain_id, addr)
+		AccountConverter::<
+			altair_runtime::Runtime,
+			altair_runtime::xcm::LocationToAccountId,
+		>::convert_evm_address(chain_id, addr)
```
lemunozm, line 697: These were wrong. When adding a new node to the network, does the node initialize its configuration with this genesis?

libs/traits/src/fees.rs
```diff
@@ -0,0 +1,90 @@
...
+pub trait FeeKey:
+	FullCodec + TypeInfo + MaybeSerializeDeserialize + sp_std::fmt::Debug + Clone + PartialEq
+{
+}
+
+impl<
+		T: FullCodec + TypeInfo + MaybeSerializeDeserialize + sp_std::fmt::Debug + Clone + PartialEq,
+	> FeeKey for T
+{
+}
```
wischli, line 16: Formatting seems off

wischli (approved):
Really nice cleanup! Thanks so much for this 🫶

node/src/chain_spec.rs
wischli: This section does not effect Altair but only any Altair-based chain spec from now on, e.g. Altair local or a theoretical Altair-based testnetwork which we might create in the future. 

This function alters storage and thus leads to a different genesis hash. Since we haven't used any converter for the genesis Altair, there is nothing to worry. Thanks for pointing this out and fixing though!

libs/traits/src/fees.rs
lemunozm: The line is not small enough to not fit in the same line as `<` but not big enough to be split in several lines. 😆

node/src/chain_spec.rs
lemunozm: Thanks for the explanation!

PR #1739, opened 2024-02-20
Loans: linear accrual on settlement price

### LINKED ISSUE #1738: Feat: Linear Accrual on Settlement Price
### Description
Currently, when we are using the `ExternalPricing` we are allowing the `last_settlement_price` to be used as a base for valuation in cases where there is no oracle, or the oracle itself is outdated. 

As an intermediate step between the current logic and [loans v4](https://centrifuge.hackmd.io/1OXnuqyBRIuZK9iNozuFvA?view) the goal is to have an estimate of the value of an asset based on the `settlement_price`, `time_till_maturity` and the `notinal`. 

See calculation example for formula: 
image (3)

### Research/based on
* Issuer needs to update asset price valuations without needing to submit oracles continuously

* Increase complexity

* Migration of loans needed (*probably*) IF we need to store the last time updated. But I think we can get around that by just calculating the value dynamically and not iteratively

lemunozm:
# Description

Fixes #1738

## Changes and Descriptions

- [x] Add `current_price` to RuntimeAPI loans portfolio for cases where the loan is external priced.
- [x] Update present value computation to use the current price instead of settlement price. (This present value affects portfolio computations)
- [x] Avoid erroring out in case `T::PriceRegistry::collection()` fails due to outdated collection.
- [x] ~Add UT for this case~ (Not added but fixed current failing test because of these changes)
- [x] Migration for the addition of `settlement_price_updated` field.

### DIFF
--- a/runtime/altair/src/migrations.rs
+++ b/runtime/altair/src/migrations.rs
@@ -81,6 +81,13 @@ pub type UpgradeAltair1034 = (
 		crate::Runtime,
 		AnnualTreasuryInflationPercent,
 	>,
+	// Bump balances storage version from v0 to v1 and mark balance of CheckingAccount as inactive,
+	// see https://github.com/paritytech/substrate/pull/12813
+	pallet_balances::migration::MigrateToTrackInactive<super::Runtime, super::CheckingAccount, ()>,
+	// Assets were already migrated to V3 MultiLocation but version not increased from 0 to 2
+	runtime_common::migrations::increase_storage_version::Migration<crate::OrmlAssetRegistry, 0, 2>,
+	// Data was already moved but storage version not increased from 0 to 4
+	runtime_common::migrations::increase_storage_version::Migration<crate::Council, 0, 4>,
 );
 
 #[allow(clippy::upper_case_acronyms)]
@@ -388,7 +395,8 @@ mod ausd_to_foreign {
 	#[cfg(feature = "try-runtime")]
 	use frame_support::ensure;
 	use frame_support::{
-		pallet_prelude::ValueQuery, storage_alias, Blake2_128Concat, Twox64Concat,
+		pallet_prelude::ValueQuery, storage::unhashed, storage_alias, Blake2_128Concat,
+		StoragePrefixedMap, Twox64Concat,
 	};
 	use orml_tokens::AccountData;
 	use parity_scale_codec::{Decode, Encode};
@@ -396,6 +404,7 @@ mod ausd_to_foreign {
 	use sp_runtime::traits::Zero;
 	#[cfg(feature = "try-runtime")]
 	use sp_runtime::DispatchError;
+	use sp_runtime::Saturating;
 	use sp_std::vec::Vec;
 
 	use super::*;
@@ -533,15 +542,50 @@ mod ausd_to_foreign {
 			);
 
 			TotalIssuance::<T>::remove(DEPRECATED_AUSD_CURRENCY_ID);
+
+			// Remove undecodable storage entries
+			let (reads, writes) = remove_undecodable_storage_keys::<T>();
+			log::info!("{LOG_PREFIX} Removed {writes} undecodable storage entries from Accounts");
+
 			log::info!("{LOG_PREFIX} Done");
 
 			// Approximate weight given for every entry migration there are two calls being
 			// made, so counting the reads and writes for each call.
 			<Runtime as frame_system::Config>::DbWeight::get().reads_writes(
-				migrated_entries.saturating_mul(5),
-				migrated_entries.saturating_mul(4).saturating_add(1),
+				migrated_entries.saturating_mul(5).saturating_add(reads),
+				migrated_entries
+					.saturating_mul(4)
+					.saturating_add(writes.saturating_add(1)),
 			)
 		}
+	}
+
+	fn remove_undecodable_storage_keys<T: orml_tokens::Config>() -> (u64, u64) {
+		let prefix = orml_tokens::Accounts::<T>::final_prefix();
+		let mut previous_key = prefix.clone().to_vec();
+		let mut reads: u64 = 1;
+		let mut writes: u64 = 0;
+		while let Some(next) =
+			sp_io::storage::next_key(&previous_key).filter(|n| n.starts_with(&prefix))
+		{
+			reads.saturating_accrue(1);
+			previous_key = next;
+			let maybe_value = unhashed::get::<OldCurrencyId>(&previous_key);
+			match maybe_value {
+				Some(_) => continue,
+				None => {
+					log::info!(
+						"Removing Account key which could not be decoded at {:?}",
+						previous_key
+					);
+					unhashed::kill(&previous_key);
+					writes.saturating_accrue(1);
+					continue;
+				}
+			}
+		}
+
+		(reads, writes)
 	}
 }
 
--- a/runtime/centrifuge/src/migrations.rs
+++ b/runtime/centrifuge/src/migrations.rs
@@ -39,9 +39,6 @@ frame_support::parameter_types! {
 }
 
 pub type UpgradeCentrifuge1025 = (
-	// Burns tokens from other domains that are falsly not burned when they were transferred back
-	// to their domain
-	burn_unburned::Migration<super::Runtime>,
 	runtime_common::migrations::epoch_execution::Migration<super::Runtime>,
 	// Migrates the currency used in `pallet-transfer-allowlist` from our global currency to a
 	// special filter currency enum
@@ -54,13 +51,16 @@ pub type UpgradeCentrifuge1025 = (
 		LocalCurrencyIdUsdc,
 	>,
 	// Register new canonical USDC on Celo
-	runtime_common::migrations::update_celo_usdcs::Migration<super::Runtime>,
-	// Init local representation for all assets
+	runtime_common::migrations::update_celo_usdcs::AddNewCeloUsdc<super::Runtime>,
+	// Update custom metadata by initiating local representation for all assets
 	runtime_common::migrations::local_currency::translate_metadata::Migration<
 		super::Runtime,
 		UsdcVariants,
 		LocalAssetIdUsdc,
 	>,
+	// Change name and symbol of Celo Wormhole USDC
+	// NOTE: Needs to happen after metadata translation because expects new CustomMetadata
+	runtime_common::migrations::update_celo_usdcs::UpdateWormholeUsdc<super::Runtime>,
 	// Switch pool currency from Polkadot USDC to Local USDC
 	runtime_common::migrations::local_currency::migrate_pool_currency::Migration<
 		super::Runtime,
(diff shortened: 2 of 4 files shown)


pallets/loans/src/entities/loans.rs
```diff
@@ -533,31 +531,42 @@ pub struct ActiveLoanInfo<T: Config> {
 
 	/// Current outstanding interest of this loan
 	pub outstanding_interest: T::Balance,
+
+	/// Current price for external loans
+	/// - If oracle set, then the price is the one coming from the oracle,
+	/// - If not set, then the price is a linear accrual using the latest
+	///   settlement price.
+	/// See [`ExternalActivePricing::current_price()`]
+	pub current_price: Option<T::Balance>,
```
mustermeiszer, line 540: I think we could use the valuation method of the internal pricing here too. Removing the Option. ANd I would name it current_value.
 

cc: @denniswell

lemunozm: Valuation method is just the present value for internal pricing loans. So we should do the same for external pricing loans too if go with this change. 

This is related to my question:

> Should we modify present_value() to now use current_price instead of latest_settlement_price?

For external pricing loans `present_value = quantity * latest_settlemente_price`. So maybe we do not need to use another field here and just update how `present_value` is computed using now `current_price` instead of `latest_settlement_price`

mustermeiszer: Ah sorry, my bad! I think that is the best idea!

lemunozm: Maybe we still want to put `current_price` here to allow the UI just to draw the price. Otherwise, they should divide the present_value / outstanding_quantity to obtain it.

mustermeiszer: I think that is confusing more ATM. The total value is anyways of greatest interest.

lemunozm: Ok, then we should document that to draw the current price used for a loan, the UI should do: `present_value / outstanding_quantity` for cases were oracles are not set instead of using directly `latest_settlement_price` as before.

lemunozm: cc @onnovisser

lemunozm: I just realized that `present_value` is always affected by how much the loan has been written down. Maybe it's quite difficult to extract the "current price" for Apps.

pallets/loans/src/lib.rs
```diff
@@ -1046,7 +1046,7 @@ pub mod pallet {
 			pool_id: T::PoolId,
 		) -> Result<(T::Balance, u32), DispatchError> {
 			let rates = T::InterestAccrual::rates();
-			let prices = T::PriceRegistry::collection(&pool_id)?;
+			let prices = T::PriceRegistry::collection(&pool_id).unwrap_or_default();
```
lemunozm, line 1049: @mustermeiszer just doing this, we force to have an outdated collection to be empty, and later (comment below)...

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -152,17 +185,25 @@ impl<T: Config> ExternalActivePricing<T> {
...
 	where
 		Prices: DataCollection<T::PriceId, Data = PriceOf<T>>,
 	{
 		let price = match cache.get(&self.info.price_id) {
 			Ok(data) => data.0,
-			Err(_) => self.latest_settlement_price,
+			Err(_) => self.linear_accrual_price(maturity)?,
 		};
 		Ok(self.outstanding_quantity.ensure_mul_int(price)?)
 	}
```
lemunozm, line 209: ...the externally priced loan will fail to obtain the "real value" and will generate the missing price for itself using the linear accrual method.

Later, faking or not, the `update_portfolio()` with a list of prices is independent of this process.

pallets/loans/src/entities/loans.rs
mustermeiszer: I think all that is relevant mostly is the value anyways.

pallets/loans/src/entities/pricing/external.rs
mustermeiszer: But this way we can not fake the oracle prices, in cases where we want to simulate the nav execution with new oracles. And this would mean closing an epoch is not protected if the oracle is outdated.

pallets/loans/src/lib.rs
mustermeiszer: But this enables closing an epoch with outdated oracles, which is not wanted.

mustermeiszer: why not use the `prices` as an parameter to `fn update_portfolio_valuation_for_pool `?

pallets/loans/src/entities/pricing/external.rs
lemunozm: > But this way we can not fake the oracle prices

We can because when we fake them, the value will come from the `Ok()` branch (line 205).

pallets/loans/src/lib.rs
lemunozm: Yes, that's the idea of allowing fake Oracle values.

lemunozm: In this change, I just avoid errors. But the idea (maybe in another PR) is to modify the PoolNAV trait passing prices as a parameter

pallets/loans/src/entities/loans.rs
lemunozm: Regarding this topic, I would say that if Apps needs to draw the price, leave the `current_price` as it is. Otherwise, if they need it and we do not offer it by the RuntimeAPI, then they need to perform not-trivial computations, and future changes to present_value will break that.

@onnovisser are you drawing right now the `settlementPrice` of an external loan?

lemunozm:
I was thinking that maybe it could be useful to pass `prices` to `update_portfolio()` as follows:

```rust
enum Prices {
    Custom(Collection), // Pass a custom collection with custom prices.
    Registry, // Use the collection from the registry (it means from oracles)
    LinearAccrual, // Use prices computed from `settlement_prices` with linear accrual. It can be useful if the collection is outdated and you want to pass "orientative" prices.
}
```

WDYT?

mustermeiszer:
I would pass prices and stuff to a different api. `simulate_epch_execution()` as we also need to close the epoch in order to make this fully useful for the ui. Also the return-type is not just the nav, but a tuple of things, like tranche states, fulfillments, prices, etc.

lemunozm:
I am not following. Do you mean to add a new extrinsic for faking prices? And leave as it is `update_portfolio_valuation()` extrinsic?

mustermeiszer:
> I am not follo ing. Do you mean to add a new extrinsic for faking prices? And leave the update_portfolio_valuation() extrinsic as it is

no, I meant another runtime-api.

pallets/loans/src/entities/loans.rs
mustermeiszer: @lemunozm can we remove that one here?

pallets/loans/src/lib.rs
```diff
@@ -433,6 +435,13 @@ pub mod pallet {
 		}
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_runtime_upgrade() -> frame_support::weights::Weight {
+			migrations::migrate_from_v2_to_v3::<T>()
+		}
+	}
+
```
mustermeiszer, line 444: Can we have this as we usually do and make this a struct that is feed manually to the runtime?

mustermeiszer: But no blocker.

mustermeiszer: Just hard to see what is already getting migrated.

mustermeiszer (commented):
Two changes I would like to see, but not blocking. 

Will the adaption of the runtime-api to not error out happen in another PR?

pallets/loans/src/lib.rs
lemunozm: I did so because it seemed to be the canonical way to migrate things that belong strictly to the internals of the pallet.

Actually, there is an annoying log message in the node that tells you that loans are not migrated when it is done outside the `hooks`. Anyway, I have no strong opinion regarding this

pallets/loans/src/entities/loans.rs
lemunozm: Can we decide after getting feedback from https://kflabs.slack.com/archives/C04GARZ532T/p1709024010720729 ?

pallets/loans/src/lib.rs
wischli: I haven't seen a migration triggered this way yet. So I would prefer going the typical `OnRuntimeUpgrade` path such we can ensure storage can be decoded properly beforehand and afterwards.

wischli (commented):
As always, looks really clean! I would like the migration to be done via `OnRuntimeUpgrade` instead of the current `hooks` approach.

pallets/loans/src/lib.rs
lemunozm: 2 vs 1. Then I change it for sure 😆

For reference, I saw this here in the official docs here: https://docs.substrate.io/reference/how-to-guides/storage-migrations/basic-storage-migration/

wischli: Thanks for the reference! Honestly, it's a simple migration. Just keep as is and we can move on.

libs/utils/src/lib.rs
```diff
@@ -112,6 +112,101 @@ pub fn decode_var_source<const EXPECTED_SOURCE_ADDRESS_SIZE: usize>(
 	}
 }
 
+pub mod math {
```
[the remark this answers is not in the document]
lemunozm: It was a fun time hitting my head with maths 😄 . Thanks William!

pallets/loans/src/entities/pricing/external.rs
lemunozm: Well, it affects. Because now, when there are no oracles, the price is the linear accrual instead of `settlement_price`. Ideally, after this change, we should do a `portfolio_update()` to consolidate the changes in the present value affected by this.

We could even add it as part of the migration. WDYT @mustermeiszer?

pallets/loans/src/lib.rs
lemunozm: 👍🏻

pallets/loans/src/entities/pricing/external.rs
lemunozm: Not sure if it is for PoolFees. But it can be for `pallet-loans` because we are modifying how the present_value is computed but we are not updating the portfolio with this new computation

pallets/loans/src/lib.rs
lemunozm: FYI, the above way generates a first-line log in the following way by `frame-support`:

```
2024-02-26 09:15:02.252  INFO main runtime::frame-support: ⚠️ Loans declares internal migrations (which *might* execute). On-chain `StorageVersion(2)` vs current storage version `StorageVersion(3)`
2024-02-26 09:15:02.252  INFO main pallet_loans::migrations: Loans: Starting migration v3
2024-02-26 09:15:02.252  INFO main pallet_loans::migrations: Loans: Migrated 1 pools
```

pallets/loans/src/entities/pricing/external.rs
lemunozm: > But it can be for pallet-loans because we are modifying how the present_value() is computed but we are not updating the portfolio with this new computation

Finally added. If finally you think we shouldn't, I can revert it. Commit: [b760ad6](https://github.com/centrifuge/centrifuge-chain/pull/1739/commits/b760ad6c1fcea264a11e206d3b7d66018a01f257)

pallets/loans/src/entities/loans.rs
lemunozm: Onno is out, which is who is taking care of it, I think, so probably we do not have feedback soon regarding this.

Seeing the apps code, they need to fetch the price as `current_price` now to fulfill this part: https://github.com/centrifuge/apps/blob/beea676a936a14597f270e931387a4df1ce7ca0d/centrifuge-app/src/utils/getLatestPrice.ts#L10 which is later used to draw in the loan page. 

I would say we need to keep this here. Do you have any argument against this, @mustermeiszer? I do not see you very in favor of adding it.

mustermeiszer (approved):

wischli (approved):

lemunozm:
Thanks! I guess you then agree with the two open threads @mustermeiszer

Issue #1495, opened 2023-08-14
[Altair] Mint treasury block rewards 


wischli:
### Description

* Mint 3% p.a. block rewards into Altair Treasury instead of none
* Burn 15,752,730 AIR (equivalent of annual Treasury block rewards)
* MUST be part of the next release

### Research/based on

https://gov.centrifuge.io/t/rfc-altair-block-rewards-activation-and-treasury-burning-proposal/5503

### How will this affect the code base

* Less maintenance: Block rewards will then mirror between Altair and Centrifuge

### What are foreseen obstacles or hurdles to overcome?

* Burning of TBD: Probably handled in democracy, not in migration
* Align final numbers with DAO

wischli:
Addition:

> AIR that are in the [modlcb/bridg](https://altair.subscan.io/account/kAKdGAcqfxWX23TwKtMUyobLEnDDzftMtJrfZppeUyLt5vs59) wallet, [should be burned] along with the portion that will burned from the Altair treasury, in an upcoming RU.
These tokens were airdropped back in July 2021 to the Centrifuge bridge wallet (all CFG holders received AIR in the ratio 1:1)

cc @Rhanok09 @ImdioR

ImdioR:
Yes, sounds good!

PR #1740, opened 2024-02-21
feat: relative treasury inflation

wischli:
# Description

Fixes https://github.com/centrifuge/centrifuge-chain/issues/1495

The current treasury inflation is based on absolute numbers and thus diverging from the annual 3% on which the community decided both for Altair ([-> RFC](https://gov.centrifuge.io/t/rfc-altair-block-rewards-activation-and-treasury-burning-proposal/5503)) as well as Centrifuge ([-> CP6](https://gov.centrifuge.io/t/rfc-updating-cp6-block-rewards/5839)). This PR fills this gap by switching to a relative treasury inflation. As a result, all three runtimes have the exact same block rewards configuration now except for different collator reward amounts.

## Now

Inside `pallet-block-rewards`, the annual treasury inflation rate is stored. On each new session, collators receive their reward. Moreover, the treasury is funded with the proration based on the configured rate and session duration in seconds:

```
treasury_inflation_per_session = annual_treasury_rate / SECONDS_PER_YEAR * session_duration_in_secs
session_duration_in_secs  = now - last_update
```

Before, the rewards configuration revolved around an absolute `total_amount` per session which was first distributed to the collators and the remainder minted into the treasury, if configured in the runtime (true for Centrifuge, false for Altair). For Altair, we also need to burn 15,752,730 AIR (equivalent of annual Treasury block rewards) which can be batched with the runtime upgrade proposal

# Checklist:

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -46,7 +46,10 @@ use sc_chain_spec::{ChainSpecExtension, ChainSpecGroup};
 use sc_service::{ChainType, Properties};
 use serde::{Deserialize, Serialize};
 use sp_core::{crypto::UncheckedInto, sr25519, Encode, Pair, Public, H160};
-use sp_runtime::traits::{IdentifyAccount, Verify};
+use sp_runtime::{
+	traits::{IdentifyAccount, Verify},
+	FixedPointNumber,
+};
 use xcm::{
 	latest::MultiLocation,
 	prelude::{GeneralIndex, GeneralKey, PalletInstance, Parachain, X2, X3},
@@ -61,6 +64,7 @@ pub type DevelopmentChainSpec =
 
 use altair_runtime::evm::AltairPrecompiles;
 use centrifuge_runtime::evm::CentrifugePrecompiles;
+use cfg_types::fixed_point::Rate;
 use development_runtime::evm::DevelopmentPrecompiles;
 
 /// Helper function to generate a crypto pair from seed
@@ -661,7 +665,11 @@ fn centrifuge_genesis(
 				.map(|(acc, _)| acc)
 				.collect(),
 			collator_reward: 8_325 * MILLI_CFG,
-			total_reward: 10_048 * CFG,
+			treasury_inflation_rate: Rate::saturating_from_rational(3, 100),
+			last_update: std::time::SystemTime::now()
+				.duration_since(std::time::UNIX_EPOCH)
+				.expect("SystemTime before UNIX EPOCH!")
+				.as_secs(),
 		},
 		block_rewards_base: Default::default(),
 		base_fee: Default::default(),
@@ -756,7 +764,11 @@ fn altair_genesis(
 				.map(|(acc, _)| acc)
 				.collect(),
 			collator_reward: 98_630 * MILLI_AIR,
-			total_reward: 98_630 * MILLI_AIR * 100,
+			treasury_inflation_rate: Rate::saturating_from_rational(3, 100),
+			last_update: std::time::SystemTime::now()
+				.duration_since(std::time::UNIX_EPOCH)
+				.expect("SystemTime before UNIX EPOCH!")
+				.as_secs(),
 		},
 		block_rewards_base: Default::default(),
 		collator_allowlist: Default::default(),
@@ -939,7 +951,11 @@ fn development_genesis(
 				.map(|(acc, _)| acc)
 				.collect(),
 			collator_reward: 8_325 * MILLI_CFG,
-			total_reward: 10_048 * CFG,
+			treasury_inflation_rate: Rate::saturating_from_rational(3, 100),
+			last_update: std::time::SystemTime::now()
+				.duration_since(std::time::UNIX_EPOCH)
+				.expect("SystemTime before UNIX EPOCH!")
+				.as_secs(),
 		},
 		base_fee: Default::default(),
 		evm_chain_id: development_runtime::EVMChainIdConfig {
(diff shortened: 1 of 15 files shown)


pallets/block-rewards/src/mock.rs
```diff
@@ -161,7 +153,7 @@ orml_traits::parameter_type_with_key! {
 }
 
 impl orml_tokens::Config for Test {
-	type Amount = i64;
+	type Amount = IBalance;
```
lemunozm, line 156: I would say using here `i64` is preferible (or `type IBalanace = i64` on top of this file) instead of using `cfg_primitives::IBalance`. Actually, `cfg_primitive::IBalance` should not be a "primitive", it's our centrifuge domain type used to configure our runtimes, so it is at the runtime level. From pallets, we should not use "runtime things". These types should probably be moved to some runtime common crate in the future.

Currently it's a mess. And primitives/utilities/and types are quite mixed. No CR here. Just a comment!

pallets/block-rewards/src/mock.rs
```diff
@@ -237,35 +229,38 @@ frame_support::parameter_types! {
...
+	type TreasuryPalletId = TreasuryPalletId;
 	type Weight = u64;
 	type WeightInfo = ();
 }
 
 pub(crate) fn assert_staked(who: &AccountId) {
 	assert_eq!(
 		// NOTE: This is now the ED instead of 0, as we collators need ED now.
-		<Test as Config>::Currency::balance(<Test as Config>::StakeCurrencyId::get(), who),
+		<Test as Config>::Tokens::balance(<Test as Config>::StakeCurrencyId::get(), who),
```
lemunozm, line 259: Super NIT:
```suggestion
		Tokens::balance(BlockRewardCurrency::get(), who),
```

in some occurrences.

pallets/block-rewards/src/migrations.rs
```diff
@@ -10,144 +10,287 @@
...
 
-impl<T, CollatorReward, TotalReward> OnRuntimeUpgrade
-	for InitBlockRewards<T, CollatorReward, TotalReward>
-where
-	T: frame_system::Config + Config + pallet_collator_selection::Config,
-	<T as Config>::Balance: From<u128>,
-	CollatorReward: Get<u128>,
-	TotalReward: Get<u128>,
-{
+pub mod init {
```
lemunozm, line 34: I do not see this migration used. Is it maintained in case we want to initialize another block reward pallet for a runtime?

lemunozm (approved):
I didn't see anything weird. Good job here William! I left mostly (super) NITs and some questions.

pallets/block-rewards/src/migrations.rs
wischli: Yes, exactly!

pallets/block-rewards/src/mock.rs
wischli: Unfortunately, using `i64` leads to overflows! Will use `i128` then.

node/src/chain_spec.rs
```diff
@@ -661,7 +665,11 @@ fn centrifuge_genesis(
 				.map(|(acc, _)| acc)
 				.collect(),
 			collator_reward: 8_325 * MILLI_CFG,
-			total_reward: 10_048 * CFG,
+			treasury_inflation_rate: Rate::saturating_from_rational(3, 100),
+			last_update: std::time::SystemTime::now()
+				.duration_since(std::time::UNIX_EPOCH)
+				.expect("SystemTime before UNIX EPOCH!")
+				.as_secs(),
```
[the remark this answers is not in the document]
wischli: Yes, thats the implication. However, this only affects new networks, i.e. either Development if it needs to be reset or otherwise local ones. Therefore, it shouldn't have an impact. If we make it static, the first treasury inflation reward will huge.

pallets/block-rewards/src/mock.rs
lemunozm: Oh, that's weird. I guess it's due to the use of amounts with a lot of decimals, right?

pallets/block-rewards/src/lib.rs
```diff
@@ -404,18 +400,25 @@ impl<T: Config> Pallet<T> {
...
+						let _ = <T::Tokens as FungibleMutate<T::AccountId>>::mint_into(
+							&T::TreasuryPalletId::get().into_account_truncating(),
+							treasury_inflation,
+						)
+						.map_err(|e| {
+							log::error!(
+								"Failed to mint treasury inflation for session due to error {:?}",
+								e
+							)
+						});
```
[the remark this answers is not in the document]
lemunozm: Great then! 👍🏻

node/src/chain_spec.rs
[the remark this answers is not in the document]
lemunozm: But you're using the system clock, not the clock used on-chain (that changed each block by 12 secs). I guess we're okay with it.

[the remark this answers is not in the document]
wischli: Good point but it should be fine here. It only affects the first treasury reward and should never affect live chains/production.

pallets/block-rewards/src/mock.rs
wischli: Yes, if I use CFG denomination, then it quickly creates an overflow

lemunozm (approved):
Not sure if Frederik wants to also review this, but LGTM!

lemunozm (approved):

PR #1741, opened 2024-02-21
Cleaning: Remove `TradingPair`

lemunozm:
# Description

Removes `TradingPair` storage and removes all effects on it in the codebase. [slack thread](https://kflabs.slack.com/archives/C05TBFRBL15/p1708422983878459)

## Changes and Descriptions

- (**API change**) Remove `TradingPair` storage from `order-book`
- (**API change**) Remove `add_trading_pair()` and `rm_trading_pair()` extrinsics.
- Remove `valid_pair()` method from `TokenSwaps` and `Swaps` traits.
- (**API change**) `allow_investment_currency()` extrinsic without `tranche_id` parameter.
- (**API change**) `disallow_investment_currency()` extrinsic without `tranche_id` parameter.


lemunozm:
Blocked by https://github.com/centrifuge/centrifuge-chain/pull/1737

wischli (changes requested):
Getting cleaner and cleaner here! Looks good except for the call index overwriting.

pallets/foreign-investments/src/impls.rs
```diff
@@ -205,23 +205,11 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
...
 
 	fn accepted_payment_currency(investment_id: T::InvestmentId, currency: T::CurrencyId) -> bool {
-		if T::Investment::accepted_payment_currency(investment_id, currency) {
-			true
-		} else {
-			T::PoolInspect::currency_for(investment_id.of_pool())
-				.map(|pool_currency| T::Swaps::valid_pair(pool_currency, currency))
-				.unwrap_or(false)
-		}
+		T::Investment::accepted_payment_currency(investment_id, currency)
```
mustermeiszer, line 208: **CR**

This is not correct. Now all payment currency are accepted. The above does simply check whether the currency given is the pool currency. 

cc @wischli

pallets/foreign-investments/src/impls.rs
```diff
@@ -205,23 +205,11 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
...
 
 	fn accepted_payout_currency(investment_id: T::InvestmentId, currency: T::CurrencyId) -> bool {
-		if T::Investment::accepted_payout_currency(investment_id, currency) {
-			true
-		} else {
-			T::PoolInspect::currency_for(investment_id.of_pool())
-				.map(|pool_currency| T::Swaps::valid_pair(currency, pool_currency))
-				.unwrap_or(false)
-		}
+		T::Investment::accepted_payout_currency(investment_id, currency)
```
mustermeiszer, line 212: See above

pallets/order-book/src/lib.rs
```diff
@@ -448,58 +412,9 @@ pub mod pallet {
...
-				currency_out,
-			});
-
-			Ok(())
-		}
-
 		/// Set the market feeder for set market ratios.
 		/// The origin must be the admin origin.
-		#[pallet::call_index(6)]
+		#[pallet::call_index(4)]
```
mustermeiszer, line 417: Not the biggest fan of changing call indexes. Just for the future. I would just deprecate old ones as external integrations "might" rely on the value.

mustermeiszer (changes requested):
The changes look good mostly. The FI logic is not correct IIUC. OTherwise good to go.

pallets/foreign-investments/src/impls.rs
wischli: That was my point in the slack conversation because trading pairs are used as payment/payout gate keepers. I thought your consensus, that it would be fine.

pallets/order-book/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I think it doesn't matter because the index was also changed in the previous PR. Actually, most extrinsic names and signatures were changed in the previous PR, but they are not still on-chain.

If it's really an issue, then maybe we should look for all previous changes. Not sure how `call_index` affect here. Are we currently using their indexes?

lemunozm: Gotcha.

> external integrations

Are we using external integrations for `order-book`? If not are we free to change them? I say this because in the previous PRs, a lot of extrinsic changed, maybe we should do some extra work related to it. 

If it doesn't matter this time for `order-book`, then I will leave it as it is to keep them simple and sorted.

pallets/foreign-investments/src/impls.rs
lemunozm: I thought the same.

mustermeiszer: I think that is fine, but we must just alwaus return true. Not checking the `T::Investments`. Sorry, for the confusion

pallets/order-book/src/lib.rs
mustermeiszer: I am fine with changing

pallets/foreign-investments/src/impls.rs
lemunozm: Ah, ok, that makes sense! Good point 🙌🏻 

Then, maybe we can remove those methods directly. WDYT?

runtime/centrifuge/src/migrations.rs
```diff
@@ -63,37 +63,6 @@ pub type UpgradeCentrifuge1025 = (
...
-		UsdcArb,
-		LocalCurrencyIdUsdc,
-		MinOrderAmount,
-	>,
-	runtime_common::migrations::local_currency::add_bidirectional_trading_pair::Migration<
-		super::Runtime,
-		UsdcCelo,
-		LocalCurrencyIdUsdc,
-		MinOrderAmount,
-	>,
```
mustermeiszer, line 96: We still need a migration to kill the storage or trading pairs... WDYT @wischli ?

lemunozm: Makes sense 👍🏻

pallets/foreign-investments/src/impls.rs
lemunozm: Removed here: [e7b70c4](https://github.com/centrifuge/centrifuge-chain/pull/1741/commits/e7b70c411d8fea83311c0d3202a8abbdd45d8e3f)

pallets/liquidity-pools/src/lib.rs
```diff
@@ -721,7 +721,6 @@ pub mod pallet {
 		pub fn allow_investment_currency(
 			origin: OriginFor<T>,
 			pool_id: T::PoolId,
-			tranche_id: T::TrancheId,
```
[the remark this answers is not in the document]
mustermeiszer: @lemunozm can you ask @onnovisser @sophialittlejohn @jpangelle or @annamehr. Not sure how much this impacts the UI code.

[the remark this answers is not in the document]
lemunozm: I've added the whole braking list change in the PR description

runtime/centrifuge/src/migrations.rs
lemunozm: After another thought, I think it's safer to just reset completely `pallet-order-book`. We should close, of course, the current pending order before the RU.

wischli (commented):
LGTM but misses cleanup of the `TradingPair` storage. If you want, I can do that.

runtime/centrifuge/src/migrations.rs
wischli: I agree!

pallets/liquidity-pools/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Onno approves 👍🏻

lemunozm:
Just done here: [1200eaf](https://github.com/centrifuge/centrifuge-chain/pull/1741/commits/1200eaf4f849e2f4168a8a420c093fe61351dc17) (tested with running the node with try-runtime)

wischli (approved):
LGTM! Thanks again for all the cleanup work you have been doing! 5/5 stars please come back lol

PR #1746, opened 2024-02-27
Fix: FI swap correclty with asymmetric ratios

lemunozm:
# Description

# Description

The issue is described in this [thread](https://kflabs.slack.com/archives/C04GJQAM9P0/p1707298087515729)

> When we decrease, and there is an increasing pending swap,  we use the `convert_by_market()` to know upfront the amount we need to cancel to the inverse swap and distinguish from the amount we need to decrease the investment.
Nevertheless, later, inside the swap, `convert_by_market()` is called again in the inverse direction to perform the cancellation of the swap. This works fine. However, with asymmetric ratios (A -> B, ratio: 0.9; B -> A, ratio 0.95), it can lead to obtaining a different swap than expected. In the worst case, it would lead to a NoFund error if the amounts are very fit.

## How it works

Now `pallet-swaps` offer a new method `cancel_swap()` which allows to cancel partially a pending amount. This way, we can cancel an inverse swap (in its currency out denomination) before applying the new swap (in the other currency denomination). Saving us from using `convert_by_market()` in the opposite direction.

### DIFF
--- a/libs/traits/src/swaps.rs
+++ b/libs/traits/src/swaps.rs
@@ -150,8 +150,8 @@ pub trait Swaps<AccountId> {
 		swap: Swap<Self::Amount, Self::CurrencyId>,
 	) -> Result<SwapStatus<Self::Amount>, DispatchError>;
 
-	/// Cancel (partial or total) a swap. The amount should be expresed in the
-	/// same currency as the the currency_out of the pending amount.
+	/// Cancel a swap partially or completely. The amount should be expressed in
+	/// the same currency as the the currency_out of the pending amount.
 	/// - If there was no previous swap, it errors outs.
 	/// - If there was a swap with other currency out, it errors outs.
 	/// - If there was a swap with same currency out:
--- a/libs/traits/src/swaps.rs
+++ b/libs/traits/src/swaps.rs
@@ -150,6 +150,21 @@ pub trait Swaps<AccountId> {
 		swap: Swap<Self::Amount, Self::CurrencyId>,
 	) -> Result<SwapStatus<Self::Amount>, DispatchError>;
 
+	/// Cancel (partial or total) a swap. The amount should be expresed in the
+	/// same currency as the the currency_out of the pending amount.
+	/// - If there was no previous swap, it errors outs.
+	/// - If there was a swap with other currency out, it errors outs.
+	/// - If there was a swap with same currency out:
+	///   - If the amount is smaller, it decrements it.
+	///   - If the amount is the same, it removes the inverse swap.
+	///   - If the amount is greater, it errors out
+	fn cancel_swap(
+		who: &AccountId,
+		swap_id: Self::SwapId,
+		amount: Self::Amount,
+		currency_id: Self::CurrencyId,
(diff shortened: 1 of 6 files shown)


pallets/foreign-investments/src/tests.rs
```diff
@@ -1067,6 +1067,119 @@ mod investment {
...
+				});
+
+				util::fulfill_last_swap(Action::Investment, foreign_to_pool(3 * AMOUNT / 4));
+
+				assert_eq!(
+					ForeignInvestmentInfo::<Runtime>::get(&USER, INVESTMENT_ID),
+					None,
+				);
+				assert_eq!(ForeignInvestment::investment(&USER, INVESTMENT_ID), Ok(0));
+				assert_eq!(MockInvestment::investment(&USER, INVESTMENT_ID), Ok(0));
```
wischli, line 1122: Why does the foreign investment vanish when we fulfill only 3/4 of the order? Shouldn't there be 1/4 left?

pallets/foreign-investments/src/tests.rs
```diff
@@ -1067,6 +1067,119 @@ mod investment {
...
+					Action::Investment,
+					foreign_to_pool(3 * AMOUNT / 4) * MULTIPLIER,
+				);
+
+				assert_eq!(
+					ForeignInvestmentInfo::<Runtime>::get(&USER, INVESTMENT_ID),
+					None,
+				);
+				assert_eq!(ForeignInvestment::investment(&USER, INVESTMENT_ID), Ok(0));
+				assert_eq!(MockInvestment::investment(&USER, INVESTMENT_ID), Ok(0));
```
wischli, line 1180: Same question here.

wischli (commented):
I have a question about the new 2 FI tests. Code looks great as always!

pallets/foreign-investments/src/tests.rs
lemunozm: Note that we only fulfill the swap with the amount that was previously swapped in the opposite direction, which is 3/4. The 1/4 left has never been swapped, so when we decrease, such 1/4 is canceled.

pallets/foreign-investments/src/tests.rs
lemunozm: The same applies here

libs/traits/src/swaps.rs
```diff
@@ -150,6 +150,21 @@ pub trait Swaps<AccountId> {
 		swap: Swap<Self::Amount, Self::CurrencyId>,
 	) -> Result<SwapStatus<Self::Amount>, DispatchError>;
 
+	/// Cancel (partial or total) a swap. The amount should be expresed in the
```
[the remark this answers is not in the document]
lemunozm: Thanks!

pallets/foreign-investments/src/tests.rs
wischli: Ah yes of course, thanks for the explanation!

wischli (approved):
Clear as a crystal ball!

PR #1748, opened 2024-02-28
Loans: Runtime API for updating the portfolio with fake prices

lemunozm:
# Description

Add a Runtime API to emulate a portfolio update with custom prices.

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -73,9 +73,12 @@ pub use weights::WeightInfo;
 #[frame_support::pallet]
 pub mod pallet {
 	use cfg_traits::{
-		self, changes::ChangeGuard, data::DataRegistry, interest::InterestAccrual, IntoSeconds,
-		Permissions, PoolInspect, PoolNAV, PoolReserve, PoolWriteOffPolicyMutate, Seconds,
-		TimeAsSecs,
+		self,
+		changes::ChangeGuard,
+		data::{DataCollection, DataRegistry},
+		interest::InterestAccrual,
+		IntoSeconds, Permissions, PoolInspect, PoolNAV, PoolReserve, PoolWriteOffPolicyMutate,
+		Seconds, TimeAsSecs,
 	};
 	use cfg_types::{
 		adjustments::Adjustment,
@@ -84,7 +87,7 @@ pub mod pallet {
 	};
 	use entities::{
 		changes::{Change, LoanMutation},
-		input::{PrincipalInput, RepaidInput},
+		input::{PriceCollectionInput, PrincipalInput, RepaidInput},
 		loans::{self, ActiveLoan, ActiveLoanInfo, LoanInfo},
 	};
 	use frame_support::{
@@ -103,7 +106,7 @@ pub mod pallet {
 		traits::{BadOrigin, EnsureAdd, EnsureAddAssign, EnsureInto, One, Zero},
 		ArithmeticError, FixedPointOperand, TransactionOutcome,
 	};
-	use sp_std::{vec, vec::Vec};
+	use sp_std::{collections::btree_map::BTreeMap, vec, vec::Vec};
 	use types::{
 		self,
 		policy::{self, WriteOffRule, WriteOffStatus},
@@ -150,7 +153,7 @@ pub mod pallet {
 			+ One;
 
 		/// Identify a loan in the pallet
-		type PriceId: Parameter + Member + TypeInfo + Copy + MaxEncodedLen;
+		type PriceId: Parameter + Member + TypeInfo + Copy + MaxEncodedLen + Ord;
 
 		/// Defines the rate type used for math computations
 		type Rate: Parameter + Member + FixedPointNumber + TypeInfo + MaxEncodedLen;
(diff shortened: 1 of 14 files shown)


lemunozm:
Blocked by https://github.com/centrifuge/centrifuge-chain/pull/1739

pallets/loans/src/lib.rs
```diff
@@ -1051,11 +1057,33 @@ pub mod pallet {
...
+		pub fn update_portfolio_valuation_for_pool(
 			pool_id: T::PoolId,
+			input_prices: PriceCollectionInput<T>,
 		) -> Result<(T::Balance, u32), DispatchError> {
 			let rates = T::InterestAccrual::rates();
-			let prices = T::PriceRegistry::collection(&pool_id)?;
+			let prices = match input_prices {
+				PriceCollectionInput::Empty => BTreeMap::default(),
+				PriceCollectionInput::Custom(prices) => prices.into(),
+				PriceCollectionInput::FromRegistry => Self::registered_prices(pool_id)?,
```
mustermeiszer, line 1084: What is the benfit of using the `PriceCollectionInput` here vs just using a `BTreeMap` as the parameter?

lemunozm: You can see each use case here: https://github.com/centrifuge/centrifuge-chain/pull/1748/files#diff-efd52f7fa525763070479bd9d72c66286451a6752bed8534e83e1004e755c3e1R469

lemunozm: `Empty` is just sugar for `Custom([])` and could be removed. But the caller should choose between `Custom` and `FromRegistry`.

wischli: I prefer @lemunozm's proposal over just using a `BTreeMap` as it is more explicit which is always helpful for runtime API which can be expected to be used by externals as well.

mustermeiszer: Okay, you changed the valuation parameters of the runtime-api. Sounds good!

 The most important thing though would be to mimic closing an epoch with 
* faked prices
* Optional<Solution>

Returning:
* Fullfillment amounts for each tranche for both invest and redeem
* New pool-reserve
* New Tranche prices

pallets/loans/src/lib.rs
```diff
@@ -1208,7 +1236,8 @@ pub mod pallet {
 		}
 
 		fn update_nav(pool_id: T::PoolId) -> Result<T::Balance, DispatchError> {
-			Ok(Self::update_portfolio_valuation_for_pool(pool_id)?.0)
+			Self::update_portfolio_valuation_for_pool(pool_id, PriceCollectionInput::FromRegistry)
```
wischli, line 1239: A quick comprehension question: Our rather newer NAV calculation runtime API calls `Loans::update_nav(pool_id)` such that the Pool Fees NAV calculation can be based on that. Say the loans NAV is updated with a price collection input in block `i` and in the next block `i+1` we are querying the NAV via `PoolsApi::nav(pool_id)` which performs `Loans::update_nav(pool_id)` with an update based on `PriceCollectionInput::FromRegistry`. IIUC, the prices from the update in block `i` are still cached. What I want to  find out is if `Loans::update_nav` can make the Loans NAV less accurate if it was previously based on a custom price list.

wischli (commented):
Really clean runtime API! I have a question before I hit the approval button.

pallets/loans/src/lib.rs
lemunozm: Ok, I guess somehow this should be amalgamated with some other `pallet-pool` calls to get this, right?

Then I propose to create another in the PoolSystem RuntimeAPI, maybe called simulate_close_epoch or similar, where all these calls are performed to return all the required values above.

pallets/loans/src/lib.rs
lemunozm: It was previously based on the price registry, so both lines above are exactly identical. From a pool fees perspective, nothing has changed. The only open door (right now) is for the RuntimeAPI.

pallets/loans/src/lib.rs
lemunozm: Great, I'll do in another PR pool-related.

mustermeiszer (approved):
LGTM! Can you ping apps that the runtime api changed?

wischli (commented):
One minor nit change I would like to see: Runtime API can be versioned via `#[api_version($VERSION)]` as part of the `decl_runtime_apis!` macro ([example](https://github.com/KILTprotocol/kilt-node/blob/62a1373d20cae10a72e67973f5f6392052a6eaee/runtime-api/did/src/lib.rs#L69C2-L69C19)). However, we have not made use of this feature so far. Of course, we should to make the breaking changes explicit.

lemunozm:
Good point. Just one question: should we bump up the version even if the upgrade is retro-compatible? (it means just an addition)

lemunozm:
@mustermeiszer can we merge https://github.com/centrifuge/centrifuge-chain/pull/1739 before?

mustermeiszer (dismissed):

wischli:
> Good point. Just one question: should we bump up the version even if the upgrade is retro-compatible? (it means just an addition)

Definitely not a blocker because it's not breaking. In the future, I would still like to make real use of versioning because it can be tied to the earliest supported spec version for that API. 

For instance, SubQuery recently had issues with the new PoolsNAV API because it was attempted at blocks which didn't support it. But in order for this to fully work, Apps needs to add spec version granularity to the type and RtApi decorations.

wischli (dismissed):

lemunozm:
Good reasoning, you've convince me! => [24adcea](https://github.com/centrifuge/centrifuge-chain/pull/1748/commits/24adcea0fe32182aa7b12eadd6114b9513280247)

wischli (approved):

PR #1749, opened 2024-02-29
Release client v0.10.36, Centrifuge 1025 & Altair 1034 

wischli:
# Description

* Release candidate for upgrading both runtimes and the client to the latest version
* Draft notes: https://centrifuge.hackmd.io/70X1MDhESt2_ohn9SdSnGQ

## Changes
* Minor migration fixes
* Updates weights
* Splits `PoolFees` trait into `PoolFeesMutate` and `PoolFeesInspect` whereas the latter is only needed for weights

## TODO

- [x] Altair weights
- [x] Centrifuge weights
- [x] Dev weights

# Checklist:

- [x] I have added Rust doc comments to structs, enums, traits and functions
- [x] I have made corresponding changes to the documentation
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or that my feature works

PR #1745, opened 2024-02-26
add code coverage to pipeline

gpmayorga:
# Description

Add code coverage reports to our CI pipeline

Fixes #1617

## Changes and Descriptions

Use tarpaulin to produce coverage reports

PR #1750, opened 2024-02-29
fix: bug on oracle fetching

mustermeiszer:
# Description
The current oracle does check local representation of wrongly. 

## Changes and Descriptions
* Fix checking local representation of local to variant, instead of variant to local
* Add integration tests for the `OracleRatioProvider`

# Checklist:


mustermeiszer:
@wischli this must go in before #1749

runtime/integration-tests/src/generic/cases/oracles.rs
```diff
@@ -0,0 +1,125 @@
...
+
+		env.parachain_state_mut(|| {
+			assert_eq!(
+				orml_asset_registry::Metadata::<T>::get(LocalUSDC.id()),
+				None
+			);
+			assert_eq!(
+				orml_asset_registry::Metadata::<T>::get(DomainUSDC.id()),
+				None
+			);
```
mustermeiszer, line 91: @wischli @lemunozm these assertions are triggered. Why on earth should this storage be populated already? I either am looking at the reason without seeing it, or something is really off here.

lemunozm: This could be due to the migration being performed before initializing the tests? Our migration populates the state with those currencies IIRC

runtime/integration-tests/src/generic/utils/currency.rs
```diff
@@ -8,128 +8,158 @@ use cfg_types::{
...
+	fn symbol(&self) -> &'static str {
+		"USD6"
+	}
+
+	fn custom(&self) -> CustomMetadata {
+		CustomMetadata {
+			pool_currency: true,
+			transferability: CrossChainTransferability::Xcm(XcmMetadata {
+				fee_per_second: Some(1_000),
+			}),
```
lemunozm, line 92: Just wondering why `Usd6` has this property but `Usd12` and `Usd18` do not. Could we use the same for all of them?

runtime/integration-tests/src/generic/utils/currency.rs
```diff
@@ -8,128 +8,158 @@ use cfg_types::{
...
+
+	fn decimals(&self) -> u32 {
+		18
+	}
+
+	fn unit(&self) -> Balance {
+		10u128.pow(self.decimals())
+	}
+
+	fn symbol(&self) -> &'static str;
```
lemunozm, line 29: Most of the time, we do not care about the symbol for testing. Could we use a default method here with some random/empty string? Maybe "Unnamed" or so.

lemunozm (commented):
LGTM. I think the currencies testing API now is more versatile.

Maybe we could extend the integration-tests with some `pallet-token-mux` call to ensure the whole use case is working well

runtime/integration-tests/src/generic/cases/oracles.rs
mustermeiszer: Oh man. That could be. Haha. But then it is not working as intended. As the local representation is missing on asset 100_001. I will try to use different ids. 

Thanks @lemunozm!

runtime/integration-tests/src/generic/utils/currency.rs
mustermeiszer: I just copy pasted everything. Not sure, why it was there.

runtime/integration-tests/src/generic/utils/currency.rs
mustermeiszer: Agreed!

runtime/integration-tests/src/generic/cases/oracles.rs
wischli: But we are using id `100_0001` instead of `100_001` here. Compiling right now with my cherry-pick for disabling migrations in std: https://github.com/centrifuge/centrifuge-chain/pull/1745/commits/0e53752f0d491afb5b9c03f6a35202f706a183a7

mustermeiszer:
> Maybe we could extend the integration-tests with some pallet-token-mux call to ensure the whole use case is working well

I agree that this would be useful. But I do not have time for that. I am testing everything in demo at the moment. Thats also where I found this issue. I think the best would be to merge #1720 and then write integration test verifying the whole flow...

runtime/altair/src/migrations.rs
```diff
@@ -90,6 +91,9 @@ pub type UpgradeAltair1034 = (
 	runtime_common::migrations::increase_storage_version::Migration<crate::Council, 0, 4>,
 );
 
+#[cfg(feature = "std")]
+pub type UpgradeAltair1034 = ();
+
```
lemunozm, line 96: Maybe there exists a way to tune `RuntimeEnv` to avoid this issue from the tests? I do not like adding this addition to the runtime too much with the aim of avoiding migrations in testing. Future readers of this will not know why this is so, and we can be subject to unexpected behavior in the future. I would add a comment at least explain why this

lemunozm: Another option is tweaking the block creation in `RuntimeEnv` somehow to skip the migration. Not sure if this is something easy to do

lemunozm (changes requested):
Just want change that I think it's important to modify/clarify.
 
Everything else looks great!

runtime/altair/src/migrations.rs
wischli: Since this is my forced change which should not have been pushed to this branch in the first place, I will just revert it. However, it improves integration test time because it blocks migrations for each node we spawn. 

What unexpected behaviour are you fearing here @lemunozm? I think it is unexpected that migrations run for integration tests which is fixed this way. It's not the nicest solution but it does the job quickly without imposing any risk.

mustermeiszer: Fixed in variant_to_local_rate_set

runtime/integration-tests/src/generic/envs/runtime_env.rs
```diff
@@ -66,6 +67,14 @@ impl<T: Runtime> Env<T> for RuntimeEnv<T> {
 
 		let mut parachain_ext = sp_io::TestExternalities::new(parachain_storage);
 
+		// NOTE: Setting the current on-chain runtime version to the latest one, to
+		//       prevent running migrations
+		parachain_ext.execute_with(|| {
+			frame_system::LastRuntimeUpgrade::<T>::put(LastRuntimeUpgradeInfo::from(
+				<T as frame_system::Config>::Version::get(),
+			))
+		});
```
wischli, line 76: Awesome change, thanks for that!

runtime/integration-tests/src/generic/utils/currency.rs
wischli: Seems like we want to have default transferability here instead?

wischli (approved):
The `Usd6` still has XCM transferability enabled. Apart from that LGTM!

runtime/altair/src/migrations.rs
lemunozm: > However, it improves integration test time because it blocks migrations for each node we spawn.

Good point!

> What unexpected behaviour are you fearing here @lemunozm?

Could some uses of the node underlying it exist that compile the runtime as `std` and where do we  expect some kind of migration, maybe some migration testing or so? Probably not now but it scared me 🙏🏻

lemunozm:
TBH, I do not remember. Maybe a copy-paste from another place. I would put there what you think is cleaner/defaulted/simpler (what @wischli said)

lemunozm:
So we know now the reason 😆, we needed it for the proxy cross-change tests. 

I think this is a change added by @cdamian (who could explain the whys better) in the PR I opened about proxies, that's why git blame tell about me. Maybe we can create a currency in our own proxy tests for cross-chain things with the required metadata, using the new way of doing it. WDYT?

mustermeiszer:
Will force merge as @wischli approved and @lemunozm is off the next week. I anyways addressed all issues from Luis.

PR #1758, opened 2024-03-04
Feat: Enhance Error Resolution EVM

mustermeiszer:
# Description
Currently, the EVM does never error out if the execution of the EVM failed for some reason. In most case it will be an active `Revert` faced from some checks on the EVM side. 

But Substrate does not see that. It will only see an okay, resulting in wrong assumptions. 

NOTE: 
This change does mean that the EVM side DOES change but the Substrate side is still transactional. In most cases the EVM side will only contain fee-payment and new `Pending` appends.

## Changes and Descriptions
* Check last `Pending` to fetch status code of transaction

# Checklist:

### DIFF
--- a/pallets/ethereum-transaction/src/lib.rs
+++ b/pallets/ethereum-transaction/src/lib.rs
@@ -29,6 +29,7 @@ mod tests;
 
 #[frame_support::pallet]
 pub mod pallet {
+	use ethereum::ReceiptV3;
 	use frame_system::pallet_prelude::OriginFor;
 
 	use super::*;
@@ -61,6 +62,27 @@ pub mod pallet {
 				H256::from_low_u64_be(2u64),
 				H256::from_low_u64_be(2u64),
 			)
+		}
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		EvmExecutionFailed,
+	}
+
+	impl<T: Config> Pallet<T>
+	where
+		OriginFor<T>:
+			From<pallet_ethereum::Origin> + Into<Result<pallet_ethereum::Origin, OriginFor<T>>>,
+	{
+		fn valid_code(receipt: &ReceiptV3) -> bool {
+			let code = match receipt {
+				ReceiptV3::Legacy(inner)
+				| ReceiptV3::EIP2930(inner)
+				| ReceiptV3::EIP1559(inner) => inner.status_code,
+			};
+
+			code == 1
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


cdamian (dismissed):
Nice to finally have a way of seeing EVM execution errors.

pallets/ethereum-transaction/src/lib.rs
```diff
@@ -129,7 +151,21 @@ pub mod pallet {
...
+
+			// NOTE: The Ethereuem side of things never returns a DispatchError
+			//       if the execution failed. But we can check that manually by
+			//       querying the `Pending` storage of the pallet-ethereum.
+			let pending = pallet_ethereum::Pending::<T>::get();
+			let last = pending.last().ok_or(DispatchError::Other(
+				"Ethereuem not adding pending storage. Unexpected.",
+			))?;
+
+			if Pallet::<T>::valid_code(&last.2) {
```
wischli, line 164: Ultra nit: Always in favor of destructuring tuples to be aware of breaking changes to number of tuple elements.
```suggestion
			let (_, last_receipt, _) = pending.last().ok_or(DispatchError::Other(
				"Ethereuem not adding pending storage. Unexpected.",
			))?;

			if Pallet::<T>::valid_code(&last_receipt) {
```

wischli (dismissed):
LGTM! Minor nit on the possibility of deconstructing the `Pending` tuple. Certainly not a blocker. Thanks for dealing with this so quickly 🚀

pallets/ethereum-transaction/src/lib.rs
mustermeiszer: Done.

cdamian (dismissed):

wischli (approved):
Thanks for satisfying the pedantic of the team 🥇

mustermeiszer:
@cdamian any idea how you would fix the test without not calling `send` on the routers? Do you think we need a dummy contract or can we get around that?

cdamian:
> @cdamian any idea how you would fix the test without not calling `send` on the routers? Do you think we need a dummy contract or can we get around that?

@mustermeiszer we might need a dummy contract, but lemme grab this branch and take a closer look.

mustermeiszer:
Maybe we can mock the ethereum calling logic

cdamian:
https://github.com/centrifuge/centrifuge-chain/pull/1760 has a fix for the current test failures.

lemunozm (approved):
Cool change!

PR #1760, opened 2024-03-05
lp-gateway-routers: Remove value from test data

cdamian:


pallets/liquidity-pools-gateway/routers/src/tests.rs
```diff
@@ -28,739 +28,739 @@ lazy_static! {
...
+        pub fn get_test_data() -> EVMRouterTestData {
+            let test_contract_address = H160::from_low_u64_be(1);
+            let test_contract_code = [0; 32].to_vec();
+            let test_contract_hash = BlakeTwo256::hash_of(&test_contract_code);
+
+            let evm_domain = EVMDomain {
+                target_contract_address: test_contract_address,
+                target_contract_hash: test_contract_hash,
+                fee_values: FeeValues {
+                    value: U256::from(0),
```
cdamian, line 58: @mustermeiszer This value here is what caused the failure in the original PR. The EVM logic would try to transfer this amount from the sender to the derived contract address.

PR #1764, opened 2024-03-08
FI: Add swapping events

lemunozm:
# Description

- Add `SwapCreated` and `SwapFullfilled` events from the PoV of FI.

**Should be merged before RU**

### DIFF
--- a/pallets/foreign-investments/src/impls.rs
+++ b/pallets/foreign-investments/src/impls.rs
@@ -229,7 +229,7 @@ impl<T: Config> Pallet<T> {
 					..swap.clone()
 				},
 				cancelled_in: status.swapped,
-				pending_in: T::Swaps::pending_amount(who, swap_id, swap.currency_in)?,
+				opposite_in: T::Swaps::pending_amount(who, swap_id, swap.currency_in)?,
 			});
 		}
 
--- a/pallets/foreign-investments/src/lib.rs
+++ b/pallets/foreign-investments/src/lib.rs
@@ -292,7 +292,7 @@ pub mod pallet {
 			swap_id: SwapId<T>,
 			remaining: SwapOf<T>,
 			cancelled_in: T::SwapBalance,
-			pending_in: T::SwapBalance,
+			opposite_in: T::SwapBalance,
 		},
 	}
 }
--- a/pallets/foreign-investments/src/tests.rs
+++ b/pallets/foreign-investments/src/tests.rs
@@ -387,7 +387,7 @@ mod investment {
 						currency_in: FOREIGN_CURR,
 					},
 					cancelled_in: AMOUNT / 4,
-					pending_in: 3 * AMOUNT / 4,
+					opposite_in: 3 * AMOUNT / 4,
 				}
 				.into(),
 			);


pallets/foreign-investments/src/impls.rs
```diff
@@ -209,23 +213,64 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
...
+		status: &SwapStatus<T::SwapBalance, T::SwapRatio>,
+	) {
+		if !status.swapped.is_zero() {
+			Pallet::<T>::deposit_event(Event::SwapFullfilled {
+				who: who.clone(),
+				swap_id,
+				remaining: Swap {
+					amount_out: status.pending,
+					..swap.clone()
+				},
```
lemunozm, line 230: Note that `..swap.clone()` contains the `currency_in` and `currency_out` to know the swap direction.

lemunozm (commented):
@hieronx, I think not, but tell me if you think some information is missing in the events.

libs/traits/src/swaps.rs
```diff
@@ -112,25 +118,33 @@ pub struct SwapState<AmountIn, AmountOut, Currency> {
 	/// Amount of incoming currency already swapped denominated in outgoing
 	/// currency
 	pub swapped_out: AmountOut,
+
+	/// Ratio used to swap `swapped_out` into `swapped_in`
+	pub ratio: Ratio,
```
wischli, line 123: The `OrderInfo` returned by `TokenSwaps::get_order_details` contains a `ratio` field of type `OrderRatio<Ratio>`. Maybe that suffices such that we don't need to introduce an additional ratio field?

wischli (commented):
Quick like a spaceship! One comment regarding a potential redundancy. Moreover, it would be nice to assert the newly added events in a unit test.

lemunozm:
> Moreover, it would be nice to assert the newly added events in a unit test.

Agree!  I'll do it in a follow-up PR to not block this in case I have no time to finish them

lemunozm:
Modified to return always the market ratio in a fulfilled event even when it comes from a cancelation.

libs/traits/src/swaps.rs
wischli: Thanks for the explanation! Makes sense to keep as is then.

wischli (dismissed):
LGTM!

lemunozm:
We should not merge it yet. There are some requested changes from slack

pallets/foreign-investments/src/lib.rs
```diff
@@ -263,4 +268,31 @@ pub mod pallet {
...
+		},
+		// The swap was fulfilled by cancelling an opposite swap for the same foreign investment.
+		SwapCancelled {
+			who: T::AccountId,
+			swap_id: SwapId<T>,
+			remaining: SwapOf<T>,
+			cancelled_in: T::SwapBalance,
+			opposite_in: T::SwapBalance,
+		},
+	}
```
lemunozm, line 297: @hieronx I've needed to add a new `SwapCancelled`. All events are seen from the perspective of who did the swap. 

> Wouldn’t it possible to add an optional arg to the decrease fulfilled event, with the swap ID that was used to cancel it out?

This can not be possible due that swap ID is the same for both directions. Note that swap id is not the order id, which is unknown from FI perspective. As a workaround, I've added the `opposite_in` field to `SwapCancelled,` which has the amount in the opposite direction that has not been canceled. That way, if opposite_in is 0, that means that the cancellation has also canceled the opposite order. That is what you need to know.

hieronx: Looks good to me! Only one nit: in `SwapFulfilled`, wouldn't it make more sense to have the `swapped_out` param before the `swapped_in` param?

lemunozm: What do you mean by "before" in this context? Placed in swapped line numbers? Probably, it makes more sense. But right now, we have such conventions in all places in the codebase. I'd prefer to keep it as it is everywhere. But we should probably look into it soon 👍🏻

wischli:
> Just RC as I want to see before merge

Just triggered a WASM generation and can update Dev later.

pallets/foreign-investments/src/lib.rs
hieronx: Well we should never change these event definitions again after merging this, at least not without a very strong reason ;) Would be a pain on the subquery side. But sure it is fine to keep it as is then if that is the convention now

pallets/foreign-investments/src/impls.rs
```diff
@@ -209,23 +213,66 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
...
+		if !status.pending.is_zero() {
+			Pallet::<T>::deposit_event(Event::SwapCreated {
+				who: who.clone(),
+				swap_id,
+				swap: Swap {
+					amount_out: status.pending,
+					..swap.clone()
+				},
+			})
+		}
```
mustermeiszer, line 245: Could it be that both `status.swapped` and `status.pending` are non zero? 

IF YES: 
* How would one interpret the `swap_id` of the two events?

mustermeiszer (commented):
Code looks good!

I am still a bit confused to follow the reasoning of the code, but the last time I had the same and found no issues. What would help me if @lemunozm could describe what are the possible states after `apply_swap`, how are they reflected in the `SwapStatus` and the `SwapOf`.

pallets/foreign-investments/src/impls.rs
lemunozm: Yes, it could. `swap_id` will be the same for both. Note that `swap_id` is just a tuple `(T::InvestmentId, Action)` not an `order_id` (which is a number from `order-book`).

If the above two events are dispatched, it means that when you apply a swap, part of the amount was automatically fulfilled by cancelation with an inverse order (notified by `SwapCancelled` event), and part of the swap could not be fulfilled with the inverse order: such remaining amount needs to be swapped (notified it by `SwapCreated` event).

Possible states after applying swaps are:
- `pending_amount == 0` and `swapped amount > 0` => it means applying a swap has partially or fully canceled any previous inverse swaps. The required amount is obtained from the inverse amount. No more amount needs to be swapped. OrderBook will Not create a new order.
- `pending amount > 0` and `swapped amount > 0` => It means applying a swap has fully canceled any previous inverse swaps, but more amount needs to be swapped. Orderbook will create a new order.
- `pending amount > 0` and `swapped amount == 0` => There was no inverse order, so we need to create a new order.


NOTE: Probably, a better name for `swap_status.swapped` is `swap_status.canceled`. It all depends on how you interpret it.

mustermeiszer: Thanks a lot, that helped!

mustermeiszer (approved):

wischli (approved):

PR #1772, opened 2024-03-15
Post RU: Remove old migrations and bump up versions

lemunozm:
## Changes and Descriptions

- Remove all temporal (and now unused) migrations
- Bump up versions for all runtimes.


lemunozm:
Maybe I've removed some migration we want to keep. If this is the case, please tell me, and I will return those.

wischli (approved):
And just like that, a huge pile of dust and dry sweat is wiped off. Once again, thanks for cleaning up!

cdamian (dismissed):
Nice to see so many of these getting cleaned up, thank you @lemunozm!

wischli (approved):

wischli (approved):

PR #1773, opened 2024-03-18
Fix: order book working with local currencies

lemunozm:
# Description

- Remove `DecimalConverter` trait, which does not handle Local currencies.
- Compute the min fulfillment amount directly in `order-book` generically for any currency.


pallets/order-book/src/lib.rs
```diff
@@ -444,6 +437,8 @@ pub mod pallet {
 				Ok::<_, DispatchError>(*n)
 			})?;
 
+			dbg!(amount_out);
+			dbg!(min_fulfillment_amount_out);
```
wischli, line 441: Leftovers from debugging

wischli (commented):
LGTM except for the two debugging leftovers. Thanks a lot for taking care of this so quickly! I propose to open a follow up PR which bumps spec versions.

wischli (approved):
Faster than Usain Bolt!

lemunozm:
😆, thanks for reviewing William!

lemunozm:
Not so fast, benchmarking is failing

lemunozm:
Modified the fix a bit to use the native decimals instead of the native currency itself, to avoid difficult initializations of the native currency in benchmarking of pallets that uses order-book

wischli (approved):

PR #1778, opened 2024-03-26
Rewards: move rewards api tests to generic testing

lemunozm:
# Description

- Remove old tests
- Added new tests
- Swap reward instances for `development-runtime` in order to have all three runtimes aligned

# Why we need this?
- Less code to maintain
- This help with future upgrades


cdamian (approved):
Awesome stuff @lemunozm, thanks for tackling this!

PR #1780, opened 2024-03-27
integration-tests: Fix test envs

cdamian:
Fixes compilation issues with integration tests branch for polkadot v1.1.0 update.


runtime/integration-tests/src/evm/precompile.rs
```diff
@@ -91,8 +91,14 @@ async fn axelar_precompile_execute() {
...
-		name: "Test".into(),
-		symbol: "TST".into(),
+		name: BoundedVec::<u8, <Runtime as pallet_pool_system::Config>::StringLimit>::try_from(
+			"Test".as_bytes().to_vec(),
+		)
+		.expect("Can create BoundedVec for token name"),
+		symbol: BoundedVec::<u8, <Runtime as pallet_pool_system::Config>::StringLimit>::try_from(
+			"TST".as_bytes().to_vec(),
+		)
+		.expect("Can create BoundedVec for token symbol"),
```
lemunozm, line 101: I think just `BoundedVec::default()` is enough for this test. To simplify a bit this.

```suggestion
		name: BoundedVec::default(),
		symbol: BoundedVec::default(),
```

runtime/integration-tests/src/generic/cases/account_derivation.rs
```diff
@@ -187,7 +188,10 @@ fn remote_account_on_relay<T: Runtime>() {
 				AccountId32 {
 					id: KEY_32,
 					network: Some(NetworkId::ByGenesis(
-						frame_system::BlockHash::<T>::get(T::BlockNumber::zero()).0,
+						frame_system::BlockHash::<T>::get(
+							<<T::BlockExt as BlockT>::Header as HeaderT>::Number::zero(),
+						)
+						.0,
```
lemunozm, line 194: Super NIT
```suggestion
						frame_system::BlockHash::<T>::get(Zero::zero()).0,
```

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -184,10 +186,22 @@ pub mod utils {
...
+			<T as pallet_pool_system::Config>::StringLimit,
+		> = AssetMetadata {
 			decimals: 12,
-			name: "Acala Dollar".into(),
-			symbol: "AUSD".into(),
+			name: BoundedVec::<
+				u8,
+				<development_runtime::Runtime as pallet_pool_system::Config>::StringLimit,
+			>::try_from("Acala Dollar".as_bytes().to_vec())
+			.expect("Can create BoundedVec for token name"),
```
lemunozm, line 199: Same as before

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -321,14 +336,14 @@ mod development {
 							// liquidity pools AddTranche message.
 							token_name: BoundedVec::<
 								u8,
-								<T as pallet_pool_system::Config>::MaxTokenNameLength,
+								<T as pallet_pool_system::Config>::StringLimit,
 							>::try_from("A highly advanced tranche".as_bytes().to_vec())
-							.expect(""),
+							.expect("Can create BoundedVec for token name"),
```
lemunozm, line 341: Same as before, `BoundedVec::default()`

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -5730,7 +5745,11 @@ mod altair {
 			env: &mut FudgeEnv<T>,
 			transfer_amount: Balance,
 			currency_id: CurrencyId,
-			meta: AssetMetadata<Balance, CustomMetadata>,
+			meta: AssetMetadata<
```
lemunozm, line 5748: You can use `cfg_types::tokens::AssetMetadata` which already has the expected bounds

lemunozm (commented):
Everything is perfect, just some NITs so reduce a bit the code lines

runtime/integration-tests/src/evm/precompile.rs
cdamian: Hmm, let's try this, even tho it might make life a bit harder when debugging.

runtime/integration-tests/src/generic/envs/fudge_env/handle.rs
```diff
@@ -96,7 +94,7 @@ pub trait FudgeHandle<T: Runtime> {
 	type RelayApi: BlockBuilderApi<RelayBlock>
 		+ BabeApi<RelayBlock>
 		+ ParachainHost<RelayBlock>
-		+ ApiExt<RelayBlock, StateBackend = <TFullBackend<RelayBlock> as Backend<RelayBlock>>::State>
+		+ ApiExt<RelayBlock>
```
lemunozm, line 97: I just curious: how did you get to solve the issue that appeared when you changed this?

cdamian: I just removed that and the compiler was happy, the `APIExt` trait only has 1 generic arg now - the block.

lemunozm: Wow really? When I removed it, a big big error was shown to me. Maybe something has changes with the rust version

cdamian: Weird, I still get some errors there but they don't seem related to this. Anyway, let's wait until I fix all of these, who knows what dragon will rise ^^

runtime/altair/src/lib.rs
```diff
@@ -1977,6 +2001,7 @@ construct_runtime!(
 		Treasury: pallet_treasury::{Pallet, Call, Storage, Config<T>, Event<T>} = 69,
 		Preimage: pallet_preimage::{Pallet, Call, Storage, Event<T>} = 70,
 		Uniques: pallet_uniques::{Pallet, Call, Storage, Event<T>} = 72,
+		Paras: polkadot_runtime_parachains::paras::{Pallet, Call, Storage, Event, Config<T>, ValidateUnsigned} = 73,
```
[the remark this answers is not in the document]
cdamian: This test here - https://github.com/centrifuge/centrifuge-chain/blob/polkadot-v1.1.0-it-env-fix/runtime/integration-tests/src/generic/cases/liquidity_pools.rs#L7316

```
  _config: Default::default(),
     |                              ^^^^^^^^^^^^^^^^^^ the trait `polkadot_runtime_parachains::paras::Config` is not implemented for `T`
 ```
 
 I'll try removing that from the test and see if we can skip adding this, I don't remember the setups required for most of these tests now.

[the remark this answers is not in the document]
Comment (cdamian) [reply runtime/altair/src/lib.rs:2004]: We're no longer able to open an HRMP channel for XCM messages if we remove that `GenesisConfig`:

```
Expected Ok(_). Got Err(
    DispatchErrorWithPostInfo {
        post_info: PostDispatchInfo {
            actual_weight: None,
            pays_fee: Pays::Yes,
        },
        error: Module(
            ModuleError {
                index: 60,
                error: [
                    1,
                    0,
                    0,
                    0,
                ],
                message: Some(
                    "OpenHrmpChannelInvalidRecipient",
                ),
            },
        ),
    },
)
```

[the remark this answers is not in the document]
mustermeiszer: Is this not also on Centrifug? Then why do we want to clean that up?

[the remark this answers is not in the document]
cdamian: @mustermeiszer I specifically added that to all runtimes in this PR.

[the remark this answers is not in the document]
wischli: There must be another way without adding this pallet. It is meant to be used on relay chains.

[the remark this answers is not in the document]
cdamian: OK, then I'll do a bit of digging there. Thanks for checking!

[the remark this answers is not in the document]
cdamian: On [this line](https://github.com/centrifuge/centrifuge-chain/blob/polkadot-v1.1.0-it-env-fix/runtime/integration-tests/src/generic/cases/liquidity_pools.rs#L7316), I should've used `FudgeRelayRuntime<T>` instead of `T`. 

Now it works and I will go ahead and remove the extra pallet. 

Thank you all for checking this!!! :heart:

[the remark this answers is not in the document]
lemunozm: Thanks for look deeply into this!

lemunozm (approved):
LGTM

PR #1782, opened 2024-03-28
Hotfix/1027

mustermeiszer:
# Description

* Fixes weight for `update_collection` call of `pallet-oracle-collection` which is uncallable right now due to PoV Excess
* Adds missing migration for Anemoy pool loans

## Changes and Descriptions

[List your changes here]

# Checklist:


mustermeiszer:
```
```

runtime/centrifuge/src/migrations.rs
```diff
@@ -12,4 +12,61 @@
...
+						&ANEMOY_POOL_ID,
+					) {
+						Ok(_) => {
+							log::info!("{LOG_PREFIX}: Registered PriceId: {:?}", pricing.price_id())
+						}
+						Err(e) => log::info!(
+							"{LOG_PREFIX}: Failed to register PriceId: {:?}, with error: {:?}.",
+							pricing.price_id(),
+							e
+						),
```
lemunozm, line 52: We should only consider it as an error if the loan is `BorrowRestrictions::OraclePriceRequired`. Otherwise, getting an error from `register_id()` is fine. But not sure if for Anemoy we have this case. If we do not see any error right now here, then we do not need to modify it.

pallets/oracle-collection/src/benchmarking.rs
```diff
@@ -109,7 +109,7 @@ mod benchmarks {
 	}
 
 	#[benchmark]
-	fn update_collection(n: Linear<1, 10>, m: Linear<1, 10>) -> Result<(), BenchmarkError> {
+	fn update_collection(n: Linear<1, 5>, m: Linear<1, 100>) -> Result<(), BenchmarkError> {
```
lemunozm, line 112: Why 100?

lemunozm (approved):
Thanks @mustermeiszer and @wischli for taking care of this 🙏🏻 

Just a couple of minor comments, not blocked. LGTM

lemunozm:
What is left in this PR?

wischli (approved):

wischli:
> What is left in this PR?

Nothing, just merged :)

PR #1791, opened 2024-04-02
wip: Make runtime apis work

mustermeiszer:
# Description

[description]

Fixes #(issue)

## Changes and Descriptions

[List your changes here]

# Checklist:

### DIFF
--- a/runtime/altair/src/lib.rs
+++ b/runtime/altair/src/lib.rs
@@ -2295,7 +2295,7 @@ impl_runtime_apis! {
 		fn tranche_token_price(pool_id: PoolId, tranche: TrancheLoc<TrancheId>) -> Option<Quantity>{
 			let now = <Timestamp as UnixTime>::now().as_secs();
 			let mut pool = PoolSystem::pool(pool_id)?;
-			let nav = Loans::update_nav(pool_id).ok()?;
+			let nav = Loans::update_nav(pool_id, Accuracy::RuntimeApi).ok()?;
 			let total_assets = pool.reserve.total.saturating_add(nav);
 			let index: usize = pool.tranches.tranche_index(&tranche)?.try_into().ok()?;
 			let prices = pool
@@ -2308,7 +2308,7 @@ impl_runtime_apis! {
 		fn tranche_token_prices(pool_id: PoolId) -> Option<Vec<Quantity>>{
 			let now = <Timestamp as UnixTime>::now().as_secs();
 			let mut pool = PoolSystem::pool(pool_id)?;
-			let nav = Loans::update_nav(pool_id).ok()?;
+			let nav = Loans::update_nav(pool_id, Accuracy::RuntimeApi).ok()?;
 			let total_assets = pool.reserve.total.saturating_add(nav);
 			pool
 				.tranches
@@ -2334,8 +2334,8 @@ impl_runtime_apis! {
 
 		fn nav(pool_id: PoolId) -> Option<PoolNav<Balance>> {
 			let pool = pallet_pool_system::Pool::<Runtime>::get(pool_id)?;
-			let nav_loans = Loans::update_nav(pool_id).ok()?;
-			let nav_fees = PoolFees::update_nav(pool_id).ok()?;
+			let nav_loans = Loans::update_nav(pool_id, Accuracy::RuntimeApi).ok()?;
+			let nav_fees = PoolFees::update_nav(pool_id, Accuracy::RuntimeApi).ok()?;
 			let nav = pallet_pool_system::Nav::new(nav_loans, nav_fees);
 			let total = nav.total(pool.reserve.total).unwrap_or(Balance::default());
 
--- a/libs/traits/src/lib.rs
+++ b/libs/traits/src/lib.rs
@@ -118,13 +118,18 @@ pub trait Reward {
 	) -> DispatchResultWithPostInfo;
 }
 
+pub enum Accuracy {
+	Exact,
+	RuntimeApi,
+}
+
 /// A trait that can be used to fetch the nav and update nav for a given pool
 pub trait PoolNAV<PoolId, Amount> {
 	type ClassId;
 	type RuntimeOrigin;
 	// nav returns the nav and the last time it was calculated
 	fn nav(pool_id: PoolId) -> Option<(Amount, u64)>;
-	fn update_nav(pool_id: PoolId) -> Result<Amount, DispatchError>;
+	fn update_nav(pool_id: PoolId, accuracy: Accuracy) -> Result<Amount, DispatchError>;
 	fn initialise(
 		origin: Self::RuntimeOrigin,
 		pool_id: PoolId,
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -75,8 +75,8 @@ pub mod pallet {
 		changes::ChangeGuard,
 		data::{DataCollection, DataRegistry},
 		interest::InterestAccrual,
-		IntoSeconds, Permissions, PoolInspect, PoolNAV, PoolReserve, PoolWriteOffPolicyMutate,
-		Seconds, TimeAsSecs,
+		Accuracy, IntoSeconds, Permissions, PoolInspect, PoolNAV, PoolReserve,
+		PoolWriteOffPolicyMutate, Seconds, TimeAsSecs,
 	};
 	use cfg_types::{
 		adjustments::Adjustment,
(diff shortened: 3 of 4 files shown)


mustermeiszer:
We can also think about removing the `update_nav` forever and make the calls to loans and fees more explicit. Would be easier to control which oracle values we might want to fake for some calls.

lemunozm:
I do not understand why this fixes the issue. Why using no prices for `RuntimeApi` would fix it?

pallets/pool-system/src/impls.rs
```diff
@@ -454,12 +454,20 @@ impl<T: Config> ChangeGuard for Pallet<T> {
...
 		}
 
-		allowed
+		let change = allowed
 			.then(|| {
 				NotedChange::<T>::remove(pool_id, change_id);
 				change
 			})
-			.ok_or(Error::<T>::ChangeNotReady.into())
+			.ok_or(Error::<T>::ChangeNotReady.into())?;
```
lemunozm, line 462: This is already notified in the consumer side of `released()`: https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/oracle-collection/src/lib.rs#L244

Not sure if we want another event here.

runtime/common/src/apis/loans.rs
```diff
@@ -32,3 +34,17 @@ decl_runtime_apis! {
 		fn portfolio_valuation(pool_id: PoolId, input_prices: PriceCollectionInput) -> Result<Balance, DispatchError>;
 	}
 }
+
+pub fn update_nav_api_call<T: pallet_loans::Config>(
```
lemunozm, line 38: NIT. We're using this for the API, but it seems like a common method that can be used for other purposes. I would move to `runtime/common/src/lib.rs` instead.

pallets/pool-system/src/impls.rs
wischli: AFAICS, we are not emitting such an event for pool fees for loans. IMO, it makes sense to add this here.

wischli (commented):
Changes seem reasonable to me, thanks for the quick fix!

@lemunozm Could you elaborate on the the difference between `older_value_timestamp` and `last_updated` entries of a `CachedCollection`? On Centrifuge chain, both values are the same right now. IIUC, `now - older_value_timestamp` needs to be less than the `value_lifetime` entry of the `CollectionInfo`, correct?

pallets/pool-system/src/impls.rs
mustermeiszer: Also the Events are quite confusing without showing that this came from a change that was released: 
* see: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Ffullnode.demo.k-f.dev#/explorer/query/0xdcd2cf1b78343a46ffea6b4a959e7644ae59aca614336a7648a6423c53dd6ca1

And if we are noting a change we are also emitting an event. I would keep it.

runtime/common/src/apis/loans.rs
mustermeiszer: Same name?

lemunozm:
> @lemunozm Could you elaborate on the the difference between older_value_timestamp and last_updated entries of a CachedCollection? On Centrifuge chain, both values are the same right now. IIUC, now - older_value_timestamp needs to be less than the value_lifetime entry of the CollectionInfo, correct?

Yes:
- `older_value_timestamp`: contains the timestamp of the older price of the collection. That timestamp comes from when the feeder fed the value.
- `last_updated: contains the timestamp when the collection was updated (when we call `update_collection()`).

Note in the code that `older_value_timestamp` is mutable and is overwriten with older values.

runtime/common/src/apis/loans.rs
lemunozm: No strong opinion. Maybe `infallible_update_nav()` or similar?

runtime/development/src/lib.rs
```diff
@@ -2461,21 +2445,23 @@ impl_runtime_apis! {
 		fn portfolio(
 			pool_id: PoolId
 		) -> Vec<(LoanId, ActiveLoanInfo<Runtime>)> {
+			runtime_common::update_nav(pool_id).ok();
```
lemunozm, line 2448: Same here

runtime/development/src/lib.rs
```diff
@@ -2461,21 +2445,23 @@ impl_runtime_apis! {
...
 			Loans::get_active_loans_info(pool_id).unwrap_or_default()
 		}
 
 		fn portfolio_loan(
 			pool_id: PoolId,
 			loan_id: LoanId
 		) -> Option<ActiveLoanInfo<Runtime>> {
+			runtime_common::update_nav(pool_id).ok();
```
lemunozm, line 2456: Same here

runtime/development/src/lib.rs
```diff
@@ -2503,8 +2489,7 @@ impl_runtime_apis! {
 	// PoolFeesApi
 	impl runtime_common::apis::PoolFeesApi<Block, PoolId, PoolFeeId, AccountId, Balance, Rate> for Runtime {
 		fn list_fees(pool_id: PoolId) -> Option<cfg_types::pools::PoolFeesList<PoolFeeId, AccountId, Balance, Rate>> {
-			let pool = pallet_pool_system::Pool::<Runtime>::get(pool_id)?;
-			PoolFees::update_portfolio_valuation_for_pool(pool_id, &mut pool.reserve.total.clone()).ok()?;
+			runtime_common::update_nav(pool_id).ok()?;
```
lemunozm, line 2492: Same here

runtime/altair/src/lib.rs
```diff
@@ -2412,8 +2399,7 @@ impl_runtime_apis! {
 	// PoolFeesApi
 	impl runtime_common::apis::PoolFeesApi<Block, PoolId, PoolFeeId, AccountId, Balance, Rate> for Runtime {
 		fn list_fees(pool_id: PoolId) -> Option<cfg_types::pools::PoolFeesList<PoolFeeId, AccountId, Balance, Rate>> {
-			let pool = pallet_pool_system::Pool::<Runtime>::get(pool_id)?;
-			PoolFees::update_portfolio_valuation_for_pool(pool_id, &mut pool.reserve.total.clone()).ok()?;
+			runtime_common::update_nav(pool_id).ok()?;
```
lemunozm, line 2402: Same here

runtime/altair/src/lib.rs
```diff
@@ -2371,21 +2356,23 @@ impl_runtime_apis! {
...
 		) -> Vec<(LoanId, ActiveLoanInfo<Runtime>)> {
+			runtime_common::update_nav(pool_id).ok();
 			Loans::get_active_loans_info(pool_id).unwrap_or_default()
 		}
 
 		fn portfolio_loan(
 			pool_id: PoolId,
 			loan_id: LoanId
 		) -> Option<ActiveLoanInfo<Runtime>> {
+			runtime_common::update_nav(pool_id).ok();
```
lemunozm, line 2367: Same here

runtime/altair/src/lib.rs
```diff
@@ -2371,21 +2356,23 @@ impl_runtime_apis! {
 		fn portfolio(
 			pool_id: PoolId
 		) -> Vec<(LoanId, ActiveLoanInfo<Runtime>)> {
+			runtime_common::update_nav(pool_id).ok();
```
lemunozm, line 2359: Same here

lemunozm (changes requested):
Looks good! Just some lines that can be removed because they do not affect to the computation

runtime/centrifuge/src/lib.rs
```diff
@@ -2464,8 +2450,7 @@ impl_runtime_apis! {
 	// PoolFeesApi
 	impl runtime_common::apis::PoolFeesApi<Block, PoolId, PoolFeeId, AccountId, Balance, Rate> for Runtime {
 		fn list_fees(pool_id: PoolId) -> Option<cfg_types::pools::PoolFeesList<PoolFeeId, AccountId, Balance, Rate>> {
-			let pool = pallet_pool_system::Pool::<Runtime>::get(pool_id)?;
-			PoolFees::update_portfolio_valuation_for_pool(pool_id, &mut pool.reserve.total.clone()).ok()?;
+			runtime_common::update_nav(pool_id).ok()?;
```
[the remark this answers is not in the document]
wischli: For pool fees, at least, it does though?! By updating the fees portfolio valuation and providing the current pool reserve, pool fees are updated in a simulated manner. E.g. their disbursement, pending amounts are updated.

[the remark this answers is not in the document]
lemunozm: Ah that right in this case

runtime/development/src/lib.rs
lemunozm: Do not apply

runtime/altair/src/lib.rs
lemunozm: Do not apply

runtime/centrifuge/src/lib.rs
```diff
@@ -2422,21 +2406,23 @@ impl_runtime_apis! {
 		fn portfolio(
 			pool_id: PoolId
 		) -> Vec<(LoanId, ActiveLoanInfo<Runtime>)> {
+			runtime_common::update_nav(pool_id).ok();
```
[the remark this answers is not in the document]
mustermeiszer: I do not think that is true. `Loans::update_portfolio_validation()` will update the portfolio.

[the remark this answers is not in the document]
mustermeiszer: Or could you elaborate on why not?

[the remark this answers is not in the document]
lemunozm: The below call `Loans::get_active_loans_info(pool_id).unwrap_or_default()` do not make use of the portfolio. It internally computes the present value. But it does not read computed portfolio in any case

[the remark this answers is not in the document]
mustermeiszer: Ahh, that is actually true. Did not see that code path. Will remove.

lemunozm (approved):
LGTM!

Thanks for this

wischli (approved):
LGTM!

PR #1793, opened 2024-04-03
Oracles: Correct feeder encoding

lemunozm:
# Description

This is not a current issue because the `Feeder` wrapper type does not add any byte to its serialization/deserialization, but it is a potential one if this type changes.


lemunozm:
I've also be warned because clippy for rust-1.78 complains about the Feeder `eq` implementation:
```
warning: function cannot return without recursing
  --> runtime/common/src/oracle.rs:38:2
   |
38 | /     fn eq(&self, other: &Self) -> bool {
39 | |         self.0.eq(&other.0)
40 | |     }
   | |_____^
   |
note: recursive call site
  --> runtime/common/src/oracle.rs:39:3
   |
39 |         self.0.eq(&other.0)
   |         ^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unconditional_recursion
   = note: `#[warn(clippy::unconditional_recursion)]` on by default

```

Anyway, it seems to be well defined, and I tested with the following and it pass:
```rust
#[cfg(test)]
mod tests {
use super::*;
    #[test]
    fn foo() {
        let a = Feeder::<RuntimeOrigin>::root();
        let b = Feeder::<RuntimeOrigin>::none();
        assert_ne!(a, b);
        assert_eq!(a, a);
        assert!(!(a < a));
        assert!(a <= a);
    }
}
```

mustermeiszer (approved):

wischli (approved):

lemunozm:
Regarding the above clippy warning. It seems a clippy issue: https://github.com/rust-lang/rust-clippy/issues/12245

PR #1796, opened 2024-04-04
Set correct onchain versions

lemunozm:
# Description

I'm not sure why the new Oracle pallets added have their on-chain version of 0 when it should be 1. It is not an issue now, but it can be if we update them tomorrow, and makes the new Ci job for migrations fail. This PR set them with the correct version.

My hypothesis of why this has happened is that we reused the previous indexes from the previous oracle pallets (`orml-oracle`, and `pallet-data-collection`), and because of this, it has not update the version (even when the pallet names have changed).

The same happens for `Orderbook` and `ForeignInvestments` for `altair`. Altair didn't require reseting the storage but we did not update the version numbers there.

### DIFF
--- a/runtime/altair/src/migrations.rs
+++ b/runtime/altair/src/migrations.rs
@@ -10,6 +10,11 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU General Public License for more details.
 
+use crate::{OraclePriceCollection, OraclePriceFeed};
+
 /// The migration set for Altair @ Kusama.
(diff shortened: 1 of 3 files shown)


lemunozm:
I think this changes are not required to `demo`, because soon or later it will be hard reset

wischli (approved):
Thanks for the fixes! I suppose pallets are initialized with storage version 0 instead of version at present when added to the runtime..

lemunozm:
But when adding an external pallet, it always starts with the version it has, not 0 🤔

PR #1408, opened 2023-06-19
Loans: multiple cashflows PoC

hieronx:
# Description

Implements a basic PoC for multiple cashflows, specifically monthly or semi-annual interest payments, as well as a corresponding write-off trigger.

# Checklist:

- [ ] I have added Rust doc comments to structs, enums, traits and functions
- [ ] I have made corresponding changes to the documentation
- [ ] I have performed a self-review of my code
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1797, opened 2024-04-05
Loans: multiple cashflows

lemunozm:
# Description

Continue work from https://github.com/centrifuge/centrifuge-chain/pull/1408

This PR is an MVP. We expect:
- Paid on day 1 of all months.
- Only monthly cashflows.

## Task list

- [x] Compute monthly dates 
- [x] Compute monthly intervals
- [x] Compute monthly cashflows
- [x] Cashflow date UTs
- [x] Cashflow interest UTs
- [x] Add Runtime API
- [x] Check cashflow when borrowing
- [x] UTs checking cashflow when borrowing a loan
- [x] Update `plantuml` model


pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,348 @@
...
+					InterestPayments::None => 0,
+					InterestPayments::Monthly(_) => 12,
+					InterestPayments::SemiAnnually(_) => 2,
+					InterestPayments::Annually(_) => 1,
+				};
+
+				let interest_rate_per_period = interest_rate
+					.per_year()
+					.ensure_div(Rate::saturating_from_integer(periods_per_year))?;
+				let amount_per_period = interest_rate_per_period.ensure_mul_int(principal)?;
```
lemunozm, line 249: @hieronx I want to double check with you that this computation is correct.

This is performed for each element in the `cashflows` list, which is per month. But here, if the `interest_payments` is i.e. `Annually`, we are setting the whole year's interest to that cashflow month.

So, is this incorrect, or should I expect that if the user calls this with an `Annually` interest payment, then the user also sets a cash flow per year?

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,348 @@
...
+					.with_rolling_day(reference_date.day())
+					.unwrap()
+					.into_iter()
+					// There's no interest payment expected on the origination date
+					.skip(1)
+					.collect()
+			}
+			ReferenceDate::End => DateRule::monthly(start)
+				.with_end(end)
+				.with_rolling_day(31)
```
lemunozm, line 184: For reference, here the cashflow list is computed per month

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,474 @@
...
+		&self,
+		start: Seconds,
+		end: Seconds,
+	) -> Result<Vec<(Seconds, Seconds)>, DispatchError> {
+		if start > end {
+			return Err(DispatchError::Other("cashflow must start before it ends"));
+		}
+
+		let start = seconds_to_date(start)?;
+		let end = seconds_to_date(end - 1)? + Months::new(1);
```
mustermeiszer, line 116: WHy the minus 1?

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,474 @@
...
+
+		let rolling_days = match self {
+			Self::Date(reference) => seconds_to_days(*reference)?,
+			Self::End => 31,
```
mustermeiszer, line 120: Why default to 31 days?

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,474 @@
...
+	/// period
+	End,
+}
+
+impl ReferenceDate {
+	fn monthly_cashflow_dates(
+		&self,
+		start: Seconds,
+		end: Seconds,
+	) -> Result<Vec<(Seconds, Seconds)>, DispatchError> {
```
mustermeiszer, line 110: What is the `(Seconds, Seconds)` representing?

lemunozm:
This is super still super WIP

pallets/loans/src/types/cashflow.rs
lemunozm: Still WIP, the first is the date in seconds when you need to make the payment, and the second is the interval for that payment. Normally, the interval will be the whole month, except for the first and last months that it can be fewer days

pallets/loans/src/types/cashflow.rs
lemunozm: It means you will pay at the end of the month. 31 as the highest value, if the month does not have 31, it will be adjusted to the last day of the month.

pallets/loans/src/types/cashflow.rs
lemunozm: We want as a final date for iterating months at least one month less a second.

It means, if your loan is valid until the first second of March 1, then you can pay until March 31, 23:59:59:999. That's the theoretical worst case.

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,474 @@
...
+
+fn seconds_in_month(date: NaiveDate) -> Result<Seconds, DispatchError> {
+	todo!()
+}
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeDebug, MaxEncodedLen)]
+pub enum ReferenceDate {
+	/// Payments are expected every period relative to a specific date.
+	/// E.g. if the period is monthly and the specific date is Mar 3, the
+	/// first interest payment is expected on Apr 3.
```
mustermeiszer, line 97: I just checked back with Dennis and the following would be expected for interest payments:

* Origination date: Anytime during a month
* First interest payment: Last day of month of origination
* Following interest payments: Last day of month
* Last interest payment: Day of maturity

We can neglect only counting business days as valid repayment days for now. That is an edge we can cover by adding soft boundaries to penalties later down the road.

pallets/loans/src/types/cashflow.rs
mustermeiszer: And we need the interval for calculation purposes?

pallets/loans/src/types/cashflow.rs
mustermeiszer: Where does this requirement come from? Why can't you pay on that date inclusive? I would assume it is inclusive. Cc @denniswell

pallets/loans/src/types/cashflow.rs
mustermeiszer: Ah, I was confised because of the error a few lines down, where we have an error if there is month with less then 31 days.

pallets/loans/src/types/cashflow.rs
lemunozm: If you need to pay on the last day of the month, and your overdue date is on day 1 in a month, then you can pay until that month ends.

That was my conclusion from: https://centrifugedao.slack.com/archives/C05DX0RH1T3/p1712920193116179

pallets/loans/src/types/cashflow.rs
lemunozm: Yeah, it should be something like "Paydown day must be less than 31 days"

pallets/loans/src/types/cashflow.rs
lemunozm: Yes, we would need it later to apply a proportional interest for each month

lemunozm: Suppose you pay at the end of the month, but you've borrowed on day 15. The first month you will pay only half of the month. The same will happen with the last month.

pallets/loans/src/types/cashflow.rs
lemunozm: > - First interest payment: Last day of month of origination
> - Following interest payments: Last day of month

Yes, but I talked with @hieronx that ideally, it could be any day, if specified.

> - Last interest payment: Day of maturity

Ok, this is different to what I thought 👍🏻

pallets/loans/src/types/cashflow.rs
lemunozm: Ok but make sense too to pay the same day of maturity 👍🏻

pallets/loans/src/tests/borrow_loan.rs
```diff
@@ -603,11 +599,178 @@ fn increase_debt_does_not_withdraw() {
 		let amount = ExternalAmount::new(QUANTITY, PRICE_VALUE);
 		config_mocks(amount.balance().unwrap());
 
-		assert_ok!(Loans::borrow(
+		MockPools::mock_withdraw(|_, _, _| {
+			unreachable!("increase debt must not withdraw funds from the pool");
+		});
+
+		assert_ok!(Loans::increase_debt(
```
lemunozm, line 606: This is a fix for a previous test. Not related to the topic of this PR.

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,432 @@
...
+			InterestPayments::None => (),
+			InterestPayments::Monthly(_) => {
+				let start = date::from_seconds(now)?;
+				let end = date::from_seconds(self.maturity.date())?;
+
+				// We want to avoid creating a loan with a cashflow consuming a lot of computing
+				// time Maximum 40 years, which means a cashflow list of 40 * 12 elements
+				if end.year() - start.year() > 40 {
+					return Ok(false);
+				}
```
lemunozm, line 129: We need to set a maximum to avoid computing big cashflows that could block the chain.

40 years is ok @hieronx? I do not know loans higher than that.

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,432 @@
...
+	) -> Result<Vec<NaiveDate>, DispatchError> {
+		if start_date >= end_date {
+			return Err(DispatchError::Other("Cashflow must start before it ends"));
+		}
+
+		if reference_day != DAY_1 {
+			return Err(DispatchError::Other(
+				"Only day 1 as reference is supported by now",
+			));
+		}
```
lemunozm, line 245: I cannot return custom errors without access to `T: Config` from this. I'm not sure if we're okay with the `Other` variant for these errors.

lemunozm (commented):
Ready to review. Some comments below.

I want to improve the `borrow()` benchmark to take into account the cashflow computation. Maybe in this or in a following up PR.

pallets/loans/src/entities/loans.rs
```diff
@@ -345,6 +361,18 @@ impl<T: Config> ActiveLoan<T> {
 			Error::<T>::from(BorrowLoanError::MaturityDatePassed)
 		);
 
+		let expected_payment = self.schedule.expected_payment(
+			self.origination_date,
```
hieronx, line 365: I think we should generate cashflows here from `repayments_on_schedule_until` rather than from `origination_date`. Since otherwise, if there is an additional borrow, the historic repayments before today would have also changed, and suddenly there will be payments overdue that weren't overdue before. That was the goal with the `repayments_on_schedule_until` variable.

runtime/centrifuge/src/lib.rs
```diff
@@ -2398,6 +2399,10 @@ impl_runtime_apis! {
 		) -> Result<Balance, DispatchError> {
 			Ok(runtime_common::update_nav_with_input(pool_id, input_prices)?.nav_aum)
 		}
+
+		fn cashflow(pool_id: PoolId, loan_id: LoanId) -> Result<Vec<CashflowPayment<Balance>>, DispatchError> {
```
hieronx, line 2403: nit: maybe better calling this `expected_cashflows`? Since just cashflow implies also including historic cashflows, which is not the case

pallets/loans/src/entities/loans.rs
lemunozm: That's right! Good catch 👍🏻

pallets/loans/src/types/cashflow.rs
hieronx: Yes definitely 40 years is ok :)

lemunozm:
Should be ready to be merged after reviewing it: @mustermeiszer @hieronx

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,430 @@
...
+				let principal = proportion.ensure_mul_int(principal)?;
+				let interest = interest_per_period.ensure_mul_int(principal)?;
+
+				Ok(CashflowPayment {
+					when: date::into_seconds(date)?,
+					principal,
+					interest,
+				})
+			})
+			.collect()
```
mustermeiszer, line 179: We are creating diverging truths here. 
1. The interest of the expected cashflows is based on a monthly schedule, without any compounding. 
2. The interest of the loan itself is secondly compounding on the principal

I think we need to reiterate this and rather change the interest rate field on the loan itself and with it the `pallet-interest-accrual`

Also, what is proportion doing here? Why are we only calculating the interest based on a portion of the principal?

mustermeiszer (changes requested):
Flagging for now. Details in review comments. 

* What is the reason for the `DAY_1` restriction, just out of interest?

pallets/loans/src/types/cashflow.rs
lemunozm: Mmm... good point. Let me think about this!

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,430 @@
...
+		let cashflow = self.generate_cashflows(origination_date, principal, interest_rate)?;
+
+		let total_amount = cashflow
+			.into_iter()
+			.take_while(|payment| payment.when < until)
+			.map(|payment| payment.principal.ensure_add(payment.interest))
+			.try_fold(Balance::zero(), |a, b| a.ensure_add(b?))?;
+
+		Ok(total_amount)
+	}
```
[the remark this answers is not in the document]
lemunozm: > not more interest or principal then expected is repayed.

Why this restriction? Could not anyone pay in advance for the upcoming months if they want?

> Currently, we are only checking if outstanding repayments are overdue when borrowing again.

Yes, this is the MVP of this feature I talked with @hieronx about. We're only checking that to prevent the cashflow history from being broken.

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,430 @@
...
+	///
+	/// NOTE: Only day 1 is supported by now
+	Monthly(u8),
+}
+
+/// Specify the paydown schedules of the loan
+#[derive(Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeDebug, MaxEncodedLen)]
+pub enum PayDownSchedule {
+	/// The entire borrowed amount is expected to be paid back at the maturity
+	/// date
```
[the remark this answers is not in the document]
lemunozm: It means that no matter how you pay it (one big payment or several payments during the time), you need to repay all before the maturity. Basically, it means no restrictions on the payments. Maybe we can rewrite this to avoid misunderstandings

lemunozm:
> What is the reason for the DAY_1 restriction, just out of interest?

The cashflow date/interval computation gets harder to compute with several border cases, i.e., the user sets day 31.

pallets/loans/src/types/cashflow.rs
lemunozm: > Also, what is proportion doing here? Why are we only calculating the interest based on a portion of the principal?

Proportion is (or should be) the weight a month has regarding other months. This is because the borrowing could start in the middle of the month. In that case, such a month will have less weight than other entire months

hieronx: They are different use cases
- The interest on the loan itself is the amount that needs to be repaid now to close the loan
- The expected cashflows is for other use cases, including
  - The valuation being able to account for the expected cashflow dates (which we will adapt in the future with DCF)
  - The UI knowing when cashflows are expected
  - Enabling write-off policies to act on overdue cashflows

It also makes sense that the expected cashflows are non-compounding: they are assuming that each cashflow itself is paid in time and thus there is no compounding month over month.

pallets/loans/src/types/cashflow.rs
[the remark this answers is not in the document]
lemunozm: [77dada6](https://github.com/centrifuge/centrifuge-chain/pull/1797/commits/77dada68c61c44f1af1ad42d5cb51d2330b4596c)

pallets/loans/src/types/cashflow.rs
mustermeiszer: @hieronx and I just briefly synced and we definitely need to have cashflows and interest on the loan data being aligned. The easiest would be to make the cashflows compounding in the intervals they are expected. If they are repayed on time, the payments match. 


Due to secondly compounding this requirement will be impossible to be matched, so I think we need to take a deeper look here.

pallets/loans/src/types/cashflow.rs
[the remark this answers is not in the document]
lemunozm: > If you are repaying the principal, you are decreasing the future cashflows

Yeah, ideally, the repayment should be amortized. By now, if you repay more, it means that you do not need to repay the next month. With 0 amortizations.

pallets/loans/src/types/cashflow.rs
lemunozm: > Due to secondly compounding this requirement will be impossible to be matched

This is because our monthly cashflow has day precision, right?

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,429 @@
...
+
+		let interest_per_period = interest_rate
+			.per_year()
+			.ensure_div(Rate::saturating_from_integer(periods_per_year))?;
+
+		timeflow
+			.into_iter()
+			.map(|(date, weight)| {
+				let proportion = FixedU128::ensure_from_rational(weight, total_weight)?;
+				let principal = proportion.ensure_mul_int(principal)?;
```
mustermeiszer, line 169: This should always be based on the whole `principal` amount. 

Example: 
* `principal = 100`
* 12 periods, `5%` `interest_per_period` for the sake of simplicity

Without a paydown in the `principal` amount you would expect to have an interst payment of `5` each month. 

We are having secondly compounding, so the actual interest that has to be paid will be different. As the `100` are compounding every second, and then the interest to be payed is actually above `5` and will increase each month. Slightly of course only. 

My mind model is: 
if the compounding interval is the same as the repayment interval, then there is no compounding at all. If the compounding interval is smaller, then there is compounding. If the repayment interval is smaller, then there is no compounding and the interest payments each month are getting smaller.

lemunozm: Then, in order to compute a cashflow upfront, we need to ensure the repayment interval is always <= than the compounding. If not, we can not compute an exact cashflow upfront, right?

mustermeiszer: Yes, but right now, it is anyways not exact, given the secondly think. So I think this is an estimate on the given state.

pallets/loans/src/types/cashflow.rs
mustermeiszer: It has monthly precisions IIUC

mustermeiszer:
@lemunozm in order to allow the UI and the indexers to use that feature, while we are still unsure if we want to ship it with the given inconsistency we decided to do the following.

* expose  runtime api as is
* do NO restrictions or whatsoever during borrow/repay
* calculate the cashflow as if ALL needs to be paid at maturity

This way the UI can already display the cashflow and we can seamlessly iterate and add more correct cashflows.

lemunozm:
Makes sense! I like this bottom-up approach. Will do so.

> calculate the cashflow as if ALL needs to be paid at maturity

That would increase the resulting interest a lot, right? I can imagine my mortgage taking into account that all principal has been accrued interest during all those years.

mustermeiszer:
Yes, but atm we mostly have non interest bearing assets.

lemunozm:
For future reference. Copy the following to the end of `ensure_can_borrow()` when restrictions apply again:

```rust
let expected_payment = self.schedule.expected_payment(
	self.repayments_on_schedule_until,
	self.principal()?,
	self.pricing.interest().rate(),
	now,
)?;

ensure!(
	self.total_repaid.effective()? >= expected_payment,
	Error::<T>::from(BorrowLoanError::PaymentOverdue)
);
```

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,451 @@
...
+		let total_weight = timeflow
+			.iter()
+			.map(|(_, weight)| weight)
+			.try_fold(0, |a, b| a.ensure_add(*b))?;
+
+		let lifetime = maturity.ensure_sub(origination_date)?.ensure_into()?;
+		let interest_rate_per_lifetime = ensure_pow(interest_rate.per_sec()?, lifetime)?;
+		let interest_at_maturity = interest_rate_per_lifetime
+			.ensure_mul_int(principal)?
+			.ensure_sub(principal)?;
```
lemunozm, line 180: Computed here the interest supposing everything is paid at maturity

pallets/loans/src/tests/mod.rs
mustermeiszer, line 1: Can we add a test for the cashflow, to ensure the right amount is returned by the API?

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,451 @@
...
+	/// All interest is expected to be paid at the maturity date
+	None,
+
+	/// Interest is expected to be paid monthly
+	/// The associated value correspond to the paydown day in the month,
+	/// from 1-31.
+	/// The day will be adjusted to the month.
+	///
+	/// NOTE: Only day 1 is supported by now
+	Monthly(u8),
```
mustermeiszer, line 93: I would remove that one, we are not using it atm, right?

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,451 @@
...
+				let principal = proportion.ensure_mul_int(principal)?;
+				let interest = proportion.ensure_mul_int(interest_at_maturity)?;
+
+				Ok(CashflowPayment {
+					when: date::into_seconds(date)?,
+					principal,
+					interest,
+				})
+			})
+			.collect()
```
mustermeiszer, line 195: If `InterestPayments::None` is choosen, there will be no `CashflowPayment`. 

But rather we want the `InterestPayments::None` to have a single payment at the mutrity date. Also, we could think about renaming `None` to `AtMaturity`.

pallets/loans/src/types/cashflow.rs
lemunozm: The variant is used. The value of the variant is clamped to expected to be `1`, but to avoid migrations, I'll keep it. WDYT?

pallets/loans/src/tests/mod.rs
lemunozm: They are under `borrow_loan.rs` because it (initially) was computed when borrowing from the loan.

lemunozm: The API call just wraps up that call. But as long as the API type checks, it means it is calling `expected_cashflows()` under the hood

pallets/loans/src/types/cashflow.rs
lemunozm: Mmmm... Makes sense. Doing it 👍🏻

lemunozm: WDYT `OnceAtMaturity` to express that you only have one entry in that case?

lemunozm: Done here: [f9db4fc](https://github.com/centrifuge/centrifuge-chain/pull/1797/commits/f9db4fcfe2211c47133005b0640df5437000235c)

lemunozm: BTW, the expected interest to pay in the cashflow will be different from the expected interest to pay with external loans. Are we ok with that?

lemunozm:
`breaks-api` tag due:
- Loans API to V3 (not exactly breaking)
- `InterestPayments::None` renamed to `InterestPayments::OnceByMaturity`

pallets/loans/src/types/cashflow.rs
mustermeiszer: Why?

lemunozm: Because the interest in external pricing loan is affected by the notional price.

lemunozm: https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/loans/src/entities/pricing/external.rs#L220-L226

Line `225` is equal to how we compute the interest in the `cashflows`. That contains the principal + interest. 

Nevertheless, in the cashflows & internal pricing, we subtract that by the outstanding principal, and for external pricing, the outstanding principal is based on the notional

(BTW, we should unify these 3 points of computing the interest)

lemunozm: I think I have a fast fix to unify them. Working on this

lemunozm: https://github.com/centrifuge/centrifuge-chain/pull/1864

pallets/loans/src/types/cashflow.rs
mustermeiszer: We will not use it and it complicates the logic IMO.

mustermeiszer: If we add another variant we will not need a migration.

pallets/loans/src/types/cashflow.rs
mustermeiszer: > If InterestPayments::None is choosen, there will be no CashflowPayment.

I still think that this is not working as expected. Would like to see a test where the actual `Payment` is shown made for externally priced assets.

pallets/loans/src/types/cashflow.rs
lemunozm: But we will have a mixed logic.

```rust
InterestPayments {
    OnceAtmaturity,
    Monthly, // paid at day 1
    MonthlyAt(u8) // Added in the future, collides with MonthlyAt(1) == Monthly
}
```

I would prefer remove it, you know I like to remove things 😆, just being aware of the future migration

pallets/loans/src/types/cashflow.rs
lemunozm: > If InterestPayments::None is choosen, there will be no CashflowPayment.

This is working as expected because now, for InterestPayments, we're returning one element in the cashflow. I added a test.

What is not working as expected is the external price. But with the fix above it should. Still pending the test. Will do later today or early tomorrow

pallets/loans/src/types/cashflow.rs
mustermeiszer: I am not following. I would just remove all variants besides `OnceAtMaturity`. Then we can add it as you currently have it later on, if we want to have it the same way.

pallets/loans/src/types/cashflow.rs
```diff
@@ -0,0 +1,480 @@
...
+	{
+		let Some(maturity) = self.maturity.date() else {
+			return Ok(Vec::new());
+		};
+
+		let start_date = date::from_seconds(origination_date)?;
+		let end_date = date::from_seconds(maturity)?;
+
+		let timeflow = match &self.interest_payments {
+			InterestPayments::OnceAtMaturity => vec![(end_date, 1)],
```
mustermeiszer, line 165: But this was empty the last time I checked ^^

pallets/loans/src/types/cashflow.rs
lemunozm: But then we do not have monthly cashflows? At which moment do we have to change that requirement? That's the main point of this, isn't it?

lemunozm:
External loans are not working as expected with this changes. Added the required changes in a new PR to easy review what's needed: https://github.com/centrifuge/centrifuge-chain/pull/1864

pallets/loans/src/types/cashflow.rs
mustermeiszer: We decided that the current approach probably needs rethinking and hence only want to expose the runtime api plus the `OnceAtMaturity` payment. 
So that the UI and subquery can already index via the api and we have more time to make the feature right.

lemunozm: I understood that we still wanted the monthly cashflows, but supposing all payments for all months were at maturity.

mustermeiszer: > I understood that we still wanted the monthly cashflows, supposing all payments for all months were at maturity.

I do not think we will keep the `Monthly(u8)`. I would propose we move all test and existing code to a new branch and only bring in the runtim-api and the calcualtion for `OnceAtMaturity`.

lemunozm:
Closed in favor of a simplified version: https://github.com/centrifuge/centrifuge-chain/pull/1867

Keeping this as a backup to recover the monthly cashflow

Issue #1790, opened 2024-04-02
IntegrationTests: Port non-generic tests into generic module


lemunozm:
### Description

This implies port:
- `evm` module. Done in https://github.com/centrifuge/centrifuge-chain/pull/1801
- `rewards` module. Done in https://github.com/centrifuge/centrifuge-chain/pull/1805

This allows:
- Remove all utilities needed right now just for the above modules (`runtime/integration-tests/src/util/*` modules).
- Better maintenability for future upgrades.

PR #1801, opened 2024-04-08
Integration tests. Move EVM tests to the generic framework

lemunozm:
# Description

Fix partially https://github.com/centrifuge/centrifuge-chain/issues/1790

Moving these tests to the generic framework:
- Precompile test
- Ethereum transaction test

~~**NOTE**: Blocked by [`polkadot-v1.1.0`](https://github.com/centrifuge/centrifuge-chain/pull/1756)~~ **READY**


runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
```diff
@@ -0,0 +1,104 @@
...
+				&& to == contract_address
+				&& transaction_hash == t_hash
+				&& exit_reason == ExitReason::Succeed(ExitSucceed::Returned))
+			.then_some(())
+		})
+		.unwrap();
+	});
+}
+
+crate::test_for_runtimes!([development], call);
```
lemunozm, line 104: When trying with `altair` or `centrifuge` I obtain a `BadOrigin` error (when I call `pallet_evm::call()` or `pallet_evm::create()`). Do you know what could be the reason?

cdamian (dismissed):
Thank you for cleaning these up @lemunozm!

runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
cdamian: Not from the top of my head unfortunately, no idea what might cause that.

lemunozm: I guess some constant we use in the test does not match the runtime config... any idea @mustermeiszer?

wischli (approved):
LGTM! Thanks for moving this

runtime/integration-tests/src/generic/cases/ethereum_transaction.rs
lemunozm: Leaving this for a future PR

PR #1802, opened 2024-04-09
BlockRewards: Fix how it handle ED for Staking currency.

lemunozm:
# Description

I guess this comes from the previous upgrade when balance/and currencies force `ExistentialDeposit` usage.

Issue 1: We're using an `ED` denominated in `CFG` for the `StakingCurrency`. This is not critical if both currencies use the same denomination, but we should use the correct `ED` for Staking. 

Issue 2: Because `Staking` currencies were not in the asset registry. From `orml_tokens` perspective, the `ED` is `0` while from `block-rewards` perspective it's `1 * MICRO_CFG`. I guess this is not a problem for us right now, but it should be fixed to avoid future problems.

### DIFF
--- a/pallets/block-rewards/src/lib.rs
+++ b/pallets/block-rewards/src/lib.rs
@@ -50,7 +50,7 @@ use frame_support::{
 	storage::transactional,
 	traits::{
 		fungible::{Inspect as FungibleInspect, Mutate as FungibleMutate},
-		fungibles::Mutate,
+		fungibles::{Inspect, Mutate},
 		tokens::{Balance, Fortitude, Precision},
 		OneSessionHandler,
 	},
@@ -132,9 +132,6 @@ pub mod pallet {
 		/// Type used to handle balances.
 		type Balance: Balance + MaxEncodedLen + FixedPointOperand + MaybeSerializeDeserialize;
 
-		#[pallet::constant]
-		type ExistentialDeposit: Get<Self::Balance>;
-
 		/// Type used to handle group weights.
 		type Weight: Parameter + MaxEncodedLen + EnsureAdd + Unsigned + FixedPointOperand + Default;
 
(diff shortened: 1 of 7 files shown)


pallets/block-rewards/src/lib.rs
```diff
@@ -314,10 +315,13 @@ impl<T: Config> Pallet<T> {
 	///  * deposit_stake (4 reads, 4 writes): Currency, Group, StakeAccount,
 	///    Account
 	pub(crate) fn do_init_collator(who: &T::AccountId) -> DispatchResult {
+		let existential_deposit =
+			<T::Tokens as Inspect<T::AccountId>>::minimum_balance(T::StakeCurrencyId::get());
+
 		<T::Tokens as Mutate<T::AccountId>>::mint_into(
 			T::StakeCurrencyId::get(),
 			who,
-			T::StakeAmount::get().saturating_add(T::ExistentialDeposit::get()),
```
lemunozm, line 320: Here, we were using the CFG existential deposit instead of the one from `StakeCurrencyId`

runtime/common/src/lib.rs
```diff
@@ -81,6 +85,7 @@ where
 	fn get(currency_id: &CurrencyId) -> Balance {
 		match currency_id {
 			CurrencyId::Native => T::ExistentialDeposit::get(),
+			CurrencyId::Staking(StakingCurrency::BlockRewards) => T::ExistentialDeposit::get(),
```
lemunozm, line 88: Here we were assuming `ED` for `StakingCurrency` was registered in the orml asset registry or, if not, it was `ED`. But the value was always `T::ExistentialDeposit::get()` which maps to `1 * MICRO_CFG`

lemunozm (commented):
Not sure if possible, but I guess staking currency should have an `ED` of 0 to avoid all this issues. Also for accounting, because now we have in centrifuge a total issuance of: `13,000,002,000,000,000,000` and I would say most of it is just an accumulation of existential deposits

lemunozm:
> Not sure if possible, but I guess staking currency should have an ED of 0 to avoid all these issues (and migrations each time we use a new reward system)

From some tests I'm doing, it seems possible. I'm not sure exactly what the problem we had was.

wischli:
> > Not sure if possible, but I guess staking currency should have an ED of 0 to avoid all these issues
> 
> From some tests I'm doing, it seems possible. I'm not sure exactly what the problem we had was.

AFAIR the Polkadot v0.9.43 update introduced a required ED for all currencies. Beforehand, we did not require ED for any currency but native. Unfortunately, I don't remember whether relying on unit tests was sufficient here or whether the `StakingCurrency` ED requirement was a result from testing on a local chain. I think though, some unit tests indicated a failure with ED set to 0 because some balance assertions after distributing block rewards were failing.

lemunozm:
Mmm... IIRC, when we "discovered" that `ED` requirement, we didn't have yet `CurrencyED` type and we were using the balance `ExistentialDeposit` for all currencies. Which we should't. Maybe for that we encountered that we needed `ED != 0`. 

With some minor modifications, I've got all integration tests passing with `ED = 0` for staking. Maybe it deserves after we merge this check if we have any issues in dev/demo if we use `ED = 0` for staking.

lemunozm:
Can I have your approval on this @wischli @mustermeiszer @cdamian ?

wischli (approved):
Reapproving

Issue #1804, opened 2024-04-10
IntegrationTest: `generic` tests with `assert_events!` macro support


lemunozm:
### Description

We want the macro found here: https://github.com/centrifuge/centrifuge-chain/blob/629dcfdc7495faf0427df20fb58e47fb05bccd01/runtime/integration-tests/src/utils/env.rs#L85 into the `generic` framework of `integration-test` to be able to check several events quickly and choose the `EventRange` we want to apply.

PR #1805, opened 2024-04-11
IT: rewards to generic framework. Remove previous non-generic IT utilities

lemunozm:
# Description

Fixes https://github.com/centrifuge/centrifuge-chain/issues/1790

## Changes and Descriptions

- Block rewards tests ported to run under `generic` version.
  - **WARNING**. block rewwards now uses `Instance1` instead of `Instance2` in development runtime in order to align all runtimes. This can imply some reset on dev/demo.
  - [ ] Pending to fix an event issue with this test.
  - [x] Pending to test against `altair` and `centrifuge` runtimes.
- Remove previous utilities no longer used (but It deserves to maintain this macro: https://github.com/centrifuge/centrifuge-chain/issues/1804).
- Remove `fast-runtime` feature from IT (for now, it's not required).
- Added `Blocks::JumpByNumber` to force the runtime to jump to a block in the future without passing by the previous blocks (can be used to speed up an integration-test)


lemunozm:
@gpmayorga, have you checked what we need to change to avoid showing a failure in the `codecov/patch` job?

Maybe just return 0 after calling tarpaulin?

runtime/integration-tests/src/generic/cases/block_rewards.rs
```diff
@@ -0,0 +1,199 @@
...
+fn apply_and_check_session<T: Runtime>(
+	collator_count: u32,
+	joining: Vec<AccountId>,
+	leaving: Vec<AccountId>,
+) {
+	for collator in &joining {
+		pallet_collator_selection::Pallet::<T>::register_as_candidate(
+			RawOrigin::Signed(collator.clone()).into(),
+		)
+		.unwrap();
```
lemunozm, line 111: I can't call this in `altair` and `centrifuge` because I get a `ValidatorNotRegistered` error. 

@wischli, do you have any idea why this happens?

runtime/integration-tests/src/generic/cases/block_rewards.rs
```diff
@@ -0,0 +1,199 @@
...
+		utils::{
+			currency::cfg,
+			genesis::{self, Genesis},
+		},
+	},
+	utils::accounts::Keyring,
+};
+
+crate::test_for_runtimes!(all, block_rewards_api);
+fn block_rewards_api<T: Runtime>() {
```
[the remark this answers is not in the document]
cdamian: That would be awesome!

runtime/integration-tests/src/generic/cases/block_rewards.rs
```diff
@@ -0,0 +1,202 @@
...
+	for collator in &leaving {
+		pallet_collator_selection::Pallet::<T>::leave_intent(
+			RawOrigin::Signed(collator.clone()).into(),
+		)
+		.unwrap();
+	}
+
+	frame_system::Pallet::<T>::reset_events();
+
+	pallet_session::Pallet::<T>::rotate_session();
```
lemunozm, line 123: BTW, I've played with this trying to substitute `pass_n(Period)` to `rotate_session()`, which maps to the same behavior and allow us getting rid of `fast-runtime` feature in integration-test and simplify a bit the test.

If you have some thoughts against it @wischli , I can revert this.

runtime/integration-tests/src/generic/cases/block_rewards.rs
```diff
@@ -0,0 +1,202 @@
...
+
+	// But later, if manually I create the event as follows:
+	let e = T::RuntimeEventExt::from(pallet_rewards::Event::<T, BlockRewards>::GroupRewarded {
+		group_id: 1,
+		amount: 2,
+	});
+
+	// And I call try_into(), it works.
+	let re: pallet_rewards::Event<T, BlockRewards> = e.try_into().ok().unwrap();
+	*/
```
lemunozm, line 146: **99% of developers** can not fix this nor saying what's happening here

cc: @cdamian @wischli @mustermeiszer

runtime/common/src/lib.rs
```diff
@@ -43,6 +43,11 @@ use sp_runtime::{
 };
 use sp_std::marker::PhantomData;
 
+pub mod instances {
+	/// The rewards associated to block rewards
+	pub type BlockRewards = pallet_rewards::Instance1;
```
wischli, line 48: Love it, thanks!

runtime/integration-tests/src/generic/cases/block_rewards.rs
wischli: In terms of `pallet_collator_selection::Config`, Dev and Altair/Centrifuge differ in the `ValidatorRegistration` trait implementor: Dev uses `Sessions` whereas production uses the `CollatorAllowlist`. Therefore, dev only requires session keys to be set for those addresses, which are registered. Production however requires the accounts to be whitelisted via the `pallet_collator_allowlist` beforehand. This can done via the `add` extrinsic of that pallet. This also requires the address to have their session key set.

runtime/integration-tests/src/generic/cases/block_rewards.rs
wischli: I don't have a solution at hand right now, would have to debug that myself. Can look into that next week, if desired.

runtime/integration-tests/src/utils/accounts.rs
```diff
@@ -249,65 +130,6 @@ impl std::str::FromStr for Keyring {
...
-///
-/// Accounts:
-/// * Keyring::Admin
-/// * Keyring::Alice
-/// * Keyring::Bob
-/// * Keyring::Ferdie
-/// * Keyring::Charlie
-/// * Keyring::Dave
-/// * Keyring::Eve
-/// * Keyring::TrancheInvestor(1)
```
wischli, line 261: I actually liked the comment. Of course we dont need a line for every single tranche investor but do you think we could bring back something like Admin, ... Eve, TranchInvestor(1), ..., TrancheInvestor(50)?

wischli (commented):
Thanks for slimming this down and making it much more readable! I hope one of us is the 1% dev, which can solve the mystery. I can't dive into this for now, but next week I would have capacity.

runtime/integration-tests/src/generic/cases/block_rewards.rs
lemunozm: Thanks so much for this answer @wischli, it helps me a lot!

runtime/integration-tests/src/utils/accounts.rs
lemunozm: Sure, but I'm just curious, what add the comment that does not tell you the own implementation?

runtime/integration-tests/src/generic/cases/block_rewards.rs
lemunozm: Thanks William. My guess is that something weird is happening with the `try_into` when events contain instances. But I've checked all the places where the instance is set, and we always set the correct one for that pallet. 🤔

runtime/integration-tests/src/utils/accounts.rs
wischli: Fair point. I often leverage rustdocs from hovering over. In this case, it is crystal clear so lets keep as is!

runtime/integration-tests/src/generic/cases/block_rewards.rs
lemunozm: @wischli, do you think we should align this for all runtimes? Why we are not using collator allowlist for dev?

cc @mustermeiszer

wischli: Functionally, it does not make sense to me to add an extra gate to collators for our dev runtime IMO. I suppose the only achievement for us would be that this test works for all runtimes? 

I definitely dont want to block your proposal, just questioning the effort.

lemunozm: Nono, it's not needed for the test; we can discriminate the behavior per runtime (at least it seems so by now). I asked because `development` is our way of proving stuff that later will go to `centrifuge`. So maybe it deserves that the demo also uses collator allowlist. Just a thought!

lemunozm:
Ok, now it works for all runtimes. Thanks for the help @wischli.

Could we consider this ready to be merged? and leave the *99% of developers can not fix this nor saying what's happening here* issue as a tech-debt?

lemunozm:
**WARNING: after merging this block rewards now will use `Instance1` instead of `Instance2` in development runtime. This would imply we need to do a reset in dev/demo in the next RU for them.

wischli:
> **WARNING: after merging this block rewards now will use `Instance1` instead of `Instance2` in development runtime. This would imply we need to do a reset in dev/demo in the next RU for them.**

Its definitely fine to reset on Dev. Do you want to open another PR for the cleanup and reset or do it in this one? The reset might require to execute the `pallet_block_rewards::migration::InitBlockRewards` again.

lemunozm:
I would say to open a new PR. Too many things in this one.

wischli (approved):
LGTM!

mustermeiszer (approved):
Blind approval after review of others.

lemunozm:
Thanks both!

wischli:
> I would say to open a new PR. Too many things in this one.
> 
> `InitBlockRewards` migration is done each time we reset?

I think it is because it attaches the staking currency to the group of the underlying `pallet_rewards::<Instance>`. I will handle this! I noticed Demo is super outdated with lots of pallet versions and I am working on fixing all of that right now.

wischli:
> Thanks William!
> 
> I was wondering... if we want to avoid resetting the storage in the demo to not lose everything set. Could we instead modify the politic always to reset demo but have a script that later it populates with some sane state for Apps (by adding a bunch of transactions). Just thinking out loud.

Let's discuss this at the offsite. If, Dev should be reset but Demo needs to be stable because of Axelar. I think repopulating state is quite intense as it requires multiple pools to be spawned etc. But given the current state of Demo, we might have to do a reset. It has never been planned to use that chain as our main testnet such that migrations have been ignored throughout numerous Subtrate and business logic updates. 

There are many many keys which cannot be decoded anymore..

PR #1806, opened 2024-04-11
Polkadot v1.1.0: Account converter changes

lemunozm:
# Description

From this discussion: https://github.com/centrifuge/centrifuge-chain/pull/1756#discussion_r1559105772

A refactor of `AccountConverter`:
- All methods belong to `AccountConverter` type.
- If the method requires the type to have a runtime type, then it uses the wrapper `RuntimeAccountConverter` type. This should only be used by trait demands in pallet configurations. For simple calls, use always `AccountConverter`

Are we in a middle point with this PR @mustermeiszer?


mustermeiszer (approved):
@lemunozm thanks, thats perfect!! ❤️

PR #1807, opened 2024-04-11
fix: chain spec not deterministic

wischli:
# Description

As part of my #1756 client checks, I noticed that our ephemeral chain specs did not lead to a deterministic chain spec which blocks all our scripts to register such parachain. The root of this was introduced by myself in #1740 by adding a unix timestamp for the block rewards. Since this timestamp is completely irrelevant for ephemeral chains, I changed it to the default value. This will mint a larger amount of treasury inflation in the first session and then go back to normal.

# Checklist:


lemunozm:
I remember something about this; thanks for fixing this @wischli !

lemunozm (approved):

wischli:
> I remember something about this; thanks for fixing this @wischli !

Yes you were not a fan of the reverted code path and thus right already back then 🫡

PR #1809, opened 2024-04-12
fix: pool fee event dispatch order

wischli:
# Description

* Context: https://centrifugedao.slack.com/archives/C04R42WGM7U/p1710854795914679
* Changes event order dispatch of pool creation and pool fees added when creating a new pool: Fee event needs to be after pool creation one

# Checklist:


lemunozm (approved):
Clean and with its own test. Perfect!!

hieronx (commented):
Changes re the pool fee event look good!

Something I realized that is completely unrelated to this PR, but was looking at the `pool-registry` code: shouldn't we emit an `MetadataSet` event here: https://github.com/centrifuge/centrifuge-chain/blob/9618071df487ebdbad1aafe17605b77f91a85b60/pallets/pool-registry/src/lib.rs#L289 Now we are only emitting it on updates,  which makes tracking all metadata harder as you need separate logic to track the first one.

If we do that, we should emit this event after the `T::ModifyPool::create` call

lemunozm (dismissed):
LGTM!!

pallets/pool-system/src/impls.rs
```diff
@@ -154,24 +154,6 @@ impl<T: Config> PoolMutate<T::AccountId, T::PoolId> for Pallet<T> {
...
-			// currency. Otherwise, we'd always need to convert investments to the decimals
-			// of tranche tokens and vice versa
-			let decimals = match T::AssetRegistry::metadata(&currency) {
-				Some(metadata) => metadata.decimals,
-				None => return Err(Error::<T>::MetadataForCurrencyNotFound.into()),
-			};
-
-			let metadata = tranche.create_asset_metadata(decimals, token_name, token_symbol);
-
-			T::AssetRegistry::register_asset(Some(tranche.currency.into()), metadata)
```
lemunozm, line 171: Why is it no longer registered?

wischli: It is! This was just moved down a few lines (L208+).

wischli: The reason why we want to move this closure is to change the dispatch order of the events: Pool creation should be first and then the tranche token registration events.

lemunozm (approved):

PR #1810, opened 2024-04-15
Fix: mostly non failable runtime-api

mustermeiszer:
# Description
The NAV runtime-apis need to be super stable for the UI. This defaults them to use the latest computed price if the latest calculation fails. 

## Changes and Descriptions
* use previously computed valuation amount for `nav_loans`
* use previously computed valuation amount for `nav_fees`
 
# Checklist:

### DIFF
--- a/pallets/pool-fees/src/lib.rs
+++ b/pallets/pool-fees/src/lib.rs
@@ -211,6 +211,7 @@ pub mod pallet {
 	///
 	/// Lifetime of a storage entry: Forever, inherited from pool lifetime.
 	#[pallet::storage]
+	#[pallet::getter(fn portfolio_valuation)]
 	pub(crate) type PortfolioValuation<T: Config> = StorageMap<
 		_,
 		Blake2_128Concat,
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -139,12 +139,19 @@ where
 {
 	let mut pool = pallet_pool_system::Pool::<T>::get(pool_id)
 		.ok_or(pallet_pool_system::Error::<T>::NoSuchPool)?;
-	let (nav_loans, _) =
-		pallet_loans::Pallet::<T>::update_portfolio_valuation_for_pool(pool_id, price_input)?;
+
+	let nav_loans =
+		pallet_loans::Pallet::<T>::update_portfolio_valuation_for_pool(pool_id, price_input)
+			.map(|(nav_loans, _)| nav_loans)
+			.unwrap_or_else(|| pallet_loans::Pallet::<T>::portfolio_valuation(pool_id).value());
+
 	let (nav_fees, _) = pallet_pool_fees::Pallet::<T>::update_portfolio_valuation_for_pool(
 		pool_id,
 		&mut pool.reserve.total,
-	)?;
+	)
+	.map(|(nav_fees, _)| nav_fees)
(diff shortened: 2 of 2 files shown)


runtime/common/src/lib.rs
```diff
@@ -139,12 +139,19 @@ where
 {
 	let mut pool = pallet_pool_system::Pool::<T>::get(pool_id)
 		.ok_or(pallet_pool_system::Error::<T>::NoSuchPool)?;
-	let (nav_loans, _) =
-		pallet_loans::Pallet::<T>::update_portfolio_valuation_for_pool(pool_id, price_input)?;
+
+	let nav_loans =
+		pallet_loans::Pallet::<T>::update_portfolio_valuation_for_pool(pool_id, price_input)
+			.map(|(nav_loans, _)| nav_loans)
+			.unwrap_or_else(|| pallet_loans::Pallet::<T>::portfolio_valuation(pool_id).value());
```
lemunozm, line 146: Note that `update_portfolio_valuation_for_pool()` overrides the `PortfolioValuation` storage, so calling `poltfolio_valuation()` will not return any different from what the first call has set.

You need to call `portfolio_valuation()` before line 144.

lemunozm (changes requested):

runtime/common/src/lib.rs
mustermeiszer: The unwrap only takes place if the update failed, right?

lemunozm: `update_portfolio_valuation_for_pool` is not transactional, and it can fail, leaving some storage written. Right now, seeing the implementation, it's not the case, but we can break this required "invariant" silently.

lemunozm: > so calling poltfolio_valuation() will not return any different from what the first call has set.

But yeah, what I said regarding this is not true. Although could be true in a future refactor

pallets/pool-fees/src/lib.rs
```diff
@@ -211,6 +211,7 @@ pub mod pallet {
 	///
 	/// Lifetime of a storage entry: Forever, inherited from pool lifetime.
 	#[pallet::storage]
+	#[pallet::getter(fn portfolio_valuation)]
```
lemunozm, line 214: FYI, this will be deprecated in future polkadot versions. We should use: `pallet_loans::PortfolioValuation::<T>::get()` instead.

https://github.com/paritytech/polkadot-sdk/issues/223

runtime/common/src/lib.rs
mustermeiszer: So what would you suggest?

pallets/pool-fees/src/lib.rs
mustermeiszer: We can remove it then IMO.

runtime/common/src/lib.rs
lemunozm: Something like:
```rust
let prev_nav_loans = pallet_loans::Pallet::<T>::portfolio_valuation(pool_id).value();
let nav_loans = pallet_loans::Pallet::<T>::update_portfolio_valuation_for_pool(pool_id, price_input)
    .map(|(nav_loans, _)| nav_loans)
    .unwrap_or_else(|_| prev_nav_loans);
```

mustermeiszer: fixed @lemunozm

lemunozm (approved):

wischli (commented):
Too late for the party.. thanks for fixing this!

PR #1812, opened 2024-04-15
Fix/linear pricing

mustermeiszer:
# Description
Currently, linear pricing is only used with the latest settlement price. Rather we want the linear pricing to be used IF the feature is enabled and then
* IF oracle use on oracle value
* IF NO valid oracle use on latest settlement price

## Changes and Descriptions
* Change linear pricing behaviour
* Feature Creep: Ensure epoch closing is only possible for liquidity and pool admins
  *NOTE: We need that is with the current setup Anemoy will set the validity of its oracles to always be valid which would allow anyone to close an epoch.*

## Communicate with Apps
* Need to set oracle collection ages to `None`
* ONLY pool admins or liquidity pool admins should be able to see `close epoch` fields
* `ExternalPricing` changed. There is now an additional field `with_linear_pricing` that allows to toggle that feature

## Communicate with Embrio
* Indexing of `LoanEvent::Created` needs to be filtered by block after the upgrade to include the `with_linear_pricing` info 

# Checklist:


mustermeiszer:
@lemunozm either I did something wrong or https://github.com/centrifuge/centrifuge-chain/pull/1812/commits/7b4ab7499973e834c70cd3ddddb75b8f7f922b73 proofs that the implementation of `IntoSeconds` for `Millis` does not provide us with type safety. 

I think we should make them new types rather. Wdyt? 

Furthermore, I added this feature without needing to adapt a single test. Not sure, but would you assume unit-tests should cover this?

mustermeiszer:
@lemunozm I am not sure if I implemented it wrongly, but I can not figure out how to fix the test. ^^

lemunozm:
> I think we should make them new types rather. Wdyt?

Yeah, we should use different types. That was something pending I had in mind to do it.

> I can not figure out how to fix the test

I'll check it!

pallets/loans/src/lib.rs
```diff
@@ -116,7 +116,7 @@ pub mod pallet {
 
 	pub type PortfolioInfoOf<T> = Vec<(<T as Config>::LoanId, ActiveLoanInfo<T>)>;
 	pub type AssetOf<T> = (<T as Config>::CollectionId, <T as Config>::ItemId);
-	pub type PriceOf<T> = (<T as Config>::Balance, <T as Config>::Moment);
+	pub type PriceOf<T> = (<T as Config>::Balance, Seconds);
```
lemunozm, line 119: When in `pallet_loans` we use `<T as Config>::Moment` is because we do not use seconds, or we do not know the time unit. In this case, this value comes from whatever the oracle configures for time, which is milliseconds.

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -163,10 +169,26 @@ impl<T: Config> ExternalActivePricing<T> {
...
+		)
+	}
+
+	fn current_price_inner(
+		&self,
+		maturity: Seconds,
+		oracle: Option<PriceOf<T>>,
+	) -> Result<T::Balance, DispatchError> {
+		if let Some((oracle_price, oracle_provided_at)) = oracle {
+			self.linear_accrual_price(maturity, oracle_price, oracle_provided_at.into_seconds())
```
lemunozm, line 184: Here, if the price comes from `T::PriceRegistry::get(&self.info.price_id, &pool_id).ok()` we're converting milliseconds to seconds. If the price comes from the cache, that was previously converted to seconds, we're converting to seconds again, which means divide by `1000`. Maybe this is related to the test issue you're experiencing.

Definitely, we need typed times to avoid this issue...

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -150,9 +151,14 @@ impl<T: Config> ExternalActivePricing<T> {
...
-	fn linear_accrual_price(&self, maturity: Seconds) -> Result<T::Balance, DispatchError> {
+	fn linear_accrual_price(
+		&self,
+		maturity: Seconds,
+		price: T::Balance,
+		price_last_updated: Seconds,
+	) -> Result<T::Balance, DispatchError> {
 		Ok(cfg_utils::math::y_coord_in_rect(
-			(self.settlement_price_updated, self.latest_settlement_price),
+			(price_last_updated, price),
```
lemunozm, line 161: This was wrong before. `price_last_updated` was in `millis` and maturity in `seconds`. I think maintaining everything as `<T as Config>::Moment` and just add `into_seconds()` here it should work.

pallets/loans/src/lib.rs
```diff
@@ -1071,7 +1071,10 @@ pub mod pallet {
 			let rates = T::InterestAccrual::rates();
 			let prices = match input_prices {
 				PriceCollectionInput::Empty => BTreeMap::default(),
-				PriceCollectionInput::Custom(prices) => prices.into(),
+				PriceCollectionInput::Custom(prices) => {
```
lemunozm, line 1074: Could make sense to receive the time from the input?
```rust
PriceCollectionInput::Custom(prices, now); //In the oracle denomination
```

lemunozm: I think it's the less invasive change, and also allows us to emulate old prices

mustermeiszer: mmh, on a second thought, might makes sense.

pallets/loans/src/entities/pricing/external.rs
mustermeiszer: Good catch!

pallets/loans/src/lib.rs
mustermeiszer: Yeah, but I collected it as seconds.

pallets/loans/src/tests/portfolio_valuation.rs
```diff
@@ -198,14 +198,16 @@ fn with_unregister_price_id_and_oracle_not_required() {
...
 
 		// Suddenty, the oracle set a value
+		const MARKET_PRICE_VALUE: Balance = 999;
 		MockPrices::mock_collection(|_| {
 			Ok(MockDataCollection::new(|_| {
-				Ok((PRICE_VALUE * 8, BLOCK_TIME_MS))
+				Ok((MARKET_PRICE_VALUE, BLOCK_TIME_MS))
 			}))
 		});
+		let price_value_after_half_year = MARKET_PRICE_VALUE + (NOTIONAL - MARKET_PRICE_VALUE) / 2;
```
lemunozm, line 207: I like more how you have tested this!

runtime/common/src/lib.rs
```diff
@@ -103,9 +104,21 @@ where
...
-				DispatchError::Other("Map expected to fit as it is coming from loans itself.")
-			})?)
+			PriceCollectionInput::Custom(
+				BoundedBTreeMap::<
+					T::PriceId,
+					(
+						<T as pallet_pool_system::Config>::Balance,
+						<T as pallet_loans::Config>::Moment,
+					),
+					T::MaxActiveLoansPerPool,
```
lemunozm, line 114: Super NIT. I think you can use `BoundedBTreeMap::<_,_,_>` here.

runtime/common/src/lib.rs
```diff
@@ -103,9 +104,21 @@ where
 {
 	let input_prices: PriceCollectionInput<T> =
 		if let Ok(prices) = pallet_loans::Pallet::<T>::registered_prices(pool_id) {
-			PriceCollectionInput::Custom(prices.try_into().map_err(|_| {
-				DispatchError::Other("Map expected to fit as it is coming from loans itself.")
-			})?)
+			PriceCollectionInput::Custom(
```
lemunozm, line 107: Not sure if I'm missing something but this entire block is not just the same as: `PriceCollectionInput::FromRegistry`?

lemunozm (commented):
Much cleaner now!

Just a question to double-check

runtime/common/src/lib.rs
mustermeiszer: No, unfortuntealy we need this in order to decide ourselves to only use the registry sources if it is not failing. If it is failing, we are going with empty prices. If we forwared `FromRegistry` then the whole call would fail.

lemunozm: But just for the case where we're sure it doesn't fail. We can add `FromRegistry`

lemunozm:
```rust
if let Ok(prices) = pallet_loans::Pallet::<T>::registered_prices(pool_id) {
    PriceCollectionInput::FromRegistry,
} else {
    PriceCollectionInput::Empty
};
```

mustermeiszer: Nice!

lemunozm (dismissed):
Thanks for tackle my suggestions!

lemunozm:
Probably `clippy` will complain with the `if let/else` structure to change it to a `match`

pallets/loans/src/entities/input.rs
```diff
@@ -61,6 +61,6 @@ impl<T: Config> RepaidInput<T> {
 #[scale_info(skip_type_params(T))]
 pub enum PriceCollectionInput<T: Config> {
 	Empty,
-	Custom(BoundedBTreeMap<T::PriceId, T::Balance, T::MaxActiveLoansPerPool>),
+	Custom(BoundedBTreeMap<T::PriceId, (T::Balance, T::Moment), T::MaxActiveLoansPerPool>),
```
lemunozm, line 64:
```suggestion
	Custom(BoundedBTreeMap<T::PriceId, PriceOf<T>, ::MaxActiveLoansPerPool>),
```

To fix the clippy issue

lemunozm (approved):

lemunozm (dismissed):

wischli (approved):

lemunozm:
Thanks for the awesome PR description! ❤️ Very helpful

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -76,6 +77,9 @@ pub struct ExternalPricing<T: Config> {
 	/// borrow/repay and the current oracle price.
 	/// See [`ExternalAmount::settlement_price`].
 	pub max_price_variation: T::Rate,
+
+	/// If the pricing is estimated with a linear pricing model.
+	pub with_linear_pricing: bool,
```
lemunozm, line 82: With this, I think now is much more understandable what's happening 🎉

mustermeiszer:
for me for the migration: https://github.com/centrifuge/centrifuge-chain/pull/1739/files#diff-637356fbd7809c17265d6b8ab519abc2b0ffa4df672cfb33a321795190f441baR580

pallets/pool-system/src/mock.rs
```diff
@@ -385,7 +384,40 @@ parameter_types! {
...
+					if account == DEFAULT_POOL_OWNER {
+						Ok(())
+					} else {
+						Err(RawOrigin::Signed(account).into())
+					}
+				}
+				RawOrigin::None => Err(RawOrigin::None.into()),
+			},
+		)
+	}
```
lemunozm, line 411: Why do we need this special implementation? Do not we know always `Ok()` for testing?

pallets/loans/src/entities/pricing/external.rs
```diff
@@ -150,23 +154,52 @@ impl<T: Config> ExternalActivePricing<T> {
...
-			Ok(data) => data.0,
-			Err(_) => self.linear_accrual_price(maturity)?,
-		})
+		self.current_price_inner(
+			maturity,
+			T::PriceRegistry::get(&self.info.price_id, &pool_id).ok(),
+		)
+	}
+
+	fn current_price_inner(
```
lemunozm, line 185: The current on-chain behavior is:
- If oracle is found, use oracle price
- If not found, use linear accrual price.

That behavior can not be modeled right now. We need to choose between:
- Use oracle price or if not use settlement price (`with_linear_pricing = false`)
- Use always linear accrual price (`with_linear_pricing = true`)

Do we want this? Don't we still want to model the current behavior?

wischli:
> for me for the migration: https://github.com/centrifuge/centrifuge-chain/pull/1739/files#diff-637356fbd7809c17265d6b8ab519abc2b0ffa4df672cfb33a321795190f441baR580

@mustermeiszer I will do the migration and open a PR against this base. [Context](https://kflabs.slack.com/archives/C04GJQAM9P0/p1715856686427749?thread_ts=1715849878.667849&channel=C04GJQAM9P0&message_ts=1715856686.427749)

pallets/loans/src/entities/pricing/external.rs
lemunozm: cc @mustermeiszer

pallets/pool-system/src/mock.rs
mustermeiszer: Mmmh. Good point. If this is covered in the integration test, probably not.

pallets/loans/src/entities/pricing/external.rs
mustermeiszer: No, we want to change that too. 

Get the price: 
* IF oracle &rarr; use oracle
* ELSE &rarr; use settlement price

Use the price:
* IF linear pricing &rarr; use linear pricing 
* ELSE &rarr; just use price

lemunozm: What's the difference between "Get the price" and "Use the price"?

mustermeiszer: Get is either orale or settlement. Use is just for using whatever came out of the get.

lemunozm: I would add new 2 test cases to check a loan configured with `use_linear_price = true` always uses the linear accrual with oracle value and with settlement price

lemunozm: Well, after the changes of this PR all loans comes with `use_linear_price = true`, so we should check the `false` version instead

pallets/loans/src/tests/repay_loan.rs
```diff
@@ -812,3 +812,63 @@ fn with_unregister_price_id_and_oracle_not_required() {
...
+		// AFTER: the loan overpassing maturity date
+		advance_time(DAY);
+
+		assert_ok!(Loans::repay(
+			RuntimeOrigin::signed(BORROWER),
+			POOL_A,
+			loan_id,
+			repay_amount.clone()
+		));
+		assert_eq!(current_price(), NOTIONAL);
```
lemunozm, line 872: Checked 3 cases, just before, just at the maturity date and just after

@mustermeiszer

pallets/loans/src/tests/repay_loan.rs
```diff
@@ -900,3 +902,79 @@ fn with_external_pricing() {
 		assert_eq!(current_price(), NOTIONAL);
 	});
 }
+
+#[test]
+fn with_more_than_required_external() {
```
lemunozm, line 907: Super NIT, but can we move this to line 262 to be closed to `with_more_than_required()` that works similar but with internal pricing?

lemunozm (commented):
Some few minor and NITs comments, but LGTM!

pallets/loans/src/tests/portfolio_valuation.rs
```diff
@@ -220,3 +220,63 @@ fn empty_portfolio_with_current_timestamp() {
...
+		update_portfolio();
+		expected_portfolio(QUANTITY.saturating_mul_int(SETTLEMENT_PRICE));
+
+		MockPrices::mock_collection(|_| {
+			Ok(MockDataCollection::new(|_| {
+				Ok((MARKET_PRICE_VALUE, BLOCK_TIME_MS))
+			}))
+		});
+		update_portfolio();
+		expected_portfolio(QUANTITY.saturating_mul_int(MARKET_PRICE_VALUE));
```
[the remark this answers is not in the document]
mustermeiszer: No, I want to check, that it is getting back to the oracle afterwards.

lemunozm (approved):
Thank you so much for tackling my Nits ❤️

PR #1813, opened 2024-04-17
Runtimes: use last benchmark structure, remove unused disable-api feature

lemunozm:
# Description

- Use the new way of setting a benchmarking list.
  - We no longer need an explicit whitelist. See this [thread](https://kflabs.slack.com/archives/C05TBFRBL15/p1713292059995299). I've tested anyways and the keys are in. 
- Remove the unused `disable-runtime-api` feature. (no longer used by Polkadot).


lemunozm:
I had it in the stash, and it was time to upload it 😆

PR #1814, opened 2024-04-17
Upgrade gh actions

gpmayorga:
# Description
The customary upgrade of Github Actions using Dependabot suggestions


lemunozm:
From the changes, I'm not able to know where the new six CI jobs come from. Are they internally added by the new dependencies updated?

gpmayorga:
the new checks aren't new, it is just that when an actions workflow file changes it will trigger that action for the current PR precisely to test if the workflow itself is now broken after such change, in other PRs, if the workflow file did not change, then it will only trigger the worfklows with the `pull_request:` trigger on them
Example of conditional pull request trigger for Docker build:
https://github.com/centrifuge/centrifuge-chain/actions/runs/8725365260/workflow?pr=1814#L11-L16

gpmayorga:
@wischli or @lemunozm 
PING - this should be a no-op

lemunozm (approved):
LGTM!

wischli (approved):
Thanks for handling this!

PR #1815, opened 2024-04-17
fix: looser `is_local_representation`

wischli:
# Description

Fixes issue which blocks FRAX <> USDC order fulfilments on Demo.

The current implementation of `is_local_representation` is too strict such that it throws an error if the checked `CurrencyId` has not configured a local representation. Instead, it should return `Ok(false)`. This function has not been tested directly. 

To the best of my knowledge, this does not affect production for now as it only affects checking currencies which don't have a configured local asset, e.g. FRAX.

# Checklist:

### DIFF
--- a/libs/types/src/tokens.rs
+++ b/libs/types/src/tokens.rs
@@ -431,18 +431,16 @@ where
 		let meta_variant =
 			AssetInspect::metadata(variant_currency).ok_or(DispatchError::CannotLookup)?;
 
-		let local: Self = meta_variant
-			.additional
-			.local_representation
-			.ok_or(DispatchError::Other("Missing local representation"))?
-			.into();
+		if let Some(local) = meta_variant.additional.local_representation {
+			frame_support::ensure!(
+				meta_local.decimals == meta_variant.decimals,
+				DispatchError::Other("Mismatching decimals")
+			);
 
-		frame_support::ensure!(
-			meta_local.decimals == meta_variant.decimals,
-			DispatchError::Other("Mismatching decimals")
-		);
-
-		Ok(self == &local)
+			Ok(self == &local.into())
+		} else {
+			Ok(false)
+		}
 	}
 }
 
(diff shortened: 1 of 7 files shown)


libs/types/src/tokens.rs
```diff
@@ -431,18 +431,16 @@ where
...
+		if let Some(local) = meta_variant.additional.local_representation {
+			frame_support::ensure!(
+				meta_local.decimals == meta_variant.decimals,
+				DispatchError::Other("Mismatching decimals")
+			);
 
-		Ok(self == &local)
+			Ok(self == &local.into())
+		} else {
+			Ok(false)
```
lemunozm, line 442: Nice found!

pallets/token-mux/src/tests.rs
```diff
@@ -506,3 +506,61 @@ pub(crate) mod swaps {
...
+	use cfg_types::tokens::CurrencyId;
+	use sp_runtime::DispatchError;
+
+	use super::*;
+	use crate::mock::{new_test_ext_invalid_assets, MockRegistry, USDC_WRONG_DECIMALS};
+
+	#[test]
+	fn is_local_happy_paths() {
+		new_test_ext().execute_with(|| {
+			assert_eq!(<CurrencyId as HasLocalAssetRepresentation<MockRegistry>>::is_local_representation_of(&USDC_LOCAL, &USDC_1), Ok(true));
```
lemunozm, line 521: Just curious,

```rust
CurrencyId::is_local_representation_of()
```

Does not compile? Does the compiler force you to specify the trait?

lemunozm (approved):
Thanks for researching this @wischli, and for adding related tests

mustermeiszer (approved):
Thanks for the hustle! Nice found!

mustermeiszer (approved):

pallets/token-mux/src/tests.rs
wischli: Unfortunately not, the `AssetInspect` parameter type cannot be inferred. Tried different variants but none of the shortcuts worked.

PR #1816, opened 2024-04-17
Fix: zero amount order bug

lemunozm:
# Description

There are two casuistics where an order can contain a zero amount. Such an order can not be closed, and theoretically, it should not exist.

- [x] Added test cases that generate the error. All newly added tests fail without the fix.
- [x] Added fix bypassing creating orders when they have 0 amount.

Q: Should we allow to close zero orders even though they should not be possible now?

### DIFF
--- a/pallets/swaps/src/tests.rs
+++ b/pallets/swaps/src/tests.rs
@@ -490,3 +490,87 @@ mod fulfill {
 		});
 	}
 }
+
+mod zero_amount_order {
+	use super::*;
+
+	#[test]
+	fn when_apply_over_no_swap() {
+		new_test_ext().execute_with(|| {
+			MockTokenSwaps::mock_place_order(|_, _, _, _, _| {
+				panic!("this mock should not be called");
+			});
+
+			assert_ok!(
+				<Swaps as TSwaps<AccountId>>::apply_swap(
+					&USER,
+					SWAP_ID,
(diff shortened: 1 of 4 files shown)


pallets/swaps/src/tests.rs
```diff
@@ -490,3 +490,87 @@ mod fulfill {
...
+				}
+			);
+
+			assert_eq!(OrderIdToSwapId::<Runtime>::get(ORDER_ID), None);
+			assert_eq!(SwapIdToOrderId::<Runtime>::get((USER, SWAP_ID)), None);
+		});
+	}
+
+	#[test]
+	fn when_apply_over_smaller_inverse_swap_but_math_precission() {
```
lemunozm, line 526: We do not really need this test having `pallet-swaps` tests. But I wanted to force a use case from FI PoV. We can finally remove it once we're sure it's fully solved.

mustermeiszer (commented):
Thanks!! @lemunozm do you think this fix has to live here or we could think about fixing it on the `pallet-order-book` side? 

I am also not sure why having a zero order lead to a missing FI state. Do you know the reason for that or in general explain what the issue was? I mean `pallet-order-book` does fulfill zero orders, right?

mustermeiszer:
Makes sense! Thanks

PR #1821, opened 2024-04-23
Integration test: making it cooler!

lemunozm:
# Description

```rust
#[test_runtimes([development, altair, centrifuge])]
fn foo<T: Runtime> {
    // Your test here...
}
```
or 
```rust
#[test_runtimes(all)]
fn foo<T: Runtime> {
    // Your test here...
}
```

## Benefits
- It's prettier!
- Better readability 
- We no longer confuse an utility method with the same "shape" as foo from an actual test method


runtime/integration-tests/src/generic/cases/example.rs
```diff
@@ -70,6 +71,7 @@ fn transfer_balance<T: Runtime>() {
 }
 
 // Identical to `transfer_balance()` test but using fudge.
+#[test_runtimes([development, altair, centrifuge])]
```
wischli, line 74: Nit:

```suggestion
#[test_runtimes(all)]
```

wischli (dismissed):
Awesome improvement 🚀

runtime/integration-tests/src/generic/cases/example.rs
```diff
@@ -16,6 +16,7 @@ use crate::{
 	utils::accounts::Keyring,
 };
 
+#[test_runtimes([development, altair, centrifuge])]
```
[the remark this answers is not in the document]
lemunozm: Yeah! I left this on purpose as an example of different runtimes you can set (not that this belongs to `example.rs`)

runtime/integration-tests/src/generic/cases/example.rs
lemunozm: Same as before

mustermeiszer (dismissed):
🚀

lemunozm:
@wischli @mustermeiszer @cdamian Can I have a reapproval?

PR #1833, opened 2024-05-09
Upgrade to Polkadot-v1.7.2

lemunozm:
# Description

### Checklist:

- [x] Upgraded all dependencies to `polkadot-sdk` with version `1.7.2`.
  - [x] **NOTE**: we can not use the original orml repo due to https://github.com/open-web3-stack/open-runtime-module-library/issues/1000, using by now the fork of moonbeam for that version.
- [x] `cargo check` & `cargo test` using `runtime-benchmarks` feature
  - [x] libs
  - [x] pallets
    - [x] anchors
    - [x] block-rewards
    - [x] bridge
    - [x] collator-allowlist
    - [x] ethereum-transaction
    - [x] fees
    - [x] foreign-investments
    - [x] interest-accrual
    - [x] investments
    - [x] keystore
    - [x] liquidity-pools
    - [x] liquidity-pools-gateway
      - [x] liquidity-pools-gateway-routers
      - [x] axelar-gateway-precompile
    - [x] liquidity-rewards
    - [x] loans
    - [x] oracle-collection
    - [x] oracle-feed
    - [x] order-book
    - [x] permissions
    - [x] pool-fees
    - [x] pool-registry
    - [x] pool-system
    - [x] restricted-tokens
    - [x] restricted-xtokens
    - [x] rewards
    - [x] swaps
    - [x] token-mux
    - [x] transfer-allowlist
  - [x] runtimes
    - [x] common
    - [x] development
    - [x] altair
    - [x] centrifuge
  - [x] integration-tests
  - [x] node
    - [x] Currently using some deprecated code. This needs to be fixed.
- [x] use latest fudge
- [x] check all required patched dependencies are patched
- [x] check real benchmarks
- [x] check try-runtime
- [x] check docker files
- [x] modify CI and scripts
- [x] Evaluate if we need to perform migrations from 3parties
- [x] Check node usage
- [x] Check any missing TODO in the PR
- [x] Ensure `cargo update` using latest polkadot


pallets/restricted-tokens/src/lib.rs
```diff
@@ -202,7 +202,7 @@ pub mod pallet {
 			+ LockableCurrency<Self::AccountId>
 			+ ReservableCurrency<Self::AccountId>
 			+ fungible::Inspect<Self::AccountId, Balance = Self::Balance>
-			+ fungible::InspectHold<Self::AccountId, Reason = ()>
+			+ fungible::InspectHold<Self::AccountId>
```
lemunozm, line 205: Super :Warn::

We would need to migrate the storage from using `()` as an index of hold reason to any other variant. `()` is no longer supported 😢

pallets/restricted-tokens/src/lib.rs
```diff
@@ -493,49 +492,22 @@ pub mod pallet {
 			who: <T::Lookup as StaticLookup>::Source,
 			currency_id: T::CurrencyId,
 			#[pallet::compact] new_free: T::Balance,
-			#[pallet::compact] new_reserved: T::Balance,
```
lemunozm, line 496: Why this: https://kflabs.slack.com/archives/C05TBFRBL15/p1715597422917969

Open to discussion

libs/mocks/src/asset_registry.rs
```diff
@@ -4,7 +4,7 @@ pub mod pallet {
 	use mock_builder::{execute_call, register_call};
 	use orml_traits::asset_registry::{AssetMetadata, Inspect, Mutate};
 	use sp_std::fmt::Debug;
-	use staging_xcm::{v3::prelude::MultiLocation, VersionedMultiLocation};
+	use staging_xcm::{v4::Location, VersionedLocation};
```
lemunozm, line 7: :warn: The asset registry's API has been changed to V4. Maybe we need to migrate our asset registry to transform locations to V4.

Not sure if this is mandatory because the locations are behind a `VersionedLocation` allowing to exists different versions. But if a several places in our code we expect they content an specific version, so if we do not migrate we need to check for different versions in those places.

libs/types/src/locations.rs
```diff
@@ -63,7 +63,8 @@ mod test {
 
 	#[test]
 	fn from_xcm_versioned_address_works() {
-		let xa = VersionedMultiLocation::V3(MultiLocation::default());
+		// TODO-1.7: Must be changed to V4?
```
lemunozm, line 66: TODO here

pallets/liquidity-pools/src/lib.rs
```diff
@@ -888,14 +883,15 @@ pub mod pallet {
 				Error::<T>::AssetNotLiquidityPoolsTransferable
 			);
 
+			// TODO-1.7.2, should be v4?
```
lemunozm, line 886: We need either:
- Support v3 and v4
- Migrate to asset registry to v4 and only use v4

pallets/restricted-tokens/src/lib.rs
wischli: I will deal with this!

libs/types/src/locations.rs
```diff
@@ -33,21 +33,24 @@ pub enum Location {
...
 
-impl From<VersionedMultiLocation> for Location {
-	fn from(vml: VersionedMultiLocation) -> Self {
+impl From<VersionedLocation> for RestrictedTransferLocation {
+	fn from(vml: VersionedLocation) -> Self {
 		// using hash here as multilocation is significantly larger than any other enum
 		// type here -- 592 bytes, vs 40 bytes for domain address (next largest)
 		Self::XCM(BlakeTwo256::hash(&vml.encode()))
+
+		// TODO-1.7: I'm afraid of locations translated from v3 to v4 will generate a different hash here.
```
lemunozm, line 48: TODO here

pallets/restricted-tokens/src/lib.rs
```diff
@@ -202,7 +211,7 @@ pub mod pallet {
 			+ LockableCurrency<Self::AccountId>
 			+ ReservableCurrency<Self::AccountId>
 			+ fungible::Inspect<Self::AccountId, Balance = Self::Balance>
-			+ fungible::InspectHold<Self::AccountId, Reason = ()>
+			+ fungible::InspectHold<Self::AccountId, Reason = Self::RuntimeHoldReason>
```
lemunozm, line 214: ⚠️ This will require migration because now the storage is indexed with `HoldReason::NativeIndex` instead of `()`.

I guess it will not be easy to know which amounts must be migrated to this value and which amounts must be migrated to the other variant `HoldReason::TrasnferAllowance` added in `pallet-transfer-alloance`. I think we only have those two possibilities. But not 100% if another pallet adds another variant to `RuntimeHoldReason` that we're using right now. TODO check this

lemunozm: We need to ensure that pallets using restricted tokens, which are:
- [ ] pallet-token-mux
- [ ] pallet-liquidity-pools
- [ ] pallet-order-book
- [ ] pallet-block-rewards
- [ ] pallet-rewards (BlockRewards instance)
- [ ] pallet-rewards (Liquidity instance)
- [ ] pallet-investments
- [ ] pallet-restricted-tokens (by extrinsics)
- [ ] pallet-pool-fees
- [ ] pallet-pool-system
- [ ] pallet-fees
- [ ] pallet-treasuty
- [ ] pallet-uniques
- [ ] pallet-vesting
- [ ] pallet-identity
- [ ] pallet-election-phragmen
- [ ] pallet-proxy

have not never used a hold call using `Currency::Native`.

If we can not ensure that, the migration will be very difficult to determine how to split the amount under `()` for each holding purpose of each pallet.

runtime/development/src/lib.rs
```diff
@@ -350,13 +398,13 @@ impl pallet_balances::Config for Runtime {
...
 	type MaxFreezes = ConstU32<10>;
-	type MaxHolds = ConstU32<10>;
 	type MaxLocks = MaxLocks;
 	type MaxReserves = MaxReserves;
 	type ReserveIdentifier = [u8; 8];
 	/// The overarching event type.
 	type RuntimeEvent = RuntimeEvent;
-	type RuntimeHoldReason = ();
+	type RuntimeHoldReason = RuntimeHoldReason;
+	type RuntimeFreezeReason = RuntimeFreezeReason;
```
lemunozm, line 407: It seems that we do not have any way of freezing things right now. It seems like other pallets do not use freeze. The `RuntimeFreezeReason` coming from the runtime will probably be `()`.

runtime/development/src/lib.rs
```diff
@@ -835,23 +888,29 @@ parameter_types! {
 	pub const MaxSubAccounts: u32 = 100;
 	pub const MaxAdditionalFields: u32 = 100;
 	pub const BasicDeposit: Balance = 100 * CFG;
-	pub const FieldDeposit: Balance = 25 * CFG;
+	pub const ByteDeposit: Balance = deposit(0, 1); //TODO-1.7
```
lemunozm, line 891: TODO: Check if we need to change this value

runtime/development/src/lib.rs
```diff
@@ -937,28 +997,32 @@ parameter_types! {
...
 	type SpendPeriod = SpendPeriod;
 	type WeightInfo = weights::pallet_treasury::WeightInfo<Runtime>;
+	type AssetKind = ();
+	type Beneficiary = Self::AccountId;
+	type BeneficiaryLookup = IdentityLookup<Self::Beneficiary>;
+	type Paymaster = PayFromAccount<Balances, TreasuryAccount>;
+	type BalanceConverter = UnityAssetBalanceConversion;
+	type PayoutPeriod = ConstU32<10>;
+	#[cfg(feature = "runtime-benchmarks")]
+	type BenchmarkHelper = ();
```
lemunozm, line 1025: pallet-treasury has change a lot allowing to transfer funds from the treasury [ref](https://github.com/paritytech/polkadot-sdk/pull/1333). 

By now, I've set this simple as if we are not using the new capabilities. I think if we want this, we should configure it later in a new PR after this upgrade to avoid mixing things up. Note that a correct configuration requires the `pallet_asset_rate` pallet to be added, so it's not trivial.

runtime/development/src/lib.rs
```diff
@@ -1720,10 +1786,10 @@ impl pallet_block_rewards::Config for Runtime {
 impl pallet_transfer_allowlist::Config for Runtime {
 	type CurrencyId = FilterCurrency;
 	type Deposit = AllowanceDeposit<Fees>;
-	type HoldId = HoldId;
-	type Location = Location;
+	type Location = RestrictedTransferLocation;
 	type ReserveCurrency = Balances;
 	type RuntimeEvent = RuntimeEvent;
+	type RuntimeHoldReason = RuntimeHoldReason;
```
lemunozm, line 1792: ⚠️  we probably need to migrate previous holds done by TransferAllowlist to the new `HoldReason` value

runtime/development/src/xcm.rs
```diff
@@ -71,7 +71,7 @@ impl frame_support::traits::Contains<RuntimeCall> for SafeCallFilter {
 				| RuntimeCall::PolkadotXcm(
 					pallet_xcm::Call::limited_reserve_transfer_assets { .. }
 				) | RuntimeCall::XcmpQueue(..)
-				| RuntimeCall::DmpQueue(..)
+				| RuntimeCall::MessageQueue(..)
```
lemunozm, line 74: Not sure about this.

runtime/development/src/lib.rs
```diff
@@ -283,27 +290,68 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
 	pub const ReservedDmpWeight: Weight = MAXIMUM_BLOCK_WEIGHT.saturating_div(4);
+	pub const RelayOrigin: AggregateMessageOrigin = AggregateMessageOrigin::Parent;
 }
 
 impl cumulus_pallet_parachain_system::Config for Runtime {
 	// Using AnyRelayNumber only for the development & demo environments,
 	// to be able to recover quickly from a relay chains issue
 	type CheckAssociatedRelayNumber = cumulus_pallet_parachain_system::AnyRelayNumber;
-	type DmpMessageHandler = DmpQueue;
+	type WeightInfo = (); //TODO-1.7 check weights
```
lemunozm, line 300: Now `cumulus_pallet_parachain_system` has been added to the benchmarking list so we ideally need to point to the autogenerated weights.

runtime/development/src/lib.rs
```diff
@@ -283,27 +290,68 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
-impl parachain_info::Config for Runtime {}
+impl staging_parachain_info::Config for Runtime {}
+
+parameter_types! {
+	pub MessageQueueServiceWeight: Weight = Perbill::from_percent(35) * RuntimeBlockWeights::get().max_block;
+}
+
+impl pallet_message_queue::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type WeightInfo = (); //TODO-1.7 check weights
```
lemunozm, line 319: Point to the autogenerated weights

runtime/development/src/lib.rs
```diff
@@ -2725,6 +2772,9 @@ impl_runtime_apis! {
 			use cumulus_pallet_session_benchmarking::Pallet as SessionBench;
 			impl cumulus_pallet_session_benchmarking::Config for Runtime {}
 
+			use pallet_xcm::benchmarking::Pallet as PalletXcmExtrinsicsBenchmark;
+			impl pallet_xcm::benchmarking::Config for Runtime {}
```
lemunozm, line 2776: FYI @wischli:

Independently if we go with a more complete solution or not, this fix by now the benchmark compilation and we can proceed with the upgrade pipeline

runtime/development/src/lib.rs
lemunozm: It seems the deposit is now computed by byte instead of by the entire field, so this constant must be less than before

pallets/restricted-tokens/src/lib.rs
lemunozm: Finally field restored using `HoldReason::NativeIndex`: [38377a9](https://github.com/centrifuge/centrifuge-chain/pull/1833/commits/38377a91aa8cc6c25f650ef9d975294579eae437)

runtime/common/src/lib.rs
```diff
@@ -381,44 +360,40 @@ pub mod investment_portfolios {
 			let amount = pallet_investments::InvestOrders::<T>::get(&investor, invest_id)
 				.map(|order| order.amount())
 				.unwrap_or_default();
-			let free_tranche_tokens_new: Balance = <pallet_restricted_tokens::Pallet<T> as fungibles::Inspect<
-				T::AccountId,
-			>>::reducible_balance(
+			let free_tranche_tokens_new = orml_tokens::Pallet::<T>::reducible_balance(
```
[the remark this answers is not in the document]
lemunozm: This change from `restricted_tokens` to `orml_tokens` is no longer needed since we have `HoldReason::NativeIndex` in `restricted_tokens`

libs/types/src/locations.rs
lemunozm: We're storing the `RestrictedTransferLocation` in `pallet-transfer-allowlist` so changing versions here will affect that storage

pallets/liquidity-pools/src/lib.rs
lemunozm: Current solution:
 - We convert any v3 xcm message to v4 because they are isometric, and we only treat in the code the v4 message.

libs/types/src/locations.rs
lemunozm: We need to ensure there is no `RestrictedTransferLocation::Xcm` stored here. If there are, we will need to tackle this problem.

runtime/development/src/lib.rs
lemunozm: Should we anyway check the accounts to ensure we're no freezing

libs/types/src/locations.rs
lemunozm: It could not be an issue if the person who called `add_transfer_allowance(location.v3)` calls `remove_transfer_allowance(location.v3)` with the same version.

runtime/development/src/lib.rs
lemunozm: Using the defensive weights

runtime/development/src/lib.rs
lemunozm: Using the recommended weights

libs/types/src/locations.rs
wischli: Unfortunately, there is one XCM location in the `AccountCurrencyTransferAllowance` storage:
```
key ["4dTeMxuPJCK7zQGhFcgCivSJqBs9Wo2SuMSQeYCCuVJ9xrE2",{"Specific":{"ForeignAsset":"6"}},{"XCM":"0x8938623071e6f6b3eebd4ae4587d3e71b9e94aa00e7cfdac883829cd003aa23f"}]
val {"allowedAt":"0","blockedAt":"4,294,967,295"}

key ["4dTeMxuPJCK7zQGhFcgCivSJqBs9Wo2SuMSQeYCCuVJ9xrE2","All",{"Address":{"EVM":["1","0x1a3b234fb570f93eb2244796f0527f834cd30294"]}}]
val {"allowedAt":"0","blockedAt":"4,294,967,295"}

key ["4btLaQXkQRkV6SfHk5VbHvb4Dzipbg4r2hRCfrDJYYiKCfHW",{"Specific":{"ForeignAsset":"100,001"}},{"Address":{"EVM":["1","0x1a3b234fb570f93eb2244796f0527f834cd30294"]}}]
val {"allowedAt":"0","blockedAt":"4,294,967,295"}
```

Since we cannot guarantee that new ones won't be added in between proposing and executing the RU, I suppose I should add a migration for this? I.e. decoding the V3 Location from the stored hash and storing as V4 hash.

Another solution would be adding this translation layer to the `From<VersionedLocation>` impl of `RestrictedTransferLocation`.q

runtime/development/src/lib.rs
wischli: I checked all accounts on all three chains: The `pallet_balances::Freeze` storage is empty for all of them.

lemunozm: Yeah!! Thanks!

libs/types/src/locations.rs
lemunozm: > decoding the V3 Location from the stored hash and storing as V4 hash.

And how you can know what was the initial location v3 struct? Not sure if we store these locations somewhere to rebuild them, translate them to v4 and make the hash again.

> Another solution would be adding this translation layer to the `From<VersionedLocation>` impl of `RestrictedTransferLocation`.

Could you expand?

lemunozm: > We need to set a max size then as locations can be arbitrarily large.

Good point to think about! 
In which sense it can be "arbitrarily"? It can have 9 elements as much (given by `X9`), and looking deeply they can not be nested

pallets/restricted-tokens/src/lib.rs
lemunozm: Solved!

libs/types/src/locations.rs
lemunozm: > so we can definitely due this manually for this case.

I guess we'll go with this option, right?

lemunozm: Or using directly a `Location` there. What do you prefer? I can do that

runtime/altair/src/xcm.rs
```diff
@@ -71,7 +71,7 @@ impl frame_support::traits::Contains<RuntimeCall> for SafeCallFilter {
 				| RuntimeCall::PolkadotXcm(
 					pallet_xcm::Call::limited_reserve_transfer_assets { .. }
 				) | RuntimeCall::XcmpQueue(..)
-				| RuntimeCall::DmpQueue(..)
+				| RuntimeCall::MessageQueue(..)
```
wischli, line 74: I think at this point we could drop this safe call filter. Polkadot is doing the same for system parachains: https://github.com/polkadot-fellows/runtimes/pull/285

The only reason holding me back here (i.e. _could_ wording) is that we use default benchmarks.

runtime/centrifuge/src/lib.rs
```diff
@@ -957,24 +996,30 @@ parameter_types! {
 	pub const MaxSubAccounts: u32 = 100;
 	pub const MaxAdditionalFields: u32 = 100;
 	pub const BasicDeposit: Balance = 100 * CFG;
-	pub const FieldDeposit: Balance = 25 * CFG;
+	pub const ByteDeposit: Balance = deposit(0, 1); //TODO-1.7
```
wischli, line 999: LGTM, can remove comment IMO.

runtime/common/src/lib.rs
```diff
@@ -793,3 +762,40 @@ pub mod rewards {
...
+		}
+	}
+
+	pub struct ParaIdToSibling;
+	impl sp_runtime::traits::Convert<ParaId, AggregateMessageOrigin> for ParaIdToSibling {
+		fn convert(para_id: ParaId) -> AggregateMessageOrigin {
+			AggregateMessageOrigin::Sibling(para_id)
+		}
+	}
+}
```
wischli, line 801: LGTM!

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -102,72 +107,12 @@ pub mod utils {
 		}
 	}
 
-	pub fn setup_xcm<T: Runtime + FudgeSupport>(env: &mut FudgeEnv<T>) {
```
wischli, line 105: Note: Moved here: https://github.com/centrifuge/centrifuge-chain/pull/1833/files#diff-3baaa18fad9b36a69774c73a025373847f78da6ff3458c3a088275ec1ee59f84

wischli (commented):
Looks really good already! Great work @lemunozm  🚀 I did not expect it to be this clean already. The majority of my comments are just approvals of your changes.

A few pallet configurations need to be changed. I only commented on the `Altair` runtime to keep the noise low but it corresponds for all three runtimes. Please let me know if I should do the changes myself and open a PR for that.

ci/run-check.sh
```diff
@@ -10,7 +10,7 @@ cargo --version
 
 case $TARGET in
   cargo-build)
-    cargo build -p centrifuge-chain --release "$@"
+    #cargo build -p centrifuge-chain --release "$@"
```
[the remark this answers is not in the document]
lemunozm: Thanks for the reminder because I forgot it completely 😆

pallets/block-rewards/src/migrations.rs
```diff
@@ -40,19 +40,6 @@ pub mod init {
 		PhantomData<(T, CollatorReward, AnnualTreasuryInflationPercent)>,
 	);
 
-	fn get_collators<T: pallet_collator_selection::Config>() -> Vec<T::AccountId> {
```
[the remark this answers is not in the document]
lemunozm: I think you were right at that moment, but now, there is a call that does exactly what you did

libs/types/src/locations.rs
wischli: > In which sense it can be "arbitrarily"? It can have 9 elements as much (given by `X9`), and looking deeply they can not be nested

Got me with my superficial knowledge 😅 Suppose then it's fine because it is bound by the `MaxEncodedLen` which will of course be a heavy overestimation.

runtime/altair/src/lib.rs
```diff
@@ -279,21 +293,63 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
+		staging_xcm_executor::XcmExecutor<XcmConfig>,
+		RuntimeCall,
+	>;
+	type QueueChangeHandler = NarrowOriginToSibling<XcmpQueue>;
+	type QueuePausedQuery = NarrowOriginToSibling<XcmpQueue>;
+	type RuntimeEvent = RuntimeEvent;
+	type ServiceWeight = MessageQueueServiceWeight;
+	type Size = u32;
+	type WeightInfo = ();
+}
```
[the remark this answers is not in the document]
lemunozm: The same configuration as the real polkadot IIRC

runtime/altair/src/lib.rs
```diff
@@ -884,20 +952,24 @@ parameter_types! {
...
 	type Burn = Burn;
-	// we burn and dont handle the unbalance
 	type BurnDestination = ();
 	type Currency = Tokens;
 	type MaxApprovals = MaxApprovals;
-	// slashed amount goes to treasury account
 	type OnSlash = Treasury;
 	type PalletId = TreasuryPalletId;
+	type Paymaster = PayFromAccount<Balances, TreasuryAccount>;
+	type PayoutPeriod = ConstU32<10>;
```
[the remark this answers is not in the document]
lemunozm: Good point! Ideally, I guess infinity is what it should be there, but we need some limit to free resources. 30 is ok? or do we prefer more?

runtime/altair/src/xcm.rs
lemunozm: The default benchmarks are computed with the "required" hardware. So I guess we're in the same magnitude. We could compute them by ourselves too, but we are not sure if we want such precision now

runtime/altair/src/lib.rs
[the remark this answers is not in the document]
wischli: 30 **days** is definitely sufficient, 30 **blocks** too fast for sure

runtime/integration-tests/src/generic/cases/liquidity_pools.rs
lemunozm: Yeah! Sorry for not indicating. `proxy.rs` also requires so it should be common to not link cases againts other cases

libs/types/src/locations.rs
lemunozm: Will do then!

lemunozm:
I'll tackle your changes, thanks for the review @wischli !

runtime/altair/src/lib.rs
```diff
@@ -884,20 +952,24 @@ parameter_types! {
 }
 
 impl pallet_treasury::Config for Runtime {
-	// either democracy or 66% of council votes
 	type ApproveOrigin = EnsureRootOr<TwoThirdOfCouncil>;
+	type AssetKind = ();
```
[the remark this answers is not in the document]
lemunozm: If we use `PayFromAccount` as `Paymaster` (that not sure if we want that behavior, but is simpler than the one used by polkadot using xcm), then we need to use `AssetKind = ()`

runtime/altair/src/xcm.rs
lemunozm: Wow, removing that would be great!

lemunozm:
Done here: [fb52bc4](https://github.com/centrifuge/centrifuge-chain/pull/1833/commits/fb52bc48a6498c1ae44a9b8d7fe7111bb161ec57)

Still the `RestrictedTransferLocation::XCM` migration that I'll do in another PR and some open point

libs/types/src/locations.rs
lemunozm: https://github.com/centrifuge/centrifuge-chain/pull/1853 and https://github.com/centrifuge/centrifuge-chain/pull/1858

lemunozm:
There is a last test failing. It seems like the HostConfiguration is not applied and I do not know why: https://github.com/centrifuge/centrifuge-chain/actions/runs/9386407945/job/25846936686?pr=1833

As a fast-dirty-fix to make this work I did: [f2b2934](https://github.com/centrifuge/centrifuge-chain/pull/1833/commits/f2b29349176144c784ccb5878add699f2d5eeb42)

A better solution should be configuring the xcm part correctly. By now, I left this as debt-tech. We, anyway, need to clear and organize the 8K lines of liquidity pools.

runtime/altair/src/lib.rs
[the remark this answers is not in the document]
wischli: You are right!

lemunozm:
@William could you double check the last non-checked point in the PR description? If considered checked, I think we can merge this!

runtime/development/src/lib.rs
```diff
@@ -2073,15 +2124,16 @@ construct_runtime!(
 		XcmpQueue: cumulus_pallet_xcmp_queue::{Pallet, Call, Storage, Event<T>} = 120,
 		PolkadotXcm: pallet_xcm::{Pallet, Call, Storage, Config<T>, Event<T>, Origin} = 121,
 		CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 122,
-		DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 123,
+		// Removed: DmpQueue = 123
```
mustermeiszer, line 2127: We will need to migrate any existing messages from this queue to the new once, right?

runtime/development/src/lib.rs
```diff
@@ -2073,15 +2124,16 @@ construct_runtime!(
...
 		XTokens: pallet_restricted_xtokens::{Pallet, Call} = 124,
 		XcmTransactor: pallet_xcm_transactor::{Pallet, Call, Storage, Event<T>} = 125,
 		OrmlXTokens: orml_xtokens::{Pallet, Event<T>} = 126,
+		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 127,
```
mustermeiszer, line 2131: Will this be the queue for ALL xcm messages or just DMP?

runtime/centrifuge/src/lib.rs
```diff
@@ -2089,15 +2142,16 @@ construct_runtime!(
 		XcmpQueue: cumulus_pallet_xcmp_queue::{Pallet, Call, Storage, Event<T>} = 120,
 		PolkadotXcm: pallet_xcm::{Pallet, Call, Storage, Config<T>, Event<T>, Origin} = 121,
 		CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 122,
-		DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 123,
+		// Removed: DmpQueue = 123
```
mustermeiszer, line 2145: See Dev notes!

runtime/centrifuge/src/lib.rs
```diff
@@ -1025,26 +1072,30 @@ parameter_types! {
...
 	type ProposalBondMaximum = ProposalBondMaximum;
 	type ProposalBondMinimum = ProposalBondMinimum;
-	// either democracy or more than 50% council votes
 	type RejectOrigin = EnsureRootOr<HalfOfCouncil>;
 	type RuntimeEvent = RuntimeEvent;
 	type SpendFunds = ();
 	type SpendOrigin = frame_support::traits::NeverEnsureOrigin<Balance>;
 	type SpendPeriod = SpendPeriod;
-	type WeightInfo = weights::pallet_treasury::WeightInfo<Runtime>;
+	type WeightInfo = (); // Using default weights for recomended hardware
```
mustermeiszer, line 1098: See Dev notes!

runtime/centrifuge/src/lib.rs
```diff
@@ -1010,6 +1056,7 @@ parameter_types! {
 
 	// periods between treasury spends
 	pub const SpendPeriod: BlockNumber = 14 * DAYS;
+	pub const PayoutPeriod: BlockNumber = 30 * DAYS;
```
mustermeiszer, line 1059: Config value taken from?

runtime/altair/src/lib.rs
mustermeiszer, line 1: See dev and centrifuge notes

runtime/common/src/migrations/hold_reason.rs
```diff
@@ -0,0 +1,151 @@
...
+					id: HoldReason::TransferAllowance.into(),
+					// Non-emptiness ensured above
+					amount: holds[0].amount,
+				};
+				weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
+
+				Some(NewHolds::truncate_from(vec![id_amount]))
+			} else {
+				None
+			}
```
mustermeiszer, line 81: I hope that one is not too heavy. How many entries do we have again?

runtime/centrifuge/src/xcm.rs
```diff
@@ -105,9 +74,10 @@ impl staging_xcm_executor::Config for XcmConfig {
 	type PalletInstancesInfo = crate::AllPalletsWithSystem;
 	type ResponseHandler = PolkadotXcm;
 	type RuntimeCall = RuntimeCall;
-	type SafeCallFilter = SafeCallFilter;
+	type SafeCallFilter = Everything;
```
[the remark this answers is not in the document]
wischli: Following Polkadot here: https://github.com/paritytech/polkadot-sdk/pull/4199

runtime/altair/src/migrations.rs
```diff
@@ -23,6 +23,7 @@ pub type UpgradeAltair1035 = (
 	runtime_common::migrations::increase_storage_version::Migration<OrderBook, 0, 1>,
 	runtime_common::migrations::increase_storage_version::Migration<ForeignInvestments, 0, 1>,
 	pallet_collator_selection::migration::v1::MigrateToV1<crate::Runtime>,
+	pallet_collator_selection::migration::v2::MigrationToV2<crate::Runtime>,
```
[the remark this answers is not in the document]
wischli: Both needed. The first one is sorting just our invulnerables and pushing storage version from 0 to 1, the second one translates the storage entry and bumps 1 to 2.

pallets/block-rewards/src/migrations.rs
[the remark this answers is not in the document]
wischli: See above

runtime/centrifuge/src/xcm.rs
```diff
@@ -105,9 +74,10 @@ impl staging_xcm_executor::Config for XcmConfig {
...
 	type SubscriptionService = PolkadotXcm;
 	type Trader = Trader;
+	type TransactionalProcessor = FrameTransactionalProcessor;
```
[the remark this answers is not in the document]
wischli: It just wraps the XCM call into a transactional layer and rolls back in case of errors: https://github.com/paritytech/polkadot-sdk/blob/851c824dc7a8e7e6db2d0c9fb29d232f2f45198a/polkadot/xcm/xcm-builder/src/transactional.rs#L23

runtime/development/src/lib.rs
```diff
@@ -283,27 +294,70 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
+	type ServiceWeight = MessageQueueServiceWeight;
+	type Size = u32;
+	type WeightInfo = ();
+}
+
+/// XCMP Queue is responsible to handle XCM messages coming directly from
+/// sibling parachains.
+impl cumulus_pallet_xcmp_queue::Config for Runtime {
+	type ChannelInfo = ParachainSystem;
+	type ControllerOrigin = EnsureRoot<AccountId>;
```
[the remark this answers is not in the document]
wischli: Resume and suspend XCMP Queue: https://github.com/paritytech/polkadot-sdk/blob/851c824dc7a8e7e6db2d0c9fb29d232f2f45198a/cumulus/pallets/xcmp-queue/src/lib.rs#L136

runtime/development/src/lib.rs
wischli: All messages, XCMP, DMP, UMP: 

> This is intended to be used with an XCM adapter struct implementing the ProcessMessage trait, to be made at a later time in the Polkadot codebase as a single, general queue for XCMP, DMP, UMP and local dispatch/queuing.

https://github.com/paritytech/substrate/pull/12485

runtime/centrifuge/src/lib.rs
wischli: Its a new parameter based on Polkadot et al. config: https://github.com/centrifuge/centrifuge-chain/pull/1833#discussion_r1622305942

runtime/common/src/migrations/hold_reason.rs
wischli: Just two on production luckily. They can only emerge from transfer allowlist for native currency. For more info, see here https://github.com/centrifuge/centrifuge-chain/pull/1849#discussion_r1622051322

runtime/development/src/lib.rs
wischli: Great catch! The DMP Queue has a lazy migration via `on_idle`. So we need to keep it in the runtime @lemunozm 

> The undeployment of the pallet commences through the on_idle hook step-by step and moves all the remaining messages into the DmpSink.
> 
> ```
> impl cumulus_pallet_dmp_queue::Config for Runtime {
>	type DmpSink = MessageQueue;
>	...
> ```
> The pallet can be removed after the lazy migration is done and event Completed is emitted.

Source: https://github.com/paritytech/polkadot-sdk/pull/1246

runtime/development/src/lib.rs
```diff
@@ -939,28 +1005,32 @@ parameter_types! {
...
 	type ProposalBondMaximum = ProposalBondMaximum;
 	type ProposalBondMinimum = ProposalBondMinimum;
-	// either democracy or more than 50% council votes
 	type RejectOrigin = EnsureRootOr<HalfOfCouncil>;
 	type RuntimeEvent = RuntimeEvent;
 	type SpendFunds = ();
 	type SpendOrigin = frame_support::traits::NeverEnsureOrigin<Balance>;
 	type SpendPeriod = SpendPeriod;
-	type WeightInfo = weights::pallet_treasury::WeightInfo<Runtime>;
+	type WeightInfo = (); // Using default weights for recomended hardware
```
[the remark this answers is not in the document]
lemunozm: But hardcoding them can also hit us if `pallet_treasury` updates, and suddenly, one extrinsic is super heavy. We silently still use our hardcoded outdated weights.

I think if `pallet-treasury` is updated, we want the latest recommended weights too.

runtime/common/src/lib.rs
```diff
@@ -301,76 +298,55 @@ pub mod investment_portfolios {
...
-				let pool_currency = PoolInspector::currency_for(pool_id)
-					.expect("Pool must exist; qed")
-					.into();
-				let free_balance = <pallet_restricted_tokens::Pallet<T> as fungibles::Inspect<
-					T::AccountId,
-				>>::reducible_balance(
-					currency.into(),
+			if let CurrencyId::Tranche(pool_id, tranche_id) = currency {
+				let pool_currency = pallet_pool_system::Pallet::<T>::currency_for(pool_id)
+					.expect("Pool must exist; qed");
```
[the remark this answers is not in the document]
lemunozm: Even if it can no be broken right now, we should use `DispatchError::Other` in case future changes can break this. Changing it...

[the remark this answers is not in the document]
wischli: Thanks for raising this question. I wasn't sure either so I tested this locally both with a local node and using Chopsticks for a temp testing runtime API which always panics. 

Luckily, it only creates a panic log but doesn't crash the client or of course not block production:

```
2024-06-07 11:20:11.778  INFO tokio-runtime-worker substrate: [🌀Parachain] 💤 Idle (0 peers), best: #25 (0x0992…4ffb), finalized #22 (0x2fcf…8920), ⬇ 99 B/s ⬆ 64 B/s
2024-06-07 11:20:12.050  INFO tokio-runtime-worker substrate: [Relaychain] ✨ Imported #70 (0x6065…02e4)

====================

Version: 0.10.39-22d0a3898a9

   0: backtrace::capture::Backtrace::create
   1: backtrace::capture::Backtrace::new
   2: sp_panic_handler::set::{{closure}}
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic_handler::{{closure}}
   5: std::sys_common::backtrace::__rust_end_short_backtrace
   6: _rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::option::expect_failed
   9: runtime_common::investment_portfolios::get_account_portfolio
  10: development_runtime::api::dispatch
  11: environmental::using
  12: sc_executor::executor::WasmExecutor<H>::with_instance::{{closure}}
  13: sc_executor::wasm_runtime::RuntimeCache::with_instance
  14: <sc_executor::executor::NativeElseWasmExecutor<D> as sp_core::traits::CodeExecutor>::call
  15: sp_state_machine::execution::StateMachine<B,H,Exec>::execute
  16: <sc_service::client::call_executor::LocalCallExecutor<Block,B,E> as sc_client_api::call_executor::CallExecutor<Block>>::call
  17: <sc_rpc::state::state_full::FullState<BE,Block,Client> as sc_rpc::state::StateBackend<Block,Client>>::call
  18: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
  19: tokio::runtime::task::core::Core<T,S>::poll
  20: tokio::runtime::task::harness::Harness<T,S>::poll
  21: tokio::runtime::blocking::pool::Inner::run
  22: std::sys_common::backtrace::__rust_begin_short_backtrace
  23: core::ops::function::FnOnce::call_once{{vtable.shim}}
  24: std::sys::pal::unix::thread::Thread::new::thread_start
  25: __pthread_joiner_wake


Thread 'tokio-runtime-worker' panicked at 'Pool must exist; qed', /Users/william/dev/rust/centrifuge-chain/runtime/common/src/lib.rs:314

This is a bug. Please report it at:

	https://github.com/paritytech/cumulus/issues/new

2024-06-07 11:20:12.151  WARN tokio-runtime-worker wasm-runtime: Evicting failed runtime instance error=Runtime panicked: Pool must exist; qed
2024-06-07 11:20:16.784  INFO tokio-runtime-worker substrate: [🌀Parachain] 💤 Idle (0 peers), best: #25 (0x0992…4ffb), finalized #23 (0x1500…5b5c), ⬇ 45 B/s ⬆ 45 B/s
2024-06-07 11:20:16.784  INFO tokio-runtime-worker substrate: [Relaychain] 💤 Idle (2 peers), best: #70 (0x6065…02e4), finalized #68 (0x311e…afe9), ⬇ 1.6kiB/s ⬆ 0.6kiB/s
2024-06-07 11:20:18.038  INFO tokio-runtime-worker babe: [Relaychain] 👶 New epoch 7 launching at block 0x3400…13fe (block slot 286292003 >= start slot 286292003).
2024-06-07 11:20:18.038  INFO tokio-runtime-worker babe: [Relaychain] 👶 Next epoch starts at slot 286292013
2024-06-07 11:20:18.043  INFO tokio-runtime-worker substrate: [Relaychain] ✨ Imported #71 (0x3400…13fe)
2024-06-07 11:20:18.045  INFO tokio-runtime-worker sc_basic_authorship::basic_authorship: [🌀Parachain] 🙌 Starting consensus session on top of parent 0x09920d2f8c5ce3ade2233a3252933f00beb5daac18c5c62a39320d9ca34e4ffb
2024-06-07 11:20:18.048  INFO tokio-runtime-worker sc_basic_authorship::basic_authorship: [🌀Parachain] 🎁 Prepared block for proposing at 26 (2 ms) [hash: 0x7efdb840f9b8be28758a156ce80ef05a0d8ce2bcef9a549de981143d8b864688; parent_hash: 0x0992…4ffb; extrinsics (2): [0x412f…fdf4, 0xd77a…184e]
2024-06-07 11:20:18.051  INFO tokio-runtime-worker aura::cumulus: [🌀Parachain] 🔖 Pre-sealed block for proposal at 26. Hash now 0xbd16575765e5d2053cb8f0164cc61921fad9778aa9a9f3055f2dc3c3d2a4c599, previously 0x7efdb840f9b8be28758a156ce80ef05a0d8ce2bcef9a549de981143d8b864688.
2024-06-07 11:20:18.051  INFO tokio-runtime-worker substrate: [🌀Parachain] ✨ Imported #26 (0xbd16…c599)
```

Anyways, if this code path here panics, we are in trouble anyways because it means we have a tranche currency in `orml_tokens` storage for a pool id which does not exist such that `currency_for(pool_id)` panics. We have many similar code paths outside of runtime API with the exact same assumptions of this being impossible.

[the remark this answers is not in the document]
lemunozm: Fixed: [5baab79](https://github.com/centrifuge/centrifuge-chain/pull/1833/commits/5baab79af45dcce828672a12277e61a72dca0e60)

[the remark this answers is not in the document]
wischli: Anyways, if this code path here panics, we are in trouble anyways because it means we have a tranche currency in `orml_tokens` storage for a pool id which does not exist such that `currency_for(pool_id)` panics. We have many similar code paths outside of runtime API with the exact same assumptions of this being impossible.

wischli:
> @william could you double check the last non-checked point in the PR description? If considered checked, I think we can merge this!

You linked a stranger :) Sorry for the confusing Github name. It's a reference to my birth name. I am ensuring a Docker image can be build and then all checks are done!

Only thing left at this point is re-adding the DMP Queue for the lazy migration.

lemunozm:
> Only thing left at this point is re-adding the DMP Queue for the lazy migration.

Doing it now

runtime/common/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Yeah, the problem with the non-compiled assumptions is that in the future they changed and silently... 💥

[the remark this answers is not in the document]
lemunozm: Good test the above! I suppose substrate crate a tokio task to each requires. So if one panics nothing happens. TBH, that leaves me calm, maybe we can use more `expect()` for impossible things in some parts of our code. Do you want to revert the above fix?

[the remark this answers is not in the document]
lemunozm: Although probably each panic would adds more memory leaks to the node 🤔

runtime/development/src/lib.rs
lemunozm: Done [b5a7d89](https://github.com/centrifuge/centrifuge-chain/pull/1833/commits/b5a7d895819d89bcd8d66d203fb648e388378cc3)

runtime/development/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I've used better weights anyway => [b5a7d89](https://github.com/centrifuge/centrifuge-chain/pull/1833/commits/b5a7d895819d89bcd8d66d203fb648e388378cc3)

runtime/common/src/lib.rs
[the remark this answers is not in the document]
mustermeiszer: I would error.

mustermeiszer (approved):

wischli (approved):
Awesome work everyone! So happy we can merge this bad boy 🤘

hieronx (approved):

cdamian (approved):

mustermeiszer (approved):

PR #1834, opened 2024-05-10
Remove crowdloan pallets

lemunozm:
# Description

Remove:

- [x] `pallet-croddloan-claim`
- [x] `pallet-crowdloan-reward`
- [x] `libs/proofs` (used only by them)
- [x] Unused trait
- [ ] Storage from `altair` and `centrifuge` (Tech-debt)


wischli:
> I think we also need to create a migration to remove any storage from them

Yes, we can use the `runtime_common::migrations::KillPallet` for that!

lemunozm:
We have an issue removing the storage: 

```
[2024-05-10T12:11:20Z INFO  runtime_common::migrations::nuke] Clear pallet "CrowdloanClaim": iteration result. backend: 10349 unique: 10349 loops: 10349
```

It's 259% heavier than a block, so we need to split this, which is not trivial right I guess. Maybe we can leave the migration for another PR and start cleaning the repo now.

lemunozm:
@gpmayorga the PR still appears as failed when it shouldn't 🤔. Something missing?

mustermeiszer:
> It's 259% heavier than a block, so we need to split this, which is not trivial I guess. Maybe we can leave the migration for another PR and start cleaning the repo now for the upgrade

Multiblock migrations are maybe already available: https://github.com/paritytech/polkadot-sdk/pull/1554#discussion_r1420277877

wischli:
> > It's 259% heavier than a block, so we need to split this, which is not trivial I guess. Maybe we can leave the migration for another PR and start cleaning the repo now for the upgrade
> 
> Multiblock migrations are maybe already available: [paritytech/polkadot-sdk#1554](https://github.com/paritytech/polkadot-sdk/pull/1554)

Unfortunately, multiblock migrations are part of SDK v1.9.0. 

IMO, it's absolutely feasible to keep the dead storage until then because it doesn't impose any risk. It just needs to be cleaned at some point.

lemunozm:
Then, we can merge this!

wischli (approved):
I created a follow-up task: https://github.com/centrifuge/centrifuge-chain/issues/1836

lemunozm:
I already did it, but I close mine and use yours 👍🏻

PR #1837, opened 2024-05-16
Fix: transfer debt event with updated repaid amount

lemunozm:
# Description

`TransferEvent` was set with the initial `repaid_amount`. That amount changes once it is processed and could not be the amount that it's really applied.

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -838,7 +838,7 @@ pub mod pallet {
 				Err(Error::<T>::UnrelatedChangeId)?
 			};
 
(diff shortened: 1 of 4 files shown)


hieronx (commented):
Fix looks good! Would be nice to have test :)

wischli (dismissed):
Could we add a unit test to assert event payloads? IMO, every hotfix should either include a test fix or add a new test.

mustermeiszer (approved):
Great catch!

lemunozm:
Sure! I always forget to test events properly, and I should.

mustermeiszer (approved):

PR #1840, opened 2024-05-20
Remove old migrations for development 1047

lemunozm:
# Description

Double check it's no longer needed please.


runtime/development/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "development-runtime"
-version = "0.10.46"
```
lemunozm, line 3: We had this not updated, although I guess it's not a problem

lemunozm:
Can we merge this soon in order to avoid some issues in `polkadot-1.7.2` upgrade with these migrations?

runtime/development/Cargo.toml
wischli: Unfortunately, this often falls short but it doesn't have any effect anyways.

wischli (commented):
We must not merge this before Dev and Demo are upgraded to 1047. Both are currently on 1046 and need all of these migrations. I wanted to hold on to this until #1812 is merged.

lemunozm:
Ok 👍🏻 Finally, it's not urgent for me for the upgrade by now, so we can leave this here

PR #1841, opened 2024-05-21
Feat: Add new oracle key

mustermeiszer:
# Description
The DYF will have shares of other funds as its assets. Those fund holdings are valued by an oracle, but have not a valid registered ISIN we could use. Hence, this fallback key will allow to specifically value a given object in the `pallet-loans`.

## Changes and Descriptions
* Adds a new variant in `enum OracleKey`
* Fixes all variants by using `#[codec(index = ...)`

# Checklist:


lemunozm (approved):
LGTM!

PR #1842, opened 2024-05-22
Fix: linear accrual not overpassing maturity

lemunozm:
# Description

[slack thread](https://kflabs.slack.com/archives/C04GJQAM9P0/p1715955299145219?thread_ts=1715935675.464829&cid=C04GJQAM9P0)

We need to avoid computing points in the line of linear accrual if the time overpass the maturity time, instead, we want to clamp the value to `notional` (which is the same when the time is just maturity)

### DIFF
--- a/pallets/loans/src/entities/loans.rs
+++ b/pallets/loans/src/entities/loans.rs
@@ -14,7 +14,7 @@ use sp_runtime::{
 	},
 	DispatchError,
 };
-use sp_std::{cmp::min, collections::btree_map::BTreeMap};
+use sp_std::collections::btree_map::BTreeMap;
 
 use crate::{
 	entities::{
@@ -359,8 +359,7 @@ impl<T: Config> ActiveLoan<T> {
 				inner.adjust(Adjustment::Increase(amount.balance()?))?
 			}
 			ActivePricing::External(inner) => {
-				let when = T::Time::now();
-				inner.adjust(Adjustment::Increase(amount.external()?), Zero::zero(), when)?
+				inner.adjust(Adjustment::Increase(amount.external()?), Zero::zero())?
 			}
 		}
 
@@ -434,8 +433,7 @@ impl<T: Config> ActiveLoan<T> {
 			}
 			ActivePricing::External(inner) => {
 				let principal = amount.principal.external()?;
-				let when = min(T::Time::now(), self.schedule.maturity.date());
-				inner.adjust(Adjustment::Decrease(principal), amount.interest, when)?;
+				inner.adjust(Adjustment::Decrease(principal), amount.interest)?;
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


pallets/loans/src/entities/loans.rs
```diff
@@ -359,7 +359,8 @@ impl<T: Config> ActiveLoan<T> {
 				inner.adjust(Adjustment::Increase(amount.balance()?))?
 			}
 			ActivePricing::External(inner) => {
-				inner.adjust(Adjustment::Increase(amount.external()?), Zero::zero())?
+				let when = T::Time::now();
```
mustermeiszer, line 362: I guess maturity not being passed already is checked by the `ensure_can_borrw()`?

pallets/loans/src/tests/repay_loan.rs
```diff
@@ -812,3 +812,41 @@ fn with_unregister_price_id_and_oracle_not_required() {
...
+
+		let active_loan = util::get_loan(loan_id);
+
+		let settlement_price_updated = match active_loan.pricing() {
+			ActivePricing::External(inner) => inner.settlement_price_updated(),
+			_ => unreachable!(),
+		};
+
+		// We must never overpass madurity date
+		assert_eq!(active_loan.maturity_date(), settlement_price_updated);
```
mustermeiszer, line 850: Could we also check, that the linear pricing actually clamps with giving the notional baack?

pallets/loans/src/entities/loans.rs
lemunozm: But not in `repay()`

lemunozm:
BTW, I did the fix quite bad 😆. I'm redoing the fix implementation

pallets/loans/src/tests/repay_loan.rs
lemunozm: Done

pallets/loans/src/tests/repay_loan.rs
```diff
@@ -812,3 +812,39 @@ fn with_unregister_price_id_and_oracle_not_required() {
...
+
+		assert_eq!(
+			ActiveLoanInfo::try_from((POOL_A, util::get_loan(loan_id)))
+				.unwrap()
+				.current_price
+				.unwrap(),
+			NOTIONAL
+		);
+	});
+}
```
mustermeiszer, line 850: I would still like to test the case where 

`last_updated_price` and `T::Time::now()` are both equal to `maturity`. I think that is an edge case worth testing

lemunozm:
Your migration should work without any change after rebasing @wischli 🤞🏻

pallets/loans/src/tests/repay_loan.rs
lemunozm: Well, it is repaid with one DAY of overdue, and the expected value is `notional`

PR #1856, opened 2024-06-04
Fix: `pallet-pool-system` - Ratio Calculation

mustermeiszer:
# Description
The current ratio calculation has the following bug. The logic assumes that `Sum(tranche.supply) = total_assets`, but as the residual tranche does only receive its variable interest after the ratio calculation, this is not true. 

In order to fix this, the following change does give the automatically all the remaining ratio to the residual tranche.
## Changes and Descriptions
* Adds a test to ensure ratios are actually adapting accordingly
* Changes ratio calculation during `fn do_execute_epoch`

# Checklist:

### DIFF
--- a/pallets/pool-system/src/lib.rs
+++ b/pallets/pool-system/src/lib.rs
@@ -187,6 +187,7 @@ pub mod pallet {
 		traits::{tokens::Preservation, Contains, EnsureOriginWithArg},
 		PalletId,
 	};
+	use rev_slice::SliceExt;
 	use sp_runtime::{traits::BadOrigin, ArithmeticError};
 
 	use super::*;
@@ -1245,7 +1246,11 @@ pub mod pallet {
 				.map(|tranches| {
 					tranches
 						.iter()
-						.zip(&executed_amounts)
+						// NOTE: Reversing amounts, as residual amount is on top.
+						// NOTE: Iterator of executed amounts is one time larger than the
+						//       non_residual_tranche-iterator, but we anyways push all reamaining
+						//       ratio to the residual tranche.
+						.zip(executed_amounts.rev())
 						.map(|(tranche, &(invest, redeem))| {
 							Ok(Perquintill::from_rational(
 								tranche.supply.ensure_add(invest)?.ensure_sub(redeem)?,
@@ -1260,7 +1265,12 @@ pub mod pallet {
 				.iter()
 				.try_fold(Perquintill::zero(), |acc, ratio| acc.ensure_add(*ratio))?;
 
+			// Pushing all remaining ratio to the residual tranche
 			tranche_ratios.push(Perquintill::one().ensure_sub(non_residual_tranche_ratio_sum)?);
+
+			// NOTE: We need to reverse the ratios here, as the residual tranche is on top
+			// all the time
+			tranche_ratios.reverse();
 
 			pool.tranches.rebalance_tranches(
 				T::Time::now(),
--- a/pallets/pool-system/src/lib.rs
+++ b/pallets/pool-system/src/lib.rs
@@ -1240,37 +1240,39 @@ pub mod pallet {
 
 			let executed_amounts = epoch.tranches.fulfillment_cash_flows(solution)?;
 			let total_assets = epoch.nav.total(pool.reserve.total)?;
-			let mut tranche_ratios = epoch
-				.tranches
-				.non_residual_tranches()
-				.map(|tranches| {
-					tranches
-						.iter()
-						// NOTE: Reversing amounts, as residual amount is on top.
-						// NOTE: Iterator of executed amounts is one time larger than the
-						//       non_residual_tranche-iterator, but we anyways push all reamaining
-						//       ratio to the residual tranche.
-						.zip(executed_amounts.rev())
-						.map(|(tranche, &(invest, redeem))| {
-							Ok(Perquintill::from_rational(
-								tranche.supply.ensure_add(invest)?.ensure_sub(redeem)?,
-								total_assets,
-							))
-						})
-						.collect::<Result<Vec<_>, ArithmeticError>>()
-				})
-				.unwrap_or(Ok(Vec::new()))?;
 
-			let non_residual_tranche_ratio_sum = tranche_ratios
-				.iter()
-				.try_fold(Perquintill::zero(), |acc, ratio| acc.ensure_add(*ratio))?;
+			let tranche_ratios = {
+				let mut sum_non_residual_tranche_ratios = Perquintill::zero();
+				let num_tranches = pool.tranches.num_tranches();
+				let mut current_tranche = 1;
+				let mut ratios = epoch
+					.tranches
+					// NOTE: Reversing amounts, as residual amount is on top.
+					.combine_with_non_residual_top(
+						executed_amounts.rev(),
+						|tranche, &(invest, redeem)| {
+							let ratio = if current_tranche < num_tranches {
+								Perquintill::from_rational(
+									tranche.supply.ensure_add(invest)?.ensure_sub(redeem)?,
+									total_assets,
+								)
+							} else {
+								Perquintill::one().ensure_sub(sum_non_residual_tranche_ratios)?
+							};
 
-			// Pushing all remaining ratio to the residual tranche
-			tranche_ratios.push(Perquintill::one().ensure_sub(non_residual_tranche_ratio_sum)?);
+							sum_non_residual_tranche_ratios.ensure_add_assign(ratio)?;
+							current_tranche.ensure_add_assign(1)?;
 
-			// NOTE: We need to reverse the ratios here, as the residual tranche is on top
-			// all the time
-			tranche_ratios.reverse();
+							Ok(ratio)
+						},
+					)?;
+
+				// NOTE: We need to reverse the ratios here, as the residual tranche is on top
+				//       all the time
+				ratios.reverse();
+
+				ratios
+			};
 
 			pool.tranches.rebalance_tranches(
 				T::Time::now(),
--- a/pallets/pool-system/src/lib.rs
+++ b/pallets/pool-system/src/lib.rs
@@ -1255,16 +1255,13 @@ pub mod pallet {
 							//       implementation defaults to 100% if the denominator is zero
 							let ratio = if total_assets.is_zero() {
 								Perquintill::zero()
+							} else if current_tranche < num_tranches {
+								Perquintill::from_rational(
+									tranche.supply.ensure_add(invest)?.ensure_sub(redeem)?,
+									total_assets,
+								)
 							} else {
-								if current_tranche < num_tranches {
-									Perquintill::from_rational(
-										tranche.supply.ensure_add(invest)?.ensure_sub(redeem)?,
-										total_assets,
-									)
-								} else {
-									Perquintill::one()
-										.ensure_sub(sum_non_residual_tranche_ratios)?
-								}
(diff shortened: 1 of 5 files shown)


pallets/pool-system/src/tests/ratios.rs
```diff
@@ -0,0 +1,163 @@
...
+// This file is part of the Centrifuge chain project.
+// Centrifuge is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version (see http://www.gnu.org/licenses).
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+use super::*;
```
wischli, line 14: Nit: Formatting

```suggestion
// Copyright 2024 Centrifuge Foundation (centrifuge.io).
//
// This file is part of the Centrifuge chain project.
// Centrifuge is free software: you can redistribute it and/or modify
// it under the terms of the GNU General Public License as published by
// the Free Software Foundation, either version 3 of the License, or
// (at your option) any later version (see http://www.gnu.org/licenses).
// Centrifuge is distributed in the hope that it will be useful,
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
// GNU General Public License for more details.

use sp_arithmetic::traits::EnsureSub;

use super::*;
```

pallets/pool-system/src/tests/ratios.rs
```diff
@@ -0,0 +1,163 @@
...
+		crate::Pool::<Runtime>::get(0)
+			.unwrap()
+			.tranches
+			.residual_top_slice()
+			.iter()
+			.for_each(|tranche| {
+				assert_eq!(tranche.ratio, Perquintill::from_percent(50));
+			});
+
+		// Attempt to redeem half
```
wischli, line 95: Comment is incorrect right? We are redeeming 40% which is 200 of the 500 tranche tokens.

pallets/pool-system/src/tests/ratios.rs
```diff
@@ -0,0 +1,163 @@
...
+	new_test_ext().execute_with(|| {
+		let pool_owner = DEFAULT_POOL_OWNER;
+		let pool_owner_origin = RuntimeOrigin::signed(pool_owner);
+
+		// Initialize pool with initial investments
+		let senior_interest_rate = Rate::saturating_from_rational(10, 100)
+			/ Rate::saturating_from_integer(SECONDS_PER_YEAR)
+			+ One::one();
+
+		assert_ok!(PoolSystem::create(
```
wischli, line 27: Q: Would the default pool from `crate::util::default_pool::create()` suffice or is any of the diverging configurations necessary?
* 1.1% annual senior rate vs 0%
* 10% min risk buffer  vs 0%
* 10k AUSD reserve vs 0

pallets/pool-system/src/lib.rs
```diff
@@ -1239,15 +1240,37 @@ pub mod pallet {
...
+						//       ratio to the residual tranche.
+						.zip(executed_amounts.rev())
+						.map(|(tranche, &(invest, redeem))| {
+							Ok(Perquintill::from_rational(
+								tranche.supply.ensure_add(invest)?.ensure_sub(redeem)?,
+								total_assets,
+							))
+						})
+						.collect::<Result<Vec<_>, ArithmeticError>>()
+				})
```
wischli, line 1261: Can"t we simplify this by using existing `Tranches` API or does this not work because all API includes the residual tranche?

wischli (commented):
AFAICT, looks good! Just wondering whether the single test is sufficient.

pallets/pool-system/src/tests/ratios.rs
mustermeiszer: Yes, adapted.

pallets/pool-system/src/tests/ratios.rs
mustermeiszer: well, `max_reserve` is a problem. The rest would be fine.

pallets/pool-system/src/lib.rs
mustermeiszer: it is already using tranches, but not the combine as I, as you said, need to exclude the residual tranche.

mustermeiszer: Happy to add a cleaner solution if you have one at hand.

pallets/pool-system/src/tests/ratios.rs
mustermeiszer: I adapted the default pool and used that one.

mustermeiszer:
@wischli what level of testing for the ratios would be fine for you? 3 tranche pool and 1 tranche pool additionally, or more movement in the ratios?

mustermeiszer:
@wischli tried to clean it up a bit [here](https://github.com/centrifuge/centrifuge-chain/pull/1856/commits/7886f40345db4e438b72ab5182b2e7fa2df1394f). What do you think - better or worse? ^^

pallets/pool-system/src/lib.rs
mustermeiszer: See https://github.com/centrifuge/centrifuge-chain/pull/1856/commits/7886f40345db4e438b72ab5182b2e7fa2df1394f

wischli:
> @wischli tried to clean it up a bit [here](https://github.com/centrifuge/centrifuge-chain/pull/1856/commits/7886f40345db4e438b72ab5182b2e7fa2df1394f). What do you think - better or worse? ^^

Thanks a lot for providing an alternate solution! IMO the new version is easier to follow and thus better readable but the first version cleaner codewise. So I vote for keeping the current version 2. Which one do you prefer?

mustermeiszer:
Would keep the first version

pallets/pool-system/src/lib.rs
```diff
@@ -1239,15 +1240,39 @@ pub mod pallet {
...
+					.combine_with_non_residual_top(
+						executed_amounts.rev(),
+						|tranche, &(invest, redeem)| {
+							let ratio = if current_tranche < num_tranches {
+								Perquintill::from_rational(
+									tranche.supply.ensure_add(invest)?.ensure_sub(redeem)?,
+									total_assets,
+								)
+							} else {
+								Perquintill::one().ensure_sub(sum_non_residual_tranche_ratios)?
```
lemunozm, line 1260: The sum of tranche ratios can never be higher than one, right?

lemunozm: I think this line is annulated by line 1263. No matter which value `sum_non_residual_tranche_ratios` has, at the end of this iteration `sum_non_residual_tranche_ratios` will always have the value `one()`. Is this the expected behavior?

pallets/pool-system/src/lib.rs
```diff
@@ -1239,15 +1240,39 @@ pub mod pallet {
...
+				let mut sum_non_residual_tranche_ratios = Perquintill::zero();
+				let num_tranches = pool.tranches.num_tranches();
+				let mut current_tranche = 1;
+				let mut ratios = epoch
+					.tranches
+					// NOTE: Reversing amounts, as residual amount is on top.
+					.combine_with_non_residual_top(
+						executed_amounts.rev(),
+						|tranche, &(invest, redeem)| {
+							let ratio = if current_tranche < num_tranches {
```
lemunozm, line 1254: When this condition will be `false`? Only for the last element?

lemunozm (commented):
I left some questions to know if I understand this correctly, but apart from them, LGTM!

pallets/pool-system/src/tests/ratios.rs
```diff
@@ -0,0 +1,130 @@
...
+		assert_ok!(PoolSystem::close_epoch(pool_owner_origin.clone(), 0));
+		assert_ok!(Investments::collect_redemptions(
+			RuntimeOrigin::signed(0),
+			TrancheCurrency::generate(0, SeniorTrancheId::get()),
+		));
+
+		let new_residual_ratio = Perquintill::one();
+		let mut next_ratio = new_residual_ratio;
+
+		// Ensure ratios are 100/0
```
lemunozm, line 103:
```suggestion
		// Ensure ratios are 100/100
```
maybe?

pallets/pool-system/src/lib.rs
mustermeiszer: Yes

pallets/pool-system/src/tests/ratios.rs
```diff
@@ -0,0 +1,404 @@
...
+			],
+		);
+
+		let check_ratios = [
+			Perquintill::from_rational_with_rounding(1u64, 3, Rounding::Up).unwrap(),
+			Perquintill::from_rational(1u64, 3),
+			Perquintill::from_rational(1u64, 3),
+		];
+
+		// Ensure ratios are 100
```
wischli, line 274: IIUC

```suggestion
		// Ensure ratios are 1/3 each
```

pallets/pool-system/src/tests/ratios.rs
```diff
@@ -0,0 +1,404 @@
...
+			0,
+			vec![
+				(0, JuniorTrancheId::get(), 500 * CURRENCY),
+				(0, SeniorTrancheId::get(), 500 * CURRENCY),
+				(0, SecondSeniorTrancheId::get(), 500 * CURRENCY),
+			],
+		);
+
+		let check_ratios = [
+			Perquintill::from_rational_with_rounding(1u64, 3, Rounding::Up).unwrap(),
```
wischli, line 269: NIT: Any reason why this diverges from the lines below?
```suggestion
			Perquintill::from_rational(1u64, 3),
```

pallets/pool-system/src/tests/ratios.rs
```diff
@@ -0,0 +1,404 @@
...
+		assert_ok!(PoolSystem::close_epoch(pool_owner_origin.clone(), 0));
+		assert_ok!(Investments::collect_redemptions(
+			RuntimeOrigin::signed(0),
+			TrancheCurrency::generate(0, JuniorTrancheId::get()),
+		));
+
+		let check_ratios = [
+			Perquintill::from_rational(1u64, 5),
+			Perquintill::from_rational(2u64, 5),
+		];
```
wischli, line 299: Q: Why does this drop to two ratios? Shouldn't it be three at `20%, 40%, 40%`?

wischli (commented):
AFAICT, the 3 tranche test needs some minor fixes! Thanks for adding these three tests. Definitely makes everything much clearer for me.

pallets/pool-system/src/tests/ratios.rs
lemunozm: Ahh ok

pallets/pool-system/src/tests/ratios.rs
mustermeiszer: `Perquintill` usually rounds down, but we need to have the round up version here as the residual takes the rest.

pallets/pool-system/src/tests/ratios.rs
mustermeiszer: Fixed

wischli (approved):
LGTM!

wischli (approved):

PR #1858, opened 2024-06-04
chore: extend v1.7.2 transfer location migration

wischli:
# Description

Extends #1853 by 
* Adding pre- and post-hooks
* Adding a migration for Dev and Demo
* Moving from hardcoded account id migration to account list per runtime
  * Dev has no storage entries which need to be migrated
  * Demo has 2 storage entries for the same account id which need to be migrated but since we don't use XCM messages on Demo, we can just purge them IMO

## To be discussed

This PR is more aggressive with storage removal than the base one: It removes any old storage entry which cannot be decoded versus only removing, if it can be decoded and a new v4 location can be created.

My argument is that undecodable storage keys are always a pain to clean up. If new storage entries pop up in between proposing the RU and executing it, we can either push a hotfix PR afterwards migrating those keys or communicating that these users need to re-do their transfer allowlist call manually. 

Another option would be exposing a `migrate` extrinsic in the `pallet-transfer-allowlist` for  undecodable keys and keeping them in storage.

# Checklist:


lemunozm:
> My argument is that undecodable storage keys are always a pain to clean up. If new storage entries pop up in between proposing the RU and executing it, we can either push a hotfix PR afterwards migrating those keys or communicating that these users need to re-do their transfer allowlist call manually.

But how can we know we need to migrate them or re-do them if we remove all of them? Just because the transfer filter is not filtering as expected?

lemunozm (approved):
Great improvement over the base migration @wischli! All cases are well tied 💯

runtime/centrifuge/src/migrations.rs
```diff
@@ -10,11 +10,26 @@
...
 
 // Number of identities on Centrifuge Chain on 29.05.2024 was 34
 const IDENTITY_MIGRATION_KEY_LIMIT: u64 = 1000;
 
+parameter_types! {
+	// Address used by Anemoy to withdraw in AssetHub
+	// 4dTeMxuPJCK7zQGhFcgCivSJqBs9Wo2SuMSQeYCCuVJ9xrE2 --> 5Fc9NzKzJZwZvgjQBmSKtvZmJ5oP6B49DFC5dXZhTETjrSzo
+	pub AccountMap: Vec<(AccountId, AccountId)> = vec![
+		(
+			AccountId::new(hex_literal::hex!("5dbb2cec05b6bda775f7945827b887b0e7b5245eae8b4ef266c60820c9377185")),
```
lemunozm, line 27: This account is `4dTeMxuPJCK7zQGhFcgCivSJqBs9Wo2SuMSQeYCCuVJ9xrE2` in hex, right?

wischli: Yes
```
❯ subkey inspect 4dTeMxuPJCK7zQGhFcgCivSJqBs9Wo2SuMSQeYCCuVJ9xrE2
Public Key URI `4dTeMxuPJCK7zQGhFcgCivSJqBs9Wo2SuMSQeYCCuVJ9xrE2` is account:
  Network ID/Version: centrifuge
  Public key (hex):   0x5dbb2cec05b6bda775f7945827b887b0e7b5245eae8b4ef266c60820c9377185
  Account ID:         0x5dbb2cec05b6bda775f7945827b887b0e7b5245eae8b4ef266c60820c9377185
  Public key (SS58):  4dTeMxuPJCK7zQGhFcgCivSJqBs9Wo2SuMSQeYCCuVJ9xrE2
```

wischli:
> But how can we know we need to migrate them or re-do them if we remove all of them? Just because the transfer filter is not filtering as expected?

I would say by looking at the logs at the time of the migration which I always do for runtime upgrades and we have access to historic storage by querying it for an older block hash in which it still existed. 

The only upside for keeping it is that it is easier to reconstruct and check against on the runtime level.

lemunozm:
Ok, then we can remove them. I think the probability to have new entries is low

lemunozm:
You can merge it when you want

PR #1859, opened 2024-06-05
integration-tests: Use fudge-test-relay

cdamian:


runtime/integration-tests/src/generic/cases/liquidity_pools.rs
```diff
@@ -7452,7 +7452,7 @@ mod centrifuge {
 					pallet_balances::Pallet::<FudgeRelayRuntime<T>>::free_balance(
 						&Keyring::Alice.into()
 					),
-					109978937205
```
cdamian, line 7455: This difference seems big tho.

lemunozm: It makes sense because the relay has sent 3 and later received 1: 10 - 3 + 1 = 8 - fees

cdamian: Cool, thank you very much for checking, I forgot how these things were supposed to work :D

lemunozm: Same result in rococo by the way

runtime/integration-tests/Cargo.toml
```diff
@@ -22,6 +22,7 @@ serde = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
 tracing-subscriber = { workspace = true }
+log = { workspace = true }
```
mustermeiszer, line 25: Nit: Do we need that?

mustermeiszer (approved):

runtime/integration-tests/Cargo.toml
lemunozm: Not, but it's true that it's confortable to have it there not to add that crate each time you need to add logs to debug something.


BTW, I think we will close this PR and will go with rococo

lemunozm:
Closing in favor of https://github.com/centrifuge/centrifuge-chain/pull/1833/commits/8410d16fb592f2684f8c23edba7a6801fd1236b1

PR #1860, opened 2024-06-05
Feat: Liquidity pools inter domain tranche token transfer

mustermeiszer:
# Description
The PR allows to send tranche tokens from `Domain A` to `Domain B` by relaying the message over `Centrifuge`. The Solidity side of the LP logic already is capable of doing so, this enables this feature on the chain itself.

NOTE: 
Pushing this change before the RU as the DYF pool will most likely need that feature.

## Changes and Descriptions
* Adds a new `DomainAccountToDomainAddress` converter to ease conversion from message to type
* Add different handling logic for tranche token transfers
  * Always transfer from `DomainAccount` to `T::DomainAddressToAccountId::convert(receiver)`
     &rarr; for local receiver this will just be the normal account they choose to send to
     &rarr; for remote receiver this will be a derived account that temporarily holds the tokens
  * If centrifuge is not the final receiving domain &rarr; use normal local transfer logic of LP to forward to actual domain

# Checklist:


mustermeiszer:
Should we start adding test to the pallet itself now? ^^ Integration test will be hard until the LP integration test is not done.

runtime/altair/src/liquidity_pools.rs
```diff
@@ -1,99 +0,0 @@
-// Copyright 2023 Centrifuge Foundation (centrifuge.io).
```
lemunozm, line 1: Why are these pallets removed?

mustermeiszer: unused config, right?

pallets/liquidity-pools/src/inbound.rs
```diff
@@ -78,12 +81,25 @@ where
...
+		// to the right domain from the holdings of the receiver we just transferred
+		// them to.
+		if receiver.domain() != Domain::Centrifuge {
+			Pallet::<T>::transfer_tranche_tokens(
+				T::RuntimeOrigin::signed(local_representation_of_receiver),
+				pool_id,
+				tranche_id,
+				receiver,
+				amount,
+			)?;
```
wischli, line 100: Love the simplicity of this feature!

wischli (commented):
Code LGTM! We could add a simple integration test which ensures the 3 hop of tranche token amounts via events.

runtime/altair/src/liquidity_pools.rs
lemunozm: Ah, that's right! I forgot to remove those files after migrating them 🤦🏻‍♂️. Thanks

lemunozm (dismissed):
LGTM!

wischli (approved):
I am fine with adding tests later and verifying this change on Demo. Code LGTM!

PR #1862, opened 2024-06-06
fix: collator selection v2 storage migration

wischli:
# Description

Fixes [unmentioned breaking storage change](https://github.com/paritytech/polkadot-sdk/pull/1340) in collator selection pallet by manually backporting an improved version of [their fix](https://github.com/paritytech/polkadot-sdk/pull/4229).

Unfortunately, we bump the storage version here higher than the max version of the `pallet-collator-selection@polkadot-v1.7.2` such that try-runtime throws an error:

```
[2024-06-06T09:06:24Z ERROR runtime::frame-support] CollatorSelection: On chain storage version StorageVersion(2) doesn't match current storage version StorageVersion(1).
```

This will lead to unsuccessful try-runtime CI in our codebase until we upgrade to SDK v1.11.0. I still prefer this over applying the migration and staying at v1. WDYT @lemunozm @mustermeiszer?

# Checklist:


lemunozm:
> we bump the storage version here higher than the max version of the pallet-collator-selection@polkadot-v1.7.2 such that try-runtime throws an error

As a workaround, can we later force v1 again with a new migration?

wischli:
> > we bump the storage version here higher than the max version of the pallet-collator-selection@polkadot-v1.7.2 such that try-runtime throws an error
> 
> 
> 
> As a workaround, can we later force v1 again with a new migration?

Then, we could also just not bump the storage version with this migration.

lemunozm:
Does that make sense to you?

wischli:
> Does that make sense to you? Or could it lead to future issues? I guess when the correct 1.11 migration enters, then it will migrate 0 entries and everything will work

Since we drain the `Candidates` storage here, re-applying this migrations shouldn't do any harm. So I am also fine with going the path of applying the migration in an unchecked fashion in which we don't bump the storage version. Will push after lunch!

lemunozm (approved):
LGTM!

lemunozm:
Thanks for your good research about this topic

lemunozm:
Does it mean this PR is no longer needed?

wischli:
> @wischli does it mean this PR is no longer needed?

We still need the migration but the diff is now only 3 lines :)

PR #1864, opened 2024-06-06
Loans: multi cashflows fix external loan

lemunozm:
# Description

External pricing loan with correct cashflow computation.

For computing interest of:
- internal loans: we use `debt - principal`
- external loans: we use `debt - notional`

From the cashflow perspective, we can emulate the `debt` at maturity given the `interest` and `principal` parameters given to the `expected_cashflows()`. Nevertheless, in the case of external loans, the principal used for computing the debt is different from the value substracted to the debt to obtain the interest. We need to pass a new parameter to `expected_cashflows()`. Called `principal_base`, which is the unaltered principal. It maps to `principal` for internal pricing and `notional` for external pricing.

### DIFF
--- a/pallets/loans/src/entities/loans.rs
+++ b/pallets/loans/src/entities/loans.rs
@@ -243,26 +243,21 @@ impl<T: Config> ActiveLoan<T> {
 		}
 	}
 
-	pub fn principal(&self, pool_id: T::PoolId) -> Result<T::Balance, DispatchError> {
-		match &self.pricing {
-			ActivePricing::Internal(inner) => Ok(self
-				.total_borrowed
-				.ensure_sub(self.total_repaid.principal)?),
-			ActivePricing::External(inner) => {
-				let maturity = self.maturity_date();
-				inner.outstanding_principal(pool_id, maturity)
-			}
-		}
+	pub fn principal(&self) -> Result<T::Balance, DispatchError> {
+		Ok(self
+			.total_borrowed
+			.ensure_sub(self.total_repaid.principal)?)
 	}
 
-	pub fn expected_cashflows(
-		&self,
-		pool_id: T::PoolId,
-	) -> Result<Vec<CashflowPayment<T::Balance>>, DispatchError> {
+	pub fn expected_cashflows(&self) -> Result<Vec<CashflowPayment<T::Balance>>, DispatchError> {
 		self.schedule.generate_cashflows(
 			self.repayments_on_schedule_until,
-			self.principal(pool_id)?,
-			self.pricing.interest().current_debt()?,
+			self.principal()?,
+			match &self.pricing {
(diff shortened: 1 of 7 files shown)


pallets/loans/src/types/cashflow.rs
```diff
@@ -173,11 +171,7 @@ impl RepaymentSchedule {
 			.map(|(_, weight)| weight)
 			.try_fold(0, |a, b| a.ensure_add(*b))?;
 
-		let lifetime = maturity.ensure_sub(origination_date)?.ensure_into()?;
-		let interest_rate_per_lifetime = ensure_pow(interest_rate.per_sec()?, lifetime)?;
-		let interest_at_maturity = interest_rate_per_lifetime
-			.ensure_mul_int(principal)?
-			.ensure_sub(principal)?;
+		let interest_at_maturity = current_debt.ensure_sub(principal)?;
```
lemunozm, line 174: Basically, the principal is different depending on the kind of pricing. The current debt is what previously it was:

```rust
let lifetime = maturity.ensure_sub(origination_date)?.ensure_into()?;
let interest_rate_per_lifetime = ensure_pow(interest_rate.per_sec()?, lifetime)?;
let interest_at_maturity = interest_rate_per_lifetime
	.ensure_mul_int(principal)?
```

We can do this simplification now due it is computed as paid at maturity

pallets/loans/src/entities/loans.rs
```diff
@@ -243,26 +243,21 @@ impl<T: Config> ActiveLoan<T> {
...
-				.ensure_sub(self.total_repaid.principal)?),
-			ActivePricing::External(inner) => {
-				let maturity = self.maturity_date();
-				inner.outstanding_principal(pool_id, maturity)
-			}
-		}
+	pub fn principal(&self) -> Result<T::Balance, DispatchError> {
+		Ok(self
+			.total_borrowed
+			.ensure_sub(self.total_repaid.principal)?)
```
mustermeiszer: Why the change? Does this mean, that the gains we are making through price increases are repayed as unscheduled?

lemunozm: It seems like you have reviewed something outdated

lemunozm: But I get your point, it should be just the borrowed amount, not taking into account the repaid one, right?

mustermeiszer: Now its compeltly gone, right? But looks good now

lemunozm: It's still there, I need to remove the `.ensure_sub(self.total_repaid.principal)?` part, to not reduce the cashflow expected amount with the paid amount

lemunozm:
I'll merge anyway both PRs into the same one due you already have reviewed this, and fixing this in the main one.

PR #1865, opened 2024-06-06
feat: `CheckMetadataHash` extension

wischli:
# Description

⚠️ Must not be merged before at least Polkadot Js Extension supports this feature, see https://github.com/paritytech/polkadot-sdk/issues/4659 ⚠️ 

* Adds https://github.com/paritytech/polkadot-sdk/pull/4274 to all runtimes which is necessary for the [Polkadot Generic Ledger App](https://forum.polkadot.network/t/polkadot-generic-ledger-app/4295)
* Changes based on the [corresponding guide](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_metadata_hash/index.html).
* Increases compile time, but only in `on-chain-relese-build` mode
* AFAIU, the signed extension is only compiled in release mode

## TODO

- [x] Before we propose the Centrifuge Release, we need to ensure transactions can be signed properly in Polkadot Apps and Centrifuge App on Demo! cc @mustermeiszer 
- [x] Update CI to build WASMs using `on-chain-release-build` feature
--> Seems to be used by `srtool` [per default](https://github.com/paritytech/srtool/blob/23a401074c9b7f0d62d1da99766643a8e381d6dc/scripts/getBuildOpts.sh#L23) which we use for release builds.

# Checklist:


runtime/altair/src/lib.rs
```diff
@@ -1940,6 +1940,7 @@ pub type SignedExtra = (
 	frame_system::CheckWeight<Runtime>,
 	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
 	runtime_common::transfer_filter::PreBalanceTransferExtension<Runtime>,
+	frame_metadata_hash_extension::CheckMetadataHash<Runtime>,
```
mustermeiszer, line 1943: NIT: It should not make a difference as `runtime_common::transfer_filter::PreBalanceTransferExtension<Runtime>::Extra` is `()`, but could we switch the two? Then we have the same order as the polkadot chains IIRC. 

Same for all other runtimes

lemunozm:
Fixing IT: https://github.com/centrifuge/centrifuge-chain/pull/1866

runtime/altair/src/lib.rs
wischli: [22d0a38](https://github.com/centrifuge/centrifuge-chain/pull/1865/commits/22d0a3898a9622424dac70799120839847bfec41)

runtime/integration-tests/src/generic/config.rs
```diff
@@ -282,6 +282,7 @@ pub trait Runtime:
 				frame_system::CheckWeight<Self>,
 				pallet_transaction_payment::ChargeTransactionPayment<Self>,
 				runtime_common::transfer_filter::PreBalanceTransferExtension<Self>,
+				frame_metadata_hash_extension::CheckMetadataHash<Self>,
```
lemunozm, line 285: You would need to change the order also here

runtime/integration-tests/src/generic/env.rs
```diff
@@ -187,6 +187,7 @@ pub mod utils {
 			frame_system::CheckWeight::<T>::new(),
 			pallet_transaction_payment::ChargeTransactionPayment::<T>::from(0),
 			runtime_common::transfer_filter::PreBalanceTransferExtension::<T>::new(),
+			frame_metadata_hash_extension::CheckMetadataHash::<T>::new(true),
```
lemunozm, line 190: and here

lemunozm (commented):
Not super expert about this but LGTM!

mustermeiszer (commented):
@wischli does this mean that we need to build local wasms always with `--features on-chain-relese-build` in order to get the correct wasm?

wischli:
> @wischli does this mean that we need to build local wasms always with `--features on-chain-relese-build` in order to get the correct wasm?

That appears to be the case, yes. Bot the [Polkadot SDK](https://github.com/paritytech/polkadot-sdk/blob/851c824dc7a8e7e6db2d0c9fb29d232f2f45198a/.github/workflows/build-and-attach-release-runtimes.yml#L26) as well as [Fellowship runtimes](https://github.com/polkadot-fellows/runtimes/blob/2539cb587ed67ac83fff1b02d86c6b7eb7542333/.github/workflows/release.yml#L121) uses this feature for release builds.

I am a bit confused though because for testing purposes, I used the default `release` WASM which did not work with signature verification.

wischli:
Reconverted to draft until Fellowships SDK v1.2.5 is near and [all extensions fixed](https://github.com/paritytech/polkadot-sdk/issues/4659).

runtime/integration-tests/src/generic/config.rs
wischli: Thanks, of course!

runtime/development/src/migrations.rs
```diff
@@ -10,142 +10,6 @@
...
-				.saturated_into();
-			assert_eq!(n, n_count);
-
-			log::info!("{LOG_PREFIX} post_upgrade done with {n} remaining storage keys!",);
-
-			Ok(())
-		}
-	}
-}
+pub type UpgradeDevelopment1102 = ();
```
lemunozm, line 15: 👍🏻

runtime/integration-tests/src/generic/env.rs
```diff
@@ -247,7 +247,7 @@ pub mod utils {
 			frame_system::CheckNonce::<T>::from(nonce),
 			frame_system::CheckWeight::<T>::new(),
 			pallet_transaction_payment::ChargeTransactionPayment::<T>::from(0),
-			frame_metadata_hash_extension::CheckMetadataHash::<T>::new(true),
+			frame_metadata_hash_extension::CheckMetadataHash::<T>::new(false),
```
wischli, line 250: @lemunozm Unfortunately, multiple of our integration tests fail with `CheckMetadataHash` enabled: https://github.com/centrifuge/centrifuge-chain/actions/runs/9665090255/job/26661355826

I am not fully sure why. Based on the `build.rs`, it is only enabled for `--features=std,metadata-hash`: https://github.com/centrifuge/centrifuge-chain/blob/2e8e44cb814d7fa5c0f18a0a121ff82527909f15/runtime/altair/build.rs#L1-L6

Moreover, it is enabled for `on-chain-release-build`: https://github.com/centrifuge/centrifuge-chain/blob/2e8e44cb814d7fa5c0f18a0a121ff82527909f15/runtime/development/Cargo.toml#L450-L457

I tried both running integration tests with `-F metadata-hash` as well as enabling it for the runtimes in the `integration-tests/Cargo.toml` but ran into polkadot-sdk compiler issues. So it appeared to be the incorrect approach. Upon further investigating the SDK repo, I saw all their integration tests disable the metadata hash. However, this might be due to performance reasons. Apparently, the WASM needs to be built twice if the feature is enabled (see [guide](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_metadata_hash/index.html#enable-metadata-hash-generation)).

wischli: Yeah I also think it is some missing feature for a polkadot crate. Tried out a lot and still haven't surpassed the error
```
error[E0433]: failed to resolve: could not find `metadata_ir` in `__private`
   --> /Users/william/.cargo/git/checkouts/polkadot-sdk-cff69157b985ed76/c00ca35/substrate/primitives/api/src/lib.rs:797:1
    |
797 | / decl_runtime_apis! {
798 | |     /// The `Core` runtime api that every Substrate runtime needs to implement.
799 | |     #[core_trait]
800 | |     #[api_version(4)]
...   |
827 | |     }
828 | | }
    | |_^ could not find `metadata_ir` in `__private`
```

It can be quickly checked by trying to build any runtime with `on-chain-release` feature. In the Polkadot SDK this works for the template runtime. For us, it doesn't for any runtime.

lemunozm: Fails only the test who makes a call to `submit_now()` or `submit_later()`, because those calls sign the transaction. Other calls just mutate the state

lemunozm: I'm also stuck in the above error. Quite weird. It seems like `frame-metadata` feature is missing for `sp-api`. But even if you try adding it, the error still happens 🤔. It seems like who add `sp-api` first in the `cargo tree` should add the `frame-metadata` to it.

runtime/integration-tests/src/env.rs
```diff
@@ -255,7 +255,7 @@ pub mod utils {
 			frame_system::CheckNonce::<T>::from(nonce),
 			frame_system::CheckWeight::<T>::new(),
 			pallet_transaction_payment::ChargeTransactionPayment::<T>::from(0),
-			frame_metadata_hash_extension::CheckMetadataHash::<T>::new(false),
+			frame_metadata_hash_extension::CheckMetadataHash::<T>::new(true),
```
lemunozm, line 258: If we have this `false`, then IT tests work?

wischli: Yes because it disables the metadata hash check.

runtime/integration-tests/src/generic/env.rs
wischli: Yeah I have been there as well. Good to know I havent missed anything obvious and sad this doesnt seem to work right now for us. Will probably open an issue in the SDK then.

runtime/integration-tests/src/env.rs
lemunozm: I think we can skip checking the metadata for IT. What knowledge gives us IT if that value is `true`?

wischli: I wanted to make the same proposal. In the end, we expect the SDK to ensure the validity of this API.

lemunozm: For signed-related things, we also mock the nonce and similar things for IT, so I think we're fine not checking that. It feels like out of scope (?)

runtime/integration-tests/src/generic/env.rs
lemunozm: Yeah, Although I think this is more a "rust" problem. It seems unmaintenable that each time you add a feature to a crate, you "need" to propagate that feature to the crates that use the featured gated crate. Ideally, you need to add it even to those crates that don't require that feature. Just for the case that a possible user want to enable it.

Maybe the solution here from polkadot would be to just remove the `frame-metadata` feature. Include `metadata_ir` always. The only reason to add the feature is to avoid the double wasm compilation. But here, it could be avoided.

wischli:
During engineering sync, we decided to add this feature into the release due to the signature success on/with Polkadot Js, Polkadot Js Extension, Polkadot Vault, Talisman and as of last night, Centrifuge Apps.

PR #1867, opened 2024-06-07
Loans: cashflows simple version

lemunozm:
# Description

A simplified version of https://github.com/centrifuge/centrifuge-chain/pull/1797 when only one payment at maturity is expected. No monthly cashflows


mustermeiszer (approved):

pallets/loans/src/entities/loans.rs
```diff
@@ -364,6 +384,8 @@ impl<T: Config> ActiveLoan<T> {
 			}
 		}
 
+		self.repayments_on_schedule_until = T::Time::now();
```
lemunozm, line 387: @mustermeiszer @hieronx should I add this line into the `repay()` method as well?

mustermeiszer: It is atm not having any effect, right? Then I would do so.

lemunozm: It has, because the `cashflow()` is computed to start from that point: https://github.com/centrifuge/centrifuge-chain/pull/1867/files#diff-637356fbd7809c17265d6b8ab519abc2b0ffa4df672cfb33a321795190f441baR254

Then I'll add it

mustermeiszer (approved):

wischli:
Could you add a very short description of the breaking API change to the PR description please @lemunozm? Will use this for the release.

lemunozm:
Sorry! it was in the old closed PR:

> `breaks-api` tag due:
> - Loans API to V3 (not exactly breaking)
> - `InterestPayments::None` renamed to `InterestPayments::OnceByMaturity`

PR #1871, opened 2024-06-13
Fix FMT linter

gpmayorga:
# Description

For some reason the FMT linter is broken in main and thus all PRs merging from main are failing to pass the linter. 
The only plausible explanation is that the FMT version has changed and not has a default rule that is different from the past because some of the code changes suggested by fmt are lines from years ago.


gpmayorga:
It's still puzzling that it woulnd't have failed in on of the previous PRs and made it to main because the way I udnerstand rust the fmt version is totally pinned down?

lemunozm:
> The only plausible explanation is that the FMT version has changed

Agree. I can figure out another option.

> It's still puzzling that it woulnd't have failed in on of the previous PRs and made it to main because the way I udnerstand rust the fmt version is totally pinned down?

Here is my hypothesis. We now configure the toolchain for stable rust. But we still want to use the nightly capabilities of `cargo fmt` (that why we use `+nightly`). But when fetching the nightly version associated, because the nightly does not belong to the toolchain it's not pinned down.

But if that is true, I do not know how to fix it. Maybe forcing installing some nightly tagged version (?)

lemunozm (approved):

lemunozm:
It seems like Frederik already fixed these issues in his PR and is merged automatically if the diff is 0 😆

PR #1878, opened 2024-06-18
pallets: Add Anchors V2 pallet

cdamian:
# Description

Adds a new anchors pallet, required for the data protocol.

# Checklist:


runtime/development/src/lib.rs
```diff
@@ -2127,6 +2142,7 @@ construct_runtime!(
 		ForeignInvestments: pallet_foreign_investments::{Pallet, Storage, Event<T>} = 117,
 		OraclePriceFeed: pallet_oracle_feed::{Pallet, Call, Storage, Event<T>} = 118,
 		OraclePriceCollection: pallet_oracle_collection::{Pallet, Call, Storage, Event<T>} = 119,
+		AnchorsV2: pallet_anchors_v2::{Pallet, Call, Storage, Event<T>} = 130,
```
cdamian, line 2145: Are we ok with this index?

runtime/development/src/weights/pallet_anchors_v2.rs
```diff
@@ -0,0 +1,51 @@
+
```
cdamian, line 1: Reminder to add actual weights.

lemunozm: They are computed automatically before each RU, so we're safe even if we forget

pallets/anchors-v2/src/benchmarking.rs
```diff
@@ -0,0 +1,89 @@
...
+use frame_system::RawOrigin;
+use parity_scale_codec::EncodeLike;
+use scale_info::prelude::format;
+use sp_core::H256;
+use sp_runtime::traits::Hash;
+
+use super::*;
+
+benchmarks! {
+	where_clause {
```
lemunozm, line 25: There is a `benchmarking::v2` that we should use instead, for reference: `pallet_investments` among others.

pallets/anchors-v2/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+	deposit: Balance,
+}
+
+pub type AnchorOf<T> = Anchor<
+	<T as frame_system::Config>::AccountId,
+	<T as Config>::DocumentId,
+	<T as Config>::DocumentVersion,
+	<T as frame_system::Config>::Hash,
+	<T as Config>::Balance,
+>;
```
lemunozm, line 48: NIT. Could we use `Anchor<T: Config>` instead? Much easier to refactor in the future

pallets/anchors-v2/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+			origin: OriginFor<T>,
+			document_id: T::DocumentId,
+			document_version: T::DocumentVersion,
+			hash: T::Hash,
+		) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+
+			// Only one anchor should be stored for a particular document ID and version.
+			ensure!(
+				Anchors::<T>::iter_prefix_values((document_id, document_version)).count() == 0,
```
lemunozm, line 195: This iterates the whole storage, which could lead to a big weight. 
I think `Anchors::<T>::get((document_id, document_version)).is_none()` do the expected, isn't? As you did below

pallets/anchors-v2/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+				deposit: anchor.deposit,
+			});
+
+			Ok(())
+		}
+
+		/// Set a new anchor deposit.
+		#[pallet::weight(T::WeightInfo::set_deposit())]
+		#[pallet::call_index(2)]
+		pub fn set_deposit(origin: OriginFor<T>, new_deposit: T::Balance) -> DispatchResult {
```
lemunozm, line 277: NIT: I would call it `set_deposit_value` or similar, if not it seems like this call makes the deposit when sets the value for future deposits.

pallets/anchors-v2/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+		/// The type used to identify a document.
+		type DocumentId: Member
+			+ Parameter
+			+ AtLeast32BitUnsigned
+			+ Default
+			+ Copy
+			+ MaxEncodedLen
+			+ FixedPointOperand
+			+ From<u64>
+			+ From<u128>
```
lemunozm, line 84: Does it really need to be a number?

pallets/anchors-v2/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+	pub type PersonalAnchors<T: Config> = StorageNMap<
+		_,
+		(
+			NMapKey<Blake2_256, T::AccountId>,
+			NMapKey<Blake2_256, T::DocumentId>,
+			NMapKey<Blake2_256, T::DocumentVersion>,
+		),
+		AnchorOf<T>,
+		OptionQuery,
+	>;
```
lemunozm, line 134: To avoid storing the anchor twice (which leads to maintaining the same data in two storages), I would create a simple `StorageMap<AnchorId, Anchor>` and would store in these two maps the `AnchorId` instead of the `Anchor`.

You would have another Read operation but the main data is stored just once.

lemunozm: Another option, because the keys of both maps are the same, then if the anchor is found in `PersonalAnchors,` you know that it can also be found in `Anchors` So `PersonalAnchors` could have `()` as a value. Then if `Some(())`, then you can retrieve the anchor data in `Anchors`.

lemunozm (commented):
I think the logic is perfect! 💯 Very clean and easy to follow. 

Some NITs, questions and suggestions, nothing severe

runtime/development/src/lib.rs
lemunozm: I'll place it where we write: `// out pallets part 2` to no mix indexes

pallets/anchors-v2/src/benchmarking.rs
cdamian: Yup, thanks for the pointer ❤️

pallets/anchors-v2/src/lib.rs
cdamian: Yup, I also agree that the current approach needs improvement. I can't do the call you suggested since it's a double map, but I'll look into removing the account ID key and try out what you suggested in a different comment.

pallets/anchors-v2/src/lib.rs
cdamian: Yes, given the current spec that we have, the id should be a u128 and the version a u64.

pallets/anchors-v2/src/lib.rs
cdamian: It's what I did initially and got a warning:

```
warning: bounds on generic parameters are not enforced in type aliases
  --> pallets/anchors-v2/src/lib.rs:42:22
   |
42 | pub type AnchorOf<T: Config> =
   |                      ^^^^^^
```

pallets/anchors-v2/src/lib.rs
cdamian: Thank you for the suggestions, I will definitely explore these and get back to you!

pallets/anchors-v2/src/lib.rs
lemunozm: I meant in the struct:

```rust
pub struct Anchor<T: Config> {
	account_id: T::AccountId,
	document_id: T::DocumentId,
	document_version: T::DocumentVersion,
	hash: T::Hash,
	deposit: T::Balance,
}
```

pallets/anchors-v2/src/lib.rs
lemunozm: But from the pallet perspective, is it needed? Or is it something like we want to configure later the id as `u128` and the version as `u64` in the runtime?

pallets/anchors-v2/src/lib.rs
lemunozm: We had a similar approach initially in the `pallet_order_book`, and later we changed it because it did not scale well. You can check `Orders` and `UserOrders` storages for reference

pallets/anchors-v2/src/lib.rs
lemunozm: Ahh you're right. Then I think this can be useful: https://paritytech.github.io/polkadot-sdk/master/frame_support/storage/types/struct.CountedStorageNMap.html

lemunozm: I was thinking about this.

If we only allow calling `set_anchor()` if the count is 0, it means that the total entries can only be 0 or 1, no more. Does it means that the storage could be modified to a single `StorageMap`?:

```rust
    pub type Anchors<T: Config> = StorageMap<
		_,
		Blake2_256,
		(T::DocumentId, T::DocumentVersion),
		(T::AccountId, AnchorOf<T>),
	>;
```

pallets/anchors-v2/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+				sufficients: 0,
+				data: AccountData::<Balance> {
+					free: 0,
+					reserved: deposit,
+					frozen: 0,
+					flags: Default::default(),
+				},
+			};
+
+			Account::<Runtime>::insert(origin, account_info);
```
lemunozm, line 215: Out of curiosity, why not `Balance::reserve(origin, deposit)` here?

pallets/anchors-v2/src/lib.rs
cdamian: Hmm, I think that we'll always use `u128`/`u64`.

pallets/anchors-v2/src/tests.rs
cdamian: Copy/pasta from keystore tests ^^

Gonna change this and other similar case.

pallets/anchors-v2/src/lib.rs
lemunozm: Then, do we need associated types?

cdamian: Nope, I removed these.

pallets/anchors-v2/src/lib.rs
cdamian: I updated the storage setup.

pallets/anchors-v2/src/lib.rs
cdamian: Updated, thank you!

pallets/anchors-v2/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+	/// Storage for anchors.
+	#[pallet::storage]
+	#[pallet::getter(fn get_anchor)]
+	pub type Anchors<T: Config> = StorageMap<_, Blake2_256, T::AnchorIdNonce, Anchor<T>>;
+
+	/// Storage for document anchors.
+	#[pallet::storage]
+	#[pallet::getter(fn get_document_anchor)]
+	pub type DocumentAnchors<T: Config> =
+		StorageMap<_, Blake2_256, (DocumentId, DocumentVersion), T::AnchorIdNonce>;
```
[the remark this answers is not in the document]
cdamian: Yes, but then storing personal anchors might be problematic, or we can just use a double map and store `()`. Lemme try that out.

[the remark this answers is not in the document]
cdamian: Done, please let me know what you think.

cdamian:
Not sure why the fmt lint is acting up, I don't get the same errors locally...

pallets/anchors-v2/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Love it! Super simple now! 💯

pallets/anchors-v2/src/benchmarking.rs
```diff
@@ -21,7 +21,7 @@ use super::*;
 
 #[benchmarks(
 	where
-		T: Config<Balance = u128, Hash = H256, AnchorIdNonce = u128>,
+		T: Config<Balance = u128, Hash = H256>,
```
lemunozm, line 24: This is correct, and probably this pallet is only for us, so it will not be an issue. But to explain the why:

These bounds force the user of this pallet to always choose `Balance = u128` and `Hash = H256` in their runtime if they want to perform the benchmarks. Ideally, we do not want to restrict the usage of the bounds to certain concrete types. We want to allow *any* type that can be configured by the pallet.

The solution for that is to add extra bounds to those associated types as:
```rust
T: Config,
T::Balance: Into<u128>, // Or maybe T::Balance: Default is just enough
T::Hash: Default, // And use later T::Hash::default() instead of H256::from_low_u64_be(1)

lemunozm (approved):
LGTM @cdamian! Good job with this pallet, everything done in less than 1K lines, super clean! 💯

pallets/anchors-v2/src/benchmarking.rs
lemunozm: Yeah, not blocker 👍🏻

lemunozm (approved):

PR #1828, opened 2024-05-03
feat: Open Gov for Development and Altair Runtimes

wischli:
# Description

* Adds OpenGov for Development and Altair Runtimes which coexists with current Gov structure
* Altair config based on https://gov.centrifuge.io/t/rfc-altair-migration-to-opengov/6011
  * More thorough [Google Doc for Altair tracks and parametrization](https://docs.google.com/document/d/1asoDHtIT3bhSnwfQir3SZ7NN-5D1i02s01clh9lEd0k)
* Development config based on Centrifuge Chain configuration: https://gov.centrifuge.io/t/rfc-centrifuge-migration-to-opengov/5984
  * Time scale entirely mapped to 1 minute: 
```diff
// centrifuge runtime
x_period: 2 * MINUTES,
- y_period: 3 * HOURS,
- z_period: 4 * DAYS,

// development runtime
x_period: 1 * MINUTES,
+ y_period: 1 * MINUTES,
+ z_period: 1 * MINUTES,
```
  * More thorough [Google Doc for Centrifuge tracks and parametrization](https://docs.google.com/document/d/1aPH1sTfAttD3K0f4AoOF2iDb3nH1GHlx5TfIP4Xi4SM)

## Changes and Descriptions

* Adds `origins` module to `runtime_common` necessary for gov origin declaration which is the same across all runtimes
* Adds necessary OpenGov pallets to Development and Altair runtimes (Centrifuge will follow in ~1 month)
* Initializes Technical Committee for `WhitelistedCaller` track via migration based on [CP86](https://gov.centrifuge.io/t/cp86-opengov-technical-committee-mandate-request/6085)
  * Requires `pallet_membership` for membership handling within the collective and 
  * `pallet_collective::Instance2` for membership handling via root as well origin derivation for the `WhitelistedCaller` track

## TODO

All done 14.05.2024

- [x] OpenGov Development Runtime
- [x] OpenGov Altair Runtime
- [x] Revert OpenGov addition to Centrifuge Runtime
- [x] Write migration to initialize Technical Committee

## Altair migrations

```
[2024-05-14T13:00:39Z INFO  remote-ext] initialized state externalities with storage root 0xa907b41b327bb68ffcdd9902a0b5cffe8bccae5a2ccc7b7cab8fde43bc18c272 and state_version V0
[2024-05-14T13:00:40Z INFO  try-runtime::cli] Original runtime [Name: RuntimeString::Owned("altair")] [Version: 1034] [Code hash: 0xbdc0...5150]
[...]

PR #1879, opened 2024-06-19
Liquidity-pools: Add unitary tests

lemunozm:
# Description

We didn't have UTs for the liquidity pool. Although the Integration Tests cover more or less the basics of `liquidity-pools`, having those checks there makes Integration Tests more boilerplate. We can simplify some tests if the basics are moved to the unitary tests.

### Changes:
- Some cleanings / simplifications
- Added UT tests to check all errors when transferring
- Removed `liquidity::gateway` tests, which are correctly checked in UTs.
- Removed IT `utils::democracy` which are no longer needed (see comment below)
- Removed `liquidity_pools::transfer` covered in UTs.

**BLOCKED BY https://github.com/centrifuge/centrifuge-chain/pull/1869**


runtime/common/src/account_conversion.rs
```diff
@@ -89,20 +94,6 @@ impl Convert<(Domain, [u8; 32]), DomainAddress> for AccountConverter {
...
-			// EVM AccountId20 addresses are right-padded to 32 bytes
-			Domain::EVM(chain_id) => {
-				let mut bytes20 = [0; 20];
-				bytes20.copy_from_slice(&account[..20]);
-				Self::convert_evm_address(chain_id, bytes20)
-			}
-		}
-	}
-}
-
```
lemunozm, line 105: We can implement this converter as a composition of the above two converters. This is valid for `liquidity pools, ' and we do not need to pass it as a `Convert` trait to that crate.

runtime/common/src/account_conversion.rs
```diff
@@ -65,6 +65,11 @@ impl AccountConverter {
 		let chain_id = pallet_evm_chain_id::Pallet::<R>::get();
 		Self::convert_evm_address(chain_id, address.0)
 	}
+
+	pub fn domain_account_to_account(domain: Domain, account_id: AccountId) -> AccountId {
+		let domain_address = Self::convert((domain, account_id.into()));
+		Self::convert(domain_address)
+	}
```
lemunozm, line 72: A utility is still here for testing or other stuff

runtime/common/src/account_conversion.rs
```diff
@@ -61,10 +61,15 @@ impl AccountConverter {
 		}
 	}
 
-	pub fn into_account_id<R: pallet_evm_chain_id::Config>(address: H160) -> AccountId {
+	pub fn evm_to_account<R: pallet_evm_chain_id::Config>(address: H160) -> AccountId {
```
lemunozm, line 64: Modified the name (open to other names that match better) for two reasons:
- The `into_*` names are usually used when the own type is converted into something, not when the thing to convert is used as a parameter. The name `into_account_id()` was inherited from the name used by the trait `AddressMapping`, not chosen by us. Anyways, the `into_account_id()` is still accessible through the trait.
- The new name follows the format: `FROM_to_TO`, as other methods:
  - `location_to_account()`
  - `domain_account_to_account()`

runtime/integration-tests/src/generic/utils/democracy.rs
```diff
@@ -1,276 +0,0 @@
-use std::ops::Add;
```
lemunozm, line 1: I've removed this file because:
- No longer used in our gateway tests, which were correctly covered by its UTs.
- It tests just democracy stuff which is not under our umbrella. We do not want to maintain this.
- No longer used after OpenGov (?): https://github.com/centrifuge/centrifuge-chain/pull/1828

Anyways, I am open to discussing this.

pallets/liquidity-pools/src/mock.rs
```diff
@@ -0,0 +1,143 @@
...
+	pub enum Runtime {
+		System: frame_system,
+		Time: cfg_mocks::time::pallet,
+		Permissions: cfg_mocks::permissions::pallet,
+		Pools: cfg_mocks::pools::pallet,
+		AssetRegistry: cfg_mocks::asset_registry::pallet,
+		ForeignInvestment: cfg_mocks::foreign_investment::pallet,
+		Gateway: cfg_mocks::outbound_queue::pallet,
+		DomainAddressToAccountId: cfg_mocks::converter::pallet::<Instance1>,
+		DomainAccountToDomainAddress: cfg_mocks::converter::pallet::<Instance3>,
```
cdamian, line 28: Why do we have to use `Instance3` here? IIRC we used maximum 2 instances in tests.

runtime/common/src/account_conversion.rs
cdamian: Nit - `evm_address_to_account_id` ^^

cdamian (dismissed):
Ultra-clean, as always :D Thank you @lemunozm!

runtime/common/src/account_conversion.rs
lemunozm: Like it more!

pallets/liquidity-pools/src/mock.rs
lemunozm: Right! I had an `Instance2` before. I will change it! Good eye 😆

pallets/liquidity-pools/src/lib.rs
```diff
@@ -1097,25 +1068,3 @@ pub mod pallet {
 		}
 	}
 }
-
-#[cfg(test)]
-mod tests {
```
lemunozm, line 1102: moved to `domain_address.rs` tests

runtime/integration-tests/src/generic/utils/democracy.rs
wischli: I agree with you!

wischli (dismissed):
Great improvement! Loving the clean distinction between all test scenarios instead of the previous huge blocks. I have two minor nits which can be ignored.

pallets/liquidity-pools/src/tests.rs
```diff
@@ -0,0 +1,622 @@
...
+				);
+				Ok(())
+			});
+
+			assert_ok!(LiquidityPools::transfer(
+				RuntimeOrigin::signed(ALICE),
+				CurrencyId::ForeignAsset(1),
+				EVM_ADDRESS,
+				AMOUNT
+			));
```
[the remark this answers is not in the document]
lemunozm: Initially I removed that check because it's checking that orml_tokens is correctly working (which is out of scope from this UT test), but thinking again, it makes sense to check that actually this method calls correctly to orml_tokens, with the correct amount. So, will do 👍🏻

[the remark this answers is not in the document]
lemunozm: Very good point BTW

pallets/liquidity-pools/src/tests.rs
```diff
@@ -0,0 +1,622 @@
...
+				Ok(())
+			});
+
+			assert_ok!(LiquidityPools::transfer_tranche_tokens(
+				RuntimeOrigin::signed(ALICE),
+				POOL_ID,
+				TRANCHE_ID,
+				EVM_ADDRESS,
+				AMOUNT
+			),);
```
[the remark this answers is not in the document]
lemunozm: Done! I've also simplified in the process some minor thing regarding this: [ce939ca](https://github.com/centrifuge/centrifuge-chain/pull/1879/commits/ce939caad83e852209f6703fe146481b22fdb91e)

Generally, the `Convert` trait is quite obscure when you have different conversions, and in that case, we didn't need it.

lemunozm:
@wischli @wischli, ready for your re-approvals and get this in 🙏🏻

wischli (approved):
Thanks for adhering to my nits!

PR #1880, opened 2024-06-20
IT: Panic if event is not found in the expected blocks

lemunozm:
# Description

The current behavior for `env.pass(Blocks::UntilEvent{event, limit})` evolve the chain a maximum blocks of `limit` looking for the event. Nevertheless if the event is not found nothing happens. 

That means that later, we need to ensure that the event is found. We can avoid that if we panic if the event is not found during `pass()` method.

### DIFF
--- a/runtime/integration-tests/src/generic/cases/routers.rs
+++ b/runtime/integration-tests/src/generic/cases/routers.rs
@@ -140,12 +140,9 @@ fn check_submission<T: Runtime>(mut env: impl Env<T>, domain_router: DomainRoute
 	});
 
 	env.pass(Blocks::UntilEvent {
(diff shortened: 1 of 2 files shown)


cdamian (dismissed):

lemunozm:
CI issues will be fixed after https://github.com/centrifuge/centrifuge-chain/pull/1879

lemunozm:
Finally, it is ready for review. Can I have your reapproval @cdamian?

cdamian (approved):

lemunozm:
@cdamian sorry I dismissed your review 😅

cdamian (approved):

PR #1881, opened 2024-06-20
Switch to full wasm execution

bkchr:
# Description

Stops using the `NativeElseWasmExecutor` and replaces it with the `WasmExecutor`. Thus, block production/import etc is also using the wasm execution as the validation on the relay chain.

Fixes https://github.com/paritytech/polkadot-sdk/issues/4808

## Changes and Descriptions

[List your changes here]

# Checklist:

### DIFF
--- a/node/src/command.rs
+++ b/node/src/command.rs
@@ -27,7 +27,7 @@ use sp_runtime::traits::AccountIdConversion;
 use crate::{
 	chain_spec,
 	cli::{Cli, RelayChainCli, Subcommand},
-	service::{evm, AltairRuntimeExecutor, CentrifugeRuntimeExecutor, DevelopmentRuntimeExecutor},
+	service::evm,
 };
 
 pub const LOCAL_PARA_ID: ParaId = ParaId::new(2000u32);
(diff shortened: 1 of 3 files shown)


mustermeiszer:
First of all, thanks a lot for all the effort taken into researching and solving this. I am still not sure I follow the process here: 

* Should we anyways switch to full wasm execution - we could also defer to use the `MaxEncodedLen` in the feed size?
* Why is the chain able to make progress? 
  *. IIUC the parachain tried to include said extrinsic in the block - events were popping up - but the relay-chain could not validate such block - still block production continued, just without said extrinsic being included.

bkchr:
> * Should we anyways switch to full wasm execution - we could also defer to use the `MaxEncodedLen` in the feed size?

Yes you should switch to full wasm execution. This is the future any way and the `NativeElseWasmExecutor` is tagged as deprecated in master already.



> * Why is the chain able to make progress?
>       *. IIUC the parachain tried to include said extrinsic in the block - events were popping up - but the relay-chain could not validate such block - still block production continued, just without said extrinsic being included.

Yeah without the transaction it could progress, because then the mismatch wasn't triggered.

mustermeiszer:
Replaced by https://github.com/centrifuge/centrifuge-chain/pull/1882. All kudos to @bkchr!

PR #1884, opened 2024-06-23
Script to upgrade a Centrifuge parachain live

gpmayorga:
# Description

Upgrade a tesnet Centrifuge runtime by connecting to the node and uploading a new wasm using JS

Tested with sudo rights on development and demo.

In case we have control over the council on Altair or we have a parachain testnet with council and democracy in the future I left the code that allows for performing a RU using council keys


scripts/js/runtime-upgrade-remote/package-lock.json
```diff
@@ -0,0 +1,853 @@
+{
```
lemunozm, line 1: I was wondering if we needed to commit this, and yes, we need it: https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

gpmayorga:
We'll wait for @wischli to try this out and upgrade Demo, if it works I'd consider it ready

gpmayorga:
Since we don't have an environment where we control the council keys, I'm going to remove the council motion from the automation but leave the govermnent proposal

scripts/js/runtime-upgrade-remote/package-lock.json
wischli: I would prefer the `yarn.lock` to follow the `runtime-upgrade-local` path. AFAIK, yarn locks are preferred.

wischli: done in 4e38fda8b5550d955907c227ca518476171054bc

wischli (approved):
@gpmayorga Thanks a lot for making our lives easier!

I removed the council related code paths because in the foreseeable future, we don't need to upgrade a non-local chain without sudo. If we need it at some point, we can migrate it over from the local script.

Later this year, I will refactor both scripts to use the same code paths as mentioned in the sync to decrease maintenance overhead. In the end, both scripts are very very similar. For now, there are higher priorities and this script worked well for Dev and Demo.

PR #1889, opened 2024-06-26
LP Message with custom data format for serialization/deserialization

lemunozm:
# Description

Current serialization/deserialization methods don't scale well and are error-prone. This PR adds a custom `serde` data format that implements our liquidity pool's Generic Message Passing Format. This would remove entirely the need for manual serialization/deserialization code


lemunozm:
Currently, the serialization is the same except for the 0 padding used for enum variants and the `Domain` type. I think both can be tricked to fit the standard.

wischli (commented):
I wish this PR existed 24 hours ago when I decided to start the message refactoring while waiting for compilers 🥲

Would be awesome if this worked out!

lemunozm:
I think once we get the correct serialization, it can be easily ported to your branch.

lemunozm:
Both issues are now fixed and the serialization & deserialization is the same as expected

pallets/liquidity-pools/src/message.rs
```diff
@@ -367,562 +408,54 @@ where
...
-pub fn deserialize<const S: usize, O: Codec, I: Input>(
-	input: &mut I,
-) -> Result<O, parity_scale_codec::Error> {
-	let mut bytes = [0; S];
-	input.read(&mut bytes[..])?;
+	pub fn deserialize(data: &[u8]) -> Result<Self, DispatchError> {
+		// Bincode serializes the enum tag variants as 4 bytes.
+		// But solidity side expect 1 byte.
+		// We emulate here as if it was so.
+		let data = [vec![0, 0, 0], data.into()].concat();
```
lemunozm, line 455: I do not like this, ideally we should overwrite the `serialize_enum` and `deserialize_enum` from `bincode` serializer to force 1 bytes per variant instead of 4 bytes.

Ref: https://docs.rs/bincode/latest/src/bincode/de/mod.rs.html#284

pallets/liquidity-pools/src/message.rs
```diff
@@ -50,13 +89,13 @@ where
 	///
 	/// Directionality: Centrifuge -> EVM Domain.
 	AddPool {
-		pool_id: PoolId,
+		pool_id: u64,
```
mustermeiszer, line 92: Why the change to raw types. I do not really like that.

mustermeiszer (commented):
Not sure I like the change. The raw types look weird, but the spec itself matches the code now better. 

No strong opinion as long as it is compatible. The main issue for me is that we do not control the serialization here. But tests should cover everything.

pallets/liquidity-pools/src/message.rs
lemunozm: We could avoid that, but the message is who dictates the serialization format. If tomorrow we change the `PoolId` type in the runtime, then the message format will change silently. So this module should know exactly how to serialize/deserialize each type, because this module is the module that implements the spec.

mustermeiszer: Yes, sounds good!

mustermeiszer:
THen lets go with this approach.

pallets/liquidity-pools/Cargo.toml
```diff
@@ -50,6 +53,7 @@ std = [
   "orml-traits/std",
   "staging-xcm/std",
   "scale-info/std",
+  "serde/std",
 ]
```
wischli, line 57:
```suggestion
  "serde/std",
  "bincode/std"
]
```

Cargo.toml
```diff
@@ -61,6 +61,8 @@ hex = { version = "0.4.3", default-features = false }
 smallvec = "1.11.0"
 serde = { version = "1.0.195", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.111" }
+serde-big-array = { version = "0.5" }
+bincode = { version = "1.3.3" }
```
wischli, line 65: This crate has `std` feature ([source](https://github.com/bincode-org/bincode/blob/trunk/Cargo.toml)), so I think we should disable default features and add it to the LP's std feature section?

```suggestion
bincode = { version = "1.3.3", default-features = false }
```

pallets/liquidity-pools/Cargo.toml
lemunozm: `bincode` doesn't have it 😓

Cargo.toml
lemunozm: Ok, it seems to have it in version 2, not in `1.3.3` which is the latest release in `crates.io`. The version 2 is still marked as RC, last one is: https://github.com/bincode-org/bincode/releases/tag/v2.0.0-rc.3

I see the following options:
- We wait until they make a formal release as `2.0.0`.
- We target to the github branch `2.0.0-rc3`. I understand there are no issues with the library just, the API can change. At least they say it's "usable" 🤔

wischli:
As I am currently working on the message reordering, I have added all known fields for the new messages already. In order to not do redundant work, I would like this PR to get merged ASAP. However, it should be based on the [new LP v2 feature branch](https://github.com/centrifuge/centrifuge-chain/tree/feat/lp-v2) `feat/lp-v2` which I just branched off from `main`.

 It appears to me that we have found consensus and can finish this PR today @lemunozm @mustermeiszer?

Cargo.toml
lemunozm: It was not noise! It was a certain comment. We can not keep this merged as 1.3.3 because it will not be compiled for wasm.

lemunozm:
Why not leave this directly in `main`? as this does not add anything LPv2 related

lemunozm:
And rebase later LPv2 over it. I understand that LPv2 is a branch that should be rebased over main continuously, right?

wischli:
> Why not leave this directly in `main`? as this does not add anything LPv2 related

Also fine with that!

wischli:
> I understand that LPv2 is a branch that should be rebased over main continuously, right?

Yes

lemunozm:
I'm on the way to clean this up and have it ready.

libs/traits/src/liquidity_pools.rs
```diff
@@ -12,34 +12,35 @@
...
 			vec![0x42]
 		}
 
-		fn deserialize<I: Input>(input: &mut I) -> Result<Self, parity_scale_codec::Error> {
-			match input.read_byte()? {
-				0x42 => Ok(Self),
-				_ => Err("unsupported message".into()),
+		fn deserialize(input: &[u8]) -> Result<Self, DispatchError> {
+			match input.first() {
+				Some(0x42) => Ok(Self),
```
cdamian, line 41: Nit - can we also have a const for this value and use that in the tests?

lemunozm: In which places in the test are you referring?

The only place where we use that value (but as a string) is here: https://github.com/centrifuge/centrifuge-chain/pull/1891/files#diff-0f02c17540413a2c2eb2d6554089fac0df041bbc8f248e5a3945b600761f0fe4R409 (which is true that with some work we can tied both)

In the rest of the tests what we do is check the serialization and deserialization matches.

lemunozm: Actually, the constant is not needed: [4cc4899](https://github.com/centrifuge/centrifuge-chain/pull/1889/commits/4cc4899bb265f3ff6be747a044700e4ebbd71de8)

cdamian: Saw [this](https://github.com/centrifuge/centrifuge-chain/blob/exp/lp-message-bincode/pallets/liquidity-pools-gateway/routers/src/tests.rs#L347) and thought that a different value here might affect the test?

lemunozm:
**Important note:**

`bincode` "almost" matches with the serialization we need. Just the variant enum tag is encoded as `u32` but we use `u8`. This is not an issue because we are "patching" that for the message ids. But something to be aware of.

Ideally, the correct solution should be, creating our own serializer, that calls to all serializer method implementations from `bincode`, except [this](https://docs.rs/bincode/latest/src/bincode/de/mod.rs.html#284) method that we would overwrite it using `u8` encoding for the enum tag. 

We're fine with the current solution as long as we don't use variants/enums in the serialization messages. Also if we use someone, we can create a conversion type that serialize as we want (as it currently happens with the `Domain` type).

wischli:
> We're fine with the current solution as long as we don't use variants/enums inside the serialization messages. But if we use someone, we also can create a conversion type that serialize as we want (as it currently happens with the `Domain` type).

Are you referring to nested enums of our Message enum such as the requirement change [just shared here](https://kflabs.slack.com/archives/C06FRHCRQL9/p1719820472226499)?

lemunozm:
Ok, after some thoughts, I think you need to implement custom serialization for the `InnerMessage`, to be able to force the enum variant as `u8`

```rust
struct Message {
    UpdateRestriction(UpdateRestriction),
}

enum UpdateRestriction {
    UpdateMember {},
    Freeze {},
    Unfreeze {},
}

impl Serialize for UpdateRestriction {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
         match self {
             UpdateMember {} => {
                 serializer.serialize_u8(1);
                 serializer.serialize_field(..)
             },
             Freeze {} => {
                 serializer.serialize_u8(2);
                 serializer.serialize_field(..)
             },
             ..
         }
    }
}

impl<'de> Deserialize<'de> for UpdateRestriction {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        match deserializer.deserialize_u8(u8Visitor) {
            1 => //Deserialize UpdateMember fields
            2 => //Deserialize Freeze,
            ..
        }
    }
}

**Comment by wischli** [comment]
> Ok, after some thoughts, I think you need to implement custom serialization for the `InnerMessage`, to be able to force the enum variant as `u8`
[...]
> ```

That's similar to what I had in mind with my limited understanding of the `bincode` changes impact. I suppose I should try this out on a fork of my `lpv2/message-reorder` branch rebased to this `exp/lp-message-bincode` branch?!

**Comment by lemunozm** [comment]
I guess so. Or maybe merge this today and then rebase the whole feat/lpv2 branch with all sub-branches. Let's see if I can solve the CI issues from this

**Comment by wischli** [comment]
> I guess so. Or maybe merge this today and then rebase the whole feat/lpv2 branch with all sub-branches. Let's see if I can solve the CI issues from this

That would be better overall of course. The lpv2 branch is in sync with `main` right now btw.

**Comment by lemunozm** [comment]
I'm not getting to overpass the CI issue. It seems like when `bincode` has not `std` enable it still imports something from `std` 😭. Maybe we should skip this PR until the `bincode` v2.0.0 formal release 😢. I do not want to block your work.

**Comment by lemunozm** [comment]
If I make:
```
cargo clean
cargo build -p centrifuge-chain -r
```
I obtain the CI issue locally.

If I do:
```
// Comment bincode in cargo.toml
cargo clean
cargo build -p centrifuge-chain -r
// Uncomment bincode in Cargo.toml
cargo build -p centrifuge-chain -r
```
It works 😅

File: pallets/liquidity-pools/src/lib.rs [hunk 6]
```diff
@@ -136,29 +127,34 @@ pub mod pallet {
...
 		type PoolId: Member
 			+ Parameter
 			+ Default
 			+ Copy
 			+ HasCompact
 			+ MaxEncodedLen
-			+ core::fmt::Debug;
+			+ core::fmt::Debug
+			+ Into<u64>
+			+ From<u64>;
```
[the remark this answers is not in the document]
lemunozm: Before closing this, do you have any opinion, or should I leave it as it is?

lemunozm:
I couldn't bypass the CI issue because something is off with `bincode` and `no_std`. 

So, I've decided to implement a custom serializer. It has been simpler than I expected (around 500 lines in total, quite straightforward). I think it's the purest and most scalable solution, allowing us to define the GMPF encoding correctly and clean the `message` module from handling bytes directly, so I would go with it.

cc @mustermeiszer

libs/traits/src/liquidity_pools.rs
```diff
@@ -12,34 +12,35 @@
...
+	use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 	use scale_info::TypeInfo;
 
-	use super::Codec;
+	use super::*;
 
 	#[derive(Debug, Eq, PartialEq, Clone, Encode, Decode, TypeInfo, MaxEncodedLen)]
 	pub struct Message;
-	impl Codec for Message {
+	impl LPEncoding for Message {
```
lemunozm, line 34: BTW, I've changed the name because `Codec` can be easily confused with the `Codec` from `parity-scale-codec`

pallets/liquidity-pools/src/message.rs
lemunozm: Fixed with the custom serializer

mustermeiszer (commented):
Why do we need to always call `gmpf::to_vec()` for a message and not just `serialize()`?

lemunozm:
> Why do we need to always call gmpf::to_vec() for a message and not just serialize()?

Both are the same, actually `serialize()` calls `gmpf::to_vec()` but without failing. I've used directly `gmpf::to_vec()` in the tests because I want to know the explict error if it can not be serialized well.

wischli (commented):
I'm honestly surprised to see that implementing the byte serializer suffices even though it is missing for the deserializer. Thanks for putting so much work into this! From my perspective, we can merge this but I'd like to wait for @mustermeiszer's opinion.

pallets/liquidity-pools/src/message.rs
```diff
@@ -367,616 +412,70 @@ where
...
-		Ratio: Encode + Decode,
-	> Message<Domain, PoolId, TrancheId, Balance, Ratio>
-{
-	/// The call type that identifies a specific Message variant. This value is
-	/// used to encode/decode a Message to/from a bytearray, whereas the head of
-	/// the bytearray is the call type, followed by each message's param values.
-	///
-	/// NOTE: Each message must immutably  map to the same u8. Messages are
-	/// decoded in other domains and MUST follow the defined standard.
-	fn call_type(&self) -> u8 {
```
wischli, line 389: Just to be sure: The call type is now inherited from the order of the enum variants, right? So when I have a nested enum `Y` inside an LP Message `X`, `Y` will have its own inner call type such that `X` has an outer and an inner one.

lemunozm: > The call type is now inherited from the order of the enum variants, right?

Yes

> So when I have a nested enum Y inside an LP Message X, Y will have its own inner call type such that X has an outer and an inner one.

And yes too

lemunozm: Probably we should have a test checking that each msg has the expected index in the spec, to avoid errors swapping variants in the future.

wischli: Yeah I was thinking of exactly that. It is more or less tested in the message serialization cycle UTs in `Message.rs` but these are too opaque to guarantee call type safety. No need for you to do it as part of this PR because I am changing almost all indices anyway. Will do it as part of #1892 then.

lemunozm: > It is more or less tested in the message serialization cycle UTs in Message.rs but these are too opaque to guarantee call type safety

fully agree!

lemunozm: To avoid boilerplate in testing, maybe each test construction can go to a function: i.e. `increase_invest_order_msg()`, and later:

```rust
`assert_eq!(msg_increase_invest_order().serialize()[0], 5)`
```

or similar.

Also, not sure if the it's more manageable to have the message as:

```rust
enum Message {
    AddPool(AddPoolMessage),
    AllowInvestCurrency(AllowInvestCurrencyMessage),
    ..
},
```

It seems like "more" code at first, but later, you can reuse each variant for different purposes. I'm just thinking out loud.

wischli: > Also, not sure if the it's more manageable to have the message as:
> 
> ```rust
> enum Message {
>     AddPool(AddPoolMessage),
>     AllowInvestCurrency(AllowInvestCurrencyMessage),
>     ..
> },
> ```

Not sure if I see the win for doing it this way right now.

wischli (approved):
LGTM!

mustermeiszer (approved):
Enum call type encoding should be checked by integration tests implicitly. If we cover all messages in the next month.

PR #1890, opened 2024-06-27
Liquidity pools: Add UTs to for `update_token_price()`

lemunozm:
# Description

UTs for `update_token_price()`

NOTE: It's not directly related to LPv2

## Changes and Descriptions

- Token price simplification removed some unused types & traits.
- Remove the IT related to this.
- Add UTs covering the possible errors.


libs/traits/src/lib.rs
```diff
@@ -206,31 +206,6 @@ pub trait PoolWriteOffPolicyMutate<PoolId> {
...
-
-	/// Retrieve the latest price of `base` currency, denominated in the `quote`
-	/// currency If `quote` currency is not passed, then the default `quote`
-	/// currency is used (when possible)
-	fn get_latest(
-		base: CurrencyId,
-		quote: Option<CurrencyId>,
-	) -> Option<PriceValue<CurrencyId, Self::Rate, Self::Moment>>;
-}
-
```
lemunozm, line 233: This was not used, but if it's intended to be used very soon, I can revert this. Otherwise, I would be removed.

wischli (dismissed):
LGTM!

pallets/pool-system/src/impls.rs
```diff
@@ -84,21 +84,9 @@ impl<T: Config> TrancheTokenPrice<T::AccountId, T::CurrencyId> for Pallet<T> {
...
 			.ok()?;
 
-		let base = pool
-			.tranches
-			.tranche_currency(TrancheLoc::Id(tranche_id))?
-			.into();
-
 		let price = prices.get(tranche_index).cloned()?;
 
-		Some(PriceValue {
```
mustermeiszer, line 94: @hieronx you added this change. Why did you wanted to have the `TrancheCurrency` and the `PoolCurrency` named here?

pallets/liquidity-pools/src/lib.rs
```diff
@@ -426,7 +426,7 @@ pub mod pallet {
 			// TODO(future): Once we diverge from 1-to-1 conversions for foreign and pool
 			// currencies, this price must be first converted into the currency_id and then
 			// re-denominated to 18 decimals (i.e. `Ratio` precision)
-			let price_value = T::TrancheTokenPrice::get(pool_id, tranche_id)
+			let (price, computed_at) = T::TrancheTokenPrice::get_price(pool_id, tranche_id)
```
mustermeiszer, line 429: @lemunozm can we in the same run also close this todo and computed the conversion?

mustermeiszer (commented):
Please do not merge until hieronx answered. I also would like to close the conversion in the same run too

pallets/pool-system/src/impls.rs
hieronx: I don't fully remember. I think we just set up the `PriceValue` for some other purpose, and then used that here.

pallets/liquidity-pools/src/lib.rs
lemunozm: I'm not sure if it's super straightforward. Do we need to add the order-book dependency to know the market price used for foreign/pool currencies?

mustermeiszer: I agree it isn't. But we are not only touching the UTs here but also chaning the return types, naming etc. Then we could also close the todo IMO.

lemunozm: Done here: [15424a8](https://github.com/centrifuge/centrifuge-chain/pull/1890/commits/15424a8ea2d379f164ac44d750af2574d6d0af8f) @mustermeiszer

mustermeiszer (changes requested):
The `MarketRatio` type should provide a `Rate::one()` in case the currencies are equal. From my perspective we can adapt `OracleRatioProviderLocalAssetExtension` or write a new one.

Otherwise, looks good.

lemunozm:
Good catch!

I've decided to merge the behavior into `OracleRatioProviderLocalAssetExtension` to avoid having different providers for each change we'll do to the prices.

Commit: [662b8a1](https://github.com/centrifuge/centrifuge-chain/pull/1890/commits/662b8a180f5502375e07053053056e6dfadf87be)

mustermeiszer (approved):
Thanks! LGTM!

PR #1891, opened 2024-06-27
Refactor: Decoupling routers from gateway

lemunozm:
# Description

Preparing changes for https://github.com/centrifuge/centrifuge-chain/pull/1889 where the `solidity_pools::Codec` trait will no longer be used.

NOTE: It's not directly related to LPv2

## Changes
- Moved serialization from `routers` to the `gateway`. Now both, serialization and deserialization happen in the `gateway`.
- Routers only understand about `Vec<u8>` no longer for an specific message.
- Decoupled routers crate from gateway crate. Now, they do not know each other.


wischli (approved):
Clean like a fresh pair of shoes!

PR #1893, opened 2024-06-28
Add multi-router support in LP gateway.

cdamian:
# Description

The LP gateway queue logic was moved to a new pallet which is used for handling both inbound/outbound messages.

## Outbound messages

Changed processing of outbound messages by sending the actual message once, using one router, and sending message proofs using the remaining routers.

## Inbound messages

Inbound messages are sent for further processing (i.e. to the LP pallet) only when the required number of message proofs  is received.

# Checklist:


pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -213,6 +224,14 @@ pub mod pallet {
 	#[pallet::getter(fn domain_routers)]
 	pub type DomainRouters<T: Config> = StorageMap<_, Blake2_128Concat, Domain, T::Router>;
 
+	/// Storage for domain multi-routers.
+	///
+	/// This can only be set by an admin.
+	#[pallet::storage]
+	#[pallet::getter(fn domain_multi_routers)]
+	pub type DomainMultiRouters<T: Config> =
```
cdamian, line 232: Added a separate storage and extrinsic since I'm not sure how to proceed in this case - do we eventually remove the old storage after a migration or can we nuke/replace it now?

cdamian:
Re-basing on main.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: You can replace it now. If needed, the migration can go along with this PR, which I think is quite straightforward: just adding the current router, if it exists, into the vector.

mustermeiszer (commented):
Pretty straightforward change. Wanted to propose a few things if we are anyways working on this: 

* Add a dedicated `Queue` for Inbound and Outbound that is currently served by on idle, so that we can easily switch to `Tasks` in the future. 
* (NOT SURE IF GOOD IDEA) Instead of iterating over all routers we could have a `DoubleMap` and then serving the message in a queue does take care of 
  * 1. Ensuring that a message can be send over each `Router` only once

**Furthermore**
We will need to have the same for inbound logic. So a the `Gateway` must take care of a few things

* Outgoing
  * Each Message is sent only once over each existing `Router`

* Incoming
  * Each Message is only forwarded to the `LiquidityPools` pallet once some information has been verified. 
    *NOTE: It is to be determined how this will happen, but we should already take care right now that messages are "queued" until this verification has happened.*

mustermeiszer:
I think `https://github.com/paritytech/polkadot-sdk/blame/master/substrate/frame/message-queue/src/lib.rs` could be a solution, but I think the queue itself is overly complicated and in general very opaque an raw. 

`trait ProcessMessage` is using a `&[u8]` as its message type for some reason.

cdamian:
> Pretty straightforward change. Wanted to propose a few things if we are anyways working on this:
> 
> * Add a dedicated `Queue` for Inbound and Outbound that is currently served by on idle, so that we can easily switch to `Tasks` in the future.
> * (NOT SURE IF GOOD IDEA) Instead of iterating over all routers we could have a `DoubleMap` and then serving the message in a queue does take care of
>   
>   * 1. Ensuring that a message can be send over each `Router` only once
> 
> **Furthermore** We will need to have the same for inbound logic. So a the `Gateway` must take care of a few things
> 
> * Outgoing
>   
>   * Each Message is sent only once over each existing `Router`
> * Incoming
>   
>   * Each Message is only forwarded to the `LiquidityPools` pallet once some information has been verified.
>     _NOTE: It is to be determined how this will happen, but we should already take care right now that messages are "queued" until this verification has happened._

@mustermeiszer we already have a queue in place for outbound messages - do you mean that we should replace that with a queue implementation that will serve both the LP gateway and the main LP pallet?

mustermeiszer:
> we already have a queue in place for outbound messages - do you mean that we should replace that with a queue implementation that will serve both the LP gateway and the main LP pallet?

I think now it makes sense, because there will be much more processing needed. Having a general `Inbound/Outbound` queue would make sense. 

The general idea of the multi router logic is. 

Sending:
* 1. Send full message over one `Router`
* 2. Send `MessageProof`s over the other `Router`s

Receiving
*NOTE: That proofs CAN be received prior to the message itself*

* Receive `Message`/`MessageProof` 
* Check whether message can be already forwarded to the  `pallet-liquidity-pools`

I really, really strongly am against exposing the `MessageProof`s and any `Multi-Router` aggregation to the `pallet-liquidity-pools`. The pallet itself should always simply assume to process the next message wrt. to business logic. 

I would have split the the enum into an `enum LpProtocol` containing messages related to ensuring the transport and also the batching, and one `enum LpMessages` containing only the business logic related parts. For now, we could ensure that `LpMessage` is a subset of `LpProtocol` so that `pallet-liquidity-pools` only handles business parts. 

For sending batches I would also say we have an `fn batch()` in `pallet-liquidity-pools` and the gatway takes care of creating such a batch. Easiest for all purposes would be to have an `Gateway::start_pack_batch()` on the gateway and an `Gateway::end_pack_batch()`. All calls to `Gateway::submit()` that happen between those two will then be boundled into a batch by the gateway. This way we can keep the current non-batch extrinsic logic and just iterate over the `Vec<LiquidityPools::Call>` that we receive in the `fn batch()`

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -715,45 +760,69 @@ pub mod pallet {
...
+							&mut post_dispatch_info,
+							e.post_info.actual_weight,
+							read_weight,
+						);
+
+						return Err(DispatchErrorWithPostInfo {
+							post_info: post_dispatch_info,
+							error: e.error,
+						});
 					}
```
mustermeiszer, line 805: Also for that: 
* The first router must send the message
* The next routers must send a `MessageProof` of said message

pallets/liquidity-pools/src/message.rs
```diff
@@ -415,6 +419,22 @@ pub enum Message {
...
+	// DEPRECATED_Unfreeze,
+	/// Trigger a redeem request.
+	///
+	/// Note - placeholder that should be updated.
+	TriggerRedeemRequest,
+	/// A multi-directional message used for verifying
+	/// both inbound and outbound messages.
+	MessageProof {
+		proof: [u8; 32],
+	},
```
lemunozm, line 437: Sorry @cdamian I didn't know, you're feature came with new messages from LPv2. In that case, maybe it makes more sense to add those to lpv2 branch.

libs/traits/src/liquidity_pools.rs
```diff
@@ -22,6 +22,11 @@ pub trait LPEncoding: Sized {
 	fn deserialize(input: &[u8]) -> Result<Self, DispatchError>;
 }
 
+pub trait LPMessage: LPEncoding {
+	fn get_message_proof(&self) -> Option<[u8; 32]>;
+	fn to_message_proof(&self) -> Self;
+}
```
lemunozm, line 28: I think there is nothing wrong with add them directly to the LPEncoding trait, or did you found something against this?

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_idle(_now: BlockNumberFor<T>, max_weight: Weight) -> Weight {
+			Self::service_message_queue(max_weight)
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
```
lemunozm, line 113: Curious, why a new pallet? Can not this behavior be embedded in the gateway?

lemunozm (commented):
First fast look to this. Will do another one more deeply soon!

pallets/liquidity-pools/src/message.rs
lemunozm: Maybe this PR can be split into something which is idempotent over how `main` works with LPv1 (it means, not sending any proof, just the original message by one router). And a new very small PR over LPv2 sending the proof.

That way most changes are done over `main` avoiding future git conflicts

cdamian: Yea, I was afraid we might have to do this. Will look into doing a smooth transition.

libs/traits/src/liquidity_pools.rs
cdamian: It made sense to me to have a separate trait that defines a separate set of behaviours, but I'm also OK with having just one big trait for this.

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian: This was mentioned in one of the previous discussions around this topic, having the queue logic separately will make it easier for us to change it to `Tasks` in the future. Furthermore, it makes the gateway lighter and hopefully the only logic in there is the one dealing with messages and message proofs.

libs/traits/src/liquidity_pools.rs
lemunozm: I have no strong opinion, what you think is better!

pallets/liquidity-pools-gateway/queue/src/lib.rs
lemunozm: I didn't read it. Thanks! makes sense 👍🏻

pallets/liquidity-pools/src/message.rs
cdamian: OK, am currently thinking about splitting this in 3 parts, and (roughly) do the following on top of the LPv2 branch:
1 - add the new queue pallet;
2 - use the new queue pallet in the gateway;
3 - add proof processing for inbound/outbound messages;

cdamian:
This PR will eventually be closed in favor of separate PRs on top of the LPv2 branch.

mustermeiszer (commented):
Just a brief scan. Nothing to complain, sounds and looks like a solid plan.

PR #1895, opened 2024-07-01
LPv2: ForeignInvestments changes

lemunozm:
# Description

Required changes in `pallet-foreign-investments` for LPv2

Checkout the sequence diagrams before reviewing this: https://centrifuge.hackmd.io/H4v-SV0zRuCxC29kxoqOgQ

## Changes and Descriptions

- Modify `decrease` by `cancel`
  - This allows big simplifications if we know that no more increases/decreases can occur from a cancel if request until the Fulfilled cancel msg is sent.
- Remove the no longer used `collect_*` and `investment()` and `redemption()` methods from `ForeignInvestment` trait. 
- Remove the `remaining` amount from output messages.
- The correlation has been simplified. We no longer need to store both foreign and pool amounts to correlate. We know that the previous `correlation.pool_amount` always matches to `T::Investment::investment()` so we only need to store the foreign amount. The correlation is only done in just one point in the code: in `post_collect()`, so we no longer need a correlation struct to handle *any* case generically. We can just simplify for the one needed case. 
- Swapping calls and hooks are now called in `entities.rs`, leaving the `impl.rs` file much more readable.
- Unify hooks into the same new trait `ForeignInvestmentHooks`
- Remove `pallet-swaps`
  - We no longer need to apply a swap over an existent inverse swap. From now on we only need to create, increase, and cancel a swap. That means that most valuable parts from `pallet-swap` are no longer used, so we can just drop that pallet and handle the `order-book` direclty from  `foreign-investment` again. Also, not having `pallet-swaps` makes easy the deal with different ratios in opposite swaps.

**Blocked by https://github.com/centrifuge/centrifuge-chain/pull/1892**


libs/traits/src/investments.rs
```diff
@@ -268,52 +267,9 @@ pub trait ForeignInvestment<AccountId> {
...
 		investment_id: Self::InvestmentId,
 		foreign_payout_currency: Self::CurrencyId,
 	) -> Result<(), Self::Error>;
-
-	/// Returns, if possible, the currently unprocessed investment amount (in
-	/// pool currency) of who into the given investment class.
-	///
-	/// NOTE: If the investment was (partially) processed, the unprocessed
-	/// amount is only updated upon collecting.
-	fn investment(
```
lemunozm, line 305: We only used this to know how much we need to cancel, so it is no longer needed

pallets/foreign-investments/src/tests.rs
```diff
@@ -904,33 +981,95 @@ mod investment {
...
+			assert_ok!(ForeignInvestment::increase_foreign_investment(
+				&USER,
+				INVESTMENT_ID,
+				AMOUNT,
+				FOREIGN_CURR
+			));
+
+			util::fulfill_last_swap(Action::Investment, AMOUNT);
+
+			let foreign_fulfilled = Arc::new(Mutex::new(0));
```
wischli, line 1026: Curious question: It's not really necessary to do this over using `mut` right because we are not leveraging concurrency yet in UTs?

pallets/foreign-investments/src/tests.rs
```diff
@@ -1426,30 +1600,23 @@ mod redemption {
...
-						amount_currency_payout: pool_to_foreign(tranche_to_pool(
-							3 * TRANCHE_AMOUNT / 4
-						)),
-						amount_tranche_tokens_payout: 3 * TRANCHE_AMOUNT / 4,
-						amount_remaining: TRANCHE_AMOUNT / 4,
-					}
-				);
-				Ok(())
-			});
+			MockHooks::mock_fulfill_collect_redemption(
```
wischli, line 1607: Isn't this mock ineffective because we don't collect anymore?

pallets/foreign-investments/src/tests.rs
```diff
@@ -1458,12 +1625,13 @@ mod redemption {
 				Some(RedemptionInfo {
 					foreign_currency: FOREIGN_CURR,
 					swapped_amount: 0,
-					collected: CollectedAmount::default(),
+					collected_tranche_tokens: 0,
```
wischli, line 1628: I don't understand why this should be 0. First, we process 3/4 of the redemption and then collect it. So why would it be 0?

pallets/foreign-investments/src/tests.rs
```diff
@@ -1483,26 +1651,18 @@ mod redemption {
...
-					ExecutedForeignCollect {
-						currency: FOREIGN_CURR,
-						amount_currency_payout: pool_to_foreign(tranche_to_pool(TRANCHE_AMOUNT)),
-						amount_tranche_tokens_payout: TRANCHE_AMOUNT,
-						amount_remaining: 0,
-					}
-				);
-				Ok(())
-			});
+			MockHooks::mock_fulfill_collect_redemption(
```
wischli, line 1656: Same here, we don't collect after this.

wischli (commented):
This definitely improves comprehensibility of foreign investments by orders of magnitude 🎉 I have quite an amount of questions.

I will create a branch which combines both our changes in order to continue working on the ITs which need the fulfilled changes on our side.

pallets/foreign-investments/src/entities.rs
```diff
@@ -1,135 +1,65 @@
-//! Types with Config access. This module does not mutate FI storage
+//! Perform actions over ForeignInvestInfo and ForeignRedemptionInfo types
+//! - This module does not handle FI storages
+//! - This module does not call hooks
+//! - This module does not directly OrderBooks
+//! - This module does not directly OrderIdToSwapId storage
```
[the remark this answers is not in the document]
lemunozm: Thanks, re-done that part a bit!

pallets/foreign-investments/src/entities.rs
```diff
@@ -203,89 +118,96 @@ impl<T: Config> InvestmentInfo<T> {
...
-			return Ok(Some(ExecutedForeignDecreaseInvest {
-				foreign_currency: self.foreign_currency,
-				amount_decreased: sp_std::mem::take(&mut self.decrease_swapped_foreign_amount),
-				amount_remaining: self.remaining_foreign_amount(who, investment_id)?,
-			}));
-		}
+		if self.foreign_currency != pool_currency {
+			let increase_foreign = match self.order_id {
+				Some(order_id) => {
+					let increase_foreign = cancel_swap::<T>(who, swap_id, &order_id)?.into();
```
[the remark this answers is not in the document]
lemunozm: > How do we know at this point that if there exists an order

We now there is an order because we have stored the `order_id`

> it is an order from foreign to pool which we can cancel in cancel_swap?

Yes

> It could also be an order which swaps pool to foreign from a previous cancel?

No, because you can not call cancel again if there is a pending cancel in progress

pallets/foreign-investments/src/entities.rs
```diff
@@ -203,89 +118,96 @@ impl<T: Config> InvestmentInfo<T> {
...
 
-		Ok(None)
-	}
+					// When cancelling, we no longer need to correlate.
+					// The entire amount returned in the cancel msg will be the entire foreign
+					// amount in the system, so we add here the not yet tracked pending amount.
+					self.foreign_amount.ensure_add_assign(increase_foreign)?;
```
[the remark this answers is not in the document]
lemunozm: This is probably the most tricky part of the whole crate:

> When we create or update a Orderbook order foreign -> pool, we increase self.foreign_amount by the foreign amount in post_increase_swap.

Note that we only increase once the swap is fulfilled, so if the swap is fulfilled, there is nothing to cancel here. If the swap was not fulfilled, then `self.foreign_amount` was not incremented (`post_increase_swap` was never called), so we increment that amount here.

pallets/foreign-investments/src/impls.rs
```diff
@@ -29,104 +25,33 @@ impl<T: Config> ForeignInvestment<T::AccountId> for Pallet<T> {
 		foreign_amount: T::ForeignBalance,
 		foreign_currency: T::CurrencyId,
 	) -> DispatchResult {
-		let msg = ForeignInvestmentInfo::<T>::mutate(who, investment_id, |entry| {
+		ForeignInvestmentInfo::<T>::mutate_exists(who, investment_id, |entry| {
 			let info = entry.get_or_insert(InvestmentInfo::new(foreign_currency));
 			info.ensure_same_foreign(foreign_currency)?;
+			info.ensure_no_pending_cancel(investment_id)?;
```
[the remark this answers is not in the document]
lemunozm: Do you mean on the solidity side? If so, yes, they do not call increase again until they have all increased amount to 0 (waiting for the cancel msg response).

pallets/foreign-investments/src/impls.rs
```diff
@@ -344,121 +175,26 @@ impl<T: Config> StatusNotificationHook for CollectedRedemptionHook<T> {
...
-				(who.clone(), investment_id),
-				msg,
-			)?;
-		}
-
-		Ok(())
+/// Avoiding boilerplate each time the entry needs to be removed
+fn remove_entry<Entry>(condition: bool, entry: &mut Option<Entry>) -> DispatchResult {
+	if condition {
+		*entry = None;
```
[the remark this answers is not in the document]
lemunozm: Yeah, was quite annoying having that code-structure copy-pasted everywhere 😆

pallets/foreign-investments/src/tests.rs
```diff
@@ -227,19 +262,14 @@ mod investment {
 				ForeignInvestmentInfo::<Runtime>::get(&USER, INVESTMENT_ID),
 				Some(InvestmentInfo {
 					foreign_currency: FOREIGN_CURR,
-					correlation: Correlation::new(0, 0),
+					foreign_amount: 0,
```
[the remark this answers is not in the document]
lemunozm: > we call info.increase which increments the InvestmentInfo.foreign_amount by the foreign amount if pool and foreign currency mismatch.

No, `InvestmentInfo.foreign_amount` is only increased in `post_increase_swap`, in order to maintain the correlation with the real pool amount increased in Investments

pallets/foreign-investments/src/tests.rs
lemunozm: Theoretically, there are not any concurrency issues, but the `mock-builder` expects any moved reference to the closure to be `'static`, so we can not pass a `&mut` reference to mutate it inside the closure. Also, `mock-builder` is not thread-safe (although theoretically, it could be for the same tests), so the closure can not be `Sync` and `Send`. So we need to pass an `Arc` (`Rc` is not supported)

pallets/foreign-investments/src/tests.rs
lemunozm: No, because we only collect when the entire swap is swapped, and this happens in the next line.

Note that who performs the `fulfill_collect_redemption is the last partial swap from the collected amount.

pallets/foreign-investments/src/tests.rs
lemunozm: We process 3/4 of the total amount. So, the total possible collectable amount is 3/4.
We later make two swaps 1/2 and 1/4, so after both, we no longer have any pending tranche tokens to notify to the user.

Note that the pending 1/4 amount is not collected. If the user wants those funds, they need to cancel the redemption or wait until they are collected.

lemunozm:
Thanks a lot for your super deep review @wischli!!

> I will create a branch which combines both our changes in order to continue working on the ITs which need the fulfilled changes on our side.

Before doing this, if you want, to avoid extra branches, we can 
- merge this into your branch
- merge your branch (even with the failing tests) into `feat/lp-v2` and later merge this as another PR on that branch.

pallets/foreign-investments/src/entities.rs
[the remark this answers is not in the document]
lemunozm: So it never can be incremented twice

wischli:
> Thanks a lot for your super deep review @wischli!! 

Thanks for answering all my questions so quickly! IMO, this branch can be merged to the feature branch indeed.

> > I will create a branch which combines both our changes in order to continue working on the ITs which need the fulfilled changes on our side.
>
> Before doing this, if you want, to avoid extra branches, we can
> 
> * merge this into your branch
> * merge your branch (even with the failing tests) into `feat/lp-v2` and later merge this as another PR on that branch.

I would prefer this because I can imagine that fixing all Solidity ITs won't happen in the near future. Especially, if minor Solidity code changes might be required, which could be the case.

wischli (approved):
I'm fine with merging this into my branch. The few clippy complaints can be fixed by me then.

lemunozm:
CI for this is still blocked, I'm gonna merge @wischli

PR #1896, opened 2024-07-01
Altair: release v0.12.0

wischli:
# Description

Release candidate for v0.12.0 which adds OpenGov to Development and Altair runtimes

* Updates OpenGov weights for Development and Altair runtimes
* Fixes benchmark compilation issue in Altair chain spec

# Checklist:


runtime/altair/src/weights/pallet_preimage.rs
```diff
@@ -2,23 +2,23 @@
...
 // pallet
-// --chain=altair-local
+// --chain=altair
 // --steps=50
 // --repeat=20
 // --pallet=pallet_preimage
 // --extrinsic=*
 // --wasm-execution=compiled
 // --heap-pages=4096
-// --output=/tmp/runtime/altair/src/weights/pallet_preimage.rs
```
lemunozm, line 21: Are we adding `development` weights into `altair`?

Why does the output path change to `development` if the file is under Altair? 🤔

wischli: Eagle eyes! Altair benchmarks didnt work until last night so I just used the development ones because they are more or less the same and I just updated gov related benchmarks. However, I will do another benchmark execution and push true Altair weights.

lemunozm: Makes sense! Actually, I guess we could just just compute weights only for `centrifuge` and use them for all runtimes

wischli: Yes I agree. In the end, we want to ensure Centrifuge weights are feasible.

So are you fine with merging the PR without recomputing the weights for Altair as well here?

lemunozm (approved):
Yeah! I'm fine with the weights

PR #1898, opened 2024-07-05
Release v0.13.0

wischli:
# Description

* Removes migrations which are expected to have been executed
* Bumps spec version to 1300
* Updates deps to enable metadata hash check as recommended [on Element](https://matrix.to/#/!pYjTIdxhocIQxcOXdb:matrix.parity.io/$rgj5rvCJXXitBp49-8Umx6bD-VMqoanZFefY7vgiSaw?via=matrix.parity.io&via=matrix.org&via=parity.io)
* Updates srtool to latest version
* Adds `on-chain-release-build` to srtool build opts
  * Longterm we should switch to using the explicit github action: https://github.com/polkadot-fellows/runtimes/blob/bb52c327360d1098d3b3d36f4eafb40a74636e80/.github/workflows/release.yml#L59

*Pipeline enhancements*
- Limit sanity_checks to 90m to avoid consuming minutes on stuck jobs
- Modify wasm build to output srtool information
- Upload artifacts when creating a release (wasm and json summary)
- Upload json summary to GCP bucket too
- Add a manual workflow to delete extra containers from GH registry (verified working)
- Remove updating the release with the docker tag as appended comment (preferred to do manually)

# Checklist:


runtime/development/src/migrations.rs
```diff
@@ -21,7 +21,7 @@ parameter_types! {
 
 /// The migration set for Development & Demo.
 /// It includes all the migrations that have to be applied on that chain.
-pub type UpgradeDevelopment1200 = (
+pub type UpgradeDevelopment1300 = (
 	// Initialize OpenGov Technical Committee with Alice
 	runtime_common::migrations::technical_comittee::InitMigration<crate::Runtime, InitialTcMembers>,
 	runtime_common::migrations::increase_storage_version::Migration<crate::Referenda, 0, 1>,
```
wischli, line 27: NOTE: Still needed because we haven't upgraded Dev and Demo to 1200.

Cargo.toml
```diff
@@ -53,7 +53,7 @@ license = "LGPL-3.0"
 homepage = "https://centrifuge.io/"
 repository = "https://github.com/centrifuge/centrifuge-chain"
 documentation = "https://reference.centrifuge.io/centrifuge_chain/index.html"
-version = "0.12.0"
+version = "0.13.0"
```
lemunozm, line 56: Why a major version instead of 0.12.1?

lemunozm (approved):
LGTM

Cargo.toml
wischli: That's something I battled with internally myself. I thought this would be cleaner because v0.12.0 was related to enabling OpenGov on Altair.

Since this enables the metadata hash check, I think it deserves its major version. For similar releases in the future, I will take the minor version approach then since you seem to prefer that as well.

mustermeiszer (dismissed):
That was fast! LGTM!

runtime/altair/Cargo.toml
```diff
@@ -157,6 +157,7 @@ std = [
   "serde/std",
   "log/std",
   # Substrate related
+  "substrate-wasm-builder",
```
lemunozm, line 160: I think it should be `"substrate-wasm-builder/std"` to enable also std, if not, only the basic crate is added

runtime/altair/Cargo.toml
```diff
@@ -459,7 +460,7 @@ try-runtime = [
 # generate the metadata hash and then a second time with the
 # `RUNTIME_METADATA_HASH` environment variable set for the `CheckMetadataHash`
 # extension.
-metadata-hash = ["substrate-wasm-builder/metadata-hash"]
+metadata-hash = ["substrate-wasm-builder?/metadata-hash"]
```
lemunozm, line 463: Why the `?`?. is not your intention to always set the `metadata-hash` of `substrate-wasm-builder` each time you set `metadata-hash`

wischli:
@lemunozm Seems like WASM building succeeds now. Can you please to the nth re-approval so we can get this merged finally? Thanks for the patience! And thanks @gpmayorga for putting so much energy in improving our WASM CI.

.github/workflows/build-docker.yml
```diff
@@ -111,14 +112,14 @@ jobs:
...
-            **Docker tags (${{ env.NOW }}):**
-               ${{ steps.meta.outputs.tags }}
+      # - name: Update GitHub release
+      #   if: github.event_name == 'release' && github.event.action == 'released' && matrix.target == 'release'
+      #   uses: softprops/action-gh-release@v1
+      #   with:
+      #     append_body: true
+      #     body: |
+      #       **Docker tags (${{ env.NOW }}):**
+      #          ${{ steps.meta.outputs.tags }}
```
lemunozm, line 122: Tech-debt, should be maintained in the repo or removed? @gpmayorga

lemunozm (approved):
Let's merge this! Thanks for your work on this 🚀

wischli:
@gpmayorga We need your approval as CI code owner

.github/workflows/build-docker.yml
wischli: Good catch. IMO, we can remove this!

gpmayorga (approved):

PR #1905, opened 2024-07-10
LP: Add Unitary Tests for update_member()

lemunozm:
# Description

Add UTs to `update_member()` and remove unitary testing from integration tests related to it.

@wischli I've done over `main` because it should not conflict with the current lpv2 work. Maybe just a minor import.


wischli (approved):
Thanks for the changes! LGTM but let's not merge this before #1898 please

PR #1907, opened 2024-07-11
Experimental: Strongly typed types of time

lemunozm:
# Description

The idea of this PR is to have `Millis` and `Seconds` as different types for the compiler. This helps to never mix by error different duration units and improve the readability of some parts of the codebase.

To achieve this, we need:
- A wrapper type over `u64` that allows us different instances to be able to compile to not mix types (check `num_wrapper.rs` file).
- Some explicit conversions from seconds to millis that will change the value accordingly
- Implement everything from `BaseArithmetic` to be able to use this types as normal `u64` (i.e. give to `pallet_timestamp`'s `Moment` the new `Millis` type.

### Blockers
It seems like `parity_scale_codec` doesn't allow to implement `HasCompact` trait required by `BasicArithmetic`:
  - https://github.com/paritytech/parity-scale-codec/issues/613
  - https://github.com/paritytech/polkadot-sdk/issues/5004


libs/utils/src/num_wrapper.rs
```diff
@@ -0,0 +1,477 @@
...
+///
+/// ```
+/// # use crate::num_wrapper::NumWrapper;
+///
+/// struct Id1;
+/// struct Id2;
+///
+/// type FooU64 = NumWrapper<u64, Id1>;
+/// type BarU64 = NumWrapper<u64, Id2>;
+/// ```
```
lemunozm, line 32: FooU64 and BarU64 are different types for the compiler, and you can implement different traits to them.

libs/utils/src/time.rs
```diff
@@ -0,0 +1,41 @@
...
+			pub const fn into_millis(self) -> Millis<$t> {
+				Millis::from(self.get().saturating_mul(1000))
+			}
+		}
+	};
+}
+
+into_millis!(Seconds<u32>);
+into_millis!(Seconds<u64>);
+into_millis!(Seconds<u128>);
```
lemunozm, line 41: We can also add types for days/months/years... interesting for interests and cashflows computation

PR #1909, opened 2024-07-12
WIP: Liquidity Pools v2

lemunozm:
# Description

LP-V2 base branch where we've merged the individual PRs. For more information, check the individual PRs merged in this branch


lemunozm:
Modify this PR to ready to review because we want to be sure this branch has always CI in green

lemunozm:
@wischli are the current failing tests already fixed in https://github.com/centrifuge/centrifuge-chain/pull/1915 or should I fix them (I think I should fix two and ignore the rest)?

wischli:
> @wischli are the current failing tests already fixed in https://github.com/centrifuge/centrifuge-chain/pull/1915 or should I fix them (I think I should fix two and ignore the rest)?

They are not part of #1915 yet but I can definitely include that! Apologies that I didn't check those.

lemunozm:
No worries at all! It's more. I think the two tests from `cases::liquidity_pools::foreign_investments::mismatching_currencies` have been probably broken by me after the foreign changes.

I will try to get this green, and then in your PR, fix whatever remains

lemunozm:
I think the conflicts are due to https://github.com/centrifuge/centrifuge-chain/pull/1926 will and I'll fix on Monday when I'm back

lemunozm:
Rebased! CI should pass 🤞🏻

pallets/liquidity-pools/src/tests.rs
```diff
@@ -545,6 +556,27 @@ mod add_tranche {
...
 			})
 		}
+
+		#[test]
+		fn with_no_hook_address() {
+			System::externalities().execute_with(|| {
+				Permissions::mock_has(|_, _, _| true);
+				Pools::mock_pool_exists(|_| true);
+				Pools::mock_tranche_exists(|_, _| true);
+				AssetRegistry::mock_metadata(|_| Some(util::default_metadata()));
```
lemunozm, line 566: NIT (I think it comes from a merge where does not applied the config_mocks() call)
```suggestion
				config_mocks();
```

runtime/integration-tests/src/cases/lp/investments.rs
```diff
@@ -165,7 +155,7 @@ mod with_pool_currency {
 	use super::{utils, *};
 	use crate::cases::lp::utils as lp_utils;
 
-	#[test_runtimes(all)]
+	#[test_runtimes([development])]
```
lemunozm, line 158: @wischli I am curious why these are changed only for development; they should be changed to `all`, right?

runtime/integration-tests/src/cases/lp/utils.rs
```diff
@@ -0,0 +1,310 @@
...
+
+		(left[31] == 1u8, unsigned64)
+	}
+}
+
+impl<T: Input> Decoder<H160> for T {
+	fn decode(&self) -> H160 {
+		assert_eq!(self.input().len(), 32usize);
+
+		H160::from(to_fixed_array(&self.input()[12..]))
```
lemunozm, line 218: Our `convert_evm_address()` starts with the H160 and the last 12 are the not used. But here it is the opposite

runtime/integration-tests/src/cases/lp/setup_evm.rs
```diff
@@ -0,0 +1,578 @@
...
+	/* NOTE: This rely is NOT needed as the LocalRouter is not permissioned
+	evm.call(
+		Keyring::Alice,
+		Default::default(),
+		names::ADAPTER,
+		"rely",
+		Some(&[Token::Address(evm.deployed(names::ROOT).address())]),
+	)
+	.unwrap(); */
+
```
lemunozm, line 27: Should be then removed?

runtime/integration-tests/src/cases/lp/setup_lp.rs
```diff
@@ -0,0 +1,974 @@
...
+	// 		&evm.view(
+	// 			Keyring::Alice,
+	// 			names::POOL_A_T_1,
+	// 			"restrictionManager",
+	// 			None,
+	// 		)
+	// 		.unwrap()
+	// 		.value,
+	// 	),
+	// );
```
lemunozm, line 300: TODO

runtime/integration-tests/src/cases/lp/setup_lp.rs
```diff
@@ -0,0 +1,974 @@
...
+	// 		&evm.view(
+	// 			Keyring::Alice,
+	// 			names::POOL_B_T_1,
+	// 			"restrictionManager",
+	// 			None,
+	// 		)
+	// 		.unwrap()
+	// 		.value,
+	// 	),
+	// );
```
lemunozm, line 358: TODO

runtime/integration-tests/src/cases/lp/setup_lp.rs
```diff
@@ -0,0 +1,974 @@
...
+	// 		&evm.view(
+	// 			Keyring::Alice,
+	// 			names::POOL_B_T_2,
+	// 			"restrictionManager",
+	// 			None,
+	// 		)
+	// 		.unwrap()
+	// 		.value,
+	// 	),
+	// );
```
lemunozm, line 416: TODO

lemunozm (commented):
Done my self-review. Some comments below

runtime/integration-tests/src/cases/lp/investments.rs
wischli: Yeah that was done to simplify CLI during debugging. We can switch back all to `all`!

runtime/integration-tests/src/cases/lp/setup_evm.rs
wischli: Let's please keep this. Right now, we can only replicate the Solidity setup manually. This comment should signal that it is not necessary to perform this action at this point but that the author was aware of it.

runtime/integration-tests/src/cases/lp/setup_lp.rs
wischli: Should I include fixing these as part of https://github.com/centrifuge/centrifuge-chain/pull/1933?

lemunozm: What you think is better

runtime/integration-tests/src/cases/lp/setup_evm.rs
lemunozm: Ok, I see! Could we then add this message as a comment to it?

runtime/integration-tests/src/cases/lp/investments.rs
lemunozm: Let's see how time CI will take after that 😆

lemunozm:
Regarding centrifuge-chain state:
- None of the involved pallets whose storages can be affected has data at this moment:
  - `pallet-foreign-investments`
  - `pallet-swaps`
  - `pallet-liquidity-pools-gateway`
  
Anyway, I think we should bump up their version numbers. Not sure if `pallet-swaps` need to be purged even if there is no data on it, in order to remove any possible associated value (as the version number, which is 0 🤔)

lemunozm:
Demo is clean also except for `pallet-swaps` which are some opened swaps that must be closed I guess before the RU

runtime/integration-tests/src/cases/lp/setup_evm.rs
wischli: Doesn't the current line suffice? 

> NOTE: This rely is NOT needed as the LocalRouter is not permissioned

runtime/integration-tests/src/cases/lp/utils.rs
wischli: Good catch. Running tests now over lunch break to see whether this breaks any ITs.

lemunozm:
BTW @wischli I see 6 tests ignored. Did you say 5 in the sync?

runtime/integration-tests/src/cases/lp/setup_evm.rs
lemunozm: Thanks! I didn't catch the meaning at first without being involve on it

runtime/integration-tests/src/cases/lp/utils.rs
lemunozm: To be honest, I'm not sure why in some Decode Methods, it is expected to be 32 bytes, but some bytes are skipped, such as when decoding the `Balance`.

wischli: Yeah it is quite magic indeed and I also don't fully understand it. Regarding this code path, it seems to be correct. At least it has been like this since its genesis when Frederik initialized LP integration tests ([blame](https://github.com/centrifuge/centrifuge-chain/blame/e4e048a574e393061f0169e88d2d66b7e1970760/runtime/integration-tests/src/generic/cases/lp/mod.rs#L244)).

It is used for decoding addresses returned by the byte vector `evm.view(...).unwrap().value`: https://github.com/centrifuge/centrifuge-chain/blob/020cf59f18209768afbb8a844867b976f66ab65d/runtime/integration-tests/src/cases/lp/pool_management.rs#L90

Under the hood, it is the result of https://github.com/centrifuge/centrifuge-chain/blob/e95db3922e81e371571cb899e3c80bfcac255f4c/runtime/integration-tests/src/envs/evm_env.rs#L181

So it appears to me that `pallet_evm` prefixes 20-byte addresses with zeroes. After changing the above line 
```diff
- 		H160::from(to_fixed_array(&self.input()[12..]))
+ 		H160::from(to_fixed_array(&self.input()[..20]))
```

many tests fail with weird reasons. So let's keep as is.

pallets/liquidity-pools/src/tests.rs
wischli: In `config_mocks()`, we mock a hook address for the `AddTranche` message. This test requires this hook to not have been set. If we execute `config_mocks`, then we need to revert https://github.com/centrifuge/centrifuge-chain/pull/1909/files/5a63633955ba13282a2448dd2a15262e5173c9ba#diff-57577a9a8bc154a3e3b81af63e11d2b863c6716120411b1480117c8c44ee02d5R446-R452

runtime/integration-tests/src/cases/lp/utils.rs
lemunozm: Ok, it seems like all call answers from Ethereum return 32 bytes. And in those bytes there are coded different types. So always need to be 32 bytes and the standard of this will dictate how a H160 is represented in those 32 bytes. Thanks for the clarification!

pallets/liquidity-pools/src/tests.rs
lemunozm: The idea is to overwrite the mock make it fail, so:
```rust
config_mocks();
Gateway::mock_get(|_| None);
```
make the tests work as expected.

The idea is not to copy-past all initialization for all these tests and not create an order in how the mocks are defined. I did that refactor, but I think this test was added later and was skipped by me in the merge.

wischli (commented):
Looks super clean except for the leftovers from me which @lemunozm already pointed out. Working on a cleanup PR which will also fix the last failing test `transfer_tranche_tokens_domain_to_local_to_domain`.

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+	/// Storage for messages that will be processed during the `on_idle` hook.
+	#[pallet::storage]
+	#[pallet::getter(fn message_queue)]
+	pub type MessageQueue<T: Config> = StorageMap<_, Blake2_128Concat, T::MessageNonce, T::Message>;
+
+	/// Storage for messages that failed during processing.
+	#[pallet::storage]
+	#[pallet::getter(fn failed_message_queue)]
+	pub type FailedMessageQueue<T: Config> =
+		StorageMap<_, Blake2_128Concat, T::MessageNonce, (T::Message, DispatchError)>;
```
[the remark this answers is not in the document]
wischli: cc @cdamian

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
```diff
@@ -0,0 +1,64 @@
...
+		T: Config,
+		T::AccountId: EncodeLike<<T as frame_system::Config>::AccountId>,
+)]
+mod benchmarks {
+	use super::*;
+
+	#[benchmark]
+	fn process_message() -> Result<(), BenchmarkError> {
+		let caller: T::AccountId = account("acc_0", 0, 0);
+		let message = T::Message::default();
```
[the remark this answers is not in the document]
cdamian: That will be tricky since the message itself is irrelevant given that it gets passed to a mock.

pallets/liquidity-pools-gateway/queue/src/lib.rs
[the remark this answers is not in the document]
cdamian: Yes, the 3rd part of the multi-router work has logic that involves hashes, see a potential solution [here](https://github.com/centrifuge/centrifuge-chain/pull/1893/files#diff-74702c928dc1b5f82b0ba7573573ef25bc263370ba55e341eeeadc7dec1ef255R228).

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use crate::{
+	config::Runtime,
+	env::{Blocks, Env},
+	envs::fudge_env::{FudgeEnv, FudgeSupport},
+	utils::{currency::cfg, genesis, genesis::Genesis},
+};
+
+#[test_runtimes(all)]
+fn submit_and_process<T: Runtime + FudgeSupport>() {
```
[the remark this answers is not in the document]
cdamian: For now this suffices as the LP gateway doesn't actually process anything. In the 2nd part that I'm tackling right now I will extend this whole suite.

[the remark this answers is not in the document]
wischli: Awesome!

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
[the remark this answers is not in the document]
lemunozm: > it gets passed to a mock

Note that when the benchmark is executed for the real runtimes, the mock is no longer used.

@wischli we already talked about adding a PoV for the MaxEncodedLen to the resulting weights

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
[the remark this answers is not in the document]
lemunozm: From my understanding, testing the biggest happy path in IT is enough. Anything else can be tested with UTs in a much simpler way.

pallets/liquidity-pools/src/tests.rs
lemunozm: I changed it in a commit after for your work on this, so you indeed followed the status-quo!!

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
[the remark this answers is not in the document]
cdamian: > Note that when the benchmark is executed for the real runtimes, the mock is no longer used.

Oh, that explains a lot then. Where does this magical switch from mock runtime to actual runtime happen?

Yes, we should definitely add that there before merging this.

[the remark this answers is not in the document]
lemunozm: The magic happens here: https://github.com/centrifuge/centrifuge-chain/blob/main/runtime/centrifuge/src/lib.rs#L2775

That macro calls the benchmarks of those pallets using the real runtime instead of the mock. That's why sometimes the benchmarking is so difficult to achieve. You should be able to create a benchmark that works fine for any runtime configuration.

Right now, the maximum message is no more than 256 bytes. So we're "safe", maybe just adding to the extrinsic is enough:

#[pallet::weight(T::WeightInfo::process_message().saturating_add(Weights::from_parts(0, T::Message::max_encoded_len())))]

I would not go with a huge change because the batch process will need some special treatment regarding this. I'll do it in the batch PR once I rebase over this.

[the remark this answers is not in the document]
cdamian: Thank you for the pointer, I'll expand that macro and take a look.

As for the `process_message` weight, with the changes that I'm currently working on, this will be a bit easier since that extrinsic will now submit the message to the queue. It will be then processed `on_idle` and return a `DispatchResultWithPostInfo` that can have more granular weights, as per [this](https://github.com/centrifuge/centrifuge-chain/pull/1937/files#diff-74702c928dc1b5f82b0ba7573573ef25bc263370ba55e341eeeadc7dec1ef255R567).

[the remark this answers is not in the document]
lemunozm: If you're able to return directly the weight there, then I would say we're fine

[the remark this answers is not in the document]
lemunozm: BTW (NIT), if you're only interested in the `weight`, maybe it's interesting to return a tuple `(DispatchResult, Weight)` instead. That way, you can get rid of the unused `pay_fees`, and you have the `weight` for both successful processing and failures. I was experimenting with that in the batch PR; maybe it's useful in that part too.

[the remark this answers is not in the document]
cdamian: Awesome idea, it should make our lives easier.

lemunozm:
Finally, this is green! I think we can just rebase and merge

wischli (approved):
Approving our great teamwork here! Once CI is green, we can merge this!

PR #1910, opened 2024-07-12
Lpv2: ignore failing tests

lemunozm:
# Description

We want to mark them as `ignore` not to forget them later when they work again


wischli (approved):
Thanks!

PR #1912, opened 2024-07-15
LP: Unitary testing for add_currency

lemunozm:
# Description

## Changes and Descriptions

- Remove related tests from IT in `cases/liquidity_pools.rs`
- Add associated UTs
- Simplify wrapped token struct as tuple.


libs/types/src/tokens.rs
```diff
@@ -395,44 +395,6 @@ impl CrossChainTransferability {
...
-	Eq,
-	Debug,
-	Encode,
-	Decode,
-	TypeInfo,
-	MaxEncodedLen,
-	Serialize,
-	Deserialize,
-)]
-pub enum LiquidityPoolsWrappedToken {
```
lemunozm, line 418: It seems it can be removed. 

Its only current purpose is to return the inner values in the private method of `pallet-liquidity-pools`: `try_get_wrapped_token()`. From my PoV, it adds a new layer of understanding for the reading; doing that as a tuple makes things easier. 

I can revert if you have some point against it.

lemunozm:
This is ready to be merged if you agree in the above clean up

mustermeiszer (commented):
This keeps the test in the lp folder, right?

lemunozm:
Yeah, `lp` tests are not modified

wischli (approved):
LGTM! I agree with you on the unnecessarily complexity of the wrapped liquidity token. It has been a longer discussion in a past but I dont see why we can just work with the tuple.

mustermeiszer (approved):

PR #1913, opened 2024-07-15
ci: disable checks for drafts

wischli:
# Description

Goal: Reduce CI overload

* Disables CI checks on draft PRs
* Removes `altair` runtime from benchmark checks

### DIFF
--- a/.github/workflows/sanity-checks.yml
+++ b/.github/workflows/sanity-checks.yml
@@ -9,6 +9,7 @@ permissions:
   contents: read
 jobs:
   test-n-lint:
+    if: ${{ !github.event.pull_request.draft }}
     timeout-minutes: 90
     name: ${{ matrix.target }}
     runs-on: ubuntu-latest-8-cores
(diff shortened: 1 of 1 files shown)


.github/workflows/sanity-checks.yml
```diff
@@ -41,12 +42,13 @@ jobs:
...
 
   benchmark-check:
+    if: ${{ !github.event.pull_request.draft }}
     # timeout-minutes: 90 # <- Not needed as ubunt-latest is free of charge
     name: bench-check-${{ matrix.runtime }}
     runs-on: ubuntu-latest #-4-cores
     strategy:
       matrix:
-        runtime: [ altair, centrifuge ]
+        runtime: [ centrifuge ]
```
wischli, line 51: It's really not necessary to also check Altair here. If it works for Centrifuge, we have ensured validity. In the worst case, something might be off in the Altair runtime config but we can always just use the Centrifuge weights anyways.

hieronx (approved):

.github/workflows/sanity-checks.yml
lemunozm: Good point regarding the weights! Maybe even make sense to move the weights to `runtime-common` 🤔

PR #1918, opened 2024-07-17
LPv2: Fix foreign-investments ITs

lemunozm:
# Description

Fix failing FI-related tests found in `feat/lp-v2`

### DIFF
--- a/runtime/integration-tests/src/cases/liquidity_pools.rs
+++ b/runtime/integration-tests/src/cases/liquidity_pools.rs
@@ -2837,7 +2837,7 @@ mod foreign_investments {
 								investor: investor.clone().into(),
 								currency: general_currency_index::<T>(foreign_currency),
 								currency_payout: invest_amount_foreign_denominated,
-								fulfilled_invest_amount: 0,
+								fulfilled_invest_amount: invest_amount_foreign_denominated,
 							},
 						}
 						.into()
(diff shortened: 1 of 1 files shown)


runtime/integration-tests/src/cases/liquidity_pools.rs
```diff
@@ -2949,10 +2942,7 @@ mod foreign_investments {
 								investor: investor.clone().into(),
 								currency: general_currency_index::<T>(foreign_currency),
 								currency_payout: invest_amount_foreign_denominated / 2,
-								// TODO(@luis): Should be `invest_amount_pool_denominated / 4` with
-								// deltas because we have processed 50% of the investment with
-								// conversion rate 50%
```
lemunozm, line 2954: The conversion rate used here is only for pool <-> tranches. Not for pool <-> foreign. Actually we do not have ITs with prices other than 1:1 for pool <-> foreign currencies. (We should)

lemunozm (commented):
@wischli, maybe the lp related tests that are also failing in `feat/lp-v2` are failing due the lp repo dependency used is not the correct one. Could you confirm what's the commit/branch of that repo that we should use to get these tests pass?

If it's working now fine, I'll add the `ignore` tag

wischli (approved):

wischli:
> @wischli, maybe the lp related tests that are also failing in `feat/lp-v2` are failing due the lp repo dependency used is not the correct one. Could you confirm what's the commit/branch of that repo that we should use to get these tests pass?

We have been using the intended submodule head, so this is not the issue unfortunately. I have consistently updated the submodule to stay in sync with the LP repo.

lemunozm:
My fear is that I could overwrite wrongly the submodule hash in some merge/rebase over `main`

PR #1920, opened 2024-07-17
LPv2: Batch Message serialization

lemunozm:
# Description
This PR is related to the batch message support. The batch logic itself will be in a next PR.

### Serialization Notes
We need a custom serialization/deserialization due to the spetial requirements of how batches needs to be serialized:
 - REQ 1: Have the size of each submessage before the submessage as `u16`.
   - Technically, this is not needed for us because we know the structure of the message to deserialize.
   - Our workaround to get the message size is to serialize with our data format inside of the the serialization process.
 - REQ 2: Not having a number indicating the size of the batch
   - Technically, this is required for us. We need to know the size of the batch and how many submessages we need to deserialize. Without that information means that we enter the deserialization logic with no idea of how to finish.
   - The workaround solution is to finish when we find an error (an "unexpected EOF") deserializing the next length of a submessage.

### Changes
- [x] Custom Batch serialization/deserialization
- [x] Type system protection to avoid nested batches

### Next PR
- Add `Gateway::start_pack_batch()` and `Gateway::end_pack_batch()`
- Gateway required changes (new storage to keep at least the boolean state of above)
- LPEncoding utilities to allow the Gateway to compose messages into batches without knowing the concrete `Message` type.


lemunozm:
@mustermeiszer @wischli this is ready to review/merge. The logic itself goes in a nex PR

pallets/liquidity-pools/src/message.rs
```diff
@@ -79,7 +200,7 @@ impl TryInto<Domain> for SerializableDomain {
 	TypeInfo,
 	MaxEncodedLen,
 )]
-pub enum Message {
+pub enum Message<BatchContent = BatchMessages> {
```
lemunozm, line 203: This allows us to create different versions of Message where we can stop the recursive process of `max_encoded_len()`. By default, we want the `Message` to be able to batch another messages. Except when the message already belongs to a batch.

pallets/liquidity-pools/src/message.rs
```diff
@@ -30,6 +34,8 @@ pub const TOKEN_NAME_SIZE: usize = 128;
 // The fixed size for the array representing a tranche token symbol
 pub const TOKEN_SYMBOL_SIZE: usize = 32;
 
+pub const MAX_BATCH_MESSAGES: u32 = 256;
```
mustermeiszer, line 37: 256 messages or bytes? I think for messages, that is too much ^^

pallets/liquidity-pools/src/message.rs
```diff
@@ -66,6 +72,121 @@ impl TryInto<Domain> for SerializableDomain {
...
+						.try_add(msg)
+						.map_err(|e| A::Error::custom::<&'static str>(e.into()))?;
+				}
+
+				Ok(batch)
+			}
+		}
+
+		let limit = MAX_BATCH_MESSAGES as usize * 2; // Lengths and messages
+		deserializer.deserialize_tuple(limit, MsgVisitor)
```
mustermeiszer, line 147: Shouldn't we for deserialze also take care of handling the `BoundedVec` or will the internal scale-codec handle that?

pallets/liquidity-pools/src/message.rs
```diff
@@ -406,6 +528,28 @@ pub enum Message {
...
+	}
+
+	/// Decompose the message into a list of messages
+	pub fn unpack(&self) -> Vec<Self> {
+		match self {
+			Message::Batch(content) => content.clone().into_iter().collect(),
+			message => vec![message.clone()],
+		}
+	}
+}
```
mustermeiszer, line 551: I like!

pallets/liquidity-pools/src/message.rs
```diff
@@ -406,6 +528,28 @@ pub enum Message {
...
+				Message::Batch(content)
+			}
+			this => Message::Batch(BatchMessages::try_from(vec![this.clone(), other])?),
+		})
+	}
+
+	/// Decompose the message into a list of messages
+	pub fn unpack(&self) -> Vec<Self> {
+		match self {
+			Message::Batch(content) => content.clone().into_iter().collect(),
```
mustermeiszer, line 547: I think a `BoundedVec` does not need to be collected. Can use `Into` or something

pallets/liquidity-pools/src/message.rs
mustermeiszer: Not sure I am following ^^

pallets/liquidity-pools/src/message.rs
lemunozm: 16 seems more reasonable?

pallets/liquidity-pools/src/message.rs
lemunozm: The `BatchMessages` knows how to create itself by increasing the internal `BoundedVec` but we can not use the substrate `BoundedVec` deserialization because any variable list of elements requires known the size upfront but we can not.

Basically, we need to redo this process due to the solidity requirement of not sending the batch size upfront.

pallets/liquidity-pools/src/message.rs
lemunozm: Good point!

pallets/liquidity-pools/src/message.rs
lemunozm: The problem: The worst case for `MaxEncodedLen` autoimplementation is that the `Message` contains a `Batch` that contains `Messages` that contain `Batch` that contains `Messages`...

Parameterizing here with the batch content we are able to create a `Message<()>` to compute correctly the `MaxEncodedLen` for cases where the message is a submessage of a batch, and avoid the infinite recursion

pallets/liquidity-pools/src/message.rs
lemunozm: Ah, I remember why I did it so. I need to iterate to remove the `NoBatchMessage` wrapper of each message. The `into_iter()` does not belong to `BoundedVec`, it belongs to `BatchMessages` which automatically unwraps the `NoBatchmessage` into normal `Message`.

mustermeiszer (approved):
LGTM! Great PR

PR #1923, opened 2024-07-22
LPv2: Batch message logic

lemunozm:
# Description

### Responsibilities
- `pallet-liquidity-pools-gateway` knows to batch/unbatch but doesn't know what a Batch message is. To handle this, the `LPMessage` trait now allows pack/unpack methods.
- `pallet-liquidity-pools` will never send or receive a batch. The gateway is the responsible for the batching process.

### Details
Batching is handled by two extrinsic: 
  - `Gateway::start_pack_messages(sender, destination)`:  When calling this, any submitted messages from LP, that match the `sender` & `destination` will be packed. _NOTE: We need to index also by destination domain because the destination information is not inside the message. So, we can never bundle two messages with two different destinations._
  - `Gateway::end_pack_messages(sender, destination)`: When calling this, the batch is finally submitted. If this is not called, no message is sent. If the batch reaches the limit size, it will dispatch an error.
  
Later, the user can call `pallet_utility::batch(calls)` and any message sent between `start_pack` and `end_pack` will be bundled in the batch.

**This PR depends on:** https://github.com/centrifuge/centrifuge-chain/pull/1920

### Changes
- Added `Gateway::start_pack_messages()` and `Gateway::end_pack_messages()` to allow creating batches.
- Renamed `process_msg()` to `receive_message()`. Added a new parameter to be able to measure the weight correctly.
- Renamed `process_message()` to `send_message()`. Measure the weight there based on the serialized message len.
- Removed `try_range` for header message deserialization and use a `BufferReader`.
- Removed `defensive_weights.rs` file and use more adjusted weights that take into account batch messages.
- Added a `weight()` method to `OutboundQueue` trait that allows to compute correct weights from liquidity_pools taking into account batch sizes.


lemunozm:
@mustermeiszer I had a lot of issues implementing the batch(calls) method regarding filtering the accounts (that some calls can be root), different destinations, nested calls to batch, weights…

I think this hybrid solution is simpler if you’re happy with the fulfilling requirements (I think everything we’ve talked about is fulfilled here). It gives enough control and never leaves the batch decision to the `on_idle.` Weights can be easily measured and handled out of `on_idle`

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -801,21 +853,12 @@ pub mod pallet {
...
-			);
-
-			Self::deposit_event(Event::OutboundMessageSubmitted {
-				sender: T::Sender::get(),
-				domain: destination,
-				message,
-			});
+			match PackedMessage::<T>::get((&sender, &destination)) {
+				Some(packed) => {
+					PackedMessage::<T>::insert((sender, destination), packed.pack(message)?)
```
lemunozm, line 858: @mustermeiszer Here the `pack()` checks that you don't exceed the batch limit, so you can never create more than 16 messages in a batch. Creating the 17th will fail. `on_idle` just will send the bytes.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -517,7 +521,11 @@ pub mod pallet {
 				}
 			};
 
-			T::InboundQueue::submit(domain_address, incoming_msg)
+			for msg in incoming_msg.unpack() {
+				T::InboundQueue::submit(domain_address.clone(), msg)?;
+			}
```
mustermeiszer, line 526: Weight of this function must take into account the maximum batch size possible. Defined by the deserialization implemenation.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -600,6 +608,30 @@ pub mod pallet {
...
 		}
+
+		#[pallet::call_index(8)]
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1).ref_time())]
+		pub fn start_pack_messages(origin: OriginFor<T>, destination: Domain) -> DispatchResult {
+			let sender = ensure_signed(origin)?;
+
+			if PackedMessage::<T>::get((&sender, &destination)).is_none() {
+				PackedMessage::<T>::insert((sender, destination), T::Message::empty());
+			}
```
mustermeiszer, line 619: `try_mutate` could be cleaner.

mustermeiszer: I would also error out if the start has already happened once.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -600,6 +608,30 @@ pub mod pallet {
...
+		}
+
+		#[pallet::call_index(9)]
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1).ref_time())] //TODO: benchmark me
+		pub fn end_pack_messages(origin: OriginFor<T>, destination: Domain) -> DispatchResult {
+			let sender = ensure_signed(origin)?;
+
+			if let Some(msg) = PackedMessage::<T>::take((&sender, &destination)) {
+				Self::queue_message(destination, msg)?;
+			}
```
mustermeiszer, line 631: error out none.

pallets/liquidity-pools-gateway/src/lib.rs
mustermeiszer: Cool! We still need to check 
* what is the weight of a 16 message batch?
* do we need to adapt `process_message` weight defaults

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Yeah, I was not sure about no-op and error out. But seeing that you also opted by erroring out, I will do so 👍🏻

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -420,6 +432,8 @@ pub mod pallet {
 		/// Process an incoming message.
 		#[pallet::weight(T::WeightInfo::process_msg())]
 		#[pallet::call_index(5)]
+		// TODO: can we modify the name to receive_message() ?
+		// TODO: benchmark me again with max batch limit message
 		pub fn process_msg(
```
lemunozm, line 437: Currently when receiving a message, you call `process_msg()` and for sending it `process_message()` which is quite confusing. Do you agree to renaming those to `receive_message()` and `send_message()`?
@wischli @mustermeiszer @cdamian

Note that this would imply a breaking change due the extrinsic name is modified

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -718,60 +758,54 @@ pub mod pallet {
...
-				)),
-				Err(e) => Err(DispatchErrorWithPostInfo {
-					post_info: post_dispatch_info_fn(e.post_info.actual_weight, read_weight),
-					error: e.error,
-				}),
-			}
+			let serialized = message.serialize();
+
+			let message_weight =
+				read_weight.saturating_add(Weight::from_parts(0, serialized.len() as u64));
```
lemunozm, line 775: Message PoV size evaluated here

pallets/liquidity-pools-gateway/src/lib.rs
mustermeiszer: No strong opinion. No external integrations, as the extrinsic anyways can only be called by a special origin. So fine for me.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: This is really difficult to measure because, right now, it's not measured. It's using hardcoded weights. The real weight of this does not depend on how many bytes we read from the message, it mostly depends of what messages are in the batch that will dispatch certain actions. i.e., if this message is an increase it will call to foreign investments and would imply swaps having a weight increase.

I'm not sure what kind of solution we want to follow here.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -801,23 +777,20 @@ pub mod pallet {
...
+			}
 
 			Ok(())
 		}
+
+		fn weight() -> Weight {
+			T::DbWeight::get()
+				.reads_writes(3, 4)
+				.saturating_add(Weight::from_parts(0, T::Message::max_encoded_len() as u64))
+		}
```
lemunozm, line 794: Here we compute the weight needed to enqueue a message the could be a batch. This size is different to the used when sending the message, which use the serialized sized.

pallets/liquidity-pools/src/lib.rs
```diff
@@ -951,6 +946,12 @@ pub mod pallet {
 			let domain_address = T::DomainAccountToDomainAddress::convert(domain_account);
 			T::DomainAddressToAccountId::convert(domain_address)
 		}
+
+		fn outbound_weight(reads: u64, writes: u64) -> Weight {
+			Weight::from_parts(1_000_000, 256) //defensive base weights
+				.saturating_add(T::DbWeight::get().reads_writes(reads, writes))
+				.saturating_add(T::OutboundQueue::weight())
+		}
```
lemunozm, line 954: This takes into account the correct `OutboundQueue::submit()` weights even when the message is a batch. We no longer need `defensive_weights.rs`

lemunozm (commented):
Some comments from the tricky parts:

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: I've got one which multiplies the current weight used by the number of messages in the batch

lemunozm:
Maybe we should wait to merge @Cosmin PR first if they have more priority to avoid wasting time with conflict (that will be a lot 😄). Anyways this is ready to be reviewed in the meantime

lemunozm:
Closed in favor of https://github.com/centrifuge/centrifuge-chain/pull/1949

PR #1926, opened 2024-07-24
Refactor: Transform `TrancheCurrency` type in a tuple

lemunozm:
# Description

Remove the TrancheCurrency type & trait, which represent exact tuple behavior, and use a tuple directly.

## Why this change?
- Less knowledge layers
- Easier to create mocks in pallets that need this type. Previously, either you imported the `TrancheCurrency` type (which is a centrifuge domain type), or you created your own by adding boilerplate in the tests. But your own implementation doesn't work well with the `CurrencyId` type used everywhere, which needs a conversion.
- This change is a step to clean all domain dependencies from our pallets and unitary tests, and get close to the ideal here: https://github.com/centrifuge/centrifuge-chain/issues/1380

### This change comes in two parts:

- Part 1 (this PR): 
  - Refactor to use a tuple instead of `TrancheCurrency` type which is fully removed. 
  - The tuple still implements the `TrancheCurrency` trait which is still used by 2 pallets: `pallet-pool-system` and `pallet-foreign-investments`
- Part 2 (new PR over `lpv2` branch): 
  - We clean `pallet-foreign-investments` from the usage of `TrancheCurrency` in this branch to avoid git conflicts later. 
  - Finishing some refactors required to `pallet-pool-system` to remove unused generics from this change.
  - Remove the `TrancheCurrency` trait.

### DIFF
--- a/libs/mocks/src/pools.rs
+++ b/libs/mocks/src/pools.rs
@@ -25,7 +25,6 @@ pub mod pallet {
 		type Balance;
 		type BalanceRatio;
 		type CurrencyId;
-		type TrancheCurrency;
 	}
 
 	#[pallet::pallet]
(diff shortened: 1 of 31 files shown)


wischli (dismissed):
We should have done this much earlier as it improves readability while being much simpler. Gone are the days in which we need to add a bunch of restrictions on `TrancheCurrency` and in which it's sometimes a bit fuzzy, whether the trait or the struct was meant 🎉

I also really appreciate the PR description and splitting this refactor into two PRs.

lemunozm:
I see I'm not the one who has fought with these types/traits before. Hopefully, it will become more straightforward to deal with now. Thanks!

lemunozm:
Can I have a reapproval here? 🙏🏻

mustermeiszer (approved):

PR #1927, opened 2024-07-24
LP: Unitary testing for inbound messages

lemunozm:
# Description

Unitary tests for inbound messages

NOTE. The decrease and collection-related inbound messages are not tested because they are deprecated in LPV2

### DIFF
--- a/pallets/liquidity-pools/src/tests/inbound.rs
+++ b/pallets/liquidity-pools/src/tests/inbound.rs
@@ -1,0 +1,791 @@
+use cfg_traits::liquidity_pools::InboundQueue;
+use cfg_types::permissions::{PermissionScope, PoolRole, Role};
+use frame_support::{assert_noop, assert_ok, traits::fungibles::Mutate as _};
+
+use crate::{mock::*, Error, Message};
+
+mod handle_transfer {
+	use super::*;
+
+	#[test]
+	fn success() {
+		System::externalities().execute_with(|| {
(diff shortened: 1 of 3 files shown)


pallets/liquidity-pools/src/tests/inbound.rs
```diff
@@ -0,0 +1,792 @@
...
+				assert_eq!(foreign_currency, CURRENCY_ID);
+
+				// Side effects of this call
+				ForeignInvestment::mock_redemption(|_, _| Ok(0));
+				Tokens::mint_into(TRANCHE_CURRENCY, &ALICE, AMOUNT).unwrap();
+				Ok(())
+			},
+		);
+		Gateway::mock_submit(|sender, destination, msg| {
+			assert_eq!(sender, TreasuryAccount::get());
```
lemunozm, line 688: Double-check, should it not be the local representation of the receiver? similar to what is used in `TransferTrancheTokens` message: https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/liquidity-pools/src/inbound.rs#L95

wischli: I think that works as well but it is not necessary because assets are not supposed to be transferred in `n-hops`, e.g. from one EVM domain to another with Centrifuge as middleman. However, we support those hops for tranche token transfers.

wischli (approved):
I love the increasing UT code coverage! Thanks so much for all your work @lemunozm 🤘

PR #1928, opened 2024-07-24
feat: add domain hook storage

wischli:
# Description

* Replaces hardcoded `AddTrancheHookAddress` with `DomainHookAddress` storage because it is different for every domain
  * @mustermeiszer recommended adding the storage to the gateway instead of the LP pallet
* Extends gateway tests by also checking LP Admin account origin

# Checklist:

### DIFF
--- a/libs/mocks/src/outbound_queue.rs
+++ b/libs/mocks/src/outbound_queue.rs
@@ -1,8 +1,9 @@
 #[frame_support::pallet(dev_mode)]
 pub mod pallet {
-	use cfg_traits::liquidity_pools::{DomainHook, OutboundQueue};
+	use cfg_traits::liquidity_pools::OutboundQueue;
 	use frame_support::pallet_prelude::*;
 	use mock_builder::{execute_call, register_call};
+	use orml_traits::GetByKey;
 
 	#[pallet::config]
 	pub trait Config: frame_system::Config {
@@ -28,8 +29,8 @@ pub mod pallet {
 			register_call!(move |(a, b, c)| f(a, b, c));
 		}
 
-		pub fn mock_get_address(f: impl Fn(T::Destination) -> Option<T::Sender> + 'static) {
-			register_call!(move |a| f(a));
+		pub fn mock_get(f: impl Fn(&T::Destination) -> Option<T::Sender> + 'static) {
+			register_call!(f);
 		}
 	}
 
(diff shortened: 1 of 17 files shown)


pallets/liquidity-pools/src/lib.rs
```diff
@@ -250,12 +250,10 @@ pub mod pallet {
...
-		/// The type for processing outgoing messages.
-		type OutboundQueue: OutboundQueue<
-			Sender = Self::AccountId,
-			Message = Message,
-			Destination = Domain,
-		>;
+		/// The type for processing outgoing messages and retrieving the domain
+		/// hook address.
+		type OutboundQueue: OutboundQueue<Sender = Self::AccountId, Message = Message, Destination = Domain>
+			+ DomainHook<Domain = Domain, AccountId = Self::AccountId>;
```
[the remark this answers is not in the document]
wischli: Ah I was only looking in Polkadot SDK which does not have this trait but just `TypedGet` which doesn't support key args. Will implement your prop!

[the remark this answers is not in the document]
wischli: [fa74c6f](https://github.com/centrifuge/centrifuge-chain/pull/1928/commits/fa74c6fc4457119204b5ccca467bec18846a4c01)

libs/mocks/src/outbound_queue.rs
```diff
@@ -3,11 +3,12 @@ pub mod pallet {
...
 	use frame_support::pallet_prelude::*;
 	use mock_builder::{execute_call, register_call};
+	use orml_traits::GetByKey;
 
 	#[pallet::config]
 	pub trait Config: frame_system::Config {
-		type Sender;
-		type Destination;
+		type Sender: Parameter;
+		type Destination: Parameter;
```
lemunozm, line 11: Why are these bounds needed in the mock?

lemunozm (commented):
Thanks for the extra coverage!

libs/mocks/src/outbound_queue.rs
wischli: The compiler required both types to be encodable and decodable. I started with single traits but then just went with `Parameter` to have peace.

lemunozm: Mmm, it's weird that the mock mandates those 🤔

libs/mocks/src/outbound_queue.rs
```diff
@@ -17,12 +18,20 @@ pub mod pallet {
 	#[pallet::storage]
 	type CallIds<T: Config> = StorageMap<_, _, String, mock_builder::CallId>;
 
+	#[pallet::storage]
+	type DomainAddressHook<T: Config> =
+		StorageMap<_, _, <T as Config>::Destination, <T as Config>::Sender>;
```
lemunozm, line 23: Ahh it's because of that. I think this can be removed from the mock

lemunozm: Copy-paste probably (?)

wischli: Great catch! Yes, copy-paste and a blind eye.

mustermeiszer (commented):
Why do we go with T::AccountId vs [u8,20]?

PR #1930, opened 2024-07-29
Feat/lp v2 gateway queue

cdamian:
# Description

Moved queue logic from LP gateway to a new pallet.

# Checklist:


lemunozm:
Is this ready to review @cdamian?

cdamian:
> Is this ready to review @cdamian?

@lemunozm yes, I don't think I want to add anything else here for now.

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
```diff
@@ -0,0 +1,65 @@
...
+use cfg_traits::liquidity_pools::test_util::Message as LPTestMessage;
+use frame_benchmarking::{account, impl_benchmark_test_suite, v2::*};
+use frame_system::RawOrigin;
+use parity_scale_codec::EncodeLike;
+
+use super::*;
+
+#[benchmarks(
+	where
+		T: Config<Message = LPTestMessage>,
```
lemunozm, line 23: This will probably not pass for the real runtimes. My suggestion is to add a `Default` bound for Message only when benchmarking it. (which would be an `Invalid` one)

Option2. Add a worst-case method under `runtime-benchmarks` in `LPEncoding`, which allows creation of the worst message possible.

Option3. Add a parameter to the benchmarks that allows to pass the `Message`, so each message is benchmarked for itself. This will be the most accurate/clean, but not sure if this is possible to pass non-integer parameters to benchmarking functions 🤔

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+		/// The message type.
+		type Message: Clone + Debug + PartialEq + MaxEncodedLen + TypeInfo + FullCodec;
+
+		/// Type used for message identification.
+		type MessageNonce: Parameter
```
lemunozm, line 48: Super NIT. Using `BaseArithmetic` I think you can get rid of most of the bounds below

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+					Ok(())
+				}
+				Err(e) => {
+					FailedMessageQueue::<T>::insert(nonce, (message.clone(), e.error));
+
+					Self::deposit_event(Event::<T>::MessageExecutionFailure {
+						nonce,
+						message,
+						error: e.error,
+					});
```
lemunozm, line 145: I think the `deposit_event` calls can factorized along with `T::MessageProcessor::process`to avoid duplication each time you call `process()`. Maybe an auxiliary function?

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+					Ok(())
+				}
+				Err(e) => {
+					Self::deposit_event(Event::<T>::MessageExecutionFailure {
+						nonce,
+						message,
+						error: e.error,
+					});
+
+					Ok(())
```
lemunozm, line 179: Why not make failing the extrinsic here?

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+				Err(e) => {
+					FailedMessageQueue::<T>::insert(nonce, (message.clone(), e.error));
+
+					Self::deposit_event(Event::<T>::MessageExecutionFailure {
+						nonce,
+						message,
+						error: e.error,
+					});
+
+					Ok(())
```
lemunozm, line 147: Same here

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+							// 1 write for the event
+							// 1 write for the failed message
+							// 1 write for the message removal
+							.saturating_add(T::DbWeight::get().reads_writes(1, 3))
+					}
+				};
+
+				weight_used = weight_used.saturating_add(weight);
+
+				if weight_used.all_gte(max_weight) {
```
lemunozm, line 232: First time I've seen this! Whats the difference with `>=`?

runtime/altair/src/lib.rs
```diff
@@ -1831,6 +1832,14 @@ impl pallet_liquidity_pools_gateway::Config for Runtime {
 	type WeightInfo = ();
 }
 
+impl pallet_liquidity_pools_gateway_queue::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Message = pallet_liquidity_pools::Message;
+	type MessageNonce = LPGatewayQueueMessageNonce;
+	type MessageProcessor = LiquidityPoolsGateway;
+	type WeightInfo = ();
```
lemunozm, line 1840: TODO We should add the generated weights here

runtime/centrifuge/src/lib.rs
```diff
@@ -1927,6 +1928,14 @@ impl pallet_liquidity_pools_gateway::Config for Runtime {
 	type WeightInfo = ();
 }
 
+impl pallet_liquidity_pools_gateway_queue::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Message = pallet_liquidity_pools::Message;
+	type MessageNonce = LPGatewayQueueMessageNonce;
+	type MessageProcessor = LiquidityPoolsGateway;
+	type WeightInfo = ();
```
lemunozm, line 1936: Same here

runtime/development/src/lib.rs
```diff
@@ -1931,6 +1932,14 @@ impl pallet_liquidity_pools_gateway::Config for Runtime {
 	type WeightInfo = ();
 }
 
+impl pallet_liquidity_pools_gateway_queue::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Message = pallet_liquidity_pools::Message;
+	type MessageNonce = LPGatewayQueueMessageNonce;
+	type MessageProcessor = LiquidityPoolsGateway;
+	type WeightInfo = ();
```
lemunozm, line 1940: Same here

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::envs::fudge_env::{FudgeEnv, FudgeSupport};
+use crate::utils::currency::cfg;
+use crate::utils::genesis;
+use crate::utils::genesis::Genesis;
+use cfg_traits::liquidity_pools::MessageQueue as MessageQueueT;
+use frame_support::assert_ok;
+use sp_runtime::traits::One;
+
+#[test_runtimes(all)]
+fn submit_and_process<T: Runtime + FudgeSupport>() {
```
lemunozm, line 12: Just to the point! Super simple 💯

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+							.saturating_add(T::DbWeight::get().reads_writes(1, 2))
+					}
+					Err(e) => {
+						Self::deposit_event(Event::MessageExecutionFailure {
+							nonce,
+							message: message.clone(),
+							error: e.error,
+						});
+
+						FailedMessageQueue::<T>::insert(nonce, (message, e.error));
```
lemunozm, line 215: I think here it makes a lot of sense to enqueue the failing messages because we're inside the `on_idle()`, and we can not respond to the errors. But in the process_*_message() extrinsic, I think we can just fail if the message is not processed. I think there is no need to enqueue messages there, because the user can acts accordingly. Actually, they would want a fast feedback of their processing message. WDYT?

lemunozm (commented):
Really love how clean and organize is this!!

Some questions below, but it has a ver good look!

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian:
```
error[E0369]: binary operation `>=` cannot be applied to type `frame_support::weights::Weight
```

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian: The idea of enqueueing the failed message even in extrinsics is that it takes them out of the normal queue that gets processed on idle. 

For example:
- a message gets added to the queue
- the extrinsic to process this manually is called
- the execution fails due to some router misconfiguration (message is not removed from `MessageQueue`)
- `on_idle` gets called, causing the message to be re-executed, consuming weight but still failing.

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian: Returning error on these extrinsics will revert any storage changes that are done during processing and we want to avoid that.

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian: I'll look into make this whole thing a bit more DRY, thank you!

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
cdamian: I would go with option 2 since that makes the most sense for benchmarks IMO.

cdamian: > so each message is benchmarked for itself.
Would this be possible? More specifically, does the benchmark test suite execute for each variant of an enum?

pallets/liquidity-pools-gateway/queue/src/lib.rs
lemunozm: Ohh, I though they implemented `Ord` on it. Thanks!

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
lemunozm: The benchmark doesn't know the `Message` itself. But if it receive the Message as a parameter it would compute the effect of that `Message` for that specific call

lemunozm: But not sure if it would work. Option 2 seems good

cdamian: Oh, but wait. I forgot about - https://github.com/centrifuge/centrifuge-chain/blob/66e3252e13e3c632142c0af22d2acfc3e68515ab/pallets/liquidity-pools-gateway/queue/src/mock.rs#L59

Isn't it easier if we just return some actual weight here that is then used by the benchmarks?

lemunozm: That's other option. Just the weight you think is higher than that call. We do that in liquidity-pools related pallets

cdamian: Done, added a default weight ref time that's also used in the pallet instead of default. PTAL

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian: Done, cheers!

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian: Done.

runtime/altair/src/lib.rs
cdamian: Done.

runtime/development/src/lib.rs
cdamian: Done.

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+			match Self::process_message_and_deposit_event(nonce, message.clone()) {
+				Ok(p) => {
+					FailedMessageQueue::<T>::remove(nonce);
+
+					Ok(p)
+				}
+				Err(e) => Ok(e.post_info),
+			}
+		}
+	}
```
lemunozm, line 184: So, for example, if we return an error instead of `Ok(e.post_info)` what's the difference in the storage? If I read this part well, nothing has changed. Then, why not just error out and warn the user that it was wrong instead of telling them that "extrinsics was succesful" ?

pallets/liquidity-pools-gateway/queue/src/lib.rs
lemunozm: It is now super readable!! Thanks, great cleaning improvement 🎉

pallets/liquidity-pools-gateway/queue/src/lib.rs
lemunozm: I thought more bounds were tied to `BaseArithmetic`, almost anything has changed 😆

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+
+	impl<T: Config> Pallet<T> {
+		fn process_message_and_deposit_event(
+			nonce: T::MessageNonce,
+			message: T::Message,
+		) -> DispatchResultWithPostInfo {
```
lemunozm, line 190: Just one consideration about `DispatchResultWithPostInfo`. If possible, it's better not to use it as a way of measure the extrinsic. It should only be used for cases where you can optimize the extrinsic. It means only when the consumed weight is less than the expected weight. The reason is that the user (and the node) wants to know the fee cost of extrinsic costs upfront, to know if they have enough funds or if the extrinsic fit into the block.

In which cases, getting the weight from `T::MessageProcessor::process()` is reducing the weight from the original benchmarked weight?

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
lemunozm: Ok, just one thing, the issue is still there because the benchmark is tied to `LPTestMessage`

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian: In this case, for this extrinsic, we won't have the `MessageExecutionFailure` event.

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian: > In which cases, getting the weight from T::MessageProcessor::process() is reducing the weight from the original benchmarked weight?

We would have to implement the `MessageProcessor` trait in the LP gateway and figure out a way to properly test that then.

But coming back to your initial point - `process_message_and_deposit_event` should return `DispatchResultWithPostInfo` since it's used in the hook. I would drop it from the extrinsics though.

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
cdamian: Been trying things out for a while now and I'm still a bit confused...

Option 2 doesn't really work since `LPEncoding` is not a trait used as a constraint for the `Message` type of this pallet, as per [these constraints](https://github.com/centrifuge/centrifuge-chain/blob/8e5171185b78f3b1cfb894f0b756430e43790303/pallets/liquidity-pools-gateway/queue/src/lib.rs#L53).

> Ok, just one thing, the issue is still there because the benchmark is tied to LPTestMessage

This will remain true if we use the mocked runtime anyway. The message type that we use [here](https://github.com/centrifuge/centrifuge-chain/blob/8e5171185b78f3b1cfb894f0b756430e43790303/pallets/liquidity-pools-gateway/queue/src/mock.rs#L57) is the `LPTestMessage`.

pallets/liquidity-pools-gateway/queue/src/lib.rs
lemunozm: Ok, so you want to record that event in those cases

lemunozm: As a retry history

cdamian: Yes, it's basically the exact same behavior we have in the gateway, and given that these are more like convenience methods, I was thinking that we can live with them.

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
cdamian: Updated this btw, took me a while to figure out what was happening when running tests and then the check benchmark scripts.

libs/traits/src/liquidity_pools.rs
```diff
@@ -30,6 +30,10 @@ pub mod test_util {
 	use super::*;
 
 	#[derive(Debug, Eq, PartialEq, Clone, Encode, Decode, TypeInfo, MaxEncodedLen)]
+	#[cfg_attr(
+		any(test, feature = "std", feature = "runtime-benchmarks"),
+		derive(Default)
+	)]
```
lemunozm, line 36: NIT. Because this type is never used for production, it's ok to add `Default` normally

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
```diff
@@ -0,0 +1,64 @@
...
+		T: Config,
+		T::AccountId: EncodeLike<<T as frame_system::Config>::AccountId>,
+)]
+mod benchmarks {
+	use super::*;
+
+	#[benchmark]
+	fn process_message() -> Result<(), BenchmarkError> {
+		let caller: T::AccountId = account("acc_0", 0, 0);
+		let message = T::Message::default();
```
lemunozm, line 31: I am a bit worried about the PoV with batch messages, but I think we can go with it now 👍🏻

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -866,3 +872,11 @@ impl<T: Config> GetByKey<Domain, Option<[u8; 20]>> for Pallet<T> {
...
 	}
 }
+
+impl<T: Config> MessageProcessor for Pallet<T> {
+	type Message = T::Message;
+
+	fn process(_: Self::Message) -> DispatchResultWithPostInfo {
+		Ok(PostDispatchInfo::default())
+	}
+}
```
lemunozm, line 882: Why this is implemented as it is?

pallets/liquidity-pools/src/message.rs
```diff
@@ -201,7 +201,9 @@ impl BatchMessages {
 	TypeInfo,
 	MaxEncodedLen,
 )]
+#[cfg_attr(feature = "runtime-benchmarks", derive(Default))]
 pub enum Message<BatchContent = BatchMessages> {
+	#[cfg_attr(feature = "runtime-benchmarks", default)]
```
lemunozm, line 206: I think it's ok to add a `default` tag without the burden of `runtime-benchmarks`. Dealing with features always end up with issues in the future. It must be used only if there is a security issue or a high increase of code in the wasm. Any of those is an issue here

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+	pub trait Config: frame_system::Config {
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+		#[cfg(not(feature = "runtime-benchmarks"))]
+		/// The message type.
+		type Message: Clone + Debug + PartialEq + MaxEncodedLen + TypeInfo + FullCodec;
+
+		#[cfg(feature = "runtime-benchmarks")]
+		/// The message type.
+		type Message: Clone + Debug + PartialEq + MaxEncodedLen + TypeInfo + FullCodec + Default;
```
lemunozm, line 58: And we can simplify this

lemunozm (commented):
Just a last thing about `Default`, but everything else looks great! 👏🏻

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Actual implementation will be done in the following PR. I can drop this if it's in the way.

pallets/liquidity-pools/src/message.rs
cdamian: OK, definitely gonna make our lives easier if we do that. On it.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Also, it's for integration tests. Let's keep it actually and I will change it when implementing the queue in the gateway in the next step.

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
cdamian: Maybe in the next steps we can have something like a `MaxEncodedLenSelf` trait or similar for cases like these.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Ok! 👍🏻

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
lemunozm: Yeah, something like:

```rust
#[pallet::weight(T::WeightInfo::process_message().saturating_add(Weights::from_parts(0, T::Message::max_encoded_len())))]
```

lemunozm (approved):
Everything looks super good! thanks for tackling all my comments/suggestions/nits 🙌🏻

pallets/liquidity-pools-gateway/queue/src/lib.rs
lemunozm: > these are more like convenience methods

Makes more sense having that in mind!

PR #1934, opened 2024-07-30
LPv2: Bump-up foreign investment. Fix failing investment ITs 

lemunozm:
# Description

### Changes 
- Versioning
  - Bump-up version for `pallet-foreign-investment` 
  - Finally, I've not bump-up the `pallet-liquidity-pools-gateway` because their storage does not change in a breaking way. Modifying the configuration of the pallet should no imply modifying the version number of a pallet.
- Fix an issue of updating an investment with zero amount when it already has zero amount.
  - `invest_cancel_full_before_swap` passing
  - `invest_cancel_full_after_swap_partially_inter_epoch_close` passing
  - `transfer_tranche_tokens_domain_to_local_to_domain` passing
- Getting 1 min of less computing for each test case after reducing the amount of configured investors


wischli (approved):
Unfortunately, `transfer_tranche_tokens_domain_to_local_to_domain` still fails but since Jeroen has confirmed this worked in practice, we can reconvert the tast back to `ignore` and finish up the LP branch

runtime/integration-tests/src/utils/accounts.rs
```diff
@@ -224,7 +224,7 @@ pub fn default_accounts() -> Vec<Keyring> {
 
 /// Returns a Vector of default investor accounts
 pub fn default_investors() -> Vec<Keyring> {
-	(0..=50).map(Keyring::TrancheInvestor).collect()
+	(1..=2).map(Keyring::TrancheInvestor).collect()
```
lemunozm, line 227: This reduces the time for each test case to more than a minute on my computer (around 70 seconds less per test case). I think we can enable again the `all` tag for these tests.

runtime/integration-tests/src/cases/lp/transfers.rs
```diff
@@ -271,7 +272,9 @@ fn transfer_tranche_tokens_domain_to_local_to_domain<T: Runtime>() {
 				Token::FixedBytes(pool_a_tranche_1_id::<T>().into()),
 				Token::Uint(DOMAIN_EVM.into()),
 				Token::Uint(EVM_DOMAIN_CHAIN_ID.into()),
-				Token::Address(Keyring::TrancheInvestor(2).into()),
+				Token::FixedBytes(
+					AccountId::from(as_h160_32bytes(Keyring::TrancheInvestor(2))).to_raw_vec(),
```
wischli, line 276: Damn, you were faster! Just came to the same conclusion after some debugging. Well done!

runtime/integration-tests/src/cases/lp/transfers.rs
```diff
@@ -271,7 +272,9 @@ fn transfer_tranche_tokens_domain_to_local_to_domain<T: Runtime>() {
...
+				),
```
lemunozm, line 277: This makes the test pass. We need an account coded in some way that later when obtaining the local representation the extra padding does not change. I think the current account conversions are quite confusing, and we should need to simplify this into the `DomainAddress`: [ref](https://kflabs.slack.com/archives/C05TBFRBL15/p1721802319814229)

wischli: Absolutely! I am wondering why this test has worked in the past .

lemunozm: Previously a `Token::Address` was passed as a parameter instead, which seems to be processed differently

runtime/integration-tests/src/cases/lp/transfers.rs
lemunozm: Happy to reach the same solution, double confident about this!

runtime/common/src/evm/mod.rs
```diff
@@ -112,19 +112,19 @@ impl<T: pallet_aura::Config<AuthorityId = AuraId>> FindAuthor<H160> for FindAuth
...
+//  2. Build with `forge-build`
+//  3. Go to `./out/PassthroughAdapter.sol` and copy-paste the
+//     `deployedBytecode` here.
+//  4. Run tests and update mismatching hashes.
+//  5. On Development chain, you might also have to update the
+//     `evm.accountCodes` storage via raw writing.
+//
+// Blake256 hash of the deployed passthrough router contract code as
+// Encoded::encode(Vec<Code>):
+// `0x283d01c648e109952e3120e8928a19614c5c694477c780920ac29a748f96babf`
```
lemunozm, line 127: The `<` character makes docs fail in they are computed as docs instead of normal comments. 

Modified this to be a coder comment instead of a user docs.

wischli: Ah sorry for merging back that error. This branch had an outdated version though.

lemunozm: Probably I think just pushed something in between your work. Now should pass 🤞🏻

runtime/integration-tests/src/cases/liquidity_pools.rs
```diff
@@ -156,9 +156,7 @@ mod utils {
 							token_name: BoundedVec::<
 								u8,
 								<T as pallet_pool_system::Config>::StringLimit,
-							>::try_from(
-								"A highly advanced tranche".as_bytes().to_vec()
-							)
+							>::try_from("A highly advanced tranche".as_bytes().to_vec())
```
lemunozm: I think this breaks `cargo fmt`?

wischli: Yeah, already fixed. I don't know why but my IDE Rustrover keeps introducing this fmt error and I can only commit the fixed version manually without the IDE.... Small indie company JetBrains..

lemunozm: Could it be done regarding the cargo version? 

For me, what's works as CI is `cargo +nightly fmt` which is the action I've configured in my IDE

PR #1937, opened 2024-07-31
Feat/lp v2 use gateway queue

cdamian:
# Description

Use the new Gateway Queue pallet in LP and LP gateway.

## Changes and Descriptions
- Use `pallet-liquidity-pools-gateway-queue` for queueing inbound and outbound message in the LP gateway.
- Add new handler traits for inbound and outbound messages that are implemented by LP and LP gateway respectively.

# Checklist:

- [ ] Address the benchmark weight suggestion [here](https://github.com/centrifuge/centrifuge-chain/pull/1909#discussion_r1699614979)
- [x] Skip the LP gateway queue in as many integration tests as possible.
- [x] Replace the fudge runtime with the normal runtime in the integration tests, where possible.


lemunozm (commented):
Everything looks pretty good! I think gateway is much more organized now and love how the gateway interacts with the queue.

For reference to other reviewers:
- `MessageQueue` trait is used by the gateway and implemented by the queue to enqueue both inbound and outbound messages.
- `MesageProcessor` trait is used by the queue and implemented by the gateway to unqueue both inbound and outbound messages
- `OutboundMessageHandler` liquidity-pools sending a outbound message to the gateway
- `InboundMessageHandler` the gateway sending inbound messages to liquidity-pools

libs/types/src/gateway.rs
```diff
@@ -0,0 +1,17 @@
+use frame_support::pallet_prelude::{Decode, Encode, MaxEncodedLen, TypeInfo};
+
+use crate::domain_address::{Domain, DomainAddress};
+
+/// Message type used by the LP gateway.
+#[derive(Debug, Encode, Decode, Clone, Eq, MaxEncodedLen, PartialEq, TypeInfo)]
+pub enum GatewayMessage<AccountId, LPMessage> {
```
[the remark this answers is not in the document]
lemunozm: I think it's here because in the first draft, this was used by `liquidity-pools` IIRC

pallets/liquidity-pools-gateway/src/mock.rs
```diff
@@ -32,6 +35,7 @@ frame_support::construct_runtime!(
 impl frame_system::Config for Runtime {
 	type AccountId = AccountId32;
 	type Block = frame_system::mocking::MockBlock<Runtime>;
+	type DbWeight = RocksDbWeight;
```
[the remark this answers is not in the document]
cdamian: Nope, the `DefaultConfig` is using `()` for this.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -539,88 +493,12 @@ pub mod pallet {
...
-					});
+			let gateway_message = GatewayMessage::<T::AccountId, T::Message>::Inbound {
+				domain_address,
+				message: incoming_msg,
+			};
 
-					Ok(())
-				}
-			}
+			T::MessageQueue::submit(gateway_message)
```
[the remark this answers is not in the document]
cdamian: It's mostly adhering to the requirement of using the queue for inbound and outbound. The proofs will be checked when processing messages in queue, initial implementation [here](https://github.com/centrifuge/centrifuge-chain/pull/1893/files#diff-74702c928dc1b5f82b0ba7573573ef25bc263370ba55e341eeeadc7dec1ef255R557).

libs/types/src/gateway.rs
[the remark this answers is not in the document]
cdamian: Yes, it's only used by the gateway and the queue in the runtime definition (gonna add that in a bit). Should I move it over to the LP gateway folder or keep it here?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -685,128 +563,85 @@ pub mod pallet {
...
 
-		/// Retrieves the router stored for the provided domain and sends the
-		/// message, calculating and returning the required weight for these
-		/// operations in the `DispatchResultWithPostInfo`.
-		fn process_message(
-			domain: Domain,
+		/// Retrieves the router stored for the provided domain, sends the
+		/// message using the router, and calculates and returns the required
+		/// weight for these operations in the `DispatchResultWithPostInfo`.
+		fn process_outbound_message(
```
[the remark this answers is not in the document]
cdamian: We can also merge that first and not have it linger for too long and rebase it too often. Up to you.

pallets/liquidity-pools-gateway/src/mock.rs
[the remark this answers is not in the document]
lemunozm: And why not use `()`?

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: No worries, I would need to redo the weights for the batch over your changes, and I would need to rethink those

libs/types/src/gateway.rs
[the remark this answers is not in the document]
lemunozm: I would move it to the gateway, yeah!

pallets/liquidity-pools-gateway/src/mock.rs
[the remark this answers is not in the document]
cdamian: Couldn't retrieve read weights during tests by calling:
```
<Runtime as frame_system::Config>::DbWeight::get().reads(1)
```

```
error[E0282]: type annotations needed
    --> pallets/liquidity-pools-gateway/src/tests.rs:1122:7
     |
1122 |                         <Runtime as frame_system::Config>::DbWeight::get().reads(1),
     |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `T` declared on the trait `Get`
```

Which is weird, am I missing something?

libs/types/src/gateway.rs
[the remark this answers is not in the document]
cdamian: OK, will move. I wasn't sure about when to add stuff to libs and when to add to pallets. Now it's a bit clearer, cheers!

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -244,31 +208,34 @@ pub mod pallet {
...
+	//TODO(cdamian): Migration?
+	// /// Storage for failed outbound messages that can be manually re-triggered.
+	// #[pallet::storage]
+	// #[pallet::getter(fn failed_outbound_messages)]
+	// pub type FailedOutboundMessages<T: Config> = StorageMap<
+	// 	_,
+	// 	Blake2_128Concat,
+	// 	T::OutboundMessageNonce,
+	// 	(Domain, T::AccountId, T::Message, DispatchError),
+	// >;
```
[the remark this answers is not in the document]
cdamian: Hmm, depends on how the RU update happens. I would imagine that it would happen after a block is produced, thus after an `on_idle` hook is called.

pallets/liquidity-pools-gateway/src/mock.rs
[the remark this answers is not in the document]
lemunozm: I see... make sense then!

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
```diff
@@ -1,41 +1,128 @@
...
-
-	let expected_event = env.parachain_state_mut(|| {
-		let message = pallet_liquidity_pools::Message::AddPool { pool_id: 1 };
-		let nonce = <T as pallet_liquidity_pools_gateway_queue::Config>::MessageNonce::one();
-
-		assert_ok!(
-			<pallet_liquidity_pools_gateway_queue::Pallet<T> as MessageQueueT>::submit(
-				message.clone()
-			)
+mod inbound {
```
cdamian, line 20: The outbound part is already tested in most LP test but I'm thinking about adding some simple tests here as well. WDYT?

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
```diff
@@ -1,41 +1,128 @@
...
+
+		env.pass(Blocks::UntilEvent {
+			event: expected_event.into(),
+			limit: 3,
+		});
+	}
+
+	#[test_runtimes([development])]
+	fn failure<T: Runtime + FudgeSupport>() {
+		let mut env = FudgeEnv::<T>::from_parachain_storage(Genesis::default().storage());
```
lemunozm, line 96: Do we need `fudge` for this test?

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
```diff
@@ -1,41 +1,128 @@
...
+			let message = GatewayMessage::Inbound {
+				domain_address: DEFAULT_DOMAIN_ADDRESS_MOONBEAM,
+				message: Message::DepositRequest {
+					pool_id,
+					tranche_id: default_tranche_id::<T>(pool_id),
+					investor: investor.clone().into(),
+					currency: general_currency_index::<T>(currency_id),
+					amount,
+				},
+			};
```
lemunozm, line 74: - Could we use a simple `Message::Invalid` and avoid all this setup?

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
```diff
@@ -1,41 +1,128 @@
...
-			)
+mod inbound {
+	use super::*;
+
+	/// This test is basically `increase_deposit_request` but instead of
+	/// handling the message directly via the LP pallet, we submit it via
+	/// the LP Gateway Queue and confirm that it's processed accordingly.
+	#[test_runtimes([development])]
+	fn success<T: Runtime + FudgeSupport>() {
+		let mut env = FudgeEnv::<T>::from_parachain_storage(
```
lemunozm, line 28: I think we need here fudge to generate a correct `MessageExecutionSuccess` by the router, right?

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
lemunozm: Added below some comments regarding this

lemunozm: I think it makes sense, at least one, to be sure the happy path is working correctly 👍🏻

lemunozm: Indeed, I think it makes sense to have the test here checking the end-to-end with the router configuration. That way, we can test all `liquidity_pools.rs` tests avoiding the last event check and then remove all fudge usage from `liquidity_pools.rs` tests, because this test already ensures to be working correctly

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
cdamian: Most likely not, gonna double-check that.

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
cdamian: That will trigger a failure but it will mean that the message did reach the LP pallet. I can drop the success/failure tests and just have 2 tests one for inbound and one for outbound, and we confirm that the message is processed one way or another by LP / LP gateway.

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
cdamian: I'll give it a try without.

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
lemunozm: I'm curious, why does it trigger a failure?

cdamian: Because we'll be reaching [this](https://github.com/centrifuge/centrifuge-chain/blob/916c1e3a5182fb7194d1d9e6bc8c4e4f1ff8e42c/pallets/liquidity-pools/src/lib.rs#L1032).

lemunozm: Ah ok, in the case of an inbound message, yeah. But as you said that's the expected behavior, so the tests have passed 👍🏻

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I meant over a human period of time. Basically we need to know if we need to maintain the message state because the users are using it at that moment or we can just assume that they will not be using it when the RU is enacted.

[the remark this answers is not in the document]
lemunozm: Asumming there're no storage or just purge it

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
cdamian: Yes, I guess that'll do. We are testing to confirm that the message reaches the destination. Will add something similar for outbound and I guess that should suffice.

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
```diff
@@ -1,41 +1,128 @@
...
+	#[test_runtimes([development])]
+	fn failure<T: Runtime + FudgeSupport>() {
+		let mut env = FudgeEnv::<T>::from_parachain_storage(Genesis::default().storage());
+
+		let expected_event = env.parachain_state_mut(|| {
+			let nonce = <T as pallet_liquidity_pools_gateway_queue::Config>::MessageNonce::one();
+
+			let message = GatewayMessage::Inbound {
+				domain_address: DEFAULT_DOMAIN_ADDRESS_MOONBEAM,
+				message: Message::TransferAssets {
```
lemunozm, line 103: Even this can also be an `Invalid` right?

cdamian: Yes, I'm gonna drop this tho, as soon as I figure out why no events are triggered when using the normal runtime.

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
cdamian: OK, I updated most integration tests around LP and dropped fudge where possible.

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
cdamian: Done.

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
cdamian: We have 2 tests now for LP gateway queue - 1 for inbound, 1 for outbound, and both confirm that the message is reaching the correct destination - LP for inbound, LP Gateway for outbound.

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: @wischli, how do you see adding this to the current RU? THat way we avoid future migration problems because now we need to ensure we have not message there.

If we go with it, just we need to upgrade the version number of this pallet. And adding a reset migration to clean the Nonce storage. WDYT? Of course, I do not want to delay the RU, just if we have time to do it.

runtime/integration-tests/src/cases/lp/investments.rs
```diff
@@ -561,16 +567,20 @@ mod with_foreign_currency {
...
+							pool_id: POOL_A,
+							tranche_id: pool_a_tranche_1_id::<T>(),
+							investor: vec_to_fixed_array(lp::utils::remote_account_of::<T>(
+								Keyring::TrancheInvestor(1)
+							)),
+							currency: utils::index_lp(evm, names::USDC),
+							currency_payout: DEFAULT_INVESTMENT_AMOUNT,
+							fulfilled_invest_amount: DEFAULT_INVESTMENT_AMOUNT,
+						},
+					},
```
lemunozm, line 583: Maybe here, the sender/destination can just checked in `process_outbound` method, leaving each test as before without that overhead

lemunozm (commented):
Really really love the test improvements here. Just one minor comment in order to be more DRY

runtime/integration-tests/src/cases/lp/investments.rs
lemunozm: I mean, I think for all cases the `sender` and `destination` are always the same

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Forgot the above, I think we can ensure the same conditions easily in the next RU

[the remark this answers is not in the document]
wischli: My understanding was that we can assume there won't be any pending LP messages when we trigger the execution of the update. However, after triggering the RU, it takes one hour for it to be live. Right now, we cannot block those permissionless extrinsics which create outgoing messages. So theoretically, there is a chance. Let's discuss this in the Eng sync today.

cdamian:
Tons-o-conflicts 🎉 Gonna close this PR and open a new one after I manage to rebase successfully...

PR #1943, opened 2024-08-01
Feat/lp v2 use gateway queue

cdamian:
# Description

Use the new Gateway Queue pallet in LP and LP gateway.

## Changes and Descriptions
- Use `pallet-liquidity-pools-gateway-queue` for queueing inbound and outbound message in the `LP gateway`.
- Add new handler traits for inbound and outbound messages that are implemented by `LP` and `LP gateway`, as follows:

`MessageQueue` trait is implemented by the `LP gateway queue` and used by the `LP gateway`  to enqueue both inbound and outbound messages.
`MesageProcessor` trait is implemented by the `LP gateway` and used by the `LP gateway queue` when processing queued messages.
`OutboundMessageHandler` trait is implemented by the `LP gateway` and used by `LP` when sending an outbound message.
`InboundMessageHandler` trait is implemented by `LP` and used by the `LP gateway` when processing inbound messages.

# Checklist:

- [x] Address the benchmark weight suggestion [here](https://github.com/centrifuge/centrifuge-chain/pull/1909#discussion_r1699614979)
- [x] Skip the LP gateway queue in as many integration tests as possible.
- [x] Replace the fudge runtime with the normal runtime in the integration tests, where possible.


pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -141,9 +135,9 @@ pub mod pallet {
 			let message = MessageQueue::<T>::take(nonce).ok_or(Error::<T>::MessageNotFound)?;
 
 			match Self::process_message_and_deposit_event(nonce, message.clone()) {
```
cdamian, line 137: Curios if there's a way to add the weight that we now return from this call to the total call weight.

runtime/integration-tests/src/cases/lp/investments.rs
```diff
@@ -703,10 +742,28 @@ mod with_foreign_currency {
...
 			);
 
-			lp_utils::process_outbound::<T>(|msg| {
+			// Required given that the deposit request address for this account
+			// does not have the extra chain id + EVM tag.
+			let tranche_investor_remote_account =
+				lp::utils::remote_account_of::<T>(Keyring::TrancheInvestor(1));
+			let tranche_investor_evm_account = &tranche_investor_remote_account.as_slice()[..20];
+
+			lp_utils::process_gateway_message::<T>(|msg| {
```
cdamian, line 751: This was added on top of what we already had. Needed to check the initial inbound message that is sent.

pallets/liquidity-pools-gateway/queue/src/lib.rs
lemunozm: Maybe in the expansion of that we can check if it's correctly reverted

cdamian:
``` impl<T: Config> frame_support::dispatch::GetDispatchInfo for Call<T> {
        fn get_dispatch_info(&self) -> frame_support::dispatch::DispatchInfo {
            match *self {
                Self::process_message { ref nonce } => {
                    let __pallet_base_weight = T::WeightInfo::process_message()
                        .saturating_add(T::process_message_and_deposit_event().1);
                    let __pallet_weight = <dyn frame_support::dispatch::WeighData<
                        (&T::MessageNonce,),
                    >>::weigh_data(&__pallet_base_weight, (nonce,));
                    let __pallet_class = <dyn frame_support::dispatch::ClassifyDispatch<
                        (&T::MessageNonce,),
                    >>::classify_dispatch(&__pallet_base_weight, (nonce,));
                    let __pallet_pays_fee = <dyn frame_support::dispatch::PaysFee<
                        (&T::MessageNonce,),
                    >>::pays_fee(&__pallet_base_weight, (nonce,));
                    frame_support::dispatch::DispatchInfo {
                        weight: __pallet_weight,
                        class: __pallet_class,
                        pays_fee: __pallet_pays_fee,
                    }
                }
```

cdamian: But it doesn't compile, gotta play with this for a bit.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -243,32 +204,6 @@ pub mod pallet {
 	pub type RelayerList<T: Config> =
 		StorageDoubleMap<_, Blake2_128Concat, Domain, Blake2_128Concat, DomainAddress, ()>;
 
-	#[pallet::storage]
```
cdamian, line 246: Removed these.

pallets/liquidity-pools-gateway/queue/src/lib.rs
cdamian: Using the standard weight calculation that we have in there, I ran:

```
target/release/centrifuge-chain benchmark pallet \
  --chain=development \
  --steps=50 \
  --repeat=20 \
  --pallet=pallet-liquidity-pools-gateway-queue \
  --extrinsic="*" \
  --wasm-execution=compiled \
  --heap-pages=4096 \
  --output=test.rs
```

and got [these weights](https://gist.github.com/cdamian/e1aa677e6e0046662a3caa28ef3be913)

libs/types/src/domain_address.rs
```diff
@@ -53,6 +53,12 @@ pub enum DomainAddress {
 	EVM(EVMChainId, [u8; 20]),
 }
 
+impl Default for DomainAddress {
```
lemunozm, line 56: Super NIT: add `#[default]` to the variant

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
```diff
@@ -28,6 +28,7 @@ mod benchmarks {
 	#[benchmark]
 	fn process_message() -> Result<(), BenchmarkError> {
 		let caller: T::AccountId = account("acc_0", 0, 0);
+		//TODO(cdamian): Add a way of retrieving the "heaviest" message.
```
lemunozm, line 31: Reminder. But I would leave it as it is in this PR. I need to know how to fit this with the batch solution. It will need another thought

pallets/liquidity-pools-gateway/queue/src/lib.rs
lemunozm: I think the PoV is enough in those weights 👍🏻

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -684,128 +530,67 @@ pub mod pallet {
...
+			message: T::Message,
+		) -> (DispatchResult, Weight) {
+			let weight = Weight::from_parts(0, T::Message::max_encoded_len() as u64);
 
-			for entry in processed_entries {
-				OutboundMessageQueue::<T>::remove(entry);
+			match T::InboundMessageHandler::handle(domain_address, message) {
+				Ok(_) => (Ok(()), weight),
+				Err(e) => (Err(e), weight),
 			}
```
lemunozm, line 543: The computation of `T::InboundMessageHandler::handle(domain_address, message)` is not counted in the `weight` and it should. Note that the inner computation of `T::InboundMessageHandler` can be a lot. i.e. if the message is something investment-related it can imply changes in foreign investment, opening swap orders, investments, etc...

Nevertheless, there is no easy solution here. I would add manually an extra weight. I would say 200_000_000 and 4096 of PoV can be enough for all cases by now. Interesting in your thoughts @wischli

lemunozm (commented):
I really think this is a super great improvement in our database in several aspects. Love it! Thanks for your work on this @cdamian

I left some minor NIT things and one major related to the weights, that IMO should be fixed before merging.

libs/types/src/domain_address.rs
cdamian: That can only be used with unit type enums :(

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: I would add that extra weight to the 2 extrinsics of the LP gateway queue that can end up processing this message when calling `MessageProcessor::process`. The only other place where this is called is `on_idle` but in that case we use the weight that's returned.

cdamian: Done.

libs/types/src/domain_address.rs
lemunozm: Ohh, didn't know, thanks 👍🏻

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: But we still skip the weight on the on_idle, right?

- `on_idle` uses the weight from `T::MessageProcessor::process`,  which only returns ` Weight::from_parts(0, T::Message::max_encoded_len() as u64)`. We're not taking into account the processed message weight. I think we should add the above weights to `process_inbound_message` as well. To get `on_idle` to compute correctly the weights

cdamian: Oh, right, for inbound messages we only return the `MaxEncodedLen` for the message, I will add the defensive weight in there as well.

lemunozm (approved):
Cool! anything LSGTM (looks Super great to me!!)

wischli (commented):
Beautiful and clean PR - great job!

Just one tiny thing bothers me. The new hardcoded weight `Weight::from_parts(200_000_000, 4096)` seems opaque and IMO we should declare it as a const at least because it is used quite often.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: @wischli, this is the thread about the weights

pallets/liquidity-pools-gateway/queue/src/mock.rs
```diff
@@ -61,13 +57,11 @@ impl Config for Runtime {
 	type WeightInfo = ();
 }
 
+const DEFENSIVE_WEIGHT_REF_TIME: u64 = 5_000_000_000;
```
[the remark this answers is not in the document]
cdamian: Will re-add a const for it and use that everywhere, Luis and I weren't sure if this value makes sense. 

Should we go for `Weight::from_parts(5_000_000_000, 4096)`?

runtime/integration-tests/src/cases/lp/investments.rs
```diff
@@ -162,6 +162,8 @@ mod with_pool_currency {
 			utils::invest(evm, Keyring::TrancheInvestor(1), names::POOL_C_T_1_USDC);
 		});
 
+		env.pass(Blocks::ByNumber(1));
```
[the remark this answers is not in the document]
cdamian: Inbound messages are queued as well, we pass 1 block to ensure that they get processed accordingly.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: The weight is something we need to discuss; I thought `5_000_000_000` was excessively high for what, in my mind, is the worst case. Although probably `200_000_000` is pretty low. But I'm choosing the `5_000_000_000`

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -130,20 +124,23 @@ pub mod pallet {
...
 		///
-		/// NOTE - this extrinsic does not error out during message processing
+		/// NOTES:
+		///   - this extrinsic does not error out during message processing
 		/// to ensure that any storage changes (i.e. to the message queues)
 		/// are not reverted.
-		#[pallet::weight(T::WeightInfo::process_message())]
+		///   - an extra defensive weight is added in order to cover the weight
+		/// used when processing the message.
+		#[pallet::weight(T::WeightInfo::process_message().saturating_add(Weight::from_parts(200_000_000, 4096)))]
```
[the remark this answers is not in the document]
lemunozm: Thread above

pallets/liquidity-pools-gateway/src/lib.rs
wischli: @lemunozm IMO being defensive is desired and not an issue for now since we don't expect LP messages to explode in cardinality for the time being. As long as we don't have somewhat of more realistic overestimate via benchmarks, this value seems feasible to me, especially since it has not caused trouble so far. `5_000_000_000` computational weight equals 1% of the maximum we allow. Does not feel too defensive to me that we can easily fit processing 50+ messages into, at least, the majority of our blocks. WDYT @lemunozm?

wischli: Though knowing that we could fit 1000-1500 transfers into a block if batched (bottleneck is signature verification if submitted in isolation),  I suppose we could lower it by half. Nevertheless, I cast my vote for the safer path.

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -130,7 +131,7 @@ pub mod pallet {
 		/// are not reverted.
 		///   - an extra defensive weight is added in order to cover the weight
 		/// used when processing the message.
-		#[pallet::weight(T::WeightInfo::process_message().saturating_add(Weight::from_parts(200_000_000, 4096)))]
+		#[pallet::weight(T::WeightInfo::process_message().saturating_add(Weight::from_parts(LP_DEFENSIVE_WEIGHT_REF_TIME, LP_DEFENSIVE_WEIGHT_POV)))]
```
lemunozm, line 134: NIT. Maybe the constant could be the own `Weight::from_parts(LP_DEFENSIVE_WEIGHT_REF_TIME, LP_DEFENSIVE_WEIGHT_POV)` (?)

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: It makes sense what you say above! Let's be super defensive by now 👍🏻

pallets/liquidity-pools-gateway/queue/src/benchmarking.rs
lemunozm: BTW, being defensive means that we can avoid these benchmarks I think. So there is no longer an issue here. We can even remove the benchmarks & weights from this 🤔

cdamian: I'm OK with this, @wischli?

lemunozm (approved):
Awesome stuff! Love it!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Just curious, if the storage is totally empty, do we need to remove it then? I think  only `OutboundMessageNonceStore` need to be removed

wischli (approved):
LGTM! Let's merge this beauty

PR #1944, opened 2024-08-01
Add increase & decrease debt to borrow proxy

lemunozm:
# Description


lemunozm:
Should we add also this to the `NonTransfer` proxy? It seems like other loans actions are there

wischli (approved):

hieronx:
> Should we add also this to the NonTransfer proxy? It seems like other loans actions are there

Yes I would say so

wischli (approved):
Thanks for the quick fix!

PR #1945, opened 2024-08-02
Mock builder update. Improved FI coverage

lemunozm:
# Description

- Added new version of mock builder with method handlers, supporting check how many times a mock was called.
- Fixed some wrong configured `default-features` (I do not know why Cargo did not complained)
- Improved FI tests checking the times the hooks are called.
- ~Cargo update~ Something is really off with the cargo update. Just doing `cargo update` on `main` makes the compilation of runtime fail due to std issues. I was not able to find the guilty one.

### DIFF
--- a/libs/traits/Cargo.toml
+++ b/libs/traits/Cargo.toml
@@ -45,11 +45,14 @@ std = [
 runtime-benchmarks = [
   "cfg-primitives/runtime-benchmarks",
   "frame-support/runtime-benchmarks",
+  "frame-system/runtime-benchmarks",
   "sp-runtime/runtime-benchmarks",
   "cfg-mocks/runtime-benchmarks",
 ]
 try-runtime = [
   "frame-support/try-runtime",
+  "frame-system/try-runtime",
   "cfg-primitives/try-runtime",
   "sp-runtime/try-runtime",
+  "cfg-mocks/try-runtime",
 ]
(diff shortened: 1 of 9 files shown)


libs/traits/Cargo.toml
```diff
@@ -25,9 +25,9 @@ sp-std = { workspace = true }
 strum = { workspace = true }
 
 [dev-dependencies]
-cfg-mocks = { workspace = true, default_features = true }
-frame-system = { workspace = true, default_features = true }
-mock-builder = { workspace = true, default_features = true }
+cfg-mocks = { workspace = true, default-features = true }
```
cdamian, line 28: Oh I just noticed the difference here. This is scary lol

cdamian (dismissed):
Awesome improvement, thank you @lemunozm!

libs/traits/Cargo.toml
lemunozm: And no warnings at all by cargo 😅

lemunozm:
I've needed to also update all inner dependencies to get rid of a cargo bug 🫤

cdamian (dismissed):

cdamian (approved):

hieronx (approved):

lemunozm:
Could you force @hieronx ? It seems like we also need the Frederik approval

PR #1946, opened 2024-08-05
migration: Add LP Gateway migration for outbound messages

cdamian:
# Description

Add a migration that removes the LP gateway storages related to outbound messages.

# Checklist:


runtime/common/src/migrations/liquidity_pools_gateway.rs
```diff
@@ -0,0 +1,131 @@
...
+			log::warn!("{LOG_PREFIX}: Found outbound message:\nnonce:{nonce}\nentry:{entry:?}");
+			weight = weight.saturating_add(<T as frame_system::Config>::DbWeight::get().reads(1));
+		}
+
+		for (nonce, entry) in FailedOutboundMessages::<T>::iter() {
+			log::warn!(
+				"{LOG_PREFIX}: Found failed outbound message:\nnonce:{nonce}\nentry:{entry:?}"
+			);
+			weight = weight.saturating_add(<T as frame_system::Config>::DbWeight::get().reads(1));
+		}
```
lemunozm, line 67: I'm pretty sure if they're empty, we do not need to remove them. 

cc @wischli

cdamian: Yea, this is more like an extra check to ensure that they're empty. Not doing any specific removals here apart from the nonce storage.

cdamian:
Testing this locally using:
```
try-runtime --runtime existing create-snapshot --uri [wss://fullnode.centrifuge.io:443](wss://fullnode.centrifuge.io/) cfg.snap

try-runtime --runtime target/release/wbuild/centrifuge-runtime/centrifuge_runtime.wasm on-runtime-upgrade --disable-spec-version-check snap -p cfg.snap
```

However, it errors out:
```
[2024-08-05T09:30:44Z ERROR runtime] panicked at /Users/cdamian/codebase/centrifuge-chain/runtime/common/src/migrations/liquidity_pools_gateway.rs:97:9:
    assertion `left == right` failed: OutboundMessageNonce should be empty!
      left: 55
     right: 0
thread 'main' panicked at cli/main.rs:324:10:
called `Result::unwrap()` on an `Err` value: Input("failed to execute TryRuntime_on_runtime_upgrade: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed\nWASM backtrace:\nerror while executing at wasm backtrace:\n ...
```

lemunozm:
That's normal. you have a `ValueQuery` in that storage. So if no entry is found, a default is returned. But that does not add any real entry in the storage. I think you should check that the value is 0.

runtime/common/src/migrations/liquidity_pools_gateway.rs
```diff
@@ -0,0 +1,127 @@
...
+			0,
+			"FailedOutboundMessages should be empty!"
+		);
+
+		Ok(Vec::new())
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(_: Vec<u8>) -> Result<(), TryRuntimeError> {
+		assert_eq!(
```
cdamian, line 93: @lemunozm this is the assert that fails when running `try-runtime`.

Not sure where that panic comes from tho, that's a different story.

lemunozm: What's the error it tells you? `"OutboundMessageNonceStore should be 0!"` ?

cdamian: Yes, with the latest version that's what I get in the `post` check. But, I guess that makes sense since the `on_runtime_upgrade` doesn't really get triggered or am I missing something?

wischli: Looking into this now

lemunozm: Mmm.. I think the assert should pass IIUC this 🤔

lemunozm: The previous message was failing correctly because "there is" entry if you read. But if you check the value it should be the default one, 0

wischli (commented):
Wrong prefix constructor for clearing. Apart from that I would like to add a storage value to the pallet and bump it by one because this is a breaking storage change.

Luckily, you don't need to extend the code much because you can wrap your migration into a `VersionedMigration` exposed from the Polkadot SDK: https://paritytech.github.io/polkadot-sdk/master/frame_support/migrations/struct.VersionedMigration.html

Unfortunately, the `try-runtime` will still fail because of two unsynchronized pallets:

```
[2024-08-06T09:58:25Z ERROR runtime::frame-support] ForeignInvestments: On chain storage version StorageVersion(1) doesn't match current storage version StorageVersion(2).
[2024-08-06T09:58:25Z ERROR runtime::frame-support] DmpQueue: On chain storage version StorageVersion(0) doesn't match current storage version StorageVersion(2).
```

For FI, we just need to bump and the DMP Queue can actually be removed because it is deprecated but we needed to keep it in the migration to Polkadot v1.7.2 because that included a migration to the new pallet `MessageQueue` / `pallet_message_queue`. 

I am happy to open a PR with all the fixes including wrapping the gateway migration into a versioned one. WDYT @cdamian ?

runtime/common/src/migrations/liquidity_pools_gateway.rs
```diff
@@ -0,0 +1,127 @@
...
+
+pub struct Migration<T>(sp_std::marker::PhantomData<T>);
+
+impl<T> OnRuntimeUpgrade for Migration<T>
+where
+	T: Config + frame_system::Config,
+{
+	fn on_runtime_upgrade() -> Weight {
+		let mut weight = Self::clear_storage(
+			OutboundMessageNonceStore::<T>::storage_prefix(),
```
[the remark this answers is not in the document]
cdamian: Oh, I thought the `storage_alias` takes care of that.

wischli:
The `try-runtime` CLI [works now](https://github.com/centrifuge/centrifuge-chain/actions/runs/10297766903/job/28501512290)! It failed for Altair because in the `pre_upgrade` we are assuming the `OutboundMessageNonceStore` to be greater than 0 which obviously isn't the case for Altair. Normally, I would fix that but given that Altair is low priority, especially currently with the strict audit deadline, and is not expected to utilize LPs, we can ignore it IMO.

wischli (approved):
Approving our mutual work. Great job 😅

lemunozm (approved):
Thanks!

PR #1948, opened 2024-08-06
lp-gateway: Add multi-router support

cdamian:
# Description

### Multi-router
Added new extrinsic for adding multi-routers, and 2 new storages for multi routers:
  - one storage that maps a router hash to a router.
    - the `Router` trait now has a `hash()` method that allows us to create a hash using certain router fields.
  - one storage that maps a domain to a bounded vec of router hashes.

#### Message proofs
The `LPEnconding` trait has 2 new methods used for:
- retrieving a message proof from a message proof message, if any.
- converting a message to a message proof message.

#### Outbound
The `OutboundMessageHandler` implementation is now queueing one message using one router and message proofs using the remaining routers.

#### Inbound
Added 2 new storages for keeping track of inbound messages and message proofs.

The inbound message processing logic is checking that the required number of proofs are received before submitting the message to the `InboundMessageHandler`.

# Checklist:

- [ ] Invalidate all message proofs if the routers for a domain change


pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -116,7 +116,7 @@ pub mod pallet {
 		type Message: LPEncoding + Clone + Debug + PartialEq + MaxEncodedLen + TypeInfo + FullCodec;
 
 		/// The message router type that is stored for each domain.
-		type Router: DomainRouter<Sender = Self::AccountId>
+		type Router: DomainRouter<Sender = Self::AccountId, Hash = Self::Hash>
```
lemunozm, line 119: Are we sure the Hash used by Solidity is the same as the Hash used by `frame_system::Config`?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -208,6 +223,32 @@ pub mod pallet {
 	pub type DomainHookAddress<T: Config> =
 		StorageMap<_, Blake2_128Concat, Domain, [u8; 20], OptionQuery>;
 
+	/// Storage for routers.
+	///
+	/// This can only be set by an admin.
+	#[pallet::storage]
+	#[pallet::getter(fn routers)]
+	pub type Routers<T: Config> = StorageMap<_, Blake2_128Concat, T::Hash, T::Router>;
```
lemunozm, line 231: Does this storage overlap somehow with `DomainRouters`? Could/should the previous be removed?

libs/traits/src/liquidity_pools.rs
```diff
@@ -23,6 +23,9 @@ use sp_std::vec::Vec;
 pub trait LPEncoding: Sized {
 	fn serialize(&self) -> Vec<u8>;
 	fn deserialize(input: &[u8]) -> Result<Self, DispatchError>;
+
+	fn get_message_proof(&self) -> Option<[u8; 32]>;
```
lemunozm, line 27: NIT. Can we use a type alias for `[u8; u32]`, maybe called `Proof` or similar? I would add it just before LPEncoding trait, as "part" of the interface

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -485,6 +541,45 @@ pub mod pallet {
...
+		#[pallet::call_index(9)]
+		pub fn set_domain_multi_router(
+			origin: OriginFor<T>,
+			domain: Domain,
+			routers: BoundedVec<T::Router, T::MaxRouterCount>,
+		) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
+
+			ensure!(domain != Domain::Centrifuge, Error::<T>::DomainNotSupported);
+			ensure!(routers.len() > 0, Error::<T>::InvalidMultiRouter);
```
lemunozm, line 556: Why not `== T::MaxRouterCount`? Later in the incoming message, we do not need to check for the router counts. We need that it's always well configured.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -485,6 +541,45 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		/// Set routers for a particular domain.
+		#[pallet::weight(T::WeightInfo::set_domain_multi_router())]
+		#[pallet::call_index(9)]
+		pub fn set_domain_multi_router(
+			origin: OriginFor<T>,
+			domain: Domain,
+			routers: BoundedVec<T::Router, T::MaxRouterCount>,
```
lemunozm, line 551: If we also need the `MaxRouterCount`, we not an array instead?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -530,9 +625,74 @@ pub mod pallet {
...
+
+					(
+						message_proof,
+						InboundMessageProofCount::<T>::get(message_proof),
+					)
+				}
+				Some(message_proof) => {
+					let message_proof_count =
+						match InboundMessageProofCount::<T>::try_mutate(message_proof, |count| {
+							*count += 1;
```
lemunozm, line 661: Could the same router send the proof 3 times and we will count 3?

lemunozm (commented):
I left some comments but I think the idea is the correct one!

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -530,9 +625,74 @@ pub mod pallet {
...
+			let expected_proof_count: u32 = (routers_count - 1) as u32;
+
+			let (message_proof, message_proof_count) = match message.get_message_proof() {
+				None => {
+					let message_proof = message
+						.to_message_proof()
+						.get_message_proof()
+						.expect("message proof ensured by 'to_message_proof'");
+
+					InboundMessages::<T>::insert(message_proof, message.clone());
```
cdamian, line 651: This has to be changed in order to support the same message being sent more than once.

libs/traits/src/liquidity_pools.rs
cdamian: OK, was thinking about doing that but wasn't sure if we need extra constraints on it so that we can use it in storage. Gonna try it out.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: I don't think that these hashes have to match, mainly because these are specific to the LP gateway here.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Yes, I'm also not sure if we should add this new storage or just change the old one and write a migration for it.

libs/traits/src/liquidity_pools.rs
lemunozm: I think by now, we can force the type to be that type. Everything regarding LP is quite tied to our use case. It's not as generic as other pallets.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: I'm not sure I understand your question.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: That's better, I'm gonna change this.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: We can then force the type to use the `[u8; 32]` (the `Proof` alias) as the standard dictates so

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: In this current state, yes. Especially given that we can get the same message more than once.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: The user who call `set_domain_multi_router` ideally needs to put always 3 routers in the call?

lemunozm: If so, then `[T::Router, T::MaxRouterCount>]` already force that

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: This is not the message proof/hash. This is simply a hash for the router that we use in storage to avoid unnecessary storage of routers when queueing message and so on.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Both solutions will need a migration anyways. But luckily, we can just purge the previous configuration and start using this new. So whatever you think it's simpler. 

I think any migration could be done after the audit, because the audit only requires to have the code submitted in the repo

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Ok, is it like an Id of the router, right?

cdamian: Yes. 

And the idea of having it as a hash that we choose how to calculate allows us to change/update a router without necessarily changing its hash. If we have a router with some invalid gas limit, we can replace it with one with a correct gas limit but with the same hash. Was thinking that this might make our lives easier if we are to retry the execution of a failed message. WDYT?

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: I also tend to be in favor of using the new one and dropping the old one. Mainly because this migration is easier than updating an existing storage to one with a different layout.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -530,9 +625,74 @@ pub mod pallet {
...
+				None => return (Err(Error::<T>::MultiRouterNotFound.into()), weight),
+			};
+
+			weight = weight.saturating_add(T::DbWeight::get().reads(1));
+
+			if routers_count <= 1 {
+				return (Err(Error::<T>::NotEnoughRouters.into()), weight);
+			}
+
+			let expected_proof_count: u32 = (routers_count - 1) as u32;
```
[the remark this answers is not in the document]
hieronx: Yes Solidity has full implementation of the message proofs already. As well as initiate/dispute message recovery.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian:
```
error: generic parameters may not be used in const operations
   --> pallets/liquidity-pools-gateway/src/lib.rs:555:25
    |
555 |             routers: [T::Router; T::MaxRouterCount],
    |                                  ^^^^^^^^^^^^^^^^^ cannot perform const operation using `T`
 ```
 
 I propose to go back to using the `BoundedVec` and ensure that it's length is always `T::MaxRouterCount`.

File: pallets/liquidity-pools-gateway/src/lib.rs (hunk 1 above)
Comment (lemunozm) [reply pallets/liquidity-pools-gateway/src/lib.rs:119]: TBH, it all depends on how this will be used. 

However taking into account the last thing Jeroen has said about "votes" and how each time a router is updated we need invalidate all. Maybe just allowing the user to set all routers each time and use the index arrays is enough. But whatever you think it's easier/simpler!

File: pallets/liquidity-pools-gateway/src/lib.rs (hunk 5 above)
Comment (lemunozm) [reply pallets/liquidity-pools-gateway/src/lib.rs:551]: Ouch, I forgot about that is not constant. Ideally you should be able to write `T::MaxRouterCount::get()`, but any method from a trait can not be `const`, that's a pity.

Nevertheless, there is still a solution, which is adding a `const MaxRouterCount: u32;` to the Config trait. But again, what you think is better!

File: pallets/liquidity-pools-gateway/src/lib.rs [hunk 9]
```diff
@@ -485,6 +536,46 @@ pub mod pallet {
...
+
+				Routers::<T>::insert(router_hash, router);
+			}
+
+			<DomainMultiRouters<T>>::insert(
+				domain.clone(),
+				BoundedVec::try_from(router_hashes).map_err(|_| Error::<T>::InvalidMultiRouter)?,
+			);
+
+			Self::clear_storages_for_inbound_messages();
```
cdamian, line 573: At the moment, we're clearing both storages for inbound messages when a new set of routers is added.

@lemunozm @hieronx you discussed about using a `session_id` but I fail to see how that will help us here? Do we need to keep track of which routers were stored/used at a certain point in time?

Another interesting question is - what is the expected behavior when a router set is changed but there are still messages in the queue?

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Array indexes might be problematic if the new set of routers has a different ordering. I would still prefer to have hashes tbh.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Added some tests with different nonproof/proof messages to ensure that these storages remain sane, please let me know what you think.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: The idea with the `session_id` is to not clear the storages, just if the storage has not the last `session_id`, then, you treat its data as an invalid (as if it was removed). Saving you from iterate over all entries removing the data. But, what you think is easier on our side. Just note that clear a storage will iterate over all entries and we can exceed the block weight, so we need to be careful here.

Regarding the second question, I would say the message can still be there. If the routers has changed you will need to wait to the new proofs of those routers. But there is nothing wrong with the message.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Note that because all router information is reset each time you set new routers, there is no problem there. There is nothing wrong with the hashes, but it's true that it requires a bit more code (in the router side) and a bit more storage on-chain because a hash is bigger than a number.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -525,14 +616,114 @@ pub mod pallet {
...
+					let message_proof = message
+						.to_message_proof()
+						.get_message_proof()
+						.expect("message proof ensured by 'to_message_proof'");
+
+					match InboundMessages::<T>::try_mutate(message_proof, |storage_entry| {
+						match storage_entry {
+							None => {
+								*storage_entry =
+									Some((message, Self::get_expected_message_proof_count()));
```
lemunozm, line 645: Should not be `(message, 1)`

lemunozm (changes requested):
I think the overall flow is the correct one, but I think there are some math issues on it

pallets/liquidity-pools-gateway/src/message.rs
```diff
@@ -1,21 +1,21 @@
...
+pub enum GatewayMessage<AccountId, Message, Hash> {
 	Inbound {
 		domain_address: DomainAddress,
 		message: Message,
 	},
 	Outbound {
 		sender: AccountId,
-		destination: Domain,
 		message: Message,
+		router_hash: Hash,
```
lemunozm, line 14: Why is this need it?

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: > The idea with the session_id is to not clear the storages, just if the storage has not the last session_id, then, you treat its data as invalid (as if it was removed). 

I think we can avoid using a session ID for now, mainly because we clear the inbound message storages once a message received the required number of proofs and was sent to the LP for further processing.

> Saving you from iterate over all entries removing the data. But, what you think is easier on our side. Just note that clear a storage will iterate over all entries and we can exceed the block weight, so we need to be careful here.

Once a new set of routers is added, all previous proofs are no longer valid right? So we'll have to clear the proof storage completely anyway.

> Regarding the second question, I would say the message can still be there. If the routers has changed you will need to wait to the new proofs of those routers. But there is nothing wrong with the message.

My bad, I phrased that question in a wrong way. What I meant was - on a new set of routers, we're clearing the inbound message storages which keep track of what messages are waiting to be processed and the current number of proofs that we got for them. However, given your explanation, it seems that we should only clear the storage that keeps track of the number of proofs and expect that those proofs will be re-sent. Is this correct?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -525,14 +616,114 @@ pub mod pallet {
 			Ok((address, incoming_msg))
 		}
 
-		/// Sends the message to the `InboundMessageHandler`.
+		fn clear_storages_for_inbound_messages() {
+			let _ = InboundMessages::<T>::clear(u32::MAX, None);
+			let _ = InboundMessageProofCount::<T>::clear(u32::MAX, None);
```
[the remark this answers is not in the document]
cdamian: True, should we maybe add a flag that marks this storage for clearing and then attempt to clear this `on_idle`?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -525,14 +616,114 @@ pub mod pallet {
...
+		}
+
+		fn get_expected_message_proof_count() -> u32 {
+			T::MultiRouterCount::get() - 1
+		}
```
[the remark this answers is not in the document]
cdamian: No, if we have 3 routers - we expect 1 message and 2 proofs.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -525,14 +616,114 @@ pub mod pallet {
...
+						match storage_entry {
+							None => {
+								*storage_entry =
+									Some((message, Self::get_expected_message_proof_count()));
+							}
+							Some((_, expected_proof_count)) => {
+								// We already have a message, in this case we should expect another
+								// set of message proofs.
+								expected_proof_count
+									.ensure_add_assign(Self::get_expected_message_proof_count())?;
```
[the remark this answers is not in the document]
cdamian: Nope, see my message above.

pallets/liquidity-pools-gateway/src/message.rs
cdamian: In the `OutboundMessageHandler` implementation we queue one message per router. Later on when the queue is processing messages, we retrieve the router based on its hash - https://github.com/centrifuge/centrifuge-chain/blob/feat/lp-v2-gateway-multi-router/pallets/liquidity-pools-gateway/src/lib.rs#L743

libs/traits/src/liquidity_pools.rs
```diff
@@ -46,11 +51,22 @@ pub mod test_util {
...
 			}
 		}
+
+		fn get_message_proof(&self) -> Option<Proof> {
+			Some([1; 32])
+		}
+
+		fn to_message_proof(&self) -> Self {
+			Self
+		}
```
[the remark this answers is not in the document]
lemunozm: Ah ok, the `None` makes sense 👍🏻.

I was thinking in something like:

```rust
trait LPEncoding { // Probably this should be call differently now
    fn create_proof(proof: Proof) -> Message;
    fn get_proof(&self) -> Proof;
}
```

But yours works fine

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Ah, right!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: > Once a new set of routers is added, all previous proofs are no longer valid right? So we'll have to clear the proof storage completely anyway.

That's the key, you don't need it. You simply doesn't nothing at that point. Later, when a new inbound message is read, you check before the session id, if it's different than the last one, then you can remove that entry, just that entry, and continue as usual

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I think this is when the `session_id` has to the rescue. `on_idle` should be as small as possible.

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Makes sense now!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: But maybe the hashes makes sense because you need them later to know which one to choose in order to send the message 👍🏻

pallets/liquidity-pools-gateway/src/message.rs
lemunozm: Makes sense!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: I'll check in the next review pass!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: > However, given your explanation, it seems that we should only clear the storage that keeps track of the number of proofs and expect that those proofs will be re-sent. Is this correct?

Yes. 

Suppose you have received a message and one proof (still pending another proof). In that moment, routers are updated, and proofs are invalidated, but not the message AFAIU.

Do we expect the new routers to send all the proofs? I think so. If not we should also remove the message, or it will be there forever.

cc @hieronx?

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
cdamian: OK, I will wrap up the normal inbound message processing logic to make sure I get it right then I can do another pass to improve storages etc.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -525,14 +616,114 @@ pub mod pallet {
...
+					}
+				};
+
+			weight = weight.saturating_add(T::DbWeight::get().reads(1));
+
+			if current_message_proof_count < Self::get_expected_message_proof_count() {
+				return (Ok(()), weight);
+			}
+
+			if current_message_proof_count == total_expected_proof_count {
```
lemunozm, line 720: @cdamian, just one important thing. How here do we know that the counts belong to different routers and not from the same router? I think this is where the concept of "vote" comes in

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -208,6 +223,33 @@ pub mod pallet {
...
+	#[pallet::storage]
+	#[pallet::getter(fn inbound_message_proof_count)]
+	pub type InboundMessageProofCount<T: Config> =
+		StorageMap<_, Blake2_128Concat, Proof, u32, ValueQuery>;
+
+	/// Storage that keeps track of incoming messages and the expected proof
+	/// count.
+	#[pallet::storage]
+	#[pallet::getter(fn inbound_messages)]
+	pub type InboundMessages<T: Config> = StorageMap<_, Blake2_128Concat, Proof, (T::Message, u32)>;
```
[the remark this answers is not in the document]
lemunozm: And when a message comes from a router, it will add a vote automatically. That way you can process the message when votes from all are higher than one. 

Not sure I'm explaining this well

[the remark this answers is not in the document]
cdamian: For inbound messages, we do not have a way of identifying routers, we just get a message from the queue that has a domain address and the actual message. The router hashes are only used in the context of outbound messages.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: That's the fun part, we do not...

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: That's correct, but why then the Solidity part is as it is (?). Maybe we need to specify a new parameter in the axelar part that carries that info?

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Not sure if we can use the domain_address though. If it comes via EVM it's most likely the same sender - the gateway router contract deployed on evm. Or maybe I'm mixing things up...

pallets/liquidity-pools-gateway/src/lib.rs
hieronx: > In that moment, routers are updated, and proofs are invalidated, but not the message AFAIU.

All votes should be invalidated, both message and proofs.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -525,14 +616,114 @@ pub mod pallet {
...
 		}
 
-		/// Sends the message to the `InboundMessageHandler`.
+		fn clear_storages_for_inbound_messages() {
+			let _ = InboundMessages::<T>::clear(u32::MAX, None);
+			let _ = InboundMessageProofCount::<T>::clear(u32::MAX, None);
+		}
+
+		fn get_expected_message_proof_count() -> u32 {
+			T::MultiRouterCount::get() - 1
```
wischli, line 625: Again, we must use safe math as long as we cannot restrict `MultiRouterCount` to be at least one.

```suggestion
			T::MultiRouterCount::get().saturating_sub(1)
```

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -525,14 +616,114 @@ pub mod pallet {
...
+				Some(message_proof) => {
+					let message_proof_count =
+						match InboundMessageProofCount::<T>::try_mutate(message_proof, |count| {
+							count.ensure_add_assign(1)?;
+
+							Ok(*count)
+						}) {
+							Ok(r) => r,
+							Err(e) => return Err(e),
+						};
```
wischli, line 677: Am I missing something or can we simplify readability by utilising `?`?:

```suggestion
						match InboundMessageProofCount::<T>::try_mutate(message_proof, |count| {
							count.ensure_add_assign(1)?;

							Ok(*count)
						})?;
```

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -525,14 +616,114 @@ pub mod pallet {
...
+			let mut weight = Weight::from_parts(0, T::Message::max_encoded_len() as u64)
 				.saturating_add(LP_DEFENSIVE_WEIGHT);
 
+			let (message_proof, current_message_proof_count) =
+				match Self::update_inbound_message_storages(message, &mut weight) {
+					Ok(r) => r,
+					Err(e) => return (Err(e), weight),
+				};
+
+			weight = weight.saturating_add(T::DbWeight::get().reads(1));
```
wischli, line 702: Nit
```suggestion
			weight.saturating_accrue(T::DbWeight::get().reads(1));
```

pallets/liquidity-pools-gateway/src/lib.rs
wischli: It appears to me that we cannot rely on the `domain_address` based on our Slack conversation. We need to assume that in the future each domain has more than one configured router. So I would derive it "special origin" for the router, e.g. in case of Axelar `axelar_gateway_precompile::Pallet::<T>::GatewayContract` (see my [slack message](https://kflabs.slack.com/archives/C06FRHCRQL9/p1723044834860239?thread_ts=1723038229.117829&cid=C06FRHCRQL9)).

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -544,12 +735,12 @@ pub mod pallet {
 		/// weight for these operations in the `DispatchResultWithPostInfo`.
 		fn process_outbound_message(
 			sender: T::AccountId,
-			domain: Domain,
```
wischli, line 547: Placeholder (might answer itself further down below): I am curios how or where we determine the destination domain instead now?

wischli (commented):
Overall looks good! Of course we need to iron out open questions and apply our current Slack conversation. Really nice work with the UTs!

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -485,6 +536,46 @@ pub mod pallet {
...
+			ensure!(domain != Domain::Centrifuge, Error::<T>::DomainNotSupported);
+			ensure!(
+				routers.len() == T::MultiRouterCount::get() as usize,
+				Error::<T>::InvalidMultiRouter
+			);
+
+			let mut router_hashes = Vec::new();
+
+			for router in &routers {
+				router.init().map_err(|_| Error::<T>::RouterInitFailed)?;
```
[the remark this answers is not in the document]
cdamian: Awesome suggestion. thank you!

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Yes, most of the logic for inbound messages is super-dirty, will change all of these once I figure out proof processing accordingly.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Might drop these fugly `try_mutate`s.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Ah, forgot about that, thanks!

pallets/liquidity-pools-gateway/src/lib.rs
wischli: Makes absolute sense. Better have something working with some dirt than have a shining non-functional beauty 😅

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I didn't know substrate supported custom content in the error; if it works, amazing!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Me to!

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
wischli: Substrate did not support this from Genesis but IIRC it was added early last year. However, since we have lacked behind Substrate versions so much, we haven't utilised it yet.

pallets/liquidity-pools-gateway/routers/src/routers/axelar_evm.rs
```diff
@@ -77,6 +78,10 @@ where
 
 		self.router.do_send(sender, eth_msg)
 	}
+
+	pub fn hash(&self) -> T::Hash {
+		BlakeTwo256::hash(self.evm_chain.encode().as_slice())
+	}
```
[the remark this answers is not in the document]
cdamian: Yes, this is mainly for the LP gateway to map a router hash to a router in storage.

cdamian:
Too many conflicts, gonna start fresh from main.

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
cdamian: It seems we can add other `PalletError`s in there - https://github.com/paritytech/polkadot-sdk/blob/149c70938f2b29f8d92ba1cc952aeb63d4084e27/bridges/snowbridge/pallets/inbound-queue/src/lib.rs#L185.

pallets/liquidity-pools-gateway/src/tests.rs
```diff
@@ -530,36 +649,1896 @@ mod message_processor_impl {
...
+							TEST_DATA.clone(),
+							expected_results,
+							MESSAGE_COUNT,
+						);
+
+						run_inbound_message_test_suite(suite);
+					}
+				}
+
+				mod four_messages {
```
lemunozm, line 1225: If it works for 3, can we ensure it will always work for 4?

cdamian: In this case, we are testing with 2 routers, and the minimum number of messages needed for a successful message submission is 2. So, if it works with 2, it should work with more messages. 

Initially, I added tests for up to 4 messages to confirm that duplicate messages are processed accordingly. Still thinking about generating correct expected results so we can slim these tests down a bit.

cdamian:
Will close this and rebase on https://github.com/centrifuge/centrifuge-chain/pull/1958 soon.

PR #1949, opened 2024-08-07
LPv2: Batch message logic (r2)

lemunozm:
# Description

This PR overrides the previous https://github.com/centrifuge/centrifuge-chain/pull/1923, in a simplified version

### Responsibilities
- `pallet-liquidity-pools-gateway` knows to batch/unbatch but doesn't know what a `Batch` message is. To handle this, the `LPMessage` trait allows pack/unpack methods.
- `pallet-liquidity-pools` will never send or receive a batch. The gateway is responsible for the batching process.

### Details
Batching is handled by two extrinsic: 
  - `Gateway::start_pack_messages(sender, destination)`:  When calling this, any submitted messages from LP, that match the `sender` & `destination` will be packed. _NOTE: We need to index also by destination domain because the destination information is not inside the message. So, we can never bundle two messages with two different destinations._
  - `Gateway::end_pack_messages(sender, destination)`: When calling this, the batch is finally submitted. If this is not called, no message is sent. If the batch reaches the limit size, it will dispatch an error.
  
Later, the user can call `pallet_utility::batch(calls)`, and any message sent between `start_pack` and `end_pack` will be bundled in the batch.

### Changes
- Added `Gateway::start_pack_messages()` and `Gateway::end_pack_messages()` to allow creating batches.
- Renamed `process_msg()` to `receive_message()`.
- Removed `try_range` for header message deserialization and use a `BufferReader`.

### DIFF
--- a/libs/traits/src/liquidity_pools.rs
+++ b/libs/traits/src/liquidity_pools.rs
@@ -21,8 +21,21 @@ use sp_std::vec::Vec;
 /// An encoding & decoding trait for the purpose of meeting the
 /// LiquidityPools General Message Passing Format
 pub trait LPEncoding: Sized {
+	const MAX_PACKED_MESSAGES: u32;
+
 	fn serialize(&self) -> Vec<u8>;
 	fn deserialize(input: &[u8]) -> Result<Self, DispatchError>;
+
+	/// Compose this message with a new one
+	fn pack(&self, other: Self) -> Result<Self, DispatchError>;
+
+	/// Decompose the message into a list of messages
+	/// If the message is not decomposable, it returns the own message.
+	fn unpack(&self) -> Vec<Self>;
+
+	/// Creates an empty message.
+	/// It's the identity message for composing messages
+	fn empty() -> Self;
 }
 
 #[cfg(any(test, feature = "std"))]
@@ -32,9 +45,14 @@ pub mod test_util {
 
 	use super::*;
 
+	pub const DECODING_ERR_MSG: &str = "decoding message error";
+	pub const EMPTY_ERR_MSG: &str = "empty message error error";
+
 	#[derive(Default, Debug, Eq, PartialEq, Clone, Encode, Decode, TypeInfo, MaxEncodedLen)]
 	pub struct Message;
 	impl LPEncoding for Message {
+		const MAX_PACKED_MESSAGES: u32 = 1;
+
 		fn serialize(&self) -> Vec<u8> {
 			vec![0x42]
 		}
@@ -42,9 +60,21 @@ pub mod test_util {
 		fn deserialize(input: &[u8]) -> Result<Self, DispatchError> {
 			match input.first() {
 				Some(0x42) => Ok(Self),
-				Some(_) => Err("unsupported message".into()),
-				None => Err("empty message".into()),
+				Some(_) => Err(DECODING_ERR_MSG.into()),
+				None => Err(EMPTY_ERR_MSG.into()),
 			}
+		}
+
+		fn pack(&self, _: Self) -> Result<Self, DispatchError> {
+			unimplemented!()
+		}
+
+		fn unpack(&self) -> Vec<Self> {
+			vec![Self]
+		}
+
+		fn empty() -> Self {
+			unimplemented!()
 		}
 	}
 }
--- a/pallets/liquidity-pools-gateway/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/src/lib.rs
@@ -38,7 +38,7 @@ use cfg_traits::{
 };
 use cfg_types::domain_address::{Domain, DomainAddress};
 use frame_support::{dispatch::DispatchResult, pallet_prelude::*, PalletError};
-use frame_system::pallet_prelude::OriginFor;
+use frame_system::pallet_prelude::{ensure_signed, OriginFor};
 use message::GatewayMessage;
 use orml_traits::GetByKey;
 pub use pallet::*;
@@ -208,6 +208,13 @@ pub mod pallet {
 	pub type DomainHookAddress<T: Config> =
 		StorageMap<_, Blake2_128Concat, Domain, [u8; 20], OptionQuery>;
 
+	/// Stores a packed message, not ready yet to be enqueue.
+	/// Lifetime handled by `start_pack_messages()` and `end_pack_messages()`
+	/// extrinsics.
+	#[pallet::storage]
+	pub(crate) type PackedMessage<T: Config> =
+		StorageMap<_, Blake2_128Concat, (T::AccountId, Domain), T::Message>;
+
 	#[pallet::error]
 	pub enum Error<T> {
 		/// Router initialization failed.
@@ -218,9 +225,6 @@ pub mod pallet {
 
 		/// The domain is not supported.
 		DomainNotSupported,
-
-		/// Message decoding error.
-		MessageDecodingFailed,
 
 		/// Instance was already added to the domain.
 		InstanceAlreadyAdded,
@@ -246,6 +250,14 @@ pub mod pallet {
 		/// Decoding that is essential and this error
 		/// signals malforming of the wrapping information.
 		RelayerMessageDecodingFailed { reason: RelayerMessageDecodingError },
+
+		/// Emitted when you call `start_pack_messages()` but that was already
+		/// called. You should finalize the message with `end_pack_messages()`
+		MessagePackingAlreadyStarted,
+
+		/// Emitted when you can `end_pack_messages()` but the packing process
+		/// was not started by `start_pack_messages()`.
+		MessagePackingNotStarted,
 	}
 
 	#[pallet::call]
@@ -356,16 +368,14 @@ pub mod pallet {
 		}
 
 		/// Process an inbound message.
-		#[pallet::weight(T::WeightInfo::process_msg())]
+		#[pallet::weight(T::WeightInfo::receive_message())]
 		#[pallet::call_index(5)]
-		pub fn process_msg(
+		pub fn receive_message(
 			origin: OriginFor<T>,
 			msg: BoundedVec<u8, T::MaxIncomingMessageSize>,
 		) -> DispatchResult {
-			let (domain_address, incoming_msg) = match T::LocalEVMOrigin::ensure_origin(origin)? {
-				GatewayOrigin::Domain(domain_address) => {
-					Pallet::<T>::validate(domain_address, msg)?
-				}
+			let (origin_address, incoming_msg) = match T::LocalEVMOrigin::ensure_origin(origin)? {
+				GatewayOrigin::Domain(domain_address) => (domain_address, msg),
 				GatewayOrigin::AxelarRelay(domain_address) => {
 					// Every axelar relay address has a separate storage
 					ensure!(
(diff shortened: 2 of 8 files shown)


pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -547,55 +536,29 @@ pub mod pallet {
...
-				Self::get_outbound_message_processing_weight(router_call_weight);
-
-			*weight = weight.saturating_add(router_call_weight);
-		}
+			// TODO: do we really need to return the weights in send() if later we use the
+			// defensive ones?
+			let (result, router_weight) = match router.send(sender, serialized) {
+				Ok(dispatch_info) => (Ok(()), dispatch_info.actual_weight),
+				Err(e) => (Err(e.error), e.post_info.actual_weight),
+			};
```
lemunozm, line 553: @cdamian, what do you think about this? Could we just convert `router.send()` to return just a simple `DispatchResult`?

cc @wischli

pallets/liquidity-pools-gateway/src/weights.rs
```diff
@@ -122,4 +124,26 @@ impl WeightInfo for () {
...
+	fn start_pack_messages() -> Weight {
+		// TODO: BENCHMARK CORRECTLY
+		//
+		// NOTE: Reasonable weight taken from `PoolSystem::set_max_reserve`
+		//       This one has one read and one write for sure and possible one
+		//       read for `AdminOrigin`
+		Weight::from_parts(30_117_000, 5991)
+			.saturating_add(RocksDbWeight::get().reads(1))
+			.saturating_add(RocksDbWeight::get().writes(1))
+	}
```
lemunozm, line 137: TechDebt: We should factorize the weights of this file and use some auxiliary methods instead (similar to what William has done in its PR about freezing)

lemunozm:
Still missings UTs, will do later, but the PR is ready to be reviewed

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -356,16 +368,14 @@ pub mod pallet {
 		}
 
 		/// Process an inbound message.
-		#[pallet::weight(T::WeightInfo::process_msg())]
+		#[pallet::weight(T::WeightInfo::receive_message())]
 		#[pallet::call_index(5)]
-		pub fn process_msg(
+		pub fn receive_message(
```
lemunozm, line 373: NOTE. It's true that now we can receive batches, but from @cdamian changes in the queue, this method will no longer process messages, just queuing, which means that it will only take some more bytes as PoV. The current defensive weights are enough for that slight increment

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -547,55 +536,29 @@ pub mod pallet {
...
-
-			router_call_weight
-				.unwrap_or(Weight::from_parts(LP_DEFENSIVE_WEIGHT_REF_TIME, 0))
-				.saturating_add(Weight::from_parts(0, pov_weight))
+			(
+				result,
+				router_weight
+					.unwrap_or(Weight::from_parts(LP_DEFENSIVE_WEIGHT_REF_TIME, 0))
+					.saturating_add(read_weight)
+					.saturating_add(Weight::from_parts(0, serialized_len)),
```
lemunozm, line 560: NOTE we need here the serialized len instead of `max_decoded_len()` because the size used internally by the router is the size of the vector. The vector doesn't know about `Message` type

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Below, we're adding the defensive weights

cdamian: I would only return defensive weights when there's no way of finding the weight for an operation. In this case, the router does have some weight and I think we should use that, maybe alongside the defensive ones if we really wanna be extra-defensive?

pallets/liquidity-pools-gateway/src/tests.rs
```diff
@@ -1188,31 +1094,27 @@ mod message_processor_impl {
 					assert_eq!(mock_sender, expected_sender);
 					assert_eq!(mock_message, expected_message.serialize());
 
-					Err(DispatchErrorWithPostInfo {
+					Err(dbg!(DispatchErrorWithPostInfo {
```
cdamian, line 1097: Was this a one-time change that you used when debugging or should we sprinkle it around some more? :D

pallets/liquidity-pools-gateway/src/weights.rs
cdamian: Yup, we have a todo to use real weights for most of the gateway stuff.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -485,46 +465,48 @@ pub mod pallet {
...
-			*slice = new_slice;
-
-			Ok(res)
+		/// Start packing messages in a single message instead of enqueue
+		/// messages.
+		/// The message will be enqueued once `end_pack_messages()` is called.
+		#[pallet::weight(T::WeightInfo::start_pack_messages())]
+		#[pallet::call_index(9)]
+		pub fn start_pack_messages(origin: OriginFor<T>, destination: Domain) -> DispatchResult {
+			let sender = ensure_signed(origin)?;
```
cdamian, line 475: I have a couple of questions to make sure I understand the expected flow.

- Who is expected to call these packing extrinsics? 
  - From what I can tell, in the `OutboundMessageHandler` impl below, it would be a `sender` coming from LP. An LP `sender` can be some address that comes from EVM, right? If so, how does that address manage to call these extrinsics?
- How do we handle cases where the supposed caller errors out between start and end?
  - Should we maybe consider adding a way to manually end a batch for a particular (sender, destination) pair via root?

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Mmmm... I see. The defensive one seems to already shadow anything returned by the router. So, I would either only use the router one or only use the defensive one.

But not super strong opinion

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Very good questions!

> Who is expected to call these packing extrinsics?

The same as who call the extrinsics from `pallet-liquidity-pools`. It does not need to be an address from EVM. Just a normal account.

> How do we handle cases where the supposed caller errors out between start and end?

If some call between start and end fails, the user notice that. Note that each call in between is not yet processed by the `on_idle()`, so it fails just in that moment.

As a resume of how all of this work:
```rust
Gateway::start_pack_messages(ALICE)
LiquidityPools::add_pool(ALICE)
LiquidityPools::add_tranche(ALICE) // -> If it fails, ALICE will know it and can choose to try again with the correct params before calling end
Gateway::end_pack_messages(ALICE) // -> Now, the batch message is on the queue to be sent in the `on_idle()` function.
```

The above can be done in several blocks, but you can also batch all of them. If one fails, nothing is applied.

cdamian: OK, thank you for the explanation.

lemunozm:
There are still pending batching tests (UTs and ITs). But maybe we can start merging this to avoid more conflicts with the incoming Cosmin PRs, WDYT?

We can fix some clean-ups in a next PR along with the tests.

cc @wischli @cdamian

libs/traits/src/liquidity_pools.rs
```diff
@@ -23,30 +23,17 @@ use sp_std::vec::Vec;
 pub trait LPEncoding: Sized {
 	fn serialize(&self) -> Vec<u8>;
 	fn deserialize(input: &[u8]) -> Result<Self, DispatchError>;
-}
 
-#[cfg(any(test, feature = "std"))]
-pub mod test_util {
```
lemunozm, line 29: Moved to the gateway/mock.rs

pallets/liquidity-pools-gateway/src/mock.rs
```diff
@@ -20,6 +22,64 @@ pub const SOURCE_ADDRESS: [u8; LENGTH_SOURCE_ADDRESS] = [0u8; LENGTH_SOURCE_ADDR
...
 
+pub const MAX_PACKED_MESSAGES_ERR: &str = "packed limit error";
+pub const MAX_PACKED_MESSAGES: usize = 10;
+
+#[derive(Default, Debug, Eq, PartialEq, Clone, Encode, Decode, TypeInfo)]
+pub enum Message {
+	#[default]
+	Simple,
+	Pack(Vec<Message>),
+}
```
lemunozm, line 33: We should extend this with the `Proof` variant from your PR @cdamian

cdamian:
> There are still pending batching tests (UTs and ITs). But maybe we can start merging this to avoid more conflicts with the incoming Cosmin PRs, WDYT?
> 
> We can fix some clean-ups in a next PR along with the tests.
> 
> cc @wischli @cdamian

I'm OK with this. If we merge this, I will rebase my multi-router branch, then you can continue with the rest of the stuff here.

lemunozm:
@cdamian you will have quite conflicts but wanted to merge this ASAP to avoid even more 😅 

Tell me if I can help you in the rebase 🙏🏻, or with some tricky union of some part

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -533,10 +515,17 @@ pub mod pallet {
...
+			for submessage in message.unpack() {
+				count += 1;
+
+				if let Err(e) = T::InboundMessageHandler::handle(domain_address.clone(), submessage)
+				{
+					return (Err(e), weight.saturating_mul(count));
+				}
 			}
+
+			(Ok(()), weight.saturating_mul(count))
```
lemunozm, line 528: TODO. Debt to myself for the next PR. I think this could be an issue if we exceed the maximum during the batch processing. We should check that we can use all message weight before processing anything.

cdamian (approved):

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Fixed in https://github.com/centrifuge/centrifuge-chain/pull/1951

lemunozm:
@wischli @cdamian can we merge this to avoid conflicts with the cosmin part? 🙏🏻😃

pallets/liquidity-pools-gateway/src/lib.rs
wischli: Awesome explanation, thanks for that @lemunozm !

IMO, in general we can assume that `{start, end}_packing_messages` will only be called locally from a Substrate account because EVM accounts would call it on their EVM chain of choice such that we receive the LP batch message.

pallets/liquidity-pools-gateway/src/lib.rs
wischli: I would also use either the "real" or the defensive weight. If the weights of `send` can be assumed to be representative, then let's go with that. Else, ignore it and fallback to the defensive one for now.

wischli (approved):
LGTM!

wischli:
@lemunozm Feel free to press the button. We can merge #1946 afterwards.

PR #1951, opened 2024-08-07
LPv2: Fix weights with batch messages

lemunozm:
# Description

- Add `max_processing_weight()` method to `MessageProcessor` that allows to know upfront the worst case for processing a message.
- Check the worst case before processing.
- Adapt weight to work fine with batches

### TODOS

- [x] This should be merged first: https://github.com/centrifuge/centrifuge-chain/pull/1949
- [x] We need to add `on_idle` tests for the event pallet. Will do it in this PR later


pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -128,21 +121,25 @@ pub mod pallet {
 		/// are not reverted.
 		///   - an extra defensive weight is added in order to cover the weight
 		/// used when processing the message.
-		#[pallet::weight(T::WeightInfo::process_message())]
+		#[pallet::weight(MessageQueue::<T>::get(nonce)
+            .map(|msg| T::MessageProcessor::max_processing_weight(&msg))
+            .unwrap_or(T::DbWeight::get().reads(1)))]
 		#[pallet::call_index(0)]
-		pub fn process_message(origin: OriginFor<T>, nonce: T::MessageNonce) -> DispatchResult {
+		pub fn process_message(
```
lemunozm, line 128: We compute the max required weight for the specific message upfront. Later the extrinsic could be consume less

pallets/liquidity-pools-gateway/queue/src/lib.rs
```diff
@@ -207,7 +205,13 @@ pub mod pallet {
...
 
 			for (nonce, message) in MessageQueue::<T>::iter() {
-				processed_entries.push(nonce);
+				let remaining_weight = max_weight.saturating_sub(weight_used);
+				let next_weight = T::MessageProcessor::max_processing_weight(&message);
+
+				// We ensure we have still capacity in the block to process the message
+				if remaining_weight.all_gte(next_weight) {
+					break;
+				}
```
lemunozm, line 214: We need to be sure we have capacity in the block before we start to compute the message. We can not add the weight later, it could be too late (even more risky with a batch with a lot of sub-messages)

pallets/liquidity-pools-gateway/queue/src/weights.rs
```diff
@@ -1,30 +0,0 @@
...
-
-// Centrifuge is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU General Public License for more details.
-
-use cfg_primitives::LP_DEFENSIVE_WEIGHT;
-use frame_support::weights::Weight;
-
-pub trait WeightInfo {
```
lemunozm, line 17: No longer needed, they use `T::MessageProcessor::max_processing_weight()` instead

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -542,23 +542,12 @@ pub mod pallet {
...
 				Err(e) => (Err(e.error), e.post_info.actual_weight),
 			};
 
-			(
-				result,
-				router_weight
-					.unwrap_or(Weight::from_parts(LP_DEFENSIVE_WEIGHT_REF_TIME, 0))
-					.saturating_add(read_weight)
-					.saturating_add(Weight::from_parts(0, serialized_len)),
-			)
```
lemunozm, line 561: After investigating this. We delegate the processing weight to the router, which should take care of the bytes sent. In case we need some extra weight, it should be added to the router.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -629,5 +618,15 @@ pub mod pallet {
...
+
+		/// Process a message.
+		fn max_processing_weight(msg: &Self::Message) -> Weight {
+			match msg {
+				GatewayMessage::Inbound { message, .. } => {
+					LP_DEFENSIVE_WEIGHT.saturating_mul(message.unpack().len() as u64)
+				}
+				GatewayMessage::Outbound { .. } => LP_DEFENSIVE_WEIGHT, /* TODO can be much lower */
+			}
+		}
```
lemunozm, line 630: This is the required weight the block should be to start processing the message. I think the outbound case can be much less, but not sure how many. We can leave it as it is because later, the real weights coming from the router will be used.

lemunozm (commented):
I leave some comment justifiying the changes, which could be not easy to understand.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: BTW, We need to check the current `XcmDomain::overall_weight` used in production to be sure it's higher than our needs

cdamian (approved):
Would've loved to avoid some of the conflicts here but it's too late for that now. Let's get this in asap so I can do another rebase.

lemunozm:
You should not be too many, because the parts I modify you don't 🤞🏻. @wischli tell me if you have some comments regarding this post-mortem, to merge this ASAP for @cdamian

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -615,5 +604,15 @@ pub mod pallet {
 				} => Self::process_outbound_message(sender, destination, message),
 			}
 		}
+
+		/// Process a message.
+		fn max_processing_weight(msg: &Self::Message) -> Weight {
+			match msg {
+				GatewayMessage::Inbound { message, .. } => {
+					LP_DEFENSIVE_WEIGHT.saturating_mul(message.submessages().len() as u64)
```
wischli, line 612: Nit: For production code, I would like to avoid casting with `as` and rather use the defensive `saturated_into`.

wischli (commented):
Post mortem: Would love to use defensive conversion for production code (NIT). Otherwise clean and beautiful as always!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Only when converting from `usize` to `u64`, I use `as` because our machine target is no more than `u64` (are they machines longer than that?) 

But it seems reasonable; I'll add it to my mental list of changes!

PR #1947, opened 2024-08-06
feat: add freeze & unfreeze extrinsics

wischli:
# Description

* Adds freeze & unfreeze extrinsics
* Adds new `Poolrole` variant `FrozenTrancheInvestor`
  * Alternative: Extend existing `TrancheInvestor` permission with downside of having to migrate all existing storage entries
* Slightly improves docs of `pallet-liquidity-pools`
* Simplifies and normalizes `pallet-liqudiity-pools` defensive weights 

# Checklist:

- [x] I have added Rust doc comments to structs, enums, traits and functions
- [x] I have made corresponding changes to the documentation
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or that my feature works

PR #1953, opened 2024-08-09
Remove old routers and axelar relayer

lemunozm:
# Description

This PR remove the no longer used routers, cleaning the codebase for the incoming router refactor.

2 main concepts were removed:
- `EthereumXCM` and `AxelarXCM` routers.
- The Axelar relayer


pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -29,21 +29,18 @@
 use core::fmt::Debug;
 
 use cfg_primitives::{LP_DEFENSIVE_WEIGHT, LP_DEFENSIVE_WEIGHT_REF_TIME};
-use cfg_traits::{
```
lemunozm, line 32: @cdamian sorry, I've needed to remove some lines from the gateway, but because they are removed lines, I think the git conflicts are easy to solve, by just applying always these changes in these files

pallets/liquidity-pools-gateway/routers/src/lib.rs
```diff
@@ -219,127 +185,3 @@ pub struct FeeValues {
 	/// The gas limit used when executing the EVM call.
 	pub gas_limit: U256,
 }
-
-/// A generic router used for executing XCM calls.
-#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
-pub struct XCMRouter<T: pallet_xcm_transactor::Config> {
```
cdamian, line 225: Given that we "might" use xcm in the future, would it make sense to keep this here and minimally maintain this when doing polkadot version updates etc.? I'm thinking that re-doing an xcm setup from scratch might be even more troublesome later on.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: No problem, as long as there are no big changes to the actual gateway logic, we can manage. Thank you!

pallets/liquidity-pools-gateway/routers/src/lib.rs
lemunozm: In the incoming refactor, this `routers` module will disappear. We no longer can have "outbound" routers without an "inbound" router. In case we need this logic to build a bidirectional XCM router in the future, we can just look in this PR and recover that logic to build that new router.

Note that XCM is still used by the chain, just that not by the that router.

lemunozm:
Can we merge this, guys? @cdamian @wischli

wischli (approved):
More than 2000 lines less simplifies a lot! I was a bit sceptical about removing the `RelayerList` storage but seeing it is empty on Centrifuge Chain, seems fine! Let's merge this PR before #1947 because the latter is low prio.

wischli:
> Wait yours then

No! Sorry, I phrased my message poorly. Please merge your PR first.

PR #1955, opened 2024-08-09
LPv2: Add PalletAxelarRoutet

lemunozm:
# Description

- Add a new pallet called `pallet-axelar-router`, which is the union of the current `axelar-gateway-precompile` and `routers/axelar_evm`.
- Added some simplification to the gateway (comment below)

*Currently compiling as* `cargo check -p pallet-axelar-router`

### DIFF
--- a/libs/traits/src/liquidity_pools.rs
+++ b/libs/traits/src/liquidity_pools.rs
@@ -13,7 +13,9 @@
 
 use frame_support::{
 	dispatch::{DispatchResult, DispatchResultWithPostInfo},
+	traits::Get,
 	weights::Weight,
+	BoundedVec,
 };
 use sp_runtime::DispatchError;
 use sp_std::vec::Vec;
@@ -46,6 +48,37 @@ pub trait Router {
 
 	/// Send the message to the router's destination.
 	fn send(&self, sender: Self::Sender, message: Vec<u8>) -> DispatchResultWithPostInfo;
+}
+
+/// The behavior of an entity that can send messages
+pub trait MessageSender {
+	/// The originator of the message to be sent
+	type Origin;
+
+	///The destination of the message
+	type Destination;
+
+	/// Sends a message for origin to destination
+	fn send(
+		origin: Self::Origin,
+		destination: Self::Destination,
+		message: Vec<u8>,
+	) -> DispatchResultWithPostInfo;
+}
+
+/// The behavior of an entity that can receive messages
+pub trait MessageReceiver {
+	/// The maximum lenght for a message the implementor is able to receive.
+	type MaxEncodedLen: Get<u32>;
+
+	/// The originator of the message to be sent
+	type Origin;
+
+	/// Sends a message for origin to destination
+	fn receive(
+		origin: Self::Origin,
+		message: BoundedVec<u8, Self::MaxEncodedLen>,
+	) -> DispatchResult;
 }
 
 /// The trait required for queueing messages.
--- a/pallets/axelar-router/src/lib.rs
+++ b/pallets/axelar-router/src/lib.rs
@@ -1,0 +1,402 @@
+use cfg_traits::{
+	ethereum::EthereumTransactor,
+	liquidity_pools::{MessageReceiver, MessageSender},
+	PreConditions,
+};
+use cfg_types::{domain_address::DomainAddress, EVMChainId};
+use ethabi::{Contract, Function, Param, ParamType, Token};
+use fp_evm::{ExitError, PrecompileFailure, PrecompileHandle};
+use frame_support::{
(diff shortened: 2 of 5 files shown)


pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+				chain_id_hash,
+				chain_id,
+			});
+
+			Ok(())
+		}
+
+		#[pallet::weight(Pallet::<T>::weight_for_set_method())]
+		#[pallet::call_index(2)]
+		pub fn set_outbound_config(
```
lemunozm, line 256: The router is configured here by `chain_id` instead of domain because a router can never be a Centrifuge domain. The gateway will no longer configure routers, will only handle the routers id.

pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+			let origin = DomainAddress::EVM(source_chain_id, source_address_bytes);
+
+			let msg = BoundedVec::<u8, <T::Gateway as MessageReceiver>::MaxEncodedLen>::try_from(
+				payload.as_bytes().to_vec(),
+			)
+			.map_err(|_| PrecompileFailure::Error {
+				exit_status: ExitError::Other("payload conversion".into()),
+			})?;
+
+			T::Gateway::receive(origin, msg).map_err(|e| TryDispatchError::Substrate(e).into())
```
lemunozm, line 339: Probably here we should add a `router_id` parameter or append to the origin

pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+				}
+			);
+
+			let source_chain_id = SourceChain::<T>::get(BlakeTwo256::hash(source_chain.as_bytes()))
+				.ok_or(PrecompileFailure::Error {
+					exit_status: ExitError::Other("converter for source not found".into()),
+				})?;
+
+			let source_address_bytes =
+				cfg_utils::decode_var_source::<EXPECTED_SOURCE_ADDRESS_SIZE>(
```
lemunozm, line 323: I've simplified this by removing the `SourceConverter` type from the original axelar precompile pallet. 

I did so, because the `EXPECTED_SOURCE_ADDRESS_SIZE` is 20, which means that we always parse the 20bytes address. Makes sense because we should not expect a Centrifuge address call coming from Axelar.

pallets/axelar-router/src/lib.rs
cdamian: Yup, atm I have a `T::Hash` for inbound routers but we can change to whatever.

libs/traits/src/liquidity_pools.rs
```diff
@@ -48,6 +50,37 @@ pub trait Router {
...
+
+	/// The originator of the message to be sent
+	type Origin;
+
+	/// Sends a message for origin to destination
+	fn receive(
+		origin: Self::Origin,
+		message: BoundedVec<u8, Self::MaxEncodedLen>,
+	) -> DispatchResult;
+}
```
lemunozm, line 82: @Cosmin, this will be the new interface. Basically, the routers are no longer instances. They are pallet, but each pallet can have different configurations in they want.

I'm not fully sure yet, but my intention is that the Self::Origin will be a tuple `(RouterId, Domain)` or similar. Still thinking about this, but somehow though the receive()/send() method you could determine the router.

Hopefully, it will require few changes in the gateway.

Also the router storage and configuration could be removed from the gateway.

cdamian: > Also the router storage and configuration could be removed from the gateway.

OK, but then we need a way to somehow still "add" the routers to the gateway. Maybe add a tuple of all the available router pallets to the gateway config or something.

lemunozm:
Regarding this implementation, I'm still not sure if the whole axelar should be inside this new router pallet, because Axelar is supposed to be connected to other chains, so maybe the precompile logic itself is shared among different routers.

libs/traits/src/liquidity_pools.rs
lemunozm: Yeah, the way would be by adding a list of router ids

lemunozm: And from the gateway perspective, only router ids exist

cdamian: OK, but then how would the gateway call router X to send an outbound message?

lemunozm: `send(router_id, domain, message)` something like that

lemunozm:
Close in favor of: https://github.com/centrifuge/centrifuge-chain/pull/1956

PR #1956, opened 2024-08-12
Refactor: Bidirectional routers

lemunozm:
# Description

Basic design to support bidirectional routers. From our perspective, we do not have a router entity, just inbound and outbound methods (which are still called routers), at the runtime level, we combine those methods under the same `RouterId`, later the gateway just handles those `RouterId`s transparently.

### Changes
- Added a new pallet: `pallet-axelar-evm-router` which is basically the same implementation as `routers/axelar-evm` but as a pallet, allowing to create configurations in the own pallet. It only supports outbound messages as before.

- Slightly modidied `axelar-precompile` to be able to call the new `MessageReceiver` method.

- **The core of this change**. Add a `common/routing.rs` module where our inbound and outbound methods for sending messages are joined under the same router Id.


runtime/common/src/routing.rs
```diff
@@ -0,0 +1,59 @@
...
+	BoundedVec,
+};
+use pallet_axelar_evm_router::AxelarEvmKind;
+use sp_std::marker::PhantomData;
+
+/// Identification of the router where the message is sent and received
+pub enum RouterId {
+	/// The message must be sent/received by EVM using axelar
+	Axelar(AxelarKind),
+}
```
lemunozm, line 15: We can not base our decision in the `Domain`, because we can have a router that connect to the same `Domain::Evm` by Axelar or by another kind of middleware. We can not use the `Domain` to distinguish the routers

lemunozm (commented):
I will need to update the specs with these changes a bit.

runtime/common/src/routing.rs
cdamian: Would it make more sense to have it in `libs` and then use this type in the gateway? Or do we want to specify the `RouterId` as an associated type in the gateway pallet config etc?

lemunozm: From the gateway perspective, this type is unknown. It is a runtime configuration type.

If the pallets use directly this type, then the pallet can only be used in runtimes with that type, which would mean runtimes with the same configured router. Also, each pallet would know about other routers ids, and some weird things.

Yeah, as you said, it should be passed as a config type in the gateway

lemunozm:
Not fully ready, still WIP. I want to find a way to join the inbound and outbound crates more closely.

runtime/common/src/routing.rs
cdamian: Cool, thank you very much!

lemunozm:
Will close to submit a better solution

PR #1958, opened 2024-08-12
Support for bidirectional routers

lemunozm:
# Description

Support configuring routers for inbound and outbound directions, based on a `RouterId` that identify it for both directions

## Main changes of this PR:

- Added `pallet-axelar-router` which connects in both directions with Axelar. It allows the configuration of different domains under Axelar for inbound and outbound messages. It is a mix of `axelar-gateway-preocompile` and `routers/axelar_evm` which will be removed.

- Added `RouterId`. It is like `Domain` but with information about by where the message needs to be sent.

- Added `MessageSender` trait and `MessageReceiver`. The middleware parameter represents by where the message must be sent. Normally, it will be the `RouterId` but not mandatory, it could be any subset of `RouterId`, i.e: `AxelarId`.

- Added a `RouterSupport` trait (name can change), to allow mapping any `Domain` in a list of supported routers that can lead the message to that `Domain`. Will be used by the gateway to obtain the `RouterId` list for a `Domain`.

- Added `RouterDispatcher` in runtime level to redirect a message to the correct `router_id`.

## Use case we want to support

Suppose gateway is configured with the following (dynamically configured by an operator):
  - `AxelarXcm`
  - `AxelarEvm`
  - `SnowbridgeEvm`
  
And the chain offers the following routers (statically configured in the runtime):
  - `AxelarXcm`
  - `AxelarEvm`
  - `SnowbridgeEvm`
  - `OtherEvm`
  
If a message with `Domain::Evm` needs to be sent, it must be sent only to `AxelarEvm` and `SnowbridgeEvm`.
  - `AxelarXcm` will not used because that router is not retrieved by `for_domain(Domain::Evm)`
  - `OtherEvm` will not used because it's not configured in the gateway


pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+
+	/// The `BlakeTwo256` hash of the target contract code.
+	///
+	/// This is used during router initialization to ensure that the correct
+	/// contract code is used.
+	pub target_contract_hash: H256,
+
+	/// The values used when executing the EVM call.
+	pub fee_values: FeeValues,
+}
```
lemunozm, line 74: Some fields have been merged with the previous `axelar_precompile` configuration. For example, `set_gateway()` previously set in `axelar_precompile` was the same as `liquidity_pools_contract_address` for the outbound router.

Anyways, maybe we need to polish it a bit when updating the ITs.

libs/traits/src/liquidity_pools.rs
```diff
@@ -37,6 +39,7 @@ pub trait LPEncoding: Sized {
 }
 
 /// The trait required for sending outbound messages.
+/// TODO: remove this
```
wischli, line 40: Do we really want to remove the `Router` trait or is this comment outdated?

pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Domain not found.
+		DomainNotFound,
+
+		/// Router not found.
+		RouterNotFound,
+
+		/// Emit when the evm account code is not registered
+		ContractCodeNotMatch,
```
wischli, line 147: Nit: To align with other error naming scheme

```suggestion
		ContractCodeMismatch,
```

pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+	}
+
+	impl<T: Config> Pallet<T> {
+		fn weight_for_set_method() -> Weight {
+			Weight::from_parts(50_000_000, 256).saturating_add(RocksDbWeight::get().writes(2))
```
wischli, line 152: Can you elaborate how those values were chosen?

runtime/common/src/routing.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+impl RouterSupport<Domain> for RouterId {
+	fn for_domain(domain: Domain) -> Vec<Self> {
+		match domain {
+			Domain::EVM(chain_id) => vec![RouterId::Axelar(AxelarId::Evm(chain_id))],
+			Domain::Centrifuge => vec![],
+		}
+	}
+}
```
wischli, line 30: I am not sure we can safely map any EVM domain to the Axelar router. AFAICT, we need to set up a storage for that in the future. For instance, there exists Gnosis chain which will not have a direct router to Centrifuge Chain but use the unimplemented message relaying in which we wrap forwarded LP messages [like this](https://centrifuge.hackmd.io/TPH6TWh7SJat_pT0Fyty7A?both#Centrifuge-to-EVM). We want to include message relaying into LPV2.

______

In the following, I am thinking out loud how to extend your design for message relaying.

We need to be able to determine whether for a given domain, a forwarder is set. Though I think your design is fully compatible if we just add a `StorageMap<Domain, Path<Domain>>` where

```rust
	enum Path<Domain> {
		Domain(Domain),
		Forwarder {
			forwarding_domain: Domain,
			forwarding_contract: H160
		}
	}

	fn wrap_into_forwarded_msg(message: Vec<u8>, source_chain: Domain, forwarding_contract: H160) -> Result<Vec<u8>, DispatchError>;

```

and extend the `Middleware` of `pallets-axelar-router` to use `Path` instead of `AxelarId`

```diff
	impl<T: Config> MessageSender for Pallet<T> {
-		type Middleware = AxelarId;
+		type Middleware = Path<Domain>;
		type Origin = DomainAddress;

		fn send(
-			axelar_id: Path<Domain>,
+			path: Path<Domain>,
			origin: Self::Origin,
			message: Vec<u8>,
		) -> DispatchResultWithPostInfo {
+			let axelar_ids = match path {
				Domain(domain) => domain.for_domain(),
				Forwarder { forwarding_domain, .. } => forwarding_domain.for_domain()
			};
+			// NOTE: Not sure how to commit to an id if vec length > 1, for simplicity here taken first element
+			let axelar_id = axelar_ids.first().ok_or(Error::<T>::PathNotFound)?;

			let chain_name =
				ChainNameById::<T>::get(axelar_id).ok_or(Error::<T>::DomainNotFound)?;
			let config = Configuration::<T>::get(&chain_name).ok_or(Error::<T>::RouterNotFound)?;

			match config.domain {
				DomainConfig::Evm(evm_config) => {
					let sender_evm_address = H160::from_slice(&origin.address());

					let message = wrap_into_axelar_msg(
						message,
						chain_name.into_inner(),
						config.liquidity_pools_contract_address,
					)
					.map_err(DispatchError::Other)?;

+					let message = match path {
						Forwarder { forwarding_domain, forwarding_contract } => wrap_into_forwarded_msg(message, Domain::Centrifuge, forwarding_contract),
						_ => Ok(message),
					}?;

					T::Transactor::call(
						sender_evm_address,
						evm_config.target_contract_address,
						message.as_slice(),
						evm_config.fee_values.value,
						evm_config.fee_values.gas_price,
						evm_config.fee_values.gas_limit,
					)
				}
			}
		}
	}
```

Do you think this draft is somewhat reasonable based on your design proposal?

wischli (commented):
Overall, the design looks really good! I am mostly curious on how to extend this to support message relaying, see my long comment.

pallets/axelar-router/src/lib.rs
cdamian: I don't think we need this to begin with, the set method is pretty straightforward and doesn't do any extra processing.

pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(Pallet::<T>::weight_for_set_method())]
+		#[pallet::call_index(2)]
```
cdamian, line 159: `#[pallet::call_index(0)]` :D

pallets/axelar-router/src/tests.rs
```diff
@@ -0,0 +1,8 @@
+use crate::mock::*;
+
+#[test]
+fn configure_router() {
+	new_test_ext().execute_with(|| {
+		//TODO
```
cdamian, line 6: Are you OK with adding unit/integration tests after we eventually merge this? Would love to have these changes in the gateway.

cdamian:
I like how this is implemented, thank you for taking care @lemunozm!

libs/traits/src/liquidity_pools.rs
lemunozm: We need it 👍🏻 along with the previous axelar precompile and routers

pallets/axelar-router/src/lib.rs
lemunozm: Ideally, it should be benchmarked, but I'm not sure if I will have time to do it. 

There are mostly 2 writes, which should not exceed 256 bytes. The check will be just a compare + hash + storage read. `50_000_000` should be enough for that.

runtime/common/src/routing.rs
lemunozm: Trying to understand this topic.

If what you need is, for a domain, knowing if there is a forwarded already set; why not retrieve that information in `RouterDomain::for_domain()`.

I think I get the point in the code you've shared but why the forwarder action needs to be in `pallet-axelar-precompile`. If we end up with 5 routers similars to the axelar one, we would like that section in a common part, isn't it? 

Why not implement this in `RouterDispatch`? if you need to store the map, maybe `RouterDispatch` can be a super simple pallet with just that map of forwarding info. That entity is in charge of determining the router, which seems similar to what you need.

pallets/axelar-router/src/lib.rs
lemunozm: Increase to 512 to be more paranoic

pallets/axelar-router/src/tests.rs
lemunozm: Will try to have it for tomorrow @cdamian 🤞🏻

runtime/common/src/routing.rs
wischli: > I think I get the point in the code you've shared but why the forwarder action needs to be in `pallet-axelar-precompile`. If we end up with 5 routers similars to the axelar one, we would like that section in a common part, isn't it?
> 
> Why not implement this in `RouterDispatch`? if you need to store the map, maybe `RouterDispatch` can be a super simple pallet with just that map of forwarding info. That entity is in charge of determining the router, which seems similar to what you need.

Yeah that makes perfectly sense. Seems like I can very well align that feature with the new design then! I hope I can start implementing that tomorrow based on either this branch or #1948

lemunozm: Super good news!

runtime/integration-tests/src/cases.rs
```diff
@@ -8,10 +8,10 @@ mod investments;
 mod liquidity_pools;
 mod liquidity_pools_gateway_queue;
 mod loans;
-mod lp;
```
lemunozm, line 11: @William TODO

pallets/axelar-router/src/tests.rs
lemunozm, line 6: TODO: pending

pallets/liquidity-pools-gateway/src/tests.rs
```diff
@@ -1,3 +1,4 @@
+/*
```
lemunozm, line 1: TODO: pending

runtime/integration-tests/src/cases.rs
```diff
@@ -8,10 +8,10 @@ mod investments;
...
+//mod lp;
 mod oracles;
-mod precompile;
+//mod precompile;
```
lemunozm, line 13: TODO: pending

runtime/integration-tests/src/cases.rs
```diff
@@ -8,10 +8,10 @@ mod investments;
...
 mod proxy;
 mod restricted_transfers;
-mod routers;
+//mod routers;
```
lemunozm, line 16: TODO: pending

lemunozm (commented):
List of pendings todos

lemunozm:
If nothing wrong CI should be green with the commented sections in the code

runtime/development/src/lib.rs
```diff
@@ -2219,7 +2225,7 @@ construct_runtime!(
 		BaseFee: pallet_base_fee::{Pallet, Call, Config<T>, Storage, Event} = 162,
 		Ethereum: pallet_ethereum::{Pallet, Config<T>, Call, Storage, Event, Origin} = 163,
 		EthereumTransaction: pallet_ethereum_transaction::{Pallet, Storage} = 164,
-		LiquidityPoolsAxelarGateway: axelar_gateway_precompile::{Pallet, Call, Storage, Event<T>} = 165,
+		AxelarRouter: pallet_axelar_router::{Pallet, Call, Storage, Event<T>} = 165,
```
lemunozm, line 2228: Reusing index, so we should remove the previous Axelar state on-chain.

Also, another TODO: we should remove the code from axelar and routers crates

libs/traits/src/liquidity_pools.rs
lemunozm: We need it to remove I wanted to say 😆

pallets/axelar-router/src/lib.rs
lemunozm: For the curious one, the size of the longer configuration right now is 208 bytes.

libs/traits/src/liquidity_pools.rs
```diff
@@ -36,16 +33,38 @@ pub trait LPEncoding: Sized {
 	fn empty() -> Self;
 }
 
-/// The trait required for sending outbound messages.
-pub trait Router {
-	/// The sender type of the outbound message.
-	type Sender;
+pub trait RouterSupport<Domain>: Sized {
```
cdamian, line 36: One nit though - maybe we can adjust this a bit so we don't have to implement it for the `RouterId`, which also has the potential adding more to the confusion.

How about something like:

```
pub trait RouterProvider<Domain>: Sized {
	type RouterId;
	
	/// Returns a list of router IDs supported for the given domain.
	fn for_domain(domain: Domain) -> Vec<Self::RouterId>;
}
```

Then we can have a simple enum as a router ID, and a different type for getting routers for a domain, all in runtime common.

WDYT @lemunozm ?

runtime/development/src/lib.rs
wischli: I have added it to the post-audit TODO list.

libs/traits/src/liquidity_pools.rs
lemunozm: I like it! I was not comfortable with the correct way. You can change this in your PR, or we can do it after merging everything, as a clean PR. As you want.

lemunozm: Maybe `for_domain()` can also be renamed to be more specific: `routers_for_domain()`

cdamian: I'll add these in my PR to have a clean state there. Thank you!

runtime/integration-tests/src/cases.rs
lemunozm: Done

runtime/integration-tests/src/cases.rs
lemunozm: Done

pallets/axelar-router/src/tests.rs
lemunozm: Done

pallets/liquidity-pools-gateway/src/tests.rs
lemunozm: @cdamian, are you in charge of this TODO, right?

cdamian: Yup, they're green as of now in the [multi-router PR](https://github.com/centrifuge/centrifuge-chain/pull/1961).

runtime/integration-tests/src/cases.rs
lemunozm: I will take care of this

lemunozm: It was pretty straighforward: [6c1b30f](https://github.com/centrifuge/centrifuge-chain/pull/1958/commits/6c1b30f1fe51c4e824da16b0ad0a8cafa274ab12)

pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+		ConfigSet {
+			name: ChainName,
+			config: Box<AxelarConfig>,
+		},
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Emit when the router configuration is not found.
+		RouterNotFound,
```
cdamian, line 159: Nit - `RouterConfigurationNotFound`.

pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+				DomainConfig::Evm(evm_config) => {
+					ensure!(
+						T::EvmAccountCodeChecker::check((
+							evm_config.target_contract_address,
+							evm_config.target_contract_hash,
+						)),
+						Error::<T>::ContractCodeMismatch
+					);
+
+					ChainNameById::<T>::insert(
```
cdamian, line 195: Do we really need a separate storage for the chain name? Why not make that part of the `AxelarConfig`?

pallets/axelar-router/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+		}
+	}
+}
+
+/// Encodes the provided message into the format required for submitting it
+/// to the Axelar contract which in turn calls the LiquidityPools
+/// contract with the serialized LP message as `payload`.
+///
+/// Axelar contract call:
+/// <https://github.com/axelarnetwork/axelar-cgp-solidity/blob/v4.3.2/contracts/AxelarGateway.sol#L78>
```
cdamian, line 355: Nit - maybe we need to update these links? cc @hieronx

pallets/axelar-router/src/lib.rs
lemunozm: Makes sense!

pallets/axelar-router/src/lib.rs
lemunozm: We need it because for sending we index by `AxelarId` but for receiving we index by `ChainName` (basically, the latest is an Axelar requirement in their API). So we have two identifications that redirect to the same config.

cdamian: Ah, I looked at that but didn't notice it, maybe I need to start drinking coffee ^^. My bad, sorry for the noise.

lemunozm: No worries! good question anyways 🫶

pallets/axelar-router/src/lib.rs
wischli: API still looks the same for the most recent v6.3.1, only the line changed: https://github.com/axelarnetwork/axelar-cgp-solidity/blob/v6.3.1/contracts/AxelarGateway.sol#L150-L156

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -93,15 +93,11 @@ pub mod pallet {
...
-		/// The message router type that is stored for each domain.
-		type Router: DomainRouter<Sender = Self::AccountId>
-			+ Clone
-			+ Debug
-			+ MaxEncodedLen
-			+ TypeInfo
-			+ FullCodec
-			+ EncodeLike
-			+ PartialEq;
+		/// The target of of the messages comming from this chain
```
wischli, line 96: Nit
```suggestion
		/// The target of the messages coming from this chain
```

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -143,13 +136,6 @@ pub mod pallet {
 		},
 	}
 
-	/// Storage for domain routers.
-	///
-	/// This can only be set by an admin.
-	#[pallet::storage]
-	#[pallet::getter(fn domain_routers)]
-	pub type DomainRouters<T: Config> = StorageMap<_, Blake2_128Concat, Domain, T::Router>;
```
wischli, line 151: I'm happy I have just finished writing the migration 🥲. Though I can imagine I can adapt it rather quickly to the final outcome of all the refactors.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -209,27 +192,6 @@ pub mod pallet {
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
-		/// Set a domain's router,
-		#[pallet::weight(T::WeightInfo::set_domain_router())]
-		#[pallet::call_index(0)]
-		pub fn set_domain_router(
```
wischli, line 215: IIUC, this is replaced with `pallet_axelar_router::Pallet::<T>::set_config`?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -379,28 +332,32 @@ pub mod pallet {
...
 			domain: Domain,
 			message: T::Message,
 		) -> (DispatchResult, Weight) {
-			let read_weight = T::DbWeight::get().reads(1);
+			let router_ids = T::RouterId::for_domain(domain);
 
-			let Some(router) = DomainRouters::<T>::get(domain) else {
-				return (Err(Error::<T>::RouterNotFound.into()), read_weight);
-			};
+			// TODO handle router ids logic
```
wischli, line 341: I suppose this will be done in #1961?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -379,28 +332,32 @@ pub mod pallet {
...
-			let (result, router_weight) = match router.send(sender, message.serialize()) {
-				Ok(dispatch_info) => (Ok(()), dispatch_info.actual_weight),
-				Err(e) => (Err(e.error), e.post_info.actual_weight),
-			};
+			let mut count = 0;
+			let bytes = message.serialize();
 
-			(result, router_weight.unwrap_or(read_weight))
+			for router_id in router_ids {
+				count += 1;
```
wischli, line 347: Nit: Use safe math
```suggestion
				count.saturating_accrue(1);
```

runtime/common/src/routing.rs
```diff
@@ -0,0 +1,73 @@
...
+	pallet_prelude::{Decode, Encode, MaxEncodedLen, TypeInfo},
+};
+use pallet_axelar_router::AxelarId;
+use sp_core::{H160, H256};
+use sp_runtime::traits::{BlakeTwo256, Hash};
+use sp_std::{marker::PhantomData, vec, vec::Vec};
+
+/// Identification of the router where the message is sent and received
+/// RouterId is more specific than Domain, because RouterId also identify by
+/// where the message is sent/received
```
wischli, line 17: Nit: Make it a bit clearer.
```suggestion
/// Identification of the router where the message is sent and received.
/// 
/// NOTE: RouterId is more specific than Domain, because RouterId also identifies
/// the source and destination of the message.
```

runtime/integration-tests/src/cases.rs
wischli: Cant find the commit you have linked but I am happy to hear it wasn't a nightmare!

runtime/integration-tests/src/cases/lp/setup_lp.rs
```diff
@@ -67,49 +64,31 @@ pub fn setup<T: Runtime, F: FnOnce(&mut <RuntimeEnv<T> as EnvEvmExtension<T>>::E
...
-			RawOrigin::Root.into(),
-			evm.deployed(names::ADAPTER).address()
-		));
-
-		assert_ok!(axelar_gateway_precompile::Pallet::<T>::set_converter(
+		assert_ok!(pallet_liquidity_pools_gateway::Pallet::<T>::add_instance(
 			RawOrigin::Root.into(),
-			BlakeTwo256::hash(EVM_DOMAIN_STR.as_bytes()),
-			SourceConverter::new(EVM_DOMAIN),
+			DomainAddress::evm(EVM_DOMAIN_CHAIN_ID, EVM_LP_INSTANCE)
```
wischli, line 91: I love seeing the reduced amount of setup steps! Now it's only two calls that need to be made 🎉

runtime/integration-tests/src/cases/restricted_transfers.rs
```diff
@@ -425,22 +423,6 @@ mod eth_address {
 				curr_contract,
 				curr.val(TRANSFER),
 			));
-
-			assert_ok!(
-				pallet_liquidity_pools_gateway_queue::Pallet::<T>::process_message(
-					RawOrigin::Signed(Keyring::Alice.into()).into(),
-					MessageNonceStore::<T>::get(),
-				)
-			);
```
wischli, line 434: Shouldn't we still process the message and check for an expected error?

wischli (commented):
Super straight-forward and clear - awesome work! A few nits and one question regarding a test.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Awesome! Thanks William

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Correct!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Thanks for the reminder! I think @cdamian already fixed in its PR, if not could you Cosmin?

runtime/common/src/routing.rs
lemunozm: Thanks for always polishing the docs

runtime/integration-tests/src/cases/restricted_transfers.rs
lemunozm: From the point of view of the restriction we're trying to check if `pallet_liquidity_pools::Pallet::<T>::transfer()` passes. If it does, then the restriction passed.

Basically, in those tests, we want to check the restriction is well configured by the runtime, not that the transfer itself is done.

The end-to-end tests are in `routers.rs`

wischli: Thanks for elaborating on that, makes sense!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: I think Cosmin can fix this if it has not changed in its PR. To avoid conflicts with your PR @cdamian

runtime/common/src/routing.rs
lemunozm: Just a though, not sure if RouterId identifies the source/destination, it identifies the "channel" to reach that destination

lemunozm: Not sure the exact word instead of channel, maybe the medium? or, "how the message is sent/received"

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: I think I already fixed this.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Sure.

runtime/common/src/routing.rs
lemunozm: In other words, to explain better, the `Domain` represents the source and destination, but `RouterId` also represents how to reach them

runtime/integration-tests/src/cases.rs
lemunozm: I rebase changed the hash 😅

pallets/axelar-router/src/lib.rs
lemunozm: Done

runtime/common/src/routing.rs
lemunozm: I've rewrote over your changes the comment, tell me if now it makes more sense

wischli: Perfect!

wischli (approved):
LS(uper)M(ega)GTM!

cdamian (approved):
The multi router PR is based on this and I'm super happy with everything so far.

Awesome job @lemunozm!

lemunozm:
Thanks for your reviews! Should be merged once CI passes 🤞🏻

PR #1959, opened 2024-08-12
Ft/gateway changes for routers

lemunozm:
# Description

Required changes for pallet-liquidity-pools-gateway to works with https://github.com/centrifuge/centrifuge-chain/pull/1958

(currently compiling for the gateway)

### DIFF
--- a/pallets/liquidity-pools-gateway/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/src/lib.rs
@@ -30,8 +30,8 @@ use core::fmt::Debug;
 
 use cfg_primitives::LP_DEFENSIVE_WEIGHT;
 use cfg_traits::liquidity_pools::{
-	InboundMessageHandler, LPEncoding, MessageProcessor, MessageQueue, OutboundMessageHandler,
-	Router as DomainRouter,
+	InboundMessageHandler, LPEncoding, MessageProcessor, MessageQueue, MessageReceiver,
+	MessageSender, OutboundMessageHandler, RouterSupport,
 };
 use cfg_types::domain_address::{Domain, DomainAddress};
 use frame_support::{dispatch::DispatchResult, pallet_prelude::*};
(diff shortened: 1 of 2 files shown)


lemunozm:
@Cosmin, this are the only required changes in the gateway to works with this PR: https://github.com/centrifuge/centrifuge-chain/pull/1958

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -118,18 +114,15 @@ pub mod pallet {
 		/// The sender account that will be used in the OutboundQueue
 		/// implementation.
 		#[pallet::constant]
-		type Sender: Get<Self::AccountId>;
+		type Sender: Get<DomainAddress>;
```
lemunozm, line 117: If instead of `AccountId` we can use `DomainAddress::centrifuge(AccountId)`, which is isometric, it would work better with the internals for sending messages.

pallets/liquidity-pools-gateway/src/message.rs
```diff
@@ -3,19 +3,19 @@ use frame_support::pallet_prelude::{Decode, Encode, MaxEncodedLen, TypeInfo};
...
 #[derive(Debug, Encode, Decode, Clone, Eq, MaxEncodedLen, PartialEq, TypeInfo)]
-pub enum GatewayMessage<AccountId, Message> {
+pub enum GatewayMessage<Message> {
 	Inbound {
 		domain_address: DomainAddress,
 		message: Message,
 	},
 	Outbound {
-		sender: AccountId,
+		sender: DomainAddress,
```
lemunozm, line 12: Normally it will be `DomainAddress::Centrifuge(AccountId)`, but if in the future we want to relay some message we should be able to storage a different origin

cdamian:
> @cosmin, this are the only required changes in the gateway to works with this PR: #1958

OK, but we can close this and I can do these changes on my branch.

lemunozm:
Yeah, that was my intention 👍🏻, just to show you the required changes

cdamian:
> Yeah, that was my intention 👍🏻, just to show you the required changes

Absolute legend <3

pallets/liquidity-pools-gateway/src/message.rs
wischli: The forwarding wrap includes the source domain as well as the forwarding contract. But that needs to be part of the message such that EVM can distinguish between forwarded and normal messages.

lemunozm: Makes sense

lemunozm:
No longer a legend 🥲, I need to merge this to be able to compile the other one

cdamian (approved):

PR #1961, opened 2024-08-13
Feat/lp v2 gateway multi router

cdamian:
# Description

This PR adds support for multiple routers in the LP gateway.

### Extrinsics

- Added `set_routers` extrinsic that sets the set of valid routers for both inbound and outbound messages.
	- Apart from populating the according `Routers` storage, a new session ID is created to keep track of incoming messages or proofs. 
- Added `execute_message_recovery` extrinsic that is used to manually increase the proof count for a particular message proof and router ID, and execute the message if the required proof count is met.

#### Inbound
For every inbound message that is received via the `MessageReceiver` implementation, we check the `Allowlist` storage to confirm that the origin is allowed to submit message, if successful, we queue an inbound gateway message that contains the necessary information for processing i.e. origin address, message and router identifier.

For every inbound gateway message we create an `InboundEntry` that specifies whether the message is a message type or proof type. 

After validation, this `InboundEntry` is then added to storage, if a previous entry of this type is found, the expected counts are updated accordingly - for a message type, the total expected proof count is increased by the value required for that domain, similarly, for a proof type, the proof count is increased by 1.

If the `PendingInboundEntries` storage contains one message entry from the first router, and the required proof entries from each of the remaining routers, the counts are decreased and the message is forwarded to the `InboundMessageHandler` for execution.

#### Outbound
For every outbound message received from the called of the `OutboundMessageHandler::handle` we retrieve all routers for the destination `Domain`.

The queueing is done using the `LP Gateway Queue` as follows:
- the message is queued as-is using the first router.
- the proof for the message is queued using the remaining routers.

Upon eventual receipt of the queued messages/proofs from the `LP Gateway Queue`, they are sent for further processing to the `MessageSender`.

---

## Other Notable Changes
- The `LPEncoding` trait now has 2 more methods for retrieving a message proof and converting a message type to a proof type.
- The `RouterSupport` trait was slightly adjusted to better fit with the requirement of providing a statically defined list of `RouterId`s for a specific domain.

# Checklist:


pallets/liquidity-pools-gateway/routers/src/lib.rs
```diff
@@ -76,6 +76,7 @@ where
 	OriginFor<T>:
 		From<pallet_ethereum::Origin> + Into<Result<pallet_ethereum::Origin, OriginFor<T>>>,
```
cdamian, line 77: TODO: This file should be removed.

pallets/liquidity-pools-gateway/routers/src/routers/axelar_evm.rs
```diff
@@ -21,7 +21,8 @@ use scale_info::{
 	prelude::{format, string::String},
```
cdamian, line 21: TODO: This file should be removed.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -134,8 +164,32 @@ pub mod pallet {
...
+
+		/// Message recovery was executed.
+		MessageRecoveryExecuted {
+			domain: Domain,
+			proof: Proof,
+			router_id: T::RouterId,
+		},
 	}
 
+	// TODO(cdamian): Add migration to clear this storage.
```
cdamian, line 176: Marked for removal after the audit.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -188,10 +272,77 @@ pub mod pallet {
...
+		pub fn set_domain_routers(
+			origin: OriginFor<T>,
+			domain: Domain,
+			router_ids: BoundedVec<T::RouterId, T::MaxRouterCount>,
+		) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
+
+			ensure!(domain != Domain::Centrifuge, Error::<T>::DomainNotSupported);
+
+			//TODO(cdamian): Outbound -  Call router.init() for each router?
```
cdamian, line 321: @lemunozm Just to be ultra-sure. There's no `init` logic in the refactored routers right?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -305,65 +456,58 @@ pub mod pallet {
...
-		/// weight for these operations in the `DispatchResultWithPostInfo`.
-		fn process_outbound_message(
-			sender: DomainAddress,
+		/// Manually increase the proof count for a particular message and
+		/// executes it if the required count is reached.
+		///
+		/// Can only be called by `AdminOrigin`.
+		#[pallet::weight(T::WeightInfo::execute_message_recovery())]
+		#[pallet::call_index(11)]
+		pub fn execute_message_recovery(
```
cdamian, line 466: @hieronx Given that with the new LP logic, routers are now stored under domains, we need an extra arg here. Is this OK?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -428,20 +574,19 @@ pub mod pallet {
 		type Origin = DomainAddress;
 
 		fn receive(
-			_router_id: T::RouterId,
+			router_id: T::RouterId,
 			origin_address: DomainAddress,
 			message: Vec<u8>,
 		) -> DispatchResult {
-			// TODO handle router ids logic with votes and session_id
```
cdamian, line 435: @lemunozm this logic is handled when the message is processed from the queue, just like we do it for outbound.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Exists when configures a router in `pallet-axelar-router` under `set_config()`

cdamian: OK, but the gateway won't have anything to do with that. Thank you!

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: > routers are now stored under domains

Maybe I could create some confusion with this before, but a router already has the domain information on it (it comes from our latest conversations). So you would not need to store a router under domains.

In the map: `message_hash -> router_id -> (message, votes, session)` the `router_id` already takes care of the domain

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Ok, then you save the `router_id` to process it later, right?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -305,65 +456,58 @@ pub mod pallet {
...
-			for router_id in router_ids {
-				count += 1;
-				if let Err(e) = T::MessageSender::send(router_id, sender.clone(), bytes.clone()) {
-					return (Err(e), LP_DEFENSIVE_WEIGHT.saturating_mul(count));
-				}
-			}
+			let routers = Routers::<T>::get(&domain).ok_or(Error::<T>::RoutersNotFound)?;
 
-			// TODO: Should we fix weights?
-			(Ok(()), LP_DEFENSIVE_WEIGHT.saturating_mul(count))
```
lemunozm, line 354: We can think later if we need a more accurate weight for outbound. I think for now it's ok having this. The previous version skipped some weights in the returned value

lemunozm: By now, the gateway is in charge of putting defensive weights in both inbound and outbound directions. So we can have centralized the weight management until polishing it

lemunozm: @wischli also interested in your thoughts

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Please check the [routers storage](https://github.com/centrifuge/centrifuge-chain/blob/5c8059756c60d0f1d04b8b4e0f9bd55971850359/pallets/liquidity-pools-gateway/src/lib.rs#L190). 

My understanding was that we can have multiple sets of routers, if so, it made sense to me that the `Domain` is the one that differentiates them.

cdamian: Also `message_hash -> router_id -> (message, votes, session)` is not the layout for inbound messages. I think it makes more sense to store these under the `session_id` key, but let's discuss that separately.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Yes, it's part of the inbound gateway message.

pallets/liquidity-pools-gateway/src/lib.rs
wischli: > By now, the gateway is in charge of putting defensive weights in both inbound and outbound directions. So we can have centralized the weight management until polishing it

Sounds perfect! 

> We can think later if we need a more accurate weight for outbound. I think for now it's ok having this. The previous version skipped some weights in the returned value

For now, seems fine. I think the weight is just noise at this point for the audit. Can iterate over that once everything is settled.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: > My understanding was that we can have multiple sets of routers, if so, it made sense to me that the Domain is the one that differentiates them.

Or is it expected that we always have one set of routers regardless of bridge setups etc?

hieronx: I think an additional extrinsic arg is not an issue at all. Just the messages need to match

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -134,8 +164,32 @@ pub mod pallet {
...
+	// pub type DomainRouters<T: Config> = StorageMap<_, Blake2_128Concat, Domain,
+	// T::Router>;
+
+	/// Storage for routers specific for a domain.
+	///
+	/// This can only be set by an admin.
+	#[pallet::storage]
+	#[pallet::getter(fn routers)]
+	pub type Routers<T: Config> =
+		StorageMap<_, Blake2_128Concat, Domain, BoundedVec<T::RouterId, T::MaxRouterCount>>;
```
lemunozm, line 191: Ok, regarding this, because `RouterId` already contains the `Domain` information, the `set_domain_routers()` could be `set_routers()`.

Later, you use `routers.for_domain(domain)` to get the correct router list for a domain, but you should not store a domain per router. Basically because the gateway does not know if that domain is supported for that router

lemunozm: @cdamian referring to the other thread ^^

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: > My understanding was that we can have multiple sets of routers, if so, it made sense to me that the Domain is the one that differentiates them.

https://github.com/centrifuge/centrifuge-chain/pull/1961#discussion_r1715099023

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: OK, lemme try something on a different branch to see if I can wrap my mind around this.

cdamian: Let's discuss more [here](https://github.com/centrifuge/centrifuge-chain/pull/1962/files)

lemunozm: Nice! As far as I could see, you got the idea perfectly! 🙌🏻

runtime/common/src/routing.rs
```diff
@@ -35,8 +35,13 @@ impl From<RouterId> for Domain {
...
 
-impl RouterSupport<Domain> for RouterId {
-	fn for_domain(domain: Domain) -> Vec<Self> {
+/// Static router provider used in the LP gateway.
+pub struct LPGatewayRouterProvider;
+
+impl RouterProvider<Domain> for LPGatewayRouterProvider {
+	type RouterId = RouterId;
+
+	fn routers_for_domain(domain: Domain) -> Vec<Self::RouterId> {
```
lemunozm, line 44: Super great! 💯

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -70,6 +74,13 @@ pub mod pallet {
 	#[pallet::origin]
 	pub type Origin = GatewayOrigin;
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_idle(_now: BlockNumberFor<T>, max_weight: Weight) -> Weight {
+			Self::clear_invalid_session_ids(max_weight)
```
lemunozm, line 80: @cdamian, I'm not sure if we need to remove the previous session IDs in `on_idle()`; I think we should just leave them there. We should remove them only when a message for the same hash has arrived and sessions are different. In that way, the fees/weight of doing that are computed at that moment.

As a rule of thumb, we should avoid `on_idle` unless it needs to be strictly needed.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -118,17 +128,34 @@ pub mod pallet {
...
+		type MaxRouterCount: Get<u32>;
+
+		/// Type for identifying sessions of inbound routers.
+		type SessionId: Parameter
+			+ Member
+			+ BaseArithmetic
+			+ Default
+			+ Copy
+			+ MaybeSerializeDeserialize
+			+ TypeInfo
```
lemunozm, line 147: Super NIT. I think we do not need `MaybeSerializeDeserialize` here and `TypeInfo` is under Parameter IIRC

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -118,17 +128,34 @@ pub mod pallet {
...
 	#[pallet::event]
 	#[pallet::generate_deposit(pub (super) fn deposit_event)]
 	pub enum Event<T: Config> {
-		/// The router for a given domain was set.
-		DomainRouterSet { domain: Domain, router: T::Router },
+		/// The routers for a given domain were set.
+		RoutersSet {
+			router_ids: BoundedVec<T::RouterId, T::MaxRouterCount>,
+			session_id: T::SessionId,
+		},
```
lemunozm, line 158: Nice event!

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -141,14 +168,21 @@ pub mod pallet {
...
-	/// Storage for domain routers.
+	/// Storage for routers.
 	///
 	/// This can only be set by an admin.
 	#[pallet::storage]
-	#[pallet::getter(fn domain_routers)]
-	pub type DomainRouters<T: Config> = StorageMap<_, Blake2_128Concat, Domain, T::Router>;
+	#[pallet::getter(fn routers)]
+	pub type Routers<T: Config> =
+		StorageValue<_, BoundedVec<T::RouterId, T::MaxRouterCount>, OptionQuery>;
```
lemunozm, line 185: NIT. Maybe `ValueQuery` does the trick? if it's not configured is the same as configured with an empty router list.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -168,18 +202,39 @@ pub mod pallet {
...
+	#[pallet::storage]
+	#[pallet::getter(fn pending_inbound_entries)]
+	pub type PendingInboundEntries<T: Config> = StorageDoubleMap<
+		_,
+		Blake2_128Concat,
+		T::SessionId,
+		Blake2_128Concat,
+		(Proof, T::RouterId),
+		InboundEntry<T>,
+	>;
```
lemunozm, line 222: I think having the `session_id` as key of the map makes that you can not clean the entry later when the same message hash has arrived after the session_id changed.

Why did you choose this instead of something like the solidity implementation?: 
```rust
Map<Proof, T::RouterId, (T::SessionId, InboundEntry)>
```

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -168,18 +202,39 @@ pub mod pallet {
...
+
+	/// Storage for inbound message session IDs.
+	#[pallet::storage]
+	pub type SessionIdStore<T: Config> = StorageValue<_, T::SessionId, OptionQuery>;
```
lemunozm, line 226: NIT `ValueQuery`

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -168,18 +202,39 @@ pub mod pallet {
...
+
+	/// Storage that keeps track of invalid session IDs.
+	///
+	/// Any `PendingInboundEntries` mapped to the invalid IDs are removed from
+	/// storage during `on_idle`.
+	#[pallet::storage]
+	#[pallet::getter(fn invalid_message_sessions)]
+	pub type InvalidMessageSessions<T: Config> = StorageMap<_, Blake2_128Concat, T::SessionId, ()>;
```
lemunozm, line 234: We do not need it if the entries are removed one new messages are added for those entries

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,498 @@
...
+			.filter(|x| stored_routers.iter().any(|y| *x == y))
+			.map(|x| x.clone())
+			.collect::<Vec<_>>();
+
+		if res.is_empty() {
+			return Err(Error::<T>::NotEnoughRoutersForDomain.into());
+		}
+
+		Ok(res)
+	}
```
lemunozm, line 67: Perfect!

Maybe just returning an empty vector instead of error?

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,498 @@
...
+				Ok(())
+			}
+			InboundEntry::Proof { .. } => {
+				ensure!(
+					router_ids.get(0) != Some(&router_id),
+					Error::<T>::ProofNotExpectedFromFirstRouter
+				);
+
+				Ok(())
+			}
```
lemunozm, line 142: Not fully get these conditions. Why is there a restriction on receiving the message from the first router in the list and the proofs for the rest? Is it a requirement?

lemunozm (commented):
I have not finished in deep the review. Leaving some thoughts regarding the `session_id` and how the entries are cleaned

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: I mean, I understand for sending message we should choose one of them, let's say, the first one, but why for the inbound messages we should expect the first router in the list? Could not by any of them?

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Not sure if that was a hard requirement when setting new routers. Maybe we can skip this and add some extra check when upserting inbound entries. I'll try this out on a different branch and get back to ya.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Will try removing it, thank you!

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: I had `ValueQuery` for this and the session ID but somehow it felt a bit weird to deal with empty vectors in this case + not having a state where no routers are set. Maybe I'm overthinking this and paranoia got to me...

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: > Why did you choose this instead of something like the solidity implementation?

Mainly due to the aspect of cleaning the map when new routers are set, but will look into it as I said in a different comment.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Yup, will try that out.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Requirement as per [slack discussion](https://kflabs.slack.com/archives/C06FRHCRQL9/p1723054166639529?thread_ts=1723038229.117829&cid=C06FRHCRQL9).

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: Thanks Cosmin!

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: Thanks for the ref. 

I've re-read the thread, and I still do not see why the router configured with index 0 is the router that must send the message. From what I understand from the conversation is that a router can not send a message and also a proof, they must be different routers. But the router who send the message can be router with index 1 and the router who send the proof, router with index 0, isn't?

cdamian: I'm not sure if this is a hard requirement or not. @hieronx can you please clarify?

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done separately [here](https://github.com/centrifuge/centrifuge-chain/pull/1965).

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done separately [here](https://github.com/centrifuge/centrifuge-chain/pull/1965).

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Done separately [here](https://github.com/centrifuge/centrifuge-chain/pull/1965).

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,498 @@
...
+
+	/// Retrieves the information required for processing an inbound
+	/// message.
+	pub(crate) fn get_inbound_processing_info(
+		domain_address: DomainAddress,
+		weight: &mut Weight,
+	) -> Result<InboundProcessingInfo<T>, DispatchError> {
+		let router_ids = Self::get_router_ids_for_domain(domain_address.domain())?;
+
+		weight.saturating_accrue(T::DbWeight::get().reads(1));
```
lemunozm, line 335: To make your life easier, because we're using defensive weights, we do not need to take into account these computations, just returning `LP_DEFENSIVE_WEIGHT` for any processed (inbound or outbound) message is enough

cdamian: Hmm, gotta double-check this since some of these weights are actually returned when the message is handled in the message queue.

cdamian: Oh right, according to [this](https://github.com/centrifuge/centrifuge-chain/blob/feat/lp-v2-gateway-multi-router/pallets/liquidity-pools-gateway/src/message_processing.rs#L405), we can definitely exceed the max processing weight defined [here](https://github.com/centrifuge/centrifuge-chain/blob/88a902d4d2fa5b38ae704f14d5eb5117e7edf444/pallets/liquidity-pools-gateway/src/lib.rs#L540).

So should I go ahead and just remove these small weight accruals for reads/writes?

lemunozm: > So should I go ahead and just remove these small weight accruals for reads/writes?

I would do to leave things cleaner to the auditers

lemunozm: Ideally, this will be benchmarked so we will end up not having weights mixed with the production code

lemunozm: I think leaving the same implementation as it is now in `main` is enough

pallets/liquidity-pools-gateway/src/message_processing.rs
hieronx: The design is such that 
- The set of routers on the EVM side to Centrifuge, must match the set of routers from Centrifuge that that EVM domain
- The first router in the list must send the message, and the remaining routers must send the proofs

Thus it logically follows that the incoming handler should only accept messages from the first router and proofs from the other routers.

lemunozm: So, when configuring routers, the configuration order matters 👍🏻

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,489 @@
...
+			.filter(|x| stored_routers.iter().any(|y| *x == y))
+			.map(|x| x.clone())
+			.collect::<Vec<_>>();
+
+		if res.is_empty() {
+			return Err(Error::<T>::NotEnoughRoutersForDomain.into());
+		}
+
+		Ok(res)
+	}
```
lemunozm, line 64: We should invert this @cdamian, if configuration list priority matter, then we should iterate `stored_routers` first and filter with `router_for_domain`

cdamian: Awesome catch!

cdamian: Coming back to:

>Suppose gateway is configured with the following (dynamically configured by an operator):
AxelarXcm
AxelarEvm
SnowbridgeEvm
>
>And the chain offers the following routers (statically configured in the runtime):
AxelarXcm
AxelarEvm
SnowbridgeEvm
OtherEvm
>
>If a message with Domain::Evm needs to be sent, it must be sent only to AxelarEvm and SnowbridgeEvm.
AxelarXcm will not used because that router is not retrieved by for_domain(Domain::Evm)
OtherEvm will not used because it's not configured in the gateway

The operator now has to ensure the correct order of routers while also keeping filtering in mind.

pallets/liquidity-pools-gateway/src/message_processing.rs
wischli: > I think leaving the same implementation as it is now in `main` is enough

I agree!

cdamian: Yup, will change these soon.

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,489 @@
...
+
+	/// Calculates and returns the proof count required for processing one
+	/// inbound message.
+	fn get_expected_proof_count(router_ids: &Vec<T::RouterId>) -> Result<u32, DispatchError> {
+		let expected_proof_count = router_ids
+			.len()
+			.ensure_sub(1)
+			.map_err(|_| Error::<T>::NotEnoughRoutersForDomain)?;
+
+		Ok(expected_proof_count as u32)
```
wischli, line 74: Nit: Use safe math. This will be flagged in audit, I presume.
```suggestion
		Ok(expected_proof_count.saturated_into())
```

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,489 @@
...
+				);
+
+				Ok(())
+			}
+		}
+	}
+
+	/// Upserts an inbound entry for a particular message, increasing the
+	/// relevant counts accordingly.
+	fn upsert_pending_entry(
```
wischli, line 148: Nit: I have to admit it's really difficult to follow this function. If we have the time on Friday, let's improve readability by splitting into smaller chunks if possible, e.g. impl compare functions for `InboundEntry`.

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,489 @@
...
+						..
+					} => match inbound_entry {
+						InboundEntry::Message {
+							session_id: new_session_id,
+							expected_proof_count: new_expected_proof_count,
+							..
+						} => {
+							if *stored_session_id != new_session_id {
+								*stored_session_id = new_session_id;
+								*stored_expected_proof_count = new_expected_proof_count;
```
wischli, line 176: Q: Shouldn't we reset the proof count if the session id does not match? I cannot tell if this always happens here, i.e. can `new_expected_proof_count` be at most 1?

Update: Ah I see that happens here: https://github.com/centrifuge/centrifuge-chain/pull/1961/files#diff-82099ad70b6f867f15a4b001aa9cfad2a5a8bf0a84cc133ec1d46f5bf78d3f26R314

Will still post this comment to be sure we are all alligned.

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,489 @@
...
+		let mut weight = Default::default();
+
+		let inbound_processing_info =
+			match Self::get_inbound_processing_info(domain_address.clone(), &mut weight) {
+				Ok(i) => i,
+				Err(e) => return (Err(e), weight),
+			};
+
+		weight.saturating_accrue(
+			Weight::from_parts(0, T::Message::max_encoded_len() as u64)
```
wischli, line 406: Nit: Use safe math

```suggestion
			Weight::from_parts(0, T::Message::max_encoded_len().saturated_into())
```

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: Agree! It's a ordered whitelist of routers

wischli (commented):
Another beautiful PR to review! Really liking the explicit errors! A question and a few nits. I hope we have time to improve readability on the marathon matches before the audit 😅

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -188,10 +238,67 @@ pub mod pallet {
...
+
+		/// Not enough routers are stored for a domain.
+		NotEnoughRoutersForDomain,
 	}
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
+		/// Sets the router IDs used for a specific domain,
+		#[pallet::weight(T::WeightInfo::set_routers())]
+		#[pallet::call_index(0)]
```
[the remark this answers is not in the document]
cdamian: Another reason would be XCM right? Do we call this pallet via XCM in any of the multi-chain setups?

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Damn, I always forget about these, will fix. Thank you!

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: I agree, I've seen it too much lately so it's a bit easier for me but still... I'll take a stab at this soon and try to break it into smaller pieces.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Yes, will change this to be explicitly 1 not to leave any door open for any bugs in the future.

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
wischli: > Another reason would be XCM right? 
Yes, XCM is the main reason because the indices enable access (i.e. extrinsics, assets etc) to remote chains without commiting to specific types.

>  Do we call this pallet via XCM in any of the multi-chain setups?
No we don't and this particular extrinsic, as well as the deprecated one, require root origin which cannot be met via XCM.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Nvm, this is correct. If the session changed, we take the new count from the entry, if not, we add it to the existing count.

cdamian: A safer and clearer way would be to just overwrite the entry if the session changed.

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: I think nested matches generally make things difficult to read. Maybe 
```rust
match (stored_entry, entry)
// four cases
```
works better? Not sure.

Also the struct-enum like forces you to always match in the fields, which becomes a bit noisy. Maybe splitting these enums into an enum with `MessageEntry` and `MessageProof` structures makes things much simpler:

```rust
enum InboundEntry {
   Message(MessageEntry)
   Proof(ProofEntry)
}
```

cdamian: I already moved the update logic to a `fn update(&mut self, other: Self)` for inbound entries. Will try out your suggestions there and hopefully we will have something that doesn't cause migraines on review.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Done.

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
cdamian: OK, I'm glad we get away with it this time.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Done, PTAL <3

pallets/liquidity-pools-gateway/src/tests.rs
```diff
@@ -42,812 +50,4349 @@ mod utils {
...
+							MESSAGE_COUNT,
+						);
+
+						run_inbound_message_test_suite(suite);
+					}
+				}
+			}
+
+			#[test]
+			fn same_message_multiple_domain_addresses() {
```
cdamian, line 3148: This one shouldn't pass... it confirms that 2 identical messages from 2 separate domain addresses are not processed separately i.e. there is only 1 inbound entry, for the first one.  We should re-think how we map inbound entries in storage.

lemunozm: Good catch!

I think the runtime thought the `RouterProvider` ensures that the case can not exist. From the gateway perspective, you do not have a way to determine it can not exist. But thanks to the `RouterProvider`, if the received messages have different domains, that forcefully implies that they will have different routers. You need to trust that `RouterProvider` is properly mapping domains into routers, though.

So, I would say, from the gateway point of view, it's ok that the same messages from different domains are processed together. That, ideally, would mean that somehow the router handles both domains, so they should be processed together. But really that case will never exists if there are no errors in `RouterProvider`

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+	pub fn create_post_voting_entry(
+		inbound_entry: &InboundEntry<T>,
+		inbound_processing_info: &InboundProcessingInfo<T>,
+	) -> Result<Option<Self>, DispatchError> {
+		match inbound_entry {
+			InboundEntry::Message(message_entry) => {
+				if message_entry.session_id != inbound_processing_info.current_session_id {
+					return Ok(None);
+				}
+
```
lemunozm, line 117: At this stage, this can never happen, right?

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+						expected_proof_count: updated_count,
+						..message_entry.clone()
+					}
+					.into(),
+				))
+			}
+			InboundEntry::Proof(proof_entry) => {
+				if proof_entry.session_id != inbound_processing_info.current_session_id {
+					return Ok(None);
+				}
```
lemunozm, line 137: Same here

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+	/// The LP message.
+	pub message: T::Message,
+
+	/// The expected proof count for processing one or more of the provided
+	/// message.
+	///
+	/// NOTE - this gets increased by the `expected_proof_count` for a set of
+	/// routers (see `get_expected_proof_count`) every time a new identical
+	/// message is submitted.
+	pub expected_proof_count: u32,
```
lemunozm, line 39: Do we really need this field? My rationale:

This information already exists indirectly in the `PendingInboundEntries`, and can be computed for a specific message on-the-fly in `process_inbound_message()`, checking the storages for the routers that message domain belong and counting them.

So maybe we do not need to store it, and later read it and update that field along the code. WDYT?

lemunozm (commented):
First, amazing work here @cdamian, don't get scared for my lists of questions/NITs. I didn't find any fault 🙌🏻 

My comments are mostly regarding some minor simplification of readabilidy improvements (it's already quite clear, but the topic is quite complex, so maybe we can help auditors a bit more). The algorithm itself is great and things overall are well organized! I feel like you realized or every minor detail, awesome!

lemunozm:
Regarding the tests, I think we should simplify them and/or factorize them. 5k lines are too many lines to refactor tomorrow 😅. 

I think that by using two messages in the test case and 2 routers, we can have all the safety we're searching for.

At least at this stage where we actually only support 1 router 😆

pallets/liquidity-pools-gateway/src/tests.rs
cdamian: OK, thank you for double-checking and clarifying!

pallets/liquidity-pools-gateway/Cargo.toml
```diff
@@ -35,6 +36,9 @@ cfg-utils = { workspace = true }
 [dev-dependencies]
 cfg-mocks = { workspace = true, default-features = true }
 sp-io = { workspace = true, default-features = true }
+itertools = { workspace = true, default-features = true }
+lazy_static = { workspace = true, default-features = true }
+mock-builder = { workspace = true, default-features = true }
```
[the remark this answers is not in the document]
cdamian: Removed.

pallets/liquidity-pools-gateway/src/mock.rs
```diff
@@ -47,38 +73,59 @@ impl LPEncoding for Message {
...
+impl RouterProvider<Domain> for TestRouterProvider {
+	type RouterId = RouterId;
+
+	fn routers_for_domain(domain: Domain) -> Vec<Self::RouterId> {
+		match domain {
+			Domain::Centrifuge => vec![],
+			Domain::EVM(_) => vec![ROUTER_ID_1, ROUTER_ID_2, ROUTER_ID_3],
+		}
 	}
 }
```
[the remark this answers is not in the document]
cdamian: Yup, I just added this here for the sake of testing, we're still covering cases with routers that are not part of this list.

libs/traits/src/liquidity_pools.rs
```diff
@@ -31,11 +33,20 @@ pub trait LPEncoding: Sized {
 	/// Creates an empty message.
 	/// It's the identity message for composing messages with pack_with
 	fn empty() -> Self;
+
+	/// Retrieves the message proof, if any.
+	fn get_message_proof(&self) -> Option<Proof>;
+
+	/// Converts the message into a message proof type.
+	fn to_message_proof(&self) -> Self;
```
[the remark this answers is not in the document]
cdamian: I never was a big fan of the names that I picked for these haha. Your suggestion is definitely better, thank you!

pallets/liquidity-pools-gateway/src/mock.rs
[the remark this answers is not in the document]
lemunozm: Amazing!

libs/traits/src/liquidity_pools.rs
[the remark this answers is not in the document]
lemunozm: Maybe `message_hash()` is more exact than `proof_hash()` because, actually, it's just a hash that does nothing to do with proof. Later, the proof uses that hash, but that's all. WDYT?

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -91,13 +98,23 @@ pub mod pallet {
...
+			+ MaxEncodedLen
+			+ TypeInfo
+			+ FullCodec;
+
+		/// The target of the messages coming from this chain
 		type MessageSender: MessageSender<Middleware = Self::RouterId, Origin = DomainAddress>;
 
 		/// An identification of a router
-		type RouterId: RouterSupport<Domain> + Parameter;
+		type RouterId: Parameter + Default + MaxEncodedLen;
```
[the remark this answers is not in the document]
cdamian: This was a remnant of the LP gateway queue benchmarks that required defaults for message, thus gateway message, thus router ID. I removed it for now and it seems OK so far. Thank you!

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -188,10 +248,74 @@ pub mod pallet {
...
+		/// messages.
+		#[pallet::weight(T::WeightInfo::set_routers())]
+		#[pallet::call_index(0)]
+		pub fn set_routers(
+			origin: OriginFor<T>,
+			router_ids: BoundedVec<T::RouterId, T::MaxRouterCount>,
+		) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
+
+			ensure!(router_ids.len() > 0, Error::<T>::InvalidRouters);
```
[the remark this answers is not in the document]
cdamian: OK, then I'll remove this check.

pallets/liquidity-pools-gateway/src/message_processing.rs
wischli: > This information already exists indirectly in the `PendingInboundEntries`, and can be computed for a specific message on-the-fly in `process_inbound_message()`, by checking the storages for the routers to which the message domain belongs and counting them.

Can you help me and tell me which storage(s) in the router we would check for this?

IIUC, this would be at least one storage read per router right. Given that is is recommended to keep the weight of processing inbound messages low for `on_idle`, I currently support keeping this field here. Looking forward to understanding your path better @lemunozm!

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -301,69 +425,60 @@ pub mod pallet {
...
 
-		fn queue_message(destination: Domain, message: T::Message) -> DispatchResult {
-			// We are using the sender specified in the pallet config so that we can
-			// ensure that the account is funded
-			let gateway_message = GatewayMessage::<T::Message>::Outbound {
-				sender: T::Sender::get(),
-				destination,
-				message,
-			};
+			Self::deposit_event(Event::<T>::MessageRecoveryExecuted { proof, router_id });
```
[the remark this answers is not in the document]
cdamian: I would say we want it as long as `execute_if_requirements_are_met` doesn't error out, regardless of whether the message gets dispatched or not. Imagine a case where we have one message and we need 2 proofs that we recover via this extrinsic. On the first call the message won't be executed but this event should still be there.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -301,69 +425,60 @@ pub mod pallet {
...
-				}
-			}
+			proof: Proof,
+			router_id: T::RouterId,
+		) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
 
-			(Ok(()), LP_DEFENSIVE_WEIGHT.saturating_mul(count))
-		}
+			let inbound_processing_info: InboundProcessingInfo<T> = domain_address.try_into()?;
```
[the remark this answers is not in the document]
cdamian: These funcs were uglier in previous commits and I didn't get to tackle this yesterday, hopefully, I can remove it completely now.

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+				message,
+				expected_proof_count: inbound_processing_info.expected_proof_count_per_message,
+			}),
+			Some(_) => InboundEntry::Proof(ProofEntry {
+				session_id: inbound_processing_info.current_session_id,
+				current_count: 1,
+			}),
+		}
+	}
+}
```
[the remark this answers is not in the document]
wischli: > As a note, From/TryFrom are very cool but they're a double-edged sword, because you lose traceability. They are usually used only for very isometric types.

Definitely agree. Feels like having a dedicated fuction such as `into_inbound_entry` is clearer. Though this is definitely a nit! Since we have more `From` impls below, no need to change this for now IMO.

libs/traits/src/liquidity_pools.rs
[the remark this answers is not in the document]
cdamian: `message_proof` would be the best fitting for this case, `message_hash` can also mean other things IMO.

How about - `message_proof` and `to_proof_message`?

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+			InboundEntry::Message(_) => Err(Error::<T>::ExpectedMessageProofType.into()),
+		}
+	}
+
+	/// A pre-dispatch update involves increasing the `expected_proof_count` or
+	/// `current_count` of `self` with the one of `other`.
+	///
+	/// If a session ID change is detected, `self` is replaced completely by
+	/// `other`.
+	pub fn pre_dispatch_update(&mut self, other: Self) -> DispatchResult {
```
wischli, line 217: Much cleaner and easier to follow now, thanks for that!

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+				None => return Ok(()),
+				Some(stored_inbound_entry) => match stored_inbound_entry {
+					InboundEntry::Message(message_entry) => message = Some(message_entry.message),
+					InboundEntry::Proof(proof_entry) => {
+						if proof_entry
+							.has_valid_vote_for_session(inbound_processing_info.current_session_id)
+						{
+							votes.ensure_add_assign(1)?;
+						}
+					}
```
wischli, line 400: Nit/food for thought: Could pull out the if condition, but not sure if it improves readability since then we have to handle the remainder case.
```suggestion
					InboundEntry::Proof(proof_entry) if proof_entry .has_valid_vote_for_session(inbound_processing_info.current_session_id) => {
						votes.ensure_add_assign(1)?;
					},
					_ => {},
```

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
cdamian: Hmm, these fields are still used all around the inbound processing logic, if we remove this we'll have more args over there.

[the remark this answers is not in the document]
cdamian: I'll remove it and see how everything looks.

wischli (commented):
Haven't gone through all tests yet and still feel a bit lost understanding the new intertwined connections together with  #1958 and $1959.

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Right! I remember now, thanks!

libs/traits/src/liquidity_pools.rs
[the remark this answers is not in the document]
lemunozm: Mmm... But the proof itself is not a hash? More specifically, a `keccak_256` hash? Later, that hash is used as proof by the message proof. Probably this would mean we should rename the type `Proof` as `MessageHash` in order to split clearly both concepts.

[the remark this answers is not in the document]
wischli: This is really not that simple picking well descriptive names! I actually liked having the `get_message_proof` because it returns an option which is usually the case for getters.

I will throw another option in the fictional bin of options: Keep the first one slightly reduced `get_proof` and name the other one `to_proof_message` as proposed by @cdamian.

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
lemunozm: > If we remove this we'll have more args over there

From what I can see, everything can be deducted from the domain, right? So just passing the domain could be enough

pallets/liquidity-pools-gateway/src/message_processing.rs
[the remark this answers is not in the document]
lemunozm: > Since we have more From impls below

I think the below ones makes sense because they are straightforward, just wrappers and things like that

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: You anyway needs to look into all entries later to check if the message must be executed counting votes. Anyways the reads will be 2-3, the same as configured routers, so very very low.

lemunozm: In `PendingInboundEntries`. If, for a message proof hash, you get a list with the second keys: the `router_ids`, and count them, you have the `expected_proof_count` value

libs/traits/src/liquidity_pools.rs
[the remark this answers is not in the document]
cdamian: Ok, renamed this.

pallets/liquidity-pools-gateway/src/lib.rs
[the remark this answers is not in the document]
cdamian: Drop that struct completely, PTAL.

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+		message: T::Message,
+		router_id: T::RouterId,
+	) -> (DispatchResult, Weight) {
+		let weight = LP_DEFENSIVE_WEIGHT;
+
+		match T::MessageSender::send(router_id, sender, message.serialize()) {
+			Ok(_) => (Ok(()), weight),
+			Err(e) => (Err(e), weight),
+		}
+	}
```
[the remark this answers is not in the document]
cdamian: Moved this logic back to the pallet. Dunno about renaming that, let's keep it for now.

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+				message_proof,
+				router_id.clone(),
+			) {
+				return (Err(e), weight.saturating_mul(count));
+			}
+
+			match Self::execute_if_requirements_are_met(&inbound_processing_info, message_proof) {
+				Err(e) => return (Err(e), weight.saturating_mul(count)),
+				Ok(_) => continue,
+			}
```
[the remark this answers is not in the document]
cdamian: To me this seems like a 2 part kind of process:
1 - get the initial entry and update counts;
2 - check the counts, execute if needed and update counts again.

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+				},
+			};
+		}
+
+		if votes < inbound_processing_info.expected_proof_count_per_message {
+			return Ok(());
+		}
+
+		match message {
+			Some(msg) => {
```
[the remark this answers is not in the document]
cdamian: Done.

pallets/liquidity-pools-gateway/src/message_processing.rs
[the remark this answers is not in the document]
lemunozm: But the above without the second one can not exist, right? Because it will leave the state in a wrong state. Not a strong opinion about this point! Too much nitpick from my side

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,539 @@
...
+	/// counts reach 0.
+	pub(crate) fn execute_post_voting_dispatch(
+		inbound_processing_info: &InboundProcessingInfo<T>,
+		message_proof: Proof,
+	) -> DispatchResult {
+		for router_id in &inbound_processing_info.router_ids {
+			PendingInboundEntries::<T>::try_mutate(message_proof, router_id, |storage_entry| {
+				match storage_entry {
+					None => {
+						Err::<(), DispatchError>(Error::<T>::PendingInboundEntryNotFound.into())
```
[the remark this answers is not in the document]
cdamian: True, this can only happen if one is to call it right after an initial message is received is something, but also for consistency's sake, I would like to have a clear error for this case.

pallets/liquidity-pools-gateway/src/message_processing.rs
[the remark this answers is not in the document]
cdamian: I stil changed some things in there, mainly dropped `validate_and_upsert_pending_entries`, most of the logic is now in `process_inbound_message`.

pallets/liquidity-pools-gateway/src/message_processing.rs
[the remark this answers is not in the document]
cdamian: I'll admit that I used and abused from/try_from :D. Dropped those and now there's a specific `create` fn in the `InboundEntry` impl.

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,505 @@
...
+				}
+			})?;
+		}
+
+		Ok(())
+	}
+
+	/// Iterates over a batch of messages and checks if the requirements for
+	/// processing each message are met.
+	pub(crate) fn process_inbound_message(
```
lemunozm, line 415: Added a TODO for me to find a way to factorize those `Err`, should there be some.

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	/// Checks if the number of proofs required for executing one message
+	/// were received, and if so, decreases the counts accordingly and executes
+	/// the message.
+	pub(crate) fn execute_if_requirements_are_met(
+		message_proof: Proof,
+		router_ids: &[T::RouterId],
+		session_id: T::SessionId,
+		expected_proof_count: u32,
```
lemunozm, line 340: These 3 can be obtained from the domain to avoid passing them, right? No strong opinion anyways

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: > Do we really need this field?
Yes, again required for ensuring proper processing of duplicate messages. For example, if we wouldn't do this for a case such as [this](https://github.com/centrifuge/centrifuge-chain/blob/88bdad7b58cbb3661c405711fba34a959b5589ae/pallets/liquidity-pools-gateway/src/tests.rs#L1817), the message would be "lost" after the first proof is received.

The name is a bit confusion tho, maybe we should rename it to `total_expected_proof_count`.

pallets/liquidity-pools/src/message.rs
```diff
@@ -557,6 +562,19 @@ impl LPEncoding for Message {
...
+			Message::MessageProof { hash } => Some(*hash),
+			_ => None,
+		}
+	}
+
+	fn to_proof_message(&self) -> Self {
+		let hash = keccak_256(&LPEncoding::serialize(self));
+
+		Message::MessageProof { hash }
+	}
```
lemunozm, line 577: I think my naming confusion comes from here. In my mind:

```rust
fn message_hash(&self) -> MessageHash {
    //Independently of the kind of message, I should be able to get a hash of it
    keccak_256(&LPEncoding::serialize(self))
}

fn message_proof(&self) -> Self {
    match self {
        Message::MessageProof(..) => self,
        other_msg => Message::MessageProof(other_msg.message_hash())
    }
}
```

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,505 @@
...
+		Ok(expected_proof_count.saturated_into())
+	}
+
+	/// Gets the message proof for a message.
+	pub(crate) fn get_message_proof(message: T::Message) -> Proof {
+		match message.get_proof() {
+			None => message
+				.to_proof_message()
+				.get_proof()
+				.expect("message proof ensured by 'to_message_proof'"),
```
lemunozm, line 307: Because if not, I'm afraid here we can get a hash proof of a proof message instead of the message itself, which will be different: `hash(MessageProof(msg))= != hash(msg)`

lemunozm (commented):
Much more readable @cdamian, thanks!! Just some thought about the message proofs methods, that maybe we can have an issue there that we're not catching in UTs

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Some evm crates have macros like:
```

macro_rules! try_or_fail {
	( $e:expr ) => {
		match $e {
			Ok(v) => v,
			Err(e) => return Control::Exit(e.into()),
		}
	};
}
```

Was thinking that we can do something similar.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Removed from UTs.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Yes, but we also want to do the minimum amount of reading from storage. I can pass the domain here and retrieve the router IDs again but that will feel off no?

pallets/liquidity-pools-gateway/src/message_processing.rs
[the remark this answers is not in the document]
lemunozm: As you think is better 👍🏻, could you comment saying that this can not be produced?

libs/traits/src/liquidity_pools.rs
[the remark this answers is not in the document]
lemunozm: Some notes below regarding this

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Only if we screw up the implementation in the LP message.

pallets/liquidity-pools/src/message.rs
lemunozm: Just explaining my thoughts about this ^^

pallets/liquidity-pools-gateway/src/message_processing.rs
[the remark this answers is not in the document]
cdamian: Sure, thank you.

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: Ah, okay. revisiting that code, the `to_proof_message().get_proof()` is only done by first checking that the message was not proof previously. So we're safe then 👍🏻 

Sorry for the noise here!

pallets/liquidity-pools/src/message.rs
wischli: > I think my naming confusion comes from here. In my mind:
> 
> ```rust
> // Nothing to do with proofs, just give the hash
> fn message_hash(&self) -> MessageHash {
>     //Independently of the kind of message, I should be able to get a hash of it
>     keccak_256(&LPEncoding::serialize(self))
> }
> 
> // Creates a new message containing the hash of message
> // or, return the itself if it's already a proof, without recomputing the hash
> fn message_proof(&self) -> Self {
>     match self {
>         Message::MessageProof(..) => self,
>         other_msg => Message::MessageProof(other_msg.message_hash())
>     }
> }
> ```

Are you proposing this as an alternative @lemunozm ?  AFAICT, this is something different than the current impl of `get_proof` which extracts the proof hash if it exists. Your method will always return a hash independent of the message variant.

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: Mmm... yeah, no strong opinion. Makes sense what you said 👍🏻.

Just as a note, reading multiple times the same storage only increments the reads by 1.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Yes, same as the case we discussed for `execute_post_voting_dispatch`. It can only happen if the function is called without confirming that a message is ready for processing. I think we can skip these checks. Same goes for the other note below.

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: Interesting 😲.

Will think about this in a next PR with well-weight treatment 👍🏻

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Changed. Thank you!

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: No worries, we should look under every rock and try to make this logic as clear as possible!

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Awesome, thank you!

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: I'm not sure if I fully understood what you said. Even if the first message is a proof, we still have the information about the required routers for that domain, isn't? So we can compute the count. 

But don't let this to block us from merging this 👍🏻

pallets/liquidity-pools-gateway/src/tests.rs
```diff
@@ -42,812 +48,4185 @@ mod utils {
...
+									(
+										vec![
+											(ROUTER_ID_1, Message::Simple),
+											(ROUTER_ID_2, Message::Proof(MESSAGE_PROOF)),
+										],
+										ExpectedTestResult {
+											message_submitted_times: 1,
+											expected_storage_entries: vec![
+												(ROUTER_ID_1, None),
+												(ROUTER_ID_2, None),
```
lemunozm, line 1333: Instead of `None`, should it then be removed? Could then the entry be there forever?

cdamian: We don't store options in storage, the value you're referring to gets compared to the output of `get` on the storage here:
https://github.com/centrifuge/centrifuge-chain/blob/584e7e8eb4333942a1093bc5f3bb2c9b06a5114c/pallets/liquidity-pools-gateway/src/tests.rs#L1038

cdamian: So that should mean that the storage is empty.

cdamian:
Gonna add the fixes for ITs on a different branch to reduce the noise here.

pallets/liquidity-pools-gateway/src/tests.rs
lemunozm: Ah great! Thanks!

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: Sorry, Cosmin, maybe I am pointing you to a bad state. Now I think those checks could be in charge of removing the message entry. Could that be?

cdamian: No, those were the ones related to the post voting updates, and in those cases, all pending inbound entries are valid.

pallets/liquidity-pools-gateway/src/message_processing.rs
wischli: Let's sort this out tomorrow and merge this beauty!

pallets/liquidity-pools-gateway/src/message_processing.rs
wischli: Is this still a relevant issue?

lemunozm (approved):
Thanks @Cosmin for flighting with this beast! And also for being so patient with all my questions/suggestions and NITs.

Amazing work about to be merged!

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: We can even merge this, and solve this (if it's required) in a next up PR

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -0,0 +1,495 @@
...
+				None => message_proof.clone(),
+			};
+
+			// We are using the sender specified in the pallet config so that we can
+			// ensure that the account is funded
+			let gateway_message = GatewayMessage::<T::Message, T::RouterId>::Outbound {
+				sender: T::Sender::get(),
+				message: router_msg,
+				router_id,
+			};
```
lemunozm, line 488: NIT for the next PR.

We're storing outbound messages with a `sender` field, which is a constant. We can remove the `sender` field from here, and when the message is unqueued, use it there.

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: > Basically, when we get the second message, the expected count gets increased by the expected counts for processing a message for this case, which is one. This ensures that the message entry doesn't get removed from storage when the first proof is received.

Now I understand you Cosmin. I think you're right 👍🏻. Such a complex casuistics has this feature!

To resume, basically, if we receive two identical messages with no proofs, we're storing the same message twice in the same entry, so we anyways need some kind of counter to identify that state. We use the above counter, and we take advance of having a counter to also count the expected proofs 👍🏻

wischli (approved):
Thanks for all the effort you have put into slaying this enormous beast! IMO, we can put an end to this chapter and iron out any minor leftovers in a subsequent PR.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Will take care of this in the following PRs for message recovery, cheers!

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: No, this was related to removing entries if a session change is detected, however, that won't be the case in this scenario since this logic only gets called if all storage entries are present for the current session.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Will be done in - https://github.com/centrifuge/centrifuge-chain/pull/1970

PR #1962, opened 2024-08-13
Dont use domain when storing routers

cdamian:


pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -309,35 +307,28 @@ pub mod pallet {
...
 			origin: OriginFor<T>,
-			domain: Domain,
 			router_ids: BoundedVec<T::RouterId, T::MaxRouterCount>,
 		) -> DispatchResult {
 			T::AdminOrigin::ensure_origin(origin)?;
 
-			ensure!(domain != Domain::Centrifuge, Error::<T>::DomainNotSupported);
-
-			//TODO(cdamian): Outbound -  Call router.init() for each router?
+			<Routers<T>>::set(router_ids.clone());
```
cdamian, line 316: Still a bit confused by this setup since the routers that we're storing at not the ones that we are using when processing...

Anyway, here I'm storing all the available routers right @lemunozm?

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -47,8 +49,8 @@ pub struct InboundProcessingInfo<T: Config> {
 impl<T: Config> Pallet<T> {
 	/// Calculates and returns the proof count required for processing one
 	/// inbound message.
-	fn get_expected_proof_count(domain: &Domain) -> Result<u32, DispatchError> {
-		let routers = Routers::<T>::get(domain).ok_or(Error::<T>::RoutersNotFound)?;
+	fn get_expected_proof_count() -> Result<u32, DispatchError> {
+		let routers = Routers::<T>::get();
```
cdamian, line 53: And here, for example, I should instead use `T::RouterId::for_domain...`?

cdamian:
Main question would be, when to use `Routers::<T>::get()` and when to use `T::RouterId::for_domain()`.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: So if a message comes from a not configured router, then the gateway will discard it

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: Yes, because we want to get only the valid routers for the domain where the message comes

lemunozm:
> Routers::<T>::get() and when to use T::RouterId::for_domain()

- `Routers::<T>::get()` the list of all configured routers, something like a whitelist for the gateway.
- `T::RouterId::for_domain()` the list of all configured routers for a specific domain. Used each time you need to know the routers for a domain.
  - More accurate will be to always filter this list with the `Router::<T>::get()`. I mean, each router returned by the `T::RouterId::for_domain()` that is not in `Router::<T>::get()` must be discarded

cdamian:
> > Routers::::get() and when to use T::RouterId::for_domain()
> 
> * `Routers::<T>::get()` the list of all configured routers, something like a whitelist for the gateway.
> * `T::RouterId::for_domain()` the list of all configured routers for a specific domain. Used each time you need to know the routers for a domain.
>   
>   * More accurate will be to always filter this list with the `Router::<T>::get()`. I mean, each router returned by the `T::RouterId::for_domain()` that is not in `Router::<T>::get()` must be discarded
>     
>     * Maybe an auxiliary method that does that or similar, replacing `for_domain()` for that auxiliary method

Still not sure what the benefit is from doing all of this, instead of keeping track which router IDs are set for each domain, in storage, like the current approach.

lemunozm:
Because basically, `T::RouterId::for_domain()` is representing the static relation between routers and domains, it's something that the gateway doesn't know (and something it can not ensure either). The gateway knows configured routers.

cdamian:
> Because basically, `T::RouterId::for_domain()` is representing the static relation between routers and domains, it's something that the gateway doesn't know (and something it can not ensure either). The gateway knows configured routers.
> 
> This list: `T::RouterId::for_domain()`, exists independently of the routers configured in the gateway

But what's wrong about having [this relation](https://github.com/centrifuge/centrifuge-chain/blob/807121e72789b0c8e3f89d1dfc9bf0ef77f555b8/runtime/common/src/routing.rs#L41) in storage?

If this changes at any time, we'll have to do a RU upgrade right? If we store this, we just update the storage with whatever we need.

cdamian:
> Because basically, T::RouterId::for_domain() is representing the static relation between routers and domains, it's something that the gateway doesn't know (and something it can not ensure either). The gateway knows configured routers.

In the [current implementation](https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/liquidity-pools-gateway/src/lib.rs#L151), the gateway knows about this relation. And I would also argue that they gateway should know, going forward.

lemunozm:
The gateway doesn't know because not all Domain can be mapped to all Routers

lemunozm:
The relation is not something the operator can choose, is something inherent from how the chain is coded, statically

lemunozm:
Suppose, you use that storage. The user operator can add an entry Domain::Centrifuge with router AxelarEvm, which could not exists. Also the gateway can not check the relation is wrong

cdamian:
> The gateway doesn't know because not all Domain can be mapped to all Routers
> Suppose, you use that storage. The user operator can add an entry Domain::Centrifuge with router AxelarEvm, which could not exists. Also the gateway can not check the relation is wrong

OK, this is the case in the current implementation and one can definitely screw up the domain to router relationship.

> The relation is not something the operator can choose, is something inherent from how the chain is coded, statically

^ fixes that.

Thank you for clarifying @lemunozm!

cdamian:
Will adjust some things here then merg so I can continue working on the initial branch for multi-routers.

PR #1964, opened 2024-08-13
feat: recover tokens

wischli:
# Description

* Adds root extrinsic to initiate token recovery in EVM with UTs and IT
* Minor improvements in LP inbound tests
* Add `schedule_upgrade` and `cancel_upgrade` ITs

# Checklist:


pallets/liquidity-pools/src/lib.rs
```diff
@@ -1013,6 +1013,44 @@ pub mod pallet {
...
+		/// Origin: Root.
+		#[pallet::call_index(17)]
+		#[pallet::weight(T::WeightInfo::update_tranche_hook())]
+		pub fn recover_assets(
+			origin: OriginFor<T>,
+			domain_address: DomainAddress,
+			incorrect_contract: [u8; 32],
+			asset: [u8; 32],
+			// NOTE: Solidity balance is `U256` per default
+			amount: sp_core::U256,
```
lemunozm, line 1032: But we're using `u128` in other places as `TransferAssets`. Should this be unified?

pallets/liquidity-pools/src/mock.rs
```diff
@@ -26,7 +26,7 @@ pub const CHAIN_ID: u64 = 1;
 pub const ALICE_32: [u8; 32] = [2; 32];
 pub const ALICE: AccountId = AccountId::new(ALICE_32);
 pub const ALICE_ETH: [u8; 20] = [2; 20];
-pub const ALICE_EVM_DOMAIN_ADDRESS: DomainAddress = DomainAddress::EVM(42, ALICE_ETH);
+pub const ALICE_EVM_DOMAIN_ADDRESS: DomainAddress = DomainAddress::EVM(CHAIN_ID, ALICE_ETH);
```
lemunozm, line 29: 👌🏻

pallets/liquidity-pools/src/tests.rs
```diff
@@ -2008,3 +2008,102 @@ mod update_tranche_hook {
...
+						RuntimeOrigin::signed(ALICE.into()),
+						ALICE_EVM_DOMAIN_ADDRESS,
+						CONTRACT,
+						ASSET,
+						AMOUNT.into(),
+					),
+					DispatchError::BadOrigin
+				);
+			});
+		}
```
lemunozm, line 2089: NIT. Do we really need both or maybe just one to determine that the `BadOrigin` error is correctly dispatched?

pallets/liquidity-pools/src/tests/inbound.rs
```diff
@@ -181,10 +183,8 @@ mod handle_tranche_tokens_transfer {
 				}
 			));
 
-			let origin = EVM_DOMAIN_ADDRESS.domain().into_account();
-			assert_eq!(Tokens::balance(TRANCHE_CURRENCY, &origin), 0);
-
```
lemunozm, line 186: Should it not be 0 after transferring the tranche tokens?

pallets/liquidity-pools/src/tests/inbound.rs
```diff
@@ -169,6 +169,8 @@ mod handle_tranche_tokens_transfer {
 				AMOUNT,
 			)
 			.unwrap();
+			let origin = EVM_DOMAIN_ADDRESS.domain().into_account();
+			assert_eq!(Tokens::balance(TRANCHE_CURRENCY, &origin), AMOUNT);
```
lemunozm, line 173: I feel like here we're testing the test (testing the above line was correct)

lemunozm (commented):
LGTM! Just some questions

pallets/liquidity-pools/src/tests/inbound.rs
wischli: Yes and no. The test setup was unrealistic because it tested receiving a tranche token transfer from EVM chain 1 to EVM chain id 1. Fixed in [c75a38d](https://github.com/centrifuge/centrifuge-chain/pull/1964/commits/c75a38de95ca797494a4d3621b3dd826af42b4aa)

pallets/liquidity-pools/src/tests/inbound.rs
wischli: Definitely, removed.

pallets/liquidity-pools/src/tests.rs
wischli: Not necessarily because we can assume that to be inherited from Substrate but I also think it does not hurt to be safe because both cases are different.

pallets/liquidity-pools/src/tests/inbound.rs
lemunozm: Thanks for fixing it!

pallets/liquidity-pools/src/lib.rs
wischli: I had the same question ([-> slack thread](https://kflabs.slack.com/archives/C06FRHCRQL9/p1721981066605389?thread_ts=1721940738.924059&cid=C06FRHCRQL9)). The summary would be that this message initates a recovery of an arbitrary token (e.g. erc20 USDC). Amounts in EVM are u256 per default such that it makes sense to have the amount type in u256. Unfortunately, I had to fallback converting to `[u8; 32]` in the end because of serialization issues.

lemunozm: Regarding the serialization, maybe you can a dd a wrapper type:

// Pseudocode
```rust
struct U256Wrapper(U256);
impl Serialize for U256Wrapper {
    fn serialize() {
        self.low_128().serialize()
        self.high_128().serialize()
    }
}
```

Also, maybe the `U256` already implements Serialize/Deserialize under some `serde` feature that we need to set (?)

lemunozm: Maybe you need to use directly the dependency `primitive-types` enabling `serde`: https://docs.rs/primitive-types/0.12.2/primitive_types/

lemunozm: Just some ideas, feel free to skip them

lemunozm: @wischli, just to know if you read this

wischli: Yeah I have read it but I didnt have the time yet to try it out, unblocking dev and demo is higher priority.

wischli: As discussed on Slack, unfortunately the serialization is blocked by us missing serialization from `str` and `U256` being serialzed to `str` hex unfortunately. So will keep as is!

lemunozm (approved):
Thanks William!

lemunozm:
Let's merge this!

PR #1965, opened 2024-08-14
lp-gateway: Drop session ID invalidation

cdamian:
Addressing some comments from - https://github.com/centrifuge/centrifuge-chain/pull/1961.

### DIFF
--- a/pallets/liquidity-pools-gateway/src/message_processing.rs
+++ b/pallets/liquidity-pools-gateway/src/message_processing.rs
@@ -14,24 +14,21 @@ use sp_arithmetic::traits::{EnsureAddAssign, EnsureSub};
 use sp_runtime::DispatchError;
 
 use crate::{
-	message::GatewayMessage, Config, Error, InvalidMessageSessions, Pallet, PendingInboundEntries,
-	Routers, SessionIdStore,
+	message::GatewayMessage, Config, Error, Pallet, PendingInboundEntries, Routers, SessionIdStore,
 };
-
-/// The limit used when clearing the `PendingInboundEntries` for invalid
-/// session IDs.
-const INVALID_ID_REMOVAL_LIMIT: u32 = 100;
 
 /// Type used when storing inbound message information.
 #[derive(Debug, Encode, Decode, Clone, Eq, MaxEncodedLen, PartialEq, TypeInfo)]
 #[scale_info(skip_type_params(T))]
 pub enum InboundEntry<T: Config> {
 	Message {
+		session_id: T::SessionId,
 		domain_address: DomainAddress,
 		message: T::Message,
 		expected_proof_count: u32,
 	},
 	Proof {
+		session_id: T::SessionId,
 		current_count: u32,
 	},
 }
@@ -51,7 +48,7 @@ impl<T: Config> Pallet<T> {
 	pub fn get_router_ids_for_domain(domain: Domain) -> Result<Vec<T::RouterId>, DispatchError> {
 		let all_routers_for_domain = T::RouterProvider::routers_for_domain(domain);
 
-		let stored_routers = Routers::<T>::get().ok_or(Error::<T>::RoutersNotFound)?;
+		let stored_routers = Routers::<T>::get();
 
 		let res = all_routers_for_domain
 			.iter()
@@ -91,17 +88,20 @@ impl<T: Config> Pallet<T> {
 	/// Creates an inbound entry based on whether the inbound message is a
 	/// proof or not.
 	fn create_inbound_entry(
-		domain_address: DomainAddress,
+		inbound_processing_info: &InboundProcessingInfo<T>,
 		message: T::Message,
-		expected_proof_count: u32,
 	) -> InboundEntry<T> {
 		match message.get_message_proof() {
 			None => InboundEntry::Message {
-				domain_address,
+				session_id: inbound_processing_info.current_session_id.clone(),
+				domain_address: inbound_processing_info.domain_address.clone(),
 				message,
-				expected_proof_count,
+				expected_proof_count: inbound_processing_info.expected_proof_count_per_message,
 			},
-			Some(_) => InboundEntry::Proof { current_count: 1 },
+			Some(_) => InboundEntry::Proof {
+				session_id: inbound_processing_info.current_session_id,
+				current_count: 1,
+			},
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -14,24 +14,21 @@ use sp_arithmetic::traits::{EnsureAddAssign, EnsureSub};
...
-/// The limit used when clearing the `PendingInboundEntries` for invalid
-/// session IDs.
-const INVALID_ID_REMOVAL_LIMIT: u32 = 100;
-
 /// Type used when storing inbound message information.
 #[derive(Debug, Encode, Decode, Clone, Eq, MaxEncodedLen, PartialEq, TypeInfo)]
 #[scale_info(skip_type_params(T))]
 pub enum InboundEntry<T: Config> {
 	Message {
+		session_id: T::SessionId,
```
cdamian, line 25: Instead of `Map<Proof, T::RouterId, (T::SessionId, InboundEntry)>`, I added the session entry to the actual entry. This gives us a cleaner storage layout and it's definitely easier to deal with it given all the `try_mutate` calls.

Lemme know what you think @lemunozm

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -14,24 +14,21 @@ use sp_arithmetic::traits::{EnsureAddAssign, EnsureSub};
...
 		domain_address: DomainAddress,
 		message: T::Message,
 		expected_proof_count: u32,
```
lemunozm, line 28: Could not always the `expected_proof_count` be infered by knowing the number of routers under `domain_address`?

lemunozm (commented):
Thanks Cosmin! I like the new direction. A couple of questions

Question 1. Can you elaborate on why we need the `InboundEntry` as an enum?

Could we not have something like `(Option<Message>, SessionId, DomainAddress, count)`, avoiding to deal with the enum variants?

Question 2. How are we ensuring two proofs are not comming from the same router? Should not that information be stored somehow in the `InboundEntry`? (maybe a list of `RouterId` of the read proofs or similar)

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: Awesome! yeah, that was the intention 🙌🏻

cdamian:
> Question 1. Can you elaborate on why we need the InboundEntry as an enum?
>  Could we not have something like (Option<Message>, SessionId, DomainAddress, count), avoiding to deal with the enum variants?

The enum made sense to me since it's a lot easier to deal with a struct-like entity like this instead of a tuple. Furthermore, it creates a clear distinction between message and message proof entries. Given that we have certain rules in place for defining which router can send what, such as [these](https://github.com/centrifuge/centrifuge-chain/blob/7c0678b06fb267f352a22f7e758b12b78a7293f9/pallets/liquidity-pools-gateway/src/message_processing.rs#L108), this distinction allows us to perform validation such as [this](https://github.com/centrifuge/centrifuge-chain/blob/7c0678b06fb267f352a22f7e758b12b78a7293f9/pallets/liquidity-pools-gateway/src/message_processing.rs#L184).

cdamian:
> Question 2. How are we ensuring two proofs are not comming from the same router? Should not that information be stored somehow in the InboundEntry? (maybe a list of RouterId of the read proofs or similar)

We don't need to ensure this, if we would, we wouldn't be able to support duplicate messages.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Yes, but if we want to support duplicate messages and ensure they don't get lost, we need to keep track of this and increment it accordingly, as per [this](https://github.com/centrifuge/centrifuge-chain/blob/feat/lp-v2-gateway-multi-router-drop-session-id-invalidation/pallets/liquidity-pools-gateway/src/message_processing.rs#L178).

lemunozm:
> struct-like entity like this instead of a tuple 

Yeah, when I wrote a tuple I wanted to write mostly a struct.

> We don't need to ensure this, if we would, we wouldn't be able to support duplicate messages.

I'm pretty sure we need it... This is where the "vote" concept comes into play. Suppose you have 2 duplicated messages. Then you dispatch the first one to be processed once you have at least one vote from all associated routers. At least this is how it works in Solidity.

cdamian:
> > struct-like entity like this instead of a tuple
> 
> Yeah, when I wrote a tuple I wanted to write mostly a struct.
> 
> > We don't need to ensure this, if we would, we wouldn't be able to support duplicate messages.
> 
> I'm pretty sure we need it... This is where the "vote" concept comes into play. Suppose you have 2 duplicated messages. Then you dispatch the first one to be processed once you have at least one vote from all associated routers. At least this is how it works in Solidity.

We should expect multiple proofs from the same router. Given a setup with 3 routers x, y, z, where x is supposed to send the message, y and z are supposed to send the proofs, we would expect 1 message + 2 proofs in order to process 1 message. Given this, we can have a scenario where we receive 4 proofs from y and z, 2 each, and then 2 messages from x.

cdamian:
> In the current implementation, we can receive a message from x, two proofs from y, and the message will be processed without waiting for z, or am I missing something?

No, we also have a test that covers the case that you mentioned [here](https://github.com/centrifuge/centrifuge-chain/blob/feat/lp-v2-gateway-multi-router-drop-session-id-invalidation/pallets/liquidity-pools-gateway/src/tests.rs#L1937).

The logic that counts votes can be found [here](https://github.com/centrifuge/centrifuge-chain/blob/feat/lp-v2-gateway-multi-router-drop-session-id-invalidation/pallets/liquidity-pools-gateway/src/message_processing.rs#L232).

lemunozm:
Ah ok, you're right!

So, to double-check. The current implementation can:
- Receive multiple same messages.
- Receive multiple proofs for the same messages.
- Will only process a message when at least one proof for each router is received.

cdamian:
> Ah ok, you're right! I didn't see well that part 👍🏻
> 
> So, to double-check. The current implementation can:
> 
> * Receive multiple same messages.
> * Receive multiple proofs for the same messages.
> * Will only process a message when at least one proof for _each_ router is received.

Correct!

cdamian:
Anyway @lemunozm, this PR is dealing with some other changes not necessarily related to votes etc. Do you mind if I merge this in the main branch and continue discussions there?

lemunozm:
Also, I think now `InboundEntry` makes more sense to be an enum 👍🏻

PR #1967, opened 2024-08-14
Refactor: Unify DomainAddress and account conversions

lemunozm:
# Description
- Removed account conversions though `Convert` traits. Not `DomainAddress` know about to convert accounts.
- Unify DomainAddress to `[u8;32]`. See this [slack thread](https://kflabs.slack.com/archives/C06FRHCRQL9/p1723627567780339)
  - Before this PR you can transform a `DomainAddress` into `[u8;32]` in two ways:
     - Using `T::DomainAddressIntoAccount::convert()` which put 20 bytes from Eth + chain + "EVM" number.
     - Using `DomainAddress.address()` which put 20 bytes following by 0s.
     
    This is pretty confusing because is easy to use one or the other in different places, ending with different arrays values. From this PR, the second option is removed. We always add the extra information each time we extract the array from a `DomainAddress` 
      - We need to ensure this is not a problem with out current chain state. From Solidity it seems not because Solidity always reads the first 20 bytes.
- Added sugar method to `DomainAddress` to easy work with them safely
- Renamed `Centrifuge` to `Local` (or maybe I revert this)
- Renamed `EVM` to `Evm`
- (only in IT). Removed `From/Into` traits from `Keyring` to `H160` and `[u8;20]` and use instead an explicit method called `in_eth()` (open to name proposals). This method **represents the account in the Ethereum network**. Note that this differs from `Keyring::Alice.into()[0..20]`, which is used in other places where an ethereum account is stored as an `AccountId` and needs to be recovered. Because this is highly confusing, I've removed the `into/from` methods to be more explicit about when the first case happens.

### DIFF
--- a/libs/types/src/domain_address.rs
+++ b/libs/types/src/domain_address.rs
@@ -10,88 +10,99 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU General Public License for more details.
 
-use cfg_utils::vec_to_fixed_array;
 use frame_support::pallet_prelude::RuntimeDebug;
 use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
-use sp_runtime::traits::AccountIdConversion;
+use sp_runtime::{traits::AccountIdConversion, TypeId};
 
 use crate::EVMChainId;
 
+const MAX_ADDRESS_SIZE: usize = 32;
+pub type LocalAddress = [u8; 32];
+pub type EthAddress = [u8; 20];
+
+pub fn local_to_eth_address(address: LocalAddress) -> EthAddress {
+	*address
+		.split_first_chunk::<20>()
+		.expect("always fit, qed")
+		.0
+}
+
+pub fn evm_to_local_address(chain_id: u64, address: EthAddress) -> LocalAddress {
+	// We use a custom encoding here rather than relying on
+	// `AccountIdConversion` for a couple of reasons:
+	// 1. We have very few bytes to spare, so choosing our own fields is nice
+	// 2. AccountIdConversion puts the tag first, which can unbalance the storage
+	//    trees if users create many H160-derived accounts. We put the tag last
+	//    here.
+	let tag = b"EVM";
+	let mut bytes = [0; 32];
+	bytes[0..20].copy_from_slice(&address);
+	bytes[20..28].copy_from_slice(&chain_id.to_be_bytes());
+	bytes[28..31].copy_from_slice(tag);
+	bytes
+}
+
 /// A Domain is a chain or network we can send a message to.
-#[derive(Encode, Decode, Clone, Eq, MaxEncodedLen, PartialEq, RuntimeDebug, TypeInfo)]
+#[derive(Encode, Decode, Clone, Copy, Eq, MaxEncodedLen, PartialEq, RuntimeDebug, TypeInfo)]
 pub enum Domain {
-	/// Referring to the Centrifuge Parachain. Will be used for handling
-	/// incoming messages.
-	///
-	/// NOTE: messages CAN NOT be sent directly from the Centrifuge chain to the
-	/// Centrifuge chain itself.
-	Centrifuge,
+	/// Referring to the Local Parachain.
+	Local,
 	/// An EVM domain, identified by its EVM Chain Id
-	EVM(EVMChainId),
+	Evm(EVMChainId),
+}
+
+impl TypeId for Domain {
+	const TYPE_ID: [u8; 4] = crate::ids::DOMAIN_ID;
 }
 
 impl Domain {
 	pub fn into_account<AccountId: Encode + Decode>(&self) -> AccountId {
-		DomainLocator {
-			domain: self.clone(),
-		}
-		.into_account_truncating()
+		self.into_account_truncating()
 	}
-}
-
-#[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
-pub struct DomainLocator<Domain> {
-	pub domain: Domain,
 }
 
 #[derive(Encode, Decode, Clone, Eq, MaxEncodedLen, PartialEq, RuntimeDebug, TypeInfo)]
 pub enum DomainAddress {
-	/// A Centrifuge-Chain based account address, 32-bytes long
-	Centrifuge([u8; 32]),
-	/// An EVM chain address, 20-bytes long
-	EVM(EVMChainId, [u8; 20]),
+	/// A local based account address
+	Local(LocalAddress),
+	/// An EVM chain address
+	Evm(EVMChainId, EthAddress),
+}
+
+impl TypeId for DomainAddress {
+	const TYPE_ID: [u8; 4] = crate::ids::DOMAIN_ADDRESS_ID;
 }
 
 impl Default for DomainAddress {
 	fn default() -> Self {
-		DomainAddress::Centrifuge([0; 32])
-	}
-}
-
-impl DomainAddress {
-	pub fn evm(chain_id: EVMChainId, address: [u8; 20]) -> Self {
-		Self::EVM(chain_id, address)
-	}
-
-	pub fn centrifuge(address: [u8; 32]) -> Self {
-		Self::Centrifuge(address)
-	}
-}
-
-impl From<(EVMChainId, [u8; 20])> for DomainAddress {
-	fn from((chain_id, address): (EVMChainId, [u8; 20])) -> Self {
-		Self::evm(chain_id, address)
+		DomainAddress::Local(LocalAddress::default())
 	}
 }
 
 impl From<DomainAddress> for Domain {
 	fn from(x: DomainAddress) -> Self {
 		match x {
-			DomainAddress::Centrifuge(_) => Domain::Centrifuge,
-			DomainAddress::EVM(chain_id, _) => Domain::EVM(chain_id),
+			DomainAddress::Local(_) => Domain::Local,
+			DomainAddress::Evm(chain_id, _) => Domain::Evm(chain_id),
 		}
 	}
 }
 
 impl DomainAddress {
-	/// Get the address in a 32-byte long representation.
-	/// For EVM addresses, append 12 zeros.
-	pub fn address(&self) -> [u8; 32] {
-		match self.clone() {
-			Self::Centrifuge(x) => x,
-			Self::EVM(_, x) => vec_to_fixed_array(x),
+	pub fn new(domain: Domain, address: [u8; MAX_ADDRESS_SIZE]) -> Self {
+		match domain {
+			Domain::Local => DomainAddress::Local(LocalAddress::from(address)),
+			Domain::Evm(chain_id) => DomainAddress::Evm(chain_id, local_to_eth_address(address)),
 		}
+	}
+
+	pub fn from_local(address: impl Into<LocalAddress>) -> DomainAddress {
+		DomainAddress::Local(address.into())
+	}
+
+	pub fn from_evm(chain_id: EVMChainId, address: impl Into<EthAddress>) -> DomainAddress {
+		DomainAddress::Evm(chain_id, address.into())
 	}
 
 	pub fn domain(&self) -> Domain {
@@ -99,24 +110,20 @@ impl DomainAddress {
 	}
 }
 
-#[cfg(test)]
-mod tests {
-	use parity_scale_codec::{Decode, Encode};
-
-	use super::*;
-
-	#[test]
-	fn test_domain_encode_decode() {
-		test_domain_identity(Domain::Centrifuge);
-		test_domain_identity(Domain::EVM(1284));
-		test_domain_identity(Domain::EVM(1));
+impl DomainAddress {
+	pub fn as_local<Address: From<LocalAddress>>(&self) -> Address {
+		match self.clone() {
+			Self::Local(x) => x,
+			Self::Evm(chain_id, x) => evm_to_local_address(chain_id, x),
+		}
+		.into()
 	}
 
-	/// Test that (decode . encode) results in the original value
-	fn test_domain_identity(domain: Domain) {
-		let encoded = domain.encode();
-		let decoded = Domain::decode(&mut encoded.as_slice()).unwrap();
-
-		assert_eq!(domain, decoded);
+	pub fn as_eth<Address: From<EthAddress>>(&self) -> Address {
+		match self.clone() {
+			Self::Local(x) => local_to_eth_address(x),
+			Self::Evm(_, x) => x,
+		}
+		.into()
 	}
 }
--- a/pallets/liquidity-pools/src/hooks.rs
+++ b/pallets/liquidity-pools/src/hooks.rs
@@ -12,7 +12,7 @@
 // GNU General Public License for more details.
 
 use cfg_traits::{investments::ForeignInvestmentHooks, liquidity_pools::OutboundMessageHandler};
-use cfg_types::domain_address::DomainAddress;
+use cfg_types::domain_address::{Domain, LocalAddress};
 use frame_support::traits::{
 	fungibles::Mutate,
 	tokens::{Fortitude, Precision, Preservation},
@@ -24,7 +24,7 @@ use crate::{pallet::Config, Message, Pallet};
 
 impl<T: Config> ForeignInvestmentHooks<T::AccountId> for Pallet<T>
 where
-	<T as frame_system::Config>::AccountId: Into<[u8; 32]>,
+	T::AccountId: From<LocalAddress> + Into<LocalAddress>,
 {
 	type Amount = T::Balance;
 	type CurrencyId = T::CurrencyId;
@@ -39,8 +39,8 @@ where
 		fulfilled: Self::Amount,
 	) -> DispatchResult {
 		let currency = Pallet::<T>::try_get_general_index(currency_id)?;
-		let wrapped_token = Pallet::<T>::try_get_wrapped_token(&currency_id)?;
-		let domain_address: DomainAddress = wrapped_token.into();
+		let (chain_id, ..) = Pallet::<T>::try_get_wrapped_token(&currency_id)?;
+		let domain = Domain::Evm(chain_id);
 
 		T::Tokens::burn_from(
 			currency_id,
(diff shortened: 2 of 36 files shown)


lemunozm:
Current implementation is done as:

```rust
pub enum DomainAddress
    Centrifuge([u8; 32]),		
    Evm(EVMChainId, [u8; 20]),
}
```

But I'm thinking of changing this into:
```rust
pub enum DomainAddress
    Centrifuge(AccountId32),		
    Evm(EVMChainId, H160),
}
```

I do not like that we expose the `AccountId32` type (which, theoretically, is set at runtime), but we would gain much more type safety. It's a pain to deal with arrays. Maybe it makes sense to expose an `AccountId32` because speaking about a "Centrifuge domain" inherently carries some "domain information"

runtime/common/src/gateway.rs
```diff
@@ -10,22 +10,18 @@
...
-	H160::from_slice(&<AccountId32 as AsRef<[u8; 32]>>::as_ref(&sender_account)[0..20])
-}
+	let sender_account: AccountId32 = Sibling::from(para_id).into_account_truncating();
 
-pub fn get_gateway_account<T: pallet_evm_chain_id::Config + staging_parachain_info::Config>(
-) -> DomainAddress {
-	DomainAddress::Centrifuge(
-		AccountConverter::evm_address_to_account::<T>(get_gateway_h160_account::<T>()).into(),
-	)
+	DomainAddress::new(Domain::Evm(chain_id), sender_account.into())
```
lemunozm, line 26: Don't be scared; although it doesn't seem idempotent, only the first 20 bytes are used in `pallet-axelar-router`, and they remain the same.

libs/types/src/domain_address.rs
```diff
@@ -10,113 +10,120 @@
...
+#[derive(Encode, Decode, Clone, Copy, Eq, MaxEncodedLen, PartialEq, RuntimeDebug, TypeInfo)]
 pub enum Domain {
-	/// Referring to the Centrifuge Parachain. Will be used for handling
-	/// incoming messages.
-	///
-	/// NOTE: messages CAN NOT be sent directly from the Centrifuge chain to the
-	/// Centrifuge chain itself.
-	Centrifuge,
+	/// Referring to the Local Parachain.
+	Local,
```
cdamian, line 50: I prefer `Centrifuge` since it's a clear distinction between the domains. Local can also mean local EVM in some cases. Those are my 2 cents :D

lemunozm: Interesting! You could be right 🤔

lemunozm: To be aligned, what do you mean exactly with local EVM?

cdamian: The contracts that we have deployed in the centrifuge-evm.

wischli: I also prefer `Centrifuge` as it is more explicit than `Local` which is relative.

libs/types/src/domain_address.rs
```diff
@@ -10,113 +10,120 @@
...
-			Self::Centrifuge(x) => x,
-			Self::EVM(_, x) => vec_to_fixed_array(x),
+	pub fn new(domain: Domain, address: [u8; MAX_ADDRESS_SIZE]) -> Self {
+		match domain {
+			Domain::Local => DomainAddress::Local(LocalAddress::from(address)),
+			Domain::Evm(chain_id) => DomainAddress::Evm(chain_id, local_to_eth_address(address)),
 		}
 	}
 
+	pub fn from_local(address: impl Into<LocalAddress>) -> DomainAddress {
```
wischli, line 100: Nit: I am not sure I am fully happy with the naming here because the returned domain address implicitly local even though it could also be EVM from a 32 byte array input.

libs/types/src/domain_address.rs
```diff
@@ -10,113 +10,120 @@
...
-use sp_runtime::traits::AccountIdConversion;
+use sp_runtime::{traits::AccountIdConversion, TypeId};
 
 use crate::EVMChainId;
 
+const MAX_ADDRESS_SIZE: usize = 32;
+pub type LocalAddress = [u8; 32];
+pub type EthAddress = [u8; 20];
+
+pub fn local_to_eth_address(address: LocalAddress) -> EthAddress {
```
wischli, line 24: Nit: Let's stick to one namespace for Eth, Evm, Solidity
```suggestion
pub fn local_to_evm_address(address: LocalAddress) -> EthAddress {
```

libs/types/src/domain_address.rs
```diff
@@ -10,113 +10,120 @@
...
+		.into()
 	}
 
-	/// Test that (decode . encode) results in the original value
-	fn test_domain_identity(domain: Domain) {
-		let encoded = domain.encode();
-		let decoded = Domain::decode(&mut encoded.as_slice()).unwrap();
-
-		assert_eq!(domain, decoded);
+	pub fn as_eth<Address: From<EthAddress>>(&self) -> Address {
```
wischli, line 122: Nit: Namespace
```suggestion
	pub fn as_evm<Address: From<EthAddress>>(&self) -> Address {
```

pallets/liquidity-pools/src/hooks.rs
```diff
@@ -74,13 +70,13 @@ where
...
-		let wrapped_token = Pallet::<T>::try_get_wrapped_token(&currency_id)?;
-		let domain_address: DomainAddress = wrapped_token.into();
+		let (chain_id, ..) = Pallet::<T>::try_get_wrapped_token(&currency_id)?;
+		let domain = Domain::Evm(chain_id);
 
 		T::Tokens::transfer(
 			(pool_id, tranche_id).into(),
 			who,
-			&domain_address.domain().into_account(),
+			&domain.into_account(),
```
wischli, line 79: Great improvement, love it!

pallets/liquidity-pools/src/message.rs
```diff
@@ -740,18 +740,18 @@ mod tests {
...
+	fn transfer_tranche_tokens_to_chain() {
+		let domain_address = DomainAddress::Evm(1284, default_address_20());
 
 		test_encode_decode_identity(
 			Message::TransferTrancheTokens {
 				pool_id: 1,
 				tranche_id: default_tranche_id(),
 				domain: domain_address.domain().into(),
-				receiver: domain_address.address(),
+				receiver: domain_address.as_local(),
```
wischli, line 751: I am skeptical about the changed serialization here. Probably should be
```suggestion
				receiver: domain_address.as_eth(),
```

pallets/liquidity-pools/src/mock.rs
```diff
@@ -26,24 +26,12 @@ pub const CHAIN_ID: u64 = 1;
 pub const ALICE_32: [u8; 32] = [2; 32];
 pub const ALICE: AccountId = AccountId::new(ALICE_32);
 pub const ALICE_ETH: [u8; 20] = [2; 20];
-pub const ALICE_EVM_DOMAIN_ADDRESS: DomainAddress = DomainAddress::EVM(CHAIN_ID, ALICE_ETH);
-// TODO(future): Can be removed after domain conversion refactor
```
wischli, line 30: 🫶

pallets/liquidity-pools/src/mock.rs
```diff
@@ -26,24 +26,12 @@ pub const CHAIN_ID: u64 = 1;
...
-	let mut arr = [0u8; 32];
-	let mut i = 0;
-	while i < 20 {
-		arr[i] = ALICE_ETH[i];
-		i += 1;
-	}
-	AccountId::new(arr)
-};
-pub const CENTRIFUGE_DOMAIN_ADDRESS: DomainAddress = DomainAddress::Centrifuge(ALICE_32);
+pub const ALICE_EVM_DOMAIN_ADDRESS: DomainAddress = DomainAddress::Evm(42, ALICE_ETH);
```
wischli, line 29: You re-inserted the incorrect chain ID which was fixed in #1964

```suggestion
pub const ALICE_EVM_DOMAIN_ADDRESS: DomainAddress = DomainAddress::Evm(CHAIN_ID, ALICE_ETH);
```

pallets/liquidity-pools/src/tests.rs
```diff
@@ -2016,17 +1990,16 @@ mod recover_assets {
 	const ASSET: [u8; 32] = [43; 32];
 
 	fn config_mocks() {
-		DomainAddressToAccountId::mock_convert(move |_| ALICE_EVM_LOCAL_ACCOUNT);
 		Permissions::mock_has(|_, _, _| false);
 		Gateway::mock_handle(|sender, destination, msg| {
 			assert_eq!(sender, TreasuryAccount::get());
-			assert_eq!(destination, EVM_DOMAIN);
+			assert_eq!(destination, Domain::Evm(42));
```
wischli, line 1996: See above
```suggestion
			assert_eq!(destination, EVM_DOMAIN);
```

runtime/common/src/gateway.rs
wischli: I am curious why you switched the return type of the function then instead of still returning 20 bytes?

runtime/integration-tests/src/utils/accounts.rs
```diff
@@ -45,7 +45,17 @@ impl Keyring {
...
 	/// NOTE: Needs to be executed in an externalities environment
 	pub fn id_ecdsa<T: pallet_evm_chain_id::Config>(self) -> AccountId32 {
-		AccountConverter::evm_address_to_account::<T>(self.into())
+		AccountConverter::evm_address_to_account::<T>(self.in_eth())
+	}
+
+	/// Returns the Ethereum address.
+	/// The H160 retrived is NOT the local representation of that account in our
+	/// chain, it's the real address in ethereum.
+	pub fn in_eth(self) -> H160 {
```
wischli, line 54: Nit: Namespace
```suggestion
	pub fn in_evm(self) -> H160 {
```

libs/types/src/domain_address.rs
lemunozm: The tech-commite has reached consensus then 😆

wischli (commented):
I wish we did this earlier because this would have saved me quite some time! Overall, looks really good to me. I just have a few nits and couple of questions.

One change request on which I cannot comment: Please revert the submodule change because you are going back one commit there: This PR wants to change `6e8f1a29dff0d7cf5ff74285cfffadae8a8b303f` which is origin/main to `4301885b9a3b8ec36f3bda4b789daa5b115c006a`

runtime/common/src/gateway.rs
lemunozm: I did this because this address is the `DomainAddress` that will be passed through the `MessageSender`, which uses as a `DomainAddress` as origin to use a symmetry with `MessageReceiver`, which requires a `DomainAddress`. Also, in the future, some router could do more inspection in the address to make decisions if we pass it as `DomainAddress`. So I've tried to pass the most generic type.

lemunozm: Anyways, we probably need to re-react this method in the future when new routers are added.

lemunozm:
> This PR wants to change 6e8f1a29dff0d7cf5ff74285cfffadae8a8b303f which is origin/main to 4301885b9a3b8ec36f3bda4b789daa5b115c006a

Good point! It was during the rebase...

runtime/common/src/gateway.rs
wischli: Thanks for the insights, makes sense!

libs/types/src/domain_address.rs
lemunozm: I wanted to distinguish for the name: 
- eth = just the address
- evm: and ethereum address + chain information

I say this because not sure for example if `[u8; 20]` is an evm address, I think not because it does not have chain information 🤔 

But not sure if this association makes sense TBH. It seems not 😅

libs/types/src/domain_address.rs
lemunozm: Thinking now about just `eth/evm`, WDYT?

pallets/liquidity-pools/src/message.rs
lemunozm: `address()` returned a `[u8; 32]`, the `receiver` has type `[u8; 32]` but `as_eth()` returns an [u8;20]. 

Basically we use `address` as 32-bytes in the message for compatibility reasons, but later the 12 remaining bytes are discarded by solidity, see this thread: https://kflabs.slack.com/archives/C06FRHCRQL9/p1723627567780339

runtime/integration-tests/src/utils/accounts.rs
lemunozm: Will change if you confirm [the above](https://github.com/centrifuge/centrifuge-chain/pull/1967#discussion_r1718562357) makes no sense, that probably not 😆

libs/types/src/domain_address.rs
lemunozm: If the user uses a 32-byte array (that represents an eth address expanded to 32 bytes) and sets this as local, then they are using the method wrongly. Or if not wrongly, they are creating an account from an array with X data, where X is a eth address in the array.

Not sure if I got what you want to say.

libs/types/src/domain_address.rs
wischli: I definitely get what you are saying but to me, having two namespaces just causes confusion.

However, since he address format is officially referred to as the _Ethereum address format_, it appears to me your distinction is correct. So fine with keeping as is.

pallets/liquidity-pools/src/message.rs
wischli: Thanks for breaking this down. Seems like the message serialization is now correct and was poorly mocked beforehand, i.e. by using the 20 byte account address which in reality, should never occur.

libs/types/src/domain_address.rs
lemunozm: Will add a comment regarding this then to explain the why. Thanks for the ref

pallets/liquidity-pools/src/message.rs
lemunozm: Yeah, previously it was a chaos of different versions of `[u8; 20] -> [u8; 32]`  conversions

lemunozm:
Given the change from `Local` to `Centrifuge`, I should also change the `as_local()` and `from_local()`, I think both could be: `as_account()` (or just `account()`) and `from_account()`.

Do you agree or do you have some preference? @wischli @cdamian

wischli:
> Given the change from `Local` to `Centrifuge`, I should also change the `as_local()` and `from_local()`, I think both could be: `as_account()` (or just `account()`) and `from_account()`.
> 
> Do you agree or do you have some preference? @wischli @cdamian

Sounds good to me! `{as, from}_centrifuge` certainly sounds wrong. If we want to be more explicit, we could also use `..account32..` or `..account_id..`.

lemunozm:
Well, I need to think a bit more because what's returned is the 32-bit array, so the account name may not make sense in some places where the array is expected (as when we construct a message with an eth address expanded to 32 bytes). Such a dilemma!

libs/types/src/domain_address.rs
cdamian: Frontier's approach is pretty straightforward when it comes to eth/evm addresses, they name these `AccountId20` - [here](https://github.com/polkadot-evm/frontier/blob/master/primitives/account/src/lib.rs#L42). 

As someone who didn't have to deal much with `Domain`/`DomainAddress`, I find that local/eth/evm are confusing as well tbh.

Why not go for something simpler like:
- `CentrifugeAccount` - `AccountId32`.
- `EvmAccount` - `AccountId20`/`[u8;20]`.
- `CentrifugeEvmAccount` - `AccountId32` that's basically `EvmAccount` + `chain_id` + `tag`.

Then we could have implementations for:
- `CentrifugeAccount` -> `EvmAccount` - truncate;
- `EvmAccount` -> `CentrifugeAccount` - pad with 0s;
- `CentrifugeAccount` <-> `CentrifugeEvmAccount` - no need;
- `EvmAccount` -> `CentrifugeEvmAccount` - append `chain_id` + `tag`;
- `CentrifugeEvmAccount` -> `EvmAccount` - truncate;

lemunozm: Thanks for sharing this!

And these `CentrifugeAccount,` etc, should be types on their own, right? So you would directly remove the `DomainAddress` type? Which type will you use to track all of them generically?

cdamian: No, I would add them to the `DomainAddress`:

```
pub enum DomainAddress {
	/// A standard Centrifuge account.
	Centrifuge(CentrifugeAccount),
	
	/// A Centrifuge account derived from an `EvmAccount`.
	CentrifugeEvm(CentrifugeEvmAccount),
	
	/// An EVM-compatible account.
	Evm(EvmAccount),
}
```

This way we can keep the above mentioned conversions, and just have `From` implementations for `DomainAddress`, and for each of these account types.

lemunozm: Ok, I like the way you're taking it. 

Just a problem with this. The variant `Evm(EvmAccount)` will not have a `Domain` information associated with it. A DomainAddress requires it to later extract this.

For sure, I would not have time to do such a refactor for Monday 😅

cdamian: We could add a struct for the evm account that has the chain and the actual account, or keep it as it is right now? I dunno, just thinking out loud at this point ^^

lemunozm: But that would be what we have, right? 🤔 

I'm going to sketch a new draft with some ideas from this, but reusing the current `DomainAddress`, let's see. Thanks for your thoughts!

pallets/liquidity-pools/src/mock.rs
lemunozm: Ok, this has changed because `ALICE_EVM_LOCAL_ACCOUNT` no longer exists, and now we're using `ALICE_EVM_LOCA_ACCOUNT`, which has chain 42, not sure why TBH...

lemunozm: Ah ok, never mind... fixing it

lemunozm: Thanks for being aware of this!

lemunozm:
Closed in favor of: https://github.com/centrifuge/centrifuge-chain/pull/1969

PR #1968, opened 2024-08-15
integration-tests: Register routers during setup

cdamian:


cdamian:
@lemunozm @wischli merged this directly since the change set wasn't as big as expected.

cdamian:
Also, didn't wanna end for another set of checks to be executed.

PR #1969, opened 2024-08-15
Refactor: Unify DomainAddress and account conversions (r2)

lemunozm:
This PRs overwrites https://github.com/centrifuge/centrifuge-chain/pull/1967

# Description

Represents `DomainAddress` as:
```rust
enum DomainAddress {
    Centrifuge(AccountId32),
    Evm(EVMChainId, H160),
}
```

With some utility methods as:
```rust
fn account() -> AccountId32;
fn h160() -> H160;
fn bytes() -> [u8; 32];
```

### Main changes
- Removed account conversions through `Convert` traits. Now `DomainAddress` knows how to convert accounts.
- Unify `[u8; 20]` to `[u8;32]` conversions. See this [slack thread](https://kflabs.slack.com/archives/C06FRHCRQL9/p1723627567780339)
- Renamed `EVM` to `Evm` (following rust guidelines)
- (only in IT). Removed `From/Into` traits from `Keyring` to `H160` and `[u8;20]` and use instead an explicit method called `in_eth()` (open to name proposals). This method **represents the account in the Ethereum network**. Note that this differs from `Keyring::Alice.into()[0..20]`, which is used in other places where an Ethereum account is stored as an `AccountId` and needs to be recovered. Because this is highly confusing, I've removed the `into/from` methods to be more explicit about when the first case happens.

### DIFF
--- a/libs/types/src/ids.rs
+++ b/libs/types/src/ids.rs
@@ -15,10 +15,7 @@
 use frame_support::PalletId;
 use sp_runtime::TypeId;
 
-use crate::{
-	domain_address::{DomainAddress, DomainLocator},
-	investments::InvestmentAccount,
-};
+use crate::investments::InvestmentAccount;
 
 // The TypeId impl we derive pool-accounts from
 impl<InvestmentId> TypeId for InvestmentAccount<InvestmentId> {
@@ -45,10 +42,5 @@ pub const CHAIN_BRIDGE_NATIVE_TOKEN_ID: [u8; 4] = *b"xCFG";
 /// The identifier of the group eligible to receive block rewards.
 pub const COLLATOR_GROUP_ID: u32 = 1;
 
-impl TypeId for DomainAddress {
-	const TYPE_ID: [u8; 4] = *b"dadr";
-}
-
-impl<Domain> TypeId for DomainLocator<Domain> {
-	const TYPE_ID: [u8; 4] = *b"domn";
-}
+pub const DOMAIN_ID: [u8; 4] = *b"dadr";
+pub const DOMAIN_ADDRESS_ID: [u8; 4] = *b"domn";
--- a/libs/types/src/ids.rs
+++ b/libs/types/src/ids.rs
@@ -42,5 +42,5 @@ pub const CHAIN_BRIDGE_NATIVE_TOKEN_ID: [u8; 4] = *b"xCFG";
 /// The identifier of the group eligible to receive block rewards.
 pub const COLLATOR_GROUP_ID: u32 = 1;
 
-pub const DOMAIN_ID: [u8; 4] = *b"dadr";
-pub const DOMAIN_ADDRESS_ID: [u8; 4] = *b"domn";
+pub const DOMAIN_ID: [u8; 4] = *b"domn";
+pub const DOMAIN_ADDRESS_ID: [u8; 4] = *b"dadr";
--- a/pallets/liquidity-pools/src/lib.rs
+++ b/pallets/liquidity-pools/src/lib.rs
@@ -43,7 +43,7 @@ use core::convert::TryFrom;
 
 use cfg_traits::{liquidity_pools::OutboundMessageHandler, swaps::TokenSwaps, PreConditions};
 use cfg_types::{
-	domain_address::{Domain, DomainAddress, EthAddress, LocalAddress},
+	domain_address::{Domain, DomainAddress},
 	tokens::GeneralCurrencyIndex,
 };
 use cfg_utils::vec_to_fixed_array;
@@ -59,6 +59,7 @@ use orml_traits::{
 	GetByKey,
 };
 pub use pallet::*;
+use sp_core::{crypto::AccountId32, H160};
 use sp_runtime::{
 	traits::{AtLeast32BitUnsigned, EnsureMul},
 	FixedPointNumber, SaturatedConversion,
@@ -131,7 +132,7 @@ pub mod pallet {
 	pub struct Pallet<T>(_);
 
 	#[pallet::config]
-	pub trait Config: frame_system::Config {
+	pub trait Config: frame_system::Config<AccountId = AccountId32> {
 		/// Weight information for extrinsics in this pallet.
 		type WeightInfo: WeightInfo;
 
@@ -346,10 +347,7 @@ pub mod pallet {
 	}
 
 	#[pallet::call]
-	impl<T: Config> Pallet<T>
-	where
-		T::AccountId: From<LocalAddress> + Into<LocalAddress>,
-	{
+	impl<T: Config> Pallet<T> {
 		/// Add a pool to a given domain.
 		///
 		/// Origin: Pool admin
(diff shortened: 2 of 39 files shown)


lemunozm:
@wischli @cdamian I tried my best from your inputs above. I would really love to have types who knows about conversions but it's quite tedious for Monday (we also will need implement them with a lot of traits, etc).

I think this solution leaves things easier than before, but of course, it still can be improved with more sophisticated solutions.

Let me know if you're agree more in the names than the before PR.

pallets/liquidity-pools/src/lib.rs
```diff
@@ -131,7 +132,7 @@ pub mod pallet {
 	pub struct Pallet<T>(_);
 
 	#[pallet::config]
-	pub trait Config: frame_system::Config {
+	pub trait Config: frame_system::Config<AccountId = AccountId32> {
```
lemunozm, line 135: Everything works fine in IT except for a compilation error with the fudge environment. I bet you, guys, to give it a try... I had never experienced it before.

libs/types/src/ids.rs
```diff
@@ -45,10 +42,5 @@ pub const CHAIN_BRIDGE_NATIVE_TOKEN_ID: [u8; 4] = *b"xCFG";
...
 
-impl TypeId for DomainAddress {
-	const TYPE_ID: [u8; 4] = *b"dadr";
-}
-
-impl<Domain> TypeId for DomainLocator<Domain> {
-	const TYPE_ID: [u8; 4] = *b"domn";
-}
+pub const DOMAIN_ID: [u8; 4] = *b"dadr";
+pub const DOMAIN_ADDRESS_ID: [u8; 4] = *b"domn";
```
cdamian, line 46: These 2 should be swapped?

pallets/liquidity-pools/src/lib.rs
cdamian: And I assume you restricted `AccountId` to `AccountId32` so that you can easily call `domain_address.account()`.

cdamian: But I think `domain_address.account()` should return `AccountId` instead of `AccountId32`.

libs/types/src/ids.rs
lemunozm: Oh man, blessing eyes yours 🙌🏻

pallets/liquidity-pools/src/lib.rs
lemunozm: > And I assume you restricted AccountId to AccountId32 so that you can easily call domain_address.account()

Yes, anyways, we were using `From<[u8; 32]>` and `Into<[u8; 32]>` as a restriction in that pallet.

> But I think domain_address.account() should return AccountId instead of AccountId32.

`AccountId` is only a type alias for `AccountId32`, which should not affect it. Ideally, `AccountId` is (should be) a runtime definition. I used the real type `AccountId32`, just to be more close to the baseline

lemunozm:
```rust
fn foo<T: Runtime + FudgeSupport>(f: ApiRefOf<T>) {} // compiles
```

```rust
fn bar<T: Runtime + FudgeSupport, F: FnOnce(ApiRefOf<T>)>(f: F) {} //no compiles
```

I'm reaching the boundaries of the rust compiler? 😲

lemunozm: Finally, I fixed getting the type in another, more simpler, way

runtime/integration-tests/src/cases/example.rs
```diff
@@ -68,7 +68,7 @@ fn transfer_balance<T: Runtime>() {
 }
 
 // Identical to `transfer_balance()` test but using fudge.
-#[test_runtimes([development, altair, centrifuge])]
+#[test_runtimes([development, altair, centrifuge], ignore = "uncomment to run the example")]
```
lemunozm, line 71: Saving some extra CI time with those tests ignored

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
```diff
@@ -1,17 +1,21 @@
...
 use cfg_types::domain_address::DomainAddress;
 use frame_support::{assert_ok, traits::OriginTrait};
 use pallet_liquidity_pools::Message;
 use pallet_liquidity_pools_gateway::message::GatewayMessage;
+use runtime_common::routing::{AxelarId, RouterId};
+use sp_core::H160;
 use sp_runtime::{traits::One, BoundedVec};
 
 use crate::{
-	cases::liquidity_pools::{DEFAULT_DOMAIN_ADDRESS_MOONBEAM, DEFAULT_ROUTER_ID},
```
lemunozm, line 9: I wanted to enforce the rule that things should not be imported from one `cases` file to another.

My rationale for doing this:
  - If `cases::liquidity_pools` needs to be refactorized due to some new feature, we're probably breaking also the queue tests (when we shouldn't).
  - If a reader wants to understand these `queue` tests, we force them to also look into `liquidity_pools` tests to fully understand when both are theoretically unrelated.

Of course, if the common code grows, we should factorize this into some well-through enough-generic utility module. Then, both cases can point to those logic without carrying mutual information.

cc @cdamian @wischli

runtime/integration-tests/src/envs/fudge_env.rs
```diff
@@ -46,8 +46,6 @@ impl<T: Runtime + FudgeSupport> Env<T> for FudgeEnv<T> {
 		parachain_storage: Storage,
 		sibling_storage: Storage,
 	) -> Self {
-		crate::utils::logs::init_logs();
```
lemunozm, line 49: I wanted to remove this because I really think it does not offer us anything to have it by default. Most errors are pallet_timestampt errors or migrations errors that do not affect the test case, offuscating the CI results and local tests.

99% of the test case failures are not something "found" in the logs. If the test requires more debugging, we can always enable it again with...

runtime/integration-tests/src/README.md
```diff
@@ -19,3 +19,13 @@ Both environment uses the "same" interface so jumping from one to the another sh
   without hidden initializations.
 - You can add any utility that helps to initialize states for others under `utils` folder.
   Again, focus in simplity but without side effects or hidden / non-obvious state changes.
+
+## Logging
+If you want to add logs to your use case for debugging purposes, simply add
+
+```rust
+crate::utils::logs::init_logs();
```
lemunozm, line 27: ...just adding this line

wischli: Thanks for reverting the default logger which has been super noisy.

runtime/integration-tests/src/cases/liquidity_pools_gateway_queue.rs
wischli: Thanks for bringin this up. I definitely agree!

wischli (approved):
Once again, thanks for cleaning up this mess! I'd be glad to merge this, because its needed for #1971. 

As for the naming of `fn bytes()`, I am not 100% on board but that can be changed at any time, if at all, because it has zero impact.

lemunozm:
> As for the naming of fn bytes(), I am not 100% on board

I wanted to not use `address` because it's quite generic, not just an array, and for that method we're specifying that we want the account as an array of bytes. 

Rust names `*_bytes*` to actions that return an array, but I am open to other names, I can do it in a next PR if needed

cdamian (approved):

lemunozm:
I'm very happy to have this PR before the audit! 🎉

PR #1970, opened 2024-08-16
lp-gateway: Add extrinsics for initiating and disputing message recovery

cdamian:
# Description

Added:
- new extrinsics for initiating and disputing message recovery.
- new gateway events when processing inbound and outbound messages.

Changed:
- renamed `LPEncoding` to `LPMessage`.
   - added new methods from getting dispute messages.
   - changed message hash method.

# Checklist:

- [x] Added extra events for LP gateway, LP.
- [x] Removed sender field for outbound gateway message.

### DIFF
--- a/libs/traits/src/liquidity_pools.rs
+++ b/libs/traits/src/liquidity_pools.rs
@@ -15,11 +15,12 @@ use frame_support::{dispatch::DispatchResult, weights::Weight};
 use sp_runtime::DispatchError;
 use sp_std::vec::Vec;
 
-pub type Proof = [u8; 32];
+/// Type that represents the hash of an LP message.
+pub type MessageHash = [u8; 32];
 
 /// An encoding & decoding trait for the purpose of meeting the
 /// LiquidityPools General Message Passing Format
-pub trait LPEncoding: Sized {
+pub trait LPMessage: Sized {
 	fn serialize(&self) -> Vec<u8>;
 	fn deserialize(input: &[u8]) -> Result<Self, DispatchError>;
 
@@ -34,11 +35,23 @@ pub trait LPEncoding: Sized {
 	/// It's the identity message for composing messages with pack_with
 	fn empty() -> Self;
 
-	/// Retrieves the message proof hash, if the message is a proof type.
-	fn get_proof(&self) -> Option<Proof>;
+	/// Retrieves the message hash, if the message is a proof type.
+	fn get_message_hash(&self) -> Option<MessageHash>;
 
 	/// Converts the message into a message proof type.
 	fn to_proof_message(&self) -> Self;
+
+	/// Creates a message used for initiating message recovery.
+	///
+	/// Hash - hash of the message that should be recovered.
+	/// Router - the address of the recovery router.
+	fn initiate_message_recovery_message(hash: [u8; 32], router: [u8; 20]) -> Self;
+
+	/// Creates a message used for disputing message recovery.
+	///
+	/// Hash - hash of the message that should be disputed.
+	/// Router - the address of the recovery router.
+	fn dispute_message_recovery_message(hash: [u8; 32], router: [u8; 20]) -> Self;
 }
 
 pub trait RouterProvider<Domain>: Sized {
--- a/libs/traits/src/liquidity_pools.rs
+++ b/libs/traits/src/liquidity_pools.rs
@@ -45,13 +45,13 @@ pub trait LPMessage: Sized {
 	///
 	/// Hash - hash of the message that should be recovered.
 	/// Router - the address of the recovery router.
-	fn initiate_message_recovery_message(hash: [u8; 32], router: [u8; 20]) -> Self;
+	fn initiate_recovery_message(hash: [u8; 32], router: [u8; 20]) -> Self;
 
 	/// Creates a message used for disputing message recovery.
 	///
 	/// Hash - hash of the message that should be disputed.
 	/// Router - the address of the recovery router.
-	fn dispute_message_recovery_message(hash: [u8; 32], router: [u8; 20]) -> Self;
+	fn dispute_recovery_message(hash: [u8; 32], router: [u8; 20]) -> Self;
 }
 
 pub trait RouterProvider<Domain>: Sized {
--- a/libs/traits/src/liquidity_pools.rs
+++ b/libs/traits/src/liquidity_pools.rs
@@ -45,13 +45,13 @@ pub trait LPMessage: Sized {
 	///
 	/// Hash - hash of the message that should be recovered.
 	/// Router - the address of the recovery router.
-	fn initiate_recovery_message(hash: [u8; 32], router: [u8; 20]) -> Self;
+	fn initiate_recovery_message(hash: [u8; 32], router: [u8; 32]) -> Self;
 
 	/// Creates a message used for disputing message recovery.
 	///
 	/// Hash - hash of the message that should be disputed.
 	/// Router - the address of the recovery router.
-	fn dispute_recovery_message(hash: [u8; 32], router: [u8; 20]) -> Self;
+	fn dispute_recovery_message(hash: [u8; 32], router: [u8; 32]) -> Self;
 }
 
 pub trait RouterProvider<Domain>: Sized {
--- a/libs/traits/src/liquidity_pools.rs
+++ b/libs/traits/src/liquidity_pools.rs
@@ -35,8 +35,11 @@ pub trait LPMessage: Sized {
 	/// It's the identity message for composing messages with pack_with
 	fn empty() -> Self;
 
+	/// Returns whether the message is a proof or not.
+	fn is_proof_message(&self) -> bool;
+
 	/// Retrieves the message hash, if the message is a proof type.
-	fn get_message_hash(&self) -> Option<MessageHash>;
+	fn get_message_hash(&self) -> MessageHash;
 
 	/// Converts the message into a message proof type.
 	fn to_proof_message(&self) -> Self;
--- a/pallets/liquidity-pools-gateway/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/src/lib.rs
@@ -166,21 +166,27 @@ pub mod pallet {
 
 		/// An inbound message was processed.
 		InboundMessageProcessed {
+			domain_address: DomainAddress,
 			message_hash: MessageHash,
 			router_id: T::RouterId,
 		},
 
 		/// An inbound message proof was processed.
 		InboundProofProcessed {
+			domain_address: DomainAddress,
 			message_hash: MessageHash,
 			router_id: T::RouterId,
 		},
 
 		/// An inbound message was executed.
-		InboundMessageExecuted { message_hash: MessageHash },
+		InboundMessageExecuted {
+			domain_address: DomainAddress,
+			message_hash: MessageHash,
+		},
 
 		/// An outbound message was sent.
 		OutboundMessageSent {
+			domain_address: DomainAddress,
 			message_hash: MessageHash,
 			router_id: T::RouterId,
 		},
--- a/pallets/liquidity-pools-gateway/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/src/lib.rs
@@ -662,12 +662,9 @@ pub mod pallet {
 
 					(res, weight)
 				}
-				GatewayMessage::Outbound {
-					sender,
-					message,
-					router_id,
-				} => {
-					let res = T::MessageSender::send(router_id, sender, message.serialize());
+				GatewayMessage::Outbound { message, router_id } => {
+					let res =
+						T::MessageSender::send(router_id, T::Sender::get(), message.serialize());
 
 					(res, LP_DEFENSIVE_WEIGHT)
 				}
--- a/pallets/liquidity-pools-gateway/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/src/lib.rs
@@ -44,7 +44,10 @@ use sp_arithmetic::traits::{BaseArithmetic, EnsureAddAssign, One};
 use sp_runtime::SaturatedConversion;
 use sp_std::{convert::TryInto, vec::Vec};
 
-use crate::{message_processing::InboundEntry, weights::WeightInfo};
+use crate::{
+	message_processing::{InboundEntry, ProofEntry},
+	weights::WeightInfo,
+};
 
 mod origin;
 pub use origin::*;
@@ -63,7 +66,6 @@ mod tests;
 #[frame_support::pallet]
 pub mod pallet {
 	use super::*;
-	use crate::message_processing::ProofEntry;
 
 	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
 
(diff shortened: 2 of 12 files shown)


pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -461,25 +482,105 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		/// Sends a message that initiates a message recovery using the
+		/// messaging router.
+		///
+		/// Can only be called by `AdminOrigin`.
+		#[pallet::weight(T::WeightInfo::initiate_message_recovery())]
+		#[pallet::call_index(12)]
+		pub fn initiate_message_recovery(
```
cdamian, line 512: cc @hieronx

pallets/liquidity-pools/src/message.rs
```diff
@@ -226,7 +226,7 @@ pub enum Message<BatchContent = BatchMessages> {
 		/// The hash of the message which shall be recovered
```
cdamian, line 226: As of now, these are only used for outbound - Centrifuge -> Evm. Do we also expect these for inbound - Evm -> Centrifuge? @hieronx

hieronx: That is right, `Initiate/DisputeMessageRecovery` are only Centrigue -> EVM

pallets/liquidity-pools-gateway/src/lib.rs
hieronx: This looks right to me!

Only note: `Adminorigin` is our LP admin right? This should be `ensure_root`, like `Schedule/CancelUpgrade`.

cdamian: It's:

https://github.com/centrifuge/centrifuge-chain/blob/d1d823acbb9dbc0a1d7ac0aa7679ee30b34320c7/runtime/centrifuge/src/lib.rs#L1865

hieronx: That should work too!

wischli: Just to be 100% sure since you originally did not want to give that power to the LpAdmin: This includes three parties:
1. `Root`
2. 2/3 of council
3. Our LPAdminAcc

@hieronx

libs/traits/src/liquidity_pools.rs
```diff
@@ -15,11 +15,12 @@ use frame_support::{dispatch::DispatchResult, weights::Weight};
...
 use sp_std::vec::Vec;
 
-pub type Proof = [u8; 32];
+/// Type that represents the hash of an LP message.
+pub type MessageHash = [u8; 32];
 
 /// An encoding & decoding trait for the purpose of meeting the
 /// LiquidityPools General Message Passing Format
-pub trait LPEncoding: Sized {
+pub trait LPMessage: Sized {
```
cdamian, line 23: Finally renamed to `LPMessage` =D

libs/traits/src/liquidity_pools.rs
```diff
@@ -15,11 +15,12 @@ use frame_support::{dispatch::DispatchResult, weights::Weight};
 use sp_runtime::DispatchError;
 use sp_std::vec::Vec;
 
-pub type Proof = [u8; 32];
+/// Type that represents the hash of an LP message.
+pub type MessageHash = [u8; 32];
```
cdamian, line 19: In hindsight, this name makes more sense now.

pallets/liquidity-pools/src/message.rs
```diff
@@ -226,7 +226,7 @@ pub enum Message<BatchContent = BatchMessages> {
...
 		hash: [u8; 32],
 		/// The address of the router
-		address: Address,
+		router: [u8; 20],
```
cdamian, line 229: Can anyone please confirm that changing these names is OK?

pallets/liquidity-pools-gateway/src/lib.rs
hieronx: I think that is fine for now. We should probably remove the LPAdmin in the next 6 months or so and just use `Root` for all of this. But until then it works as is.

pallets/liquidity-pools/src/message.rs
lemunozm: I don't think so. The deserialize part in Solidity would expect 32 bytes 🤔, unless this message has special treatment
cc @hieronx

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -293,24 +293,24 @@ impl<T: Config> Pallet<T> {
...
+	pub(crate) fn get_message_hash(message: T::Message) -> MessageHash {
+		match message.get_message_hash() {
 			None => message
 				.to_proof_message()
-				.get_proof()
-				.expect("message proof ensured by 'to_message_proof'"),
+				.get_message_hash()
+				.expect("message hash ensured by 'to_proof_message'"),
 			Some(proof) => proof,
 		}
```
lemunozm, line 303: Really like the new name, but not sure if this reading process makes sense now here. Putting me in the mind of an auditor:

- I get the hash of a message
- If the message has no hash: (<- 🤔 could a message not have a hash?)
  - We create a proof of a message (<- 🤔  Why do I need to create a proof message to get the hash?)
  - And then we get the hash

pallets/liquidity-pools/src/message.rs
lemunozm: Thanks for explaining the whys! I didn't know the exact reason.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Yea, it might be a bit confusing, but I dunno how to solve this without adding another method to the trait like `has_message_hash`, then still do a call and do an `unwrap/expect` on it.

pallets/liquidity-pools/src/message.rs
cdamian: OK, I will revert the type, what about the field name then?

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: I'll update this in a bit and ping you.

libs/traits/src/liquidity_pools.rs
```diff
@@ -34,11 +35,26 @@ pub trait LPEncoding: Sized {
 	/// It's the identity message for composing messages with pack_with
 	fn empty() -> Self;
 
-	/// Retrieves the message proof hash, if the message is a proof type.
-	fn get_proof(&self) -> Option<Proof>;
+	/// Returns whether the message is a proof or not.
+	fn is_proof_message(&self) -> bool;
```
cdamian, line 39: @wischli @lemunozm I added this and changed `get_message_hash` below.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -162,11 +164,48 @@ pub mod pallet {
 			hook_address: [u8; 20],
 		},
```
cdamian, line 166: @hieronx the following events should roughly match the ones that you described:

```
event ProcessMessage(bytes message, address adapter);
event ProcessProof(bytes32 messageHash, address adapter);
event ExecuteMessage(bytes message, address adapter); // liquidityPools.IncomingMessage now
event SendMessage(bytes message); // liquidityPoolsGateway.MessageSubmitted now
event RecoverMessage(address adapter, bytes message);
event RecoverProof(address adapter, bytes32 messageHash);
```

We have a different naming convention for events and I wouldn't change that tbh, unless we change all of them...

Please note that:
- we're only using message hashes since using the actual bytes would require `BoundedVec`s.
- the adapter address is not something that we have access to at this point, I'm not sure how easy it would be to grab those for each router. Maybe @lemunozm @wischli can help.

hieronx: Sounds great, thanks for adapting this!

I just realized, should we add `Domain` to `InboundMessageProcessed`, `InboundProofProcessed`, and `OutboundMessageSent`? 

This is not included on the Solidity side  since it is always Centrifuge there. But here it seems we need it.

cdamian: >I just realized, should we add Domain to InboundMessageProcessed, InboundProofProcessed, and OutboundMessageSent?

Sure.

libs/traits/src/liquidity_pools.rs
lemunozm: I'm not fully sure about adding a new method... 🤔 

I do not want to impose my model, but why not this: https://github.com/centrifuge/centrifuge-chain/pull/1961#discussion_r1718314551, is there anything that does not match with the expected behavior?

cdamian: Not a big fan of the new method but we use it here now:
- https://github.com/centrifuge/centrifuge-chain/blob/feat/lp-v2-message-recovery/pallets/liquidity-pools-gateway/src/message_processing.rs#L466
- https://github.com/centrifuge/centrifuge-chain/blob/feat/lp-v2-message-recovery/pallets/liquidity-pools-gateway/src/message_processing.rs#L98

lemunozm: Ah ok 👍🏻 

Maybe my proposal doesn't fit because you need to forcibly check if the proof is proof. So you need 3 things:

- a `MessageHash`
- Create a `MessageProof`
- Check if the current message is a `MessageProof`

Yeah, let's go with this new method

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Added `domain_address` to all inbound / outbound events. Note that for outbound, that address is the `T::Sender`.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -461,25 +511,103 @@ pub mod pallet {
...
+		///
+		/// Can only be called by `AdminOrigin`.
+		#[pallet::weight(T::WeightInfo::dispute_message_recovery())]
+		#[pallet::call_index(13)]
+		pub fn dispute_message_recovery(
+			origin: OriginFor<T>,
+			domain: Domain,
+			message_hash: MessageHash,
+			recovery_router: [u8; 32],
+			messaging_router: T::RouterId,
```
lemunozm, line 575: Question. Should the `domain` field always be the domain belonging to the `messaging_router` passed?

If that is the case, we can extend `RouterId: Into<Domain>`, which is already implemented in `common/routing.rs` and save the user to write that parameter.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -461,25 +511,103 @@ pub mod pallet {
...
+		///
+		/// Can only be called by `AdminOrigin`.
+		#[pallet::weight(T::WeightInfo::initiate_message_recovery())]
+		#[pallet::call_index(12)]
+		pub fn initiate_message_recovery(
+			origin: OriginFor<T>,
+			domain: Domain,
+			message_hash: MessageHash,
+			recovery_router: [u8; 32],
+			messaging_router: T::RouterId,
```
lemunozm, line 546: Same here, and probably in any place where we're passing both `domain` and `router_id`, because `router_id` already carries the `domain` information

lemunozm (commented):
Super well organized and easy to review, thanks @cdamian

Just one thing, but apart from that, not a single extra NIT.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -461,25 +511,103 @@ pub mod pallet {
...
+		) -> DispatchResult {
+			let router_ids = Self::get_router_ids_for_domain(domain)?;
+
+			ensure!(
+				router_ids.iter().any(|x| x == &messaging_router),
+				Error::<T>::MessagingRouterNotFound
+			);
+
+			T::MessageSender::send(messaging_router, T::Sender::get(), message.serialize())
```
lemunozm, line 607: Question: does recovery messages need to go through different routers with proofs?

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: If agree, we can merge this anyways and do it in a next PR

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: No, for these kind of messages we don't send proofs.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Let's do it separately and not waste another CI run?

cdamian:
@lemunozm @wischli can we go ahead and merge it since everything is green? We can tackle the other item in a separate PR.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: For reference - [discussion](https://kflabs.slack.com/archives/C06FRHCRQL9/p1722934270095379)

lemunozm: Thanks for the reference!

lemunozm (approved):
Sure! Everything looks super great to me. Thanks Cosmin!

libs/traits/src/liquidity_pools.rs
```diff
@@ -34,11 +35,26 @@ pub trait LPEncoding: Sized {
...
+	fn get_message_hash(&self) -> MessageHash;
 
 	/// Converts the message into a message proof type.
 	fn to_proof_message(&self) -> Self;
+
+	/// Creates a message used for initiating message recovery.
+	///
+	/// Hash - hash of the message that should be recovered.
+	/// Router - the address of the recovery router.
+	fn initiate_recovery_message(hash: [u8; 32], router: [u8; 32]) -> Self;
```
wischli, line 51: Nit: Should use `MessageHash` type in signature
```suggestion
	fn initiate_recovery_message(hash: MessageHash, router: [u8; 32]) -> Self;
```

libs/traits/src/liquidity_pools.rs
```diff
@@ -34,11 +35,26 @@ pub trait LPEncoding: Sized {
...
+
+	/// Creates a message used for disputing message recovery.
+	///
+	/// Hash - hash of the message that should be disputed.
+	/// Router - the address of the recovery router.
+	fn dispute_recovery_message(hash: [u8; 32], router: [u8; 32]) -> Self;
```
wischli, line 57: Nit: Should use `MessageHash` type in signature
```suggestion
	fn dispute_recovery_message(hash: MessageHash, router: [u8; 32]) -> Self;
```

pallets/liquidity-pools-gateway/src/lib.rs
wischli: In 6 months we should be using OpenGov and then have the `Pool` track for those type of things.

wischli (approved):
Great work - LGTM! One minor nit that can be ignored or applied in another PR.

libs/traits/src/liquidity_pools.rs
cdamian: I'll add it in - https://github.com/centrifuge/centrifuge-chain/pull/1974

PR #1971, opened 2024-08-16
feat: LP V2 message forwarding

wischli:
# Description

Super early draft, please don't waste time reviewing this yet!

# Checklist:


libs/traits/src/liquidity_pools.rs
```diff
@@ -12,14 +12,17 @@
...
 use sp_std::vec::Vec;
 
 pub type Proof = [u8; 32];
 
 /// An encoding & decoding trait for the purpose of meeting the
 /// LiquidityPools General Message Passing Format
-pub trait LPEncoding: Sized {
+pub trait LpMessage: Sized {
+	type Domain;
+	type SerializableDomain;
```
wischli, line 24: @lemunozm Apologies for wrecking this trait. Regard this as WIP to get to an MVP. I am not sure yet how to properly solve this. I tried using generics for the two forwarding functions but didn't manage to please the compiler. If you have a recommendation, please speak up!

pallets/liquidity-pools/src/message.rs
```diff
@@ -530,9 +565,17 @@ pub enum Message<BatchContent = BatchMessages> {
 		/// The amount of tranche tokens which should be redeemed.
 		amount: u128,
 	},
+	Forwarded {
```
cdamian, line 568: This will be a new message type on the solidity side as well right?

wischli: Yes per https://kflabs.slack.com/archives/C06FRHCRQL9/p1723817828381219?thread_ts=1723816341.737369&cid=C06FRHCRQL9

Note: Not implemented yet in Solidity

libs/traits/src/liquidity_pools.rs
```diff
@@ -39,6 +42,16 @@ pub trait LPEncoding: Sized {
...
+
+	/// Unwraps a forwarded message
+	fn unwrap_forwarded(self) -> Option<(Self::Domain, H160, Self)>;
+
+	/// Attempts to wrap into a forwarded message
+	fn try_wrap_forward(
+		domain: Self::Domain,
+		forwarding_contract: H160,
+		message: Self,
+	) -> Result<Self, DispatchError>;
```
lemunozm, line 54: Like those methods!

libs/traits/src/liquidity_pools.rs
lemunozm: I am not sure if I follow, but I do not see the `SerializabbleDomain` associated type used by the trait

lemunozm: Regarding the domain, why not directly use a `Domain` type?

EDIT: ah because it's in `cfg-types`. I think maybe `Domain` should be treated as primitive, because all pallets use them directly. Still not having a well defined place where to put types lead us to not accurate designs, that's a pity 😢. Hopefully, after this audit we'll have time to do a cool refactor.

By now I think it's ok using `Domain` as associated type, but I think `SerializableDomain` can be removed, right?

libs/traits/src/liquidity_pools.rs
```diff
@@ -129,3 +142,38 @@ pub trait InboundMessageHandler {
...
+	/// The originator of the received message
+	type Origin;
+
+	/// Forwards a received message from origin
+	fn forward(
+		middleware: Self::Middleware,
+		origin: Self::Origin,
+		message: Vec<u8>,
+	) -> DispatchResult;
+}
```
lemunozm, line 179: Why not reuse the `MessageSender` and `MessageReceiver`?

lemunozm: Is it because the `Message` type here is as `Message`? Maybe then the message can be also an associated types in `MessageSender` and `MessageReceiver`, to reusing them? Just thinking out laud

lemunozm (commented):
Really like that the serialization and deserialization is in the same place of the LP chain.

I like the design 🚀

libs/traits/src/liquidity_pools.rs
wischli: Yeah that is one reason. Another is that this was a proposal from @cdamian which I liked to distinguish between both traits and its implementers. Not sure what is preferable: Composability or explicitness.

lemunozm: What you think is better! I think we could do that refactor if needed after the audit. So we can focus in the solution

libs/traits/src/liquidity_pools.rs
wischli: > I am not sure if I follow, but I do not see the `SerializabbleDomain` associated type used by the trait

Artifact of a previous implementation indeed!

libs/traits/src/liquidity_pools.rs
wischli: Combined: [e5dedb2](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/e5dedb2f696e470abd94fe3481ab6cb9dd316ca9)

libs/traits/src/liquidity_pools.rs
wischli: Removed: [e5dedb2](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/e5dedb2f696e470abd94fe3481ab6cb9dd316ca9)

lemunozm:
Just a NIT thought. Seems like an entity called forwarder is not the entity in charge of serializing / deserializing, could be weird to find those methods here? 🤔 

Could make sense a super thing wrapper as a type runtime that makes that serialization?

```
// To send
gateway -> forwarder -> serializer -> router-dispatcher -> router 

// To receive
router -> serializer -> forwarder -> gateway
```

The serializer could be just a `MessageSerializer` implementing `MessageSender` and `MessageReceiver` type who calls msg.serialize() and deserialize().

Just a thought, and maybe being a paranoic organizer 😆

wischli:
> Just a NIT thought. Seems like an entity called forwarder is not the entity in charge of serializing / deserializing, could be weird to find those methods here? 🤔 
> 
> 
> 
> Could make sense a super thing wrapper as a type runtime that makes that serialization?
> 
> 
> 
> ```
> 
> // To send
> 
> gateway -> forwarder -> serializer -> router-dispatcher -> router 
> 
> 
> 
> // To receive
> 
> router -> serializer -> forwarder -> gateway
> 
> ```
> 
> 
> 
> The serializer could be just a `MessageSerializer` implementing `MessageSender` and `MessageReceiver` type who calls msg.serialize() and deserialize().
> 
> 
> 
> Just a thought, and maybe being a paranoic organizer 😆 

I agree that your proposal would be the preferred design. Will look into it later after finishing the mandatory tasks for this feature.

lemunozm:
Great! but please, it is not mandatory at all for Monday 🙏🏻

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+		type Message = T::Message;
+		type Middleware = T::RouterId;
+		type Origin = DomainAddress;
+
+		fn send(
+			router_id: T::RouterId,
+			origin: DomainAddress,
+			message: T::Message,
+		) -> DispatchResult {
+			let payload = if let Some(info) = RouterForwarding::<T>::get(&router_id) {
```
cdamian, line 198: Nit - use match?

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+			router_id: T::RouterId,
+			domain_address: DomainAddress,
+			payload: Vec<u8>,
+		) -> DispatchResult {
+			let message = T::Message::deserialize(&payload)?;
+
+			// If message can be unwrapped, it was forwarded
+			if let Some((source_domain, _, lp_message)) = message.unwrap_forwarded() {
+				let router_ids = T::RouterProvider::routers_for_domain(source_domain);
+				for router_id in router_ids {
```
cdamian, line 224: Nit - use `router_ids.iter().find()`.

lemunozm: I think we "cannot" due to the return

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+					// NOTE: We can rely on EVM side to ensure forwarded messages are valid such
+					// that it suffices to filter for the existence of forwarding info
+					if RouterForwarding::<T>::get(&router_id).is_some() {
+						return T::MessageReceiver::receive(
+							router_id,
+							domain_address,
+							lp_message.serialize(),
+						);
+					}
```
lemunozm, line 233: So basically, the first router found for that domain is the chosen one, right? Could this affect to proofs, right? which expect other routers. Or is a proof never wrapped in a `ForwardedMessage`?

pallets/liquidity-pools-forwarder/src/lib.rs
cdamian: Hmm, on second though, I think we can drop this whole loop. Basically, we want to get the `ForwardingInfo` for the provided router_id, if any, and then continue processing?

wischli: Ouch, I was completely ignoring the input argument `router_id`. You are right @cdamian !

pallets/liquidity-pools-forwarder/src/lib.rs
wischli: To the best of my knowledge, if the check is binary, `if else` is preferred - which is the case here.

pallets/liquidity-pools-forwarder/src/lib.rs
cdamian: Also, would the following work?

```
match message.unwrap_forwarded() {
				Some((source_domain, _, lp_message)) => {
					let router_ids = T::RouterProvider::routers_for_domain(source_domain);

					ensure!(
						router_ids.iter().any(|x| x == &router_id),
						DispatchError::Other("router ID not found for source domain")
					);

					match RouterForwarding::<T>::get(&router_id) {
						Some(_) => T::MessageReceiver::receive(
							router_id,
							domain_address,
							lp_message.serialize(),
						),
						None => Err(Error::<T>::ForwardInfoNotFound.into()),
					}
				}
				None => T::MessageReceiver::receive(router_id, domain_address, payload),
			}
```

Mainly too see if I got this flow right.

pallets/liquidity-pools-forwarder/src/lib.rs
lemunozm: I think with the @cdamian solution above, this is no longer an issue

runtime/common/src/routing.rs
```diff
@@ -57,10 +57,11 @@ impl<Routers> MessageSender for RouterDispatcher<Routers>
 where
 	Routers: pallet_axelar_router::Config,
 {
+	type Message = Vec<u8>;
```
cdamian, line 60: I think the current flow is:
LP gateway serializes message -> LP forwarder deserializes and serializes again -> RouterDispatcher sends the serialized message bytes

Wouldn't it make more sense to have it like:

LP gateway sends the actual LP message -> LP forwarder processes it and sends an LP message -> RouterDispatcher receives an LP message and serializes it when sending

WDYT?

pallets/liquidity-pools-forwarder/src/lib.rs
wischli: Check my solution here: [ce64041](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/ce64041e4125bca42a8626418dccaebf663ce9a5)

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+			domain_address: DomainAddress,
+			payload: Vec<u8>,
+		) -> DispatchResult {
+			let message = T::Message::deserialize(&payload)?;
+
+			// Message can be unwrapped iff it was forwarded
+			//
+			// NOTE: We can rely on EVM side to ensure forwarded messages are valid such
+			// that it suffices to filter for the existence of forwarding info
+			match (
```
cdamian, line 227: Love it, thank you!

runtime/common/src/routing.rs
wischli: > I think the current flow is: LP gateway serializes message -> LP forwarder deserializes and serializes again -> RouterDispatcher sends the serialized message bytes [...]

For serialization, I removed the gateway serialization because that is redundant as you have pointed out.

The forwarder pallet is at least necessary to wrap an LP message into forwarding, in case a forwarder config exists for the given router id. However, it doesn't have to do the actual (de-)serialization. I am more in favor of @lemunozm 's [above proposal](https://github.com/centrifuge/centrifuge-chain/pull/1971#issuecomment-2294360217) of moving the (de-)serialization to a runtime configured type. WDYT?

cdamian: I somehow missed that, nevermind this then.

pallets/liquidity-pools-forwarder/src/lib.rs
wischli: Improved in [b39fa47](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/b39fa47187b959c5c8dd3d2c7c73f1b5b55b0563)

wischli:
> Great! but please, it is not mandatory at all for Monday 🙏🏻

Much cleaner now: [b39fa47](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/b39fa47187b959c5c8dd3d2c7c73f1b5b55b0563)

pallets/liquidity-pools/src/message.rs
```diff
@@ -530,9 +565,20 @@ pub enum Message<BatchContent = BatchMessages> {
...
 	},
+	/// A wrapped Message which was forwarded from the source domain via the
+	/// contract.
+	///
+	/// Directionality: Centrifuge <-> EVM Domain.
+	Forwarded {
+		source_domain: SerializableDomain,
+		forwarding_contract: H160,
+		message: ForwardContent,
+	},
```
wischli, line 576: @hieronx This would be the (de-) serialization order. One downside of this generic approach for receiving and sending the same `Forwarded` message is that when Centrifuge Chain receives a forwarded message, `forwarding_contract` is not necessary. However, I believe it is feasible to just fill this `H160` with 0s on EVM side when forwarding.

Otherwise, we would have to implement two forwarded message types which increases complexity of (de-)serialization by orders of magnitude. 

A quick response would be highly appreciated! The rest of the PR are implementation details that you can skip.

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+	/// of A.
+	pub(crate) source_domain: Domain,
+	/// Refers to contract on forwarding chain.
+	///
+	/// Example: Assume A -> B -> C, then this refers to the forwarding
+	/// contract address on B.
+	pub(crate) contract: H160,
+}
+
+#[frame_support::pallet(dev_mode)]
```
cdamian, line 57: Should we drop `dev_mode`?

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+pub mod pallet {
+	use std::fmt::Debug;
```
cdamian, line 59: Nit - we use `core::fmt:Debug` in other pallets.

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+		///
+		/// Origin: Admin.
+		///
+		/// NOTE: Simple weight due to origin requirement.
+		#[pallet::weight(T::DbWeight::get().writes(1))]
+		#[pallet::call_index(1)]
+		pub fn remove_forwarder(origin: OriginFor<T>, router_id: T::RouterId) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
+
+			if let Some(info) = RouterForwarding::<T>::take(&router_id) {
```
cdamian, line 181: Nit - for other extrinsics that perform removal, we return a `NotFound` error, it might make sense here as well to inform the user of a potential incorrect call.

pallets/liquidity-pools/src/message.rs
hieronx: Yeah that seems like a good solution!

pallets/liquidity-pools-forwarder/src/tests.rs
```diff
@@ -0,0 +1,394 @@
...
+
+			assert_eq!(
+				RouterForwarding::<Runtime>::get(ROUTER_ID),
+				Some(ForwardInfo {
+					contract: FORWARD_CONTRACT,
+					source_domain: SOURCE_DOMAIN
+				})
+			);
+
+			System::assert_last_event(RuntimeEvent::LiquidityPoolsForwarder(
```
cdamian, line 29: Cool, had no idea about this one!

pallets/liquidity-pools/src/message.rs
```diff
@@ -192,6 +193,40 @@ impl BatchMessages {
 	}
 }
 
+/// A message type that cannot be forwarded.
+
```
cdamian, line 197: Nit - extra lines here and below?

runtime/common/src/routing.rs
```diff
@@ -78,3 +80,40 @@ impl<Runtime: pallet_evm::Config> PreConditions<(H160, H256)> for EvmAccountCode
 		BlakeTwo256::hash_of(&code) == contract_hash
 	}
 }
+
+pub struct MessageSerializer<Sender, Receiver>(PhantomData<(Sender, Receiver)>);
```
cdamian, line 84: Epic!

cdamian:
Only a couple of minor nits, I'm happy we got to slay this one as well, thank you @wischli!

pallets/liquidity-pools/src/message.rs
wischli: Removed!

pallets/liquidity-pools-forwarder/src/lib.rs
wischli: Implemented, thanks!

pallets/liquidity-pools-forwarder/src/lib.rs
wischli: Fixed!

pallets/liquidity-pools-forwarder/src/lib.rs
wischli: Definitely, good catch!

pallets/liquidity-pools-forwarder/src/tests.rs
```diff
@@ -0,0 +1,393 @@
...
+		use cfg_traits::liquidity_pools::MessageReceiver;
+
+		use super::*;
+
+		#[test]
+		fn with_forwarding() {
+			System::externalities().execute_with(|| {
+				config_mocks(true);
+
+				assert_ok!(<LiquidityPoolsForwarder as MessageReceiver>::receive(
```
lemunozm, line 299: Just a NIT curiosity, could not be `LiquidityPoolsForwarder::receive()`?

pallets/liquidity-pools-gateway/src/mock.rs
```diff
@@ -124,6 +126,18 @@ impl LPMessage for Message {
...
+		unimplemented!("out of scope")
+	}
+
+	fn unwrap_forwarded(self) -> Option<(Self::Domain, H160, Self)> {
+		unimplemented!("out of scope")
+	}
+
+	fn try_wrap_forward(_: Self::Domain, _: H160, _: Self) -> Result<Self, DispatchError> {
+		unimplemented!("out of scope")
+	}
```
lemunozm, line 140: I think we can simplify a lot this by having two traits. Note that previous `LpMessage` methods and these new methods are or-exclusively used. This means we can have two traits:

- `LpMessage` as it was, used by the gateway.
- `ForwardableMessage` (or a more, similar, cool name), used by the forwarder.

I think there is no need to merge all message methods under the same trait. This will allow us to remove all `unimplemented!` methods in the two mocks, reducing noise

pallets/axelar-router/src/mock.rs
```diff
@@ -35,6 +35,8 @@ impl frame_system::Config for Runtime {
 impl cfg_mocks::router_message::pallet::Config for Runtime {
 	type Middleware = Middleware;
 	type Origin = DomainAddress;
+	type ReceiverMessage = Vec<u8>;
+	type SenderMessage = Vec<u8>;
```
lemunozm, line 39: NIT. I think we're always using the same type for both, right? could we merge them under the same `Message` type. If, in the future, we need to use them with different types, the user can create different mock instances.

runtime/common/src/routing.rs
lemunozm: I agree with Cosmin! The result was super cool!

lemunozm (commented):
Thanks @wischli for bringing this super PR along with documentation chains so fast!! 🚀 I really like the result.

Just few NITs, some questions, and just one think that I would simplify about mocking the Message in both pallets.

But the code looks super great! Thanks

pallets/liquidity-pools-gateway/src/mock.rs
lemunozm: In a more deep thought, ideally we should have (names can be changed):

- `LpMessageSerializer` with serialize/deserialize.
- `LpMessageBatch` with batching methods.
- `LpMessageProof` with proofs methods.
- `LpMessageHash` with `message_hash()`.
- `LpMessageForwarder` with forwarding methods.

Note that all the above traits will NOT be in a hierarchy. They are just behaviors that you can add to a T::Message to deal with.

That way, different entities can deal with different message behavior, i.e. spliting tomorrow the gateway in different pallets, etc.

If agree, I think we can refactor this later. But at least I will split now The `LpMessage` from the `LpMessageForwarder` to simplify the mocks

wischli:
> Thanks @wischli for bringing this super PR along with documentation chains so fast!! 🚀 I really like the result.
> 
> 
> 
> Just few NITs, some questions, and just one think that I would simplify about mocking the Message in both pallets.
> 
> 
> 
> But the code looks super great! Thanks

Absolutely agree! Sorry I didn't initiate like that but prio was to get it done. If you think, you can still fit something in, great, but entirely optional.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -416,7 +416,7 @@ pub mod pallet {
 				return Err(Error::<T>::InvalidMessageOrigin.into());
 			}
 
-			Self::receive(router_id, origin_address, msg.into())
+			Self::receive(router_id, origin_address, T::Message::deserialize(&msg)?)
```
lemunozm, line 419: I think we need to move this out from the gateway. If not, the caller of this extrinsic will skip the forwarding logic.

Still not sure where, maybe on `pallet-axelar-router` or `pallet-liquidity-pool-forwarder` (and using `Message` instead of `Vec<u8>`).

Or maybe just removing it, because it's no longer used...

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+		) -> DispatchResult {
+			// Message can be unwrapped iff it was forwarded
+			//
+			// NOTE: We can rely on EVM side to ensure forwarded messages are valid such
+			// that it suffices to filter for the existence of forwarding info
+			let lp_message = match (
+				RouterForwarding::<T>::get(&router_id).is_some(),
+				message.clone().unwrap_forwarded(),
+			) {
+				(true, Some((_domain, _contract, lp_message))) => Ok(lp_message),
```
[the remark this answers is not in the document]
wischli: Because the origin of the LP message was not `_domain` (say Celo) but the `domain_address.domain()` (say Ethereum). The gateway checks in the allowlist, if it contains Ethereum's `domain_address.domain()` in the allowlist. I don't believe that Celo's `_domain` should be in the allowlist because there is no direct connection configured for Celo.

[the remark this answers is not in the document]
wischli: Though now I remember why I initially went with the `RouterProvider`! We cannot guarantee that the `router_id` of the `receive` signature here is the same for which we have stored forwarding. Therefore, we actually need to iterate over all possible router ids!

cc @cdamian

pallets/liquidity-pools-forwarder/src/tests.rs
wischli: Sure but I wanted to make it explicit.

pallets/liquidity-pools-gateway/src/mock.rs
wischli: Absolutely agree! Luckily, this change won't influence the outcome of the audit. So we can ship it after the audit feature freeze without decreasing the effectiveness.

pallets/liquidity-pools-forwarder/src/lib.rs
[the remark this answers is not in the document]
lemunozm: Mmm... Ok, here we have two points:

First, I think we need to read `_domain` or `_contract` somehow because if the other side forwarder implementation works like ours, it will never forward a message.

Second, if we need to use the forwarder domain, how do we receive the correct router?


Crazy idea. Could it makes sense that our chain never receives a Forwarder message? I mean, we're the final endpoint, right? The forwarder wrapper should have been unwrapped before reaching centrifuge. So, as a final endpoint, we should never receive a wrapped message.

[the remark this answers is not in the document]
lemunozm: I think my above crazy idea is not as crazy, and it's related to [this thread](https://github.com/centrifuge/centrifuge-chain/pull/1971#discussion_r1720782323) above with Jeroen about not using the contract.

TLDR: I think we should not receive forwarded messages.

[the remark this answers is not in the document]
wischli: > First, I think we need to read `_domain` or `_contract` somehow because if the other side forwarder implementation works like ours, it will never forward a message.

It won't work exactly the same, see [the spec](https://centrifuge.hackmd.io/@jeroen/Sy5MbJbzA?type=view#EVM-to-Centrifuge) and [this Slack response](https://kflabs.slack.com/archives/C06FRHCRQL9/p1723833895812759?thread_ts=1723801740.442379&cid=C06FRHCRQL9).

However, rereading the spec ECM to Centrifuge

> On Centrifuge
> 
> * Receives ForwardedMessage(source_chain=Gnosis, message=payload) from Ethereum
> * Executes payload, but with **source chain as Gnosis**

It appears to me we actually have to use the `source_domain` instead when handing over to the Gateway via `receive(...)` and require this `source_domain` to also be stored in the gateway allowlist.

Current flow Celo -> Ethereum -> Centrifuge is
* Forwarder checks whether message from Ethereum is a forwarded one 
* Assuming it is, extract the `source_domain` Celo and ensure forwarding storage is set up for Celo domain
* Hand over the extracted non-forward LP message to the gateway but with source domain and address pointing to the forwarder domain Ethereum

At least for handling tranche token transfers, this will be an issue, because it thinks the source domain was Ethereum. So I will fix this now.

pallets/liquidity-pools-gateway/src/lib.rs
wischli: Great catch! I propose to add a `MessageDeserializer` AT to the Gateway config which is implemented by our runtime type `MessageSerializer`. This will deserialize the message and pass it to the forwarder pallet which will pass it back to the `MessageReiver` impl of the Gateway.

So from now:
* `process` extrinsic -> Gateway impl of `MessageReceiver`

To
* `process` in Gateway -> `receive` in `MessagSerializer` which deserializes -> `receive` in forward pallet which unwraps any forwarded message -> `receive` in Gateway which does its thing

WDYT?

pallets/liquidity-pools-forwarder/src/lib.rs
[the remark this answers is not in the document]
lemunozm: How will you then choose the correct `router_id` for the `source_domain`, @wischli ?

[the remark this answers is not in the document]
wischli: See my above comment, the forwarder pallet needs to re-add the `RouterProvider` runtime type: https://github.com/centrifuge/centrifuge-chain/pull/1971#discussion_r1720905269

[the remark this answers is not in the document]
lemunozm: But that way, we're breaking the proof checks, because all messages (even proofs) are redirected to the first found router.

cc @cdamian

pallets/liquidity-pools-gateway/src/lib.rs
wischli: This requires instantiating the `router_message` mock pallet, which I am working on right now.

lemunozm: Mmm... I am not sure about that walk...

As more I think on it as more I think we can remove this. We no longer need this since Axelar uses the trait to send message to the gateway. If we want to emulate receiving message explicitly, this extrinsic should be in axelar-router

wischli: Interested in your take on this @cdamian

wischli: See here for code changes if we added the `MessageDeserializer` to the Gateway: [a253a6f](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/a253a6fcb3db9292ca5686d2df651fd074d345a0)

However, I have reverted that change and applied @lemunozm's proposal of removing the `receive_message` extrinsic which apparently was only used by the Axelar precompile. I have refactored the existing Gateway UTs for checking the `MessageReceiver` implementation which was not tested: [cbb75b4](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/cbb75b4605c618a32885be53259db8319ce49b40)

pallets/axelar-router/src/mock.rs
wischli: [6a5cb5f](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/6a5cb5fdf6bee131476a92878f01183f3e5459e4)

pallets/liquidity-pools-forwarder/src/tests.rs
wischli: [4a0f751](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/4a0f751c6638212341ca1b344c5832e0f6a425f9)

pallets/liquidity-pools-forwarder/src/lib.rs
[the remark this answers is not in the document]
wischli: > But that way, we're breaking the proof checks, because all messages (even proofs) are redirected to the first found router, right?
> 
> cc @cdamian

It appears to me that this is a question of how to correclty configure/set up the storages. E.g. when calling `set_forwarding_info`, the correct router id needs to be stored which does not break our _first router_ invariant.

Moreover, this loop also breaks on the first found router.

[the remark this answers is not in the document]
wischli: > `Forwarder X -> Ethereum -> Centrifuge` - message that comes through router A (pallet-axelar-router)
> `Forwarder X -> Ethereum -> Centrifuge` - proof that comes through router B (some other router pallet)

This scenario seems possible.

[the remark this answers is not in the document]
cdamian: OK, so in that case we can use the router ID that's provided to the forwarder in the receive call.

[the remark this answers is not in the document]
cdamian: Summary being:

>Forwarder X -> Ethereum -> Centrifuge - message that comes through router A (pallet-axelar-router)
this will call `receive` with router ID A

> Forwarder X -> Ethereum -> Centrifuge - proof that comes through router B (some other router pallet)
this will call `receive` with router ID B

Meaning that we shouldn't get the router ids for the domain provided when unwrapping the message?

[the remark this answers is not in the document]
wischli: I see your point now. The current design wouldn't handle messages and their proofs correclty as everything is mapped to the same router id. So we are on the same page.

Revisiting our inbound setup and [my comment](https://github.com/centrifuge/centrifuge-chain/pull/1971#discussion_r1720939493) about correct forwarding configuration: `Router -> Serializer -> Forwarder -> Gateway`, it appears to me that we can actually rely on the correct `router_id` from the `receive` signature called by the `Router` and re-remove the `RouterProvider` from the forwarder pallet. 

E.g., when we store the forwarding info for a router id, we need to do that for every supported router. This way, the forwarder pallet has no impact on handling proof- vs non-proof messages and leaves this up to the Gateway based on the `router_id` which is passed all the way from the `MessageReceiver` impl of the corresponding router.

WDYT @cdamian @lemunozm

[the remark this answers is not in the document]
wischli: > `Forwarder X -> Ethereum -> Centrifuge` - message that comes through router A (pallet-axelar-router)
> `Forwarder X -> Ethereum -> Centrifuge` - proof that comes through router B (some other router pallet)

This is the counterexample to the current implementation of the forwarder pallet.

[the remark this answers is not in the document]
cdamian: > I see your point now. The current design wouldn't handle messages and their proofs correclty as everything is mapped to the same router id. So we are on the same page.
> 
> Revisiting our inbound setup and [my comment](https://github.com/centrifuge/centrifuge-chain/pull/1971#discussion_r1720939493) about correct forwarding configuration: `Router -> Serializer -> Forwarder -> Gateway`, it appears to me that we can actually rely on the correct `router_id` from the `receive` signature called by the `Router` and re-remove the `RouterProvider` from the forwarder pallet.
> 
> E.g., when we store the forwarding info for a router id, we need to do that for every supported router. This way, the forwarder pallet has no impact on handling proof- vs non-proof messages and leaves this up to the Gateway based on the `router_id` which is passed all the way from the `MessageReceiver` impl of the corresponding router.
> 
> WDYT @cdamian @lemunozm

Completely agree.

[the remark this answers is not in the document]
wischli: Copy pasting here the flows for a 2-router-setup you broke down to make this more explicit:

Flow of a non-forwarded message
```
1) Ethereum -> Centrifuge Router A -> LP Forwarder::receive(router A, domain address, message) -> noop -> LP Gateway::receive(router A, domain address, message)

2) Ethereum -> Centrifuge Router B -> LP Forwarder::receive(router A, domain address, message) - >noop -> LP Gateway::receive(router A, domain address, message)
```

Flow of a forwarded message
```
1) Celo(Ethereum) -> Centrifuge Router A -> LP Forwarder::receive(router A, domain address, message) -> unwrap message to (source_domain, contract, unwrapped_message) -> LP Gateway::receive(router A, (source domain, contract), unwrapped_message)

2) Celo(Ethereum) -> Centrifuge Router B -> LP Forwarder::receive(router B, domain address, message) -> unwrap message to (source_domain, contract, unwrapped_message) -> LP Gateway::receive(router B, (source domain, contract), unwrapped_message)
```

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+		fn send(
+			router_id: T::RouterId,
+			origin: DomainAddress,
+			message: T::Message,
+		) -> DispatchResult {
+			let msg = RouterForwarding::<T>::get(&router_id)
+				.map(|info| {
+					T::Message::try_wrap_forward(info.source_domain, info.contract, message.clone())
+				})
+				.unwrap_or_else(|| {
```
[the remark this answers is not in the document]
wischli: We can hit both elses!

When the `try_wrap` fails, `RouterForwarding::<T>::get(&router_id)` is some, such that calling

```rust
			let msg = RouterForwarding::<T>::get(&router_id)
				.map(|info| {
					T::Message::try_wrap_forward(info.source_domain, info.contract, message.clone())
				})
				.unwrap()`
```, 

the error is extracted. The `unwrap_or_else` only handles the case of missing forwarding info.

We are testing the `try_wrap` error here: https://github.com/centrifuge/centrifuge-chain/pull/1971/files#diff-2fa3688d086872e10e1a539738a4b3cd7d6ce44da1530b1f8c7818744b079becR231-R245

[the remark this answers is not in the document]
Comment (cdamian) [reply pallets/liquidity-pools-forwarder/src/lib.rs:212]: Dang, rust 1 - me 0. Sorry for the noise...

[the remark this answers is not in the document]
Comment (wischli) [reply pallets/liquidity-pools-forwarder/src/lib.rs:216]: No worries! It was good to check if my late night brain was not injecting a loop hole :D

File: pallets/liquidity-pools-forwarder/src/lib.rs (hunk 21 above)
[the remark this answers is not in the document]
Comment (wischli) [reply pallets/liquidity-pools-forwarder/src/lib.rs]: Since we all have found consensus on this matter, here's the corresponding code: [e409fd5](https://github.com/centrifuge/centrifuge-chain/pull/1971/commits/e409fd59baf891910fd4a42a5ff7138cc1735e5f)

I really hope this was the last raid boss to slay and that we can merge this beast.

File: pallets/liquidity-pools-forwarder/src/lib.rs [hunk 23]
```diff
@@ -0,0 +1,266 @@
...
+						Error::<T>::SourceDomainMismatch
+					);
+
+					let domain_address = DomainAddress::Evm(
+						info.source_domain
+							.get_evm_chain_id()
+							.expect("Domain not Centrifuge; qed"),
+						info.contract,
+					);
+					Ok((lp_message, domain_address))
```
cdamian, line 255: Nit - this one is actually the `forwarding_domain_address`. Feel free to disregard tho, can live with this.

pallets/liquidity-pools-forwarder/src/lib.rs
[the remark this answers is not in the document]
cdamian: My late night brain left that comment haha.

pallets/liquidity-pools-forwarder/src/lib.rs
wischli: No both are different:

* `forwarding_domain_address` is some address of the forwarding domain `Ethereum`, e.g. `DomainAddress::Evm(1, x)`
* `domain_address` here is the the domain address of the source domain and the forwarding contract, e.g. `DomainAddress::Evm(42220, info.contract)`

pallets/liquidity-pools-forwarder/src/lib.rs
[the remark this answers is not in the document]
lemunozm: I understand the second line of non-forwaded message is router B in all places right? @wischli

pallets/liquidity-pools-forwarder/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+				(Some(info), Some((source_domain, _contract, lp_message))) => {
+					ensure!(
+						info.source_domain == source_domain,
+						Error::<T>::SourceDomainMismatch
+					);
+
+					let domain_address = DomainAddress::Evm(
+						info.source_domain
+							.get_evm_chain_id()
+							.expect("Domain not Centrifuge; qed"),
```
lemunozm, line 252: NIT. we can have more domains in the future and break this silently. Just a tech-debt here to fix.

pallets/liquidity-pools/src/message.rs
```diff
@@ -79,9 +80,9 @@ impl TryInto<Domain> for SerializableDomain {
 
 /// A message type that can not be a batch.
 #[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo)]
-pub struct NoBatchMessage(Message);
+pub struct NonBatchMessage(Message);
```
lemunozm, line 83: Thanks for polishing my names 🫶

pallets/liquidity-pools/src/message.rs
```diff
@@ -206,7 +239,7 @@ impl BatchMessages {
 	MaxEncodedLen,
 	Default,
 )]
-pub enum Message<BatchContent = BatchMessages> {
+pub enum Message<BatchContent = BatchMessages, ForwardContent = NonForwardMessage> {
```
lemunozm, line 242: Are curious about where a message with ForwardContent != NonForwardMessage is created. I think you always create instances with `NonForwardMessage`, right?

lemunozm (commented):
LGTM! Some comments that does not block the merge! Let's add this 💯

pallets/liquidity-pools-forwarder/src/lib.rs
wischli: I get the point but actually think we wont have an additional domain variant for the time being because that requires LP compatibility with that domain and that is not on the radar outside EVM. For XCM, the domain would actually be Centrifuge.

pallets/liquidity-pools/src/message.rs
wischli: Yes. I suppose this can be reduced then? Would like to merge this PR as is assuming CI is green. Will clean up in a follow up PR. Thanks for spotting this!

pallets/liquidity-pools-forwarder/src/lib.rs
lemunozm: Sure! Not for the time being. Could only be an issue in the future

PR #1972, opened 2024-08-16
LPv2: Isolate gateway weights

lemunozm:
# Description

Small PR to move the weight checks to `MessageProcessor`.  

Like the fact that now only the `MessageProcessor` trait implementation knows about weights

### DIFF
--- a/pallets/liquidity-pools-gateway/src/lib.rs
+++ b/pallets/liquidity-pools-gateway/src/lib.rs
@@ -517,18 +517,31 @@ pub mod pallet {
 					domain_address,
 					message,
 					router_id,
-				} => Self::process_inbound_message(domain_address, message, router_id),
+				} => {
+					let mut counter = 0;
+
+					let res = Self::process_inbound_message(
+						domain_address,
+						message,
+						router_id,
+						&mut counter,
+					);
+
+					let weight = match counter {
+						0 => LP_DEFENSIVE_WEIGHT / 10,
+						n => LP_DEFENSIVE_WEIGHT.saturating_mul(n),
+					};
+
+					(res, weight)
+				}
 				GatewayMessage::Outbound {
 					sender,
 					message,
 					router_id,
 				} => {
-					let weight = LP_DEFENSIVE_WEIGHT;
+					let res = T::MessageSender::send(router_id, sender, message.serialize());
 
-					match T::MessageSender::send(router_id, sender, message.serialize()) {
-						Ok(_) => (Ok(()), weight),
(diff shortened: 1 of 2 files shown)


pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -517,18 +517,31 @@ pub mod pallet {
...
+
+					let res = Self::process_inbound_message(
+						domain_address,
+						message,
+						router_id,
+						&mut counter,
+					);
+
+					let weight = match counter {
+						0 => LP_DEFENSIVE_WEIGHT / 10,
```
lemunozm, line 531: We want to add just a bit if we fail even before starting to process the first message

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -437,27 +422,19 @@ impl<T: Config> Pallet<T> {
 				expected_proof_count,
 			);
 
-			if let Err(e) = inbound_entry.validate(&router_ids, &router_id.clone()) {
-				return (Err(e), weight.saturating_mul(count));
-			}
-
-			if let Err(e) = Self::upsert_pending_entry(message_proof, &router_id, inbound_entry) {
```
cdamian, line 444: My eyes are not bleeding anymore when looking at this, thank you <3

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Should we maybe add a `LP_MIN_WEIGHT` or something?

pallets/liquidity-pools-gateway/src/message_processing.rs
lemunozm: BTW, liked that you move the outbound processing impl into the `MessageProcessor`, now everything is quite unified 🙌🏻

pallets/liquidity-pools-gateway/src/lib.rs
wischli: I would prefer hardcoding that as well, yes. An alternative would be a simple `fn defensive_weight(counter)`.

wischli (approved):
Thanks for tackling this! LGTM but I would prefer hardcoding the reduced defensive weight.

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: I didn't read it, and I had the auto-merge enabled. Tech-debt

PR #1974, opened 2024-08-17
Latest cleanings pre audit

lemunozm:
# Description

PR with some cleanings to be merged just before the audit, after:
1. https://github.com/centrifuge/centrifuge-chain/pull/1970
2. https://github.com/centrifuge/centrifuge-chain/pull/1971

Changes:

- [x] `pallet-liquidity-pools-queue` moved to `pallets/`
- [x] extract trait queue methods from `traits/liquidity_pools.rs` (they are generic enough)
- [x] ~renamed to `pallet-queue` or `pallet-idle-queue`~
- [x] Fix docs
- [x] RuntimeEnv with support for idle events
- [x] queue ITs now do not need to use fudge.
- [x] `cases/liquidity_pools.rs` renamed to `cases/foreign_investments.rs`
- [x] extract domain from router ID - [discussion](https://github.com/centrifuge/centrifuge-chain/pull/1970#discussion_r1720717386)
- [x] change message recovery arg type - [discussion](https://github.com/centrifuge/centrifuge-chain/pull/1970#discussion_r1720762179)


Cargo.toml
```diff
@@ -235,7 +235,7 @@ pallet-investments = { path = "pallets/investments", default-features = false }
 pallet-keystore = { path = "pallets/keystore", default-features = false }
 pallet-liquidity-pools = { path = "pallets/liquidity-pools", default-features = false }
 pallet-liquidity-pools-gateway = { path = "pallets/liquidity-pools-gateway", default-features = false }
-pallet-liquidity-pools-gateway-queue = { path = "pallets/liquidity-pools-gateway/queue", default-features = false }
+pallet-liquidity-pools-gateway-queue = { path = "pallets/liquidity-pools-gateway-queue", default-features = false }
```
cdamian, line 238: I think we can leave this name as it is.

lemunozm: But then it would look weird if the pallet is being used for another kind of queues...

cdamian: That's a problem for future-us :D Plus, I think that eventually we'll drop it and use `Tasks`.

lemunozm: I'm not convinced by the name, but I will leave it as it

lemunozm:
If passes CI, that it should, we can merge it

cdamian (approved):

lemunozm:
I've moved some utility into axelar to be audited as part of the pallet: [cd7c3d0](https://github.com/centrifuge/centrifuge-chain/pull/1974/commits/cd7c3d05acf627a9dfcb3a54f8247ff79ff6c0ac)

wischli (approved):
Thanks so much for yet another cleanup!

PR #1976, opened 2024-08-18
Clean: Remove unused gateway origin

lemunozm:
# Description

We can simplify the gateway a bit more by removing the now unused origin. But feel free to skip this PR if we reach the audit deadline before


cdamian (approved):
Thank you!

wischli (approved):
🧹 You're the best cleaner!

runtime/centrifuge/src/lib.rs
```diff
@@ -2099,7 +2097,7 @@ construct_runtime!(
 		LiquidityRewardsBase: pallet_rewards::<Instance2>::{Pallet, Storage, Event<T>, Config<T>} = 104,
 		LiquidityRewards: pallet_liquidity_rewards::{Pallet, Call, Storage, Event<T>} = 105,
 		GapRewardMechanism: pallet_rewards::mechanism::gap = 106,
-		LiquidityPoolsGateway: pallet_liquidity_pools_gateway::{Pallet, Call, Storage, Event<T>, Origin } = 107,
+		LiquidityPoolsGateway: pallet_liquidity_pools_gateway::{Pallet, Call, Storage, Event<T> } = 107,
```
lemunozm, line 2100: I think this could imply a migration because we're modifying the origin enum stored by the `order_book`, through the `Feeder` wrapper 🤔. How can we check if this happens? `try-runtime` should tell this automatically?

lemunozm:
CI fails when calling this Solidity constructor: https://github.com/centrifuge/liquidity-pools/blob/6e8f1a29dff0d7cf5ff74285cfffadae8a8b303f/src/factories/TransferProxyFactory.sol#L62 which makes absolute no sense

runtime/centrifuge/src/lib.rs
wischli: Possible, yes. Thanks for bringing it up! I added it to the post audit migration [todo list](https://centrifuge.hackmd.io/_bJWZVKySHOCDkhrpbqbAQ?both#Post-audit).

wischli (commented):
@lemunozm You altered the submodule file again which should be the root of your errors.

lemunozm:
Oh, man... I can't believe it 🫠. My mind no longer works...

wischli (approved):
Spring cleaning is outdated, summer cleaning it is - estupendo!

PR #1979, opened 2024-08-19
Refactor: Remove overusage of DomainAddress

lemunozm:
# Description

We were overusing `DomainAddress` in places where we did not really need it, generating extra complexity that offuscate some parts of the codebase (i.e affecting to how the forwarder needs to be implemented).

Change list:
- Simplify sending/receiving methods:
  - Sending methods use as origin an `AccountId` instead of `DomainAddress`.
  - Receiving methods use as origin a `Domain` instead of `DomainAddress`.
- Move `gateway::Allowlist `storage for instances its methods from the `gateway` to the `axelar-router`.
  - The reason: we need to check if the address is valid or not as soon as possible. Doing it in the gateway is late. We want to only forward messages (which is reached before the gateway) for allowed instances.
  - This check could not be super ideal here. i.e, in the future, several routers can use the same allowlist information. If that is the case, this should be extracted as a precondition or as a new entity in the chain:
    - `axelar-router -> (new) allowlist-checker -> deserializer -> forwarder -> gatewat -> lp`

### DIFF
--- a/libs/types/src/domain_address.rs
+++ b/libs/types/src/domain_address.rs
@@ -21,7 +21,7 @@ use crate::EVMChainId;
 const MAX_ADDRESS_SIZE: usize = 32;
 
 /// By just clamping the value to a smaller address
-pub fn account_to_eth_address(address: AccountId32) -> H160 {
+pub fn truncate_into_eth_address(address: AccountId32) -> H160 {
 	let bytes: [u8; 32] = address.into();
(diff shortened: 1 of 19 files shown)


lemunozm:
Will create another one more clean to evaluate the diffs

lemunozm:
https://github.com/centrifuge/centrifuge-chain-internal/pull/57

PR #1980, opened 2024-08-19
lp-forwarder: Always use Centrifuge domain when wrapping outbound mes…

cdamian:
# Description

The current implementation is using the stored domain when wrapping forward messages, however, this is incorrect since we should always use `Domain::Centrifuge`.

# Checklist

### DIFF
--- a/pallets/liquidity-pools/src/message.rs
+++ b/pallets/liquidity-pools/src/message.rs
@@ -659,13 +659,9 @@ impl LpMessageForwarded for Message {
 		}
 	}
 
-	fn try_wrap_forward(
-		source_domain: Domain,
-		forwarding_contract: H160,
-		message: Self,
-	) -> Result<Self, DispatchError> {
+	fn try_wrap_forward(forwarding_contract: H160, message: Self) -> Result<Self, DispatchError> {
 		Ok(Self::Forwarded {
-			source_domain: source_domain.into(),
+			source_domain: Domain::Centrifuge.into(),
 			forwarding_contract,
 			message: message.try_into().map_err(|_| {
 				DispatchError::Other(
--- a/pallets/liquidity-pools/src/message.rs
+++ b/pallets/liquidity-pools/src/message.rs
@@ -659,9 +659,13 @@ impl LpMessageForwarded for Message {
 		}
 	}
 
-	fn try_wrap_forward(forwarding_contract: H160, message: Self) -> Result<Self, DispatchError> {
+	fn try_wrap_forward(
+		source_domain: Self::Domain,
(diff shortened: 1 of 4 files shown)


pallets/liquidity-pools/src/message.rs
```diff
@@ -659,13 +659,9 @@ impl LpMessageForwarded for Message {
...
 
-	fn try_wrap_forward(
-		source_domain: Domain,
-		forwarding_contract: H160,
-		message: Self,
-	) -> Result<Self, DispatchError> {
+	fn try_wrap_forward(forwarding_contract: H160, message: Self) -> Result<Self, DispatchError> {
 		Ok(Self::Forwarded {
-			source_domain: source_domain.into(),
+			source_domain: Domain::Centrifuge.into(),
```
lemunozm, line 664: I would move this to the forwarder pallet. If not, this means that any message we ask for to be forwarder will have a domain of `Centrifuge` by default. But this could not be the case in all scenarios. It is only for the implementation of `pallet-liquidity-pools-forwarder`

lemunozm (commented):
Super simple solution, thanks @cdamian. Just one minor thing

pallets/liquidity-pools/src/message.rs
cdamian: That was my initial approach but I couldn't think of any case where that would apply.

lemunozm: `Message` should be super dumb with no business logic

cdamian: Yup, I'll revert this.

lemunozm (commented):
LGTM! Thanks @cdamian.

A question, the check of line: https://github.com/centrifuge/centrifuge-chain/pull/1980/files#diff-d763817d4b90117d6295fc32e8fd8a66d2ee4093bfb40d90941a3368e72aa099R244-R247 makes sense? I mean, should we treat the forward info domain as a whitelist for domains?

lemunozm (approved):

cdamian:
> LGTM! Thanks @cdamian.
> 
> A question, the check of line: https://github.com/centrifuge/centrifuge-chain/pull/1980/files#diff-d763817d4b90117d6295fc32e8fd8a66d2ee4093bfb40d90941a3368e72aa099R244-R247 makes sense? I mean, should we treat the forward info domain as a whitelist for domains?

I would leave it like that for now and get back to it after we discuss the rest of the LP concerns. Furthermore, setting forwarding info can only be done by admins so we can trust it for the time being anyway IMO. TBD if we need extra/different validation.

PR #1990, opened 2024-09-11
fix: add fees in TT price calculation

### LINKED ISSUE #1989: bug: Calculate Tranche Token Prices with `NavFees`
### Description
Current implementation in the pallet-pool-system for `trait TrancheTokenPrice` does not take into account accrued fees, as the epoch execution does. 

We need to fix that by adding the same flow as for epoch-closing.

mustermeiszer:
# Description
See issue

Fixes #1989

## Changes and Descriptions
* just adds fees to tranche token price calculation
 
# Checklist:

### DIFF
--- a/pallets/pool-system/src/impls.rs
+++ b/pallets/pool-system/src/impls.rs
@@ -66,13 +66,15 @@ impl<T: Config> TrancheTokenPrice<T::AccountId, T::CurrencyId> for Pallet<T> {
 		pool_id: Self::PoolId,
 		tranche_id: Self::TrancheId,
 	) -> Option<(T::BalanceRatio, Seconds)> {
-		let now = T::Time::now();
 		let mut pool = Pool::<T>::get(pool_id)?;
 
-		// Get cached nav as calculating current nav would be too computationally
-		// expensive
-		let (nav, nav_last_updated) = T::AssetsUnderManagementNAV::nav(pool_id)?;
-		let total_assets = pool.reserve.total.ensure_add(nav).ok()?;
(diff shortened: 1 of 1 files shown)


pallets/pool-system/src/impls.rs
```diff
@@ -66,13 +66,15 @@ impl<T: Config> TrancheTokenPrice<T::AccountId, T::CurrencyId> for Pallet<T> {
...
-		// expensive
-		let (nav, nav_last_updated) = T::AssetsUnderManagementNAV::nav(pool_id)?;
-		let total_assets = pool.reserve.total.ensure_add(nav).ok()?;
+		let (nav_loans, nav_loans_updates) = T::AssetsUnderManagementNAV::nav(pool_id)?;
+		let (nav_fees, nav_fees_updated) = T::PoolFeesNAV::nav(pool_id)?;
+
+		let nav = Nav::new(nav_loans, nav_fees);
+		let total_assets = nav
+			.total(pool.reserve.total)
+			.unwrap_or(<T as Config>::Balance::zero());
```
lemunozm, line 77: Super NIT: `unwrap_or_default()`

lemunozm (approved):
Makes sense! LGTM

pallets/pool-system/src/impls.rs
mustermeiszer: I am not the biggest fan of default in these case. Zero is really explicit and also what we want.

mustermeiszer:
Why did the pipeline fcancel.

PR #1991, opened 2024-09-19
Fix/gateway message processing

cdamian:
# Description

Various fixes for LP gateway-related logic.

### LP Gateway
- Make `MessageProcessor` implementation transactional to ensure that storage changes are reverted on failure.
- Move iteration over inbound sub-messages in execution logic.
- Add session ID check for inbound message entries when checking for votes/proofs.
- Use defensive weight when executing message recovery.
- Add more tests that cover session ID changes and storage rollback.

### LP Gateway Queue
- Remove extra weight check.
- Get `MessageQueue` keys, order them, and iterate over them when servicing the `MessageQueue`. 

# Checklist:

### DIFF
--- a/pallets/liquidity-pools-gateway-queue/src/lib.rs
+++ b/pallets/liquidity-pools-gateway-queue/src/lib.rs
@@ -236,10 +236,6 @@ pub mod pallet {
 				processed_entries.push(nonce);
 
 				weight_used = weight_used.saturating_add(weight);
-
-				if weight_used.any_gte(max_weight) {
-					break;
-				}
 			}
 
 			for entry in processed_entries {
--- a/pallets/liquidity-pools-gateway-queue/src/lib.rs
+++ b/pallets/liquidity-pools-gateway-queue/src/lib.rs
@@ -22,7 +22,6 @@ use parity_scale_codec::FullCodec;
 use scale_info::TypeInfo;
 use sp_arithmetic::traits::BaseArithmetic;
 use sp_runtime::traits::{EnsureAddAssign, One};
-use sp_std::vec::Vec;
 
 #[cfg(test)]
 mod mock;
@@ -61,6 +60,12 @@ pub mod pallet {
 	#[pallet::getter(fn message_nonce_store)]
 	pub type MessageNonceStore<T: Config> = StorageValue<_, T::MessageNonce, ValueQuery>;
 
+	/// Storage that is used for keeping track of the last nonce that was
+	/// processed.
+	#[pallet::storage]
+	#[pallet::getter(fn last_processed_nonce)]
+	pub type LastProcessedNonce<T: Config> = StorageValue<_, T::MessageNonce, ValueQuery>;
+
 	/// Storage for messages that will be processed during the `on_idle` hook.
 	#[pallet::storage]
 	#[pallet::getter(fn message_queue)]
@@ -92,6 +97,11 @@ pub mod pallet {
 			nonce: T::MessageNonce,
 			message: T::Message,
 			error: DispatchError,
+		},
+
+		/// Maximum number of messages was reached.
+		MaxNumberOfMessagesWasReached {
+			last_processed_nonce: T::MessageNonce,
 		},
 	}
 
@@ -200,16 +210,42 @@ pub mod pallet {
 		}
 
 		fn service_message_queue(max_weight: Weight) -> Weight {
-			let mut weight_used = Weight::zero();
+			let mut last_processed_nonce = LastProcessedNonce::<T>::get();
 
-			let mut nonces = MessageQueue::<T>::iter_keys().collect::<Vec<_>>();
-			nonces.sort();
+			// 1 read for the last processed nonce
+			let mut weight_used = T::DbWeight::get().reads(1);
 
-			for nonce in nonces {
-				let message =
-					MessageQueue::<T>::get(nonce).expect("valid nonce ensured by `iter_keys`");
+			loop {
+				if last_processed_nonce.ensure_add_assign(One::one()).is_err() {
+					Self::deposit_event(Event::<T>::MaxNumberOfMessagesWasReached {
+						last_processed_nonce,
+					});
 
+					break;
+				}
+
+				// 1 read for the nonce
 				weight_used.saturating_accrue(T::DbWeight::get().reads(1));
+
+				if last_processed_nonce > MessageNonceStore::<T>::get() {
+					break;
+				}
+
+				// 1 read for the message
+				weight_used.saturating_accrue(T::DbWeight::get().reads(1));
+
+				let message = match MessageQueue::<T>::get(last_processed_nonce) {
+					Some(msg) => msg,
+					// No message found at this nonce, we can skip it.
+					None => {
+						LastProcessedNonce::<T>::set(last_processed_nonce);
+
+						// 1 write for setting the last processed nonce
+						weight_used.saturating_accrue(T::DbWeight::get().writes(1));
+
+						continue;
+					}
+				};
 
 				let remaining_weight = max_weight.saturating_sub(weight_used);
 				let next_weight = T::MessageProcessor::max_processing_weight(&message);
@@ -219,22 +255,28 @@ pub mod pallet {
 					break;
 				}
 
-				let weight = match Self::process_message_and_deposit_event(nonce, message.clone()) {
+				let processing_weight = match Self::process_message_and_deposit_event(
+					last_processed_nonce,
+					message.clone(),
+				) {
 					(Ok(()), weight) => weight,
 					(Err(e), weight) => {
-						FailedMessageQueue::<T>::insert(nonce, (message, e));
+						FailedMessageQueue::<T>::insert(last_processed_nonce, (message, e));
 
 						// 1 write for the failed message
 						weight.saturating_add(T::DbWeight::get().writes(1))
 					}
 				};
 
-				weight_used.saturating_accrue(weight);
+				weight_used.saturating_accrue(processing_weight);
 
-				MessageQueue::<T>::remove(nonce);
+				MessageQueue::<T>::remove(last_processed_nonce);
+
+				LastProcessedNonce::<T>::set(last_processed_nonce);
 
 				// 1 write for removing the message
-				weight_used.saturating_accrue(T::DbWeight::get().writes(1));
+				// 1 write for setting the last processed nonce
+				weight_used.saturating_accrue(T::DbWeight::get().writes(2));
 			}
 
 			weight_used
--- a/pallets/liquidity-pools-gateway-queue/src/lib.rs
+++ b/pallets/liquidity-pools-gateway-queue/src/lib.rs
@@ -100,7 +100,7 @@ pub mod pallet {
 		},
 
 		/// Maximum number of messages was reached.
-		MaxNumberOfMessagesWasReached {
+		MaxNumberOfMessagesReached {
 			last_processed_nonce: T::MessageNonce,
 		},
 	}
@@ -217,7 +217,7 @@ pub mod pallet {
 
 			loop {
 				if last_processed_nonce.ensure_add_assign(One::one()).is_err() {
-					Self::deposit_event(Event::<T>::MaxNumberOfMessagesWasReached {
+					Self::deposit_event(Event::<T>::MaxNumberOfMessagesReached {
 						last_processed_nonce,
 					});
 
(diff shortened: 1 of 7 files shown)


pallets/liquidity-pools-gateway-queue/src/lib.rs
```diff
@@ -202,9 +202,13 @@ pub mod pallet {
 		fn service_message_queue(max_weight: Weight) -> Weight {
 			let mut weight_used = Weight::zero();
 
-			let mut processed_entries = Vec::new();
+			let mut nonces = MessageQueue::<T>::iter_keys().collect::<Vec<_>>();
+			nonces.sort();
```
cdamian, line 206: Since we are collecting the keys here, it also made sense to me to sort them, just in case. Please let me know if there are any objections to this.

pallets/liquidity-pools-gateway-queue/src/lib.rs
```diff
@@ -233,17 +237,7 @@ pub mod pallet {
 					}
 				};
 
-				processed_entries.push(nonce);
-
 				weight_used = weight_used.saturating_add(weight);
-
-				if weight_used.all_gte(max_weight) {
```
cdamian, line 240: This was an extra check that I think we can remove.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -460,7 +464,7 @@ pub mod pallet {
 			domain_address: DomainAddress,
 			message_hash: MessageHash,
 			router_id: T::RouterId,
-		) -> DispatchResult {
+		) -> DispatchResultWithPostInfo {
```
cdamian, line 467: This is now a `DispatchResultWithPostInfo` given that we can get the number of inbound sub-messages when processing.

Also, are we sure that we won't be adding support for batch messages coming from EVM? If so, we can remove this whole counter logic completely.

pallets/liquidity-pools-gateway-queue/src/lib.rs
lemunozm: Although this new solution is more simple, I think there is a problem here:

```rust
MessageQueue::<T>::iter_keys().collect::<Vec<_>>();
```

It can collect many keys, making the block impossible to build. 

I think we need a complex structure here that allow us to store them already organized

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -621,34 +638,44 @@ pub mod pallet {
...
+
+				if res.is_ok() {
+					TransactionOutcome::Commit(Ok::<(DispatchResult, Weight), DispatchError>((
+						res, weight,
+					)))
+				} else {
+					TransactionOutcome::Rollback(Ok::<(DispatchResult, Weight), DispatchError>((
+						res, weight,
+					)))
 				}
```
lemunozm, line 676: Super NIT. I think adding the returned type in the closure allows to remove the `Ok<stuff>` here

```rust
with_transaction(|| -> DispatchResult {
```

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -349,9 +354,13 @@ impl<T: Config> Pallet<T> {
 		}
 
 		if let Some(msg) = message {
-			Self::execute_post_voting_dispatch(message_hash, router_ids, expected_proof_count)?;
+			for submessage in msg.submessages() {
```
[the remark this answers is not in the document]
lemunozm: Yeah, we should discuss this

pallets/liquidity-pools-gateway/src/message_processing.rs
```diff
@@ -333,7 +334,11 @@ impl<T: Config> Pallet<T> {
 				// we can return.
 				None => return Ok(()),
 				Some(stored_inbound_entry) => match stored_inbound_entry {
-					InboundEntry::Message(message_entry) => message = Some(message_entry.message),
+					InboundEntry::Message(message_entry)
+						if message_entry.session_id == session_id =>
```
lemunozm, line 338: Question: If `session_id` is different, should we remove the entry?

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: I'm more on the way of: If we do not really need this now, let's keep things simple

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -621,34 +638,44 @@ pub mod pallet {
...
+							message,
+							router_id,
+							&mut counter,
+						);
 
-					(res, weight)
-				}
-				GatewayMessage::Outbound { message, router_id } => {
-					let res = T::MessageSender::send(router_id, T::Sender::get(), message);
+						(res, Self::get_weight_for_batch_messages(counter))
```
lemunozm, line 659: Question: This can also be simplified if no inbound messages are supported right?

lemunozm (commented):
Thanks for all the fixed! 

BTW, not sure if we should move this to the internal repo, to not forking both `main` branches too much.

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -621,34 +638,44 @@ pub mod pallet {
...
-					domain_address,
-					message,
-					router_id,
-				} => {
-					let mut counter = 0;
-
-					let res = Self::process_inbound_message(
+			// The #[transactional] macro only works for functions that return a
+			// `DispatchResult` therefore, we need to manually add this here.
+			with_transaction(|| {
```
lemunozm, line 643: We do not need this if we choose to remove the inbound batch support, right?

pallets/liquidity-pools-gateway/src/lib.rs
wischli: I'm also in favor of removing it and adding back later. Less maintenance in between.

pallets/liquidity-pools-gateway/src/message_processing.rs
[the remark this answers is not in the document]
wischli: Agreeing with @lemunozm

pallets/liquidity-pools-gateway/src/message_processing.rs
wischli: We should not because then it's impossible to replay the message and funds are stuck on the EVM side. Keeping entries for an old session id can be made unstuck via `execute_message_recovery`.

wischli: > Question: If `session_id` is different, should we remove the entry?

We should not because then it's impossible to replay the message and funds are stuck on the EVM side. Keeping entries for an old session id can be made unstuck via `execute_message_recovery`.

pallets/liquidity-pools-gateway-queue/src/lib.rs
cdamian: > It can collect many keys, making the block impossible to build.

Can you elaborate on this please?

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: We do, we still need to rollback storage in case of a failure.

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: `execute_message_recovery` will only increase the proof count for a specific router ID, we will still hit this logic and be unable to execute a message from an older session. Maybe we should extend `execute_message_recovery` and either:
- set the session of a message entry to the current one;
- increase the proof count - current behavior;

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: That won't work given:
```
pub fn with_transaction<T, E, F>(f: F) -> Result<T, E>
where
	E: From<DispatchError>,
	F: FnOnce() -> TransactionOutcome<Result<T, E>>, // this
{
```

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Cool, was asking because I wasn't sure if Solidity is already working on this or not. Will remove, thank you!

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Yes, we can always return the defensive weight.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: But given that we will always return `LP_DEFENSIVE_WEIGHT` for both inbound/outbound, we can change this a bit. I'll ping you when done.

pallets/liquidity-pools-gateway-queue/src/lib.rs
cdamian: How about limiting the number of keys that we collect via:
```
let mut nonces = MessageQueue::<T>::iter_keys().take(MAX_MESSAGES_PER_BLOCK).collect::<Vec<_>>();
```

pallets/liquidity-pools-gateway/src/lib.rs
lemunozm: That's right!

pallets/liquidity-pools-gateway-queue/src/lib.rs
lemunozm: > Can you elaborate on this please?

When you collect, the iterator will make one read per item, and could be a number of items that overpass the limit for the block weight capacity.

The `take(MAX_MESSAGES_PER_BLOCK)` still does not work because could be left a message in the queue that ideally should be processed first.

lemunozm: Not sure if exists some order structure available in substrate for this. If not, we should create some complex/annoying structure to organize the way the messages are stored.

But I'm not able to see a super simple way TBH. We can leave that fix for another PR to unlock this if we see it's not easy

cdamian: Maybe there's a simpler solution that involves using the latest message nonce. I'll try something on a different branch.

cdamian: Something similar to - https://github.com/centrifuge/centrifuge-chain/pull/1992

pallets/liquidity-pools-gateway-queue/src/lib.rs
```diff
@@ -93,6 +98,11 @@ pub mod pallet {
 			message: T::Message,
 			error: DispatchError,
 		},
+
+		/// Maximum number of messages was reached.
+		MaxNumberOfMessagesWasReached {
```
wischli, line 103: Nit: Would remove `Was`  but absolutely non-blocking!

```suggestion
		MaxNumberOfMessagesReached {
```

pallets/liquidity-pools-gateway/src/lib.rs
```diff
@@ -621,45 +624,33 @@ pub mod pallet {
...
-				GatewayMessage::Inbound {
-					domain_address,
-					message,
-					router_id,
-				} => {
-					let mut counter = 0;
-
-					let res = Self::process_inbound_message(
+			// The #[transactional] macro only works for functions that return a
+			// `DispatchResult` therefore, we need to manually add this here.
```
wischli, line 628: Q: Any reason for going with the overhead of `with_transaction` and branching over the result instead of just using `#[transactional]` and returning `Ok(())`?

wischli (approved):
LGTM, thanks for tackling all critical queue related findings! Non-blocking nit and question.

mustermeiszer (approved):
Thanks for tackling this!!

pallets/liquidity-pools-gateway-queue/src/lib.rs
cdamian: Will change.

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: Not sure what you mean here - do you mean changing the return of `process` to `DispatchResult`?

pallets/liquidity-pools-gateway/src/message_processing.rs
cdamian: Bumping this again since I think we might need to add the 1st point that I mentioned above.

cc @hieronx

pallets/liquidity-pools-gateway/src/lib.rs
cdamian: We can't use the `#[transactional]` macro in this function, that is not supported. We can either change the signature as I suggested in my previous message, or go with the current approach.

PR #1992, opened 2024-09-20
lp-gateway-queue: Ensure messages are processed in order

cdamian:
Addressing the comment regarding LP gateway queue message processing:
https://github.com/centrifuge/centrifuge-chain/pull/1991#discussion_r1766659178

### DIFF
--- a/pallets/liquidity-pools-gateway-queue/src/lib.rs
+++ b/pallets/liquidity-pools-gateway-queue/src/lib.rs
@@ -100,7 +100,9 @@ pub mod pallet {
 		},
 
 		/// Maximum number of messages was reached.
-		MaxNumberOfMessagesWasReached,
+		MaxNumberOfMessagesWasReached {
+			last_processed_nonce: T::MessageNonce,
+		},
 	}
 
 	#[pallet::error]
@@ -214,8 +216,10 @@ pub mod pallet {
 			let mut weight_used = T::DbWeight::get().reads(1);
 
 			loop {
-				if let Err(_) = last_processed_nonce.ensure_add_assign(One::one()) {
-					Self::deposit_event(Event::<T>::MaxNumberOfMessagesWasReached);
+				if last_processed_nonce.ensure_add_assign(One::one()).is_err() {
+					Self::deposit_event(Event::<T>::MaxNumberOfMessagesWasReached {
+						last_processed_nonce,
+					});
 
 					break;
 				}
@@ -232,8 +236,15 @@ pub mod pallet {
 
 				let message = match MessageQueue::<T>::get(last_processed_nonce) {
 					Some(msg) => msg,
-					// No message found, we can stop.
-					None => break,
+					// No message found at this nonce, we can skip it.
+					None => {
+						LastProcessedNonce::<T>::set(last_processed_nonce);
+
+						// 1 write for setting the last processed nonce
+						weight_used.saturating_accrue(T::DbWeight::get().writes(1));
+
+						continue;
+					}
 				};
 
 				let remaining_weight = max_weight.saturating_sub(weight_used);
@@ -244,7 +255,7 @@ pub mod pallet {
 					break;
 				}
 
-				let weight = match Self::process_message_and_deposit_event(
+				let processing_weight = match Self::process_message_and_deposit_event(
 					last_processed_nonce,
 					message.clone(),
 				) {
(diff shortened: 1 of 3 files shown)


pallets/liquidity-pools-gateway-queue/src/lib.rs
```diff
@@ -231,10 +259,12 @@ pub mod pallet {
 
 				weight_used.saturating_accrue(weight);
 
-				MessageQueue::<T>::remove(nonce);
+				MessageQueue::<T>::remove(last_processed_nonce);
 
 				// 1 write for removing the message
 				weight_used.saturating_accrue(T::DbWeight::get().writes(1));
```
wischli, line 265: Nit: We need to add another write: `MessageQueue::remove` + `LastProcessedNonce::set`.

pallets/liquidity-pools-gateway-queue/src/lib.rs
```diff
@@ -200,17 +208,34 @@ pub mod pallet {
...
 
 		fn service_message_queue(max_weight: Weight) -> Weight {
-			let mut weight_used = Weight::zero();
+			let mut last_processed_nonce = LastProcessedNonce::<T>::get();
+
+			// 1 read for the last processed nonce
+			let mut weight_used = T::DbWeight::get().reads(1);
+
+			loop {
+				if let Err(_) = last_processed_nonce.ensure_add_assign(One::one()) {
```
wischli, line 217: Nit: 
```suggestion
			loop {
				if last_processed_nonce.ensure_add_assign(One::one()).is_err() {
```

pallets/liquidity-pools-gateway-queue/src/lib.rs
```diff
@@ -93,6 +98,9 @@ pub mod pallet {
 			message: T::Message,
 			error: DispatchError,
 		},
+
+		/// Maximum number of messages was reached.
+		MaxNumberOfMessagesWasReached,
```
wischli, line 103: Nit: Could add the current last processed nonce to the event payload for more info

wischli (commented):
Seems to be an appropriate solution to the comment

pallets/liquidity-pools-gateway-queue/src/lib.rs
```diff
@@ -200,17 +208,34 @@ pub mod pallet {
...
+				}
 
+				// 1 read for the message
 				weight_used.saturating_accrue(T::DbWeight::get().reads(1));
 
+				let message = match MessageQueue::<T>::get(last_processed_nonce) {
+					Some(msg) => msg,
+					// No message found, we can stop.
+					None => break,
+				};
```
lemunozm, line 237: I think the idea is fine! But here we could have an undesired behavior:

A message could be processed with priority through `process_message()` and consequently be extracted from the `MessageQueue`. If that happens, then `last_processed_nonce` could not be found, but that doesn't mean that we want to end up iterating the message queue. TLDR: Could be holes in the `MessageQueue` storage.

We should somehow jump over the holes and continue iterating until reaching the maximum nonce added to the `MessageQueue`

lemunozm: Maybe another storage value called `LastKnownNonce` or similar does the trick

cdamian: `MessageNonceStore` contains the last inserted nonce, I will adapt this as per your comment, I forgot about `process_message` when tackling this. Thank you!

lemunozm: This is one of the whys why having that method increase complexity 😅

runtime/integration-tests/src/cases/lp/utils.rs
```diff
@@ -86,15 +86,14 @@ pub fn pool_c_tranche_1_id<T: Runtime>() -> TrancheId {
...
-	let (_tx, status, receipt) = pallet_ethereum::Pending::<T>::get()
+	let (_tx, status, _receipt) = pallet_ethereum::Pending::<T>::get()
 		.last()
 		.expect("Queue triggered evm tx.")
 		.clone();
 
 	// The sender is the sender account on the gateway
 	assert_eq!(T::Sender::get().h160(), status.from);
 	assert_eq!(status.to.unwrap().0, to.0);
-	assert!(!receipt_ok(receipt));
```
cdamian, line 97: Note - given that the message processor implementation of the LP gateway is transaction, this check is no longer required.

pallets/liquidity-pools-gateway-queue/src/lib.rs
```diff
@@ -200,17 +210,43 @@ pub mod pallet {
...
+				let message = match MessageQueue::<T>::get(last_processed_nonce) {
+					Some(msg) => msg,
+					// No message found at this nonce, we can skip it.
+					None => {
+						LastProcessedNonce::<T>::set(last_processed_nonce);
+
+						// 1 write for setting the last processed nonce
+						weight_used.saturating_accrue(T::DbWeight::get().writes(1));
+
+						continue;
```
lemunozm, line 246: Nice! I think this fixes the issue of having "holes" in the queue.

lemunozm (approved):
Thanks Cosmin, I think now it works fine in all scenarios!

PR #2017, opened 2025-05-22
chore: post v0.15.0 cleanup

wischli:
# Description

* Bumps to next version v0.16.0
* Removes executed migrations
* Reverts disabled Docs CI which was pushed from downstream

# Checklist:

- [x] I have added Rust doc comments to structs, enums, traits and functions
- [x] I have made corresponding changes to the documentation
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or that my feature works

PR #2033, opened 2026-03-04
Migration 1507: Force transfer Polkadot USDC + set LPAdmin as fast tracker

wischli:
## Summary
- Includeds all changes of #2017, see here for diff: https://github.com/centrifuge/centrifuge-chain/compare/chore/post-1505-cleanup...migration-1506/force-transfer-usdc
- Consolidate all USDC (`ForeignAsset(6)`) balances from 12 holders into a single account (`4dhWqvsRVE8urtAcn3RkbT1oJnqFktGF1abfvuhyC8Z13Lnd`) via runtime migration, including clearing locks and reserves
- Grant the `LpAdminAccount` (`4eEqmbQMbFfNUg6bQnqi9zgUvQvSpNbUgstEM64Xq9FW58Xv`) the ability to propose and instantly execute referenda, enabling fast runtime upgrades without full governance cycles
- Enable instant referenda (`InstantAllowed = true`)

These changes prepare for the full deprecation of Centrifuge Chain as part of the CFG token migration to EVM, approved by governance in [CP149](https://github.com/centrifuge/cps/blob/main/cps/CP149/CP149.md).

### USDC Consolidation Migration

All USDC (`ForeignAsset(6)`) is swept from every holder into a single Centrifuge-controlled account so it can be offramped to Polkadot AssetHub before the chain is bricked. The migration:

1. Iterates all `orml_tokens::Accounts` entries for `ForeignAsset(6)`
2. Removes all non-destination entries (accounts, locks, reserves)
3. Credits the total (free + reserved) to the destination as free balance

### Instant Referenda via LpAdminAccount

To allow fast execution of the subsequent chain-killing runtime upgrade, the `LpAdminAccount` is granted the following democracy origins:

| Origin | Effect |
|---|---|
| `ExternalDefaultOrigin` | Can propose external default referenda (negative turnout bias) |
| `FastTrackOrigin` | Can fast-track referenda to a short voting period |
| `InstantOrigin` | Can trigger instant referendum execution |

Combined with `InstantAllowed = true`, this lets the `LpAdminAccount` push through a runtime upgrade in a single block.

### How to execute a fast referendum after this upgrade

1. **Propose**: `LpAdminAccount` submits `democracy.externalProposeDefault(encodedCall)` where `encodedCall` is the `system.setCode(wasm)` or `parachainSystem.authorizeUpgrade(codeHash)` call
2. **Fast-track**: `LpAdminAccount` submits `democracy.fastTrack(refIndex, votingPeriod=1, delay=0)` — the `instant=true` flag allows `delay=0`
3. **Vote**: `LpAdminAccount` votes aye on the referendum
4. **Enact**: The referendum passes and executes as Root in the next block

## Test plan

- [x] Compiles with `--release` and `--features try-runtime`
- [x] Clippy passes
- [x] Run `try-runtime on-runtime-upgrade --checks=all` against a Centrifuge state snapshot
- [x] Verify USDC consolidation: all balances moved to destination, no other holders remain
- [x] Verify `LpAdminAccount` can propose + fast-track + instantly enact a referendum on a local testnet


runtime/altair/src/lib.rs
```diff
@@ -2015,7 +2015,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	migrations::UpgradeAltair1505,
+	migrations::UpgradeAltair1506,
```
lemunozm, line 2018: FYI: spec_version is 1507 but Migration is 1506

runtime/common/src/migrations/consolidate_usdc.rs
```diff
@@ -0,0 +1,165 @@
...
+					}
+				})
+				.collect();
+
+		for (account, data) in holders {
+			let free: u128 = data.free.into();
+			let reserved: u128 = data.reserved.into();
+			let amount = free.saturating_add(reserved);
+
+			total_moved.saturating_accrue(amount);
```
lemunozm, line 61: We're not adding `data.frozen`. I understand we don't have any frozen amount

lemunozm (approved):
LGTM, just a couple of comments to double-check

runtime/common/src/migrations/consolidate_usdc.rs
wischli: Yes, exactly. Thanks for asking!

runtime/altair/src/lib.rs
wischli: Since the WASM is already built and this has no impact, I'd keep it as is. We want to teardown Altair soon anyways with #2034 where we have v0.16.0

